PR #19, opened 2014-11-16
Fix show previous version

JoeCohen:
Fixes [Pivotal Issue 12252025](https://www.pivotaltracker.com/story/show/12252025): 
Bug: "Show Previous Version" goes to current version instead of previous version
* Wrote test to find bug, placing test in glossary_controller_test.rb
* Fixed the bug by making trivial correction to application_helper.rb
* Refactored glossary_controller_test.rb to fix RuboCop violations


mo-nathan:
What you've done is very good although technically GlossaryControllerShowAndIndexTest should got in its own file. Generally tests get a bit of a pass regarding things like line count for the class although breaking up the methods into smaller functions is definitely goodness.

PR #23, opened 2015-01-16
First draft of sitemap.

pellaea:
This is one simple proposal of how a sitemap might be used to reduce robot traffic.  It:

1) bans all robots from the entire site
2) gives the robots a sitemap listing these pages:
* activity log, intro, how_to_use, how_to_help (each in all our non-beta languages, and tells robots to index weekly)
* all show_object pages (in English only, and provides a last_modified timestamp to let robots know when they need to refresh their index based on the object itself as well as comments and images)

I have for the moment included all major object types, even though Nathan only wanted names included.  It is trivial to add or remove object types the way I've written it.  I see no reason personally not to include all the object types I've added so far, but I don't have a strong opinion on the subject.  And we could potentially use traffic statistics to determine which objects to allow robots to index.

There is some question yet about whether robots will be able to access the pages in the site index if we've banned robots from the whole site.  I'm not even sure how to test whether this works.

The problem, as I understand it, is that robots.txt is extremely clunky.  It can only disallow -- not *allow* -- and it can only disallow based on the first half of the request url.  There is no way to disallow all /observer/* except for /observer/show_observation.  One must explicitly disallow every single action.  Yuck.

An alternative to consider (if the present version does not work) is to allow only a small set of fake "root" urls, and reroute them inside nginx appropriately:

/activity_logs -> /
/how_to_use -> /observer/how_to_use
/show_observation/id -> /id
/show_name/id -> /name/show_name/id

Then robots.txt might look like:

Disallow: /observer/
Disallow: /name/
etc.

This at least reduces it to simply listing all the controllers, a much more easily scaled and maintained solution.  (And this would also ensure that images would be accessible, another questionable point with the present proposal.)


pellaea:
New information: I can now say with reasonable confidence that search engines do not like a conflict between robots.txt and sitemap.xml.  (I did *a lot* of googling on the subect!)  Therefore we definitely need to find a way to block everything *but* a few urls in robots.txt.

One useful trick I just found that will help tremendously:

Disallow: *?*

This blocks all requests with query parameters.  But we'll still need to block every action *but* a handful.  I propose that we have refresh_sitemap scour our controllers and place all action-like method names in robots.txt.

Of course, now there's also a new problem, thanks to Rails not being able to choose a standard and stick to it: there is confusion between the versions of Rails whether the id is passed in as:

/obs/show_obs/1234
or
/obs/show_obs?id=1234

And it seems likely now that even with a sitemap, search engines still need to be able to get to the urls in the sitemap by crawling.

One solution for both simultaneously would be to have the root index (list_rss_logs) give robots just a huge long list of all our objects that we want search engines to index.  Or better yet stick a link in list_rss_logs which points robots at such a page.  The beauty is, we can make this page static and have refresh_sitemap create it for us, so the configuration of which pages get indexed remains all in one place (refresh_sitemap).

raysuelzer:
Jason --

I don't have a strong opinion on this either way, you seem to have done a
lot of research on this and trust your judgement.  We definitely want to
balance keeping up organic search traffic (40% of our traffic) by having
Mushroom Observer come up on google while making sure that the servers are
not being needlessly crushed.

But, I'm curious as to if the bots are a major problem for us.  I have
noticed that since Nathan made the COUNT(*) fix that the site loads
instantaneously for me from DC.  I ran a few tests on pingdom and the
average time to serve the initial HTML page is ~200ms.  50ms of which are
spent are spent on resolving DNS. Meaning it takes less than 150ms to
process the request and send it back over the wire... That is really good
for a Rails application on a single server! I have read on some Ruby blogs
that unless you use something like JRuby it's hard to get below 100ms on
any sort of complicated view.

Maybe it would make sense to set a cron job to test the server for a period
of 72 hours and gauge response times and see if we can notice any trends.
I'd be willing to write a script to do that.

Another thing that we could do is monitor the API usage.  If we do have
people running chron jobs against the API, I'd think that could be a
performance hog.

-Ray



Ray Suelzer
Mobile: (312) 869-2665

pellaea:
My stats differ from yours dramatically.  I found 90% of our traffic to be
robots, not 40%.  And I find several seconds response time on average when
looking at the production logs.  And that was after Nathan did his magic.

The other piece of information -- if robots constituting 90% of our traffic
isn't enough! -- that convinces me is that when Alan blocked all robots for
a few minutes, response time reportedly was much faster.  (I don't recall
the figures given, sorry.)

Besides, this pull request kind of assumes that we want to control robots
better as a basic assumption.  Nathan's proposal earlier today was simply
to shunt them to another server using a stale, lagging copy of the
database.  I believe this solution will be nearly as effective, and will
obviate the need for extra servers and the complications that will result
from that.  If robots truly aren't a problem, then we don't need to do
either.  But I am personally convinced that they are.

raysuelzer:
I mean that 40% of the real humans that visit MushroomObserver.org come to
the site through a search engine.  I don't doubt that 90% of the actual
server traffic is robots, but that won't show up on google analytics since
they don't have javascript.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
I have finished making the proposed updates.  This should now be ready to go.  In principle it will deny robot access to all of the site except three static pages, the first page of list_rss_logs, and all of the show_object pages.  And it will tell robots very precisely when they need to re-index all of the show_object pages.

mc-nathan:
Like Jason, I'm not seeing the amazing times that Ray reports.  According to this page:

https://www.google.com/analytics/web/?authuser=1#report/content-site-speed-overview/a1916187w3391904p3473187/

we had a nasty spike Wednesday and Thursday with average page load times over 25 seconds.

I like the approach that Jason describes of providing specific pages for the robots to checkout.  The only problem I see with that is that it may not let Google correlate links in other sites with the pages it indexes.  If that's true then it may severely impact the page rank of those referred to pages.  However, to really get our head around that we should look at what searches that 40% of users are using to find MO and what their landing pages are.  All of that info is extractable from Google Analytics.  Before getting into that I want to review Jason's commit so I better understand what has been implemented.

pellaea:
Interesting point about correlates.  For example show_observation has at
least five versions(!)

/1234
/observer/show_observation/1234
/observer/show_observation?id=1234
/observer/lookup_observation/1234
/observer/lookup_observation?id=1234

And that doesn't include the q=xxx versions we often see in external links.

But Nathan only cares about names, but that, too, has problems:

/name/show_name/1234
/name/show_name?id=1234
/observer/lookup_name/1234
/observer/lookup_name?id=1234
/observer/lookup_name/Pleurotus_populinus
/observer/lookup_name?id=Pleurotus_populinus

I wonder if there is some magic header we can set that will tell google how
to index a given page?  Shoot, but that won't solve the problem.  We need
to somehow tell google that all these variations point to the same page.  I
don't know how to do that. :(

(Note that lookup_xxx redirects to the show_xxx page, not sure if that
helps, though.)

-Jason

raysuelzer:
Probably just send a 301 permanently moved redirect to the URL we want to
have indexed.

https://support.google.com/webmasters/answer/93633?hl=en




Ray Suelzer
Mobile: (312) 869-2665

pellaea:
But how does google know that the external link redirects until it tries to
follow it?  Isn't that the whole point? Need to do some googling to
understand this issue better.

Regardless, I think it doesn't affect this PR, though, since the end-points
in the "fake" robot index pages are the same as they always have been.  The
only thing novel are the fake indexes themselves.  And there's nothing of
any value on those pages, so I can't imagine anyone accidentally seeing one
of those pages as a search result.  (There may even be a meta tag to allow
crawling but prevent indexing, that would solve even this minor problem.)

raysuelzer:
Also this: https://support.google.com/webmasters/answer/139066?hl=en

It how we keep our URLs restful on search engines where I work.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
Do the logs give us a sense of which bots are most problematic?  I see with
Google Webmaster tools you can set some specific crawling options and
reduce the crawl rate.

Ray Suelzer
Mobile: (312) 869-2665

script/refresh_sitemap
```diff
@@ -0,0 +1,353 @@
...
+  return controller, action
+end
+
+def get_list_of_controllers
+  # We could use ApplicationController.descendants, but controllers are loaded
+  # lazily, so there won't actually be any loaded yet.  This is safer.
+  result = []
+  wildcard = CONTROLLER_FILE.sub("NAME", "*")
+  for file in Dir.glob(wildcard)
+    result << match[1] if match = file.match(/(\w+)_controller.rb/)
```
mo-nathan: Wanted to create an example sitemap to check out, so I ran:

$ script/refresh_sitemap 

and got:

script/refresh_sitemap:325:in `block in get_list_of_controllers': undefined local variable or method `match' for main:Object (NameError)
	from script/refresh_sitemap:324:in `each'
	from script/refresh_sitemap:324:in `get_list_of_controllers'
	from script/refresh_sitemap:286:in `update_robots_file'
	from script/refresh_sitemap:351:in `<main>'

pellaea: Arg.  I made that tweak to avoid using $1 -- which I know you hate :) --
but never tested it!

Works now.  Just checked in the fix.

Note, however, that you don't actually get much of anything if you run it
on the base VM database.  I recommend grabbing a snapshot first.

-Jason

pellaea:
re: "link rel=canonical"

That's perfect!  Actually, the sitemap should be enough acc. to the docs at
the link you shared.  But I think it would be worthwhile setting the
canonical url explicitly for all the show_xxx pages, too.  I'll add that as
a pivotal story.

Thanks, Ray!

JoeCohen:
Can the routing changes be separated from the site map improvements?  (Would make it easier to isolate any problems.)

pellaea:
Sure, Ray had just asked me to push it out at the same time.  I'll make a
separate pr for it.

PR #26, opened 2015-01-19
Removed excess whitespace from xml and html files.

pellaea:


raysuelzer:
I just put in my pull request.  I think I finally figured out how to git it
to work using the RubyMine git interface....

Ray Suelzer
Mobile: (312) 869-2665

PR #29, opened 2015-01-21
Added before_filter to kick out robots that go where they aren't allowed.

pellaea:
Also:

Added timing stats at end of catch_errors around_filter.
Added script/refresh_sitemap to crontab.

### DIFF
--- a/test/functional/observer_controller_test.rb
+++ b/test/functional/observer_controller_test.rb
@@ -2446,4 +2446,19 @@ class ObserverControllerTest < FunctionalTestCase
     assert_no_flash
     assert_response(:success)
   end
+
+  def test_normal_permissions
+    get :intro
+    assert_equal(200, @response.status)
+    get :textile
+    assert_equal(200, @response.status)
+  end
+
+  def test_robot_permissions
+    @request.user_agent = "Googlebot"
+    get :intro
+    assert_equal(200, @response.status)
+    get :textile
+    assert_equal(403, @response.status)
+  end
 end
--- a/app/classes/robots.rb
+++ b/app/classes/robots.rb
@@ -8,19 +8,21 @@ class Robots
     end
 
     def populate_allowed_robot_actions
-      file = "#{::Rails.root}/public/robots.txt"
+      file = MO.robots_dot_text_file
       @@allowed_robot_actions = parse_robots_dot_text(file)
     end
 
     def parse_robots_dot_text(file)
       results = {}
-      pat = Regexp.new('Allow: /(\w+)/(\w+)')
-      File.open(file).readlines.each do |line|
-        match = line.match(pat)
-        if match
-          controller = match[1]
-          action     = match[2]
-          results["#{controller}/#{action}"] = true
+      if File.exist?(file)
+        pat = Regexp.new('Allow: /(\w+)/(\w+)')
+        File.open(file).readlines.each do |line|
+          match = line.match(pat)
+          if match
+            controller = match[1]
+            action     = match[2]
+            results["#{controller}/#{action}"] = true
+          end
         end
       end
       return results
--- a/app/classes/robots.rb
+++ b/app/classes/robots.rb
@@ -1,0 +1,29 @@
+# encoding: utf-8
+
+class Robots
+  class << self
+    def allowed?(controller, action)
+      populate_allowed_robot_actions if !defined?(@@allowed_robot_actions)
+      return @@allowed_robot_actions["#{controller}/#{action}"]
+    end
+
+    def populate_allowed_robot_actions
+      file = "#{::Rails.root}/public/robots.txt"
+      @@allowed_robot_actions = parse_robots_dot_text(file)
+    end
+
+    def parse_robots_dot_text(file)
+      results = {}
+      pat = Regexp.new('Allow: /(\w+)/(\w+)')
+      File.open(file).readlines.each do |line|
+        match = line.match(pat)
+        if match
+          controller = match[1]
+          action     = match[2]
+          results["#{controller}/#{action}"] = true
+        end
+      end
+      return results
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


raysuelzer:
Cool!  I've made some significant headway on the layout and views. I think
I've been able to remove a large amount of code from application helper
that was generating HTML by creating an rss_log_item partial view that
renders the needed HTML for those rss boxes.  I've put image thumbnails
into a partial view, and I've place the Ajax vote bar as a partial view.
In total it's about 500 lines that can be reduced in application helper.
I'm no where near having it read for a commit, but my progress is going
very well and the result is a design pattern much simpler  and will be a
breeze to test in just a few lines of code.

I made a new helper called rssloghelper which moved all of the logic on
translating objects into RSS feed items out of the view. I think it
represents a better separation of concerns and will make it easy to unit
test.

I'm also going to probably move what remains of image helper stuff that is
in the application module into an image helper module to get that file down
to a more manageable size.

The whole process is going to be far more complicated than I anticipated
because of how much HTML is generated in the application helper.

I am definitely making major changes, but I will leave the code better than
I found it, at least to my sensibilities.

On Wednesday, January 21, 2015, Jason Hollinger <notifications@github.com>
wrote:

> Also:
>
> Added timing stats at end of catch_errors around_filter.
> Added script/refresh_sitemap to crontab.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/MushroomObserver/mushroom-observer/pull/29
> Commit Summary
>
>    - Added before_filter to kick out robots that try to go where they
>    aren't welcome.
>
> File Changes
>
>    - *A* app/classes/robots.rb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-0>
>    (29)
>    - *M* app/controllers/application_controller.rb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-1>
>    (23)
>    - *M* config/etc/crontab
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-2>
>    (1)
>    - *M* script/refresh_sitemap
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-3>
>    (18)
>    - *A* test/fixtures/robots.txt
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-4>
>    (7)
>    - *M* test/functional/observer_controller_test.rb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-5>
>    (15)
>    - *A* test/unit/robots_test.rb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#diff-6>
>    (14)
>
> Patch Links:
>
>    - https://github.com/MushroomObserver/mushroom-observer/pull/29.patch
>    - https://github.com/MushroomObserver/mushroom-observer/pull/29.diff
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/29>.
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

test/functional/observer_controller_test.rb
```diff
@@ -2446,4 +2446,19 @@ def test_download_observation_index
     assert_no_flash
     assert_response(:success)
   end
+
+  def test_normal_permissions
+    get :intro
+    assert_equal(200, @response.status)
```
JoeCohen: Is it any better to use `assert_response` instead of `assert_equal`?  (Error messages might be more specific; with some symbols as argument to `assert response`, the assertion is more flexible.  See http://api.rubyonrails.org/classes/ActionDispatch/Assertions/ResponseAssertions.html#method-i-assert_response.)
So instead of `assert_equal(200, @response.status)`, consider
  `assert_response(200)` or
  `assert_response(:success)`  (Status code 200-299)
And instead of  `assert_equal(403, @response.status)`
  `assert_response(403)` or
  `assert_response(:Forbidden)`  (same effect, but intent might be more transparent)

raysuelzer: That sounds reasonable to me an probably why they included those helpers.

Ie. a post should technically return a 201 on success.

On Wednesday, January 21, 2015, Joseph D. Cohen <notifications@github.com>
wrote:

> In test/functional/observer_controller_test.rb
> <https://github.com/MushroomObserver/mushroom-observer/pull/29#discussion_r23332869>
> :
>
> > @@ -2446,4 +2446,19 @@ def test_download_observation_index
> >      assert_no_flash
> >      assert_response(:success)
> >    end
> > +
> > +  def test_normal_permissions
> > +    get :intro
> > +    assert_equal(200, @response.status)
>
> Is it any better to use assert_response instead of assert_equal? (Error
> messages might be more specific; with some symbols as argument to assert
> response, the assertion is more flexible. See
> http://api.rubyonrails.org/classes/ActionDispatch/Assertions/ResponseAssertions.html#method-i-assert_response.
> )
> So instead of assert_equal(200, @response.status), consider
> assert_response(200) or
> assert_response(:success) (Status code 200-299)
> And instead of assert_equal(403, @response.status)
> assert_response(403) or
> assert_response(:Forbidden) (same effect, but intent might be more
> transparent)
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/29/files#r23332869>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

mo-nathan:
Travis isn't happy with the change to robots.txt.  I think it's saying that robots can't get to the intro page which would not be a good thing.

pellaea:
Uh...  The current robots.txt blocks neither / nor /observer/intro.  What
makes you think Travis doesn't like it?

Last I checked robots are only half of our traffic now.  Google analytics
is convinced page access time has gone up. But has anyone actually used the
site??  It's way faster.  What gives?  Even average time in rails has gone
down.

-Jason

pellaea:
Oh, I see, it has nothing to do with where robots.txt is allowing robots to
go.  It's just that robots.txt no longer exists in the repo, so it failed
to read it in a unit test.  Fixed.  Testing.  Will push out in a few
minutes.  I'm eager to start gathering proper stats about which pages take
the longest in rails.  Apparently reducing robot traffic is no longer such
high priority.  Hooray!

-Jason

PR #31, opened 2015-01-23
mock script's call to blocked page on live MO website

JoeCohen:
* cherry-picked from ror4-joe 637157

### DIFF
--- a/script/make_eol_xml
+++ b/script/make_eol_xml
@@ -10,29 +10,23 @@
 #
 ################################################################################
 
-require 'tempfile'
-require 'net/http'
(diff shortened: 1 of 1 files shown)


JoeCohen:
1. Mocked in the script rather than the test because I didn't see how to mock in the test without mocking out the entire script.  That seemed to make the test pointless.
2. Tested in ror4-joe, but not in master.  To run only the test with which this interacts:
`rake test test/unit/script_test.rb test_make_eol_xml` or
`ruby -Ilib:test test/unit/script_test.rb test_make_eol_xml`

PR #35, opened 2015-01-28
Multi-image upload wasn't blocking form submission in edit_obs form.

pellaea:


raysuelzer:
Thanks for catching this.

Ray Suelzer
Mobile: (312) 869-2665

PR #38, opened 2015-01-29
fix double render bug

JoeCohen:
* same fix as master 7be246041c27078ebb9cb052ab17e4d52c0db8f6,
but with more specific assertion

### DIFF
--- a/app/controllers/observer_controller.rb
+++ b/app/controllers/observer_controller.rb
@@ -645,7 +645,6 @@ class ObserverController < ApplicationController
       show_selected_observations(query)
     else
       flash_error(:runtime_missing.t(field: "id"))
-      redirect_to(action: "list_rss_logs")
     end
   end
 
--- a/app/controllers/observer_controller.rb
+++ b/app/controllers/observer_controller.rb
@@ -638,14 +638,10 @@ class ObserverController < ApplicationController
   end
 
   # Displays matrix of User's Observation's, by date.
-  def observations_by_user # :nologin: :norobots:
-    user = params[:id] ? find_or_goto_index(User, params[:id].to_s) : @user
-    if user
-      query = create_query(:Observation, :by_user, user: user)
-      show_selected_observations(query)
-    else
-      flash_error(:runtime_missing.t(field: "id"))
-    end
+  def observations_by_user  # :nologin: :norobots:
+    return unless user = find_or_goto_index(User, params[:id].to_s)
+    query = create_query(:Observation, :by_user, user: user)
+    show_selected_observations(query)
   end
 
   # Displays matrix of Observation's at a Location, by date.
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -172,7 +172,7 @@ class ObserverControllerTest < FunctionalTestCase
 
   def test_observations_by_unknown_user
     get(:observations_by_user, id: 1e6)
-    assert_response(:success)
+    assert_redirected_to(action: :index_user)
   end
 
   def test_altering_types_shown_by_rss_log_index


pellaea:
Ugh, I thought I fixed this already.  Dejà vu!  What you just proposed was my initial solution , too.  But on further inspection, the whole logic of this action is sloppy.  For example, I think it will still crash if no id is given and no user is logged in.  How about this instead?  Much safer and cleaner.

def observations_by_user
  user = find_or_goto_index(User, params[:id].to_s)
  if user
    query = create_query(:Observation, :by_user, user: user)
    show_selected_observations(query)
  end
end

JoeCohen:
1.  You did fix it.  But the change to observation_controller.rb didn't make it into the ror4 branch.  So I just copied that (and tweaked the assertion to make it more specific).
2.  I don't think it will crash if no id is given and no user is logged in -- the first line will set user = nil.
In any event, don't we need the `else` with the `flash_error`?

pellaea:
But if no id is given it never calls find_or_goto_index, therefore nothing
is rendered or redirected so it will try to render a nonexistent view and
crash.

I don't think the "else flash_error" is necessary or even correct.  It will
complain that "id is missing" even when you pass in an id of a nonexistent
user.  Note that find_or_goto_index flashes an error already if the id is
invalid or missing.

I can find no use of observations_by_user without an id passed in, so I see
no reason to preserve that unnecessary functionality.  Better to simplify.

(I know, I bet I lost those changes when I messed up the rebase last night
or the night before.  I know I tried to fix it better for ror4, but
obviously it never made it onto ror4!)

Sorry, don't mean to criticize here.  As I said, I did the same exact thing
the first time around.  But then I noticed the deeper issues last night,
and I think we might as well fix the whole thing while we're at it.

-Jason

JoeCohen:
1. You're right about the crashing.
2. We can combine the 1st two lines of your code to 
```
if user = find_or_goto_index(User, params[:id].to_s)
```
Even shorter:
```
def observations_by_user
  return unless user = find_or_goto_index(User, params[:id].to_s)
  query = create_query(:Observation, :by_user, user: user)
  show_selected_observations(query)
end
```
But see 3; we may want an `else`
3. Based on experiment, with: that code, nobody is logged in, and no id, I end up at
observer/observations_by_user "Observations created by Administrator"
Do we want to avoid that?
My brain is going to sleep, so I think I'm off to be.

pellaea:
Now we're talking!  Yes, I like the three-line version.  I don't really see
a problem with going to "obs by admin" page if someone misuses the page.  I
just don't want it to *crash* and send me an error report! :)

Thanks for indulging me. :)

-Jason

JoeCohen:
And thanks for setting me straight. Criticism is welcome.
Either you can make the changes yourself or I'll push another commit in the
morning.

El miércoles, 28 de enero de 2015, Jason Hollinger <notifications@github.com>
escribió:

> Now we're talking! Yes, I like the three-line version. I don't really see
> a problem with going to "obs by admin" page if someone misuses the page. I
> just don't want it to *crash* and send me an error report! :)
>
> Thanks for indulging me. :)
>
> -Jason
>
> On Wed, Jan 28, 2015 at 9:07 PM, Joseph D. Cohen <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> >
> > 1. You're right about the crashing.
> > 2. We can combine the 1st two lines of your code to
> >
> > if user = find_or_goto_index(User, params[:id].to_s)
> >
> > Even shorter:
> >
> > def observations_by_user
> > return unless user = find_or_goto_index(User, params[:id].to_s)
> > query = create_query(:Observation, :by_user, user: user)
> > show_selected_observations(query)
> > end
> >
> > But see 3; we may want an else
> > 3. Based on experiment, with: that code, nobody is logged in, and no id,
> I
> > end up at
> > observer/observations_by_user "Observations created by Administrator"
> > Do we want to avoid that?
> > My brain is going to sleep, so I think I'm off to be.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/MushroomObserver/mushroom-observer/pull/38#issuecomment-71970030>
>
> > .
> >
>
> —
> <https://github.com/MushroomObserver/mushroom-observer/pull/38#issuecomment-71970462>
> .
>


-- 
Sent from Gmail Mobile

PR #41, opened 2015-01-30
fix calling link_with_query with path Symbol

JoeCohen:
* duplicate, as independent test, one assertion of test_page_loads
* fix resulting Error
* fix additional bug in app/helpers/application_helper.rb
  (change Numeric to Integer)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1409,23 +1409,25 @@ end
     return result
   end
 
-  # Draw a thumbnail image.  It takes either an Image instance or an id.  Args:
-  # link::      a hash of {controller: , action:, id: }
+  # Draw a thumbnail image.  It takes either an Image instance or an id.
+  # Args:
+  # link::      a hash of { controller:, action:, id: }
   # size::      Size to show, default is thumbnail
   # original::  Show original file name?
   # votes::     Show vote buttons?
+  def thumbnail(image, args={}) ##TODO: Add size option
+    image = Image.find(image) if image.is_a?(Integer)
+    args[:link].is_a?(Symbol) ? link = args[:link].to_s : link = args[:link]
 
-def thumbnail(image, args={}) ##TODO: Add size option
-  if image.is_a?(Numeric)
-    image = Image.find(image)
+    render(partial: "image/image_thumbnail",
+           locals: { image:    image,
+                     link:     link,
+                     votes:    args[:votes],
+                     size:     args[:size],
+                     original: args[:original]
+           }
+    )
   end
-
-  render(:partial => 'image/image_thumbnail', :locals => {:image => image,
-                                                          :link => args[:link],
-                                                          :votes => args[:votes],
-                                                          :size => args[:size],
-                                                          :original => args[:original]})
-end
 
   # Provide the copyright for an image
   def image_copyright(image, div=true)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1416,8 +1416,8 @@ end
   # original::  Show original file name?
   # votes::     Show vote buttons?
   def thumbnail(image, args={}) ##TODO: Add size option
-    image = Image.find(image) if image.is_a?(Integer)
-    args[:link].is_a?(Symbol) ? link = args[:link].to_s : link = args[:link]
+    image = Image.find(image) unless image.is_a?(Image)
+    link = args[:link].is_a?(Symbol) ? args[:link].to_s : args[:link]
 
     render(partial: "image/image_thumbnail",
            locals: { image:    image,
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -170,6 +170,11 @@ class ObserverControllerTest < FunctionalTestCase
     assert_template(:textile_sandbox)
   end
 
+  def test_page_load_user_by_contribution
+    get_with_dump(:users_by_contribution)
+    assert_template(:users_by_contribution)
+  end
+
   def test_observations_by_unknown_user
     get(:observations_by_user, id: 1e6)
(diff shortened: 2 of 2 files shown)


JoeCohen:
This fixes two bugs which did not show up in the tests.
In order to work on a Error which did show up in the tests, I duplicated the problematic assertion and duplicated it in a stand-alone test.  That test then failed, but in a new mode:
app/helpers/application_helper.rb:124:in link_with_query was called with the link as a Symbol (rather than as a String or as a Hash of Controller:, Action:, Id:.  
```
ObserverControllerTest#test_page_load_user_by_contribution:
ActionView::Template::Error: undefined method `show_user_path' for #<#<Class:0x000000060baaf8>:0x0000000665d370>
    app/helpers/application_helper.rb:124:in `link_with_query'
    app/views/image/_image_thumbnail.html.erb:11:in `_app_views_image__image_thumbnail_html_erb___2753962239459614944_50290100'
    app/helpers/application_helper.rb:1423:in `thumbnail'
    app/views/observer/users_by_contribution.html.erb:18:in `block in _app_views_observer_users_by_contribution_html_erb___3152215231590731906_53756640'
    app/views/observer/users_by_contribution.html.erb:13:in `_app_views_observer_users_by_contribution_html_erb___3152215231590731906_53756640'
    app/controllers/application_controller.rb:218:in `catch_errors'
    test/functional_test_case.rb:25:in `get'
    test/controller_extensions.rb:113:in `get_with_dump'
    test/controllers/observer_controller_test.rb:174:in `test_page_load_user_by_contribution'
```
I elected to fix this further up the chain, by changing a Symbol to a String.
There may be better/more elegant ways of dealing with this issue.  One is to change the symbol into the expected Hash.  There are several methods for which:
(1) the comments say that an expected parameter is a hash of Controller, Method, Id, but
(2) the argument is sometimes a String or Symbol.
This works if the String can be turned into a valid url with `url_for`.  But it will cause errors further down the chain if called with a Symbol.
Maybe a better way of dealing with this is to always call with the Hash.  (I'm sure there are other possibilities.)

I also noticed a use of "Numeric" to test whether an argument might be the id of an image.
```
    image = Image.find(image) if image.is_a?(Numeric)
```
which I changed to
```
    image = Image.find(image) if image.is_a?(Integer)
```
Thinking about this some more, I believe a better way to handle this wouls be:
```
    image = Image.find(image) unless image.is_a?(Image)
```

raysuelzer:
That's my code :)  And yes the last solution is the most elegant.  Go for
it!

Ray Suelzer
Mobile: (312) 869-2665

JoeCohen:
By "last" solution do you mean 
(a) `image = Image.find(image) unless image.is_a?(Image)` 
or
(b) always calling with a Hash?

pellaea:
Sounds good to me, too.  One minor stylistic comment.  I prefer:

x = a ? b : c

over

a ? x = b : x = c

(line 1420 in new app_helper)

Makes it easier to tell you're assigning "link", and less ambiguous with resp. to operation precedence.

raysuelzer:
joe: (a)

jason: Yep. Makes sense.

Ray Suelzer
Mobile: (312) 869-2665

JoeCohen:
Thanks for the help.  I just pushed a commit with those changes.

pellaea:
Done!  I'm about to return to ror4 myself, too.

JoeCohen:
I am working on the test errors like `ActionView::Template::Error: no implicit conversion of Symbol into Integer`
I now think my method of solving 
```
undefined method `show_user_path' for #<#<Class:0x000000060baaf8>:0x0000000665d370>
    app/helpers/application_helper.rb:124:in `link_with_query'
```
was not the best.  It fixed one assertion in one test.  But there are other similar problems.  
So I think I must make sure that `params[:link]` which is passed to various MO helpers is a Hash (of a type which the Rails helpers can turn into a url).

pellaea:
Yeah, I noticed those errors were still there.  Darn.  Might be a real pain
to catch all of those usages.  Maybe not.  (I'm not 100% certain I fully
understand the problem yet...)

I just fixed another enum error (already merged into ror4).  Will work on
the "undefined method `where`" next.  Down to 7 failures and 4 errors.
Pretty good!

-Jason

JoeCohen:
Example:
ApplicationController#add_query_params
```
  def add_query_param(params, query=nil)
    if browser.bot?
      # do nothing
    elsif query
      query.save if !query.id
      params[:q] = query.id.alphabetize
    elsif @query_params
      params[:q] = @query_params[:q]
    end
    params
  end
```
If `params` is a string (e.g. "show_observation"), this will throw an error when it gets to `params[:q]=`.
This can happen, for instance, if a view gives wrong arguments, e.g., app/views/species_list/show_species_list.html.erb, says:
```
<%= if observation.thumb_image_id 
          thumbnail(observation.thumb_image, :border => 0,
            :link => :show_observation, :obs => observation.id,
            :votes => true)
         end %>
```
Eventually there is a call to 
app/helpers/application_helper.rb#ink_with_query, and then to
  app/controllers/application_controller#add_query_param
which chokes as described above.

If I change the template to make the link a hash ~~:link => :show_observation~~, `link: { action: :show_observation }`, MO is happy.

JoeCohen:
PS:  I'm thinking that the issues are probably all in newer templates, so I may be able to find all of them.

pellaea:
What confuses me is why this ever worked.  But, regardless, if there prove
to be prohibitively many places to change -- it might affect things other
than thumbnail, e.g. tabs, which may be much harder to search for -- you
might also just add a line at the top of add_query_params:

params = { :action => params } if params.is_a?(Symbol)

Rails does this sort of thing all over the place in their code, so we'd at
least be in good company... :)

-Jason

JoeCohen:
Thanks for the suggestion.  I'll see what I can do to find the places to change.

PR #67, opened 2015-02-04
Added [brakeman](http://brakemanscanner.org/) static security analyzer

JoeCohen:
To see a sample report, **download** the following file and open in a browser.
(If you follow the link instead of downloading, GoogleDocs will show show you the source html instead of rendering the html.)
https://drive.google.com/file/d/0B3HuZZ4W6W1wamVPOVplUHRJYWc/view?usp=sharing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,6 @@ gem "coveralls", require: false
 # Use rubocop for code style quality control
 gem "rubocop", require: false
 
-
-
-
+# Brakeman static analysis security scanner
+# See http://brakemanscanner.org/
+gem "brakeman", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,6 +31,16 @@ GEM
     astrolabe (1.3.0)
       parser (>= 2.2.0.pre.3, < 3.0)
     blankslate (3.1.3)
+    brakeman (3.0.1)
+      erubis (~> 2.6)
+      fastercsv (~> 1.5)
+      haml (>= 3.0, < 5.0)
+      highline (~> 1.6.20)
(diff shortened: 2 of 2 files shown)


pellaea:
Yikes, that's a dismal report!

JoeCohen:
1.  There may be many false positives there.
2.  Is there a way to add these reports as part of our CI?  See http://brakemanscanner.org/docs/jenkins/

raysuelzer:
Only hesitation I would have on including it on the CI is that then those
exploits are more public.  Of course, it doesn't really matter since all of
the source code is there so anyone could run the scan.

Ray Suelzer
Mobile: (312) 869-2665

JoeCohen:
That is an issue, but -- as you say -- anyone can run the report.
Hopefully, we can get rid of the security issues.  At that point the report
will let us know if code changes introduce any new problems.

pellaea:
Many of the bugs are very real, not false positives, things I would
definitely want to clean up given time and energy.  But...  I don't think
they are serious threats.  At best someone could inject additional
parameters which might mess up one of their own observations, or more
likely result in a 500 error (I see lots of those every day from hacker
bots trying their best to find security holes like these).  But ultimately
it's absolutely impossible for someone to delete or modify someone else's
work, we've been very careful about that, and instituted safeguards at the
highest levels to ensure it.  Spam can obviously still get through, too,
and we see that frequently as well.  (Although in practice my latest
"suspicious user" notification has been very effective at helping me catch
and delete them before anyone else notices them.)

Summary: I'm mortified at how sloppy our code is... but I'm not worried
that there's any real security risk.

That's just my two cents.

PR #68, opened 2015-02-04
Turn on CSRF protection

JoeCohen:
Also deleted patch of Rails method `handle_unverified_request`
I don't know why forgery protection was not turned on in ror4.  Maybe I misunderstood the documentation and thought it was on by default.  
Questions:
1.  Will raising an exception and deleting our patch of handle_unverified_request work with the API? It would be helpful if @raysuelzer and @pellaea could review the notes at http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html to see if the changes will work with the API.
2.  Do we use persistent cookies to store user information?  See http://guides.rubyonrails.org/security.html#csrf-countermeasures

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,6 +94,10 @@ class ApplicationController < ActionController::Base
   require "csv"
   include LoginSystem
 
+# Prevent CSRF attacks by raising an exception.
+# For APIs, you may want to use :null_session instead.
+  protect_from_forgery with: :exception
+
   around_filter :catch_errors # if Rails.env == "test"
   before_filter :block_ip_addresses
   before_filter :kick_out_robots
@@ -143,16 +147,6 @@ class ApplicationController < ActionController::Base
     else
       block_given? ? yield(result) : result
     end
-  end
-
-  # The default CSRF handler silently resets the session.  The problem is
-  # autologin will circumvent this, so we would need to disable autologin
-  # temporarily.  Or we can just make forgeries fail, but leave valid requests
-  # alone.  This seems much more graceful... and it lets the user know why they
-  # are experiencing otherwise bewildering and incorrect behavior.
-  def handle_unverified_request
-    render(text: "Cross-site Request Forgery detected!", layout: false)
-    return false
   end
 
   # Physically eject robots unless they're looking at accepted pages.


raysuelzer:
This will probably break the ajax controllers.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Can probably (and probably should!) turn of csrf for api.

Cookies are used to store the user_id (for autologin, see also session
below), and as a hacky way to report the user's timezone back to the server.

We mostly use the session (stored in cookie, I think, because it has the
best performance, despite being limited in size) for the user id.  But also
a couple other non-sensitive things:

:notice -- The flash_notice/warning/error message.  (I've never liked how
this mechanism works, despite it apparently being a golden standard in
rails.  It's not RESTful and results in error message ending up displaying
in the wrong tab frequently.  Yuck.)

'return-to' -- The "back to" address in some cases, an old throwback to MO
days before even I signed on.  You've probably all seen the
"store_location" at the top of a lot of actions.  It stores "safe" fallback
points periodically so that after, for example, logging in, it knows where
to go afterward.  In general it was a bad idea, and if you use multi-tab
browsing, it frequently results in confusing behavior.  It's not RESTful, I
really need say no more, I think.

:checklist_source -- May be obsolete.  Used to have something to do with
prepopulating list of names in create_species_list.

:js_override -- I think this mechanism is now broken, and unused, too.
Used to be able to override the auto-detecction of whether you had
js-capable browser or not.

:pattern -- Last search pattern.  I'm okay with this use: it just allows
page to keep the last search pattern in the search bar from page to page as
a handy default.

:theme, :layout -- Don't remember what if anything these are used for.

:admin -- Turns on "admin" mode.  Quick and dirty, only affects Nathan and
me, I'm fine with it.

:hide_thumbnail_maps, :thumbnail_size -- Allows users who aren't logged in
to change thumbnail size (and hide maps?).  This seems entirely appropriate
and harmless.

-Jason

JoeCohen:
1.  Which will probably break the ajax controller:  
(a) forgery protection in general,
(b) forgery protection with: :exception,
(c) using Rails' native handle_unverified_request,
(d) other?
2.  Do we have tests which will detect whether this breaks the controller?  (All current tests passed.)

pellaea:
I suspect it *doesn't* break api.  CSRF is enforced by a before_filter, and
I have almost all filters disabled for api.  You could try disabling

    skip_filter   :verify_authenticity_token

in self.disable_filters in app_controller.rb.  That might cause
api_controller_test.rb to fail if it tests any posts.

-Jason

PR #70, opened 2015-02-04
Miscellaneous changes to app configuration based on Rails 4.0.13 default

JoeCohen:
* See RailsDiff http://railsdiff.org/diff/v3.2.21/v4.0.13/
* Many comments
* Also MO-specific comments added to locales/en.txt


JoeCohen:
PR not ready for prime time -- rails server won't start.

pellaea:
oops!  will wait. :)

JoeCohen:
The 2nd commit fixes the immediate problem, but might not be the right long-range solution; we probably should look as much as possible like the default Rails application, and a default Rails 4.0.13 application lacks a script/rails file.

pellaea:
I'm a little confused about the secret_token -vs- secret_key_base.  I see secret_token created in config/initializers/secret_token.rb, but the comments talk about secret_key_base.  When I added secret_key_base, I added a "fake" one to consts.rb (feel free to move to initializers if that's where you feel it belongs), then have the "real" one live privately only on the production server (in consts-site.rb, a file which is not tracked by git).

I've preformed the merge already anyway.  Just help me understand what's going on here, and I can push this last tweak directly to origin ror4.  Certainly no "secret" anything belongs in git! :)

JoeCohen:
Thanks for catching this.  I will look at it later today.

*--Joe Cohen*

JoeCohen:
secret_token is the Rails 3 scheme; secret_key_base is new to Rails 4:
>Rails 4.0 introduces ActiveSupport::KeyGenerator and uses this as a base from which to generate and verify signed cookies (among other things). Existing signed cookies generated with Rails 3.x will be transparently upgraded if you leave your existing secret_token in place and add the new secret_key_base.
    # config/initializers/secret_token.rb
    Myapp::Application.config.secret_token = 'existing secret token'
    Myapp::Application.config.secret_key_base = 'new secret key base'
>Please note that you should wait to set secret_key_base until you have 100% of your userbase on Rails 4.x and are reasonably sure you will not need to rollback to Rails 3.x. This is because cookies signed based on the new secret_key_base in Rails 4.x are not backwards compatible with Rails 3.x. You are free to leave your existing secret_token in place, not set the new secret_key_base, and ignore the deprecation warnings until you are reasonably sure that your upgrade is otherwise complete.
>If you are relying on the ability for external applications or Javascript to be able to read your Rails app's signed session cookies (or signed cookies in general) you should not set secret_key_base until you have decoupled these concerns.

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#action-pack

The standard way of keeping this secret is to reference a constant in config/initializers/secret_token.rb, and define that constant in a file that's not under version control.  Also maybe hard-code it for test and development.  See, e.g., http://daniel.fone.net.nz/blog/2013/05/20/a-better-way-to-manage-the-rails-secret-token/.
So:  What if we simply leave  secret_key_base undefined until the app is in much better shape?  We'll get one deprecation warning  I don't think that's a problem and it will remind us to fix secret_key_base when we are ready to roll out the new version.

pellaea:
Thanks, Joe.  You've got me scared now.  I better read up on exactly what
this is doing.  For the moment, I've just set secret_key_base (rails 4 key)
on the production server.  So I guess that's why nothing broke.

We can either leave it the way it is, or as you suggest, disable it and
learn to ignore the deprecation warning (I seem to remember there being a
way to suppress it without actually setting the key, but might be mixing it
up with something else).  Makes no difference to me.  Either way, when we
go live with ror4, it should either be fine the way it is and it will start
using the key then, or if we disable it now, we can enable it later once
we're sure everything's working.  (But we always test big new releases in
parallel for weeks before going live, so I'm not concerned about that,
personally.)

As for dealing with secret data, I recommend keeping the set-up we have.
Instead of distributing secret things among various files which only live
on the server - a lot to keep track of when deploying and testing
production servers, guaranteed to cause problems -- or polluting the
runtime environment (another common suggestion I've seen), I've put all the
sensitive stuff in one file, config/consts-site.rb (which I keep backed up,
encrypted, on my laptop).

-Jason

JoeCohen:
1. I agree with keeping the secrets in one place.  You (a) could either keep the current setup, or (b) define a global constant in the consts file, then reference that constant in secret_token.rb.
2. Is there any advantage to leaving it undefined until we are happy with the parallel run?
3. All deprecation warnings can be turned off by `ActiveSupport::Deprecation.silenced = true`. (In fact, I used this for a while when we were still getting the form_for_function and button_for_function warnings.)  I don't know how to silence the warnings only for secret_token.  But with only one warning, I'd just as soon leave the warnings on. One warning isn't much bother, and we'll get notice is we accidentally re-introduced something else which is deprecated.

pellaea:
3. OMG, there's a way to silence all deprecation warnings??? Now you tell
me!

2. I see no advantage to turning it off now.

1. As I designed the current setup, it should come as no surprise that I
prefer it. ;)

-Jason

PR #71, opened 2015-02-07
remove a few more deprecated or redundant lines

JoeCohen:
and re-order environments/production.rb
  * see http://railscasts.com/episodes/415-upgrading-to-rails-4
  * and see deprecation warning when starting `rails console`
  * put MO production.rb lines in same order as default production.rb
for ease of comparison


pellaea:
Nice!  I bet there wascruft in there from rails 1, still! :)  Thanks!

PR #73, opened 2015-02-07
License db

JoeCohen:
* add public domain license fixture
  * correct tests which expect only 3 licenses
  * seed development db  with fixtures
  * **developers must also run `rake db:seed`**

Fixes issues discussed in https://groups.google.com/forum/#!topic/mo-developers/hTxzv61e-Zw


pellaea:
Let's fix mo-dev to run rake db:seed, then.

pellaea:
Done.

PR #74, opened 2015-02-08
changes to pagination HTML/CSS and odd_even logic

raysuelzer:
This commit adjusts the pagination HTML and CSS and removes most of the odd_even logic used to determine coloration of table lines.  This is now handled via CSS.


app/helpers/application_helper.rb
```diff
@@ -1581,22 +1581,20 @@ def pagination_numbers(pages, args={})
       pstr = "« #{:PREV.t}"
       nstr = "#{:NEXT.t} »"
       result << pagination_link(pstr, this-1, arg, args) if this > 1
-      result << '|'                                      if this > 1
       result << pagination_link(1, 1, arg, args)         if from > 1
-      result << '...'                                    if from > 2
+      result << result << content_tag(:li, content_tag(:span, "..."), class: "disabled")  if from > 2
```
JoeCohen: Should `result <<` appear twice on this line?

app/views/location/list_locations.html.erb
```diff
@@ -1,68 +1,71 @@
...
-  @known_pages = @pages
-  @known_data = @objects
+   @known_pages = @pages
+   @known_data = @objects
 
-  if (!@known_pages or @known_pages.empty?) and
-     (!@undef_pages or @undef_pages.empty?)
-    flash_error(@error) if @error
-  end
+   if (!@known_pages or @known_pages.empty?) and
```
JoeCohen: Is it possible to use `try` here &/or other places?  See [try(*args)](http://api.rubyonrails.org/classes/NilClass.html#method-i-try).
(I haven't played around with the try method yet.)

app/views/location/list_locations.html.erb
```diff
@@ -1,68 +1,71 @@
...
+    <% end %>
+    <div class="row">
+      <div class="col-xs-12">
+                <%= pagination_numbers(@known_pages) %>
+      </div>
+      <div class="col-xs-12 color-block">
+        <% @known_data.each do |location| %>
+            <div class="col-xs-12 color-block-item slight-pad">
+              <%= link_with_query(location.display_name.t,
+                                  :action => 'show_location', :id => location.id) %>
```
JoeCohen: Here and elsewhere, new-style key: value format is a little shorter
~~`:action => 'show_location', :id => location.id) %>`~~
`action: 'show_location', id: location.id) %>`

pellaea: Yeah, it would be nice to change to the new style of key: val throughout,
but woof! that's a lot of work!  I fix it everywhere I touch the code, but
even so it will take a very long time to get 'em all...

raysuelzer: Sorry, what is the key: val approach?

Also, I'd test the performance of the try statements.  If it is anything
like "rescue" we should avoid it.

http://technicaldebt.com/the-cost-of-using-rubys-rescue-as-logic/

Ray

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer: Or maybe it would be faster if the argument isn't ever nil.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer: result << result << content_tag(:li, content_tag(:span, "..."),
class: "disabled")  if from

I will fix this.


Ray Suelzer
Mobile: (312) 869-2665

JoeCohen: For a hash, instead of
:key => value
type
key: value
(works as long as key is a Symbol)

JoeCohen:
`try` doesn't seem to be anything like `rescue`
See https://github.com/rails/rails/blob/cd2d3664e3b434d15b6c19e652befb386187642f/activesupport/lib/active_support/core_ext/object/try.rb#L93

pellaea:
I think Ray's point is that there is some (significant?) overhead to
generating exceptions, and that a simple conditional is the better way to
handle such cases.  Exceptions should be, well, exceptional! :)

That said, I use the syntax

value = expression rescue nil

all the time.  But "expression" *should* generally work.  This is just a
way of saying, "value is really unimportant in the grand scheme of things,
if we can't set it correctly because of one or two tricky boundary cases
which are ugly or difficult to test for, then so what, just leave it nil".

As far as performance goes, if it doesn't generate an exception 99% of the
time, then who cares if it takes twice as long the remaining 1%?  It won't
affect the average performance of the code.

In the case of testing if the results of a query are empty, then the page
is already going to be incredibly fast because the query was empty(!) so
who cares about a little extra overhead.

But if we're testing for an error message which will be *absent* most of
the time, then the "exceptional" case is actually the norm!  That doesn't
feel right to me.

raysuelzer:
I am all for implementing joes suggestion.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
screenshot from 2015-02-07 17 59 56
screenshot from 2015-02-07 18 00 05

raysuelzer:
I added toggle navigation for mobile devices. Pretty sweeeet.

Ray Suelzer
Mobile: (312) 869-2665

PR #76, opened 2015-02-08
Fixes image ajax fallback links not rendering adds tests to ensure they render

raysuelzer:


raysuelzer:
boo!  I wanted coverage increase.

Ray Suelzer
Mobile: (312) 869-2665

PR #83, opened 2015-02-16
huge layout work

raysuelzer:
--Added cookie to hide banner
--Made "flashes" prettier
--Started to do something with tab_sets, but I think I might reverse course.  I was attempting to render as partials.
--Removed a decent amount of code that was only used in a single view from application_helper. Mostly dealing with description links.
--Added 'theater' mode to view an image
--Still have to fix ajax vote bar displaying funky on landscaped images
--Removed tables from a whole bunch of views and replaced with responsive css
--Added top menu bar with account links, a la github style, still need to work on mobile responsiveness
--Adjust gmaps helper in application-helper to allow for responsive maps based on percentage rather than fixed with.
--Cleaned up some of the tests to use more modern methods, i.e. using assert_tag instead of a regex for extracting links.
--Tried to move the _prev_next_pager into a different view, but it can probably be put into a helper
--All the tests pass, but javascripts are still very much broken in some places, but will be easy to fix.


raysuelzer:
https://drive.google.com/file/d/0B_xUuUwXMGp9UXljMmpwc29rTDA/view?usp=sharing

Just a video of some of how it looks and stuff.

Ray Suelzer
Mobile: (312) 869-2665

PR #85, opened 2015-02-19
Converts all hashes to Ruby 1.9 style.

raysuelzer:
I am not sure if this will pass unit tests because I cannot get VM working on my windows machine.  Travis will tell us if it doesn't pass.


pellaea:
Bah, we can deal with unit tests later. ;)

pellaea:
Haha, yes, that should do it alright, but you might want to make sure you
do it on a fresh branch of ror4 in which everything is already working! :)

JoeCohen:
And -- if you can do it easily -- think about also catching the cases where the key is currently a string, but could equally well be a symbol.  E.g., `"abc" => foo` could become `abc: foo`.

pellaea:
That's trickier.  You can definitely change a few like "action" =>
"show_observation" to action: :show_observation.  But there are going to be
some that have to remain strings.  Wish I had some examples off the top of
my head.  I think acts_as_versioned had some parameters which had to be
strings(?)  You can try converting all strings to symbols, then reverting
the ones that break things... but really it doesn't seem so important(!)

raysuelzer:
Hmmm... Looking at my local copy now, and it did have the reuse_image
partial, I think I realized my mistake but it didn't get "added" by git.
I'm too tired to deal with it right now, but once i re-add that and adjust
the test to look in the right div for the links it will pass.

Ray Suelzer
Mobile: (312) 869-2665

PR #96, opened 2015-02-22
even more layout fixes

raysuelzer:
Added votes back to gollsary terms. 
Publications Page:
 make the text block on the top wrap on larger screen
 make the publications links not cause the table to go off screen.
Half way done with the image fix. CLicking image in theater now download original.
fixed media queries.
Moved theater to its own partial view since it will probably only grow in the future.


raysuelzer:
seems odd.  didn't change that much code.

Ray Suelzer
Mobile: (312) 869-2665

PR #103, opened 2015-02-22
Always use 100% of the screen!

raysuelzer:


pellaea:
done

raysuelzer:
Hmm seems to have introduced a problem with rss-detail boxes being sized
weird on load.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
Of course, some browsers consider the dom ready before images have
loaded... so it's resizing to the wrong size.

Ray Suelzer
Mobile: (312) 869-2665

PR #106, opened 2015-02-23
layout tweaks

raysuelzer:
hide the images for site states on cell phones they don't add much.
Again fixing the theater button. It was breaking the responsiveness of images to position it under the image.


raysuelzer:
One tiny one incoming.

Ray Suelzer
Mobile: (312) 869-2665

PR #108, opened 2015-02-23
I think this fixes multiimage uploader, but need to install ngnix locall...

raysuelzer:
...y. But, definitely fixes the JS up until it expects a response from the server.


pellaea:
merged, pushed to test server

PR #112, opened 2015-02-24
Carousel doesn't wrap anymore...

raysuelzer:
Finished tests in 296.870817s, 3.9916 tests/s, 57.5301 assertions/s.

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -78,19 +78,13 @@ jQuery(document).ready(function () {
         return undefined;
     }
 
-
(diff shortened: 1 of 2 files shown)


pellaea:
done, merged, pushed

raysuelzer:
Unfortunately, it will still show white spaces until you are left with one
image on the screen.  There doesn't seem to be an easy fix for this since
it's already a hack on top of the carousel.  I'd just assume add it to the
list of things to fix after release as it is of minor concern since it
won't cause an infinite wrap anymore.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Sounds good to me.

PR #128, opened 2015-03-01
html fixes and md5 check for banner.

raysuelzer:
Finished tests in 292.688283s, 4.0487 tests/s, 58.3522 assertions/s.

1185 tests, 17079 assertions, 0 failures, 0 errors, 0 skips


pellaea:
done

PR #130, opened 2015-03-02
Dropdown for languages shows now, but not sure why the overflow was set ...

raysuelzer:
...to hidden on navigation.  Are there items overflowing?

### DIFF
--- a/app/assets/stylesheets/mushroom_observer.scss
+++ b/app/assets/stylesheets/mushroom_observer.scss
@@ -331,7 +331,7 @@ a {
             padding: 1px 5px;
         }
         white-space: nowrap;
-        overflow: hidden;
+        overflow: visible;
         text-overflow: ellipsis;
     }
 


pellaea:
Hmmm, I think we need to change overflow to visible more judiciously.  I added that to prevent wrapping of "Create Observation" and such in those in-between screen sizes when the lefthand nav bar was too small to fit the text, but not small enough to flip into the next smaller tier.  Doing white-space:nowrap + overflow:hidden + text-overflow:ellipsis causes it to ellide long entries instead of wrapping them.  If I don't get out-voted, and we keep this styling, then we need to be sure to apply overflow:visible only to the language dropdown item.

What do you think about all this?  Is there a better way to deal with the wrapping?  I'd *prefer* to ensure the lefthand nav bar stays large enough to fit the longest item, but I think that requires a table(?)  The ellipsis trick was just a compromise that allowed responsive resizing of the nav bar while preventing ugly wrapping.

Thoughts?

raysuelzer:
I think your approach is just fine.  Could be solved by setting overflow-x:
hidden; overflow-y: visible;  so it can overflow below but not on the side?

On Sunday, March 1, 2015, Jason Hollinger <notifications@github.com> wrote:

> Hmmm, I think we need to change overflow to visible more judiciously. I
> added that to prevent wrapping of "Create Observation" and such in those
> in-between screen sizes when the lefthand nav bar was too small to fit the
> text, but not small enough to flip into the next smaller tier. Doing
> white-space:nowrap + overflow:hidden + text-overflow:ellipsis causes it to
> ellide long entries instead of wrapping them. If I don't get out-voted, and
> we keep this styling, then we need to be sure to apply overflow:visible
> only to the language dropdown item.
>
> What do you think about all this? Is there a better way to deal with the
> wrapping? I'd *prefer* to ensure the lefthand nav bar stays large enough
> to fit the longest item, but I think that requires a table(?) The ellipsis
> trick was just a compromise that allowed responsive resizing of the nav bar
> while preventing ugly wrapping.
>
> Thoughts?
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/130#issuecomment-76656148>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Oh that works.  So simple.  Duh.  I'll take care of it.

raysuelzer:
I also saw this was trending on GitHub, worth a read.

https://github.com/bendc/frontend-guidelines/blob/master/README.md

On Sunday, March 1, 2015, Jason Hollinger <notifications@github.com> wrote:

> Oh that works. So simple. Duh. I'll take care of it.
>
> On Sun, Mar 1, 2015 at 8:24 PM, Raymond Suelzer <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > I think your approach is just fine. Could be solved by setting
> overflow-x:
> > hidden; overflow-y: visible; so it can overflow below but not on the
> side?
> >
> > On Sunday, March 1, 2015, Jason Hollinger <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> > wrote:
> >
> > > Hmmm, I think we need to change overflow to visible more judiciously. I
> > > added that to prevent wrapping of "Create Observation" and such in
> those
> > > in-between screen sizes when the lefthand nav bar was too small to fit
> > the
> > > text, but not small enough to flip into the next smaller tier. Doing
> > > white-space:nowrap + overflow:hidden + text-overflow:ellipsis causes it
> > to
> > > ellide long entries instead of wrapping them. If I don't get out-voted,
> > and
> > > we keep this styling, then we need to be sure to apply overflow:visible
> > > only to the language dropdown item.
> > >
> > > What do you think about all this? Is there a better way to deal with
> the
> > > wrapping? I'd *prefer* to ensure the lefthand nav bar stays large
> enough
> > > to fit the longest item, but I think that requires a table(?) The
> > ellipsis
> > > trick was just a compromise that allowed responsive resizing of the nav
> > bar
> > > while preventing ugly wrapping.
> > >
> > > Thoughts?
> > >
> > > --
> > > Reply to this email directly or view it on GitHub
> > > <
> >
> https://github.com/MushroomObserver/mushroom-observer/pull/130#issuecomment-76656148
> > >
> > > .
> > >
> >
> >
> > --
> > Ray Suelzer
> > Mobile: (312) 869-2665
> >
> > --
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/MushroomObserver/mushroom-observer/pull/130#issuecomment-76656980
> >
> > .
> >
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/130#issuecomment-76657328>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

PR #133, opened 2015-03-03
Modernized app/views/account/*

pellaea:
Also fixed bug in api key manager: jquery.show() doesn't seem to be able to show something hidden with class="hidden".  Changed it to "display:none" and now jquery.show() can manage it.  Weird.  I could've sworn this worked before.

Also added data-autofocus as an alternative to plain vanilla autofocus.  I don't know about other browsers, but autofocus causes firefox to scroll stupidly down the page, hiding the title and stuff.  If instead wewait until the page is loaded and then focus on the first element with data-autofocus=true, then it just focuses without scrolling.

### DIFF
--- a/app/assets/javascripts/api_key.js
+++ b/app/assets/javascripts/api_key.js
@@ -3,13 +3,15 @@ function ApiKeyModule() {
 
     // bindings
     jQuery('[data-role="edit_api_key"]').click(function(event){
+      var keyId = $(this).data().id;
       event.preventDefault();
-      shouldShowEditFields($(this).data().id, true);
+      shouldShowEditFields(keyId, true);
     });
 
     jQuery('[data-role="activate_api_key"]').click(function(event){
+      var keyId = $(this).data().id;
       event.preventDefault();
-      activateKey($(this).data().id);
+      activateKey(keyId);
     });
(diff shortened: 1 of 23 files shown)


raysuelzer:
Yeah, it's annoying.  Check out bootstraps css for hidden.  jQuery doesn't
change the visibility or override the !important.


.hidden { display: none !important; visibility: hidden !important;}


Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Oh, is that why?  I figured jquery changed both visibility and display, but
didn't consider !important.  Out of curiosity, how would you override an
!important styling in javacsript / jquery?

raysuelzer:
you'd have to do
$('#foo').style('display', 'block');
$('#foo').style('visiblity', 'visible');

but if you really just want to hide or unhide things in bootstrap it's
easiest to use  $('#foo').toggleClass('hidden');

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
vanilla javascript would be something like:
 document.getElementById('foo').style.visibility = 'visible';      i
think...

Ray Suelzer
Mobile: (312) 869-2665

PR #139, opened 2015-03-07
Various style work:

pellaea:
Theater view: Removed side-bars tentatively in order to allow image to take up whole page.  Have js select image source appropriately based on window size. Still allows browser to upscale beyond image resolution, need to add some onload hook to img to set max-width and max-height on img to prevent this. Tried valiently for hours to get it to center image vertically, but it is impossible without js.

Matrix view: Tried out two options:

xs = 2 cols
sm = 3 cols
md = 4 cols
lg = 6 cols

But this makes for some mighty small thumbnails at times.  Went with option two instead which feels more comfortable:

xs = 1 col
sm = 2 cols
md = 3 cols
lg = 4 cols

There are resolutions where the 1-column xs view is way too large, but what can you do?  Maybe leave "small thumbnail" pref and morph it into choice between these two options?

Show image: Set default image size to 640 for all sizes, then have js watch window size to determine when to load the huge-sized image.  Should save on bandwidth for mobile devices.

Side-bar: Ray claims it's not moving with screen, but I can't reproduce this. But I did set a max width of 15em to prevent it from getting absurdly wide at largest resolutions.

Image votes: Removed spaces to tighten it up, making it take less room so it fits more comfortably when thumbnails are tiny.

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -41,21 +41,25 @@ jQuery(document).ready(function () {
     });
 
     jQuery('body').delegate('[data-toggle="theater"]', 'click', function (e) {
-        console.log('click');
         e.preventDefault();
-        var img_src = jQuery(this).data().image;
-        var img_orig = jQuery(this).data().original;
-        jQuery('.img-theater').css('top', jQuery(document).scrollTop())
-        jQuery('.img-theater').show();
         jQuery('body').addClass('theater-shown');
-        jQuery('#img_append_target').html('<a href="{{orig}}"><img src="{{src}}" style="height: {{h}}; width: auto;"><//img><//a>'
-            .replace("{{src}}", img_src)
-            .replace("{{orig}}", img_orig)
-            .replace("{{h}}", jQuery(window).height() - 20 + 'px'))
-        jQuery(document).on('keyup.hideTheater', function (e){
+        var win = jQuery(window),
+            doc = jQuery(document),
+            div = jQuery('.img-theater-div'),
+            w = win.width(),
+            h = win.height(),
+            img_orig = jQuery(this).data().orig,
+            img_src = w-40 > 1280 || h-40 > 1280 ? jQuery(this).data().full :
+                      w-40 > 960  || h-40 > 960  ? jQuery(this).data().huge :
+                                                   jQuery(this).data().large;
+        div.html('<img><//img>');
+        div.find('img').attr('src', img_src)
+                       .click(function() { window.location = img_orig });
+        jQuery('.img-theater').css('top', doc.scrollTop()).show();
+        doc.on('keyup.hideTheater', function (e) {
             if (e.keyCode == 27) {
                 hideTheater();
-                jQuery(document).unbind('keyup.hideTheater');
+                doc.unbind('keyup.hideTheater');
             }
         })
     });
--- a/app/assets/stylesheets/mushroom_observer.scss
+++ b/app/assets/stylesheets/mushroom_observer.scss
@@ -303,6 +303,10 @@ a:hover {
 // left nav bar
 // --------------------------------------------------
 
+.sidebar-offcanvas {
+    max-width: 15em;
+}
+
 #navigation {
     background-color: $LEFT_BAR_BG_COLOR;
     border-color: $LEFT_BAR_BORDER_COLOR;
@@ -540,16 +544,36 @@ a:hover {
 
 .img-theater {
     position: absolute;
-    top: 0px;
-    background: none repeat scroll 0% 0% rgba(48, 50, 48, 0.55);
+    top: 0;
+    left: 0;
+    margin: 0;
+    padding: 0;
     width: 100%;
     height: 100%;
-    z-index: 999;
+    z-index: 998;
     display: none;
-}
-
-#img_append_target {
-    padding-top: 20px;
+    background: none repeat scroll 0% 0% rgba(48, 50, 48, 0.90);
+    .img-theater-btn {
+        position: absolute;
+        top: 5px;
+        right: 20px;
+        z-index: 999;
+        @include opacity(0.50);
+    }
+    .img-theater-btn:hover {
+        @include opacity(1.00);
+    }
+    .img-theater-div {
+        width: 100%;
+        height: 100%;
+        padding: 20px;
+        text-align: center;
+        img {
+            max-width: 100%;
+            max-height: 100%;
+            vertical-align: middle;
+        }
+    }
 }
 
 //
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1714,10 +1714,18 @@ module ApplicationHelper
       output = "".html_safe
       for object in objects
         body = capture(object, &block).to_s
-        output += content_tag(:div, "", class: "hidden visible-xs-block", style: "clear:left") if idx % 2 == 0
-        output += content_tag(:div, "", class: "hidden visible-sm-block visible-md-block", style: "clear:left") if idx % 3 == 0
-        output += content_tag(:div, "", class: "hidden visible-lg-block", style: "clear:left") if idx % 4 == 0
-        output += content_tag(:div, body, class: "col-xs-6 col-sm-4 col-lg-3")
+        if false ## Maybe give users who want small thumbnails this option?
+          output += content_tag(:div, "", class: "hidden visible-xs-block", style: "clear:left") if idx % 2 == 0
+          output += content_tag(:div, "", class: "hidden visible-sm-block", style: "clear:left") if idx % 3 == 0
+          output += content_tag(:div, "", class: "hidden visible-md-block", style: "clear:left") if idx % 4 == 0
+          output += content_tag(:div, "", class: "hidden visible-lg-block", style: "clear:left") if idx % 6 == 0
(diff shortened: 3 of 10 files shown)


raysuelzer:
Did you put all the matrix box code I removed back in there?  did you use
the CSS to set te different widths. If it's the CSS solution then that is
cool with me.

Theater view: Taking up the whole page will be a problem because images are
all size differently.  What was the issue with the scaling to fit the image
on the screen before?  It should have adjusted to fit the image on the
screen and maintained proportions doth at it never went off screen
vertically or horizontally. Was that not the case?


On Friday, March 6, 2015, Jason Hollinger <notifications@github.com> wrote:

> Theater view: Removed side-bars tentatively in order to allow image to
> take up whole page. Have js select image source appropriately based on
> window size. Still allows browser to upscale beyond image resolution, need
> to add some onload hook to img to set max-width and max-height on img to
> prevent this. Tried valiently for hours to get it to center image
> vertically, but it is impossible without js.
>
> Matrix view: Tried out two options:
>
> xs = 2 cols
> sm = 3 cols
> md = 4 cols
> lg = 6 cols
>
> But this makes for some mighty small thumbnails at times. Went with option
> two instead which feels more comfortable:
>
> xs = 1 col
> sm = 2 cols
> md = 3 cols
> lg = 4 cols
>
> There are resolutions where the 1-column xs view is way too large, but
> what can you do? Maybe leave "small thumbnail" pref and morph it into
> choice between these two options?
>
> Show image: Set default image size to 640 for all sizes, then have js
> watch window size to determine when to load the huge-sized image. Should
> save on bandwidth for mobile devices.
>
> Side-bar: Ray claims it's not moving with screen, but I can't reproduce
> this. But I did set a max width of 15em to prevent it from getting absurdly
> wide at largest resolutions.
>
> Image votes: Removed spaces to tighten it up, making it take less room so
> it fits more comfortably when thumbnails are tiny.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/MushroomObserver/mushroom-observer/pull/139
> Commit Summary
>
>    - Various style work:
>
> File Changes
>
>    - *M* app/assets/javascripts/main.js
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-0>
>    (26)
>    - *M* app/assets/stylesheets/mushroom_observer.scss
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-1>
>    (38)
>    - *M* app/helpers/application_helper.rb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-2>
>    (16)
>    - *M* app/views/image/_image_thumbnail.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-3>
>    (14)
>    - *M* app/views/image/_image_vote_links.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-4>
>    (14)
>    - *M* app/views/image/_theater.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-5>
>    (16)
>    - *M* app/views/image/list_images.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-6>
>    (22)
>    - *M* app/views/image/show_image.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-7>
>    (25)
>    - *M* app/views/layouts/application.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-8>
>    (2)
>    - *M* app/views/observer/list_rss_logs.html.erb
>    <https://github.com/MushroomObserver/mushroom-observer/pull/139/files#diff-9>
>    (34)
>
> Patch Links:
>
>    - https://github.com/MushroomObserver/mushroom-observer/pull/139.patch
>    - https://github.com/MushroomObserver/mushroom-observer/pull/139.diff
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/139>.
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
Re hamburger menu button should always be visible when you scroll down on a
phone. Ie you shouldn't have to scroll back to the top to get to the menu
button. Like if I scroll halfway down the observations page the menu button
should stick to the top on a phone.

On Friday, March 6, 2015, Jason Hollinger <notifications@github.com> wrote:

> Merged #139
> <https://github.com/MushroomObserver/mushroom-observer/pull/139>.
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/139#event-247919876>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

pellaea:
We have the same goals, look at it, see what I did.

Yes, it's using css to do the matrix stuff.  But now, the number of columns
adjusts to the window size, your version always had 3 no matter what
(except for mobiles).  The trick was getting the cells to wrap nicely.
That's what the empty <div style="clear:left"> are for, ensures that the
overflow wraps starting at the left of the next row.  We can stick
make_matrix() in a partial instead if you prefer.  But we clearly shouldn't
be duplicating all that code in every view that has a matrix display.

My issue with the old theater was that some images weren't allowed to fill
the whole screen when they were the appropriate aspect ratio to do so.
There was always a bunch of dead space on left and right, and I couldn't
understand what purpose it served.  Obviously we don't want landscape
images extending off the page, and it shouldn't do that.  I was also hoping
that by simplifying the html dramatically (div{div{img}} instead of
div{div{div div {img} div}} and I might be missing some) I might be able to
convince browsers to center the image vertically, but failed.  We can throw
all those superfluous divs back in there, I just can't imagine why we would
want to.

Of course, I can only test in firefox and chrome, so I could well have
f---ed things up, in which case we can always go back.  Easy enough to do.
But things look and feel and work better on my browser.  I'd prefer we
figure out how to incorporate my changes instead of throwing them away.

Thanks for clarifying the hamburger thing.  That's got to be a trivial
thing to fix.  Maybe I fat-fingered something accidentally.
​

pellaea:
Hmmm.  The hamburger buttons are clearly styled as "position:fixed"
right in the html.  And in fact, they stay put at the top of the
window just like they should.  For me.  Arg.

When is the last time you definitely remember it working right?  How
far back do I need to go to look for diffs that might be responsible
for this bug?

raysuelzer:
Hmm I'll look at the hamburger thing. The rest sounds good. Ignore my
previous github comment. I'm on my iPhone.

No worries. Sounds like all the changes are great and thank you!

On Friday, March 6, 2015, Jason Hollinger <notifications@github.com> wrote:

> Hmmm. The hamburger buttons are clearly styled as "position:fixed"
> right in the html. And in fact, they stay put at the top of the
> window just like they should. For me. Arg.
>
> When is the last time you definitely remember it working right? How
> far back do I need to go to look for diffs that might be responsible
> for this bug?
>
> On Fri, Mar 6, 2015 at 7:37 PM, Jason Hollinger <pellaea@gmail.com
> <javascript:_e(%7B%7D,'cvml','pellaea@gmail.com');>> wrote:
> > We have the same goals, look at it, see what I did.
> >
> > Yes, it's using css to do the matrix stuff. But now, the number of
> columns
> > adjusts to the window size, your version always had 3 no matter what
> (except
> > for mobiles). The trick was getting the cells to wrap nicely. That's what
> > the empty <div style="clear:left"> are for, ensures that the overflow
> wraps
> > starting at the left of the next row. We can stick make_matrix() in a
> > partial instead if you prefer. But we clearly shouldn't be duplicating
> all
> > that code in every view that has a matrix display.
> >
> > My issue with the old theater was that some images weren't allowed to
> fill
> > the whole screen when they were the appropriate aspect ratio to do so.
> > There was always a bunch of dead space on left and right, and I couldn't
> > understand what purpose it served. Obviously we don't want landscape
> images
> > extending off the page, and it shouldn't do that. I was also hoping that
> by
> > simplifying the html dramatically (div{div{img}} instead of div{div{div
> div
> > {img} div}} and I might be missing some) I might be able to convince
> > browsers to center the image vertically, but failed. We can throw all
> those
> > superfluous divs back in there, I just can't imagine why we would want
> to.
> >
> > Of course, I can only test in firefox and chrome, so I could well have
> > f---ed things up, in which case we can always go back. Easy enough to do.
> > But things look and feel and work better on my browser. I'd prefer we
> > figure out how to incorporate my changes instead of throwing them away.
> >
> > Thanks for clarifying the hamburger thing. That's got to be a trivial
> thing
> > to fix. Maybe I fat-fingered something accidentally.
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/139#issuecomment-77672346>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
I think the reason I set the side margins was to make room for the X button
and  more importantly, so that we weren't loading 10mb images when someone
clicked on the image and went with the 1200px image.  Serving the original
image in theater view makes the feature really, really slow on even a
modest DSL connection.  Too about 15 seconds to load an image here at the
hotel.   Just my two cents.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
Some minor bugs I see:

Pagination at the bottom is in the wrong position on lg-4,  boxes are no
longer scaled to the same size.
Top navigation extra white space to the right side.

[image: Inline image 3]

[image: Inline image 2]



Ray Suelzer
Mobile: (312) 869-2665

PR #140, opened 2015-03-07
Moved make_matrix into partial.

pellaea:


raysuelzer:
like! +1

Ray Suelzer
Mobile: (312) 869-2665

PR #142, opened 2015-03-07
Various work on nav bars and theater.

pellaea:
1. Tweaks to search bar in mobile mode
2. Fixed width of left nav in mobile mode.
3. Fixed position of bottom pagination links after matrix table.
4. Tweaked theater to only load medium-sized image if can't display larger.
5. Moved theater JS out into new file, since it's gotten rather large.
6. Thumbnail partial passes in image width and height if available, allowing JS to choose and position image very precisely in most cases.
7. Have theater hide hamburger controls in mobile.

Ray, I added the search pulldown back in (mobile mode).  My reasoning is that the search bar is usually hidden, so it doesn't matter if it takes up multiple lines on mobile devices.  I think people use the pulldown often enough that it's worth ensuring that it's present even in mobile.  If you disagree, please feel free to return to the way it was.

I think the left nav issue is fixed now.  We might want to add more padding to the *elements* of the left nav.  You had added extra margin to the entire div in mobile mode, but that looks bad to me -- I prefer when the lines and hover background take up the full width of the nav bar.

I think the theater works really well now... and I've tried to test the mathematics very carefully... but I'm worried that it's too complicated now and dangerous.  I'd understand if people want to switch back to the old one, which always loaded the 960-size image, left it centered at the top of the page -- the js is *way* simpler... just so long as you don't mind that it doesn't center vertically (minor), and that it over-inflates images past 960 on large screens, potentially making users think the photograph itself is crappy quality, when in fact it's just dumb css not knowing to switch to the next higher resolution when the space available to the image grows beyond its full 960x640 resolution (more serious in my opinion).

I'm going to have to stop working on MO (I'm sure Ray is rejoicing! :).  I will try to keep the test server up-to-date if anyone creates pull-requests, at least this weekend.  I might not be able to help at all Monday through Thursday this coming week, though.  Sorry.


raysuelzer:
This all sounds great.  I don't mean I sound critical of your work in my
responses. They are just short because I'm on an iPhone.

Thanks so much for doing all of this, much improved.

On Saturday, March 7, 2015, Jason Hollinger <notifications@github.com>
wrote:

> Merged #142
> <https://github.com/MushroomObserver/mushroom-observer/pull/142>.
>
> --
> <https://github.com/MushroomObserver/mushroom-observer/pull/142#event-248035204>
> .
>


-- 
Ray Suelzer
Mobile: (312) 869-2665

PR #144, opened 2015-03-08
Make the hamburger button only disappear when theater is loaded

raysuelzer:
not when image is moused over.
made "pad" a private variable.


pellaea:
Oops!  Oh yes, I see now, hid them in the wrong callback!  How
embarrassing. :)

PR #145, opened 2015-03-08
Performance improvements(?)

raysuelzer:
I turned off the assets.debug and the assets.logger in the development.rb config in order to get a better sense of how long the partial views were actually taking to render.  Feel free to change it back.
 
I converted the matrix_box views to use the collection parameter for render :partial.  This seems to drastically improve the performance, but my tests are limited to about 20 page loads with the old version and then 20 with the new version.
 
Render time for the page went from about 1000ms to 600ms.   This could probably be further improved.  
Please take a look at it.  I'm not sure it is the silver bullet solution, but might be a step in the right direction.

### DIFF
--- a/app/views/image/list_images.html.erb
+++ b/app/views/image/list_images.html.erb
@@ -3,10 +3,8 @@
   flash_error(@error) if @error and (!@objects or @objects.empty?)
 %>
 
-<%= paginate_block(@pages) do %>
-  <%= render(layout: "shared/matrix_table", locals: {objects: @objects}) do |image| %>
-    <div class="push-down push-up">
-      <%= render(partial: "shared/matrix_box", locals: {object: image}) %>
-    </div>
-  <% end %>
-<% end %>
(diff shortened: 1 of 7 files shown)


pellaea:
The "<div class="push-down push-up">" in list_images was to add whitespace between cells, not at the top and bottom of the whole list.  Might have to add a padding parameter to matrix_box if that's still a problem.  Otherwise this looks good to me.

raysuelzer:
Hm... It seems to look OK here, but feel free to make any changes on that.
My goal here isn't to change the styles but just improve performance.  What
you have done looks great to me.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
thanks, I'm seriously out of time to work on things today, gotta prepare a
talk in the next few hours, and emails are distracting me so much I can't
even get started, haha!

good luck with this caching stuff, hopefully there will be at least a few
things that are easy to improve

PR #146, opened 2015-03-09
performance and javascript

raysuelzer:
Fixed the javascript to render based upon the view size.  Realized it was possible to have a partial with layout and a collection, which improves speed since rails can create all the partials with a single build and teardown.


raysuelzer:
err fixed javascript to adjust box heights based upon the viewport size so that they match.

i.e. when the screen is md make sure every n+2 matrix box is the same height, when it's "lg" make sure every n+3 is the same height.   This makes the matrix boxes in each row the same height.

pellaea:
sweet

PR #148, opened 2015-03-14
Adds slick slider to the project instead of hacky carousel. 

raysuelzer:
NOTE: This adds depreciation warnings becuase the gem uses scss.css as the extension.  I put in a PR to fix this issue see: https://github.com/bodrovis/jquery-slick-rails/pull/10

I also removed the Semantic Vernacular tests.


JoeCohen:
1.  Thanks for dealing with so many issues!  Also nice to see coverage increase a bit.  Hopefully we will be able to release when Jason gets back.
2.  For the Semantic Vernacular tests, I suggest leaving the test code there but disabling the tests from running.  This will (1) if we want -- make it easier to figure out why the timeout error is not being rescued, and (2) let us remove the associated SVD classes at the same time as we remove the tests
To skip a test method, just insert `skip("msg")` as the first line of a test.  For example
```
class SemanticVernacularControllerTest < FunctionalTestCase
  def test_index
    skip("Takes too long to time out.")

raysuelzer:
Sorry Joe.  I deleted them in this commit.  I'm not on my linux box anymore, but i can put them back in.

PR #150, opened 2015-03-19
Raise contrast between Black_and_white page background and boxes

JoeCohen:
* Make page bg about as dark as other themes
  * Addresses https://www.pivotaltracker.com/story/show/90607800


JoeCohen:
I played around with increasing contrast by (a) changing box colors versus (b) changing page background.  I personally preferred changing the page background, which also makes the styling more parallel to the Agaricus theme (and probably the others).
Try it and see if you like it.  We can always continue to tweak it.

pellaea:
I like it!

PR #152, opened 2015-03-30
Modernized interest/* and translation/*.

pellaea:
I think interest was already done, though. :)

Leaves the three big ones:

observer/
name/
location/

And several small ones:

naming/
pivotal/
project/
publications/
species_list/
specimen/
support/

Many of which Ray has already done, as well, esp. in observer.

I'm going to continue to whittle away at the small ones as time permits.

### DIFF
--- a/app/assets/javascripts/translations.js
+++ b/app/assets/javascripts/translations.js
@@ -6,11 +6,11 @@ function TranslationsModule(localizedText) {
       SAVING_STRING = localizedText.saving_string,
       LOADED = false,
       CHANGED = false,
-      $whirly = jQuery("#whirly"),
-      $save_button = jQuery("#save_button"),
-      $cancel_button = jQuery("#cancel_button"),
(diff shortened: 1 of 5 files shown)


JoeCohen:
My time has gotten away from me.  I will also try to whittle away at it
later this week.

*--Joe Cohen*

raysuelzer:
Sorry, didn't realize there was still outstanding work to be done on this.
I can pitch in next weekend or maybe a few nights this week. I was
wondering what was up with the radio silence for the last few weeks.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Radio silence meant that no one had the time to even think about working on
MO!

I know this work I'm doing isn't a priority... for everyone else.  But it's
very mechanistic and mindless work, about all I'm capable of at the
moment.  And it has to be done sooner or later.  At least so it seems to me.

-Jason

PR #161, opened 2015-04-25
A pull request showing a proposed fix for the create obersvation form wh...

raysuelzer:
...ich cleans it up a bit.  If we want to go this way, then I can finalize it and make ready for other themes.


raysuelzer:
screenshot from 2015-04-25 19 01 15

pellaea:
Don't know about everyone else, but this looks great to me.  My only
comment is that I think the <hr>'s are unnecessary and distract the eye and
at any rate fail to actually help group things.  But that may just be a
personal preference.
-Jason

raysuelzer:
Cool.  I can pop those out and just put in white space.

Ray Suelzer
Mobile: (312) 869-2665

pellaea:
Should I merge it?  Sounded like you wanted feedback first(?)

raysuelzer:
You can hold if on merging it if we can wait until tomorrow for me to
finish it up.  The tests failed on something I didn't touch, not sure what
is up with that.  They passed locally.

Ray Suelzer
Mobile: (312) 869-2665

raysuelzer:
Made the changes and tested in all the themes.  Looks pretty slick.  I've noticed the old image uploader might be having some issues, specifically it doesn't seem to be adding more than one image when I select an image.  I'll look into that a bit more.

PR #162, opened 2015-05-23
Fixing location controller list by country regexp

lpasquali:
As discussed here:
https://groups.google.com/forum/#!topic/mo-developers/r_vYlC6XoTk

I'm proposing a fix to get a clean list_by_country for locations in MO (Italy at the time shows spurious entier from Australia, Mexico shows New Mexico, Usa countries).

this issue was introduced by this commit:
https://github.com/MushroomObserver/mushroom-observer/commit/72c41768c27038bdad99a8588b37a6033a60f014

anyway, test_list_by_country_quote doesn't fail with the code fix.
I run script/run_tests and rake test:functionals and it looks like everything works.
If I revert the patch the regression get triggered and my test fails. 

these test fail:
Error: test_index(SemanticVernacularControllerTest)
Error: test_all_svds(SvdTest)

but they look unrelated.

Ciao!
Luca Pasquali

### DIFF
--- a/app/controllers/location_controller.rb
+++ b/app/controllers/location_controller.rb
@@ -69,7 +69,7 @@ class LocationController < ApplicationController
 
   # Displays a list of all locations whose country matches the id param.
   def list_by_country # :nologin:
-    query = create_query(:Location, :regexp_search, :regexp => "#{params[:country]}*")
+    query = create_query(:Location, :regexp_search, :regexp => "#{params[:country]}$")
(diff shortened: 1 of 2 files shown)


pellaea:
Looks great.  Thanks for writing the test, and nicely done.  Sorry, should have warned you about the possibility of SVD failures.  Those tests are disabled in the new release, I think, but still give false failures on the master branch. :(

PR #166, opened 2015-07-26
Fixes email tests failing when using a windows machine because the line endings…

raysuelzer:
are not normalized between the template file and the test file that is created. This causes problems when files get committed with improper line endings, for example our text files are committed as CLRF and the HTML files are LF.

This applies the normalizing function fix_mac_vs_pc to both the source template and the test file to normalize line endings.


raysuelzer:
of course... now it doesn't work on linux machines.

mo-nathan:
Hi Ray,

Should I reject this pull request since the Travis build is failing and your comment about it not working on Linux machines or do you plan to work this one out?

Thanks!
-Nathan

raysuelzer:
Reject it.  But there are issues with the line endings in those test files being different than all the other files in the project.  I will try and look at it again this week.

PR #168, opened 2015-08-03
Mycobank link

JoeCohen:
Creates links to MycoBank from Names and Observations
Addresses [Pivotal story #61044172](https://www.pivotaltracker.com/story/show/61044172)
Caveats:
(1) I updated the app_banner_box and observer/news in en.txt.  Please feel free to modify.  More important, I don't know how to make localizations update.  (They should initially be updated to be === English.)  
(2) The links are to a search in MycoBank.  If there's an official MycoBank translation for the MO user's language, the results are in that translation.  
The MO method uses a custom array to translate from MO language to MycoBank.  Another approach is to internationalize and localize this using I18n.  I didn'tt do that because (1) MO users should not be able to change the localizations, and I don't know how to avoid that; (2) the custom array is more flexible in that it will automatically deal with any new languages added to MO.

Some issues with the MycoBank search results.  
(a) The hits are all taxa whose name includes the text of the searched taxa.  The biggest annoyance is the search for genus, as the hits include at least all subordinate taxa of that genus.  
(b) I do not see how to search for a rank between genus and species (e.g., subgenus).  So for those ranks I just search for genus.  This is doubly annoying because of (a) above.  

All-in-all, less than ideal, but it's a big improvement, at least for my use case.  I hope it also will save MO users the trouble of trying to paste MycoBank links into descriptions, comments, etc.


JoeCohen:
**Please hold up on merging this.**  There is something very strange going on:  when I pull this branch back down to my local machine, then run the tests on the latest VM, I get 11 failures.  I will post a longer message in the MO-Developers forum.

JoeCohen:
This PR is now ready to go. 
Sorry about all the duplicate commits.  (I'm doing something wrong in my Git GUI when I rebase.)

mo-nathan:
Thanks for all the Rubocop cleanup in addition to the fix!

PR #175, opened 2015-09-03
Lightbox

raysuelzer:
This is the new light box, we will want to battle test this a bit on a few browsers before pushing it to release.   Maybe we should make a staging branch for features instead of just using master?


pellaea:
I've created a lightbox branch on origin and cherry-picked these 7 commits.  (Probably a better way to do it!  But it worked...)

raysuelzer:
Great.  That works for me.

PR #176, opened 2015-09-06
Lightbox

pellaea:
This is Ray's work, probably should have let him make the PR.  I just wanted a forum to discuss the feature...


pellaea:
This looks and feels great, Ray.  Thanks!

raysuelzer:
I have not tested it in the various themes.   But, I think it is an improvement.  We might be able to lump the theater.js file into another js file as all it does now is position the expand button.

raysuelzer:
Hmmm... I made a change to this, I noticed I had not removed some of the old theater box HTML which started showing up on the main page.  Should I put in a PR into the lightbox branch?

mo-nathan:
Closing.  I'm pretty sure this is duplicated by a later pull from Ray that doesn't have conflicts.

PR #179, opened 2015-09-06
Comment Box and CSS Fixes

raysuelzer:
Fixes some of the most egregious CSS ugliness in the non black-and-white themes.  Specifically, killing ugly borders and making sure that the active button in the btn group at the top isn't the same color as the body background.  

This also tweaks the comment boxes to be less ugly. Closing: https://www.pivotaltracker.com/story/show/98641918


pellaea:
This I will merge now, soon as I take a quick peak at how you did it...

PR #180, opened 2015-09-07
adds infinite scroll to main page

raysuelzer:


mo-nathan:
Nice improvement, but the language edit links at the bottom of the page are inaccessible when this is on.  I also noticed that the pagination links show up briefly and then vanish.  I think they shouldn't be rendered.  I also have a vague memory that this was going to be a user preference so the curmudgeons can turn it off if they want to.

app/controllers/ajax_controller.rb
```diff
@@ -56,6 +56,45 @@ def test
     render(:text => 'test')
   end
 
+
+  #Used by infite scroll
+  def index_rss_log
+    User.current = get_session_user
+    if !params[:type].blank?
```
JoeCohen: I suggest:
`unless params[:type].blank?`

app/controllers/ajax_controller.rb
```diff
@@ -56,6 +56,45 @@ def test
...
+    end
+
+    show_selected_rss_logs(query, id: params[:id].to_s, always_index: true)
+  end
+
+  #used by infinite scroll
+  def show_selected_rss_logs(query, args = {})
+    #store_query_in_session(query)
+    set_query_params(query)
+    args = {
```
JoeCohen: I suggest running Rubocop on this file to see how Rubocop would like these lines indented.

app/controllers/ajax_controller.rb
```diff
@@ -56,6 +56,45 @@ def test
...
+            species_list: [:location, :user]
+        }
+    }.merge(args)
+
+    @user = get_session_user
+    @types = query.params[:type].to_s.split.sort
+    @links = []
+
+    show_index_of_objects(query, args)
+    render(template: "/observer/list_rss_logs", layout: nil)
```
JoeCohen: I think you can shorten this to 
`render("/observer/list_rss_logs", layout: nil)`
See http://guides.rubyonrails.org/layouts_and_rendering.html, 2.2.3 Rendering an Action's Template from Another Controller.

app/helpers/application_helper.rb
```diff
@@ -124,6 +124,13 @@ def make_line(row, td_opts)
   #
   def reload_with_args(new_args)
     uri = request.url.sub(/^\w+:\/+[^\/]+/, "")
+
+    # In the infnite scroll javascript, we want to make sure that the pagination html
+    # which is returned does not lead to the ajax controller.
+    if request.url.include? "ajax"
```
JoeCohen: These 3 lines can be shortened to 
`uri = request.url.sub(/ajax/, "observer") if request.url.include? "ajax"`

app/helpers/application_helper.rb
```diff
@@ -124,6 +124,13 @@ def make_line(row, td_opts)
   #
   def reload_with_args(new_args)
     uri = request.url.sub(/^\w+:\/+[^\/]+/, "")
```
JoeCohen: What about the following in place of lines 126-132 (so that `uri` is only ever defined once):
```
if request.url.include? "ajax"
  uri = request.url.sub(/ajax/, "observer")
else
  uri = request.url.sub(/^\w+:\/+[^\/]+/, "")
end
```

raysuelzer: Thanks done.

pellaea:
Where does this stand -- aside from conflicts! :)  Does it still need user testing?  Looks like Nathan's issues were resolved(?)

orbea:
I just want to ask to please make this a optional feature if implemented, in every site I have seen infinite scroll it has caused far more problems and page load than has been worth. 

A good example is google images, when the google scripts are blocked with firefox addons like noscript or umatrix I get the classic and reliable system where you load individual pages. If I allow the google scripts I get an infinite page load which is extremely slow and problematic, it renders poorly with firefox themes and is generally ugly.

I do recognize that the current page system on MO is lacking as it only offers jumping a few pages ahead or "Next" and "Previous" buttons. I would suggest to instead of this idea to implement a simple feature that allows users to jump to any page existing page number which they could type.

Anything else is already handled well with the advanced search function.

pellaea:
Good feedback, thanks!  The feature has already been implemented on a code
branch, but has bogged down in internal discussion about the pros and cons
of the various implementation details, including not least some of your
concerns.

mo-nathan:
Closing since the concerns raised have not been addressed.

PR #182, opened 2015-09-07
Image vote fix

raysuelzer:
I added a title tag that will show the number of votes on mouse over of the vote meter.  Might not be the greatest solution, but it gives that information back to the user.


JoeCohen:
Sorry to ask again, but how will this affect performance?
In particular, what will it do to the rss log if users are displaying lots
of log entries?

raysuelzer:
That is a good question for Jason.  But, I did not notice any change at all in the load time on my machine.  It looks like image.num_votes on gets its information from some sort of hash and isn’t making extra db calls.  But, I’m not sure entirely sure what magic is happening in the background.

mo-nathan:
I couldn't measure any performance difference on the home page with 150 images loading.  Personally, I find waiting for the title to appear a bit annoying and would really like to have back the popup with the details about the voting, but this is an improvement so I'm accepting it.

PR #187, opened 2015-10-12
Observer ID

JoeCohen:
Conditionally displays Observer ID on Observation pages

* Delivers [Pivotal #104964854](https://www.pivotaltracker.com/story/show/104964854)
* Requires `rake db:migrate` (because of new user option) and `rake lang:update`
* If User opted to View Observer ID, shows Observer's ID if:
  * Observer has a single name preference, different from consensus, and not "Fungi"; and  Observer is sufficiently confident.
  * "sufficiently confident" is defined in the Vote model in owner_id_min_confidence
  * As a starting point, I set it to the minimum_positive_vote. 
* Please note:
  * There's a fair bit of refactoring of show_observation view files, so there's potential for conflict with pending Pull Requests.
  * Within the code, Observer ID is referred to as "owner_id".


JoeCohen:
Please do NOT merge yet:  Coveralls reports slight reduction in coverage (a couple of lines missed).  I will fix this morning.

JoeCohen:
OK to merge.

JoeCohen:
Oops.  Manual testing has revealed some appearance/logic issues.  I will re-open the PR once I fix them.

JoeCohen:
Ready to go again. Issue was appearance/styling issue (rather than logic). 
___
Partial screenshot of Observation without showing Observer ID:
screen shot 2015-10-13 at 07 47 32
___
Partial screenshot of Preferences, showing View Observer ID and help:
screen shot 2015-10-13 at 07 53 58
___
Same Observation, showing Observer ID:
screen shot 2015-10-13 at 07 48 30

mo-nathan:
Thanks for looking at the coverage as well as reducing the overall rubocop violations!

app/helpers/application_helper.rb
```diff
@@ -41,6 +41,21 @@ def indent(w=10)
     "<span style='margin-left:#{w}px'>&nbsp;</span>".html_safe
   end
 
+  # Wrap content in tag if condition is true, else nil
+  # 1st arg is condition, which wraps content_tag
```
mo-nathan: In my view these functions are clear enough that they don't need comments.  Not a blocker for accepting the pull.

app/helpers/interest_icons_helper.rb
```diff
@@ -0,0 +1,85 @@
+  # Draw the cutesy eye icons in the upper right side of screen.  It does it
```
mo-nathan: New file has 8 rubocop "offenses".  I recognize that you're just moving code, so again not a blocker.  Personally I try to clean things up when I make moves like these.

app/helpers/tabs_helper.rb
```diff
@@ -1,107 +1,88 @@
...
-  # This will cause the set of three icons to be rendered floating in the
-  # top-right corner of the content portion of the page.
-  #
-  def draw_interest_icons(object)
-    if @user
-      type = object.type_tag
+  # assemble HTML for "tabset" for show_observation
+  # actually a list of links and the interest icons
+  def show_observation_tabset(obs, user)
+    tabs = []
```
mo-nathan: Couldn't this just be:

    tabs = [google_images_and_dist_map_links(obs),
            general_questions_link(obs, user),
    ...

app/helpers/tabs_helper.rb
```diff
@@ -1,107 +1,88 @@
...
-      end
+    { pager_for: obs, right: draw_tab_set(tabs.reject(&:empty?)) }
+  end
 
-      def interest_tab(img1, img2, img3)
-        content_tag(:div, img1 + safe_br + img2 + img3, class: "interest-eyes")
-        # content_tag(:div, img1 + safe_br + img2 + img3, style: "position: absolute; top: 0;")
-      end
+  def google_images_and_dist_map_links(obs)
+    return unless obs.try(:name).known?
```
mo-nathan: I don't understand what real case you are guarding against with the 'try' here.  An observation with a name is broken so I would let it raise an error.  I would also create a local for name rather than dereferencing it three times in the method.  I'd also move the link calculation to another function or even a method on name.  Finally, given that you are using capture/concat, I wouldn't bother with the "+ safe_br".  Something like:

    def google_images_and_dist_map_links(obs)
      name = obs.name
      return unless name.known?
      capture do
        concat link_to(:google_images.t, google_link(name))
        concat safe_br
        concat link_with_query(:show_name_distribution_map.t,
                               controller: :name, action: :map, id: name.id)
      end
    end

    def google_link(name)
      "http://images.google.com/images?q=#{name.real_text_name}"
    end

mo-nathan: Actually you could go even one step further and simply pass in the name.

app/models/observation.rb
```diff
@@ -345,12 +345,59 @@ def owners_votes
...
+    result
+  end
+
+  # list of Owners's "favorite" votes for this Observation
+  def owner_favorite_votes
+    votes.where(user_id: user_id, favorite: true)
+  end
+
+  def owner_favorite_vote
+    owner_favorite_votes[0]
```
mo-nathan: I prefer .first over [0] in a case like this.  [0] gives me the willies since in many languages it's an error if the list is empty.  In Ruby it's OK, so this isn't a blocker.

app/models/observation.rb
```diff
@@ -345,12 +345,59 @@ def owners_votes
...
+    namings.each do |n|
       if v = n.users_vote(user)
         result << v
       end
     end
-    return result
+    result
+  end
+
+  # list of Owners's "favorite" votes for this Observation
```
mo-nathan: I'd remove most of the one line comments in this section.  The names say the same thing and we're guaranteed they are correct.  Chapter 4 in book "Clean Code" (PDF here: http://vitoex.googlecode.com/svn/trunk/Read/Clean%20Code.pdf) is what crystalized my current views on comments.

app/models/user.rb
```diff
@@ -931,9 +931,10 @@ def self.erase_user(id)
   # Does user have any unshown naming notifications?
   # (I'm thoroughly confused about what role the observation plays in this
   # complicated set of pages. -JPH)
+  # 2015-10-12 - JDC as currently written, observation appears to be lint
```
mo-nathan: I wouldn't change this file as part of the PR since these changes are entirely in comments.  I'd be OK with it as a separate tiny PR, but it would be even better if it resolved whatever the confusion is here.  I'm also not sure what you mean by "lint" in this context.

app/models/vote.rb
```diff
@@ -125,6 +125,11 @@ def self.maximum_vote
     MAXIMUM_VOTE
   end
 
+  # minimum owner's vote needed to display owner_id
+  def self.owner_id_min_confidence
+    min_pos_vote
```
mo-nathan: This seems a bit odd to me given that min_pos_vote is just returning a constant.  I've been tend towards eliminating constants like that and just returning the value from the function.  At this point I would probably replace all the ALL_CAPS constants with function that simply return the number.  Then your code could stand as it is since conceptually it's a distinct value from min_pos_vote.  However, none of this is blocking accepting the pull as it is.

app/views/account/_prefs_appearance.html.erb
```diff
@@ -0,0 +1,49 @@
...
+
+  <%= label_tag(:user_location_format, :prefs_location_format.t + ":") %>
+  <%= form.select(:location_format,
+                  [[:prefs_location_format_postal.l, :postal],
+                   [:prefs_location_format_scientific.l, :scientific]],
+                  class: "form-control") %><br/>
+
+  <%= label_tag(:user_theme, :prefs_theme.t + ":") %>
+  <%= form.select(:theme,
+                  [
```
mo-nathan: I'd get rid of the extra whitespace similar to the previous lines.  It sure feels like the code is repeating itself, but I don't see a quick, clean way to fix that.

app/views/naming/_show.html.erb
```diff
@@ -36,19 +42,27 @@
...
+                  ) if naming.votes.any? %>
               <%= "(#{naming.votes.any? == 0 ? :show_namings_no_votes.t : naming.votes.length})" %>
             </td>
             <td>
               <% if logged_in %>
                 <% @vote = @votes[naming.id]
-                menu = Vote.confidence_menu
-                if check_permission(naming) ? (!@vote || @vote.value == 0) : true
-                  menu = [Vote.no_opinion] + menu
+                   menu = Vote.confidence_menu
```
mo-nathan: I think github is indicating that there are tabs rather than spaces on these three lines.

mo-nathan: Nevermind.  Misreading github.

app/views/observer/show_observation.html.erb
```diff
@@ -1,76 +1,46 @@
...
     <div class="list-group">
       <div class="list-group-item">
-        <%= render(partial: "show_name_info", locals: {observation: @observation}) %>
+        <%= render(partial: "show_name_info",
+                   locals: { observation: @observation }) %>
       </div>
     </div>
-    <%= render(partial: "show_lists", locals: {observation: @observation}) if show_lists %>
+    <% if show_lists %>
+    <%= render(partial: "show_lists",
```
mo-nathan: I'd intend these two lines.

test/controllers/application_helper_test.rb
```diff
@@ -21,4 +21,39 @@ def test_add_args_to_url
     assert_equal("/blah\x80",
                  add_args_to_url("/blah\x80", x: "foo\xA0"))
   end
+
+  def test_textile_markup_should_be_escaped
```
mo-nathan: Why this test?  I try to avoid tests of third party tools since we have no control of them.

test/controllers/application_helper_test.rb
```diff
@@ -21,4 +21,39 @@ def test_add_args_to_url
...
+  	textile = "**Bold**"
+  	escaped = "&lt;div class=&quot;textile&quot;&gt;&lt;p&gt;&lt;b&gt;Bold&lt;/b&gt;&lt;/p&gt;&lt;/div&gt;"
+		assert_equal escaped, escape_html(textile.tpl)
+  end
+
+  def test_conditional_content_tags
+    assert_nil(content_tag_if(nil, :p, "Hello world!"))
```
mo-nathan: I don't really like sets of tests like this since the first one to fail stops all the other ones to not get run which may reveal more.  The general recommendation I've heard is one assert per test.  That's not currently true in our code base by a long shot since I didn't know that when the project got started, but I've been moving more an more in that direction.

test/controllers/observer_controller_test.rb
```diff
@@ -545,6 +545,56 @@ def test_show_observation
...
+                  id: obs.id)
+    assert_select("div[class *= 'owner-id']",
+                  { text: /#{obs.owners_only_favorite_name.text_name}/,
+                    count: 1 },
+                  "Observation should show Observer ID")
+
+    # observer_id on, observation with 1 owner favorite == consensus
+    get_with_dump(:show_observation,
+                  id: observations(:owner_only_favorite_eq_consensus).id)
+    assert_select("div[class *= 'owner-id']", { count: 0 },
```
mo-nathan: Again, I'd rather see one assert per test.

test/controllers/observer_controller_test.rb
```diff
@@ -1013,7 +1062,7 @@ def test_create_observation_with_new_herbarium
     assert(herbarium.is_curator?(rolf))
   end
 
-  def test_create_simple_observation_with_approved_unique_name
+  def test_create_simple_observation_with_approved_only_name
```
mo-nathan: Not sure why you're making this change in this PR, but "approved_only_name" seems awkward to me.  Maybe "approved_single_name"?

test/models/observation_test.rb
```diff
@@ -88,6 +87,37 @@ def test_name_been_proposed
     assert(!observations(:coprinus_comatus_obs).name_been_proposed?(names(:conocybe_filaris)))
   end
 
+  def test_vote_counts
+    obs = observations(:owner_only_favorite_eq_consensus)
+    assert_equal(0, obs.users_votes(users(:rolf)).count)
```
mo-nathan: It's not clear to me why this test cares that Rolf hasn't voted on this observation.  I'd rather see obs.user (than users(:dick)) compared to obs.owner_favorite_votes.count.

test/models/observation_test.rb
```diff
@@ -95,10 +125,23 @@ def test_specimens
     assert(observations(:detailed_unknown).specimens.length > 0)
   end
 
+  def test_observer_takes_general_email_questions_from_viewer
+    user_a = users(:rolf) # rolf takes general email questions
```
mo-nathan: It seems more intuitive if this was user_a = user_a_obs.user (obviously with the statements switched).

tmp/name_primer.test
```diff
@@ -1,6 +0,0 @@
-Agaricus campestras
```
mo-nathan: Hmmm, not clear why this file changed.

JoeCohen:
Temporarily closed while I fix things up in response to Nathan's thorough code review.
Will repoen when fixed.

app/helpers/tabs_helper.rb
JoeCohen: 1. `return unless obs.try(:name).known?` is my rewriting of master branch show_observation.html.erb, line 13, `if name && !name.unknown?`.  I was also puzzled about the first condition, and thought it might be best to throw an error.  But I eventually decided not to monkey with it.  Perhaps @pellaea knows?  Maybe at one time there was a Name fixture without a name? But now all tests pass if the first condition is removed, so I will remove it.

pellaea: Not I.  I never saw the "obj.try(:method)" trick.  I use "obj.method rescue
nil" instead (if I'm being lazy).

JoeCohen: What  about the predecessor:  if name && !name.unknown?

*--Joe Cohen*

pellaea: Yeah, all three are similar... but I think (1) is the most clear and robust.

1) return unless obs && obs.name && obs.name.known?
2) return unless obs.name.known? rescue false
3) return unless obs.try(:name).known?

What happens in (3) if obs.name is nil for example?  And how is (2)
parsed?  Does it return if it throws an error, or does the statement
evaluate to false and continue to the next line?

JoeCohen: I think Nathan thinks an error should be thrown if observation.name is
nil.  But neither obs.try(:name).known? or if name && name.unknown? will do
that.  So, was there a reason to avoid throwing an error in that situation?

pellaea: Yeah, maybe I just have a different philosophy than Nathan on this.  If
that throws an error, no one above it will catch it, and the end user will
see a blank page with "oops! something went wrong" on it.  99.9% of the
time the user couldn't care less if the google image and distribution links
are working.  This doesn't seem optimal.  There are no doubt times when
obs.name being nil is a serious problem that prevents any useful content
from being displayed, but in this case it just blocks the most incidental
of links most users aren't even aware of.  The result will just be to drop
the links.  That sounds correct to me.  But that's just one opinion.
(However, it is the opinion of the guy who actually gets and reads and
responds to all the "no method defined for nil" errors! :)

JoeCohen: Uh oh.  To my mind, what @mo-nathan says is reasonable, and what @pellaea says is reasonable.  I'm kind of stuck in the middle.  
Unless you two come to an agreement, my inclination is to write the simplest code; so that will throw an error in the (extremely unlikely) event that observation.name is somehow nil. 
Another option would be to log it and not throw an error.  Better for the user, but the same amount of work for you.  Moreover, if there's an observation with no name, it's likely that some other part of the code will throw an error anyhow.

pellaea: Yes, that's what I figure, too.  This is not the place to get whiny about
nils.  There are other pages, like show_observation, where that should
happen.

mo-nathan: I've realized that what actually bothers me about this function is that it can't decide if it's about the observation or the name.  If you want to check that observation.name is nil you should do that in a separate function from where you are essentially operating on the name.  I also just realized that this function as written is violated the 'do one thing' rule.  Even the name of the function has an 'and' in it.  You could do:

    def google_links(obs)
        return unless obs.try(:name).known?
        name = obs.name
        capture do
            concat name.google_image_link
            concat safe_br
            concat name.google_map_link
        end
    end

(I did not check the above for correctness)

JoeCohen: Thanks!  I'll catch on ... sooner or later.  
Question:  Should  `google_map_link` be moved to the name model?

pellaea: No, Name should only contain stuff exclusively about names.  Google URLs
belong in either a google module somewhere (if we needed or had one) or,
next best thing, a helper or view.  That's where stuff dealing with HTML or
details of presentation in general belongs.

app/models/observation.rb
JoeCohen: `.first` is clearer, so I will change it to `.first`.  Indeed, when I did the 3.2 => 4.0 upgrade, I changed a bunch of `[0]`s to `.first`s.  I used `[0]` hear because I recently read that it's faster than `.first`.  See https://github.com/JuanitoFatas/fast-ruby?utm_source=rubyweekly&utm_medium=email. I guess that my personal recency bias.

app/helpers/tabs_helper.rb
mc-nathan: Jason is right.  It is better to keep the models as simple as possible (just providing data accessors and such) and keep the HTML stuff in another location.  Another case of old bad habits emerging.  It does seem like we have enough Google stuff that we could separate it out into it's own thing.  A better approach might be to just consider Google one instance of third party URL providers and have a module for that.  Given that there are clearly two distinct examples, that's enough these days for me to think it should go in its own module especially since I know we use a Google services in other places and we have the links to MycoPortal etc.  So you might end up with something name_image_link(:google, name) and name_map_link(:google, name).  The :google stuff could even be a backend configuration option so you'd end up with just name_image_link(name), name_map_link(name) and leave it up to those functions to decide who the provider is going to be.  Sort of depends on how far you want to go.

app/models/observation.rb
mc-nathan: In my view clarity is far more important than squeezing out the last last sub-millionth of a second.

app/helpers/tabs_helper.rb
JoeCohen: I don't want to go that far right now, but I will add it to Pivotal.

app/models/observation.rb
JoeCohen: Will remove most comments in this section.  
Please feel free to remove more or to restore any which I removed;  someone who did not write the code is probably better judge of whether the code is self-explanatory.

app/models/user.rb
JoeCohen: 1.  Will delete my comment.  
2.  I meant "lint" in the same way as used by Rubocop.  
Here, `has_unshown_naming_notifications?(observation=nil)` doesn't seem to do anything with its argument.
3. I'd prefer not to fix it as part of this PR, as fixing it seems unrelated, and I think I've already put too much stuff into this PR.  I will create a Pivotal Issue so we can deal with it in due course.  (Moreover, I cannot resolve the confusion right now because I'm also confused about the intent of the method.  Is it trying to determine if the user has (a) any unshown naming notifications, or (b) any unshown naming notifications for a single observation?

app/models/vote.rb
JoeCohen: Also seems odd to me.  I prefer the functional approach.  
But as fixing this will take changes to multiple files and it's not a PR blocker, I will leave as-is and create a Pivotal story.

app/views/account/_prefs_appearance.html.erb
JoeCohen: Removed extra white space (blank lines) within the div.  I think that's what you meant, but the way in which GitHub displays comments makes it unclear.  Perhaps you meant only remove the one instance of a double blank line.

test/controllers/application_helper_test.rb
JoeCohen: This is a pre-existing test (see https://github.com/MushroomObserver/mushroom-observer/commit/a38d6cc69e94f4eb7eebc5a852c77f8af4eb37d7) which I moved to tests/**controllers**/application_helper_test.rb from test/**helpers**/application_helper_test.rb.  
The name which I gave to the test is awful.  It implies that that what's being tested is Textile.  But it's really testing MO's escape_html helper method (or maybe the MO `tpl` extension which lives in app/extensions/string_extensions.rb.
I have to think about this.  I will at least find a better name for the test.

test/controllers/application_helper_test.rb
JoeCohen: Will change to 1 assertion/test

app/views/account/_prefs_appearance.html.erb
mo-nathan: To be specific, I'd do something like:

    <%= form.select(:theme,
                    [[:theme_random.l, "NULL"]] +
                    MO.themes.map { |t| [t.to_sym.l, t] },
                    class: "form-control") %><br/>

test/controllers/application_helper_test.rb
mo-nathan: Just to be clear, the 1 assertion/test is a good rule, but it doesn't always make complete sense.  The real rule is a given test should test one thing.  If you have more than one assertion in a test it's worth making sure it doesn't violate the real rule.  Examples where more than one assertion is fine are cases where more than one thing happen as the result of a single action.  For example you may want to test that an action created a particular objects and the it redirected to a particular page.  Another time may use additional assertions in a single test is if I want to make sure that some precondition is setup before the action of the test happens.

test/controllers/observer_controller_test.rb
JoeCohen: I'm also unsure why/how this changed.  Will revert.

test/models/observation_test.rb
JoeCohen: Will do 1 assertion/test and rewrite.  For instance as:
```ruby
  def test_observer_accepts_general_email_questions
    obs_by_user_taking_questions = observations(:coprinus_comatus_obs)
    assert(obs_by_user_taking_questions.
             observer_takes_email_questions_from?(users(:dick)),
           "User with email_general_question should take questions from others")
  end
```
I hope that's also somewhat more intuitive.
But it's arbitrary that the owner of the observation accepts general emails.
I think the only way to make it completely obvious would be some db queries setting up the assertion. 
Something like (pseudocode):
```ruby
obs = Observation.where(email_general_question: true)
viewer = User.where.not(obs.user.id)
assert(obs.observer_takes_email_questions_from?(viewer),
           "User with email_general_question should take questions from others")
```
But I don't want to hit the db in what should be a unit test.

test/models/observation_test.rb
JoeCohen: It's also not clear to me.  As best I can reconstruct events, I wrote this to test either (a) my changes to fixtures, and/or (b) a change to `users_votes` which I was contemplating, but never implemented.  See [MO Developers › Loop vs Finder](https://groups.google.com/forum/#!topic/mo-developers/uBfhRMGbV3k).
So I will delete the test.

tmp/name_primer.test
JoeCohen: Don't remember why this file changed.  I think at some point (maybe after some combination of: migration, switching branches, trying to reset the db) Git refused to let me make commits unless I did something about the file.  I think I removed it.
Perhaps more important:  (a) why are these changes showing up? -- I thought everything in this directory was outside of version control; (b) do changes to files in this directory matter?

test/models/observation_test.rb
mo-nathan: Since we're using static fixtures, the way to do this is create a separate observation fixture for this test that has a relevant name.  At some point we should consider moving to Factory Girl factories rather than static fixtures.  They essentially let you do what you want to do here without undo database cost.

JoeCohen:
Ready to merge after changes in response to code review

Manual testing script:
* Create _user a_ (or use a preexisting user as _user a_) and login as _user a_.
* Create _observation a_ with _name a_ as What and with Confidence Level "Could Be"
* Observation title should be "Observation" _n_: _name a_ like:
image
* Create _user b_, login, show _observation a_, propose <name b> with Confidence Level "I’d Call It That", cast "As If" vote for _name a_.  Name in Observation title should change to _name b_.
* Logout, login as _user a_. In Preferences check "View Observer identification of Observations?" 
Observation title should change to add Observer ID and designate Site ID, like:
image
* Play with votes to see if title changes appropriately.

test/models/observation_test.rb
JoeCohen: Got this note about
observation_test#test_observer_takes_general_email_questions_from_viewer
just after I re-opened the PR, and while Travis was running.
I will re-write the test and push it up, so there will be more msgs from
Travis.  May not happen until tonight Pacific Time.
Writing fixtures can be a royal pain, but there seems to be lots of
controversy about Fixtures vs. Factories. See e.g.:
http://brandonhilkert.com/blog/7-reasons-why-im-sticking-with-minitest-and-fixtures-in-rails/
https://whatdoitest.com/getting-friendly-with-fixtures
https://semaphoreci.com/blog/2014/01/14/rails-testing-antipatterns-fixtures-and-factories.html
https://www.google.com/search?q=rails+fixtures+versus+factories&rlz=1C5CHFA_enUS512US512&oq=rails+fixtures+versus+factories&aqs=chrome..69i57.8767j0j7&sourceid=chrome&es_sm=119&ie=UTF-8

mo-nathan:
I think you're intention for the example is that it should have switch to:

Observation 4: Gastrosuillus (Site ID)
Observer ID: Suillus caerulescens

That's what happened for me and what seems like the right thing to have happened.

I'll accept the pull now.

PR #192, opened 2015-11-09
Convert Relation to array when calling "blacklisted" "mutator" methods

JoeCohen:
- Delivers [Pivotal #107666274](https://www.pivotaltracker.com/story/show/107666274)
- fixes test Error: NoMethodError: undefined method `map!'
at test/models/query_test.rb:16

### DIFF
--- a/test/models/query_test.rb
+++ b/test/models/query_test.rb
@@ -12,6 +12,7 @@ class QueryTest < UnitTestCase
   end
 
   def assert_query(expect, *args)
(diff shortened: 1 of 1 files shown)


JoeCohen:
I checked all MO instances of all the blacklisted mutators.  
Only one was clearly called on a Relation (and I fixed it).
Almost all the rest were clearly not called on a Relation.
There were a couple that I was unsure of, but thought were not called on a Relation.
I did not change them.  If I was wrong, they'll be easy to diagnose and fix.

PR #199, opened 2015-11-23
Do not override maps_collapsible#max_objects for every test in Collap…

JoeCohen:
Do not override max_objects for all tests in CollapsibleMapTest

- Fixes order-dependent Errors thrown by rake test:models TESTOPTS="--seed=42066"
- Prevents overridden definition of maps_collapsible#max_objects from persisting to tests where max_objects it is not explicitly defined
- max_objects is overridden by the class TestCollapsible, so use that class only in tests which must override max_objects
- In other tests, use the superclass of TestCollapsible -- CollapsibleCollectionOfMappableObjects
- Fixes [Pivotal #108723288](https://www.pivotaltracker.com/story/show/108723288)
- For more details see that Story

### DIFF
--- a/test/models/collapsible_map_test.rb
+++ b/test/models/collapsible_map_test.rb
@@ -13,6 +13,8 @@ class BigDecimal
   end
 end
 
+# class for the sole purpose of allowing selected tests to override
+# app/classes/map_collapsible.rb#max_objects
 class TestCollapsible < CollapsibleCollectionOfMappableObjects
   class << self
     attr_accessor :max_objects
@@ -360,7 +362,7 @@ class CollapsibleMapTest < UnitTestCase
   def test_mapping_one_observation_with_gps
     obs = observations(:amateur_observation)
     assert(obs.lat && obs.long && !obs.location)
-    coll = TestCollapsible.new(obs)
+    coll = CollapsibleCollectionOfMappableObjects.new(obs)
     assert_equal(1, coll.mapsets.length)
     mapset = coll.mapsets.first
     assert_mapset_is_point(mapset, obs.lat, obs.long)
@@ -373,7 +375,7 @@ class CollapsibleMapTest < UnitTestCase
   def test_mapping_one_observation_with_location
     obs = observations(:minimal_unknown)
     assert(!obs.lat && !obs.long && obs.location)
-    coll = TestCollapsible.new(obs)
+    coll = CollapsibleCollectionOfMappableObjects.new(obs)
     assert_equal(1, coll.mapsets.length)
     mapset = coll.mapsets.first
     assert_mapset_is_box(mapset, *obs.location.edges)
@@ -385,7 +387,7 @@ class CollapsibleMapTest < UnitTestCase
 
   def test_mapping_one_location
     loc = locations(:albion)
-    coll = TestCollapsible.new(loc)
+    coll = CollapsibleCollectionOfMappableObjects.new(loc)
     assert_equal(1, coll.mapsets.length)
     mapset = coll.mapsets.first
     assert_mapset_is_box(mapset, *loc.edges)


mo-nathan:
The PR works, but highlighted some style things that I didn't like.  Specifically it seems really weird to me to be creating a child class of a production class just so you can override something in the production class and to have a method in the production class specifically to allow this.  So I changed the original class so it takes max_objects as argument to init and it all got a lot simpler.  I also cleaned up the tests a little bit.

JoeCohen:
1.  That also seemed odd to me.  I thought about eliminating the subclass,
but I no longer remember why I didn't do that.  Your solution is
(obviously) much cleaner.

2. After looking into the history of this some more, it seems that  the
tests themselves can be marginally simplified by omitting the max_objects
argument:
For instance, instead of

coll = CollapsibleCollectionOfMappableObjects.new(observations,
observations.length)

do

coll = CollapsibleCollectionOfMappableObjects.new(observations)

Looks to me like there's a maximum number of objects because of problems in
development mode (as opposed to production or testing) .
Might need to configure MAX_MAP_OBJECTS. My local server truncates the page
if I do more than 100.

I'm not sure why this is, because it can get away with 1000+ on the
production server.  Bear in mind,
though, that we have to use webrick locally, while we use passenger in
production.

See
https://github.com/MushroomObserver/mushroom-observer/commit/2167c637682e8486eb6885e6b5bbb20d04d6b612
And, in any event, omitting the argument in the tests will work because all
the the tests have far fewer than 100 objects.  So the tests are okay
simply using the default.
We would only need to supply the argument when we want to test mapping >
100 objects.

3.  And if we're going to limit it to 100 objects in production, why would
we want to test > 100?  Maybe just have a hard limit that cannot be
overridden???

mo-nathan:
It seemed to me that the argument is controlling how much collapsing gets done, hence resulting in different data.  I haven't tried it, but I would guess if you removed the max_objects argument, then the existing tests would fail.

JoeCohen:
The tests do pass if the max_objects argument is removed.
@max_objects is used in one place -- a `while` loop in MapCollapsible#group_objects_into_sets.  See below.  The inside of that loop does not get executed for the tests as currently written because -- for all tests -- @sets.length <= max_objects.
So, at the moment, the overriding of max_objects has no effect and the tests pass without a second argument to CollapsibleCollectionOfMappableObjects.new.
I'm ignorant enough so I don't understand the intent of the relevant part of tests or the model.  So maybe one and/or the other is broken.
```
  def group_objects_into_sets
    prec = next_precision(MAX_PRECISION)
    while @sets.length > @max_objects && prec >= MIN_PRECISION
      old_sets = @sets.values
      @sets = {}
      for set in old_sets
        add_box_set(set, set.objects, prec)
      end
      prec = next_precision(prec)
    end
  end
```
For all the tests, if

pellaea:
There are client-side packages that handle mapping huge numbers of
points/polygons nowadays, I think.  (See the excellent interface CNALH
has developed over top of google maps, for example... if only ours was
so slick! :)  But at the time I couldn't find any.  This server-side
code is what I came up with.

It groups nearby objects into a "box set" if there are too many
objects.  And it continues to do this, bumping the precision (i.e.
loosens the definition of "nearby") each time, until there are
sufficiently few objects to display safely without causing a browser
embolism.  The "real work" is presumably done in add_box_set, which I
assume must modify @sets.  Some better (if longer!) names might be:

group_objects_into_sets ->
group_nearby_objects_into_sets_if_there_are_too_many_objects
add_box_set -> add_set_of_objects_to_sets_while_merging_nearby_objects

My apologies if I've missed other emails directed at me.  I can't
begin to read all the recent MO-related email in the time I have
available.  I've just deleted most of them wholesale in order to
preserve some semblance of sanity. :(  Please try writing to me
directly with a catchy subject line if it's something you want me to
see...

-Jason

mo-nathan:
I'm not sure what you mean.  I just tried removing the second argument from the two relevant tests (test_mapping_a_bunch_of_points and test_mapping_a_bunch_of_points_straddling_date_line) in collapsible_map_test.rb and they both failed (see below).

I do see that I missed some of the calls to TestCollapsible, but I'm not sure if you are about to check that in or not.

I did get to the bottom of the Nokogiri issue so we're down to 8 failures and 38 errors.

-Nathan

Started with run options --seed 63083

 FAIL["test_mapping_a_bunch_of_points_straddling_date_line", CollapsibleMapTest, 2015-11-24 20:35:25 +0000]
 test_mapping_a_bunch_of_points_straddling_date_line#CollapsibleMapTest (1448397325.53s)
        Mapsets are wrong: expect -vs- actual
           0.0000    0.0000  175.0000  175.0000       0.0000    0.0000  175.0000  175.0000  10.1000   10.0000 -175.0000 -175.1000      10.0000   10.0000 -175.0000 -175.0000 (*)  20.0000   20.0000 -165.0000 -165.0000      10.1000   10.1000 -175.1000 -175.1000 (*)  20.0000   20.0000 -175.0000 -175.0000      20.0000   20.0000 -165.0000 -165.0000 (*)  22.0000   22.0000 -167.0000 -167.0000      20.0000   20.0000 -175.0000 -175.0000 (*)  70.0000   70.0000 -145.0000 -145.0000      22.0000   22.0000 -167.0000 -167.0000 (*) -10.0000  -10.0000 -175.0000 -175.0000      70.0000   70.0000 -145.0000 -145.0000 (*) -12.0000  -12.0000 -177.0000 -177.0000     -10.0000  -10.0000 -175.0000 -175.0000 (*) -90.0000  -90.0000 -135.0000 -135.0000     -12.0000  -12.0000 -177.0000 -177.0000 (*)                                            -90.0000  -90.0000 -135.0000 -135.0000 (*)                                                                                   
        test/models/collapsible_map_test.rb:90:in `assert_list_of_mapsets'
        test/models/collapsible_map_test.rb:464:in `test_mapping_a_bunch_of_points_straddling_date_line'

 FAIL["test_mapping_a_bunch_of_points", CollapsibleMapTest, 2015-11-24 20:35:25 +0000]
 test_mapping_a_bunch_of_points#CollapsibleMapTest (1448397325.59s)
        Mapsets are wrong: expect -vs- actual
           0.0000    0.0000    0.0000    0.0000       0.0000    0.0000    0.0000    0.0000  10.1000   10.0000   10.1000   10.0000      10.0000   10.0000   10.0000   10.0000 (*)  20.0000   20.0000   10.0000   10.0000      10.1000   10.1000   10.1000   10.1000 (*)  20.0000   20.0000   20.0000   20.0000      20.0000   20.0000   10.0000   10.0000 (*)  22.0000   22.0000   22.0000   22.0000      20.0000   20.0000   20.0000   20.0000 (*) -10.0000  -10.0000   10.0000   10.0000      22.0000   22.0000   22.0000   22.0000 (*) -12.0000  -12.0000   12.0000   12.0000     -10.0000  -10.0000   10.0000   10.0000 (*) -70.0000  -70.0000  -30.0000  -30.0000     -12.0000  -12.0000   12.0000   12.0000 (*) -90.0000  -90.0000   50.0000   50.0000     -70.0000  -70.0000  -30.0000  -30.0000 (*)                                            -90.0000  -90.0000   50.0000   50.0000 (*)                                                                                   
        test/models/collapsible_map_test.rb:90:in `assert_list_of_mapsets'
        test/models/collapsible_map_test.rb:416:in `test_mapping_a_bunch_of_points'

  11/11: [==================================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.89190s
11 tests, 936 assertions, 2 failures, 0 errors, 0 skips
[Coveralls] Outside the CI environment, not sending data.

JoeCohen:
Hmm.  I must have tested something else (or the wrong branch.
I will check in the missing calls.

PR #201, opened 2015-11-26
Use RoR4.2 Selector matchers in LurkerTest

JoeCohen:
- should Finish [Pivotal #108925894](https://www.pivotaltracker.com/story/show/108925894), but I cannot test immediately:
- This fix exposes another Error in the LurkerTest. See [Pivotal #108971510](https://www.pivotaltracker.com/story/show/108971510)


mo-nathan:
Fixes several cases, but there's still one out there.

PR #206, opened 2015-12-04
Enable web console for MushroomObserver VM

JoeCohen:
- Finishes [Pivotal #109534868](#109534868)


mo-nathan:
I think the goal of this PR is to allow us to use the web console gem while running the server off of a Vagrant box.  If that's not right, let me know what the goal is.  In any case, this branch should be moved to go against master and a new PR created.

PR #212, opened 2015-12-24
Web console

JoeCohen:
Enables web console in dev mode on VM

- Takes advantage of new Rails feature which opens console when an error is thrown in dev mode
- Also gets rid of log entries like "Cannot render console from 10.0.2.2! ..."


JoeCohen:
Manual testing script:
- On VM in dev mode:
- Throw an error.  
  - E.g., (1) Search for "Cortinarius" in Specimens, (2) on results page click "Date Created" Tab.  (Or, if you prefer, temporarily some line of code to create an error, then do something to invoke that code.
- The resulting error page should include an interactive web Ruby 
console, for example, the dark part of the attached screen shot.
screen shot 2016-01-10 at 07 44 12

PR #209, opened 2015-12-15
Ask users to check spam folder for verification or new password email.

JoeCohen:
- May avoid some questions to MO webmaster
- Finishes [Pivotal #105363482](https://www.pivotaltracker.com/story/show/105363482)

PR #214, opened 2015-12-26
Add notice re check spam folder for email from MO

JoeCohen:
Add notice re check spam folder for email from MO

- Should avoid some questions to MO webmaster
- Rewrite 7 separate spam notices as one notice, tweak some text to work with rewritten notice. See #209
- Refinishes [Pivotal #105363482](https://www.pivotaltracker.com/story/show/105363482)

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -105,7 +105,7 @@ class AccountController < ApplicationController
           flash_object_errors(@new_user)
         else
           group = UserGroup.create_user(@new_user)
-          flash_notice :runtime_signup_success.t
+          flash_notice (:runtime_signup_success.tp + :email_spam_notice.t)
           VerifyEmail.build(@new_user).deliver_now
           redirect_back_or_default(action: :welcome)
         end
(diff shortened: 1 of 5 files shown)


JoeCohen:
Manual test script:
- If logged in, logout.
- Click "Create Account"
  - the account/signup page should include a notice to check spam folder when looking for confirmation email:
screen shot 2016-01-10 at 08 11 12
- Complete the Account signup page, click Signup.
  - The resulting flash message should include the same notice:
screen shot 2016-01-10 at 08 14 44
- On resulting page, click "Login" **without verifying password**.
- On resulting login page: fill in name, password, click Login
 - Resulting welcome page should include the same notice.
- Click Reverify
 - The flash message should include the same notice.
- Click "Login"
- on the resulting login page, click Email me a new one.
 - resulting Email New Password page should include the notice

app/controllers/account_controller.rb
```diff
@@ -105,7 +105,7 @@ def signup # :nologin: :prefetch:
           flash_object_errors(@new_user)
         else
           group = UserGroup.create_user(@new_user)
-          flash_notice :runtime_signup_success.t
+          flash_notice (:runtime_signup_success.tp + :email_spam_notice.t)
```
pellaea: I'm concerned about mixing xxx.tp and xxx.t.  tp adds a <p> wrapper around the text.  If one needs it, don't both?  Same thing below.

pellaea:
Other than minor quibble about .tp in flash warning, this looks fine to me.

JoeCohen:
Thanks for looking at these.  I will check and re-submit or re-comment.

PR #215, opened 2016-01-11
Always show owner_id line if user opts to show observer preference

JoeCohen:
- Avoids ambiguity/confusion about missing line
  - If no clear owner preference, line says "no clear preference"
  - If consensus = observer preference, shows that name
- Other changes to more clearly state intent of option and code
 - "Observer ID" => "Observer Preference"
 - Re-order and re-name some methods and variables
- I want to merge these changes before working more on this branch

Manual test script (essentially duplicated in ObserverControllerTest)
- While logged in, goto Preferences, check "View Observer’s Preferred ID?", Save Edits
- Click Create Observation, Fill in form with:  
  - Where: any existing location
  - What: **leave blank**
  - Confidence: Could Be
  - Create
- Observation should have Observer Preference line, which says "no clear preference"
screen shot 2016-01-10 at 16 32 01
- Propose a name:  _Coprinus comatus_, Confidence: Could Be
  - Result:
screen shot 2016-01-10 at 16 46 47
- Propose another name:  _Agaricus_, Confidence: Could Be
  - Result:  no clear preference (because of tie vote)
- Destroy Proposed Name _Agaricus_;  Edit Proposed Name  _Coprinus comatus_ Confidence: Doubtful
   - Result:  no clear preference (because confidence is negative)
- Change Preferences to **un**check "View Observer’s Preferred ID?", Save Edits, 
- In browser go back to Observation and refresh
 - Result: no Observer Preference line

### DIFF
--- a/app/helpers/show_observation_helper.rb
+++ b/app/helpers/show_observation_helper.rb
@@ -28,7 +28,7 @@ module ShowObservationHelper
     return unless obs.show_owner_id?
     capture do
       concat(:show_observation_owner_id.t + ": ")
-      concat(obs.owners_only_favorite_name.format_name.t)
+      concat(obs.owner_favorite_or_explanation.t)
     end
   end
 end
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -364,12 +364,12 @@ class Observation < AbstractModel
     result
   end
 
-  def owner_favorite_votes
-    votes.where(user_id: user_id, favorite: true)
-  end
-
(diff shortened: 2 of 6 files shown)


JoeCohen:
I will fix test Failure.  But first I will open an issue.  (When I test locally everything passes, but that's a bug.)

JoeCohen:
Tests passing.

pellaea:
Looks good to me.

PR #228, opened 2016-06-13
Advanced fixtures

JoeCohen:
Refactor fixtures and tests to use advanced fixtures.

See generally http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
Delivers [Pivotal #110805910](https://www.pivotaltracker.com/story/show/110805910).
Change fixtures as follows:
- Replaces manual fixture IDs with autogenerated (stable) IDs.
- Uses label references instead of manual fixtures numbers for associations (belongs_to, has_one, has_many)
- Creates HABTM associations as inline lists instead of HABTM fixtures.

Change tests as required to work with above.


JoeCohen:
Temporarily closing PR while I investigate Coverage decrease (7 lines) and hopefully restore coverage.

PR #234, opened 2016-07-22
Test MinimalMapObservation.

JoeCohen:
Improves test coverage.  (MinimalMapObservation had 0% coverage.)

Manual Test Script:  n/a.

### DIFF
--- a/test/models/observation_test.rb
+++ b/test/models/observation_test.rb
@@ -150,6 +150,10 @@ class ObservationTest < UnitTestCase
 
     min_map.location = locations(:albion)
     assert_equal(locations(:albion).id, min_map.location_id)
+
+    min_map.location = false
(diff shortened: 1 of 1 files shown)


JoeCohen:
I can't figure out how to test, or make sense of, MinimalMapObservation#initialize lines13-15:
```
    elsif location
      @location = location_or_id
      @location_id = location_or_id.id
```
If I'm understanding Ruby correctly: During initialization, location is always undefined.  Thus the `elsif` condition is always false, and the nested code is never executed.
Maybe what was intended was `elsif location_or_id.is_a?(Location)`.

pellaea:
Agreed, Joe, that's what that second clause in MinimalMapObservation initializer must've been intended to do.  I'm fixing it and adding test clause to your new test...

Issue #222, opened 2016-04-06
MO Species List


orbea:
This is a re-post of a part of a comment made in MO observation 228965.
http://mushroomobserver.org/228965
```
If such lists are to be used at MO I think they should be entirely segregated from 
more usefulobservations. A text file uploadable and viewable on MO in a different
section of the site would probably work well for species lists, they could look like
conventional species lists found at actual forays. If there are any real observations 
associated with them there should be a way to link them. For miscellaneous “Faith” 
based observations perhaps should be further separated, I’m not sure what the best 
way to represent them would be, but they might be useful for the maps, but only if 
they are visibly distinguishable on the map (Different color?) and I would strongly 
suggest they be off by default.

Additionally previous practices such as refusing “Lists” from new, unknown, or users 
with questionable experience would need to stop. I at least think we should accept them 
all or none of them, it should not be something only available to a privileged few. Also 
name changes would become less important for the species lists, but a way to highlight 
clearly wrong ones could nice and perhaps even link better names would be nice. I would 
imagine the lists would be most useful if they represented our knowledge at the time and 
not what we currently know as that is always in flux.
```
Please view the other comments for more background. I am posting this here in hopes that it exposes the MO developers to issues usually only discussed in MO observation comments when they might be more useful to be discussed here.

I hope this or perhaps something expanded from it could be implemented in hopes that "Imageless" observations can be more useful for what they are and out of the way for what they aren't.

orbea:
An additional idea would be to make it possible to directly connect names to species lists without any associated observations so that observations like 93995 can be removed.

orbea:
This should be considered "Won't fix", MO is not a serious mycological organization unfortunately.

PR #237, opened 2016-08-09
Don't enforce imageless.

orbea:
This just opens the door for 1000s of other names for `Imageless` and will get really silly fast.

A better solution will be to be to address issue https://github.com/MushroomObserver/mushroom-observer/issues/222. 

As for observations with a herbarium specimen, if a specimen exists than an image of the dried specimen should also be possible. If its not then that is enough reason to suspect the specimen either does not exist or has been lost.


orbea:
Here is a prime example of why these commits were a bad idea.

http://mushroomobserver.org/246992?q=2oKv8

Part of a species list so that "Imageless" can not be posted, but without any images or even name...
http://mushroomobserver.org/246992?q=2oKv8

pellaea:
This is not the place for debate about MO policy.  In fact I'm a little
tired of getting of hearing nothing but complaints from users who do
nothing to contribute to the development or running of MO.  Please send the
webmasters / admins your comments.  When I'm done with my field season I
promise I'll read them.  Meanwhile it pollutes the one channel of email
that in principle I should be trying to keep up with in my few precious
spare moments.

Besides, the debate about imageless observations was closed years ago.
Nathan made his position crystal clear.  At this point we're just playing a
stupid game with a troll who was walking the line between annoyance and
vandalism.  Thanks to this commit, the troll was forced to cross that line
and can now be removed from the site without any concerns of it being
called censorship.  We've given him his last warning.  As far as I'm
concerned it's over and I will remove all this silliness in a few months
once everything calms down again.  And I have the time to play games again.

Now I'm deleting this "pull request" and hopefully won't have to play
*this* game again, either.

orbea:
Another example...this qualifies under both your rules 3 & 4....

http://mushroomobserver.org/50248?q=2oKv8

pellaea:
Look, they aren't my damned rules.  You accomplish nothing by pestering me
with this issue.  I'm going out of town now, and I hope to hell I don't
have a mountain of Imageless BS in my inbox when I get back.

PR #240, opened 2016-09-11
Fix author standardization; add support for "comb".

JoeCohen:
Add support for "comb", e.g., "comb. nov."  Delivers [Pivotal #127954231](#127954231).

Fix Name author standardization. Delivers [Pivotal #130136909](https://www.pivotaltracker.com/story/show/130136909).

Also standardizes "nov" and "prov" when they are suffixes to "comb" or "nom".

I put these changes in the same PR because they're related and I want to avoid local merge conflicts.

Manual test:
Create a name with Author field: "comb nov" (no periods)
  Result:  The name should be created with author including "comb. nov." (with periods).
Create a name with Author field "nomen".
  Result:  Name should be created with Author "nom."


JoeCohen:
I put these changes in the same PR because they're related and I want to avoid local merge conflicts.

PR #243, opened 2016-10-05
Pkg update 2016 10 03

JoeCohen:
Update gems. Update release notes.


JoeCohen:
Don't know why this Travis build is still failing (other than the explanation that it's not completing).  When I first made the PR, it failed because of a Travis issue with the gem index.  They fixed that, and it still failed because they had not cleared the index cache.  Once Travis cleared the cache, the build did complete in my repo.  Don't know why it's not completing as a PR.
I will see if I can restart the job.  If I can't do that or it continues to not complete, I will close the PR, delete the branch, then try it all as a new branch.

PR #246, opened 2016-12-21
Added ability to add/remove results of observation query to/from list.

pellaea:
Adds link in upper right tab_set of all observation index pages "Add/Remove from Species List".  This takes you to a page which asks you for a species list id or title (auto-completion enabled).  Then you click on either "Add" or "Remove" submit buttons to do the deed.

It will ignore observations which are already in the list when adding.  And it will obviously ignore observations which are not in the list when removing.

It might be nice to actually show the list of observations at the bottom of the "Add/Remove" form, but this seemed unnecessary.

### DIFF
--- a/app/controllers/species_list_controller.rb
+++ b/app/controllers/species_list_controller.rb
@@ -406,11 +406,7 @@ class SpeciesListController < ApplicationController
     pass_query_params
     id = params[:species_list].to_s
     query = find_query(:Observation)
-    if id =~ /^\d+$/
-      species_list = SpeciesList.safe_find(id)
-    else
-      species_list = SpeciesList.where(title: id).first
-    end
+    species_list = lookup_species_list(id)
     flash_error(:species_list_add_remove_no_query.t) if !query
     flash_error(:species_list_add_remove_bad_name.t(name: id.inspect)) if !species_list
     if !query && !species_list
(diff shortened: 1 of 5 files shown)


app/controllers/species_list_controller.rb
```diff
@@ -390,6 +392,53 @@ def destroy_species_list # :norobots:
...
+    pass_query_params
+    @id = params[:species_list].to_s
+    @query = find_query(:Observation)
+    if !@query
+      flash_error(:species_list_add_remove_no_query.t)
+      redirect_to(action: "list_species_lists")
+    end
+  end
+
+  def post_add_remove_observations # :prefetch: :norobots:
```
mo-nathan: I'd like to see this function broken up and I'm a bit concerned that Rubocop is reporting 17 new "offenses".

pellaea:
Oh sure, how long does rubocop allow?

app/controllers/species_list_controller.rb
```diff
@@ -390,6 +392,61 @@ def destroy_species_list # :norobots:
...
+    pass_query_params
+    id = params[:species_list].to_s
+    query = find_query(:Observation)
+    species_list = lookup_species_list(id)
+    flash_error(:species_list_add_remove_no_query.t) if !query
+    flash_error(:species_list_add_remove_bad_name.t(name: id.inspect)) if !species_list
```
mo-nathan: I'd add this flash_error to the local lookup_species_list function.

app/controllers/species_list_controller.rb
```diff
@@ -390,6 +392,61 @@ def destroy_species_list # :norobots:
...
+      redirect_to(action: "list_species_lists")
+    end
+  end
+
+  def post_add_remove_observations # :prefetch: :norobots:
+    pass_query_params
+    id = params[:species_list].to_s
+    query = find_query(:Observation)
+    species_list = lookup_species_list(id)
+    flash_error(:species_list_add_remove_no_query.t) if !query
```
mo-nathan: I'd make a small local function that combines find_query and this flash_error.  This could also be reused in add_remove_observations.

app/controllers/species_list_controller.rb
```diff
@@ -390,6 +392,61 @@ def destroy_species_list # :norobots:
...
+    flash_error(:species_list_add_remove_bad_name.t(name: id.inspect)) if !species_list
+    if !query && !species_list
```
mo-nathan: Switching this to something like:

```
if species_list && query:
  do_add_remove_observations(species_list, query)
else:
  determine_redirect(species_list, query)
end
```

would be cleaner and clearer in my view.

PR #247, opened 2016-12-21
Have vote select menus automatically submit new vote.

pellaea:
Only affects javascript in vote_popup.js.  (Tweaked a comment in ajax controller, too.)

Should disable affected select menus and prevent leaving page until success confirmed.  If there is an error it will reset the menus to their previous state.

### DIFF
--- a/app/assets/javascripts/vote_popup.js
+++ b/app/assets/javascripts/vote_popup.js
@@ -29,18 +29,47 @@ function VotePopupModule(showNamingsLostChangesText) {
     });
 
     $change_vote_selects.change(function (event) {
-      var _this = $(this); //references the vote being changed
-      _haveVotesChanged = true;  //votes have been changed
-      getUserAttention();
+      var _this = $(this);
+      var value = _this.val();
+      var naming_id = _this.data("id");
+      _haveVotesChanged = true;
+      _this.attr("disabled", "disabled");
+
       // If setting vote to 3.0, go through all the rest and downgrade any
       // old 3.0's to 2.0.  Only one 3.0 vote is allowed.
-      if (_this.val() == "3.0") {
-        var changedNamingId = _this.data().id;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/vote_popup.js
```diff
@@ -29,18 +29,47 @@ function VotePopupModule(showNamingsLostChangesText) {
...
-      if (_this.val() == "3.0") {
-        var changedNamingId = _this.data().id;
+      if (value == "3.0") {
         $change_vote_selects.each(function () {
-          if ($(this).data().id != changedNamingId && $(this).val() == "3.0")
-            $(this).val("2.0");
+          var _this2 = $(this);
+          if (_this2.data("id") != naming_id && _this2.val() == "3.0") {
+            _this2.val("2.0");
+            _this2.attr("disabled", "disabled");
```
mo-nathan: I note that _this2 never gets "disabled" set to null like _this does.  Could this be a problem?

pellaea:
This logic is not perfect, but not in the way you think... I think. :)
After either error or success it will re-enable all the menus again,
regardless whether they were disabled or not.  But if the user has managed
to submit *another* vote in the meanwhile, then it will re-enable that
request's disabled menus prematurely.  Hmmm.  I should probably have it
just disable everything to restrict it to one vote at a time, shouldn't I?

mo-nathan:
I like the idea, but it's still not clear to me if this achieves that.  It now looks like it only disables if there is a second votes that's getting down voted.  I feel like I'm missing something.

pellaea:
Hmmm.  Maybe I mucked it up, but I just pushed a change which disables
everything.  I can't run it because of a migration pending in the user
highlighting PR.  Grrr.  I'll ask how to get around this problem in another
email.  I must be doing something wrong with git...

PR #248, opened 2016-12-22
Added user highlighting

pellaea:
Ripped out users.created_here - an accidental hold-out from transactions which never got used.

Added some logic to Comment#notify_users which checks for highlighted users (_user login_, @login and @Full Name@).

It considers this a "comment response" in case someone would like to prevent notifications in these cases.


mo-nathan:
This PR also increases the rubocop "offenses", but I'm OK with accepting it as is.  We might want to discuss this and come up with plan for how to deal with this in the longer term.

pellaea:
I'm working on understanding what rubocop wants.  Phew it's draconian!  But
I'm "getting" it.  Should at least, as so many other places, at least
ensure forward progress.

PR #251, opened 2016-12-23
Response to naming email

pellaea:
Very simple feature.  Just makes it so users who have proposed names but not bothered to comment will also get "comment response" emails when someone comments on the observation.  (I personally will find this invaluable!!)


pellaea:
Oops, but need to adjust a few tests apparently, one moment...

mo-nathan:
rubocop report 51 offenses in the files that were touched:

% git diff --name-only master | xargs rubocop
...
3 files inspected, 51 offenses detected

At lot of those would be really easy to clean up (lines too long, indents etc.).  The hard ones include the "Assignment Branch Condition size" and "Class has too many lines" offenses.  It would be great if the easy ones could get cleaned up and we can worry about the hard ones later.

mo-nathan:
BTW, I merged with master and pushed those changes so I could see more clearly what the new stuff is.

mo-nathan:
Hmm, got a failing test:

```
 FAIL["test_comment_notification", CommentTest, 156.52414682600647]
 test_comment_notification#CommentTest (156.52s)
        Expected: 89
          Actual: 90
        test/models/comment_test.rb:79:in `test_comment_notification'
```

pellaea:
This has to be wrong, I couldn't possibly have caused 51 new offences, I
only added one method that's 3 lines long!  And I should point out that
app/models/comment.rb is one of the only files that actually passes without
*any* warnings.  (After several hours of work the other day.)

[Yes, just checked, there are no new rubocop warnings in either comment.rb
or comment/callbacks.rb -- the only files in the code I might have touched
-- except for a request for comments which already exist in the main
Comment class which I don't know how to silence because it's clearly a
bogus warning, and it's complaining about a blank line I just added for
readability, but which I'd be happy to remove, didn't occur to me that
might be a problem.]

Your diff method didn't work for me at all.  If you add a line of comments
at the top of the file it claims all the offences in that file are new even
if you didn't touch any of the rest of the file.

I am happy to ensure that my new code follows rubocops dubious protocols
but I can't be responsible for cleaning up all the warnings in any file I
touch -- I would never get anything done and there would be a huge
activation energy to making even the tiniest contribution to the site.

-Jason

pellaea:
Something must be off.  Both my tests and the automated tests (coveralls?
travis? whatever that service is called) passed fine.

mo-nathan:
I'm not particularly worried about the rubocop offenses.  I was just hoping to leave the campsite a bit cleaner.  I'll dig into the test failure.  It's probably my fault.

mo-nathan:
Looks like the test failure is order dependent.  If I run the file by itself I get no errors.  However, if I use either the seed 36259 or 38763 I get the failure.  For example,

```
$ rake TESTOPTS="--seed=38763"
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's 'rails' settings.
Started with run options --seed=38763

 FAIL["test_comment_notification", CommentTest, 145.35456999001326]
 test_comment_notification#CommentTest (145.36s)
        Expected: 29
          Actual: 30
        test/models/comment_test.rb:79:in `test_comment_notification'

  1316/1316: [===============================================================] 100% Time: 00:02:40, Time: 00:02:40

Finished in 160.22179s
1316 tests, 17896 assertions, 1 failures, 0 errors, 0 skips
[Coveralls] Outside the CI environment, not sending data.
```

I can't dig into this tonight and tomorrow may be tricky.  After that I'll dig into it if you haven't already found the problem.

pellaea:
Thanks for that, hopefully I can reproduce it with your seeds.  It's mine,
I'll take care of it.

PR #257, opened 2016-12-26
Disable other value unless radio button is selected.

pellaea:
And cleaned up / standardized syntax, removed tabs from donate view.

(Sorry, Nathan, broke my proposed rule of not mixing cleaning and coding.  Didn't take long to prove myself hypocritical...)

### DIFF
--- a/app/assets/javascripts/donate.js
+++ b/app/assets/javascripts/donate.js
@@ -1,9 +1,8 @@
 $(window).load(function () {
-  var other = $("#donation_other_amount").attr("disabled", "disabled");
-  $("[name='donation[amount]'").change(function () {
-    other.attr("disabled", "disabled");
+  $("#donation_other_amount").click(function () {
+    $("#donation_amount_other")[0].checked = true;
   });
-  $("#donation_amount_other").change(function () {
-    other.attr("disabled", null);
+  $("#donation_other_amount").keyup(function () {
+    this.value = this.value.replace(/[^0-9]/g,'');
   });
 });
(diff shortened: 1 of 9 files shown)


mo-nathan:
This is enough to get me going in the right direction.  I don't want to merge it as is since I've realized that it already refers to a JS file, donate.js, that isn't currently in the code base (but maybe it used to be?).  That seems like a logical place to put this new code.  Also while the new interaction is better, it seems like we may as well go a bit further such as making it so that clicking on the text box moves the radio button selection rather than it being disabled and maybe the input should be constrained to be numeric.

pellaea:
So, this may be pathological, but it is possible to circumvent your onclick event on other-value.  You can _tab_ into the text field, then enter an amount and submit the form.  You could, I suppose, watch onchange for that text field, too.

Sorry to be negative, but I can also see confusion if someone quickly types "100.00" and submits then finds themselves surprised when MO wants to deduct $10K from their bank account(!)  It's kinda fun to play with this callback which prevents you from typing anything except digits, but it's a bit odd.  Brilliantly simple!  But feels strange to me personally.

Feel free to ignore both complaints.

mo-nathan:
The recurring donations are handled by PayPal.  We don't have to do anything.

I also think the other concerns are also reasonably addressed by the PayPal confirmation process.  However, if you can think of an easy way to make these better, that would be great.  However, I think this is ready to ship as it is and I want to get it out before the end of the year so folks can use it if they want.

pellaea:
Agreed, I'm fine with paypal catching any (extremely rare) stingers.  Glad
this is out.  Progress, whee!

PR #258, opened 2016-12-27
Add user option for persistent content filters.

JoeCohen:
Gives user the option - via Preferences - to always exclude imageless or unvouchered Observations from Activity Feed and search results. 
(Can override via search bar filters or Advanced Search checkboxes.)

Basic Manual Test Script
* Create 4 Observations, as follows
  (a) one with an image and a specimen;
  (b) one with an image but without specimen;
  (c) one with a specimen but without images;
  (d) one without images or specimen.
* Refresh Activity Feed.

Result: all 4 Observations should be shown.

- Turn on has_images persistent filter: in /account/prefs, check the box next to “Hide Observations with no images.”, then Save Edits
- Refresh Activity Feed.

Result: Of those 4 Observations, only (a) and (b) should appear.  “(filtered)” appears next to page title. 

- Turn off has_images persistent filter, turn on has_specimen filter, Save Edits.
- Refresh Activity Feed.  

Result: Of those 4 Observations, only (a) and (c) should appear. (filtered)” appears next to page title. 

- Turn on both filters, Save Edits.
- Refresh Activity Feed.  

Result: Of those 4 Observations, only (d) should appear. (filtered)” appears next to page title. 

Note: this is a simple script.  These filters persist site-wide for all Observation queries.  For instance, with any of these filters on, from any /name/show_name/nnnnnn page, click More Observations.  The results are filtered (and that is shown in the page title).


mo-nathan:
There appear to be 23 new rubocop offenses under app/.  Specifically when on the branch after running the recommended GitHub command line instructions, I ran:

rubocop `git diff --name-only master | grep '^app/.*\.rb$'` > branch.cop

At the end of branch.cop it says:

13 files inspected, 1097 offenses detected

Then back on master and accounting for the newly added file, I ran:

rubocop `git diff --name-status JoeCohen-content-filters | grep '^M.*app/.*\.rb$' | sed 's/^M\t//'` > master.cop

At the end of master.cop it says:

12 files inspected, 1074 offenses detected

I haven't dug into what all the new ones are, but at least some of them are easy, e.g.,

```
app/classes/api/observation_api.rb:43:25: C: Align the parameters of a method call if they span more than one line.
                        limit: Range.new(Vote.minimum_vote, Vote.maximum_vote)),
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
app/classes/content_filter.rb:3:1: C: Extra empty line detected at module body beginning.
app/classes/content_filter.rb:35:1: C: Incorrect indentation detected (column 0 instead of 2).
< # There are probably other steps/files I've forgotten. JDC 2016-09-15
< ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
< app/classes/content_filter.rb:51:7: C: Rename has_images to images?.
<   def has_images
<       ^^^^^^^^^^
< app/classes/content_filter.rb:64:7: C: Rename has_specimen to specimen?.
<   def has_specimen
<       ^^^^^^^^^^^^
< app/classes/content_filter.rb:72:21: C: Use %w or %W for an array of words.
<       on_vals:      ["TRUE", "FALSE"],
<                     ^^^^^^^^^^^^^^^^^
< app/classes/content_filter.rb:74:4: C: Indent the right brace the same as the start of the line where th\
e left brace is.
<    }
<    ^
172,177d155
< app/classes/observation_report.rb:283:21: C: Unnecessary spacing detected.
<         has_specimen          = clean_boolean(row[OBS_SPECIMEN])
<                     ^^^^^^^^^
< app/classes/observation_report.rb:283:31: C: Operator = should be surrounded by a single space.
<         has_specimen          = clean_boolean(row[OBS_SPECIMEN])
<                               ^
```

Then a simple diff starts failing due to added lines in term.rb.

mo-nathan:
Functionality looks great!  It will be interesting to see how many folks actually set the user preference.  I like having it in the Advanced Search.

app/classes/content_filter.rb
```diff
@@ -0,0 +1,98 @@
...
+# * Supplement ObservationController#advanced_search_form as needed.
+# * Supplement ApplicationController#show_index_of_objects as needed.
+# * Supplement Query::RssLogBase as needed.
+#
+# To also add it as a Pattern Search
+#   Supplement PatternSearchTest
+#   Supplement classes/pattern_search/observation.rb or -- if it's not an
+#     Observation search -- add a new pattern_search file and class.
+#
+# There are probably other steps/files I've forgotten. JDC 2016-09-15
```
mo-nathan: Not needed for this PR, but if we find we are adding a lot of these it would be worth figuring out how to simplify the above process.  In theory at least a new filter could be added by just creating a new class and having it register itself with a list of available filters for the various UI pieces to use as they want.

JoeCohen: @mo-nathan -
Thanks so much for taking the time to look at this so thoroughly.
- I'll check the Rubocop offenses (I suspect that most are preexisting) and fix at least the easier ones. 
- I agree with that the process should be simplified, but not needed for this PR. (I couldn't figure out how to make it any simpler; I thought it was best to push it out now to minimize future conflicts and make the functionality available. It can be refactored in the future if more content filters are added, and the additional filters should help make for a better refactoring.)

app/controllers/application_controller.rb
```diff
@@ -1161,13 +1163,9 @@ def show_index_of_objects(query, args = {})
...
 
-#     # Apply content filter to any queries which are capable of being filtered.
-#     if query.respond_to?(:observation_filters)
-#       filter_params = @user ? @user.content_filter : MO.default_content_filter
-#       query.params.merge!(filter_params)
-#     end
+    update_filter_status_of(query)
 
-    # Tell site to come back here on +redirect_back_or_default+.
+    # Tell site to come back here on +redirect_back_or_default+.rr
```
mo-nathan: 'rr'??

app/controllers/application_controller.rb
```diff
@@ -1310,6 +1308,32 @@ def show_index_of_objects(query, args = {})
...
+      !query.has_obs_filter_params?
+  end
+
+  # Apply user defaults if they exists, else apply site-wide default.
+  def apply_default_filters_to(query)
+    default_filters = @user ? @user.content_filter : MO.default_content_filter
+    query.params.merge!(default_filters) if default_filters
+  end
+
+  def supply_list_of_filters_on_for(query)
```
mo-nathan: I don't understand the name of this method.

app/models/user.rb
```diff
@@ -114,6 +114,11 @@
 #  locale::             Language, e.g.: "en" or "pt"
 #  theme::              CSS theme, e.g.: "Amanita" or +nil+ for random
 #  layout_count::       Number of thumbnails to show in index.
+#  view_owner_id::      View Observation author's ID on Obs page
+#
+#  ==== Content filter options
+#  has_images::         Exclude imageless Observations from search results and
```
mo-nathan: Shouldn't has_specimens be here as well?

app/views/account/_prefs_filters.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<div class="form-group push-down">
+  <span class="bold"><%= :prefs_content_filters.t %></span>
+  <p><%= :prefs_content_filters_explanation.t %><p>
+</div>
+
+<!-- Observation Filters -->
+<div class="form-group indent push-down">
+  <span class="bold"><%= :prefs_obs_filters.t %></span><br>
+  <span><%= :prefs_obs_filters_explanation.t %>.</span>
+  <%= render partial: "prefs_obs_filter_checkboxes",
```
mo-nathan: +1  I like the generality of this UI and the one for Advanced Search.

app/views/observer/_form_observations.html.erb
```diff
@@ -98,16 +98,16 @@
           }) if include_naming %>
   </div>
 </div>
-<!-- HAS_SPECIMENS -->
-<% if has_specimens %>
+<!-- has_specimen -->
+<% if has_specimen %>
```
mo-nathan: Why the switch from plural to singular?  Observations can have multiple specimens (e.g., split collections in different herbaria).

app/views/shared/_title_and_tab_sets.html.erb
```diff
@@ -21,7 +21,9 @@
 
     <!-- Title Caption -->
     <!-- e.g. "Observation 5"; Can be multiple lines, e.g., with Observer ID -->
-    <%= content_tag(:div, @title, id: "title-caption") %>
+    <%= content_tag(:span, @title, id: "title-caption") -%>
+    <%= content_tag_if(@on_obs_filters.present?, :span, "(#{:filtered.t})",
```
mo-nathan: This should at least have some help text associated with it explaining that it is filtered either by the user's preference or as part of the Advanced Search interface.

db/migrate/20160829175601_add_content_filter_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentFilterToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :content_filter, :string
```
mo-nathan: This should really be handled by a separate object where many of them can point to a given user.

app/classes/content_filter.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  ### filter definitions ###
+  # In the order you want filters to appear in advanced_search
+  # name:         filter name, as string
+  # sym:          filter mame, as symbol
```
JoeCohen: Yes. Either one of them is sufficent.

app/classes/content_filter.rb
```diff
@@ -0,0 +1,98 @@
...
+  # model:        model on which filter operates.  Used by ContentFilter
+  # checkbox:     prefs form checkbox, e.g., :has_images_checkbox.
+  #               Used by prefs view
+  # checked_val:  value when checkbox checked
+  # off_val:      value when filter is off
+  # on_vals:      array of allowed values when filter is on,
+  #               in order you want them to appear in advanced_filters
```
JoeCohen: Interesting point. For the moment, I'd like to allow non-binary filters.
I originally envisioned these as non-binary and originally implemented them that way. E.g., a user could persistently choose to exclude non-vouchered Observations, to exclude vouchered Observations.
@AlanRockefeller and @pellaea convinced me that there wasn't a good use case for persistently excluding vouchered (or imaged), so I ratcheted these two back to binary.
But I keep thinking about other non-binary possibilities, e.g., a location filter based on continent, country, etc.

app/classes/content_filter.rb
```diff
@@ -0,0 +1,98 @@
...
+  # sql_cond:     predicate added to Query "where", when filter is on.  E.g.:
+  #               "observations.specimen IS #{params[:has_specimen]}"
+
+  def has_images
```
JoeCohen: I like the idea; I couldn't figure out how to do it.

app/classes/content_filter.rb
```diff
@@ -0,0 +1,98 @@
...
+   }
+  end
+
+  def filters
+    [has_images, has_specimen]
+  end
+
+  ### These reflect on above methods.
+  def observation_filters
+    filters.select { |fltr| fltr[:model] == Observation }
```
JoeCohen: My thought was these two particular filters would work only on Observations, but other other filters might work on other models (e.g., location filters would also work on locations).  
Probably premature generalization.

pellaea:
For whatever it is worth, while I was momentarily shocked by the complexity
of your solution, I quickly came to realize that if you want generality,
then I can't come up with a fundamentally better way of doing this, either.

I think the way forward is to chip away here and there as we identify
patterns that will allow inference and reduction of duplication.  Nathan
already has several good ideas along these lines.

I definitely like Nathan's suggestion of turning the big hashes into
objects.  That will allow significant improvements in time.  (Look how much
it helped Query!)

I second Joe's comment about string-based filters some day being added --
location and taxon group both come to mind immediately.

I think there is no foreseeable value to anticipating filtering other
objects.  I think MO is fundamentally about observations, and they perform
a central role in everything from site UI to query structure and database
structure.  (Examples: Every major object hangs off of the observations
table one way or another.  Almost all "coercion" involves converting
queries to and from queries on observations.)  My vote would be to lose
this generality if it simplifies things significantly, and just assume all
persistent user content filtering be associated with observations.

re: has_specimens -> has_specimen -- This is my fault.  The Query filter
parameter is "has_specimen", because it was written back before Herbarium
and Specimen classes existed.  I wish there were good ways to standardize
pluralization, but these sorts of head-scratchers seem to be inevitable in
any old, evolving code base. :(

Good work, Joe.  Thanks for persisting with this.  I would like to wait for
you to merge this into master before I gum things up even more with all my
recent refactoring.  It just wouldn't be fair to make you keep playing
catch up.  (And this latest work last night was a doozie.)

-Jason

app/controllers/application_controller.rb
JoeCohen: Will see if I can find better name.

pellaea:
This is definitely my fault.  Shall we just arbitrarily choose one set?
Initially I figured it would be harmless for Query and the API to accept
all those alternatives, no harm there.  (They are all internally translated
into honest true and false boolean values by Query validation.)  But we
should definitely choose a single consistent set for internally-generated
filters.  If it makes it less confusing all around, then let's remove the
flexibility of Query and API validators.  If it breaks someone's code,
hopefully they will forgive us and find it easy to fix.

app/models/user.rb
JoeCohen: Quite right.  Will add.

app/views/shared/_title_and_tab_sets.html.erb
JoeCohen: Will think about how to implement this.

db/migrate/20160829175601_add_content_filter_to_user.rb
JoeCohen: 1.  Not sure I understand. is the suggestion that there's a class for each type of filter, each class having one instance per user?
2. I thought a single serialized column might better anticipate a wider variety of future developments:
On Tue, Aug 2, 2016 at 1:36 AM Jason Hollinger <pellaea@gmail.com> wrote:
...
Again, I think it's worth looking ahead to other uses of this mechanism.  For example, instead of a boolean column in users for this one filter, it could be a serialized hash of arbitrary parameters and values.  For example it might read "{ has_images: yes }" for users who wish to hide imageless observations.  Or Adolf might like "{ has_images: yes, has_specimen: yes }".

app/classes/content_filter.rb
JoeCohen: Also interested to see who uses it. I suspect it will see little use, and that the users who have complained the most about unvouchered and/or imageless Observations will ignore it or find some excuse not to use it.

JoeCohen:
@pellaea: I will look at these values and see if I can simplify.

db/migrate/20160829175601_add_content_filter_to_user.rb
pellaea: I think I agree with Joe on this point.  This is a perfect example of a single complex piece of data attached to the user which you will never need to search or filter on.  The only time anyone ever looks at this is when you need to filter content or adjust the filter parameters.  I see no advantage in making a particular combination of filter settings a separate stand-alone entity which can be shared by multiple users.  Instead I think it is more natural to think of this as a description of a filter.  Each user can describe their own unique filter.  It always seemed to me that a serialized hash would be the natural implementation.  Especially since a hash like that can be passed in verbatim to any observation query.  Then it's just a matter of creating a generalized and extensible UI for adjusting this hash of filter parameters appropriately.

mo-nathan: I'm ok with a serialized hash (but not thrilled by it).  However, if we really expect folks to use this in more complex cases, then varchar 255 seems kind of small and it's not clear what would happen if the serialized hash happens to be more than 255 characters long.  Maybe this should be a text field like the other fields in User.

app/classes/content_filter.rb
mo-nathan: Location based filter is a great example.  There have definitely been requests for that (Martin Osis of the Alberta club comes to mind).  However, I'm not sure that gets handled well by "on_vals".  It seems like a different type of filter that has a regular expression of some sort or maybe a lat/long and a distance.  Filtering by taxon would also be really nice.  The more I think of examples like this the more I think that users will want to switch between filters dynamically.  They will also want them to be in the URL so they can share them with folks or put them on their local club website etc.  However, all of that is for another day!

config/locales/en.txt
```diff
@@ -1629,6 +1637,7 @@
   rss_log_title: Activity Log
   rss_by: by [user]
   rss_title: "[:app_title]"
+  rss_filtered_mouseover: "[:app_title] has applied filter(s), including any filter(s) in: (a) your Preferences and (b) Advanced Search or the search bar."
```
mo-nathan: Sorry to be a pest about this, but the "(a) ... and (b) ... or ..." is hard for me to parse.  How do filters get applied through the search bar?  I must have missed that.  Would it work to just say "[:app_title] has applied filter(s), including any filter(s) in your preferences or search parameters."?

JoeCohen: Your help suggestion is perfect & easy to implement; will do.  
(In search bar, `images:` and `specimen:` apply filters.)

db/migrate/20160829175601_add_content_filter_to_user.rb
JoeCohen: Good point about field length. But will leave this alone for now (unless you want me to deal with it now).

config/locales/en.txt
JoeCohen: Done. 59508e1358cd875c4fdf43bd36bbae5834ee0d74

mo-nathan:
Tested the mouse over text.  All tests pass.  Rubocop offenses have gone down by 51.  Merging!

PR #259, opened 2016-12-28
Added checks to make sure all "Show <Type>s" links are actually valid.

pellaea:
Added checks for all :show_objects links to ensure that the present query is coercable to the appropriate type.
Cleaned two minute rubocop offences in app/model/comment/callbacks.rb.
(I added no additional tests to ensure these new checks are working, in a hurry.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -924,6 +924,19 @@ class ApplicationController < ActionController::Base
     url_for(add_query_param(args))
   end
 
+  def coerced_query_link(query, model)
+    return nil unless query && query.is_coercable?(model.name.to_sym)
+    link_args = {
+      controller: model.show_controller,
+      action: model.index_action
+    }
+    return [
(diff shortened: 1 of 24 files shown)


JoeCohen:
Looks good.  
Is it worth DRYing this a bit with new helper method(s) something like
query_coercable_to?(query, new_model)
add_link_with_coercable_query_to_tab(query, new_model)

pellaea:
Sure, that shouldn't be a problem

JoeCohen:
Will you do that, or do you want me to?

pellaea:
I've pushed the branch to origin.  Do you have permission?  If not, I'd be
happy to.  Was just looking into refactoring Query#coerce and how it can
possibly be kept DRY.

JoeCohen:
Not sure if I have permission. Maybe you should go ahead.

pellaea:
Will do.

pellaea:
Merged with PR 260 (refactor-coerce).

PR #261, opened 2016-12-30
External links prototype.

pellaea:
1) Added external_links and external_sites tables.
2) Added section for external_links on show_observation page under species_lists.
3) Added ajaxy controls for creation, editing and deletion of external_links.
4) Presently the owner of the observation can create links to any external_site; only members of the project associated with an external_site can add links to that site to arbitrary observations.
5) There is no API for grabbing these links yet.


pellaea:
In order to test this PR, one needs to create a project for MycoPortal, and an entry in the external_sites table (by hand unfortunately).  Then you should see controls for adding a link to MycoPortal on any observations you own.

JoeCohen:
If I'm understanding this correctly:here could be multiple external sites which actually go to the same place.

app/classes/query/initializers/observation_filters.rb
```diff
@@ -3,67 +3,40 @@ module Initializers
...
-        end
-      end
-
-      # Does params[:x] == one of x's on_vals?  For example:
-      # is_on?(:has_images) is true if params[:has_images] == "NOT NULL" || "NULL"
-      def is_on?(filter_sym)
-        return unless params[filter_sym]
-        filter = send(filter_sym)
-        filter[:on_vals].include?(params[filter_sym])
+        !params[:has_specimen].nil? || !params[:has_images].nil?
```
JoeCohen: maybe
```params[:has_specimen] || params[:has_images]```

File: app/classes/content_filter.rb [hunk 2]
```diff
@@ -1,96 +1,117 @@
...
+  #
+  #   Add tests, e.g., to TestUserContentFilter
+  #   Supplement fixtures as needed by added tests
+  #   Add a filter definition below, and add that definition to #filters below
+  #   Add a new checkbox_val method in User
+  #   In /config/locales/en.txt define text to be displayed before and next to
+  #     check boxes, using prefs_obs_filters_has_images: as a model.
+  # * Supplement _prefs_filters.html.erb as needed
+  # * Add filter to AccountController#prefs_types & #update_content_filter
+  # * For Observation filter, supplement Query::Initializers::ObservationFilters
```
JoeCohen: No longer done automatically. (Developer must add observation_filter_parameter_declarations manually.) So asterisk should be deleted.
```#   For Observation filter, supplement Query::Initializers::ObservationFilters```

File: app/classes/query/modules/validation.rb [hunk 3]
```diff
@@ -21,27 +21,42 @@ def validate_params
...
+    begin
+      val = pop_param_value(old_args, arg_sym)
+      unless val.blank?
+        if arg_type.is_a?(Array)
+          val = array_validate(arg_sym, val, arg_type.first)
+        else
+          val = scalar_validate(arg_sym, val, arg_type)
+        end
+      end
+      if !val.nil?
```
JoeCohen: What about
```if val```

pellaea:
This has changed a bit in the latest version of content filters.

pellaea:
The whole procedure of adding a new filter is changed much in the newer
versions of content filters.

pellaea:
Presumably this was to differentiate param == false from param == nil(?)  I
would like to clean all this up with a refactoring of
Query::Modules::Validation.  I think rubocop would faint if we were to run
it on this file.

mo-nathan:
I resolved the conflicts, but now two tests are failing and I don't understand why.  I'm not sure where this work got left in particular with respect to Joe's concerns about potentially problematic uses of the new feature.

PR #262, opened 2017-01-02
Revised content filters

pellaea:
See email.


app/classes/content_filter.rb
```diff
@@ -1,96 +1,117 @@
...
 
-  def filters
-    [has_images, has_specimen]
+  # Array of all available filters.
+  def self.all
+    @@filters ||= [
+      ContentFilter.new(
+        sym:          :has_images,
+        model:        Observation,
+        checkbox:     :has_images_checkbox,
```
mo-nathan: In my view (and I think this agrees with a comment you made in email) this object shouldn't know about checkboxes.  I'd rather the UI generated the name/symbol :has_images_checkbox based on the filter's sym.

app/classes/content_filter.rb
```diff
@@ -1,96 +1,117 @@
...
   end
 
-  def filters
-    [has_images, has_specimen]
+  # Array of all available filters.
+  def self.all
+    @@filters ||= [
+      ContentFilter.new(
+        sym:          :has_images,
+        model:        Observation,
```
mo-nathan: What's the current opinion about filters being relevant to anything other than Observations?  It seemed to me that we were converging on the opinion that these are only relevant to Observations.

app/classes/content_filter.rb
```diff
@@ -1,96 +1,117 @@
...
-    observation_filters.select { |fltr| fltr[:checkbox].present? }
+  # Array of observation filter names as symbols.
+  def self.observation_filter_keys
+    observation_filters.each_with_object([]) { |fltr, keys| keys << fltr.sym }
   end
 
-  module_function(:has_images, :has_specimen, :filters, :observation_filters,
-                  :observation_filter_keys)
+  # Array of filters which use checkboxes.
+  def self.observation_filters_with_checkboxes
```
mo-nathan: Do we have a motivating example for a filter without a checkbox?  I could see this once we get to location or taxon based filters, but it's not clear to me what happens at that point and whether the has_images and has_specimen filters remain special.

app/classes/query/initializers/observation_filters.rb
```diff
@@ -3,67 +3,40 @@ module Initializers
...
-      end
-
-      # Lets ApplicationController check whether to add default filters to
-      # current query; defaults are added unless query already has
-      # observation filter params, even if those filters are off.
-      # This allows for overriding of default filters.
-      def has_obs_filter_params?
-        observation_filter_keys.any? {|k| params[k] != nil}
+        {
+          has_specimen?: :boolean,
```
mo-nathan: Shouldn't this be leveraging ContentFilter.observation_filters() or a related method?

app/classes/query/initializers/observation_filters.rb
```diff
@@ -3,67 +3,40 @@ module Initializers
...
-        end
-      end
-
-      # Does params[:x] == one of x's on_vals?  For example:
-      # is_on?(:has_images) is true if params[:has_images] == "NOT NULL" || "NULL"
-      def is_on?(filter_sym)
-        return unless params[filter_sym]
-        filter = send(filter_sym)
-        filter[:on_vals].include?(params[filter_sym])
+        !params[:has_specimen].nil? || !params[:has_images].nil?
```
mo-nathan: Again should be leveraging something from ContentFilter.

app/classes/query/initializers/observation_filters.rb
```diff
@@ -3,67 +3,40 @@ module Initializers
...
-      # array of literal sql conditions to be included in query
+      # Array of literal sql conditions to be included in query.
       def obs_filter_sql_conds
-        on_obs_filters.each_with_object([]) do |filter_sym, conds|
-          filter = send(filter_sym)
-          conds << filter[:sql_cond]
+        conds = []
+        unless params[:has_specimen].nil?
+          val = params[:has_specimen] ? "TRUE" : "FALSE"
+          conds << "observations.specimen IS #{val}"
```
mo-nathan: I agree with keeping the SQL separate from ContentFilter, but this should be derived based on a property of the ContentFilter rather than being explicit for each key.

app/classes/query/modules/validation.rb
```diff
@@ -21,27 +21,42 @@ def validate_params
...
+          val = scalar_validate(arg_sym, val, arg_type)
+        end
+      end
+      if !val.nil?
+        new_args[arg_sym] = val
+      elsif !optional
+        fail("Missing :#{arg_sym} parameter for #{model} :#{flavor} query.")
       else
-        val = scalar_validate(arg_sym, val, arg_type)
+        new_args[arg_sym] = nil
```
mo-nathan: Given that you are failing a line above this, I believe this line is dead.

pellaea:
re: observations only -- See use case in other email.  Filtering out
lichens should apply to names as well as observations.

re: ContentFilter shouldn't know about checkboxes -- Agreed.  I still don't
have a concrete recommendation for an alternative.  Maybe record the data
type in the ContentFilter instead?  Then account/prefs and
observer/advanced_search can implement UI however they see fit?

re: non-checkbox filter -- Location is the only compelling use case so far.

re: Query::ObservationFilters leveraging ContentFilter -- Arg, yes, it
should, but I got really confused when I tried to do revise this file.  We
can definitely automate the declarations.  And the
"any_observation_filters_on?" methods.  But I couldn't figure out how to
automatically deduce in obs_filter_sql_conds that one boolean filter is
implemented with a "column IS TRUE" or "column IS FALSE" condition, and the
other (syntactically identical!) boolean filter is implemented with a
"column IS NOT NULL" or "column IS NULL" condition.  In the end it added a
large amount of complexity for virtually no gain.  Feel free to disagree.
I wasn't happy with how this file turned out.  The best I can think of id
create a "has_images_sql_condition" and "has_specimen_sql_condition" and
have obs_filter_sql_conds automatically call and AND these two conditions
together based on the filter.sym of the filters which are on.  Definitely
an improvement of what I left it as, but still requires explicit
specialized code for each filter.  Is there any way around at least that
much, though??

re:"dead" line in Query#validate_param -- Hmmm.  I don't think so.  It's a
bit ugly, and certainly won't pass rubocop and therefore should be
refactored, though.  This was a last minute thing to get the tests passing
so I could go to sleep. :(

Thanks for the comments.  This is important work.  (And very well done,
Joe, sorry I keep forgetting to say that!)  I think together we're going to
make this really robust. Well worth the effort.  Sorry it took me a long
night of living with the new code to understand it finally!

-Jason

mo-nathan:
> re: observations only -- See use case in other email.  Filtering out
> lichens should apply to names as well as observations.

OK, I can see that as well as real taxa (e.g., Boletaceae).

> re: ContentFilter shouldn't know about checkboxes -- Agreed.  I still don't
> have a concrete recommendation for an alternative.  Maybe record the data
> type in the ContentFilter instead?  Then account/prefs and
> observer/advanced_search can implement UI however they see fit?

Yes, it seems to me that we should be saying more about the "type" of the ContentFilter.  I think this would help here and with ObservationFilters even potentially with getting rid of the explicit reference to the Observation or Name classes, or maybe we need some more booleans.  For example, .filter_type could be one of (:boolean, :id, :list, :name, :chained_filter).  :boolean and :id would get simple checkboxes and :boolean would use "IS TRUE"/"IS FALSE" whereas :id would use "IS NULL"/"IS NOT NULL".  :list would be fore the lichen case where there is an explicit list somewhere of lichen species.  :name be for matching as a regular expression against a .name field.  :chained_filter would take another field and another filter e.g., (:where, (filter :name "California"))).

> re: non-checkbox filter -- Location is the only compelling use case so far.

Generalizing the lichen case to other taxa or lists also seems compelling.

> re: Query::ObservationFilters leveraging ContentFilter ...

Does the .filter_type idea above resolve this?

> re:"dead" line in Query#validate_param ...

My bad.

pellaea:
I like the "filter_type" field.  I think it's the right direction.
Actually, maybe we should be handling it instead as different subclasses of
ContentFilter.  That's probably the "right" way to go.

But I still don't like putting the SQL into ContentFilters, even if it's
half-hidden behind :id and :boolean type. I think the following case might
help make my point.  Right now we are still using the now-redundant
observations.specimen flag to implement the "has_specimen" filter.
Someday, me might like to see this go away and instead have it check if
there are attached specimen records.  That means the SQL condition will
change from "observations.specimen IS TRUE" to "JOIN specimens ON
specimens.observation_id = observations.id".  This sort of implementation
detail, in my opinion, belongs in Query, not the ContentFilter.

But I will support leveraging what we can from ContentFilter for the rest.
Right now with only two simple boolean filters it will, of course, increase
complexity.  But I'm glad to see interest in expanding the number and
complexity of filters.  It will probably be worth it eventually, if not
immediately.

I've been waiting for this feature for many years, very exciting!

-Jason

app/classes/content_filter.rb
JoeCohen: Pasting @pellaea comments here, to keep thread together:
> re: observations only -- See use case in other email. Filtering out
lichens should apply to names as well as observations.

and for locations. If I'm interested in Observations only if they're from North America, I probably don't want my RSS feed to include new Locations in Laos.

But am I guilty of premature generalization?

app/classes/content_filter.rb
JoeCohen: Yes.  
When I wrote all this code, I was thinking like a procedural programmer rather than an OOP program -- which I am not.  I love the ContentFilters as a class; couldn't figure out how to do that on my own.
Also note @pellaea email comment (pasted here for completeness):

> d) It is possible to have the User#has_specimen_checkbox and User#has_image_checkbox methods created dynamically based on ContentFilter using a construct like this:
```ruby
  ContentFilter.all.each do |fltr|
    next unless fltr.checkbox.present?
    define_method(fltr.checkbox) do
      content_filter[fltr.sym] == fltr.checked_val ? 1 : 0
    end
  end
```

pellaea:
Not premature, we're just struggling with understanding how it will work!

mo-nathan:
Coverage change is due to reduced total line count in the file due to the simplification of a coverage function.  Not going to worry about this coverage decrease.

PR #265, opened 2017-01-08
Rubocop ApplicationController

JoeCohen:
Reduces the number of Rubocop offenses from 184 to 9.

Some questions:
- For almost all the "Method has too many lines" offenses, the fix included extracting methods.  **Should I make the extracted methods Private?**
- I changed some method names to make Rubocop happy, and -- if the the method was called in another file -- aliased the old name. **Should I change the methods names in all files?** (I think that would be better, but might confuse someone looking for the old name.)
- There are two hard remaining issues which I did not address:
  - Class has too many lines. [1039/100].  Not sure what to about this.  Break off modules? Would that really make the controller slimmer, or would it merely make Rubocop happier?
  - `show_index_of_objects` is too large and complex. Refactoring it will require a lot of thought. I will circle back to it for a future PR.
- There are some spots where Rubocop complained about `html_safe`.  I didn't know if there was a reason for calling that method, so I left things alone.
- I shrunk`log_memory_usage` to 10 lines (at least not without using `eval`), so I left it at 15.

I will also add separate comments on a few lines of code.


test/controllers/account_controller_test.rb
```diff
@@ -421,7 +421,7 @@ def test_add_mugshot
   end
 
   def test_no_email_hooks
-    for type in [
+    [
```
JoeCohen: Makes Rubocop happier, but my real motive for changing this is that it make my text editor happier. (Folds behave better.)

app/controllers/application_controller.rb
```diff
@@ -1611,36 +1735,34 @@ def default_thumbnail_size
...
 
-  # Set the default thumbnail size, either for the current user if logged in,
-  # or for the current session.
-  def set_default_thumbnail_size(val)
-    if @user
-      if @user.thumbnail_size != val
-        @user.thumbnail_size = val
-        @user.save_without_our_callbacks
-      end
+  def default_thumbnail_size=(val)
```
JoeCohen: Does thumbnail size have any real effect? Is there a UI for changing it? 
If the answers are "no", maybe the code should be ripped out (or at least rip out anything that calls `default_thumbnail` or `default_thumbnail=`.

pellaea:
There is UI for changing it.  On the show_observation page, I believe.  I
think smaller thumbnails probably do make a significant difference on very
limited bandwidth connections.  It looks like 410 out of 8193 users have
requested the small thumbnails (5%).  We can still have the browser
downsize and serve the same size thumbnails to everyone, if it's just a
matter of screen real estate usage.  But this seems to be enough users to
make it worth keeping, at least to me.

mo-nathan:
> For almost all the "Method has too many lines" offenses, the fix included extracting methods. Should I make the extracted methods Private?

Personally I like private methods so I would be fine with that.  I generally hold non-private methods to a high standard in terms of defensiveness against bad input.  However, private methods should never get called directly in a test (but they should have full coverage).

> I changed some method names to make Rubocop happy, and -- if the the method was called in another file -- aliased the old name. Should I change the methods names in all files? (I think that would be better, but might confuse someone looking for the old name.)

I would rather the name was just changed everywhere.  In my view aliases lead to more confusion than global name changes.

> Class has too many lines. [1039/100]. Not sure what to about this. Break off modules? Would that really make the controller slimmer, or would it merely make Rubocop happier?

This suggests (and it's true) that ApplicationController has become a "God Object" (https://en.wikipedia.org/wiki/God_object) and does too much.  In my view it should get broken up into multiple classes not just separate modules.  However, that type of refactor is often difficult.

> show_index_of_objects is too large and complex. Refactoring it will require a lot of thought. I will circle back to it for a future PR.

Understandable.

> There are some spots where Rubocop complained about html_safe. I didn't know if there was a reason for calling that method, so I left things alone.

Interesting.  I'm not seeing that problem when I run Rubocop on the current version.

> I shrunk log_memory_usage to 10 lines (at least not without using eval), so I left it at 15.

In my view, the counters should get switched to a dictionary rather than using a case statement.  However, I also note that this method does not appear to be in use any more, so maybe it should just be deleted.

> I will also add separate comments on a few lines of code.

Wasn't able to find these extra comments.  Let me know if there's other stuff you want me to look at.

Great work!

mo-nathan:
This is just an error message, so there should not be a concern with this
line.

app/controllers/application_controller.rb
```diff
@@ -1307,7 +1411,8 @@ def show_index_of_objects(query, args = {})
                                          name: name.display_name)
         when :pattern_search
           :runtime_no_matches_pattern.t(type: type,
-                                        value: query.params[:pattern].to_s).html_safe
+                                        value: query.params[:pattern].
+                                                     to_s).html_safe
```
mo-nathan: Reasonable point.  However, it's just an error message so it would be worth removing the html_safe and seeing if any tests fail or any weird things happen when you run the code.  If not, then we should remove it.  If it does cause a failure of some sort, then we'll need to see if there's some way this could turn into an HTML injection issue which seems like the only real risk here.

app/controllers/application_controller.rb
```diff
@@ -1549,25 +1666,33 @@ def extra_gc
...
-        line.match(/^Private_Dirty/) ? (pd += val) :
-        line.match(/^Private_Clean/) ? (pc += val) : 1
+      next unless line =~ /\d+/
+      val = $&.to_i
+      case line.match(/^[\w]+/)[0]
+      when "Shared_Dirty"  then sd += val
+      when "Shared_Clean"  then sc += val
+      when "Private_Dirty" then pd += val
+      when "Private_Clean" then pc += val
+      else 1
```
mo-nathan: I'm pretty sure it does nothing.  As mentioned elsewhere, I think this would be better as a dictionary where the keys are "Shared_DIrty" etc., and the values are the counts rather than separate local variables.  It could even be a little counter object.  I would also get rid of $& and use a match object.  I'm surprised Rubocop doesn't complain about that.

JoeCohen:
I am ready for this to be merged.  What I did was:
- Get rid of the method name`alias`es which I previously added, using a single method name throughout the app.  (That's why so many files are now changed.)
- Privatize a bunch of methods.

I did not remove the two instances of `.html_safe`. One (at line 753) is tested. When I removed it, tests failed with "Unsafe HTML found!" The other (at line 1431) is untested, so I let it be.

I did not do anything with log_memory_usage. I didn't understand how a dictionary would work to fix the Rubocop issues. (Maybe I didn't under what you meant by dictionary.) And I didn't want to spend the time to deal with it (and to write a test for this method) given that it's unused and might be removed.  (I'll ping Jason and ask if he wants it left in.  If not, I'll remove it in a separate PR.)

pellaea:
Fixing conflicts now...

PR #267, opened 2017-01-11
Replaces fakeweb gem with webmock.

JoeCohen:
* Allows easier stubbing and setting expectations on HTTP requests. That will let me test some corners of the app which I was having trouble testing with just fakeweb and minitest.
* Disables external networks requests during testing.


pellaea:
Looks like this will merge easily.  Just running tests on general principle...

PR #270, opened 2017-01-17
Test BackwardsCompatibility and fix one bug

JoeCohen:
- Also add test of BackwardsCompatibility methods, and
- Fixes bug exposed by test.

The bug is that `/observer/old_method/nnn` redirects to `/obj/new_method/\u001` instead of `/obj/new_method/nnn`
The problem is the second argument to `sub` was `”\1”`. This was intended to be the  regexp capture group, but actually was the *string* `\l`.  To get the capture group, we need single quotes or an extra backslash.  See https://ruby-doc.org/core-2.2.0/String.html#method-i-sub.

There's still at least one issue with `BackwardsCompatibility#action_has_moved`
(assuming that we should not change the login requirements on any route):
None of the methods created by `action_has_moved` are excepted from the `before_action :login_required` callback, but many should be excepted.

Four ways to address this (and there are probably more):
- Add a skip_before_action to `backwards_compatibility` with a manual list of methods to except.
- Manually add those methods to the exceptions in observer_controller.rb (and maybe move all the `action_has_moved` calls to that file).
- More metaprogramming, e.g., add a parameter to `action_has_moved` which is a list of filters to be excepted, with `action_has_moved`to reflect on that list and create appropriate exceptions.

Another possibility (suggested by @nwilson-eol) is ditching BackwardsCompatibility and adding routes.  (And another is to keep BackwardsCompatibility but have it metaprogramatically add routes instead of methods. Ugh!)

### DIFF
--- a/app/controllers/observer_controller/backwards_compatibility.rb
+++ b/app/controllers/observer_controller/backwards_compatibility.rb
@@ -8,7 +8,7 @@ class ObserverController
       args = url.sub(/^[^?]*/, "")
     elsif url.match(/\/\d+$/)
       base = url.sub(/\/\d+$/, "")
-      args = url.sub(/.*(\/\d+)$/, "\1")
+      args = url.sub(/.*(\/\d+)$/, '\1')
     else
       base = url
       args = ""
--- a/test/integration/observer_controller_supplemental_test.rb
+++ b/test/integration/observer_controller_supplemental_test.rb
@@ -1,0 +1,43 @@
+require "test_helper"
+
+# Tests which supplement controller/observer_controller_test.rb
+class ObserverControllerSupplementalTest < IntegrationTestCase
+  # ----------------------------
+  #  BackwardsCompatibility
+  # ----------------------------
+  def test_backward_compatibility
+    obj = "name"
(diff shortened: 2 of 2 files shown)


pellaea:
First observation -- ObserverController#action_has_moved creates redirects,
and of course the require_login filter will still apply to the action it
redirects to, so that one thing is not a problem.

I personally find routes opaque to put it mildly.  I like the
easily-readable format of the current setup.  One line per action, obvious
at a glance what the old methods were called and what the new ones are.

But that said, I never actually have to look at this stuff.  So whatever
the rest of you decide is the right way to do this is just fine by me.

JoeCohen:
I think I wasn't very clear about the login problem. The before_action callback seems to be invoked as soon as a method is called -- it doesn't wait for the redirect.  So, for example, `show_user_observations` is demanding a login, even though `observations_by_user` is excepted from `:login_required`.
Try this:  
1.  While logged in, click http://mushroomobserver.org/observer/show_user_observations?id=4468.
2. Logout, click it again.

JoeCohen:
Obviously, this is going nowhere fast, as BackwardsCompatibility will be ripped out.

PR #271, opened 2017-01-17
Remove backwards_compatibility portion of partioned ObserverController.

JoeCohen:
See Slack discussion:
>[09:00]
jdcohenesq Some questions/comments about BackwardsCompatibility
1. *Purpose*. Is it to:
(a) not break links from external websites and/or not break other requests from external sources?
(b) keep things running smoothly while ObserverController is being refactored?
2. If the purpose includes not breaking links from external websites or other external requests, should some/all methods be excepted from `:login_required`? If that’s the only purpose, do we need an `action_has_moved` for actions which write to the MO db?
…
[15:11]
nathan Regarding backwards compatibility, we try to keep all old links valid, at least if they are getting used.  action_has_moved is a reasonable approach, but I wonder if we should really be using routes for this.  In any case we should probably add some sort of logging for the use of old routes.  We should not change the login requirements on any route.
[15:41]
jdcohenesq Thanks for the help and advice.
A.  I was also wondering about using routes. But the downside is it would really complicate our current, wonderfully simple routes.
B.  Do you think that urls which require login (e.g., add_comment, bulk_name_edit, destroy_species_list) are getting used?
C.  There’s a bug in BackwardsCompatibiity#rewrite_url.  I’ll shortly create a PR to address the bug.  But maybe the PR should not be merged until the next issue is addressed.
D.  If we want to keep login requirements constant, then there’s another bug:  None of the BackwardsCompatibiity#action_has_moved is excepted from before_action :login_required, but many should be. In the PR comments, I’ll propose ways to fix this.
[16:09]
nathan I don't see that moving stuff out of observer_controller/backwards_compatibility would be more complicated than using routes.rb for it's intended purpose.  routes.rb is certainly where current Rails developers expect stuff like that to happen.  I don't believe it was supported back when we first created the site, but it is the prevailing standard for this purpose.  Removing that stuff from observer_controller would be a step in demoting ObserverController from it's historic status as a God Object.
[16:09]
I'm sure add_comment is being used every day.
[16:10]
Unless you specifically mean the route through observer/add_comment and then, I'm not sure.
[16:11]
It's hard tell how much the old routes are being used.  We could well have dangling references to them.
[16:12]
Oh, the point about action_had_moved not being excepted is interesting.  Now I think I understand what you are driving at.
…
[16:27]
Sounds like this is actually dead code and should be deleted.
[16:28]
That's even simpler than moving it all to routes!
[16:31]
jdcohenesq I think you’re probably right about the dead code. But maybe we should wait a few days and see if people are complaining to the webmaster.  And I also hope Jason will weigh in — he may have had some other purpose in mind.
…
[16:56]
jasonphollinger One last comment on action_has_moved -- Yes, I agree, all the old routes are almost certainly dead now, after what? 7 or 8 year! :slightly_smiling_face:  But we will have extensive need for some sort of equivalent mechanism when we finally create all those new controllers segregated out of observer_controller.  If this is easy and clean to do with routes, and that's the sanctioned way to do it nowadays, then I'm all for it.
[16:56]
nathan w3c_tests is also a best practices thing.  If when you run the tests there is a directory called html, then copies of all the pages should get dumped there.  I think this page gives you a convenient index to view them so they can be send to the W3C validator.  Something we probably haven't done in a long time and would be a great thing to do sometime soon.
[16:57]
My understanding is routes.rb is the way to go.
[16:58]
jasonphollinger Good, then rip out observer_controller/backwards_compatibility.rb, whee!


JoeCohen:
Oh no, another order-dependent-ish failure! (Passed on my local VM, and when Travis ran it in my Github repo.  See https://travis-ci.org/JoeCohen/mushroom-observer/builds/192810323.)
I'll look into it.

JoeCohen:
Passes on my local VM, using the same seed (55101) as the failing Travis build.  
The test includes:
```ruby
    post(:add_specimen, params)
    assert_equal(specimen_count + 1, Specimen.count)
    # specimen = Specimen.find(:all, order: "created_at DESC")[0] # Rails 3
    specimen = Specimen.all.order("created_at DESC")[0]
    assert_equal(params[:specimen][:herbarium_name], specimen.herbarium.name)
```
and it's failing at the last line above.  Travis said:
```ruby
FAIL["test_add_specimen_post", SpecimenControllerTest, 1.348250276999977]
 test_add_specimen_post#SpecimenControllerTest (1.35s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"Strobilurus diminutivus det. Rolf Singer - NYBG 1234567"
        +"specimen_with_notes"
        test/controllers/specimen_controller_test.rb:94:in `test_add_specimen_post'
``` 
The message indicates that -- under some condition(s) -- Rails thinks that the last-created specimen is not the one created by the `post`, but rather the last one in the fixtures. I see that the last fixture lacks an explicit created_at.  But why should that matter? Rails auto-fills that:
>If your table/model specifies any of Active Record's standard timestamp columns (created_at, created_on, updated_at, updated_on), they will automatically be set to Time.now.

See http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html. The Specimen table does specify created_at (and this test has never before failed). Maybe some Ruby or Rails time zone issue? Some Travis timing problem?  Perhaps I should add an explicit created_at date to the last fixture?

First try again to reproduce the error.  Can I tell Travis which seed to use?

JoeCohen:
I tried restarting the job, but it ran with a different minitest seed (and passed).

JoeCohen:
The minitest seed can be set by viewing the Travis job, clicking "More Options", Settings, and setting the TESTOPTS  environvment variable to "--seed=55101".
I did that and the job passed.  So the problem is not order, not local vs Github, not Joe's Github repo versus the project repo. I don't know how else to try replicating the failure, so:
**I give up.**

PR #274, opened 2017-01-21
Add coverage to observer_controller/user_controller

JoeCohen:
Coverage had been at 40.21.  See [Coveralls report](https://coveralls.io/builds/9716352/source?filename=app%2Fcontrollers%2Fobserver_controller%2Fuser_controller.rb). This branch should cover pretty much all the red lines in the report.

Questions/Comments:
In general, I preferred adding integration tests rather than functional tests. (This is partly due to Rails changes which we will hit when we move to 5.0. See, e.g., http://blog.bigbinary.com/2016/04/19/changes-to-test-controllers-in-rails-5.html.)
QUESTION: Do you prefer functional tests?

When testing whether the correct page was being displayed, I matched against <title> in <head>. I thought it would be the most stable (least brittle) match. 
QUESTIONS: Do you think the response url is more stable? Can you suggest a more stable match?

### DIFF
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -2579,4 +2579,19 @@ class ObserverControllerTest < FunctionalTestCase
     actual = @controller.external_sites_user_can_add_links_to(obs)
     assert_equal(expect.map(&:name), actual.map(&:name))
   end
+
+  # -------------------------------------
+  #  observer_controller/user_controller
+  # -------------------------------------
+
+  # Prove that user_index is restricted to admins
+  def test_index_user
+    login("rolf")
+    get(:index_user)
+    assert_redirected_to(:root)
+
+    make_admin
+    get(:index_user)
+    assert_response(:success)
+  end
 end
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -2580,9 +2580,11 @@ class ObserverControllerTest < FunctionalTestCase
     assert_equal(expect.map(&:name), actual.map(&:name))
   end
 
-  # -------------------------------------
+  # ------------------------------------------------------------
+  #  User
   #  observer_controller/user_controller
-  # -------------------------------------
+  #  Also see test/integration/observer_user_controller_test.rb
+  # ------------------------------------------------------------
 
   # Prove that user_index is restricted to admins
   def test_index_user
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -2596,4 +2596,35 @@ class ObserverControllerTest < FunctionalTestCase
     get(:index_user)
     assert_response(:success)
   end
+
+  def test_change_user_bonuses
+    user = users(:mary)
+    old_contribution = mary.contribution
+    bonus = "7 lucky \n 13 unlucky"
+
+    # Prove that non-admin cannot change bonuses and attempt to do so
+    # redirects to target user's page
+    login("rolf")
+    get(:change_user_bonuses, id: user.id)
+    assert_redirected_to(action: :show_user, id: user.id)
+
+    # Prove that admin posting bonuses in wrong format causes a flash error,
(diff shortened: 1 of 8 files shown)


pellaea:
I was under the impression that integration tests were _much_ slower than
functional tests(?)  Otherwise I don''t care.

I think url and title are equally brittle (or not brittle).

JoeCohen:
Will see if I can do some very simple quick-and-dirty benchmarks later today. (I think I read somewhere that Capybara tests are speedy. But I can't find where I might have read that. So maybe I'm imagining things.)

pellaea:
That would be great, worth checking though.

JoeCohen:
In a very simple test, an integration test was ~2x slower than its functional counterpart.  
So I will re-write the integration tests as functionals, then re-open the PR.  (And maybe at that point Travis will be more cooperative.)
Thanks for prodding me on this!

Details:  
Functional test with 1 `get`, Integration test with 1 `visit` I looped 1000 times (`1000.times do`) around the tests.  Then separately ran each. 
Timings:
- Functional: 42.45760s
- Integration: 82.91319s
And adding a single user action (clicking on a link) to the integration test made it even slower:  178.98536s.

Even more details:
Functional test:
```ruby
  def test_benchmark_user_checklist
    1000.times { test_user_checklist }
  end

  def test_user_checklist
    user = users(:rolf)
    expect = Name.joins(observations: :user).
                  where("observations.user_id = #{user.id}
                         AND names.rank = #{Name.ranks[:Species]}").
                  uniq

    get(:checklist, id: user.id)
    assert_match(%r{Checklist for #{user.name}}, css_select("title").text, "Wrong page")

    prove_checklist_content(expect)
  end

  def prove_checklist_content(expect)
    missing_names = (
      expect.each_with_object([]) do |taxon, missing|
        next if /#{taxon.text_name}/ =~ css_select("body").text
        missing << taxon.text_name
      end
    )

    assert_select(".checklist a", count: expect.size)
    assert(missing_names.empty?, "Species List missing #{missing_names}")
  end
```
Integration test (using Capybara):
```ruby
  def test_benchmark_user_checklist
    1000.times { test_user_checklist }
  end

  def test_user_checklist
    user = users(:rolf)
    expect = Name.joins(observations: :user).
                  where("observations.user_id = #{user.id}
                         AND names.rank = #{Name.ranks[:Species]}").
                  uniq

    visit("/observer/checklist/#{user.id}")
    assert_match(%r{Checklist for #{user.name}}, page.title, "Wrong page")

    prove_checklist_content(expect)
  end

  def prove_checklist_content(expect)
    missing_names = (
      expect.each_with_object([]) do |taxon, missing|
        missing << taxon.text_name if page.has_no_content?(taxon.text_name)
      end
    )

    page.assert_selector(".checklist a", count: expect.size)
    assert(missing_names.empty?, "Species List missing #{missing_names}")
  end
```
Functional test run:
```
VM:mushroom-observer$ rake test test/controllers/observer_controller_test.rb test_benchmark_user_checklist
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's 'rails' settings.
Started with run options -n test_benchmark_user_checklist --seed 21271

  1/1: [====================================================================] 100% Time: 00:00:42, Time: 00:00:42

Finished in 42.45760s
1 tests, 4000 assertions, 0 failures, 0 errors, 0 skips
```
Integration test run:
```
VM:mushroom-observer$ rake test test/integration/observer_user_controller_test.rb test_benchmark_user_checklist
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's 'rails' settings.
Started with run options -n test_benchmark_user_checklist --seed 36206

  1/1: [====================================================================] 100% Time: 00:01:22, Time: 00:01:22

Finished in 82.91319s
1 tests, 3000 assertions, 0 failures, 0 errors, 0 skips
```

pellaea:
Yikes.  I wonder what capybara is quick compared to??  That's even worse
than I suspected. Thanks so much for testing this.

PR #276, opened 2017-01-25
Create ThemeController from part of ObserverController

JoeCohen:
Delivers [Pivotal #137757419](https://www.pivotaltracker.com/story/show/137757419). Also see recent Slack discussion.

**Comments/Questions**:

Checklist/Completeness. 
I used [this checklist](https://docs.google.com/document/d/1Tn4uwFHXLMpNv12oKMkKgqz8Fiw07qCvDSpwnGbaphc/edit?usp=sharing) (which augments @pellaea’s Slack post), except as noted below.  If there’s something that I should add to the checklist please let me know.  

Callbacks and exceptions
- I used the current scheme (list callbacks and excepts near top of controller), as we don’t yet have a replacement). 
- None of the ThemeController methods invoke `before_action :login_required`.  Therefore I could have omitted that callback.  Instead, I included it, excepting all methods. I did it this way to serve as a reminder in the unlikely circumstance that we one day add a method which does require login.

Actions
Doing this conversion made me really learn something which I’d read before.  One doesn’t need to define controller actions if there’s a template which matches that name.  
- There were no controller actions defined for the individual themes (Agaricus, etc.).  I left things that way, adding a comment explaining the absence of a definition.
QUESTION: would you prefer that I create actions for those themes.  The definitions will be empty, e.g.,
```
  def Agaricus
  end
```
- There was an empty action defined for `color_themes`.  I retained it.  I could delete it.  
QUESTION: Should I do so? (Then we’ll have a controller with no defined actions.)

Routes
I did not add routes to direct the old actions to the new controller.  This isn’t hard to do, but -- given that these actions have been invisible for a while -- seemed unnecessary clutter in the routes file.
QUESTION:  Should I add these routes? (If so, I also intend to add route tests.)

Other:
- I will update the observer/news item right after I’m done creating this PR. 
- I’m trying to make my commit log and messages more useful via the practices suggested [here](http://chris.beams.io/posts/git-commit/) and [here](http://who-t.blogspot.com/2009/12/on-commit-messages.html).  (I can still improve; I see at least one thing I messed up -- mixing code changes with whitespace changes.)
Suggestions are welcome.


pellaea:
Nathan, do you want to look at this one, too, before I merge it?  Looks good to me.

PR #279, opened 2017-01-28
Add coverage to RssLogController

JoeCohen:
Coverage was at 78.05.  See [Coveralls Report](https://coveralls.io/builds/9876869/source?filename=app%2Fcontrollers%2Fobserver_controller%2Frss_log_controller.rb). This PR should bring it to 100%.

QUESTION: Can you suggest meaningful assertions which are simpler and/or faster than the ones I have added? (When I run either of the new test methods on its own, minitest reports that it Finishes in ~5 seconds.)


JoeCohen:
Forget to cover two methods.  I'll look at this tomorrow to see if I can get them covered.

JoeCohen:
The untested methods are RssLogController#next_rss_log and #prev_rss_log.
I think they are not used.
Am I missing something? If they are really not used, can I remove them?

pellaea:
Hmmm.  Probably not.  There is no way to show a single rss log entry, is
there?  If you click on one you go to the target object, not the rss log.
... No!  Here's how you do it: Go to "Latest Changes", click on an
observation, go to the bottom of the page and click on "Show Log".  The
resulting page is a show_rss_log, and on top are "«prev | index | next»"
links.  Probably no one ever uses them, so if you prefer, you can just rip
those out and rip out prev and next_rss_log.  Otherwise, I'll bet
show_rss_log is not tested, either(?)

JoeCohen:
`next_rss_log` and `prev_rss_log` require login.  I don't think that's intentional, but wanted to double check.
Both actions do say `:no_robots:` but don't say `:nologin:` (and are not excepted from `:login_required`).

pellaea:
Yes, accident.  This is what I mean about it being very easy to forget to
add things to the login_required exception!

JoeCohen:
Thanks for confirming. Of course, it's also possible to forget `:nologin:', but - as you pointed out - less likely because it's right next to the method definition.

pellaea:
I'd like to see some stats on that. :)

JoeCohen:
Added test of `next_rss_log` and `prev_rss_log`, and fixed bug (that they required login) exposed by test.

pellaea:
Is this ready to go now?

JoeCohen:
Yes. It's ready.

PR #280, opened 2017-01-28
Cover ObserverController/ShowObservation

JoeCohen:
Covers the last few previously uncovered lines of this file.

QUESTIONS: `show_obs` simply redirects to `show_observation`. 

1. Do we redirect, or are requests to this action better handled by routing (like @nwilson-eol suggested for the now-defunct backwards_compatibility stuff)? 
2. If a redirect, should the html status code be 301 (Moved Permanently), rather than 302 (Found)? See https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html. 
3. If anyone actually does anything with 301, is that generally a way to handle backwards compatibility?

### DIFF
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -103,6 +103,12 @@ class ObserverControllerTest < FunctionalTestCase
     assert_equal(:thumbnail, user.thumbnail_size)
   end
 
+  def test_show_obs
+    obs = observations(:fungi_obs)
+    get(:show_obs, id: obs.id)
+    assert_redirected_to(action: :show_observation, id: obs.id)
+  end
+
   def test_page_loads
     get_with_dump(:index)
     assert_template(:list_rss_logs, partial: :_rss_log)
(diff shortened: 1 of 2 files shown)


pellaea:
It seems to me all backwards compatibility should be handled with "permanently moved" redirects.  (I thought that was how "action_has_moved" did work.  I didn't realize redirect_to didn't serve a "permanently moved" status.  Although, it does make sense now that I think of it.)  Is there a way to make the routing behave like that?  I agree in principle that the router does feel like the right place to handle all these sorts of things.  But if it can't...  Hmmm.  Maybe it just doesn't matter if browsers are left thinking incorrectly that the old actions still work.

pellaea:
PS. This PR looks fine.

JoeCohen:
Good question about routing and redirects. (Wish I'd thought of that.)  And the answer is yes, we can make routes behave the way you suggest. (Moreover, in the routing system 301 is the default status for a redirect.  See http://guides.rubyonrails.org/routing.html#redirection.

Is that what we want to do with `show_obs`?

pellaea:
Almost certainly.  I can't remember the original purpose of show_obs.  But
I'm pretty sure it's deprecated now.  I'm glad you understand routes!

JoeCohen:
Thanks again.
I suspect that this PR should be merged as is so it doesn't hang around.
I'll add a ticket to replace the method with a route to redirect to show_observation.

pellaea:
Great, will do.

PR #282, opened 2017-02-01
Cover missed lines in MarkupController

JoeCohen:
ObserverController/MarkupController#advanced search had 3 lines which were missed by the test suite. They involve: (a) rescuing an error, and (b) two cases where a name lookup finds no matches but does find spelling suggestions.

This commit adds a test which forces an error by stubbing a method invoked by lookup_general 
(using mocha gem for stubbing.) The commit also tests the edge cases.


pellaea:
Looks good to me, okay if I merge it?

PR #284, opened 2017-02-03
Remove list_notifications

JoeCohen:
- This method was unused.
- It also had the only line in this file which was not covered by the test suite. Therefore its removal finishes https://www.pivotaltracker.com/story/show/137757299 (Improve NotificationController coverage).


JoeCohen:
@pellaea - I think you can merge this despite the supposed 0.01% decrease in coverage.

PR #290, opened 2017-03-10
Improve group names

JoeCohen:
## Procedure for dealing with this PR
This PR just improves “group”. (it does not deal with “clade”.)
I suggest this procedure:
1. Code review.
2. I make any desired fixes. 
3. Merge into master but do not roll out.  
4. I do another PR which  
- adds “clade” (should be very simple: add 1 or 2 tests, adjust GROUP_ABBR, tweak `parse_group`;
- expands instructions in the create_name template (tell people how to define a group or clade, where to put “sensu X”, where to put status designation (e.g., “nom. prov.”), gives some more examples.
- updates release news.
5. Once that PR passes muster, merge and roll out.
(Alternatively - if you prefer -I could add the items in #4 to this PR. Then just a single code review, merge, and rollout.)
6. Once rolled out, I will manually adjust the funny group Names in the db which you cited in https://github.com/MushroomObserver/mushroom-observer/pull/289

### Changes to transformation methods
I couldn’t complete https://www.pivotaltracker.com/story/show/141125771 just by tweaking `ParsedName#parse_group` and the Name parsing constants; I had to adjust the methods which transform the various types of names.  In general, I changed these transformation methods to: 
1. remove “group” from a method input;
2. transform a string;
3. restore group to the end of the string.
Perhaps there’s a more elegant way of doing this, at least for some of the transformation methods.

### Where to define transformation methods
These transformations methods are Name class methods.  This feels somewhat awkward; at their core they’re just String transformations. If they were String instance methods (defined in StringExtensions) we could send them to the string being transformed.  E.g., instead of
  Name.display_to_real_search(self)
we would do
  _string_.display_to_real_search
On the other hand, these transformations are used only in the context of names; so it would also be odd to make them general StringExtensions and it’s handy to have them in the Name file.
** What do you think?  And is there another alternative? ** (It doesn’t look like we can make them Name instance methods, as they are sometimes used before instantiation of a Name.)

### Miscellaneous
In this PR, I didn't try to deal with users' putting "group" into Author.  I'd rather they put "group" where it belongs.  I will add more instructions and examples to the create_name template as part of the next PR.

I don’t try to handle “Xxx sp. group”; I don’t think this form should be allowed.

### DIFF
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -442,10 +442,10 @@ class Name < AbstractModel
       gsub(/__\*?\*? [^_\*]* \*?\*?__/x, " "). # (this part should be unnecessary)
       # Because "group" was removed by the 1st gsub above,
       # tack it back on (if it was part of display_name)
-      concat(group_designation(name))
+      concat(group_suffix(name))
   end
 
-  def self.group_designation(name)
+  def self.group_suffix(name)
     / group\b/.match(name.display_name).to_s
   end
 
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -57,7 +57,7 @@
 #  SUBSPECIES_PAT::     (Xxx yyy ssp. zzz) (Author)
 #  VARIETY_PAT::        (Xxx yyy ... var. zzz) (Author)
 #  FORM_PAT::           (Xxx yyy ... f. zzz) (Author)
-#  GROUP_PAT::          (Xxx yyy ...) group
+#  GROUP_PAT::          (Xxx) | (Xxx yyy ...) group
 #  AUTHOR_PAT:          (any of the above) (Author)
 #
 #  * Results are grouped according to the parentheses shown above.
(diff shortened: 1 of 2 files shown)


app/models/name.rb
```diff
@@ -439,7 +439,14 @@ def real_search_name
   def self.display_to_real_text(name)
```
JoeCohen: 1.Move to StringExtensions?
2. Is there an elegant, transparent way to do remove and restore group_abbr using only regexp?

JoeCohen:
Re decrease in coverage. Just ignore it. The alleged decrease in coverage is due to an old, previously merged PR.  See https://coveralls.io/builds/10534859/source?filename=app%2Fclasses%2Fquery%2Finitializers%2Fcontent_filters.rb

JoeCohen:
Hold on!  Just added some more tests and one failed.  I will re-open when I get to green.

pellaea:
Thanks for your work on this, Joe.  Sounds like you're on the right track.
I won't be able to look at the PR today, but promise to check it out
tomorrow.

JoeCohen:
That's perfect timing!  I expect to have it fixed by this evening.
Tomorrow, I'll be off-line from about sunrise to mid-afternoon.

PR #291, opened 2017-03-11
New improved group names

JoeCohen:
## Procedure for dealing with this PR
This PR replaces #290, and just improves “group”.  (it does not deal with “clade”.)  
I suggest this procedure:
1. Code review.
2. I make any desired fixes. 
3. Merge into master but do not roll out.  
4. I do another PR which  
- adds “clade” (should be very simple: add 1 or 2 tests, adjust GROUP_ABBR, tweak `parse_group`;
- expands instructions in the create_name template (tell people how to define a group or clade, where to put “sensu X”, where to put status designation (e.g., “nom. prov.”), gives some more examples.
- updates release news.
5. Once that PR passes muster, merge and roll out.
(Alternatively - if you prefer -I could add the items in #4 to this PR. Then just a single code review, merge, and rollout.)
6. Once rolled out, I will manually adjust the funny group Names in the db which you cited in https://github.com/MushroomObserver/mushroom-observer/pull/289

### Miscellaneous
In this PR, I didn't try to deal with users' putting "group" into Author.  I'd rather they put "group" where it belongs.  I will add more instructions and examples to the create_name template as part of the next PR.

The purported coverage decrease should not stand in the way of merging this PR.  (The decrease was caused by a prior PR and merger, and has nothing to do with this PR.  See https://coveralls.io/builds/10544157/source?filename=app%2Fclasses%2Fquery%2Finitializers%2Fcontent_filters.rb.)

### DIFF
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -1598,8 +1598,15 @@ class Name < AbstractModel
                           )? "?
                         )
                       )
-                      \s #{GROUP_ABBR}
-                      (\s (#{AUTHOR_START}.*))?
+                      (
+                        ( # group, optionally followed by author
+                          \s #{GROUP_ABBR} (\s (#{AUTHOR_START}.*))?
+                        )
+                        | # or
+                        ( # author followed by group
+                          ( \s (#{AUTHOR_START}.*)) \s #{GROUP_ABBR}
+                        )
+                      )
                     $/x
 
   class ParsedName
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -1692,9 +1692,9 @@ class Name < AbstractModel
     if result.author.present?
       # Add "group" before author
       author = Regexp.escape(result.author)
-      result.search_name.sub!( /(.*)(#{author})/, '\1group \2')
-      result.sort_name.sub!(   /(.*)(#{author})/, '\1 group  \2')
-      result.display_name.sub!(/(.*)(#{author})/, '\1group \2')
+      result.search_name.sub!( /(#{author})$/, 'group \1')
+      result.sort_name.sub!(   /(#{author})$/, ' group  \1')
+      result.display_name.sub!(/(#{author})$/, 'group \1')
     else
       # Append "group" at end
       result.search_name += " group"
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -57,7 +57,7 @@
 #  SUBSPECIES_PAT::     (Xxx yyy ssp. zzz) (Author)
 #  VARIETY_PAT::        (Xxx yyy ... var. zzz) (Author)
 #  FORM_PAT::           (Xxx yyy ... f. zzz) (Author)
-#  GROUP_PAT::          (Xxx yyy ...) group
+#  GROUP_PAT::          (Xxx) | (Xxx yyy ...) group
 #  AUTHOR_PAT:          (any of the above) (Author)
 #
 #  * Results are grouped according to the parentheses shown above.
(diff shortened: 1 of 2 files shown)


app/models/name.rb
```diff
@@ -439,7 +439,14 @@ def real_search_name
   def self.display_to_real_text(name)
     name.display_name.gsub(/ ^\*?\*?__ | __\*?\*?[^_\*]*$ /x, "").
       gsub(/__\*?\*? [^_\*]* \s (#{ANY_NAME_ABBR}) \s \*?\*?__/x, ' \1 ').
-      gsub(/__\*?\*? [^_\*]* \*?\*?__/x, " ") # (this part should be unnecessary)
+      gsub(/__\*?\*? [^_\*]* \*?\*?__/x, " "). # (this part should be unnecessary)
+      # Because "group" was removed by the 1st gsub above,
+      # tack it back on (if it was part of display_name)
+      concat(group_suffix(name))
```
JoeCohen: This line is necessary because the first `gsub` removes everything in the name after the taxon, including “group” (and author). For “group” names, `display_name` now includes “group” and so should `real_text_name`.
There's probably a way to revise the `gsub` so it does not remove "group".  But I couldn't figure it out, so it's probably convoluted.

pellaea:
Wow you're up late!

JoeCohen:
Too late.  But I wanted to get this done before morning.  Now off to bed.

app/models/name.rb
pellaea: Don't worry, display_to_real_text is already an ugly kludge to fix something I couldn't figure out how to fix "correctly" years ago.  Some day in the distant future when we have nothing better to do -- ha! -- we can revise the whole process of parsing and multiply-formatting names.

app/models/name.rb
```diff
@@ -1639,22 +1679,33 @@ def self.parse_author(str)
...
-        sort_name: format_sort_name(text_name, "group"),
-        display_name: format_name(name, deprecated) + " group",
-        parent_name: parent_name,
-        rank: :Group,
-        author: ""
-      )
+    return unless match = GROUP_PAT.match(str)
+
+    # Parse the string without "group"
+    ungrouped_str = str.sub(/\s#{GROUP_ABBR}/, "")
```
pellaea: Should this have a slash-b at the end of the regex to prevent premature match of, say, a species epithet which begins with "group" (or later "clade")?

app/models/name.rb
```diff
@@ -1639,22 +1679,33 @@ def self.parse_author(str)
...
+    # Adjust the parsed name
+    result.text_name += " group"
+
+    if result.author.present?
+      # Add "group" before author
+      author = Regexp.escape(result.author)
+      result.search_name.sub!( /(.*)(#{author})/, '\1group \2')
+      result.sort_name.sub!(   /(.*)(#{author})/, '\1 group  \2')
+      result.display_name.sub!(/(.*)(#{author})/, '\1group \2')
+    else
```
pellaea: This approach worries me a bit.  It's hard to predict how a strange author might match some part of the genus or species in very rare cases.  Can we assume that the author has to come at the end?  I think that's safe.  Then we can calculate exactly where to insert "group" in each of search_name, sort_name and display_name.  Or just anchor the regexen to the end of the string, that would also feel safer to me.

app/models/name.rb
JoeCohen: Great catch.  Just pushed the fix, including a test.

app/models/name.rb
JoeCohen: Another great catch. I think we should make that assumption.  (And I had already planned to tweak the instructions in the create_name template under Scientific Name to tell users where to put "group" or "clade"):
>Unformatted taxon name <i>without</i> author info or status info. E.g.: <br>&nbsp;&nbsp; Amanita muscaria var. flavivolvata <br>For a <b>group or clade</b>, select <b>[:RANK]</b> “[:RANK_GROUP]” above, and add either “group” or “clade” at the end of <b>[:form_names_text_name]</b>. E.g.: <br>&nbsp;&nbsp; Amanita muscaria group <br>&nbsp;&nbsp; Morchella elata clade.

(I can only think of one case where Author doesn't legitimately come at the end: autonyms like _Amanita_ Pers. sect. _Amanita_. But "group" will blow up that case in any event.
And, of course, there will always be a few Names where users ignore the instructions and try to do things like Amanita crocea sensu auct. amer. group.)

JoeCohen: I will modify and push up.

pellaea:
Oh right, what happens if the user places the word "group" in the wrong
place?  Only two possibilities make sense: the correct placement before
author, and at the end.  It might not be hard to modify GROUP_PAT to accept
either.  Something like this?

(
  \s #{GROUP_ABBR} (\s #{AUTHOR_START}.*)? |
  (\s #{AUTHOR_START}.*) \s #{GROUP_ABBR}$
)

(Note that we'd have to change how the author is extracted from the match
slightly.  Namely, we'd have to remove GROUP_ABBR from the beginning and
end of the author paren group in the suggested regex above.)

I can just see this being a common source of confusion and annoyance,
because "group" after author will still parse successfully, just
incorrectly.  If we can have MO handle both cases, it might save some
heartburn.

JoeCohen:
I'll have a look at fixing things per your comment on the possibility of user typing "group" at the very end (after author).

app/models/name.rb
JoeCohen: As I was getting ready to replace
`result.search_name.sub!( /(.*)(#{author})/, '\1group \2')`, etc., 
with
`result.search_name.sub!(/(#{author})$/, 'group \1')`, etc.,
I finally realized -- given our assumption that author is at the end -- that the two `sub`s have the same functionality. The `*` is greedy, so `(.*)` gobbles up everything until the **last** occurrence of `author`.
Your suggestion is simpler (only 1 capture group), shorter, and clearer.
The other way is marginally more flexible -- it would still work if we ever decided to allow something (I have no clue what) to appear after author. Maybe that's just premature generalization.
**What do you think?**

pellaea:
That's my whole point, yours allows something to come after... oh! I see,
you *know* at least one occurrence of the author is at the end, therefore
the greedy .* will ensure that it matches the one at the end, not some echo
of it in the middle.  You're absolutely right, these are equivalent.  My
bad.  Might want to add one simple test case to make sure it parses it
correctly, though, say: "Foo group Foo".  Just to be safe!  I personally
like to add "^" and "$" anchors to regexen whenever I can, just makes
surprises less likely.

pellaea:
I'm content with this.  I'll let it simmer overnight, just to be sure neither of us thinks of anything (although sounds like you will be to busy to be thinking about this PR!).  Will merge it tomorrow.  Thanks again, Joe.

PR #293, opened 2017-03-16
Fix bug in calculation of user's favorite naming

JoeCohen:
Delivers https://www.pivotaltracker.com/story/show/124224561

This bug happened when:
- A user changed his vote for one naming to a lower value, and
- That change resulted in another naming(s) have the highest positive value for that user.
In those circumstances MO would not recognize the other naming(s) as the user's favorite.
So if the user owned the Observation, then the Big Eyes would disappear, and Observer Preference would be calculated incorrectly.

The PR
- re-enables ObserverTest#test_vote_favorite, fixing bugs in the test itself
- fixes the problem in Observer#change_name which problem is exposed by the re-enabled test
  - change_vote now checks to see if other vote(s) should become the user's favorite(s).
- re-factors ObserverTest#test_vote_favorite.

Manual Test Script
- Create Observation with 3 Proposed Names:
  - A with confidence "I'd Call It That"
  - B, with confidence "Promising"
  - C, with confidence "Could Be"
- Edit A to make its confidence "Could Be"
Result: The "big eyes" should appear next to B. (In master, they disappear. )

Other
It might be convenient to review this PR by stepping through the commits.


test/models/observation_test.rb
```diff
@@ -683,4 +703,8 @@ def test_imageless
     # not enough notes
     assert_false(observations(:agaricus_campestrus_obs).has_backup_data?)
   end
+
+  # Prove that dump_votes debugging routine correctly dumps votes
+  def test_dump_votes
```
pellaea: Did you forget to add some code here?

pellaea:
Looks good, just that one test stub which I think you'd intended to fill out but lost track of.

test/models/observation_test.rb
JoeCohen: Sorry.  I did not mean to commit those lines.  
I will delete them.

JoeCohen:
Just pushed the deletion.

PR #294, opened 2017-03-19
Add support for clade names

JoeCohen:
- Delivers https://www.pivotaltracker.com/story/show/141072769
- Clarifies and expand instructions for Create Name and Edit Name.

There are also 2 tasks which must await roll-out of this PR (assuming that it gets merged):  
- Update the non-English translations of `RANK_GROUP: Group or Clade`, etc.
- Fix the poorly formatted group names in the DB which @pellaea found as part of the review of PR #289.

I will fix these manually after roll-out.


app/models/name.rb
```diff
@@ -1688,25 +1693,26 @@ def self.parse_author(str)
...
-    # Parse the string without "group"
-    ungrouped_str = str.sub(/\s#{GROUP_ABBR}\b/, "")
-    result = parse_name(ungrouped_str)
+    result = parse_name(str_without_group(str))
     return nil unless result
 
     # Adjust the parsed name
-    result.text_name += " group"
+    group_chunk = GROUP_CHUNK.match(str)
+    group_type = group_type(group_chunk[:group_abbr])
```
pellaea: I think you might have forgotten to change these two lines to
```
group_type = Name.group_type(str)
```

pellaea:
Looks good, just that one tiny accidental omission.

app/models/name.rb
JoeCohen: I will re-write and ping you when I'm done.
(Those two lines are what I meant, but I had reservations about them and `GROUP_CHUNK`. I'm going to re-write to try to make things more transparent.)

PR #295, opened 2017-03-19
Fix bulk editor invalid date bug

JoeCohen:
Fixes bug in SpeciesList Bulk Editor, delivering [Pivotal #141748815](https://www.pivotaltracker.com/story/show/141748815)

Not much code here, but the SpeciesList Bulk Editor is a bit hair, so this could use a good review.

- The Bulk Editor would throw an error when user tried to change an Observation date to a non-date (e.g., "blah") or something in YYYY-MM-DD format which is a non-existent date (e.g., 2017-02-31).
- The bug was caused by lines like
```
  if @when_str && !Date.parse(@when_str)
    errors.add(:when_str, :runtime_date_should_be_yyyymmdd.t)
```
This doesn't work because Date.parse is not a validator, see https://ruby-doc.org/stdlib-2.3.1/libdoc/date/rdoc/Date.html#method-c-_parse.
If Date.parse cannot parse its argument then you get an Error: at least a TypeError if x can’t be converted to a String (e.g. x == nil), and ArgumentError if x is an invalid string (e.g., x == "banana" or x == "2017-02-31").
- Incidentally improves coverage of Observation. Trying to improve that coverage is what made me fix the bug.  I found the bug while trying to cover:
screen shot 2017-03-19 at 16 45 40
My solution was to Rescue the error, detect its type, and give the appropriate flash message.

I did not provide a news release update. Users will likely not notice the change because -as far as I know- I'm the only one who report the bug, .

### Manual Test
- Go to any Species List you own.
- select Bulk Editor
- fill in the Date field of the 1st Observation with  a non-existent date (e.g., 2017-02-31)
- select Update
Result: Flash message including "Invalid date", and the nonsense date remains in the date field of the 1st Observation.  (In master this would throw an error).
- fill in the date field of the 1st Observation with  a  non-date (e.g., "nonsense")
- select Update
Result: Flash message including "Date should be in year-month-day format.", and the non-date remains in the date field of the 1st Observation.  (In master this would throw an error).

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -230,7 +230,10 @@ class Observation < AbstractModel
 
   def when_str=(x)
     @when_str = x
-    self.when = x if Date.parse(x) rescue ArgumentError
+    begin
+      self.when = x if Date.parse(x)
+    rescue ArgumentError
+    end
     x
   end
 
@@ -1139,15 +1142,14 @@ class Observation < AbstractModel
       errors.add(:alt, :runtime_altitude_error.t)
     end
 
-    if @when_str
-      begin
-        Date.parse(@when_str)
-      rescue ArgumentError
-        if @when_str =~ /^\d{4}-\d{1,2}-\d{1,2}$/
-          errors.add(:when_str, :runtime_date_invalid.t)
-        else
-          errors.add(:when_str, :runtime_date_should_be_yyyymmdd.t)
-        end
+    return unless @when_str
+    begin
+      Date.parse(@when_str)
+    rescue ArgumentError
+      if @when_str =~ /^\d{4}-\d{1,2}-\d{1,2}$/
+        errors.add(:when_str, :runtime_date_invalid.t)
+      else
+        errors.add(:when_str, :runtime_date_should_be_yyyymmdd.t)
       end
     end
   end
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -1139,8 +1139,16 @@ class Observation < AbstractModel
       errors.add(:alt, :runtime_altitude_error.t)
     end
 
-    if @when_str && !Date.parse(@when_str)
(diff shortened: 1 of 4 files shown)


app/models/observation.rb
```diff
@@ -230,8 +230,8 @@ def when_str
 
   def when_str=(x)
     @when_str = x
-    self.when = x if Date.parse(x)
-    x
+    self.when = x if Date.parse(x) rescue ArgumentError
+    self.when
```
pellaea: Are you sure there's no subtle reason why we used to return the input string here, and not the parsed result?

pellaea:
Only that one question.  If you know that having
```
obs.when_str = "blah"
```
evaluating to "blah" is a bug, then I'm fine with changing it.  Or if you know that having assignments return the parsed value is the standard way these assignments work in Rails, then I'm good with it.  If the question is fuzzy, then I think it should revert.  Two reasons:

1. Don't fix what ain't broke.
2. Having the assignment evaluate to something other than the value assigned seems highly non-intuitive to me personally.  I don't think of "=" typically being a closet operator(!)

That said, I do know it is common in Ruby for 
```
obj.attr = val
puts "gotcha!" if obj.attr != val
```
to be true.  But
```
var = obj.attr = val
puts "really??" if var != val
```
would really catch me off guard.  But that's Ruby for you!  Despite the language's motto about not doing anything surprising, I find it surprises me regularly! :)

JoeCohen:
Thanks for noticing this and making me really focus on it.  I think you're right:
- Returning the former value of `when` is an unexpected result
- Returning the rvalue of `when=` works (doesn't break anything) in all cases. 
- There probably is a subtle reason to return the rvalue.  (I think it influences whether the bulk editor stays on the same page.)
(Of course, it's a little hard to tell because when the input string was unparsable, we did not **actually** "return the input string"; instead things just blew up.

PR #296, opened 2017-03-23
Interactive lat lng

scottymeyers:
(#9630845) Users should be able to interactively add lat/long for observations -- Adds a map to the observation form which geocodes the Where field. Users may manipulate the marker, latitude & longitude fields.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,6 +12,7 @@
 //= require theater
 //= require validate_file_input_fields
 //= require main
+//= require observation_form_map
 //= require lightbox
 //= require popup_exif
 
--- a/app/assets/javascripts/observation_form_map.js
+++ b/app/assets/javascripts/observation_form_map.js
@@ -1,0 +1,96 @@
+// Observation Form Map - Lat/Long/Alt Helper
+
+// ./observer/create_observation
+if ( $('#observationFormMap').length ) {
+
+    (function() {
+        var searchInput = $('#observation_place_name'),
+            latInput = $('#observation_lat'),
+            lngInput = $('#observation_long');
+
+        // init map object
+        var map = new google.maps.Map(document.getElementById('observationFormMap'), {
+            zoom: 1,
+            center: { lat: 8.38321792370903, lng: -66.56186083652352 }
+        });
+
+        // init marker
+        var marker = new google.maps.Marker({
+            title: 'Location',
+            map: map,
+            draggable: true,
+            visible: false
+        });
+
+        // disable user from submitting on this element
+        $(searchInput).keypress(function(e) {
+            if (e.keyCode == 13) { e.preventDefault(); }
+        });
+
+        // dragged marker event
+        google.maps.event.addListener(marker, 'drag', function(e) {
+            latLngInputs.update(e.latLng.lat(), e.latLng.lng());
+        });
+
+        // clear the text inputs
+        $('.map-clear').click(function(e) {
+            e.preventDefault();
+
+            latLngInputs.empty();
+        });
+
+        // updates map on text input change
+        $([latInput, lngInput]).each(function(index, value) {
+            $(this).keyup(function() {
+                var latLng = new google.maps.LatLng($(latInput).val(), $(lngInput).val());
+
+                marker.setPosition(latLng);
+            });
+        });
+
+        // find on map
+        $('.map-locate').click(function(e) {
+            var address =  $(searchInput).val(),
+                geocoder = new google.maps.Geocoder();
+
+            // we need a value, even a single letter will suffice
+            if ( $(searchInput).val().length <= 0 )
+                return false;
+
+            latLngInputs.empty();
+
+            geocoder.geocode({'address': address }, function(results, status) {
+                if (status === google.maps.GeocoderStatus.OK && results.length > 0) {
+                    var location = results[0].geometry.location;
+
+                    marker.setPosition(location);
+                    marker.setVisible(true);
+
+                    if (results[0].geometry.viewport) {
+                        map.fitBounds(results[0].geometry.viewport);
+                    } else {
+                        map.setZoom(1);
+                    }
+
+                    latLngInputs.update(location.lat(), location.lng());
+                } else {
+                    console.log('Invalid address');
+                }
+            });
+        });
+
+        var latLngInputs = {
+            update: function(lat,lng) {
+                $(latInput).val(lat);
+                $(lngInput).val(lng);
+            },
+            empty: function() {
+                $(latInput).val('');
+                $(lngInput).val('');
+
+                // hide the marker
+                marker.setVisible(false);
+            }
+        };
+    })();
+}
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -44,4 +44,5 @@ jQuery(document).ready(function () {
         // If file field immediately followed by span, show selection there.
         if (next.is('span')) next.html(val);
     });
+
 });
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/application.js
```diff
@@ -12,6 +12,7 @@
 //= require theater
 //= require validate_file_input_fields
 //= require main
+//= require observation_form_map
```
pellaea: Would it be better to have app/views/observer/_form_observations.html.erb load this instead?  See how Ray had app/views/observer/_form_multi_image_upload.html.erb load the interactive image uploader.  What do you think?  I guess if it's really short it may be better to have it in the default JS slug that every page loads, than to have create/edit_observation have to load several additional files -- a few extra bytes in one download is quicker than a bunch of tiny downloads.  Not sure what's best.

app/assets/javascripts/observation_form_map.js
```diff
@@ -0,0 +1,96 @@
...
+
+    (function() {
+        var searchInput = $('#observation_place_name'),
+            latInput = $('#observation_lat'),
+            lngInput = $('#observation_long');
+
+        // init map object
+        var map = new google.maps.Map(document.getElementById('observationFormMap'), {
+            zoom: 1,
+            center: { lat: 8.38321792370903, lng: -66.56186083652352 }
```
pellaea: Haha!  Just curious, how did you settle on the middle of Venezuela being the optimal default centerpoint??

pellaea:
This looks great.  I still want to run it locally to see what it looks like, too.  Nice and simple, I like it.

app/assets/javascripts/application.js
scottymeyers: I'm completely open to that! I've still got quite a bit to learn about Rails (and MO's) conventions. I can quickly adjust to include directly from the template.

What I generally do is compile all my JS from component/template specific files into one global JS file (concatenated, minified, cached etc.). Then I apply a specific class to the body (controller/action or page name) that I need to add functionality to. That way I don't have to dig through HTML templates to find the relevant JS when unfamiliar!

app/assets/javascripts/observation_form_map.js
scottymeyers: I'm debating where to take my next trip! Honestly, it probably doesn't need to be so specific... just felt it setup for a nice vantage point.

pellaea:
Hmmm.  My browser is behaving strangely.  A few observations and suggestions:

First, it took me a while to realize that I couldn't do anything with the map until I entered a text location.  How can we make this clearer?  Can we maybe hide the map until the location is entered?

I was debating whether we could simplify out of existence the "Locate on Map" and "Clear Map" buttons.  While I think "Locate on Map" is unnecessary at first, I can imagine the user changing their mind about the text name for the location, and I wouldn't want the map to wipe out their gps choice and re-geolocate every time the user edits the text field; however there are times when I think the user would like to ask MO to re-geolocate.  So I think I like the "Locate on Map" button.

But what is your envisioned use-case for the "Clear Map" button?  I can't think of one off the top of my head.

Another note: if the user has recently created an observation, it will prefill the "where" field with the value from the last location.  And lat/long as well.  Will your code pick that up on page load and initialize the google map correctly?  It looks like it only does geolocation in event of "where" changing, and only repositions the pin in event of lat/long changing.

Note, that if the geolocation service fails to return a lat/lng, there is no way to use the map!

So I definitely think we should let the user initiate a pin on the map with a click.  Would that be weird?  Would that interfere with dragging the map around?  Maybe double-click would be better?  Along these lines, I'm also imagining a case where geolocation is totally wrong, so the user zooms out, pans halfway around the world to the correct location (this has happened to me!), zooms in to the right location... only to realize that they've left the pin far behind in India, with no way to move it to the presently shown location.  I would like to see a single peck or click-and-hold or double-click move the pin to a given spot, regardless where the pin is presently, or if there is a pin at all.

Lastly, my browser is doing strange things with the drag event.  Half the time it thinks I'm trying to drag the map into the image uploader(!)  And that steals focus from google maps and leaves the map is odd states.  Like one time the pin started drifting slowly north while I watched with me not doing a thing, haha!  And frequently google doesn't get the "stop drag" event, and the map continues to scroll around tracking the mouse even though I'm not holding the button down anymore.

Maybe the best solution is to change the multi-image form dropzone to a more restricted spot, say, around the file browse input, for example?  I think if all of the drag/drop event handlers for the image uploader are totally separate from (i.e., not on a parent of) the google map, things will work more smoothly.  But whatever works is fine with me.

Last comment -- elevation service... I think there is a great way to integrate it here after all!  Once you've picked a pin location, I think it should be safe to request the elevation of that exact point.  My objection was providing a default basd only on the vague extents of the geolocate service.  If the user explicitly plops the pin down somewhere (maybe require that they be at a minimum zoom level, too?) then I think it's safe to accept the elevation google gives us.

Along those lines, I just realize another concern.  A lowe-power-user typical use case might be: enter "San Francisco", pick "San Francisco, California, USA" from the autocomplete menu, google automatically locates SF on the map and plops a precise pin down smack in the middle of the city, wherever that is.  The user ignores that completely, drags a photo in, and submits the observation.  The resulting observation will appear to have a precise lat/long, but in fact it's totally arbitrary.  I think this is dangerous.  I think we should consider not placing a pin on the map for the user.  Force them to do it explicitly.

Instead, how about this alternate work flow:  User enters SF, picks SF from autocomplete, MO zooms in automatically because there was no pre-existing gelocation info, map now shows SF nice and conveniently but there is no pin on the map yet and no lat/long has been chosen.  If the user just drags their photo in and submits now, the observation is left (correctly) vague without a falsely accurate lat/long/elev.  If the user instead clicks/control-clicks/click-and-holds/double-clicks (whatever we choose!) on the map, a pin is dropped at a precise location, a lat/long/elev is filled in for them, and now we can trust that a human being has actually signed off on that lat/long/elev.

What do you think?

pellaea:
That actually makes a lot of sense.  Should we summarily decide right now
to combine all the little js's into one big universal js?  The only thing I
think we need to be careful of is not including google maps api, for
example, from pages that don't actually have a map!

I think it will be both quicker and, as Scott says, easier to understand if
all of our js from app/assets/javascripts be included on every page.

(Just to be clear, I do think that we should continue to keep them split up
into lots of small js files.  That is definitely in line with rubocop and
most other modern "clean code" practices.  Shouldn't be too hard to grep
for function names in all the js files in that directory.)

scottymeyers:
Thanks for all the concerns and suggestions.

**Can we maybe hide the map until the location is entered?**
That’s a good idea. Although further down in your comments you suggest using the ‘Where’ field solely to zoom into a particular region of the map w/o placing a marker. If that was the case (which I think should be), I feel it’d be okay to leave the map visible initially.

**Buttons?**
Agreed, ‘Clear Map’ has no reason to be included, I'll go ahead and remove that!

**If the user has recently created an observation, it will pre-fill the "where" field with the value from the last location.**
No, I will hook that in.

**So I definitely think we should let the user initiate a pin on the map with a click. Would that be weird?**
I agree adding a double click seems to be a good solution.

**Lastly, my browser is doing strange things with the drag event.**
I haven’t experienced that, what browser are you using? and are you using a trackpad by chance?

**Last comment -- elevation service...**
Great. This wasn’t very difficult to implement, I’ll go ahead and include it.

**Map now shows SF nice and conveniently but there is no pin on the map yet**
Beautiful, I love that idea. I was definitely concerned about situations with generic long/lat and wasn’t too sure how to handle it aside providing some additional instructions to its helper block.

pellaea:
Yeah, sorry, that was a rambling comment!  I think I'd talked myself out of
hiding the map by the end. :)  In any case, it was all intended to be just
discussion, not firm requests.  If you like any of the suggestions, please
make some changes, and let's see how it "feels" then.  I was never very
good at designing UIs, and this is why.

PR #297, opened 2017-03-25
Interactive lat lng

scottymeyers:
(#9630845) - Adds map to the observation form.

- Users can use the 'Locate' Button which takes the value of the 'Where' field and focuses into that region of the map. I've removed the 'Clear Map' Button. You now have to manually add a marker, by clicking on the map. We discussed using the double click event, but that's already default for zoom in functionality.

- If a user has recently created an observation then reloads this form, and the 'Latitude' & 'Longitude' fields are both populated, a marker will be shown at the same location. If those fields are empty, but the 'Where' field is populated, we use the same function as the 'Locate' Button, which simply zooms us into that region of the map.

- Integrated the elevation service.


pellaea:
Thanks for working on this.  I'll give it a test drive later this afternoon.  Looking forward to it.  This is cool stuff. (So much nicer than endless tweaking of how names are parsed and cleaning up corrupt database records, haha!)

scottymeyers:
My pleasure, let me know if you have any suggestions!

JoeCohen:
Scott:
1. I sent you an invite so you can see the details of codeclimate's complaints. I have no clue whether they're justified, so your choice whether to ignore them.
2. How can I checkout your interactive-lat-lon branch? (Is it in a publicly-readable on-line repo?)
-- Joe

scottymeyers:
Thanks @JoeCohen, I'll take a look @ Code Climate.

I believe you can checkout the PR locally by creating a new branch and pulling the PR in like so:

> git checkout -b interactive-lat-lng && git pull origin pull/297/head

JoeCohen:
Thanks for the instructions.  They worked fine.  
The functionality is neat!

pellaea:
This works perfectly for me now.  Very intuitive and easy to use.

You're gonna hate me, but...  I realize now that there is a case for the old "Clear Map" button.  Once you've placed a pin, there's now no way to opt out of attaching a GPS to your observation.  There really needs to be some way to change your mind and clear the lat/long/elev fields (and pin on the map, of course, too).

So sorry!  But this is otherwise perfect, at least to my tastes.  Anyone else had time to test it out yet?

JoeCohen:
I looked at it briefly, and I liked it.
I agree there should be a way to change your mind: either clearing the map or getting rid of it. 
Some options: a Clear Map button; a Close Map button, re-clicking the Locate button; when the map is displayed, the Locate button changes to Clear Map or Close Map.
How about putting the Locate button next to the Map, rather than being separated by "Is Collection Location" its accompanying text (as it appears on my screen). 
screen shot 2017-03-25 at 18 30 27

pellaea:
I think the "Locate" (I liked "Locate on Map" a bit better if it fits,
since it's more self-explanatory) button needs to remain, just in case the
user changes the text and wants to re-locate the spot on the map.

I think Scott's original "Clear Map" button really was ideal.  (Sorry I
said anything!)

Joe has a point about where the button(s) go.  "Clear Map" would make more
sense next to the map and gps coordinates.  But "Locate (on Map)" makes
more sense where it is.  If both must be together, I vote for keeping them
next to the "where" text field.

On my screen, I think it was more comfortable to have the "Locate" button
below the where field rather than to the right of it.  Just because it
otherwise doesn't give enough space for the text field.  And there's dead
space (at least on my screen) below the text field anyway.

Just mild preferences, whatever you guys like is fine with me, too.

scottymeyers:
Closing for now - I'm going to include the 'Clear Map' button again and mess around with the placement.

PR #300, opened 2017-03-26
Fix google query string

raysuelzer:
Fix the google search so that it actually shows results.  

Resolves:  https://www.pivotaltracker.com/story/show/137410985


raysuelzer:
Unfortunately, running the tests using a windows host produces tons of errors, so I missed this one.  Will update test.

PR #301, opened 2017-03-27
Interactive lat lng

scottymeyers:
I've included the 'Clear Map' button again, placing it directly under the map this time around. Also, changed the 'Locate' button to say 'Locate on Map' and place directly underneath the 'Where' field.

Screenshot


JoeCohen:
Question about initial focus.  
I pulled your branch and started it. When I did create_observation, the initial focus was in the What field. On my screen this means Date, Collection Location, and Where. are hidden. 

screen shot 2017-03-31 at 17 03 44

I think this encourages the User to 
(a) do things in the wrong order, pinning the lat/lon before entering Where, 
(b) maybe pinning a lat/lon which has nothing to do with Where.

Could the initial focus be changed to Where?  
(We've had a similar conversation before.  See https://www.pivotaltracker.com/story/show/99865100.  But the presence of the interactive map makes things worse because it ends up hiding the top fields if the focus is on What.

pellaea:
It is supposed to focus first on where unless it pre-fills where, in which
case it should focus on what.

JoeCohen:
That helps with the problem of lat/lon being outside of Where.  
But it still hides those 3 fields, which makes it easy to unintentionally re-use them.

pellaea:
Fair enough.  How about always leaving where, lat, long, elev blank, and
asking the user if they want to re-use the last location, or maybe even the
last few locations.  How about a short list of checkboxes or a pulldown,
such as:

Recent locations: [Apply this location data] <-- button
 [Blah, Tulare co., California, USA - 39.1234°N 118.5678°W 1234m] <--
pulldown

or

Recent locations:
1. Blah, Alameda Co., California, USA [Use this]
2. Blah, Tulare co., California, USA - 39.1234°N 118.5678°W 1234m [Use this]
3. etc.

Clearly this belongs in a different PR and pivotal story.  But the
conversation seems to be happening here regardless!

JoeCohen:
i suggest doing the simplest thing which will make this branch work for users, then doing the other stuff in another story.  
The added functionality works as intended; Scott has spent a fair bit of time on this; but the mission keeps expanding.
What's the simplest thing we can do to make this mergeable?

pellaea:
I'm personally just waiting for a way to clear the lat/long/elev.
Otherwise I think it's great.  The rest of this conversation really has
nothing to do with the interactive google map, per se.

JoeCohen:
There's a Clear Map button.  Will that work?

pellaea:
Oh!  Sorry, did that get added while I was away?  I'll check out the latest
version tonight.

PR #303, opened 2017-04-01
JSON format for API

pellaea:
Allows client to request JSON format (request.env['CONTENT-TYPE'] = "application/json").

I tweaked the XML version to make all data not requiring joins available at low detail level.

XML is still the default for backwards compatibility.

Not all the API views are tested.  In some cases we'll need to add Query classes to handle them, e.g., Naming, Vote.  In at least one other case (ExternalLink) there is something broken which I can't make sense of.

I'm making a PR now so that it doesn't continue to rot in my personal repo for another several months.  Either we can (safely) accept it now despite the probably inevitable complaints about code style and coverage decrease, or we can leave this open for someone else to take over and polish off.  Either works for me.

I suspect there are clients out there who will love to switch from clunky XML to slick JSON.  So the sooner the better.


app/classes/api/results.rb
```diff
@@ -46,12 +45,12 @@ def page_length
       high_detail_page_length
     elsif method == "PUT"
       put_page_length
-    elsif method == "PUT"
+    elsif method == "DELTE"
```
JoeCohen: Should this be `DELETE`?

pellaea:
Yes, obviously still untested.  No PUT or DELETE services are tested yet. :(

JoeCohen:
Sorry I can't contribute more; I am woefully ignorant about this side of things.

pellaea:
The summary is the API class is a wrapper for Query, and the ApiController
just renders the results of the query.  Try these for example:

http://mushroomobserver.org/api/observations?locations=antarctica
http://mushroomobserver.org/api/observations?locations=antarctica&detail=low
http://mushroomobserver.org/api/observations?locations=antarctica&detail=high

Mostly this PR just messes with the rendering side.  And that's mostly just
a matter of deciding which info to include in the XML / JSON output.

The details of how views and partials and layouts and formats work is very
obscure to me, and I've probably gone about it slightly or completely
wrong.  But this seems to work at least...

raysuelzer:
Jason, do you want me to continue working on this?

pellaea:
By all means, if you'd like to take this over, go right ahead.  Or if you
want to use this fancy swagger blocks thing and start over from scratch,
you're welcome to do that, too.

I thought I already merged a relatively recent commit from master into this
branch.  I don't think it will be a huge problem merging it into master.
Besides it shouldn't conflict with anything we've done in months, anyway.

I'm not sure how much more there is to do, anyway.  Isn't it essentially
done?  Of course, if you want to change things, then go ahead.  My only
requirement is that we be careful to structure it so that joins are done
sparingly and correctly (i.e. with eager-includes where joins are
necessary, e.g. names and locations in observations).  My implementation
attempted to conform as closely as possible to the existing XML version.
Just seemed natural at the time.  But I suppose that's not a strict
requirement.  It's sort of a blank slate.  If there is a compelling reason
to change it subtly or radically, and you want to do the work, I'll almost
certainly support you.

-Jason

pellaea:
Oh yeah, look at that, there are no conflicts with master.

raysuelzer:
I’m putting a pull request that fixes the issue where the headers are not being respected for application/json.

The correct header for requesting JSON should be Accept.  Content-Type indicates the type of content that is being sent.

request.format = "json" if request.headers["ACCEPT"] == "application/json"

pellaea:
Thanks for catching that!

pellaea:
Hey Ray, where does this stand?  Are you working on it in a private branch somewhere?

raysuelzer:
I felt like I put in a pull request to you branch, but I can only find the one which fixes the total number of pages being calculated improperly.  I will dig around more when I get back to my house.  But, a good start would be to check the Pull Requests on your branch and pull in my page count fix if you want.

pellaea:
Ah!  PR to *my* repo?  Never thought to check for that...  I see it, okay,
will update the PR to the MO repo now...

pellaea:
Okay, I've pulled in Ray's fixes, and fixed the code climate issues.  This should be fully backwards compatible for all existing users of the API.  I think it should be safe to push out.  Any last thoughts?

JoeCohen:
RuboCop hates you.  Do you want me to deal with the 5 new issues (+ anything else which it kicks out when I fix them)?

pellaea:
That's so weird, I thought it had passed all the tests.  I'll do it, probably easy.

JoeCohen:
Yes, probably easy.  That's why I volunteered.  😉 
My guess is that when one touches something (file? method?) Rubocop complains about supposedly  "new" issues which aren't truly new.

pellaea:
No I mean I waited for the little green check mark to arrive before claiming I'd fixed all the problems.  I was very surprised I hadn't introduced any problems.  Anyway, it claims to be passing codeclimate again.  For now.

pellaea:
Curious: how do I run by hand myself  the equivalent of what codeclimate is doing?  If I run "rubocop test/controllers/api_controller_test.rb" I get a ton of complaints in areas I've not touched, and which codeclimate is somehow ignoring.

JoeCohen:
I'm going to reply in Slack.

app/controllers/api_controller.rb
```diff
@@ -17,6 +17,10 @@ class ApiController < ApplicationController
 
   disable_filters
 
+  # Don't know what wrapped parameters are supposed to be for, but in practice
```
JoeCohen: Wrapped parameters are described here:  http://api.rubyonrails.org/v5.1.1/classes/ActionController/ParamsWrapper.html

PR #304, opened 2017-04-04
adds filter to PivotalTracker API to specify stories with unscheduled…

scottymeyers:
The Pivotal Tracker API will only return 500 stories. Currently, we grab them regardless of their state, so we are grabbing 'Approved' stories which will be filtered out locally.

I've adjusted the API call to specifically grab the last 500 stories with an 'Unscheduled', 'Started' or 
'Unstarted' state.

### DIFF
--- a/app/classes/pivotal/http.rb
+++ b/app/classes/pivotal/http.rb
@@ -8,7 +8,7 @@ class Pivotal
 
   class << self
     def get_stories(_verbose = false)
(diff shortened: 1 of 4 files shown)


pellaea:
Awesome!  Thanks for looking into this, Scott.  I really really want to get this feature tracker back up and working.  Looks like this will be a big help.  I glanced through it and it looks good.  Are you ready, or should I wait a day before merging and pushing out?

PR #305, opened 2017-04-05
[136653151] - Feature Tracker page is missing stories

scottymeyers:
This applies fixes to the initial codeclimate issues.

-----

The Pivotal Tracker API will only return 500 stories. Currently, we grab them regardless of their state, so we are grabbing 'Approved' stories which will be filtered out locally.

I've adjusted the API call to specifically grab the last 500 stories with an 'Unscheduled', 'Started' or
'Unstarted' state.

### DIFF
--- a/app/classes/pivotal/http.rb
+++ b/app/classes/pivotal/http.rb
@@ -8,7 +8,7 @@ class Pivotal
 
   class << self
     def get_stories(_verbose = false)
-      json = get_request("stories?limit=500&filter=state:unscheduled,started,unstarted&#{story_fields}")
+      json = get_request("stories?limit=500&#{story_filters}&#{story_fields}")
       stories = []
       JSON.parse(json).each do |obj|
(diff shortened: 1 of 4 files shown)


app/classes/pivotal/http.rb
```diff
@@ -8,17 +9,25 @@ class Pivotal
 
   class << self
     def get_stories(_verbose = false)
-      json = get_request("stories?limit=10000&#{story_fields}")
       stories = []
+
+      request_stories("", stories)
+      request_stories("&offset=501", stories) if stories.count == 500
```
pellaea: Hey, so what's the approx. number of stories currently?  Do you happen to know off the top of your head?  Now that you're removing accepted stoires, etc. maybe it's not so prohibitively many?

scottymeyers: It was around 498

PR #306, opened 2017-04-19
Change authored/authorless Name pairs

JoeCohen:
User facing changes:
- Allow author to be deleted via edit Name without forcing a merger.  Example in Manual Script below.
- Allow create Name to create an authored :Group-level Name even if an unauthored one exists (but not vice versa).
My thought was that the creation of an imprecise :Group Name should not prevent creation of a precise one (but not vice versa).
This differs from non-:Group Names because; for non-:Group Names the user can easily find the author in an official nomenclature db (e.g., MycoBank or Index Fungorum), and then edit the extant Name to add the author.  This procedure doesn't work with Groups because they're not in those db's.)
- Display correct flash message when trying to create an authored non-:Group name when an unauthored one exists.

These changes involve re-writing NameController#create_name and NameController#edit_name. That in turn involves changes to Name.  And the changes to Name and NameController require correcting old tests and adding new ones, which require new fixtures.  There was also incidental refactoring.

Delivers https://www.pivotaltracker.com/n/projects/224629/stories/142363931
Follows up on PRs #291 and #294.

**manual test script**
1. (Create an authored group although an unauthored one exists)
- Create Name _Aa_ group
- Create Name _Aa_ group sensu Bb
result: Successfully created name ‘_Aa_ group sensu Bb’. 
(in master, result was "Name already exists: ‘_Aa_ group’")
2. (Edit Name, removing author without forced merger)
- Create Name, e.g.,  _Aa aa_ Author
- Create Name, e.g., _Aa bb_ Author.
- Edit _Aa bb_ Author, changing the Scientific Name to _Aa aa_, and removing Author.
result: Successfully updated name ‘_Aa aa_’.  (giving us an authored/unauthored pair.)
(In master the result was "Successfully merged name _Aa bb_ Author into _Aa aa_ Author.")
3. (Display correct error message when trying to create authored/unauthored non-Group name)
- Create Name _Aa bb_
- Create Name _Aa bb_ Author
result: "Name already exists: ‘_Aa bb_’ "
(in master result was "Name already exists: ‘_Aa bb_ Author‘)


JoeCohen:
In case anyone is thinking about merging this, please hold off a little.  I will deal with some of the codeclimate and ccveralls issues.

JoeCohen:
Fixed coverage, still need to deal with CodeClimate.

JoeCohen:
Dealt with CodeClimate issues (mostly by saying won't fix ABC offenses in tests).
So ready for review.

app/models/name.rb
```diff
@@ -2287,6 +2222,129 @@ def change_deprecated(deprecated)
 
   ################################################################################
   #
+  #  :section: Merging
```
pellaea: Are there any changes in this section?  Did you just move it all verbatim?  I can do a diff by hand if there are changes, no worries, just want to know if I need to take the time.

JoeCohen: No changes.  I just moved it near the other stuff which is used in editing.

pellaea:
Cool.  I like name.rb.  Have to run out for an errand.  Will check
name_controller.rb next.  All the rest  of the changes look like tests and
translations(?)

JoeCohen:
Exactly (the remainder are on the test side: tests, fixtures, etc.).  
The big substantive changes are in NameController.

app/models/name.rb
```diff
@@ -2219,12 +2102,64 @@ def self.new_name(params)
...
+  # Returns a Name instance, *UNSAVED*!!
+  def self.new_name_from_parsed_name(parsed_name)
+    new_name(parsed_name.params)
+  end
+
+  # Return extant Names matching a desired new Name
+  # Used by NameController#create_name
+  def self.names_matching_desired_new_name(parsed_name)
+    # unauthored ParsedName matches Names with or w/o authors
+    if parsed_name.author.empty?
```
pellaea: I haven't actually tested this, but it looks like it might be impossible to create "Blah group" if "Blah group sensu Author" already exists.  Maybe just moving the "if parsed_name.rank == :Group" to the top of the conditional would work?  I think the behavior is correct for other ranks.  That is, if "Genus Author" already exists, then it shouldn't let you create "Genus" without author.  I realize that you're allowed to modify an existing name to not have an author.  And that's fine by me.  But it smells funny that you have to create "Blah group Bogus Author" first, then edit the name to remove "Bogus Author". ...  Or you've already foreseen this case and I'm just not following you right!  In that case just ignore me.

JoeCohen: You are correct about being unable to create "_Blah_ group" if "_Blah_ group Author" already exists.  The question is whether that's the right behavior.

I did it that way because I thought on the create side, it should behave like non-Group Names for the same reason:  When editing a Name to remove an author we know the user has made a deliberate choice with respect to author, but when creating a Name it's more probable that the user simply overlooked the author.

I'm not sure that's the right thing to do, partly for the reason you pointed out.  What do you think? Should we ask other people to weigh in?

JoeCohen: And you are also right about (a) being easy to change and (b) how to change NameController. (I don't even have to change any tests -- I didn't test creating an unauthored group when an authored one exists. If we're going to change it, I should add a test.)

pellaea:
People very rarely use create_name, so it may be a non-issue.

So, if someone enters "Blah group" when creating an observation,
absolutely, have it auto-fill the author.  This seems still, after all
these years, to be obviously the right thing to do.

But if someone uses create_name... blah... yeah, I wonder if the correct
behavior is to allow it -- assuming a deliberate choice because they
actually know how to reach create_name (I keep forgetting!) and are
probably doing so for a reason -- but to warn them that another version of
that group exists with an author.

Lastly, if someone uses edit_name... all good, we're totally on the same
page.

Sure, by all means bring others in.  Shouldn't be nearly so hard to ask
"what should MO do if..." as it is to ask "is this code right?"  Haha!

JoeCohen:
1. I'll post something in MO Developers.  
2. You get to create_name from any Name page or from any Name Search results page by following the "Add a Name" link.

JoeCohen:
I posted a [question in MO-Developers](https://groups.google.com/forum/#!topic/mo-developers/cpSkT-CIN9c), but no responses.  So I made @pellaea's suggested change: 
When creating a :Group Name, a desired new unauthored Name is matched only by exact matches (and not by an authored counterpart).  
This will make things a little easier for the users (if any) who create :Group Names via name_create.

pellaea:
Yikes, I'm not prepared to dive back into this, since I've completely forgotten the changes.  I'm willing to assume that the changes you made after my earlier review are minor.  As I remember, it really only affects people using create/edit_name.  It shouldn't affect how names are created in, say, the obs form.  Again, I'm willing to field any further complications we've somehow missed after it's gone into production.

Does this sound good to you, too, Joe?  Should I go ahead and merge this?

JoeCohen:
1. Yes, changes after your last review are minor.
2. Yes, it doesn't affect how names are created in the Observation.  
3. But the big issue is that I refactored create_name, and you didn't have a chance to review that (a big job, and not easy). So it's a balance between:

- Merge without further review. The downside is that if I messed up something in create_name or edit_name, it could have big consequences. On the other hand it's close to 100% covered by tests, so likely very small probability of downside.
- Wait for complete review. The downside is 
  (a) Possibility of conflicts if another dev works on any of these files.  For instance @mo-nathan is working on a [name-controller-coverage branch](https://github.com/MushroomObserver/mushroom-observer/tree/name-controller-coverage), and there are (already) a few conflicts.  
  (b) Blocking other stories. (Currently blocking 2 stories.)

If -- after considering the above -- you're still willing to help field unexpected complications, then I think we should merge it.  But if you're at all uncomfortable with that, then let's wait for a complete code review.

pellaea:
Actually, sorry, I did finally review all of it.  Sorry I didn't say.  I
remember liking the rewrite, thought it was a great improvement over the
mess that was there before.

I'm fine with merging it.  Now's the time to do it, because I'm around off
and on all day every day.  I can help with anything that comes up, within a
few hours generally at most.

JoeCohen:
I will merge later this evening (unless you want to do it now).

Naive question:  How do the merged changes get into production?

pellaea:
Aha, I'm not entirely superfluous yet.  Someone has to log into the
production server and pull the latest changes from origin/master.  Then be
very careful to run the correct set of updates:

1. "bundle install" if Gemfile.lock has changed
2. "rake lang:update" if en.txt has changed
3. "rake assets:precompile" if any assets have changed or been added
4. "rake db:migrate" if there are any migrations
5. "uni reload" or "uni restart" to update the running rails processes
(unicorn)

And I'm undoubtedly missing something somewhere.

pellaea:
Good point, there's no harm in running them all every time.  Just takes
time.  It's probably best to make as much of the update process atomic as
possible.  But it's probably pretty hard to corrupt anything because
different processes are running different versions.  I think it will
generally just cause a few intermittent errors.  The affected users just
need to reload the page and it will be "magically" all better.

PR #307, opened 2017-04-20
adds a warning to feature tracker page

scottymeyers:
Displays a warning if more than 900 stories are displayed and that there is a limit of 1,000.

### DIFF
--- a/app/views/pivotal/index.html.erb
+++ b/app/views/pivotal/index.html.erb
@@ -48,7 +48,9 @@
 
 <div class="row">
   <div class="col-xs-12 max-width-text">
+    <% if @stories.count > 900 %>
       <b>Displaying <%= @stories.count %> of 1,000 available Feature Requests. Please <%= link_to 'let the admins know', '/observer/ask_webmaster_question', title: "Contact Us" %> that we're close to the limit.</b>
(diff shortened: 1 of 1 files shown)


pellaea:
Being picky here but...  Would you mind rewording it a bit?  Maybe adding: "Please let admins know that we're close to the limit."  Perhaps with a link to /observer/ask_webmaster_question.

scottymeyers:
For sure!

app/views/pivotal/index.html.erb
```diff
@@ -48,9 +48,7 @@
 
 <div class="row">
   <div class="col-xs-12 max-width-text">
-    <% if @stories.count > 900 %>
-        <b>Displaying <%= @stories.count %> of 1,000 available Feature Requests.</b>
-    <% end %>
+      <b>Displaying <%= @stories.count %> of 1,000 available Feature Requests. Please <%= link_to 'let the admins know', '/observer/ask_webmaster_question', title: "Contact Us" %> that we're close to the limit.</b>
```
pellaea: Wait... don't we want to keep the if statement?  I just want to someone to warn me when we get close to the limit, in case I don't look at the page myself regularly.

PR #308, opened 2017-05-01
mentions 'Locations' in Advanced Search Form helper text

scottymeyers:
(#129803613) It now says "Result Type: Do you want observations, images or names/descriptions?" But Locations is also a valid result type.

It's been included like so:
"Result Type: Do you want observations, images, locations or names/descriptions?"


JoeCohen:
Thanks for catching this.  And it's simple enough so that I'm confident merging it on my own.  (One less task for @pellaea.)

scottymeyers:
Great, thanks Joe!

PR #309, opened 2017-05-04
Page title messed up for any index sorted in reverse

scottymeyers:
(#144750485) Would like some thoughts on this approach, perhaps there is a better way?!


JoeCohen:
Scott:  I'll try to look at this this weekend. (I've been out all day and will be out most of tomorrow. I usually forget to set my Slack status to Away.)
- Joe

PR #311, opened 2017-05-06
Create name layout should be full width

raysuelzer:
Resolves pivotal issue: "create_name template should use full width, just like edit_name"

https://www.pivotaltracker.com/story/show/143807293


JoeCohen:
Probably more important:
Could the styling of _form_name be changed so that the help text (e.g. `form_names_text_name_help`) takes advantage of the full available width of the screen?

raysuelzer:
Will make suggested changes and look at the _form_name styling.  The only thing about full width text is that it is very hard to read.

raysuelzer:
I have a feel the help text was intentionally set to avoid taking up the full screen.  Which on my monitor would mean everything was a single line, which is hard for the eye to track.  There is a specific setting in the CSS that defines this.  I don't think we want to override it. 

image

JoeCohen:
Fair point about full-width text. 
I was thinking about avoiding/minimizing scrolling and making better use of the screen real-estate.  
What do you think about moving the help text to the right.  Or even (much harder) flowing everything into columns, the number of which would depend on screen width.

PR #312, opened 2017-05-06
Move news to a model

JoeCohen:
Delivers [Pivotal #142020237](https://www.pivotaltracker.com/story/show/142020237)

- Creates Article model whose instances are news articles.  
- Articles are added to the RssLog.
- Article write permission is administered via the "News Articles" Project.  
- Changed the "Release" link in the LH navbar:
  - renamed it "News"
  - moved it to the "Other" section, which I renamed "Indexes" and which I alphabetized.

**To manually test this branch  in dev mode you must create a Project with Title: "News Articles"**
This is a magic Project used to control Article write permission. I manually created this Project in the production db (and it there's fixture in the test db).  (It seemed simpler to do it this way versus a data migration.)

When (and if) the branch is put into production, I will create Articles:
(a) for the Releases prior to this PR (it will probably be one big Article, like "Releases 2016-12-18 to 2017-04-01"
(b) describing the changes in this PR.  
I will have limited web access ~ May 16-21, so it probably shouldn't get rolled out during that time.


pellaea:
This all looks good.

One general question: Why did you choose to go with a magic project instead of just falling back on admin mode?

I haven't tested it on the dev VM.  I trust you.  Besides, it's noncritical, so if bugs show up in production, no big deal, we;ll just fix it then.

Good exercise, creating a new model, controller, queries, etc. etc.  You clearly have a good grasp of the entire structure at this point.  Good work!

JoeCohen:
Comments below:
> One general question: Why did you choose to go with a magic project instead of just falling back on admin mode?

Good question.  I went back and forth on this.  I decided on a project so that we could let every developer (and maybe some others, e.g., Alan) write brief Articles without requring that they all be full Admins. 
> I haven't tested it on the dev VM. I trust you. 

Trusting me is a probably wrong. :smiley: Might be okay for this PR because it's non-critical and walled off from non-devs.  But don't get carried away. :smiley: 
> Besides, it's noncritical, so if bugs show up in production, no big deal, we;ll just fix it then.

Agree.  **Should I merge it?**

> Good exercise, creating a new model, controller, queries, etc. etc. You clearly have a good grasp of the entire structure at this point. Good work!

:blush: I'm learning, but slowly.  Still have a long way to go.  And Queries still confuse me. And there many parts of App and the stack where I haven't a clue.

pellaea:
Ha!  Then we're probably on about the same page now.  I'm forgetting faster
than you're learning.  Sure, please go ahead and merge this.  Whee!  I'm
now entirely superfluous!

PR #313, opened 2017-05-08
removes reverse_ from influencing page title

scottymeyers:
(#144750485) - Page title messed up for any index sorted in reverse

Title arguments [:order] was using the value for params[:by] which, for example, could be sort_by_* or sort_by_reverse_* and then pulling the translation.

None of those values existed for sort_by_reverse_* so I've removed it from title argument, since they're displaying the same copy.


JoeCohen:
Thanks for the fix, and for all the hidden time you put into this.
Looks good.  And I think it's simple enough for me to merge, so I will do so.

PR #314, opened 2017-05-11
Relax RuboCop

JoeCohen:
- Relaxes various RuboCop metrics application-wide: uses CodeClimate's suggested max's instead of RuboCop's
- Ignores various RuboCop metrics in test files: disables various Cops which we regularly ignore in test files and which make less sense there.


JoeCohen:
Closing this because I accidentally included some commits which belong in a separate branch.

PR #318, opened 2017-06-05
Make CodeClimate report results

JoeCohen:
- Eliminate stray leading space.
This space made CodeClimate (engines version) say  "No results to report" and also made OS version add files for  bootstrap.js and
lightbox.js.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -12,7 +12,7 @@
 # For configuration of specific engines, see CodeClimate documentation
 #   for that engine.
 
- engines:
+engines:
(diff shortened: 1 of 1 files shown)


JoeCohen:
I will merge because I want to fix the CodeClimate reporting  problem ASAP.
(The "3" "new" issues are which duplicated (triplicated) in 3 places in   app/classes/observation_report.rb.  The line tells how one row should be mapped for exporting. I don't want to take the time right now to figure out how to fix this.)

JoeCohen:
Delivers Pivotal #146618407.

PR #319, opened 2017-06-07
Refactor observation reports.

pellaea:
Add mycoflora format.
Move species_list report here.
Get it all to pass rubocop (as much as possible).
Add unit tests.


JoeCohen:
IMO: Just fix the easy offenses, mark the others Invalid or Wontfix.

pellaea:
Oooh!  There is a way to mark them "invalid" or "wontfix"???  Yes!  We need
to in this case, I think.  I tried valiently, but could not find any sane
way to reduce a method which is comprised of a single array of labels(!)
Even when I moved every value in the array into a method, it *still*
complains about too many lines and ABC overcomplexity.  But I don't see any
way to avoid creating a big array.  That's what the CSV file is -- a bunch
of rows of long arrays of values.  How does one break that up??

JoeCohen:
Yes, you can do that.  
What do you see when you click on Details on the codeclimate line in the report above?
screen shot 2017-06-08 at 14 22 29

JoeCohen:
If you see lines like this:
screen shot 2017-06-08 at 14 24 59
then hover over New.

pellaea:
Thanks, Joe.  Yes, I see four options:

New
Confirmed
Invalid
Wontfix

Would you mind telling me what the lower three mean?

JoeCohen:
See https://docs.codeclimate.com/docs/issues under Issue Statuses.

pellaea:
Thanks!

JoeCohen:
Another approach is [disabling individual cops for part of a file](https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#disabling-cops-within-source-code) by magic comments.

The relevant cops here are Metrics/AbcSize and Metrics/MethodLength.

pellaea:
Thanks.  Depends if it recomplains each time I touch one of these too-long
methods.

JoeCohen:
I think it will do that; it hates you.

pellaea:
It's only fair; I hate it.

pellaea:
Weird, it did it to me again.  There were 6 temporary complaints.  Started looking them up, then discovered the file in question passes rubocop locally.  Looked back and the 6 errors were gone.  Bizarre.

Anyway, rubocop seems content for the moment.  Still want to add a unit test for the content of at least one line of each report, but otherwise done.

JoeCohen:
Yes that's weird.  I didn't see it, but I wasn't watching the report all the time.  The only thing I saw was that it passed.

pellaea:
Let's go with "it hates me".

JoeCohen:
I looked at this, but can't really help.  Nothing jumped out at me from the code.

But we should ask Steve about lat/lon.  My understanding of the code is that for `decimalLatitude` and `decimalLongitude` we export the "best" figures: 
- If the user has supplied exact lat/lon we export those,
- else we export the lat/lon of the center of the Observation location. 
 
The report doesn't distinguish between those two.  So there's no way to know the precision of `decimalLatitude` and`decimalLongitude` -- could be up to 1/2 the diameter of the planet.
Steve might prefer something else. Maybe the observation.lat and .long, plus the Location bounding points. Maybe just leave `decimalLatitude` and`decimalLongitude``nil` if there's no observation.lat and .long.

pellaea:
Good point, I'll ask.  Thanks!

PR #321, opened 2017-06-11
Name controller coverage

mo-nathan:
- Coverage improvements for NameController.
- Also enables local coverage report generation, but should keep coveralls support going.


app/controllers/application_controller.rb
```diff
@@ -1234,15 +1234,15 @@ def redirect_to_next_object(method, model, id)
...
-  private ##########
-
   def find_query_and_next_object(object, method, id)
     # prev/next in RssLog query
     query_and_next_object_rss_log_increment(object, method) ||
       # other cases (normal case or no next object)
       query_and_next_object_normal(object, method, id)
   end
 
+  private ##########
```
JoeCohen: I think `find_query_and_next_object` is called only by ApplicationController#1224. Should it continue to be private?
(I have only a rudimentary understanding when to use `private`.)

test/controllers/name_controller_test.rb
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
JoeCohen: Hmm. Because utf-8 is Ruby's default encoding, I've been omitting (and sometimes removing) similar magic comments in other files.
Should I add/restore them?

test/controllers/name_controller_test.rb
```diff
@@ -3445,6 +3507,14 @@ def test_email_tracking_disable
   end
 
   # ----------------------------
+  #  Mushroom App
+  # ----------------------------
+
+  def test_names_for_mushroom_app
```
JoeCohen: What's the mushroom app?

test/test_helper.rb
```diff
@@ -10,9 +10,17 @@
...
 
-# Coveralls.wear! must occur before any of your application code is required
+# Code to allow both local and coveralls coverage.  From:
+# https://coveralls.zendesk.com/hc/en-us/articles/201769485-Ruby-Rails
+require "simplecov"
 require "coveralls"
-Coveralls.wear!("rails")
+
+# Coveralls.wear!("rails")
+SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
```
JoeCohen: This line caused a deprecation notice when I ran rake.  I don't understand the notice (or how to fix it).
```ruby
$ rake
/vagrant/mushroom-observer/test/test_helper.rb:19:in `<top (required)>': [DEPRECATION] ::[] is deprecated. Use ::new instead.
```

JoeCohen:
PS: I edited this branch directly in GitHub because 
1. I've been having trouble with my VM and my local repo; I didn't want to risk transmitting the problems to this repo; and
2. My editing was trivial.

Please let me know if messed up your work flow.

pellaea:
My take on the subject initially was that character encoding is far too
complex and far too magical; the more things I could write explicitly the
better.  But it feels to me like we've got a good handle on it, and we
haven't seen any strange, unexpected encoding-related bugs in a long, long
time.  So I've be letting you delete those magic comments without comment
(sorry).  I'm good with either.

JoeCohen:
Re character encoding.
I'm curious to hear @mo-nathan's take.
I now see that ruby-docs contains the sentence: "Ruby source files should declare its script encoding by a magic comment even when they only depend on US-ASCII strings or regular expressions." https://ruby-doc.org/core-2.2.0/Encoding.html#class-Encoding-label-Script+encoding
But I can't tell whether that applies only to files which are using non-default encoding, or also to apps which use the default.  If you're supposed to add `# encoding: utf-8`, what's the point of a default?

mo-nathan:
I don't really have an opinion about the character encoding issue.  I lean slightly towards depending on the default since it's what we want and it's one less thing we have to remember, but if either of you feel strongly about it, I'm happy to try to remember to add the encoding to files when I touch them.

app/controllers/application_controller.rb
mo-nathan: It depends on how you count. I added calls to test/controllers/name_controller.rb in order to get coverage and it wouldn't let me do that if it was declared private.

JoeCohen:
On encoding. I also favor depending on the default.

app/controllers/application_controller.rb
JoeCohen: Aha!

test/controllers/name_controller_test.rb
mo-nathan: Interesting.  It's actually my editor, emacs, that adds it.  I just tried deleting it and it re-added it.  I'm sure I can figure out how to delete it if anyone cares.  Otherwise I'm just going to leave it the way it is.

test/test_helper.rb
mo-nathan: Fixed.

test/controllers/name_controller_test.rb
JoeCohen: Perhaps it's something Ray was talking about, but put on hold because MO now works so well responsively.  I'll ask him.  It shouldn't make any difference for this PR, but perhaps we can strip out some code at a later date.

JoeCohen:
The latest CodeClimate/Robocop complaint is the same Cop as last time.  
 If this is also being done by your editor, maybe we should turn off the Cop.
 Do you want me to fix the offense?

test/controllers/name_controller_test.rb
JoeCohen: It predates Ray's involvement with MO development.

PR #322, opened 2017-06-15
Increase size of Name.author to 255 [Delivers #147238161]

JoeCohen:
The limit needed to be increased for a recent Name. See [Pivotal #147238161](https://www.pivotaltracker.com/story/show/147238161).

- Write migration.
- Hand-test migration by migrating up and down, checking resulting schema.
- Migrate


JoeCohen:
Forgot to see if there was a limit in Rails (in addition to the DB).  There is, so I have to fix this.

PR #323, opened 2017-06-16
Fix :string limits

JoeCohen:
Delivers [Pivotal #147238161](https://www.pivotaltracker.com/story/show/147238161)

- Increase Name :string field limits
  - Bump up Name#author limit to accommodate _Imperator_ Authors
  - Increase other limits to accommodate largest possible Scientific Name and Author
  - Validate Name :string fields based on character size (instead of bytesize)  
 
- Validate and truncate :string fields in other tables based on character size (instead of bytesize); update tests accordingly. 
- Remove now-uncalled bytesize StringExtension methods

[Delivers #147238161]


pellaea:
Thanks for taking care of this, Joe!

JoeCohen:
codeclimate report 
Please ignore it.  This PR page says "**1 new issue** (5 fixed)", but details say: w"Passed No new issues were found." https://codeclimate.com/repos/58cfc16d1035280287001d93/pull/323  (I think the difference is one issue I marked WONTFIX because it was preexisting, time-consuming, and not the concert of this PR.)

JoeCohen:
coverage/coveralls report
Please ignore it. There are no newly uncovered lines. The decrease in the Coveralls percentage grade is minute. (I think it's due to the removal of tests which covered methods removed in this PR.)

app/models/name.rb
```diff
@@ -2555,29 +2561,75 @@ def save_with_log(log = nil, args = {})
...
+    255
+  end
+
+  def self.display_name_limit
+    285
+  end
+
+  def self.text_name_limit
+    100
+  end
```
JoeCohen: It would be nice to grab these numbers from the db (instead of hard-coding them).  But I couldn't figure out how to do that.

JoeCohen:
I removed all but one use of `bytesize`.  The remaining one is at line 10 in https://github.com/MushroomObserver/mushroom-observer/blob/master/app/classes/api/parse_string.rb. I left it there because I wasn't sure if it could be removed.

pellaea:
Yes, you can remove that, change that to a normal char length.  Its purpose
there is to give the api clients appropriate error message if they are
trying to set a string in the db too long.  Ideally all the set_xxx
parameters should be declared with appropriate limits based on the db
limits.  I don't think any of the set_xxx parameters are tested outside of
what's needed to post an observation, and there are no limited strings
there (I think).

JoeCohen:
Oops! Forgot to Rake.  Will re-open when ready.

JoeCohen:
Thanks.  Will do.

JoeCohen:
I think it's fixed. (But there seems to be >= 5 minutes delay between (a) when I push commits to my personal GitHub repo and (b) when those commits show up in the MO repo.)

app/controllers/account_controller.rb
```diff
@@ -160,7 +160,7 @@ def verify # :nologin:
           elsif password != confirmation
             @user.errors.add(:password_confirmation,
                              :validate_user_password_no_match.t)
-          elsif password.length < 5 || password.bytesize > 40
+          elsif password.length < 5 || password.size > 40
```
JoeCohen: At this point is `password` the real password or the hash?  
If it's the hash then can we validate its size here?  Also see same issue in User.

app/models/user.rb
```diff
@@ -969,28 +969,24 @@ def crypt_password # :nodoc:
...
+    elsif login.length < 3 || login.size > 40
       errors.add(:login, :validate_user_login_too_long.t)
     elsif (other = User.find_by_login(login)) && (other.id != id)
       errors.add(:login, :validate_user_login_taken.t)
     end
 
-    if password.to_s.blank?
-      # errors.add(:password, :validate_user_password_missing.t)
-    elsif password.length < 5 || password.bytesize > 40
+    if password.to_s.present? && (password.length < 5 || password.size > 40)
```
JoeCohen: At this point is `password` the real password or the hash?  
If it's the hash then can we validate its size here?  Also see same issue in Account Controller.

pellaea:
No, it is plain text here.  It gets hashed automagically by rails in a
before_save hook, I think.

JoeCohen:
Yup. I should have figured that out.  (it's `before_create :crypt_password`.)

I'm done with this branch unless someone has suggestions.

pellaea:
Great, thanks again, Joe.  I'll take a look later tonight maybe.

JoeCohen:
Thank you Jason. I'll try to get started on creating more work for you; no good deed goes unpunished.

app/models/name.rb
```diff
@@ -2555,29 +2561,75 @@ def save_with_log(log = nil, args = {})
...
-      errors.add(:text_name, :validate_name_text_name_too_long.t)
+    if author.to_s.size > Name.author_limit
+      errors.add(:author, (:validate_name_author_too_long.t +
+        " " + :MAXIMUM.t + ": " + Name.author_limit.to_s))
     end
-    if search_name.to_s.bytesize > 200
-      errors.add(:search_name, :validate_name_search_name_too_long.t)
+    if text_name.to_s.size > Name.text_name_limit
+      errors.add(:text_name, (:validate_name_text_name_too_long.t +
+        " " + :MAXIMUM.t + ": " + Name.text_name_limit.to_s))
```
pellaea: Good, thank you for adding the limit to the error message.

app/models/name.rb
```diff
@@ -2555,29 +2561,75 @@ def save_with_log(log = nil, args = {})
...
+  #     Adds the space between name and author; "**__" and "__**" (8 chars)
+  #     around every epithet, grouping genus and species. Infrageneric ranks
+  #     win, making as many as 5 separate bold epithets or epithet pairs.
+  #     That adds up to 8*5 + 1 = 41.
+
+  # Numbers are hard-coded (rather than calculated) to make it easier to copy
+  # them to migrations.
+  def self.author_limit
+    # max if text_name_limit == 100, and we want to keep sort_name_limit < 256
+    134
```
pellaea: But we don't care if display_name > 255?  Can I assume that if we shortened author to keep display_name < 256, then author would be too short for a specific case already seen?

JoeCohen: Short answer:  Yes.  To be precise:
If I increase author to accommodate the specific case, keep text_name unchanged, and adjust the other limits (search_name, display_name, sort_name) to their theoretical maxima,  then
 display_name == 261.

Long answer:
I want to increase `author` because of the specific case.  
Keeping `text_name` constant, I have to increase `search_name`, `display_name`, and `sort_name`.  
Because of the specific case, `author` must be increased to at least 120.  
With text_name unchanged at 100, the theoretical maxima are 
search_name: 120 + 100 + 4 = 224
display_name: 120 + 100 + 41 = 261
sort_name: 120 + 100 + 21 = 241

display_name is necessarily over the default :string limit, so I don't care if it goes up a little more.  The next largest is sort_name, which I can keep to 255. Therefore I kept sort_name at 255, and worked backwards from there.

author: 255 - 100 - 21  = 134
search_name: 134 + 100 + 4 = 238
display_name: 134 + 100 + 41 = 275

**I need to fix `display_name_limit` and the migration. They should be 275, not 285.**

JoeCohen:
Please give me a few minutes to  fix display_name_limit and the migration.

JoeCohen:
Will probably take more than a few minutes -- I have to do some branch switching, cleaning, migrating, fixing, etc. 
So I will close this, and re-open when it's fixed.

JoeCohen:
Corrected display_name_limit.

pellaea:
Thanks, Joe.  Yeah, I thought that might be the reason.  My thought, in
that case, was "what are the odds we will have to do this again when an
even longer author comes around?"  Of course, bumping up the size again
with another very short PR will be trivial, so maybe "who cares?"

JoeCohen:
My guess is it's not likely to happen again soon. (But P it will happen sometime => 100%.)
Should we just forget about this PR?  
What effect will it have on disk storage? (I don't understand whether MySQL automatically consumes the :limit for each column for each record, or whether it uses only the space it needs for the actual field data.) We have close to 60,000 Names.
What about search speed?

pellaea:
I think you misunderstand.  I was wondering if we shouldn't just bump it up
way bigger, and have done with it.  I am definitely in favor of this PR!

As for performance, space, etc... Who knows?  I suppose you could do some
benchmarks before and after to at least quantify any hit to either.  But
I'm having a hard time being very concerned about it.  Don't do it on my
account.

JoeCohen:
I'll think about it overnight.

JoeCohen:
I am now inclined to: 
- leave the Author`limit` as it was (100);
- give a flash error suggesting use of "et al." when a user inputs an `author` longer than the limit (or when search_name, display_name, and sort_name would be too long);
- keep other improvements of this branch (e.g., `size` instead of `bytesize`). 

This goes back to something @pellaea said (in an email?) about using "& al." and [ICN Recommendation 46C.2](http://www.iapt-taxon.org/nomen/main.php?page=art46) (which I somehow overlooked until now):
>46C.2. After a name published jointly by more than two authors, the citation should be restricted to the first author followed by “et al.” or “& al.”, except in the original publication.
>>46C.Ex.2. _Lapeirousia erythrantha_ var. _welwitschii_ (Baker) Geerinck, Lisowski, Malaisse & Symoens (in Bull. Soc. Roy. Bot. Belgique 105: 336. 1972) should be cited as _L. erythrantha_ var. _welwitschii_ (Baker) Geerinck & al.

pellaea:
Yes!  I didn't know about the ICN recommendation, but I'm totally behind
using "et al." for these ridiculous molecular-defined genera!

Good idea, Joe.

JoeCohen:
Okay!  I will close this PR, modify the branch, and then make a new PR.

JoeCohen:
Leave author limit at 100, adjust other limits
- Leave author as it was before this PR
- Adjust search_name, sort_name, and display_limits to their theoretical maxima, given author and text_name limits.

Revise Name validation
- Revise flash message for `author` too long -- suggest using first author plus "et al."
- Remove validation of search_name, sort_name, and display_limits. These fields are always calculated by the app based on :text_name and :author; they are never supplied by the user or API. The :limits have been increased to the theoretical max, so length should never be exceeded. The old message no longer makes sense. If there's an error, it now gets treated like any other glitch which occurs when trying to save a Name.

Lots of RuboCopping
If this is too much to review, it's easy to revert.

JoeCohen:
Connection to CodeClimate broken. I will seek help fixing this.
Once connection is restored, my RuboCopping will likely provoke new CodeClimate complaints. I will fix them, and post a GitHub Comment when this is ready to review.

app/models/interest.rb
```diff
@@ -83,9 +83,9 @@ def summary
       target_type.underscore.to_sym.l + ": " +
       (target ? target.unique_format_name : "--")
   end
-  alias_method :text_name, :summary
+  alias text_name summary
```
pellaea: What is the difference between alias_method and alias?  Why is alias preferred here?  Is it always preferred to alias_method, or are there still cases where alias_method is used?

JoeCohen: The difference is lexical scope.  See https://github.com/bbatsov/ruby-style-guide#alias-method-lexically for: discussion, including why alias is preferred here (and I suppose why it's the RuboCop default), and when `alias_method` is preferred.  (RuboCop tries to catch those cases.)

PR #324, opened 2017-06-17
Fix display of deprecated :Group Names

JoeCohen:
Delivers [Pivotal #147206581](https://www.pivotaltracker.com/story/show/147206581)

Deprecated :Group Names were displayed instead of roman.   The problem was that when Name#parse_group called parse_name(str_without_group(str)), it neglected to pass in `deprecated`.

In order to test this, I ended up doing the following:
- Changing Name#parse_name to use keyword parameters
- When needed, change calls to Name#parse_name throughout the app to use those keywords
- Add an additional `deprecated` parameter to NameTest#do_name_parse_test
- Add test of parsing some deprecated Names

### DIFF
--- a/app/controllers/name_controller.rb
+++ b/app/controllers/name_controller.rb
@@ -504,7 +504,7 @@ class NameController < ApplicationController
     in_str = Name.clean_incoming_string("#{text_name} #{author}")
     in_rank = params[:name][:rank].to_sym
     old_deprecated = @name ? @name.deprecated : false
-    parse = Name.parse_name(in_str, in_rank, old_deprecated)
+    parse = Name.parse_name(in_str, rank: in_rank, deprecated: old_deprecated)
     if !parse || parse.rank != in_rank
       rank_tag = :"rank_#{in_rank.to_s.downcase}"
       fail(:runtime_invalid_for_rank.t(rank: rank_tag, name: in_str))
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -1572,7 +1572,7 @@ class Name < AbstractModel
   end
 
   # Parse a name given no additional information.  Returns a ParsedName instance.
-  def self.parse_name(str, rank = :Genus, deprecated = false)
+  def self.parse_name(str, rank: :Genus, deprecated: false)
     str = clean_incoming_string(str)
     parse_group(str, deprecated) ||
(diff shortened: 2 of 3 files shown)


JoeCohen:
I could have done this without adding keyword parameters to Name#parse_name, which would have required fewer changes, like this:
>    def do_name_parse_test(str, args)
>>~~parse = Name.parse_name(str)~~
    _# If final key of arg is :deprecated", remove it and_
    _# separately pass it to Name.parse_name as the final argument_
    _deprecated = args.delete(:deprecated) || false
    _# :Genus is the default 2nd argument of Name.parse_name
    _parse = Name.parse_name(str, :Genus, deprecated)

This looked kludgy to me, and creates an unnecessary coupling between the test and the model.  The test knows 3 more things about the model: Name.parse_name has a `rank` argument; it's the 2nd argument; and its default is :Genus.  

I eventually decided that adding keyword parameters to Name#parse_name was cleaner. But -- as it requires changes throughout the app -- may be riskier; I may have unintentionally created a problem which isn't exposed by the tests.

What do you think?

JoeCohen:
This PR is not retroactive; it has no effect on deprecated :Group Names with incorrect display_name's.  Is it worth someone's trouble to write a script to fix them in production? (If I used the UI, they would be laborious to find, and -- once found -- would require an involved fix.)  Alternatively, is it worth someone's trouble to write a script to simply find them.  (I don't want to try that myself; too much security risk in getting to the production server before I fix my router issues, and I don't know the first thing about scripts.)

pellaea:
I think you did great.  Thanks for warning me.  I can pay special attention
to your conversion of all the calls to Name.parse_name to help minimize the
risk.

pellaea:
There are only four: 25230, 27051, 30521, 58368.

Question: is it just me, or should "group" also be in boldface if not
deprecated?  I guess, I feel (now) like authors and everything should be in
boldface if not deprecated.  Yes, only the name should be in italics, but
"group" is important!  Oh well, ignore me, beyond the scope of this PR.
Apparently I feel very critical today...

JoeCohen:
Re existing bad names.  I'll fix by hand after this gets into production.

Re bolding group. Interesting point. We've never done that, so I didn't think about it when I made the changes to groups/clades back in #291. I lean toward not bolding: less ink on the screen.

PR #325, opened 2017-06-24
Changed MycoFlora report format.

pellaea:


app/classes/observation_report/mycoflora.rb
```diff
@@ -29,41 +27,63 @@ def labels
...
+      return nil unless row.obs_lat.blank?
+      r1 = lat_radius(row)
+      r2 = long_radius(row)
+      return nil if !r1 || !r2
+      (r1 > r2 ? r1 : r2).to_f.round
+    end
+
+    def lat_radius(row)
+      return nil if row.loc_north.blank? || row.loc_south.blank?
+      (row.loc_north - row.loc_south) / 360 * 2 * 3.14159 * 6_371_000 / 2
```
JoeCohen: Maybe `Math::PI` instead of `3.14159`?

app/classes/observation_report/mycoflora.rb
```diff
@@ -29,41 +27,63 @@ def labels
...
+    end
+
+    def long_radius(row)
+      return nil if row.loc_east.blank? || row.loc_west.blank?
+      (row.loc_east - row.loc_west) / 360 * 2 * 3.14159 * 6_371_000 *
```
JoeCohen: See above question re `MATH::PI`.

pellaea:
Oh yes, thanks, didn't know that was there.  Will do.

BTW, I'm going to push this through immediately, since Steve already has my
specimens and is waiting for the report.  Thanks so much for reviewing it
so quickly.

JoeCohen:
Only issue is whether the added precision of Math::PI might require changing expectations in any of the tests.

pellaea:
Naw, it's not precise enough to make a difference.  Tests still passed just
fine.

I'm sorry I haven't reviewed your PRs yet.  I was about to, but realized
that I'm too tired to give them the attention I need to.  Just want to get
this stupid report out tonight so I have one fewer thing hanging over me...

PR #326, opened 2017-06-30
Sequences

JoeCohen:
Lets Users add Sequences to Observations
[Delivers #121269103]


app/helpers/sequence_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+        prefix: "https://www.ncbi.nlm.nih.gov/nuccore/" },
+      { name:   "ENA",
+        home:   "http://www.ebi.ac.uk/ena",
+        prefix: "http://www.ebi.ac.uk/ena/data/view/" },
+      { name:   "UNITE",
+        home:   "https://unite.ut.ee/",
+        prefix: "https://unite.ut.ee/search.php?qresult=yes&accno=" }
+    ]
+  end
+
```
pellaea: Should this go in config/consts.rb instead?

pellaea:
Looks great, even query stuff.

Request: can we add sequences to observer/show_user, and also give a contribution for them?  Look in app/classes/site_data.rb, FIELD_WEIGHTS especially.  Since it's a simple "has_many" relationship between users and sequences, it should be trivial to add sequences in.  (Also, feel free to put this request off for another PR.)

app/helpers/sequence_helper.rb
JoeCohen: That's a great question.  I hadn't thought about where this belongs, or whether there's a better design.  
Maybe pull this, and similar stuff into one place, with the same design for each. (e.g., MycoBank, MycoPortal, eventually will need MycoMap, and maybe even also add Google, Google Images, EOL).  
I don't know where, or what the design should be. What do you think -- you're infinitely more adept at that than I.

JoeCohen:
Re Contribution for Sequence  
Sounds good to me, but how many points? And what does @nwilson-eol think? 
I'd prefer to do it in another PR so that Nathan can weigh in.

pellaea:
Simplest solution -- and possibly even a good one -- is just take this
entire structure and move it into config/consts.rb.  Then can refer to it
as MO.sequence_archives (or whatever you call it) from anywhere, including
views.

Another option is to move it into app/models/sequence.rb, e.g., a class
method like Sequence.archives.

I'm just thinking about where I would go looking for this if I didn't know
where it was and wanted to add another entry to the structure.  First guess
would be consts.rb, second the model.  Maybe it's just me, but I would not
think "view helper".

pellaea:
Separate PR works just fine.  I just saw the show_user page a little while
ago and noticed that a number of other things probably should be added,
too, e.g., specimens and glossary_terms.

JoeCohen:
Re archive stuff:
Agree about moving it out of the helper.
If I'm right about combining all the outside url portions in one place, what about in its own Class? (Don't have a good name for it. Maybe WebLink?)

PS: I see this branch now has a conflict in the schema (because of the merger of  :string limits branch.) I'll have to resolve it. Trivial, but probably won't have time until tomorrow a.m.

JoeCohen:
Latest commit:
- Moves the archive stuff out of the helper
- Validates the format of bases
- Does some refactoring of Sequence

I decided to move the archive stuff to its own class (classes/web.rb) rather than config/consts.rb.  Please see if you like that as a location, and whether you have any ideas for better design.

PR #328, opened 2017-08-04
Notes templates and fields

JoeCohen:
- Lets Users create Observation notes via a notes template in User preference.
- Revises internal representation of Observation notes from a String to a serialized Hash, with the hash keys comprising the items in the notes template plus `:Other`.
- Includes a data migration of Observation notes from the old String to the new Hash.

The Hash keys are:
  - key(s) from the User's notes template, and
  - a general Other key supplied by the system.

Keys with empty values are not saved.

The notes template is a comma-separated list of arbitrary keys (except for
the following which are reserved for the system: "Other", "other", etc., and
translations thereof.
Sample observation.notes
```ruby
 { }                                        # no notes
 { Other: "rare" }                          # generalized notes
 { Cap: "red", stem: "white" }              # With only user-supplied keys
 { Cap: "red", stem: "white", Other: rare } # both user-supplied and general
```
The create Observation form displays separate fields for the keys in the
following order:
  - each key in the notes template, in the order listed in the template; and
  - Other.

The edit Observation form displays separate fields in the following order:
  - each key in the notes template, in the order listed in the template;
  - each "orphaned" key -- one which is neither in the template nor Other;
  - Other.

The show Observation view displays notes as follows, with Other underscored:
  > no notes - nothing shown

  only generalized notes:
  > Notes: value

  only user-supplied keys:
  > Notes:
    <ins>First user key</ins>: value
    <ins>Second user key</ins>: value
    ...

  both user-supplied  and general Other keys:
  > Notes:
    <ins>First user key</ins>: value
    <ins>Second user key</ins>: value
    ...
    <ins>Other</ins>: value

Because keys with empty values are not saved in observation.notes, they are
not displayed with show Observation.

Notes are exported as shown, except that the initial "Notes:" caption is
omitted, and any markup is stripped from the keys.


app/models/observation.rb
```diff
@@ -242,6 +264,192 @@ def lat_long_dubious?
...
+  #
+  # Notes are exported as shown, except that the intial "Notes:" caption is
+  # omitted, and any markup is stripped from the keys.
+
+  serialize :notes
+
+  # value of observation.notes if there are no notes
+  def self.no_notes
+    {}
+  end
```
JoeCohen: Question: should this method (and some others) be promoted to AbstractModel?

pellaea:
I don't see the advantage.  Would other models benefit from having a
no_notes method?  Feels like it would just clutter other models
unnecessarily.  But maybe I'm missing something.

app/views/api/_observation.xml.builder
```diff
@@ -18,7 +18,14 @@ xml.tag!(tag,
   xml_detailed_object(xml, :consensus_name, object.name)
   xml_confidence_level(xml, :confidence, object.vote_cache)
   if detail
-    xml_html_string(xml, :notes, object.notes.to_s.tpl_nodiv)
+    xml.notes do
```
JoeCohen: Let's review this to see if one of your suggested renderings makes more sense, e.g., 
```
<fields>
  <field>
    <key>Cap</key>
    <value>red</value>
  </field>
  etc.
</fields>
<notes>Other notes, like before.</notes>
```
Does that feel any more intuitive to you?

You know what would be really sweet?  Try this for kicks:
```
<notes>
  <cap>red</cap>
  etc.
  <other>Other notes.</other>
</notes>
```
Of course this requires that all field names be simple ascii alphanumeric strings, maybe with spaces replaced by underscores.  ... Yeah, blah, that really is begging for trouble, isn't it?  I'm just trying to imagine use-cases like accessing common herbarium fields like habitat, substrate, etc. It would be really cool to be able to refer to them directly like this:

observation.fields.substrate

as opposed to having to look through all the fields searching for the one you want.  Of course, in the JSON format this is a moot point, and that's probably how everyone is going to be using the API in the future, anyway...

JoeCohen: Re CodeClimate issues.  Will deal with those tomorrow.
Re promotion.  You're right. I had this my notes, but don't know what I was thinking.

JoeCohen:
I'm closing this PR because the graph of this branch is becoming too messy.  I will delete the branch in my remote repo, then push it from my local machine (where it has been rebased on master), and make a new PR.

PR #331, opened 2017-08-11
Notes template

JoeCohen:
- Lets Users create Observation notes via a notes template in User preference.
- Revises internal representation of Observation notes from a String to a serialized Hash, with the hash keys comprising the items in the notes template plus `:Other`.
- Includes a data migration of Observation notes from the old String to the new Hash.

The Hash keys are:
  - key(s) from the User's notes template, and
  - a general Other key supplied by the system.

Keys with empty values are not saved.

The notes template is a comma-separated list of arbitrary keys (except for
the following which are reserved for the system: "Other", "other", etc., and
(some of ) their translations).
Sample observation.notes
```ruby
 { }                                        # no notes
 { Other: "rare" }                          # generalized notes
 { Cap: "red", stem: "white" }              # With only user-supplied keys
 { Cap: "red", stem: "white", Other: rare } # both user-supplied and general
```
The create Observation form displays separate fields for the keys in the
following order:
  - each key in the notes template, in the order listed in the template; and
  - Other.

The edit Observation form displays separate fields in the following order:
  - each key in the notes template, in the order listed in the template;
  - each "orphaned" key -- one which is neither in the template nor Other;
  - Other.

The show Observation view displays notes as follows, with Other underscored:
  > no notes - nothing shown

  only generalized notes:
  > Notes: value

  only user-supplied keys:
  > Notes:
    <ins>First user key</ins>: value
    <ins>Second user key</ins>: value
    ...

  both user-supplied  and general Other keys:
  > Notes:
    <ins>First user key</ins>: value
    <ins>Second user key</ins>: value
    ...
    <ins>Other</ins>: value

Because keys with empty values are not saved in observation.notes, they are
not displayed with show Observation.

Notes are exported as shown, except that the initial "Notes:" caption is
omitted, and any markup is stripped from the keys.


app/views/api/_observation.xml.builder
```diff
@@ -30,6 +30,18 @@ xml.tag!(tag,
         xml_detailed_object(xml, :sequence, sequence)
       end
     end
+    xml.notes do
```
JoeCohen: Let's review this to see if one of @pellaea's suggested renderings makes more sense, e.g., 
```
<fields>
  <field>
    <key>Cap</key>
    <value>red</value>
  </field>
  etc.
</fields>
<notes>Other notes, like before.</notes>
```
Does that feel any more intuitive to you?

You know what would be really sweet?  Try this for kicks:
```
<notes>
  <cap>red</cap>
  etc.
  <other>Other notes.</other>
</notes>
```
Of course this requires that all field names be simple ascii alphanumeric strings, maybe with spaces replaced by underscores.  ... Yeah, blah, that really is begging for trouble, isn't it?  I'm just trying to imagine use-cases like accessing common herbarium fields like habitat, substrate, etc. It would be really cool to be able to refer to them directly like this:

observation.fields.substrate

as opposed to having to look through all the fields searching for the one you want.  Of course, in the JSON format this is a moot point, and that's probably how everyone is going to be using the API in the future, anyway...

app/views/observer/_form_observation_notes.html.erb
```diff
@@ -0,0 +1,70 @@
+<% # Notes section of create_observation form %>
+<div class="row push-down">
+  <%= content_tag(:span, "#{:NOTES.t} ", class: "bold") %>
+  <% if @observation.form_notes_parts(@user) == [Observation.other_notes_part]
+       part = Observation.other_notes_part %>
+
+    <div class="row">
+      <div class="col-xs-12 col-sm-6 max-width-text">
+        <div>
+          <textarea class="form-control" rows="10"
```
JoeCohen: This ugly bit of html is because I want the the notes to appear in the params hash as params[:observation][:notes][:caption1] etc., rather than as params[:notes][:caption1]. 

Is there a better way to do this using Rails helpers?  I couldn't figure it out. (Maybe possible with wrap_parameters? See http://api.rubyonrails.org/classes/ActionController/ParamsWrapper/ClassMethods.html#method-i-wrap_parameters.)

app/views/observer/_form_observation_notes.html.erb
```diff
@@ -0,0 +1,70 @@
...
+          id=<%= @observation.notes_part_id(part) %>
+          name=<%= @observation.notes_part_name(part) %>
+          ><%= @observation.notes_part_value(part) %></textarea>
```
JoeCohen: This ugly bit of html (the textarea lines 10-13) is because I want the the notes to appear in the params hash as params[:observation][:notes][:caption1] etc., rather than as params[:notes][:caption1]. 

Is there a better way to do this using Rails helpers?  I couldn't figure it out. (Maybe wrap_parameters? See http://api.rubyonrails.org/classes/ActionController/ParamsWrapper/ClassMethods.html#method-i-wrap_parameters.)

app/views/api/_observation.xml.builder
```diff
@@ -30,6 +30,18 @@ xml.tag!(tag,
...
+        end
+      end
+    end
+    xml_datetime(xml, :created_at, object.created_at)
+    xml_datetime(xml, :updated_at, object.updated_at)
+    xml_integer(xml, :number_of_views, object.num_views)
+    xml_datetime(xml, :last_viewed, object.last_view)
     xml.namings(number: object.namings.length) do
       object.namings.each do |naming|
         xml_detailed_object(xml, :naming, naming, true)
```
JoeCohen: I don't understand the comment on line 51: "# Do it this way, else will not use eager-loaded image instance."

Also note the difference from  _observation.json.builder analagous lines (copied below), which:
- when finding the thumbnail, build a list of non-thumbnails (`other_images`), and
- then use that list, versus again iterating over all the images.  
```ruby
  object.images.each do |image|
    # Do it this way, else will not use eager-loaded image instance.
    if image.id == object.thumb_image_id
      json.primary_image { json_detailed_object(json, image) }
    else
      other_images << image
    end
  end
  json.images other_images.map do |image|
    json_detailed_object(json, image)
  end
```
Comment:
- The two builders really should work the same way.
- Is it worth creating the new list? The difference between `object.images` and `other_images` is just one image.

JoeCohen:
I will have to close this PR, then make a new one.  
(I messed up a rebase merge, seemingly leaving this branch with circular conflicts between my local repo and my Github repo.)

PR #334, opened 2017-08-14
Rubocop-centric refactor of model tests

JoeCohen:
This is 99% fixing of RuboCop offenses.
The goal is to reduce CodeClimate offenses during CI. 
Often if a developer touches a file or method, CodeClimate will flag pre-existing violations as new, causing a CodeClimate failure during CI.  
With all the RuboCop offenses fixed in the model test files, developers should be able to modify these files without re-incarnating old offenses. (Caveat: a few files still contain a tiny number of RuboCop offenses which I did not resolve.) 

The branch also contains a small number of non-RuboCop changes, such as:
- fixing bugs in tests;
- removing duplicate tests, moving tests to the correct file.


JoeCohen:
CodeClimate says 47 issues to fix!  Looks like there are two kinds:
- "Block has too many lines" in some rakes files
- issue with %i and %w delimiters.
I'll have to look into this tomorrow.  I'm not sure if I touched the rake files, so maybe there's a new CodeClimate configuration.
On the delimiters, it's possible that the CodeClimate default is overriding the RuboCop default. When I did this branch, I used RuboCop on the VM, and made changes accordingly.

PR #336, opened 2017-08-16
Rubocop tests controller

JoeCohen:
This is 99% fixing of RuboCop offenses.
There are two goals:
1. Reduce CodeClimate offenses during CI.  See # 334.
2. In conjunction with 1, expedite the (soon-to-start) upgrade to Rails 5.0.  That upgrade is likely to cause a bunch of deprecation warnings in test files. When fixing those, I don't want to also have to contend with pre-existing RuboCop offenses.  So I'm getting rid of them in advance.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,12 @@ AllCops:
     - public/design_test/jquery-1.5.2.min.js
     - "tmp/**/*"
 
+###################### Global Disables #########################################
+
+# Disable using Rails < 5.0; it's relevant only to >= 5.0
+Rails/HttpPositionalArguments:
+  Enabled: False
+
 ###################### Metrics #################################################
 
 Metrics/AbcSize:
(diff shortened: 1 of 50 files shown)


JoeCohen:
CodeClimate vs RuboCop.
CodeClimate says: "%w-literals should be delimited by ( and )", citing Rubocop.
But Rubocop delimiters -- which I have not changed -- are:
` {"default"=>"()", "%i"=>"[]", "%I"=>"[]", "%r"=>"{}", "%w"=>"[]", "%W"=>"[]"}`
And when I run RuboCop locally (on the VM), that's what Rubocop demands.
I don't know why behavior is different with CodeClimate on GitHub.  Is CodeClimate silently overriding the RuboCop defaults.  
I will investigate.  
There's probably an easy work-around: just repeat the defaults in .rubocop.yml.

PR #339, opened 2017-08-27
Remove conferences

JoeCohen:
Removes all code and tables related to Conferences.

It's unused, but must be maintained if we keep it. (E.g., in Rails updates, we'll need, at a minimum, to deal with routes.)  Because of impending update to Rails 5.0, I decided to ditch it. I think cost of maintainance > hypothetical future usefulness.

Also see following email chain:
>On Sat, Aug 12, 2017 at 2:14 PM, Joseph Cohen <jdcohenesq@gmail.com> wrote:
Nathan:
Could I delete the Conference code (controller, models, tests, views)?
> - - -
>Nathan Wilson to me, Jason Aug 13
I guess. Doesn't seem worth worrying about and some day it might be useful, but I don't really care.
-Nathan


JoeCohen:
I will let this PR sit until Monday or Tuesday. If there are no objections, I'll merge it into master (and ror-5, which catalyzed this PR).

PR #342, opened 2017-09-01
Change deprecated `render :text` to `render :plain`

JoeCohen:
**Exception**: observation_controller/other.rb, where it's changed to `render(html`

**Question**: should others be `html` instead of `plain`?

See http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#rendering-content-from-string,
https://github.com/rails/rails/issues/12374

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -801,7 +801,7 @@ class AccountController < ApplicationController
     if redirect
       redirect_to(redirect)
     else
-      render(text: "", layout: true)
+      render(plain: "", layout: true)
     end
   end
 
--- a/app/controllers/ajax_controller.rb
+++ b/app/controllers/ajax_controller.rb
@@ -51,7 +51,7 @@ class AjaxController < ApplicationController
   rescue => e
     msg = e.to_s + "\n"
     msg += backtrace(e) if Rails.env != "production"
-    render(text: msg, status: 500)
+    render(plain: msg, status: 500)
   end
 
   def prepare_parameters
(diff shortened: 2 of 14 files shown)


pellaea:
Whenever I used "render text: xxx" I meant for it to be plain text, not
html.

JoeCohen:
1.  Thanks for the comment!
This branch is a looong way from even a tentative merge with master.
I made this PR within the ror50 branch so that I could put comments on Github. It will help me keep track of things. Please don't feel obliged to look at the code, Github comments. (Of course, your comments are always welcome and helpful.)
I am going to merge this into the ror50, and begin adding comments so that I don't forget what I've done and need to do.

2. The only place where I used `html` instead of `plain` is observer_controller/other.rb#test_flash_redirection:
```ruby
      # (sleight of hand to prevent localization_file_text from complaining
      # about missing test_flash_redirection_title tag)
      @title = "test_flash_redirection_title".to_sym.t
      render(html: "", layout: "application")
```
If I use `plain` here I get:
>  1) Error:
AmateurTest#test_language_tracking:
NoMethodError: undefined method 'document' for nil:NilClass
    test/session_extensions.rb:266:in 'click'
    test/integration/amateur_test.rb:307:in 'run_test'
    test/integration/amateur_test.rb:301:in 'test_language_tracking'

pellaea:
Sure, I understand, was just answering your question.  Good catch re:
test_flash_redirection.  Strange test, but no doubt important for some
reason!

JoeCohen:
Your answer (re intending plain/text was indeed very helpful. It means that
I can put this set of deprecation fixes to bed.

JoeCohen:
My proposed method of leaving GitHub comments doesn't work. I can comment only on commits and PRs, not code generally.  Maybe I just need to leave comments in the code.

PR #343, opened 2017-09-02
Fix API to_hash deprecation warning

JoeCohen:
>DEPRECATION WARNING: #to_hash unexpectedly ignores parameter filtering, and will change to enforce it in Rails 5.1. Enable `raise_on_unfiltered_parameters` to respect parameter filtering, which is the default in new applications. For the existing deprecated behaviour, call #to_unsafe_h instead. (called from params_to_api_args at /vagrant/mushroom-observer/app/controllers/api_controller.rb:83)

Fixed by using existing deprecated behavior.
But in the long run it would be better to filter the params.
Would something like this work:
 Each model has an identically named method (e.g., `permitted`)
 (unlike the current whitelisted_x_argument)s which method is
called here depending on type


app/controllers/api_controller.rb
```diff
@@ -80,10 +80,16 @@ def rest_query(type)
...
   # Massage params hash to proper args hash for api
   def params_to_api_args(type)
-    args = params.to_hash.symbolize_keys.except(:controller)
+    # In Rails 4, was:
+    #  args = params.to_hash.symbolize_keys.except(:controller)
+    # TODO: Can it be made safer?
+    # Perhaps each model has a method with same name (e.g., `permitted`)
+    # (insteead of the current `whitelisted_x_arguments`) which method gets
+    # called here depending on type
+    args = params.to_unsafe_h.symbolize_keys.except(:controller)
```
pellaea: Note that API class validates arguments, so it doesn't need to be done here.

PR #345, opened 2017-11-04
Sequence codes

JoeCohen:
- Fixes "NoMethodError: undefined method `first' for nil:NilClass" which sometimes occurred when editing a Sequence
- Allows all IUPAC valid base codes - including ambiguous bases - in MO Sequences. (Fixes https://www.pivotaltracker.com/story/show/152022995)
- Prevents long Sequence from scrolling right indefinitely. (Fixes https://www.pivotaltracker.com/story/show/152023063)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -23,6 +23,12 @@ ActiveRecord::Schema.define(version: 20170827000729) do
     t.datetime "verified"
   end
 
+  create_table "ar_internal_metadata", primary_key: "key", force: :cascade do |t|
+    t.string   "value",      limit: 255
+    t.datetime "created_at",             null: false
+    t.datetime "updated_at",             null: false
+  end
+
   create_table "articles", force: :cascade do |t|
     t.string   "title",      limit: 255
     t.text     "body",       limit: 65535
--- a/app/models/sequence.rb
+++ b/app/models/sequence.rb
@@ -41,13 +41,14 @@ class Sequence < AbstractModel
   ##############################################################################
 
   # matchers for bases
-
   BLANK_LINE_IN_MIDDLE = /(\s*)\S.*\n # non-blank line
                           ^\s*\n      # followed by blank line
                           (\s*)\S/x   # and later non-whitespace character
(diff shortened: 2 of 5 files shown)


db/schema.rb
```diff
@@ -23,6 +23,12 @@
     t.datetime "verified"
   end
 
+  create_table "ar_internal_metadata", primary_key: "key", force: :cascade do |t|
```
JoeCohen: This is Rails 5 stuff. I'm guessing it leaked into master because of an issue in my changing branches while developing.  I don't think its doing any harm, so I'll leave it for now.

JoeCohen:
@pellaea: If you don't have a chance to look at this, I will probably merge it and push it out this afternoon. 
It responds to several good comments from users; fixes an error which recently popped up in the Chron Daemon; and It's small and contained, so it should not create new problems.

pellaea:
In the gentlest way possible... I would like to suggest that you're rebasing your PRs incorrectly!  Have you documented what you do somewhere?  I seem to remember doing it successfully once, and it basically just involved fetching master from origin, and merging it into my feature branch.  But my PRs are never as complicated as yours.

No question, in any case, that git makes this task -- updating feature branches with respect to recent changes on master -- very confusing.  (Wouldn't say difficult, because if only we knew WTF we were doing, it would no doubt be easy!! :)  So it would be worth figuring out how to do it right once and for all and documenting it.

JoeCohen:
I'm clearly doing something wrong, but am uncertain exactly what.
It's probably due to some combination of interim pushes of my feature
branch from my local repo to my GitHub repo, plus rebasing more than once.
(When I rebase, I'm doing this:

   - pull master from origin to my local repo;
   - rebase (merge) master into my feature branch in my local repo.

That's where the trouble seems to start. At that point, my GitGUI tells me
that my local feature branch is both behind and ahead of my remote feature
branch.

I'm working on a feature branch now. I'll see what happens if I do only one
rebase. I.e., when it's absolute ready for prime time, I will do a single
rebase, and only then push it to my GitHub repo.

PR #346, opened 2017-11-04
Fix bug: EditObservation dropped Notes fields

JoeCohen:
It was dropping contents of Notes Template fields where the template entry was multiple words, e.g. "Nearest tree".

Problem:
* the view filled those fields using Observation#notes_part_value(part), where part was "Nearest tree",
* notes_part_value got the value of the key `:Nearest tree`, but
* the actual key is `:Nearest_tree`, i.e, with an underscore instead of a space.

Fix is to translate spaces to underscores when looking for the value of the key.  Also added a simple test to prevent reversion.

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -375,9 +375,17 @@ class Observation < AbstractModel
 
   # value of notes part
   #   notes: { Other: abc }
-  #   obervation.notes_part_value("Other") #=> "abc"
+  #   observation.notes_part_value("Other") #=> "abc"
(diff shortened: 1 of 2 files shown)


pellaea:
Tricky.  Not sure why that fixes it, but good job figuring it out.  Glad it
is so simple, too.

JoeCohen:
Pushed to production. 
I'll do a News Article tomorrow late afternoon.  (I didn't want to publicize this until the bug was fixed. I'm out of time today, and will be participating in a Lichen-Moss-Mushroom bioblitz tomorrow until 2 or 3pm.)

PR #347, opened 2017-11-11
Fleshed out and tested sequence API.

pellaea:
It was already written, as it turns out.  Sequences were already being returned in high-detail responses to observation queries.  And there was already a sequences entry point.  I just completed the set of parameters sequence queries can take, including a fair subset of the standard parameters one can apply to any observation query, then added support for those new parameters in the sequences API.  And, of course, added tests for all the above, since there were essentially no tests either for sequence queries or sequence API.


app/classes/query/modules/initialization.rb
```diff
@@ -46,10 +46,9 @@ def initialize_model_do_range(arg, col, args = {})
       min, max = params[arg]
       @where << "#{col} >= #{min}" unless min.blank?
       @where << "#{col} <= #{max}" unless max.blank?
-      if (val = args[:join]) &&
-         (!min.blank? || !max.blank?)
-        # TODO: is this used? convert to piecewise add_join
-        @join << val
+      if val = args[:join]
```
JoeCohen:
```
if (val = args[:join])
```
Parenthesizing the condition makes Rubocop happy.

app/classes/query/modules/initialization.rb
```diff
@@ -82,12 +81,16 @@ def initialize_model_do_misspellings
...
   end
 
-  def initialize_model_do_objects_by_id(arg, col = nil)
+  def initialize_model_do_objects_by_id(arg, col = nil, args = {})
     if ids = params[arg]
       col ||= "#{arg.to_s.sub(/s$/, "")}_id"
       col = "#{model.table_name}.#{col}" unless col.to_s.match(/\./)
       set = clean_id_set(ids)
       @where << "#{col} IN (#{set})"
+      if val = args[:join]
```
JoeCohen: `if (val = args[:join])`

app/classes/query/modules/initialization.rb
```diff
@@ -287,6 +291,10 @@ def initialize_model_do_date(arg = :date, col = arg)
         initialize_model_do_date_half(true, vals[0], col)
         initialize_model_do_date_half(false, vals[1], col)
       end
+      if val = args[:join]
```
JoeCohen:
```ruby
(if val = args[:join])
```

pellaea:
Having come to this fresh, I understand your confusion.  Ironically (to me)
it's precisely because API and Query are so closely linked... yet not
precisely the same.  That is, the parameters passed into API are passed
through *almost* but not quite verbatim to Query.  It's the differences
which were confusing to me.  The design decision reasoning was simple: I
didn't want it to be verbatim, because any internal change to Query would
then necessarily affect the external API and all the API users.  I want to
feel free to change implementation details internally without risk of
affecting external users.  But the result is admittedly confusing!  Perfect
example was how dates are handled.  Some places accept a date range, in
some places "2016" implicitly means "2016-01-01 through 2016-12-31", in
others it just means anything after 2016-01-01.  And the syntax varies.
And parameters like "children_names" are especially opaque to me now.  Once
I remembered how it's implemented I thought it was obvious, but coming from
outside it's completely obscure!  Sorry!  So much for the much-vaunted
self-documenting code theory.  A terrible idea if I've ever heard one!

PR #349, opened 2017-11-18
Fully tested API GET, POST, PATCH and DELETE reqeusts for almost all objects.

pellaea:
Pretty straight forward.  Required some significant changes to Query as well, of course, because API and Query are so closely related.

Biggest unexpected change was a tweak to User.current_location_format.  Since most apps will not have access to the users' preferences, they would have no way of knowing which location format to use on behalf of users.  Therefore I've had the API override user preferences by setting User.current_location_format explicitly right after authenticating the user and setting User.current.  The result is that the location format is universally and consistently "postal" for the API.  I check the last year of logs and everyone POSTing to the API over the last year has chosen to use postal format.  So this shouldn't break anyone's apps.

As for the rest, it was mostly just writing a s---t-ton of tests in test/models/api_test.rb.  And fixing the numerous bugs uncovered thereby.

Oh right, and I refactored parsing in API completely.  I added extensive documentation to API::Base to help new developers understand the overall design and how everything fits together, as well as explaining how new subclasses should be written to implement API for additional models in the future.


app/classes/api/base.rb
```diff
@@ -42,60 +254,55 @@ def self.instantiate_subclass(params)
...
-    elsif !key.user.verified
-      fail UserNotVerified.new(key.user)
+      User.current_location_format = :postal
     else
+      key = ApiKey.find_by_key(key_str)
+      raise BadApiKey.new(key_str)        unless key
+      raise ApiKeyNotVerified.new(key)    unless key.verified
+      raise UserNotVerified.new(key.user) unless key.user.verified
       User.current = self.user = key.user
+      User.current_location_format = :postal
```
JoeCohen: Duplicates l.279. But maybe that's OK for clarity?

pellaea: I went along with rubocop for a long time on this sort of thing, but I'm increasingly firm in my disagreement.  DRY is a great principle, but like any other principle, needs to be applied in the real world.  There are cases where a tiny bit of innocuous repetition helps readability.  In this case, I think it is clearer this way because I prefer to keep the two lines User.current = xxx and User.current_location_format = xxx together.  Otherwise every time I see this code I will have to verify that there is no real reason to have them separated, that the intervening code between the lines doesn't somehow take advantage of some obscure momentary difference.  Yes, I could create yet another class method in User which sets the two simultaneously.  If that works better for you, I could swallow that.  But another thing rubocop strongly encourages which drives me out of my tree is taking single simple lines of code and moving them into single-line methods in distant parts of the code.  It just makes me have to track down dozens of silly one-line methods every time I want to figure out why a method is failing.  It's a nightmare!  (But rubcop likes it because it reduces the ABC ratio or length of whatever that cursed metric is called.)  That's why I'd prefer to keep this in-line and repeated.  I believe API still passes rubocop 100% (except where I've judiciously disabled a few metrics here and there, ha!)  Although granted I haven't run rubocop in a long time...

JoeCohen:
1. Understood. And I've also been bitten by having to track down stuff in
distant parts of the code, or worse, in different files.  And if RuboCop is
being too big a pain, let's adjustment the metrics.  We can also mark
individual offenses as "Wontfix".

2. I'm not going to be any substantial help on your latest PRs. It's fine
with me if you merge them into Master.  (I'd actually prefer that, as I can
work on other code without fear of conflict.)

pellaea:
Thanks, Joe.  I keep forgetting about the "wontfix" thing.  I have been
making use of the temporary disable pragma comments, though.

Short term, since I've gotten cold feet about releasing all this work and
am taking days to fully test it and re-test it, I recommend making a branch
off of the PR.  I think that will work perfectly.  The PR is working great
and passes all the test and all that.  So it's a good branch to work off
of.  I'm just finding lots of little bugs here and there which our tests
are not catching.  I won't be making any more sweeping changes.

Does that make sense?  I'm sorry if you've been waiting a long time.

PR #350, opened 2017-11-20
Collection numbers

pellaea:
[NOTE: this builds on top of the API PR]

Clearing up some of the confusion about what a "specimen" really is in MO.  Renames the old "specimen" to "herbarium record".  Creates a new entity called "collection number", comprising a collector and a collection number.  There is no object encapsulating a real physcial collection ("specimen").  A specimen in this sense can have multiple collection numbers (this is not unusual actually, especially if an amateur sends a specimen in to an expert, and the expert is forced to give the specimen their own collection number in order to keep track of it, then later they discover the original collector does in fact have a collection number for it, too).  A specimen ca certainly have multiple herbarium records, usually one record in each herbarium it has passed through, but also possibly multiple records at the same institution (if it has been split for example).  Note that neither herbarium_record nor collection_number record the actual physical location(s) of the specimen.  And lastly, a specimen can belong to multiple observations, and multiple specimens can belong to a single observation.  So it's a grand many-to-many relationship all around.  At the moment there is no use-case for actually needing a true "specimen" entity on MO, so I'm leaving it out.

Note that the typical case will, however, continue to be quite simple: one observation corresponds to one collection number and one herbarium record.  And typically the herbarium record will belong to the user's personal herbarium with the specimen id being their collection number.  And this will be easy to set up in create_observation.  All the same fields basically -- some called slightly different things -- specimen checkbox, herbarium, collector and collector number.  (Collector is the only one that's new.)  This will create a collector number, and it will create an herbarium_record with the initial determination coming from the observer's choice of name, the collector (defaulting to the observer) and the collector number.

Presently, all I have done is rename "specimen" to "herbarium_record".  I wanted that all in one monster commit after the migration and tests all passed.  Tomorrow I'll work on creating CollectorNumber and tweaking the observation forms and display pages... and test and translation strings and API and ...  Ugh.


pellaea:
Okay, this branch is more or less finished.  There are still some unit tests to write (like the entire collection_number controller!), and I think there is still work to do in the API.  But if anyone wants to test the work flow of adding, changing, removing collection numbers, herbarium records, sequences, and creating, editing herbaria, now is the time.  I will be out for a week maybe.  Will finish off the unit tests when I get back if not before.  This PR solves a *lot* of bugs, and it's important for the Mycoflora initiative, so I want it to go out sooner rather than later.  I was pushing for end of November, since that seemed to be a deadline for Mycoflora, too.  But something has come up requiring me to push it back another week.

PS. I have not been running rubocop, so I expect I've inadvertently introduced a lot of petty new complaints throughout the code.  Tough.  I grow very weary of rubocop.  Keeping methods short and simple, enforcing uniform style -- okay! I'm on board.  But 90% of my time is spent fixing stupid things that don't help the code in the slightest.  Sometimes even hurts it.  In my opinion, rubocop is a textbook case of how it's impossible to come up with a single set of hard and fast rules that apply to every case.  It gets an A for good intentions, though.  If I had way more time I would continue to fight with it.  But I don't.  No bitterness though... ;)

PR #368, opened 2017-12-24
Log specimen changes better

pellaea:


pellaea:
I'll merge this after merging the "specimens anyone" branch, just in case there are any conflicts.

PR #370, opened 2017-12-25
Blast links

JoeCohen:
Adds links from Observation and from Sequence to an NCBI page to create a BLAST report for the Sequence.

### DIFF
--- a/app/models/sequence.rb
+++ b/app/models/sequence.rb
@@ -6,21 +6,25 @@
 #
 #  == Attributes
 #
-#  id::               unique numerical id, starting at 1.
-#  observation::      associated Observation record
-#  user::             user who created the Sequence
-#  locus::            description of the locus (region) of the Sequence
-#  bases::            nucleotides in FASTA format (description lines optional)
-#  archive::          on-line database in which Sequence is archived
-#  accession::        accession # in the Archive
-#  notes::            free-form notes
+#  id::                unique numerical id, starting at 1.
+#  observation::       associated Observation record
+#  user::              user who created the Sequence
+#  locus::             description of the locus (region) of the Sequence
+#  bases::             nucleotides in FASTA format (description lines optional)
+#  archive::           on-line database in which Sequence is archived
+#  accession::         accession # in the Archive
+#  notes::             free-form notes
 #
 #  == Class Methods
 #
-#  locus_width        Default # of chars (including diaresis) to truncate locus
+#  blast_url_prefix    part of url prepended to BLAST QUERY
+#  locus_width         Default # of chars (including diaresis) to truncate locus
 #
 #  == Instance Methods
 #
+#  accession_url       url of a search for accession
+#  blast_url           url of NCBI page to create a BLAST report
+#  blastable?          Can we easily create a blast_url for the Sequence?
 #  deposit?            Does sequence have a deposit (both Archive && Accession)
 #  format_name         name for orphaned objects
 #  locus_width         Default # of chars (including diaresis) to truncate locus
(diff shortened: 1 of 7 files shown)


app/models/sequence.rb
```diff
@@ -82,6 +86,39 @@ def locus_width
...
+  def blast_url
+    if blastable_by_accession?
+      "#{blast_url_prefix}#{accession}"
+    else
+      # wrap QUERY in quotes because it can contain whitespace
+      %(#{blast_url_prefix}"#{bases}")
+    end
+  end
+
+  def self.blast_url_prefix
```
JoeCohen: Is this better implemented as a constant? 
If so, where to put it?

pellaea:
My approach to these sorts of things is -- If it's not obvious how to
parameterize something, then don't.  Wait for additional concrete use cases
to come up, then it will typically become obvious.  If you have to guess
now, chances are you will guess wrong and just have to redo it later anyway!

app/views/sequence/show_sequence.html.erb
```diff
@@ -36,18 +36,26 @@
...
+        <% if @sequence.deposit? %>
+          <tr>
+            <td class="bold"><%= :DEPOSIT.t %>:</td>
+            <td>
+                <%= sequence_archive_link(@sequence) %>:
+                <%= sequence_accession_link(@sequence) %>
+            </td>
+          </tr>
+        <% end %>
+        <tr align="center">
```
JoeCohen: I'm not in love with the way this looks in show_sequence. 
Any suggestions?
How about a button like this: 
blastbuttondown
And how do I implement a button outside of a form without using js?

pellaea:
I haven't seen it on a browser yet, so can't comment on its current
appearance.  But re: button outside of form -- That's just a CSS thing.  I
think you just want to use class="btn btn-xs" on the anchor tag.  Or
something like that.  It's still technically just a link, but it looks like
a button.

JoeCohen:
1. I intended to ask only about blast_url_prefix (a sort of
functional-style constant). E.g., should it be implemented as a constant
(or some way other than what I did).
2. I did add a blastable_by_accession param to the archives. It's used in
Sequence#blastable_by_accession?. Should I implement that method
differently, e.g.,
  def blastable_by_accession?
    archive.present? && archive == "GenBank" || archive == "ENA"
  end

pellaea:
Sorry, I'm trying to take care of all of Pulk's requests, and haven't had a
chance to look at your PR in detail.  Ignore my advice if it didn't make
sense in the context.  I'll look at it again later when I have time to
figure out what you mean.

pellaea:
Sequence#blast_url -- I think you will need to parse "bases" and probably should escape it for url, as well, just to be safe.  At least remove white space and comments(?) even if no other additional parsing is required.

On that subject, I've been wondering of we shouldn't be removing comments and whitespace from Sequence#bases anyway.  Because of the possible whitespace we can't actually allow searches on that column.  I think it would make sense to remove whitespace for storage in database, then when displaying the bases introduce formatting.  Maybe even allow user to choose between various formats, e.g., for downloading a bunch of sequences from a sequence index (although that's getting a bit beyond the scope of MO's mission).  Or alternatively, store bases twice, once verbatim, and once without any extra crap so that it's searchable.  Am I making any sense?

pellaea:
re: Sequence.blast_url_prefix -- Yeah, it doesn't feel right there to me, either, but I also don't know where it belongs.  My advice stands: wait until we have more of these sorts of external links.  Eventually it will be obvious how to generalize the task and pull them out of models, etc.

(Personally, I would have thought of this as a view thing, not a model thing, but such a minor preference.)

JoeCohen:
Re buttons. In Rails it's `button_to`.

pellaea:
Interesting.  I remember that from a long time ago.  But also note that
there's not a single occurrence of that helper anywhere in our views now.
Might not play well with bootstrap.  Definitely worth checking in a couple
themes just to be sure it looks the same as other buttons.

JoeCohen:
Re parsing bases in order to form URL. 
I had been thinking about this. But it turns out to be unnecessary **if bases are formatted correctly in the 1st place**; all I have to do is quote the QUERY param of the url.
Unfortunately, it looks like I'm messing up FASTA formatting when it includes numbers. I also need to accomodate at least 69 characters width in for Bases in the show_sequence page.
I'll add a Story for this.

JoeCohen:
Re making Bases column searchable.  
Hmm. Hadn't thought of that. Do we really want to do that? Feature creep? Should we instead encourage folks to  put stuff into GenBank or Steve Russell's MycoMap and search there?

pellaea:
That was my take, too.

JoeCohen:
More re parsing Bases: You are right.  
If I simply wrap the Bases in quotes and use that as NCBI's QUERY param, then NCBI can mess up some of the statistics. (NCBI will BLAST the QUERY, and will correctly report the identity statistic (Ident). But can mess up other statistics.) 
So the best approach is to use only the nucleotide codes as the QUERY param. That should be very easy.
Thanks again!

pellaea:
I'm glad you're on top of this!  My experience with NCBI is still very
limited.

JoeCohen:
More problems with Git. So to keep the history clean I will:
- close #370 
- delete this branch
- create a new branch with the same name and push some changes to it.

PR #371, opened 2017-12-27
Improve Sequence functionality & appearance

JoeCohen:
Continues #370 
Delivers https://www.pivotaltracker.com/story/show/153852725 and https://www.pivotaltracker.com/story/show/153903849

- Adds links to NCBI BLAST from Observation and from Sequence.
- Addresses issues raise in #370: uses raw base codes as BLAST's QUERY param. (We need not escape that param because the raw codes are a limited set of letters.)
- Improves appearance of Sequence and links.


app/models/sequence.rb
```diff
@@ -179,7 +221,7 @@ def other_sequences_same_obs
   # Validate proper formatting of bases
   # See BLAST documentation (shortened url: https://goo.gl/NYbptK)
   # full url in WebSequenceArchive::blast_format_help
-  def blastable
```
JoeCohen: Method name changed to avoid confusion/collision with new `blastable?` method.

PR #374, opened 2017-12-29
Set db default collation & charset

JoeCohen:
Hopefully this will fix Travis errors which occur when loading fixtures, like
```ruby
ActiveRecord::StatementInvalid: Mysql2::Error: Incorrect string value: '\xCE\x95\xCE\xBB\xCE\xBB...' for column 'name' at row 1: INSERT INTO `languages` (`locale`, `name`, `order`, `official`, `beta`, `id`) VALUES ('el', 'Ελληνικά', 'Ellenika', 0, 1, 552606384)
```


JoeCohen:
The change worked as intended: `rake db:fixtures:load` no longer errors, and the test suite runs.
But now that there are test errors/failures which had been previously been hidden.  And all of them appear to be Travis-only errors. (They don't occur on my VM.)

JoeCohen:
Looks to me like they should be failing locally. For instance, when run on Travis, the test suite reports these missing tags:
add_herbarium_record
show_observation_create_herbarium_record
edit_herbarium_save
And indeed they don't exist in en.txt.  But they do exist **locally** in en.yml (but of course not in Travis).
I vaguely recall seeing some commit about deleting the .txt files. Could these errors be related?

JoeCohen:
Tags in local en.yml, but missing from en.txt.  (I don't want to add these back until we figure out what caused the problem in the first place.)
goes after edit_herbarium_title
`  edit_herbarium_save: Save`
  
goes after log_herbarium_record_added
`  show_observation_create_herbarium_record: Add Herbarium Specimen`

goes after edit_herbarium_duplicate_label
`  add_herbarium_record: Add Specimen`

pellaea:
No, the non-English text files were only designed to make it easy to
translate the site for a new language.  They aren't used by anyone.  I am
concerned about the discrepancy between en.txt and en.yml, though.  I would
like to look into that.

pellaea:
edit_herbarium_save is only used in test, should just be :SAVE instead.
The others are used in the code.  (They should no longer use the word
"Specimen" though, as that confuses some people like Adolf.)  I thought I
removed all three of these.  And maybe that's what I should do.  But still,
not sure why they would persist in en.yml after being deleted from
en.txt...  Oh wait, I think I know.  I think the language rake task used to
put any deleted tags at the end of en.yml.  I think that was deliberate to
avoid accidentally deleting important things.  But it results in the unit
test not catching missing tags like it should.  We can always go back in
git to restore accidentally-deleted tags.  I think I should stop the rake
task from doing that.

JoeCohen:
Thanks for fixing the tags. Much improved!
Hopefully just one type of test error/failure remaining (occurs 3x). 
Travis-only.  It cannot find el.txt
```ruby
Errno::ENOENT:         Errno::ENOENT: No such file or directory @ rb_sysopen - /home/travis/build/MushroomObserver/mushroom-observer/config/locales/el.txt
```
Details here: https://travis-ci.org/MushroomObserver/mushroom-observer/builds/323025049

JoeCohen:
Do the tests simply need to create the file if it doesn't exist?  
And should we delete the non-English .txt files from our local machines?

pellaea:
I wouldn't worry about it, personally.

PR #379, opened 2017-12-30
Fix deprecated "What" loop [Fixed #107455054]

JoeCohen:
When **What** is deprecated (in create Observation), and user clicks Create,
form is reloaded without any message. Click Create again, and it keeps happening.

Problem is a typo in _form.html.erb.

- Fixed typo which caused bug
- Fixed typo in comment
- Minor reformatting for readability, e.g., align arguments, shorten long lines, etc.


pellaea:
Good catch, thanks again.

JoeCohen:
No, thank you. (As you would say.) You wrote about a zillion lines of code, and even I can understand them. My recent changes are a drop in the bucket.

PR #385, opened 2018-01-07
Gems 2018 01

JoeCohen:
Updates gems.  51 are out-of-date, and we have one security warning.

Delivers https://www.pivotaltracker.com/story/show/154110059.


JoeCohen:
Mail renders newline differently into html.
Before update:  `\r\n`
After: `\n`
This should not change how the html is rendered.
**Is this a problem?**

PR #390, opened 2018-02-05
Cleanup hitless search display

JoeCohen:
Clean up the views of search results, especially when there are no hits, or only one hit. See generally [Pivotal #154862044](#154862044) and [part of discussion, MO Developers > Synonymy etc.](#154862044)
- Prevent circular link in results of hitless Observation search;
- Improve wording of display for hitless Observation search;
- Don't show right tabset for hitless Observation search;
- On all search result pages, don't display title.
- On all search result pages, don't show sort bar unless there are at least 2 hits.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1379,7 +1379,11 @@ class ApplicationController < ActionController::Base
     @sorts = (num_results > 1 ? sorting_links(query, args) : nil)
 
     # Supply a default title.
-    @title ||= query.title
+    # If no results, then title is empty but not nil
+    # An empty title will cause the view to display nothing,
+    # overriding any title specified in the view,
+    # while keeping the application default (action name) as <title> metadata
+    num_results.zero? ? @title = "" : @title ||= query.title
 
     # Supply default error message to display if no results found.
     if (query.params.keys - query.required_parameters - [:by]).empty?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1373,18 +1373,6 @@ class ApplicationController < ActionController::Base
     # Pass this query on when clicking on results.
     query_params_set(query)
 
-    num_results = query.num_results
-
-    # Add magic links for sorting if enough results to sort
-    @sorts = (num_results > 1 ? sorting_links(query, args) : nil)
-
-    # Supply a default title.
-    # If no results, then title is empty but not nil
-    # An empty title will cause the view to display nothing,
-    # overriding any title specified in the view,
-    # while keeping the application default (action name) as <title> metadata
-    num_results.zero? ? @title = "" : @title ||= query.title
-
     # Supply default error message to display if no results found.
     if (query.params.keys - query.required_parameters - [:by]).empty?
       @error ||=
@@ -1460,9 +1448,19 @@ class ApplicationController < ActionController::Base
     @num_results = query.num_results
     @timer_end = Time.current
 
+    # Supply a default title.
+    # If no results, then title is empty but not nil.
+    # Result: No title is displayed
+    # (overriding any title specified in the view)
+    # and the html <title> metadata == a translated tag or the action name
+    # see ApplicationHelper#title_tag_contents
+    @num_results.zero? ? @title = "" : @title ||= query.title
+
+    # Add magic links for sorting if enough results to sort
+    @sorts = (@num_results > 1 ? sorting_links(query, args) : nil)
+
     # If only one result (before pagination), redirect to 'show' action.
-    if (query.num_results == 1) &&
-       !args[:always_index]
+    if (@num_results == 1) && !args[:always_index]
       redirect_with_query(controller: query.model.show_controller,
                           action: query.model.show_action,
                           id: query.result_ids.first)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1373,6 +1373,11 @@ class ApplicationController < ActionController::Base
     # Pass this query on when clicking on results.
     query_params_set(query)
 
+    num_results = query.num_results
+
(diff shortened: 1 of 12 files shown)


JoeCohen:
My implementation of no title for hitless searches involves changing the html metadata title for those pages. It's "Mushroom Observer: #{controller.action_name}", e.g.:
Mushroom Observer: observation_search
instead of 
Mushroom Observer: Observations Matching ‘asdfghjqwerty’

Is this a problem?

pellaea:
I think I see the problem -- it's getting the title from the Query, so it's
not easy to change, is it?  Still, maybe we can use query.model to turn
this new title into a translatable string?  (Damn, I don't even have the
code on my laptop any more!  Hard to look stuff up...  one second... okay,
there's what I'm looking for...)  How about something like this

if no_matches?
  @title = :pattern_search_no_results_title.t(type:
@query.model.to_s.underscore.to_sym
}
end

With the new tag defined something like this:

  pattern_search_no_results_title: "No Matching [TYPES]"

Does that work for you?

JoeCohen:
That would work as a title, but there's already an almost identical flash message:
captura de pantalla 2018-02-05 a la s 15 49 14
In my (not too strong) opinion, 
- I like the flash message, 
- I like getting rid of the title, and 
- if we're going to have a title, I think the old one is okay (Observations Matching ‘_pattern_’)

JoeCohen:
PS:
And I have only the vaguest idea of the implications of the <title> metadata.  My very uneducated notion is that it won't matter for these kinds of pages. 
Will the robots/search engines be angry with us? Can they even see these pages?

pellaea:
Robots should not see these pages.  Searches are strictly verbotten for
robots.

The <title> meta tags affects the title bar of the browser window.  It'll
be shown in the task bar, bookmarks, etc., too.  It should be reasonably
correct.  Aren't both the title on the page itself and the <title> tag set
from @title?

Agreed that "No Matching [TYPES]" is redundant.  How about remove the flash
error instead?  Or make the title just "[TYPE] Search" and leave the flash
error.  (The title still needs to show the search string if there are
actually hits, because otherwise it doesn't necessarily show anywhere on
the page!  But if you want to add that information elsewhere, e.g. as a
subtitle, that would work for me, too.  Then the title can remain the same
regardless of whether there are any hits.)

I just find:

No matching observations found.
Observations matching "xxx"
Did you mean "yyy"?

to be confusing.  But maybe it's just me, in which case feel free to ignore
me.

-Jason

JoeCohen:
Thanks so much for the metadata explanation.

Yes, both the displayed title on the page itself and the <title> tag are set from @title, 
BUT if @title is blank, then the title tag is "MO: _controller.action_name_". 
(That's how we can get a <title> tag even though @title is blank and no title is displayed on the results page.)

So I think this will work: if @title is blank, then:
1. The <title> tag is "Mushroom Observer: `controller.action_name.gsub("_", " ").titleize_`", 
e.g. "Mushroom Observer: Observation Search";
2. No title is displayed on the search results page. 

Example:
captura de pantalla 2018-02-05 a la s 17 17 14

(If there are hits, then the page title and the <title> tag are the same as before, and include the pattern.  I haven't changed that.)

pellaea:
Very close, but with apologies, one more detail(!)

  controller.action_name.gsub("_", " ").titleize

Unfortunately that doesn't translate.  If we must provide an automatic
default for it, can we run it through i18n somehow?  Maybe something like:

  :"title_for_#{controller.name}_#{controller.action}".t

Or whatever, just allow the string to be translated somehow.

Thanks!

JoeCohen:
Good catch! 
I'll try to figure it out.

JoeCohen:
Is there an easy OR official way to test whether a translation label exists (in en.txt)?

pellaea:
Sort of, indirectly.  All tests that use translation tags implicitly have
tests to assure that any tags they use actually exist.  So the way to do it
in effect is just to run

  :tag.l

on every one.  If you really want you can do

  assert_not_blank(:tag.l)

to make it more obvious that they are tests, but even that is unnecessary.
Feel free to try it -- add a bogus

  :xxx.l

to any test and run it and make sure it complains about :xxx not existing.

JoeCohen:
Thanks.  But I asked the wrong question. (Sorry.)
What I really want to know is whether there's a way for **the app** to tell if a label exists in en.txt.

pellaea:
(Haha, and assert_not_blank(:xxx.l) would actually pass anyway! even if
:xxx doesn't exist... But it makes it *look* like it's a test at least. :)

pellaea:
Okay.  Yes, just TranslationString.where(tag: "xxx").empty?  But forgive
me, I wonder if there might be a better way to do what you're doing?

JoeCohen:
You're forgiven. :-)
I want to set the <title> tag with something like the following method (pseudocode)
```ruby
  def title_tag_contents(action_name)
    if @title.present?
      @title.strip_html.html_safe
    elsif tag_exists_for(:"title_for_#{action_name}")
      :"title_for_#{action_name}".t
    else
      action_name.gsub("_", " ").titleize
    end
  end
```
AND I want to be able to test this method.

pellaea:
Okay, I see.  Well, if the unit tests work right, the third clause of your
if statement should never invoke, but...

Without that clause, :title_for_pattern_search will yield the somewhat
unsightly string "[:title_for_pattern_search]".  I agree that it would be
better for it to display "pattern search" instead.

Please ignore my nosiness. :)

JoeCohen:
I appreciate the nosiness.
Still not sure how the app itself can tell whether label exists in en.txt 
(other than seeing if `:foo.l` returns `"[:foo]"`. Ugh).

pellaea:
Sorry, must have lost my response to that question.  Do this

TranslationString.where(tag: "foo").empty?

JoeCohen:
Thanks!

PR #391, opened 2018-02-07
Added collectors' numbers and yyyy-mm-dd date to Mycoflora report.

pellaea:
I'm going to let Travis do tests for me since I can't actually run dev VM yet.


JoeCohen:
Shall I finish this branch? It might be most efficient given the issues with your development VM.

pellaea:
Sure, if you like.  It should be done, in fact.  I just need to test it
live.  But I'm going to be out until next week, so I was holding off.

pellaea:
Oh I see, must've broken the report test... yeah, just need to adjust the
tests to account for the new format.  Sorry!

JoeCohen:
That's what I thought.  But there's another issue. 
ObservationReportTest#test_mycoflora_no_exact_lat_long
expects fieldNumber "314159", but instead it's returning "".
How could changing the order of the report fields cause `fieldNumber` to be calculated incorrectly? I probably won't have time to figure that out.

JoeCohen:
I see the problem. 
The name of the project was changed, but the name of the herbarium was not. 
I'll fix it.

JoeCohen:
Grr. [The test which fails in Travis](https://travis-ci.org/MushroomObserver/mushroom-observer/builds/339253577?utm_source=github_status&utm_medium=notification) passes locally.

JoeCohen:
I restarted the Travis job, and it passed. 
Question: My understanding is that mySQLis not guaranteed to return rows in any particular order. 
Therefore, should the mySQL search in `ObservationReport::CSV::Mycoflora::add_mycoflora_ids!` also include a sort?

app/classes/observation_report/mycoflora.rb
```diff
@@ -1,12 +1,16 @@
 module ObservationReport
   # Format for export to Mycoflora.
   class Mycoflora < ObservationReport::CSV
+    MYCOFLORA_PROJECT_NAME = "North American Mycoflora Project".freeze
```
JoeCohen: Is there a better place to define this constant so that it can easily be used outside this file, e.g., tests, ObserverController::make_labels, etc.?

app/classes/observation_report/mycoflora.rb
```diff
@@ -107,5 +114,19 @@ def add_mycoflora_ids!(rows, col)
       )
       add_column!(rows, vals, col)
     end
+
+    def add_collector_ids!(rows, col)
+      vals = CollectionNumber.connection.select_rows %(
```
JoeCohen: Should a sort be added to this SQL statement? See  https://github.com/MushroomObserver/mushroom-observer/pull/391#issuecomment-364300163 and https://github.com/MushroomObserver/mushroom-observer/pull/391#issuecomment-364312470.

pellaea: Sure, if a sort is needed to derandomize the final result, I'm all in favor of it.  I hate functions that leave order random.  (The alternative is to sort the results before the assertion in the unit test.  Whatever works for you.)

JoeCohen: Hard (maybe impossible?) to tell if it's really needed. This is the only time the test failed.

JoeCohen: But I like the idea of sorting it in the test. Epsilon less of a performance hit in production.

pellaea:
I'm back, so I should take over this PR again, anyway.  I can look into
it.  I seem to remember considering the case of multiple collection numbers
for a given observation, and that it was undefined which order they would
appear in that report column.  I was just in a hurry and/or lazy and
ignored it.

PS. re: where to define the name of the magic "North America Mycoflora
Project" title -- It probably should be done differently... but all the
NAMP-specific code is presently in this file (I think), so I say let's
leave it unless you have some obviously superior solution.  I do realize
that the report will break if anyone changes the herbarium's name.  (I at
least added Steve and Bill as curators, so no one else should be allowed to
change the name now.)  Yes, ugly.  But this will not last.  And we'll
probably never need anything like this again, anyway.  Or if we do, further
data points will help us generalize it more intelligently.  That's my two
cents, at least.

JoeCohen:
Great! I was just trying to save you a few minutes by fixing the broken tests, etc.

pellaea:
And thank you for that!  I wish I hadn't botched it so badly. :)

PR #392, opened 2018-02-08
Codeclimate update

JoeCohen:
Update configuration

The new (Nov 2017) CLI broke our old configuration.  See [Pivotal #153955907](https://www.pivotaltracker.com/story/show/153955907)


JoeCohen:
I will merge this; there are finally 0 issues in this branch.
This results largely from disabling (a) CodeClimate's new "Maintainability Checks" and (b) the js linting plugin (ESLint).
If we have big problems in other areas (e.g., scss style checking) we probably should relax rules, or disable rules or plugins until we're ready to address the alleged violations. _Cf._ http://rubocop.readthedocs.io/en/latest/configuration/#automatically-generated-configuration (suggesting that if there are an "overwhelming amount of offenses" then some cops should be disabled until the offenses are worked through).

PR #394, opened 2018-02-14
Ruby 2 4

JoeCohen:
- Upgrade Ruby from 2.3.3 to 2.4 
- Update gems
Delivers [Pivotal #118390619](https://www.pivotaltracker.com/story/show/118390619), which has been around for almost 2 years.
Erases some technical debt; should give a slight speed bump; and has fewer security issues.

I wouldn't have done this on my own, but @mperham was kind enough to walk me through the critical part (upgrading Ruby, making the test pass, get rid of the deprecation warnings). 
I later did a general `bundle update`, then one-by-one, removed limits on various gem versions. I also tweaked the Travis and RuboCop configurations.

I think that pushing it to production should be easy: I think the only special step is `rvm install 2.4` (followed by the usual stuff). Newly created dev VMs should take care of themselves.


JoeCohen:
[Coveralls job](https://coveralls.io/builds/15515501) says that I've lost some lines of coverage. I'll check that out to (a) make sure it doesn't indicate a bug, and (b) try to restore the lost coverage.

CodeClimate blew up because of some problem within RuboCop.  I'll check it out.

JoeCohen:
Re Coveralls: the lines were not lost in this branch; they are not covered in master. (I switched to master, put breakpoints in the 2 lines reported as newly uncovered, and ran the tests.  Neither breakpoint was hit.)

pellaea:
Weird.  Bug in coveralls?

JoeCohen:
Yes, it seems that either Coveralls is buggy, or I don't understand its report.
For instance, for  /app/classes/query/modules/validation.rb, line 185
the [current report](https://coveralls.io/builds/15515501/source?filename=app%2Fclasses%2Fquery%2Fmodules%2Fvalidation.rb#L185) says the line is newly uncovered
captura de pantalla 2018-02-13 a la s 20 39 05
but the [prior report for master](https://coveralls.io/builds/15478741/source?filename=app%2Fclasses%2Fquery%2Fmodules%2Fvalidation.rb#L185) shows it was previously uncovered.
captura de pantalla 2018-02-13 a la s 20 41 38

pellaea:
Too bad.  But for the most part it is good.

JoeCohen:
If I get some time, I'll ask Coveralls.

JoeCohen:
CodeClimate is not completing its report. It says that this is due to excessive Rubocop issues.
I've run Rubocop locally (on my dev VM) and there are actually fewer problems in this branch than in master.
I will push some experimental changes to see if I can isolate the problem, but will probably need to get help from Codeclimate.

pellaea:
Oh I see.  Must be some configuration problem.

JoeCohen:
I hope so, but am not sure. If I run Rubocop locally (on my dev VM):
ruby-2-4 - 7265 offenses detected
master - 6288 offenses detected
Maybe the additional 1023 offenses are enough to push Codeclimate over the edge - which is where you wanted it. :-)

pellaea:
Haha!

JoeCohen:
The 1st experiment didn't work, so I reverted it. I will try something else later today or tomorrow.

JoeCohen:
A. Coveralls is flaking out again. Travis says it sent coverage data to Coveralls and started a [Coveralls job](https://coveralls.io/jobs/33764521) but the job just hangs.
B. Codeclimate is at least giving me a report. I think I can deal with 304 offenses (and maybe about 1000 more which I turned off artificially. But I'm worried about creating conflicts with your tweak_mycoflora_report branch.  Two options:
1. Merge tweak_mycoflora_report into master, merge master into ruby-2-4, then fix rub-2-4.
2. Fix ruby-2-4, merge it into master, merge master into tweak_mycoflora_report, and fix any conflicts.
I can do either of these.
**Which would you prefer?**  Any other suggestions?

pellaea:
I'm happy to merge master into my PR branch.  It can't be that bad.  We did
so little on that branch! :)

JoeCohen:
How about merging your branch into master (rather than the other way) (my suggestion 1 above)
My fixing 304-1000 issues is going to touch a lot files. If the fixes get merged into master, and master gets merged into your branch, there may be conflicts. Might be easier if I fixed them, especially since your dev VM is dead. (My suggestion 2 above.)

JoeCohen:
And one more potential issue:
Will you have problems installing Ruby 2.4.1 on your local machine?

pellaea:
Okay, sure.  What state is my PR even in, though?  Sorry, I haven't taken
even a second to look at it since our anniversary weekend out in the mtns.
If it's safe to merge, then sure.  I seem to remember only changing a
single file on that branch.  But I'm fine with whatever.

JoeCohen:
Headed off to bed. I'll look for your response tomorrow.

pellaea:
I'm not sure I have to.  Don't I only need to install it on the VM (once I
finally get virtualbox working again)?

pellaea:
Thanks for all the work you've been doing lately.  Sorry I haven't even
taken the time to see what you've been doing. :(

JoeCohen:
I think that once ruby-2-4 gets merged into master, virtual box will automatically install it.

JoeCohen:
But right now you aren't working on virtual box.

pellaea:
Right.  And right now I'm not even running the code at all!  So I don't
think it even matters if I have ruby installed on my laptop.

JoeCohen:
OK. I'll try to get rid of these 304 issues.

JoeCohen:
TLDR: This will take a long time.
Long story:  
- RuboCop has new cops which take advantage of new features in Ruby 2.4. For instance, Ruby 2.4 has `match?`, and Rubocop has `Performance/RegexpMatch`:
>In Ruby 2.4, String#match?, Regexp#match? and Symbol#match? have been added. The methods are faster than match. Because the methods avoid creating a MatchData object or saving backref. So, when MatchData is not used, use match? instead of match.
- The new cops are adding hundreds of offenses.
- Rubocop's experimental autocorrect feature can help fix many of these, but: 
  - autocorrect is experimental and sometimes breaks things
  - even when it works, it can create new offenses. For instance it can make a line too long by substituting `rescue` for `fail`.
So I'll need to go through hundreds of files, a few at a time, to: autocorrect, fix offenses caused by autocorrection, and then test. (And because of time constraints, I may need to turn off Rubocop for many files, then catch up on things later.)

JoeCohen:
**Back to Basics**
I did not actually fix thousands of CodeClimate issues. Rather, due to the number of CodeClimate/RuboCop errors, I'm following a suggestion from RuboCop:
>If you have a code base with an overwhelming amount of offenses, it can be a good idea to use `rubocop --auto-gen-config` and add an `inherit_from: .rubocop_todo.yml` in your `.rubocop.yml`. The generated file `.rubocop_todo.yml` contains configuration to disable cops that currently detect an offense in the code by excluding the offending files, or disabling the cop altogether once a file count limit has been reached.
>
>By adding the option `--exclude-limit COUNT`, e.g., `rubocop --auto-gen-config --exclude-limit 5`, you can change how many files are excluded before the cop is entirely disabled. The default COUNT is 15.
>
>Then you can start removing the entries in the generated `.rubocop_todo.yml` file one by one as you work through all the offenses in the code.

See [Automatically Generated Configuration](https://github.com/bbatsov/rubocop/blob/master/manual/configuration.md#automatically-generated-configuration)

pellaea:
I think I follow what you're doing, and e.g. the new match? cop.  This
sounds like an acceptable practical plan for moving forward.

As usual, it's all we can manage seemingly to even just keep up with the
ruby and rails, let alone actually do meaningful development.  Sigh.  So it
goes.  Maybe ruby/rails isn't unique in this regard.  I hope you are less
dispirited by it than I am.

But at least we keep moving forward.  (Mostly thanks to you these last few
years.)

PR #397, opened 2018-02-20
Other taxa this name

JoeCohen:
Fix a bug in the link to Observations of: Other Taxa, this name proposed
- Add an option to the Query `nonconsensus` parameter to allow this kind of search
- Use the new search in the link

See generally [Pivotal #155310823](https://www.pivotaltracker.com/story/show/155310823)

### DIFF
--- a/app/classes/query.rb
+++ b/app/classes/query.rb
@@ -41,12 +41,12 @@
 #  4) observations with naming @name whose consensus is not a synonym of @name
 #
 #    query = Query.lookup(:Observation, :of_name, name: @name)
-#    query = Query.lookup(:Observation, :of_name,
-#                         name: @name, synonyms: :exclusive)
-#    query = Query.lookup(:Observation, :of_name,
-#                         name: @name, synonyms: :all, nonconsensus: :exclusive)
-#    query = Query.lookup(:Observation, :of_name,
-#                         name: @name, nonconsensus: :other_taxa)
+#    query = Query.lookup(:Observation, :of_name, name: @name,
+#                         synonyms: :exclusive)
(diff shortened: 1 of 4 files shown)


JoeCohen:
I am rethinking this approach, which involved adding another option to the `nonconsensus` condition.
Maybe instead:
Add some short methods to Name creating queries which can be created using the current options for `synonym` and `nonconsensus`
>
Name::observations_of_query
Name::observations_of_other_synonyms_query
Name::observations_of_taxon_query
Name::observations_other_taxa_query
Name::observations_of_other_names_this_name_proposed

For, instance
```ruby
def observations_of_other_names_this_name_proposed
  query = Query.lookup(:Observation, :of_name, name: name, by: :confidence, 
                       synonyms: :no,
                       nonconsensus: :exclusive)
end
```
(I might do the above in any event in order to move some logic out of the view helper and into the model.)

Then add 
>Name::observations_other_taxa_this_name_proposed

which calls one of the above and adds another "where".
Something like:
```ruby
def observations_other_taxa_this name_proposed_query
  query = observations_of_other_names_this_name_proposed
  query = query.where << "observations.name NOT IN #{synonyms - name}"
end
``` 
Issue: **Would SQL handle this efficiently?**  We would generate SQL with two ANDs, e.g.:
```sql
WHERE (namings.name_id IN (1172) 
AND (observations.name_id NOT IN (1172)
AND (observations.name_id NOT IN (601, 3073, 4853, 4854, 4855, 59538)
```
whereas the approach so far in this branch generates just one:
```sql
WHERE (namings.name_id IN (1172) 
AND (observations.name_id NOT IN (601, 1172, 3073, 4853, 4854, 4855, 59538)
```

PR #398, opened 2018-02-22
Obss of name

JoeCohen:
- Fix a bug in the link to Observations of: ... Other Taxa, this name proposed. See [Pivotal #155310823](https://www.pivotaltracker.com/story/show/155310823)
- Create methods on Name which wrap or modify various Query's for Observations of name.  (It took me a long time to figure out exactly how the `:synonym` and `:nonexclusive` conditions worked. The methods have names which -- at least to me -- reveal their intentions without having to dive into the Query code.
- Simplify ShowNameHelper by using the new methods directly in ShowName
- Lots of follow-on Rubocopping in the touched files.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -511,11 +511,6 @@ Metrics/PerceivedComplexity:
 Naming/AccessorMethodName:
   Enabled: false
 
-# Offense count: 1
-Naming/BinaryOperatorParameterName:
-  Exclude:
-    - 'app/models/name.rb'
-
 # Offense count: 14
 # Configuration parameters: Blacklist.
 # Blacklist: END, (?-mix:EO[A-Z]{1})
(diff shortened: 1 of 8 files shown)


JoeCohen:
Closing this (after discussion with @pellaea)
- The approach is too much of a special case. 
- The underlying `synonym` and `nonconsensus` conditions and their options don't do enough to reveal the intent of the resulting Query's 
- Creating wrapping methods is undesirable because their names may be confusing to future coders and it means another step to find the underlying code. 

Better approach:  a set of parameters/filters and options which are more transparent. E.g. `consensus` and `proposed name` with these options: 
```ruby
consensus      = name || other synonyms || taxon || excludes name || excludes other synonyms || excludes taxon 
proposed names = name || other synonyms || taxon || excludes name || excludes other synonyms || excludes taxon 
```
(As is the case now, misspellings should go along with each name.)

pellaea:
Thanks, Joe.  I think that (proposal) will a huge improvement.

PR #402, opened 2018-05-28
Remove thumbnail map

mo-nathan:
To test:
- On master:
-- Go to the show observation page and see the Map section on the right side of the screen
-- Go to the user preferences page and see an option for toggling whether map thumbnails are visible
-- See current banner

- On branch:
-- Go to the show observation page and see no Map section
-- Go to the user preferences page and see no option for map thumbnails
-- See banner with comment about the change


JoeCohen:
Did the manual tests.  

- The banner doesn't change. I don't think that's a problem in production, as we have to change the banner via the admin interface. But I'm going to take some time to look into it.
- It would be nice to have the ability to add a thumbnail map to the Show Observation page for as long as one views that particular Observation page. (I.e., if the user reloads the page, he does not automatically get the map back.) But is this too risky in terms of billing?
- Instead of our serving a map, is there a way to get it from the client side so that it's the client, rather than us, who's using the API?

mo-nathan:
For the banner did you run `rake lang:update`?

I'm not going to invest more in the maps stuff until we see the effect of this change.

PR #403, opened 2018-05-29
Remove thumbnail map

JoeCohen:
Change banner
Various Rubocop fixes and tweaks


JoeCohen:
I'm having issues with my mac and with git. I pushed changes to the wrog place. Will try again later this morning.

PR #405, opened 2018-06-13
Restored user preference box and associated tests for thumbnail maps.

pellaea:
Anything else that got deleted I should know about???

### DIFF
--- a/app/views/account/_prefs_appearance.html.erb
+++ b/app/views/account/_prefs_appearance.html.erb
@@ -33,6 +33,10 @@
                 end
       form.select(:locale, locales, class: "form-control") %><br/>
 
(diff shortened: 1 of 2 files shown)


JoeCohen:
Thanks for running with this. I think all the deletions are here: https://github.com/MushroomObserver/mushroom-observer/pull/402/files.
Sorry not to be contributing more, but at the moment I'm being consumed by some OMS projects and crises; then serious out-of-town guests. Might not be back into the swing of things until mid-July.

pellaea:
Thanks Joe!  Okay good, I finally got them all.  Didn't remember I could
look at closed PRs to get the log of changes.  That would have saved me a
lot of time! :)

PR #408, opened 2018-07-16
Temporarily disable google maps.

pellaea:
Super-simple band-aid fix.  Just to have it here in case we decide to turn it off suddenly.


pellaea:
Clearly don't need this after all.

PR #409, opened 2018-08-11
Address most cases of Naming/UncommunicativeMethodParamName

mo-nathan:
This is a rubocop warning that was fairly widespread (54 cases).  This PR addresses all but one of them.  The one remaining on in TimerParser should really be handled by a refactor of that code so is left as a reminder.

Tests were added to cover almost all the cases that were not yet covered.  The only exception in this case is user.rb where the alerting mechanism was found to be broken so fully coverage was not attempted.


pellaea:
Thanks, Nathan!

(I love that "exp" is considered "communicative" where "e" was not. :)

mo-nathan:
Yeah, I didn't work too hard in some of the cases.  The 'val' for 'x' changes are also less than awesome, but I do think these are better.  However, I do think 'e' was particularly annoying since there are a lot of 'e's in the code and way fewer 'exp's.

My goal at this point is to get rubocop to run cleanly.  The big challenge are going to be the 254 Metrics/AbcSize "offenses".  Those basically are each some sort of refactor.  I do find it is worthwhile doing, but they can be challenging.

pellaea:
Very true: searching for "exp" or "val" is definitely better than searching
for "e" or "v"!

Good luck with the metrics violations.  Actually I'm surprised there are
only 254.  At one a day, that wouldn't even take a whole man-year.  Some of
those must really be whoppers, though.

PR #410, opened 2018-08-11
Update gems

mo-nathan:
Attempts to address concerns raised by Github about older gems.  Hopefully this will also fix the Travis-CI fails on the cop-1 branch.


JoeCohen:
Thanks for doing this. I had meant to do it monthly, but -- as you can see -- got far behind.

PR #412, opened 2018-08-16
Clarified some confusion concerning which taxa of_children queries wi…

pellaea:
…ll show when "all" is false.

### DIFF
--- a/app/classes/query/initializers/of_children.rb
+++ b/app/classes/query/initializers/of_children.rb
@@ -21,6 +21,7 @@ module Query
       end
 
       # If at genus or below, we can deduce hierarchy purely by syntax.
+      # Note, this will show species below genus, not subgenera etc.
       def add_name_condition_below_genus(name, all)
         where << "names.text_name LIKE '#{name.text_name} %'"
         return if all
--- a/app/classes/query/observation_of_children.rb
+++ b/app/classes/query/observation_of_children.rb
@@ -30,10 +30,6 @@ module Query
       do_coerce(:Location)
     end
 
-    def coerce_into_name_query
-      do_coerce(:Name)
-    end
-
     def do_coerce(new_model)
       Query.lookup(new_model, :with_observations_of_children,
                    params_plus_old_by)
--- a/app/views/name/_classification.html.erb
+++ b/app/views/name/_classification.html.erb
@@ -11,7 +11,7 @@
 <% end %>
 
 <% if @first_child %>
-  <p><%= link_to(:show_object.t(type: :"RANK_PLURAL_#{@first_child.rank.to_s.upcase}"),
+  <p><%= link_to(:show_object.t(type: @name.at_or_below_genus? && !@name.at_or_below_species? ? :rank_species : :show_subtaxa_obss),
                  add_query_param({ action: :index_name }, @children_query)) %></p>
 <% end %>
 <% if @name.below_genus? && @name.genus.try(&:classification) != @name.classification %>


JoeCohen:
Probably beyond the scope of this PR, but:
Consider a species Name which has both true subtaxa (subspecies, varieties, forms) and Group(s)/Clade(s).  Example: [_Cantharellus cibarius_ Fr.](https://mushroomobserver.org/name/show_name/404)
If I'm understanding this PR, that page will include links to:

**Show Observation of, Subtaxa** (The PR doesn't change this.)
- This shows Observations of subtaxa of the species and of groups/clades. No harm done, but illogical.

**Classification, Species**
- **What will this show?** (I'm too time-stressed to try it our.). Is there a way to show Names of only the subtaxa? Do we care?

mo-nathan:
OK, I think I understand what's going on here and it seems reasonable to me.  One bit that's confusing is that Jason's explanation about the "Show Species" and "Show Groups and Clades" links doesn't apply to the Tricholomopsis as mentioned in the email, but does for Agaricus.  I need a better local environment for testing these PRs, but that's a totally separate issue.

Jason or Joe feel free to merge this if you feel Joe's concern doesn't need to be addressed here.

pellaea:
Yeah, this was a bit slap-dash right from the beginning.  And very
confusing, even to me.  Sorry for the long email...

Just to try to be more clear, the "of children" flavor of Query has two
modes.

If "all" is true, then it's nice and easy: it just selects all subtaxa at
all ranks below the given name(s).

If "all" is false, then it attempts to get just "immediate children", that
is children taxa whose direct parent is the given name(s).  This is the
mode used by both links on the show_name page.

The confusion as concerns the UI is how to characterize this latter mode.
In the majority of cases, subtaxa of families should be genera, subtaxa of
genera should be species, and so on.  But it breaks down in certain cases.

1. Below species it can be a mixture of subspecies, varieties and forms.

2. Below genus Query skips over subgenera, etc.  And note that
classifications are not presently allowed to include subgenera, etc.  So
there is no way, using the of_children flavor, to select for subgenera, etc.

3. Further confusing things, groups (and presumably clades) confuse the
algorithm.  "Agaricus group" looks just like a species.  Apparently
"Cantharellus cibarius group" looks like a subspecies/variety/form.

Okay.  To answer your questions, then, Joe:

Cantharellus cibarius -> Observations of -> Subtaxa: Presently shows
observations of subspecies/varieties/forms, as well as (erroneously)
groups.

Cantharellus cibarius -> Classsification -> Show Subtaxa (presently called
"Show Groups and Clades"): Presently and under this PR, shows
subspecies/varieties/forms, as well as (erroneously) groups.

More generally (after this PR):

Kingdom -> Observations of -> Subtaxa: shows mostly phylla but can even
include unclassified genera
Family -> Observations of -> Subtaxa: shows genera
Genus -> Observations of -> Subtaxa: shows species (and groups like
Agaricus group)
Species -> Observations of -> Subtaxa: shows infraspecies (and groups like
Cantharellus cibarius group)

Kingdom -> Classification -> Show Subtaxa: shows mostly phylla but can even
include unclassified genera
Family -> Classification -> Show Subtaxa: shows genera
Genus -> Classification -> Show Species: shows species (and groups like
Agaricus group)
Species -> Classification -> Show Subtaxa: shows infraspecies (and groups
like Cantharellus cibarius group)

(Yes, those are exactly the same sets of names selected for under both
Observations of -> Subtaxa and Classification -> Subtaxa.  I have renamed
the latter to "species" instead of "subtaxa" for genera in order to
emphasize that it's skipping subgenera and other infrageneric subtaxa.
Maybe we should do the same for Genus -> Observations of -> Subtaxa, too?)

If we would like to remove groups from these queries -- and I think we
should, I think it feels like a bug -- then I propose we make that a
separate story.  That will require nontrivial modification of the Query
code, tests, etc.  That's beyond the scope of what I can do this summer.

If we would like to give users the ability to request a list of
infrageneric taxa under genus, or list of (just) groups / clades under any
rank taxon, then we will have to expand the of_children Query flavors or
create new Query flavor(s) or use some other mechanism altogether.  Again,
definitely requires a new story, as it will require considerably more
programming effort than I can manage.

-Jason

PR #413, opened 2018-08-16
Fixed image uploader so that it will catch network-related and other …

pellaea:
…unexpected errors correctly.

### DIFF
--- a/app/assets/javascripts/multi_image_upload.js.erb
+++ b/app/assets/javascripts/multi_image_upload.js.erb
@@ -6,7 +6,8 @@ function MultiImageUploader(localized_text) {
         image_too_big_text: "This image is too large. Image files must be less than 20Mb.",
         creating_observation_text: "Creating Observation...",
         months: "January, February, March, April, May, June, July, August, September, October, November, December",
-        show_on_map: "Show on map"
+        show_on_map: "Show on map",
+        something_went_wrong: "Something went wrong while uploading image."
     };
 
     if (localized_text == undefined) {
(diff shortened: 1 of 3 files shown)


pellaea:
Except for adding a translation string, this only affects one javascript file.  No tests or rubocop affected.

mo-nathan:
This looks reasonable, but I'd love some more context about what triggers this problem and how much we're seeing it.  However, feel free to merge it.  Thanks!

pellaea:
Context was Paul Derbyshire was having network connection problems, and -- at least in my tests once I finally was able to reproduce the problem --  it caused the existing uploader simply to lock up.  The problem was we weren't watching for errors.  That is, we trapped ajax_request.onload callback (which will catch any errors which occur on the server), but not ajax_request.onerror (network or client-side errors).  The fixed version catches all possible outcomes so that it should never miss a changed state.  If it is successful, it works just like before.  If there is any unexpected error or other state (I don't even know what all the possible state changes are), then it throws up "Something went wrong." and skips the image and moves on.  At least it will create the observation, and attach any images which do successfully get uploaded and created.

There is still room for improvement.  The most obvious is that it prompts the user after each error.  So if you attach a bunch of big images and walk away, it will be waiting for you after the first failure. I think it should append the error messages one at a time to a log, and either prompt the user at the end, or pass the error messages in to create_observation and have MO display the error message using the flash_error mechanism.

And it should say which images failed to upload (it should have the file name available for this diagnostic).

I thought it might be nice to have the uploads start immediately, and not wait for you to submit the form.  But note that this actually create image objects in our database.  If the user changes their mind and doesn't submit the form, then the images will be orphaned.  We already have lots of those, but I'm concerned that doing this "eager loading" of images while the user is filling in the create observation form would result in a significant uptick.  (Note that we have no easy way presently of determining if an orphaned image is being referred to within textile, e.g., in comments, notes, name descriptions, etc.  So there is no trivial way to clean out truly orphaned -- unused -- images.)

PR #414, opened 2018-08-18
Resolve Style/MethodMissing offenses

mo-nathan:
- Replaces fake Configuration object by using the rails runner instead.  This has some overhead that we need to discuss.
- Wrapper object used in QueryTest is replaced by using MOPaginator instead.


script/bash_include
```diff
@@ -55,7 +55,7 @@ trap clean_log EXIT
 # Grab constant from rails configuration (consts.rb and consts-site.rb).
 #   path=$( rails_constant MO.local_image_files )
 function rails_constant {
-  $app_root/script/config.rb $*
+  rails runner $app_root/script/config.rb $*
```
pellaea: I think it's totally cool that "rails runner" does exactly what my crappy config.rb script did! :)

pellaea:
Sweet!  I'm blown away that it was that easy to fix.  Right on.  I'm sure we can fix the performance problem with sufficient creativity.

PR #415, opened 2018-08-22
Resolve Style/MethodMissing offenses

mo-nathan:
- Replaces fake Configuration object with a lighter weight solution driving off a new configuration file, image_config.yml.

I am a bit nervous about these changes since it doesn't really test the production version of the image paths.  We should just be careful to manual test all the image stuff when this goes out and be ready to revert it if needed.  @pellaea may have some thoughts on how best to test this.


JoeCohen:
Is PR #414 still alive?

mo-nathan:
@JoeCohen Closed #414 since now that I have all the checks passing on this branch from a personal repo.

PR #416, opened 2018-08-23
Fix Style/ExpandPathArguments rubocop offenses

mo-nathan:


pellaea:
That *is* a lot nicer.  Didn't know you could do it with __dir__ instead of __File__.  I think we were just copying Rails code all along.  Cool.

mo-nathan:
Yeah, those are the kind of fixes I love getting out of Rubocop.  I'm pretty sure you're right about the copypasta as they say.

PR #418, opened 2018-09-07
Clean up offenses related to the placement of parens

mo-nathan:
All tests should pass and all functionality should be identical.

### DIFF
--- a/app/classes/api/herbarium_record_api.rb
+++ b/app/classes/api/herbarium_record_api.rb
@@ -91,8 +91,7 @@ class API
     def before_create(params)
       obj = HerbariumRecord.where(
               herbarium: params[:herbarium],
-              accession_number: params[:accession_number]
-      ).first
+              accession_number: params[:accession_number]).first
       return nil unless obj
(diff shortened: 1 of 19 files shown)


JoeCohen:
Nathan:
Thanks so much for doing this. (It's probably something that's more suitable for someone at my pay grade, rather than someone with your skills. But I haven't gotten around to it.)
After everything is fixed, you might consider removing the fixed Cops from .rubocop_todo.yml.  (Minimal documentation about the point of that file is at https://rubocop.readthedocs.io/en/latest/configuration/#automatically-generated-configuration.)
_Joe_

mo-nathan:
@JoeCohen The fun thing about doing volunteer work is there are no pay grades!  Personally I find fixing stuff like this kind of relaxing.  A friend of mine has described it as "comfort coding".  So don't worry about that.  I had noticed the .rubocop_todo.yml and have removed a few things from there that are now deprecated.  I think we should get the existing list down to zero before taking on those.  Right now I'm dealing with the low hanging fruit.  The hard stuff is going to be the "Metrics" offenses.

PR #419, opened 2018-10-01
Fix Style/RedundantReturn offenses

mo-nathan:
To test:
Tests pass and rubocop doesn't report any Style/RedundantReturn offenses.

### DIFF
--- a/app/classes/api/base.rb
+++ b/app/classes/api/base.rb
@@ -237,12 +237,12 @@ class API
     api.handle_version
     api.authenticate_user
     api.process_request
-    return api
+    api
   rescue API::Error => e
     api ||= new(params)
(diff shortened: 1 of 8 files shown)


mo-nathan:
It said we should use match? rather than =~ when we aren't using the Match object.  However, the old code actually was sort of using the Match object via `Regexp.last_match(1)`.  I really don't like using side-effecty globals like last_match, so the best solution in my view was to break it out into two lines.

mo-nathan:
OK, this PR now includes full coverage for the lines changed by the rubocop fixes.  I think it is ready to merge.  It would be great if someone could look it over, but if I don't hear anything by Tuesday, I'll go ahead and merge it.

PR #428, opened 2019-01-25
[Delivers ###163460580] Revive assert_response_equal_file

JoeCohen:
While reviewing files during RuboCopping, I realized that I disabled this assertion 4 years ago! and never fixed it. Aarg! See https://github.com/MushroomObserver/mushroom-observer/blame/d6dc969759744dd057c03a013ecce01b0e37a5e5/test/controller_extensions.rb
Solution:
 - Reenable the assertion
 - Fix encoding issue which baffled me 4 years ago.

### DIFF
--- a/test/controller_extensions.rb
+++ b/test/controller_extensions.rb
@@ -244,8 +244,8 @@ module ControllerExtensions
       file_name = "#{html_dir}/#{label}_#{count}.html"
       count += 1
       if count > 100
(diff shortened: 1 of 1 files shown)


JoeCohen:
**Is there a better place or better way to fix this?**
(The problem is that the encoding of the response body is wrong. See my notes at https://www.pivotaltracker.com/story/show/163460580.

PR #432, opened 2019-01-28
[Delivers ##163505539] Change numerical comparisons to predicates

JoeCohen:
Predicates are the RuboCop default style and preferred by the MO developers. See https://mushroomobserver.slack.com/archives/C040TH9FV/p1548618290041400
- Delete configuration which used non-default style
- Change all instances of numerical comparisons to numerical predicates
- Exclude one non-Ruby file from Rubocop

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,6 +24,8 @@ AllCops:
     - public/design_test/jquery-1.5.2.min.js
     # We need not check these scripts, and some of them cause Rubocop errors
     - "script/old/**/*"
+    # not Ruby
+    - script/perf_monitor
     - "tmp/**/*"
 
 ###################### Global Disables #########################################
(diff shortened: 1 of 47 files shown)


JoeCohen:
Coverage decrease https://coveralls.io/builds/21312089 is bogus. 
All the alleged "New Missed Lines" are lines with changes (e.g. from `== 0` to `zero?`, which lines were previously missed. Coveralls can't tell the difference between a new line and a changed line.

JoeCohen:
May I remove `AbstractModel#find_version`? 
It's not tested and I can't find any calls at all to that method.

mo-nathan:
I suspect the coveralls change is really because of some of the changes going from one line to multiple lines and them not being covered.  That probably triggers with warning and then they mark all the new/changed lines that aren't being covered.

mo-nathan:
Note: I removed a few rubocop:disable/rubocop:enable pairs that are no longer needed.  Rubocop once again only reports Metrics issues.

PR #440, opened 2019-01-31
Fix Rubocop Style/S... offenses

JoeCohen:


JoeCohen:
Hold on!
Forgot one Style

JoeCohen:
Coveralls' report of decreased coverage is 100% bogus.
Coveralls incorrectly includes in "New Missed Lines in Diff" lines which changed (and therefore are in the diff) and are missed (uncovered), even if they were missed before. For instance, compare:
[line 40 of app/classes/map_set.rb in master prior to this PR](https://coveralls.io/builds/21369946/source?filename=app/classes/map_set.rb#L40):
```ruby
fail "Tried to map #{obj.class}!" 
```
to [the same line in this PR](https://coveralls.io/builds/21372635/source?filename=app/classes/map_set.rb#L40)
```ruby
raise "Tried to map #{obj.class}!"
```
The line was missed both before and after. So while it's "new" in the sense of "different" it is not newly missed.

JoeCohen:
This PR uses Ruby's Safe Navigation Operator `&.` to fix Style/SafeNavigation offenses.
I'm flagging that because some people object to it for esthetic reasons. I'm good with it, and generally like the idea of being on the same page as RuboCop, and therefore lots of the Ruby community.

PR #444, opened 2019-02-01
Fix RuboCop Style/Nxx-Rxx offenses

JoeCohen:
- Fix RuboCop offenses for Style/_Nxx_ through /_Rxx_ **except Style/NestedTernaryOperator**
- Update RuboCop todo file
- A few additional changes in or near lines in which offenses were fixed, in order to reduce secondary offenses and make the code more readable as I reviewed it.

Style/NestedTernaryOperator, unlike the the Styles in this PR, does not support auto-correct. It therefore will take longer to fix and will be in a separate PR.


JoeCohen:
Will look at codeclimate and coveralls issues 2/1 or 2/2

JoeCohen:
Brakeman/CodeClimate flagged a possible SQL injection issue in app/controllers/image_controller.rb ~line 733 (which begins `#{data.map`). 
If this is an invalid issue, then it's mark it as Invalid in CodeClimate. (I think the only potential problem is with `old_holder`.)
But if it's a real issue, I'd only stand a chance of fixing it if I could use ActiveRecord instead of `Image.connection.insert`.  Would that be a problem?

JoeCohen:
To me CodeClimate/Brakeman's Possible SQL injection issue is not an actual issue, and I changed it's status to Invalid.
Here's the code, with extra newlines + my comments:
```ruby
Image.connection.insert(%(
  INSERT INTO copyright_changes
    (user_id, updated_at, target_type, target_id, year, name, license_id)
  VALUES
   #{data.map { |img_id, img_when| "(
      #{@user.id},   # just the user's id. Don't see how this can be turned into SQL
      NOW(),         # a db function. No user input here.
      'Image',       # a fixed string.
      #{img_id},     # The image id,  extracted from the db by two ActiveRecord queries
                     #  images_to_update = Image.where(
                     #    user: @user, license: row[:old_id], copyright_holder: row[:old_holder]
                     #  )
                     # data = images_to_update.pluck(:id, :when)
      #{img_when.year},       # Year extracted from the image `when` field in the db. See above.
      #{safe_old_holder}, # This came from the params. But:
                     # it's quoted, safe_old_holder = Image.connection.quote(old_holder), so isn't it safe?
      #{safe_old_license_id})" }. # This came from the params. But:
                     # we called to_i on it, safe_old_license_id = row[:old_id].to_i, so isn't it safe?
    join(",\n")}
          ))
```

JoeCohen:
Closing this PR until I:
- Get Code Climate to use the right version of RuboCop;
- Explore @mo-nathan's suggestion about SQL injection.

JoeCohen:
Reopening this because I couldn't find a general solution to the SQL injection issue.But further progress is beyond my abilities.
It doesn't seem worth more time (at least in the short run). The SQL uses only two user variables; each has been freed from possible SQL injection very simply: `quote` was called on one, `to_i` on the other.

(What I tried. (Both based on pointers from @mo-nathan)
- Extracted `{data.map ...}` to a local variable; called `Image.connection.quote` on that variable; and used  that `quoted` result as the VALUE in the SQL statement. SQL choked (with the typical useless error message: "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version ... ."
- Tried to see if there's a way to use a "?" to create an SQL bind variable (https://www.ibm.com/developerworks/library/se-bindvariables/) for the VALUES clause. Just couldn't figure this out.)

JoeCohen:
I had fixed the `values_each` RuboCop issue. But now that CodeClimate is running the current RuboCop (same one as in the dev environment), CodeClimate kicks out 38 new issues. Grr.

JoeCohen:
Ready to be merged. 
**BUT** when either this or #440 is merged it will probably have conflicts with the other.

PR #445, opened 2019-02-06
Add checkbox to request exact lat/long be hidden.

pellaea:
Pretty minimal feature.  Causes lat/long to be rounded to one decimal place for all but owner of the observation.  Causes map to display location bounds instead of exact lat/long. Affects show_observation, map_observation(s), API and observation_reports.  Wrote tests to ensure that exact lat/long is indeed present, then flip on "gps_hidden" flag and ensure that the exact lat/long is no longer present anywhere in the body of the page. I have not fiddled with species lists. Also I have this automatically inherit from the last observation you created to ensure you don't accidentally forget to check the box.


JoeCohen:
I like this feature. And it will let us get rid of a related Pivotal story [Let user hide Observation lat/lon](https://www.pivotaltracker.com/story/show/153560258).
I don't understand much of the code, so can add only RuboCop-type comments.

config/locales/en.txt
```diff
@@ -1414,6 +1414,7 @@
   # observer/_form_observations
   form_observations_specimen_available: Specimen Available
   form_observations_is_collection_location: Is this location where it was collected?
+  form_observations_gps_hidden: Hide exact coordinates?
```
JoeCohen:
```suggestion
  form_observations_gps_hidden: round off latitude and longitude to hide exact location?
```

pellaea:
Thanks. Some professor just made the request to me today. Definitely higher
priorities, but it was easy, and I had a couple of hours free.

test/controllers/observer_controller_test.rb
```diff
@@ -611,6 +623,29 @@ def test_map_observations
     assert_template(:map_observations)
   end
 
+  def test_map_observation_hidden_gps
+    obs = observations(:unknown_with_lat_long)
+    get(:map_observation, params: { id: obs.id })
+    assert_true(assigns(:observations).map(&:lat).map(&:to_s).join("").include?("34.1622"))
```
JoeCohen:
```suggestion
    assert_false(assigns(:observations).map(&:lat).map(&:to_s).join("").
                                        include?("34.1622"))
```
At the moment, we're having RuboCop mask line length offenses (max 80) because there are so many.
But one day, I'll have to fix them. So best not to add another

PR #447, opened 2019-02-08
Fix Style/Mxx offenses

JoeCohen:
Fixes RuboCop Style offenses starting with "M", e.g., Style/MultilineIfModifier

If merging to master, please use a Squash Merger in order to clean up the history.


JoeCohen:
This and PR #444 will have conflicts. 
After one of them is merged to master, I will resolve them.

JoeCohen:
Conflicts resolved. All tests pass, but there is one open issue.
See Commit 99bfd4d

app/models/name/parse.rb
```diff
@@ -207,7 +216,7 @@ def self.parse_name(str, rank: :Genus, deprecated: false)
   # Guess rank of +text_name+.
   def self.guess_rank(text_name)
     TEXT_NAME_MATCHERS.each do |matcher|
-      return matcher.second if matcher.first =~ text_name
+      return matcher.second if text_name.match?(matcher.first)
```
JoeCohen: This (and lines 356-70) below were were my response to RuboCop's [Style/MultilineTernaryOperator](https://rubocop.readthedocs.io/en/latest/cops_style/#stylemultilineternaryoperator) and [Style/NestedTernaryOperator](https://rubocop.readthedocs.io/en/latest/cops_style/#stylenestedternaryoperator).
But my solution (a dispatch table based on an array) strikes me as ugly and inefficient.
Is there a better way to do this? (Something more OO/Ruby/Rails-like?)

mo-nathan: I don't understand your concern yet.  When I compare this file to what is currently on master, I don't see any significant changes to ternary operators.  A change I would at least consider is using dictionary rather than an array.  Since Ruby 1.9 hashes are guaranteed to preserve order.  It would seem clearer to me if the ranks were the keys, but then the values would need to be arrays of possible values and you'd end having to go through those lists.  Given that I'd also consider creating a small light-weight that has a regex and rank and can do the match function.  It would be tiny bit clearer, but it wouldn't really be different from what you have here.

JoeCohen: 1. The change to deal with NestedTernaryOperator was #444. It used to read:
```ruby
  def self.guess_rank(text_name)
    / (group|clade|complex)$/.match?(text_name) ? :Group :
    text_name.include?(" f. ")         ? :Form       :
    text_name.include?(" var. ")       ? :Variety    :
    text_name.include?(" subsp. ")     ? :Subspecies :
    text_name.include?(" stirps ")     ? :Stirps     :
    text_name.include?(" subsect. ")   ? :Subsection :
    text_name.include?(" sect. ")      ? :Section    :
    text_name.include?(" subgenus ")   ? :Subgenus   :
    text_name.include?(" ")            ? :Species    :
    /^\S+aceae$/.match?(text_name)     ? :Family     :
    /^\S+ineae$/.match?(text_name)     ? :Family     : # :Suborder
    /^\S+ales$/.match?(text_name)      ? :Order      :
    /^\S+mycetidae$/.match?(text_name) ? :Order      : # :Subclass
    /^\S+mycetes$/.match?(text_name)   ? :Class      :
    /^\S+mycotina$/.match?(text_name)  ? :Class      : # :Subphylum
    /^\S+mycota$/.match?(text_name)    ? :Phylum     :
    /^Fossil-/.match?(text_name)       ? :Phylum     :
                                         :Genus
  end
```
2. There's something missing in one sentence of your comment: "'... I'd also consider creating a small light-weight that has a regex and rank and can do the match function." I assume that means "a small light-weight **class**". (I just want to be sure about any new classes, as my mind is stuck on procedural programming and still has trouble dealing with OOP.)

mo-nathan: Yes.  I meant class.  That original code is kind of scary!  Either an array of arrays or a dictionary is better.

mo-nathan: I haven't tested it, but something like:
```
class RankMatcher
  attr_reader :rank
  attr_reader :pattern

  def initialize(rank, pattern)
    @rank = rank
    @pattern = pattern
  end

  def match?(text_name)
    return text_name.match?(@pattern)
  end
end

TEXT_NAME_MATCHERS = [
  RankMatcher(:Group, / (group|clade|complex)$/),
  ...
  RankMatcher(:Genus, //)]

def self.guess_rank(text_name)
  return TEXT_NAME_MATCHERS.find{ |matcher| matcher.match?(text_name) }
end
```

JoeCohen: Thanks very much for the response and especially for the example code. 
I will try implementing it, but probably won't start until tomorrow p.m.

mo-nathan: It would also be slightly safer/more explicit to remove `RankMatcher(:Genus, //)` from TEXT_NAME_MATCHERS and add it as the `ifnone` argument to `find`.

JoeCohen: Thanks again! I had been wondering about that.

mo-nathan: If you're having fun with this, then it would be even better to move the lists into the same file as the light-weight class and just have a TEXT_NAME_RANK_MATCHER and ABBREV_RANK_MATCHER defined in the class file that which parse.rb would then use.  This would imply moving a bunch of the other constants our of parse.rb and might suggest some other ways to break it up all of which would be a good thing, but potentially quite a bit more work.

JoeCohen: I am indeed enjoying this. I'll probably have time to move the two listed matchers, not time for the rest of the constants. (I'll email you & @pellaea about my schedule.)
What do you think about keeping the lightweight class, with it's moved constants, in the same file? My thought was it might be easier to understand both parse.rb and the lightweight class if one could read them together.

mo-nathan: Personally I lean towards one class per file especially given how large the files have become.  For classes that are only used by one other class or are derived from a common base class, I sometimes create sub-folders for these classes, but only if there are multiple such classes and ideally when their are for similar cases.  The example of a TextNameRankMatcher and an AbbrevRankMatcher class would fit this case nicely.  In my view, a NameParser class would be better than stuffing all this code into the Name model.

JoeCohen:
This works; probably isn't the best way to put things together, but I don't know the right way.
- The file for the lightweight class is in app/classes.
- In addition to the class, it includes two matcher arrays
- It also includes some constants which I moved there from  app/models/names/parse.rb because they are used in the matcher arrays
- It's `require`'d in app/models/names/parse.rb

Other comments:
- I didn't understand why two classes were needed or how they would differ.
- I kept the empty `:Genus` matcher (`RankMatcher.new(:Genus, //)`. I didn't understand how to use the `ifnone` argument to `find` here. `find` returns a RankMatcher instance, and `rank` is called on that instance. So I don't know what `ifnone` would equal (other than `RankMatcher.new(:Genus, //)`.).

mo-nathan:
> This works; probably isn't the best way to put things together, but I don't know the right way.

Looks good to me.

> Other comments:
> * I didn't understand why two classes were needed or how they would differ.

This isn't critical, but the idea is that rather than using the constants `TEXT_NAME_MATCHERS` and `RANK_FROM_ABBREV_MATCHERS` you could create classes TextNameMatcher and RankFromAbbrevMatcher that encapsulate both the constants and the `.find` expression.  Then you could just do `TextNameMatcher.find_rank(text_name)` and `RankFromAbbrevMatcher.find_rank(str)` rather than requiring the client to do all the `find` and `match?` stuff each time they want to use one of these.

> * I kept the empty `:Genus` matcher (`RankMatcher.new(:Genus, //)`. I didn't understand how to use the `ifnone` argument to `find` here. `find` returns a RankMatcher instance, and `rank` is called on that instance. So I don't know what `ifnone` would equal (other than `RankMatcher.new(:Genus, //)`.).

Again, not critical and unnecessary if we did what I suggest above.  But to explain, the fact that we have to specify a default in the two constants is a little bit ugly.  It would be a bit cleaner if the invoking code provided a default (using `ifnone`) that responds reasonably to `.rank`.  For example, if you did:

`RANK_FROM_ABBREV_MATCHERS.find(ifnone=RankMatcher(:Genus, //) { |matcher| matcher.match?(str) }.rank`

then you could remove default cases from the constants.  Conceptually this gives control over the default value to the client which might be useful.  The idea of giving control over the default value to the client could be added to the `.find_rank` methods described above.  However, given that this isn't expected to be used at all outside of parse.rb, it may not be worth the effort.  If these classes were going to be in a library that was going to get more widely used, then it would definitely be worth wrestling with some of these ideas.

JoeCohen:
@mo-nathan Thanks for taking the time to look and this and give such detailed comments.
I probably won't have time to work on it further until mid-March.
So merge or wait as you please.

PR #452, opened 2019-02-11
[Delivers #163850011, #115980029] Improve preferred id

JoeCohen:
- Refactor Observer preferred id code
  - Change method names to remove confusion
  - Refactor and reorganize methods
  - Remove some methods from public interface
  - Improve comments
  - Delete unnecessary tests
- Refactor other files to use only public interface

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -689,36 +689,6 @@ class Observation < AbstractModel
     end
   end
 
-  # Observation.user's unique preferred positive Name for this observation
-  # Returns falsy if there's no unique preferred positive id
-  # Used on show_observation page
-  def owner_preference
-    owner_uniq_favorite_name if owner_preference?
-  end
-
-  private # private methods used by owner_preference ###########################
-
-  # Does observation.user have a single preferred id for this observation?
-  def owner_preference?
-    owner_uniq_favorite_vote&.value&.>= Vote.owner_id_min_confidence
-  end
-
-  def owner_uniq_favorite_name
-    favs = owner_favorite_votes
-    favs[0].naming.name if favs.count == 1
-  end
-
-  def owner_uniq_favorite_vote
-    votes = owner_favorite_votes
-    return votes.first if votes.count == 1
-  end
-
-  def owner_favorite_votes
-    votes.where(user_id: user_id, favorite: true)
-  end
-
-  public #######################################################################
-
   # Change User's Vote for this naming.  Automatically recalculates the
   # consensus for the Observation in question if anything is changed.  Returns
   # true if something was changed.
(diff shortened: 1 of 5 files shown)


JoeCohen:
This is ready to merge, despite "failing" Coveralls. Although Coveralls is "failing" because of purported epsilon decreased coverage, there are no newly missed lines.

app/models/observation.rb
```diff
@@ -653,73 +655,69 @@ def users_vote(naming, user)
...
-  def owner_sure_enough?
-    return unless owner_favorite_vote
-
-    owner_favorite_vote.value >= Vote.owner_id_min_confidence
-  end
-
-  def owner_id_known?
-    owners_only_favorite_name.try(:known?)
-  end
+  public #######################################################################
```
mo-nathan: Not a big deal, but in my experience it is unusual to explicitly use `public`.  Generally all the `private` methods just go at the end of the class.  Part of the problem is that this class is really too big so that makes it hard to logically order things.

test/controllers/observer_controller_test.rb
```diff
@@ -770,47 +770,14 @@ def test_show_observation_change_vote_anonymity
   def test_show_owner_id
     login(user_with_view_owner_id_true)
     obs = observations(:owner_only_favorite_ne_consensus)
-    get_with_dump(:show_observation, id: obs.id)
+    get(:show_observation, params: { id: obs.id })
```
mo-nathan: Why get rid of the `get_with_dump`?  We don't use it much, but it is useful to ensure compliance with HTML standards.

app/models/observation.rb
JoeCohen: Thanks for he feedback. It also bothered me to have private methods in the middle of the file.
I wanted to keep those methods relatively close to `owner_preference`.
But I'm happy to move them wherever you like.

mo-nathan:
Just a minor comment and one question.  Otherwise this looks good!

test/controllers/observer_controller_test.rb
JoeCohen: Thanks again. I didn't realize why we did it. When I initially created the tests (years ago?) i was just imitating other tests. I took it out because I thought it would be slightly more efficient.
I'll put them back.

app/models/observation.rb
JoeCohen: @mo-nathan: 
I will:
- move all the preferred ID stuff to its own section following the Namings and Votes section,
- keep the private methods close to own_preference, rather than 250 lines away. 

Of course, that leaves a `public` in the middle of the file. That's not a strong preference, so if you'd rather have the private stuff at the end, just say the word.

test/controllers/observer_controller_test.rb
mo-nathan: Thanks!  Looks good.

PR #461, opened 2019-03-05
Added support for species with authors starting with "van" or "de".

pellaea:


pellaea:
Oops, sorry, should be "added support for *genera* with authors..."

PR #463, opened 2019-03-10
Fix LineLength offenses in files thru app/classes

JoeCohen:
Fix RuboCop LineLength offenses in:
- Rakefile
- app/classes/...

### DIFF
--- a/app/classes/eol_data.rb
+++ b/app/classes/eol_data.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 #
 #  = EOL Data
 #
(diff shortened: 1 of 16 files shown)


JoeCohen:
I'll deal with the CodeClimate violations later today or tomorrow.

app/classes/eol_data.rb
```diff
@@ -47,9 +46,12 @@ def total_description_count
     @id_to_description.count
   end
 
+  # grandfather method name to avoid breaking 3rd-party use of API
```
mo-nathan: Not a blocker, but I doubt there's any 3rd-party use of this.

PR #465, opened 2019-03-11
Fix controller LineLength offenses

JoeCohen:


app/controllers/description_controller_helpers.rb
```diff
@@ -16,18 +16,26 @@
...
-#  find_description::                   Look up a description based on id and controller name.
-#  merge_description_notes::            Merge the notes fields of two descriptions.
-#  initialize_description_source::      Initialize source info before serving creation form.
-#  initialize_description_permissions:: Initialize permissions of new Description.
-#  check_description_edit_permission::  Double check that no illegal changes are being made.
-#  modify_description_permissions::     Update blah_groups based on changed to two public checkboxes.
-#  update_writein::                     Update the permissions for a write-in user or group.
-#  update_groups::                      Update one type of permissions for a Hash of groups.
-#  update_group::                       Update one type of permission for one group.
+#  find_description::                   Look up a description based on id and
```
mo-nathan: Not a blocker, but as you may already know I'm not a fan of all the extra spaces in order to get things to line up between lines.  If I were doing this fix, I'd first eliminate the extra spaces which would get rid of almost all the cases where you now have to go to two lines.

PR #467, opened 2019-03-14
Fix LineLength violations in extensions

JoeCohen:


JoeCohen:
Is there a reasonable way to (probably will need a new story) to get rid of all the calls to `html_safe`? 
E.g., could we somehow use `sanitize`?

JoeCohen:
I have reviewed this to remove extra spaces which caused LineLength to be > 80.

app/extensions/string_extensions.rb
```diff
@@ -397,6 +404,7 @@ def tp_nodiv(sanitize = true)
   def tpl_nodiv(sanitize = true)
     Textile.textilize(self, true, sanitize)
```
mo-nathan: Should this have an `.html_safe`?  Or better yet, this should leverage the proposed `.textilize_safe`.

app/extensions/string_extensions.rb
```diff
@@ -381,12 +386,14 @@ def tl(sanitize = true)
 
   # TODO: Move somewhere that content_tag is defined
   def tp(sanitize = true)
-    '<div class="textile">'.html_safe + Textile.textilize(self, false, sanitize).html_safe + "</div>".html_safe
+    '<div class="textile">'.html_safe <<
+      Textile.textilize(self, false, sanitize).html_safe <<
```
mo-nathan: I'd rather there was a `textilize_safe` for that piece and then do all the other stuff with existing safe methods.

app/extensions/string_extensions.rb
```diff
@@ -363,14 +365,17 @@ class String
...
   # an HTML tag.
   HTML_TAG_PATTERN = %r{</*[A-Za-z][^>]*>}.freeze
 
   ### Textile-related methods ###
-  #
+
+  # Disable cop because we need to use html_safe to avoid added escaping
+  # rubocop:disable Rails/OutputSafety
   def t(sanitize = true)
     Textile.textilize_without_paragraph(self, false, sanitize).html_safe
```
mo-nathan: Seems like it would be better to have a `.textlize_without_paragraph_safe` that takes the same args.  It could then be leveraged by `t` and `tl` and reduce the number of `.html_safe` calls.

app/extensions/string_extensions.rb
```diff
@@ -483,16 +494,18 @@ def html_to_ascii
       gsub(/[ \t]+(\n|$)/, '\\1').    # remove superfluous trailing whitespace
       gsub(/\n+\Z/, "").              # remove superfluous newlines at end
       gsub(HTML_TAG_PATTERN, "").     # remove all <tags>
-      gsub(/^ +|[ \t]+$/, "").        # remove leading/trailing space on each line
-      gsub(/&(#\d+|[a-zA-Z]+);/) { HTML_SPECIAL_CHAR_EQUIVALENTS[Regexp.last_match(1)].to_s }.
-      html_safe                       # convert &xxx; and &#nnn; to ascii
+      gsub(/^ +|[ \t]+$/, "").        # remove leading/trailing sp on each line
+      gsub(                           # convert &xxx; and &#nnn; to ascii
+        /&(#\d+|[a-zA-Z]+);/
+      ) { HTML_SPECIAL_CHAR_EQUIVALENTS[Regexp.last_match(1)].to_s }.html_safe
```
mo-nathan: I'd isolate this in a separate function and only disable the cop for that function.

app/extensions/string_extensions.rb
```diff
@@ -483,16 +494,18 @@ def html_to_ascii
...
   end
 
   # Surround HTML string with a span that prevents long strings from being
   # broken.
   def nowrap
     '<span style="white-space:nowrap">'.html_safe + self + "</span>".html_safe
```
mo-nathan: I think this can be done safely with `content_tag`.

mo-nathan:
> Is there a reasonable way to (probably will need a new story) to get rid of all the calls to `html_safe`?
> E.g., could we somehow use `sanitize`?

That would be awesome.  Haven't dug into this stuff recently enough to be sure though.

app/classes/textile.rb
```diff
@@ -33,17 +35,36 @@ class Textile < String
...
+  ########## Class methods #####################################################
+
   # Convenience wrapper on instance method Textile#textilize_without_paragraph.
   def self.textilize_without_paragraph(str, do_object_links = false,
                                        sanitize = true)
     new(str).textilize_without_paragraph(do_object_links, sanitize)
   end
 
+  # Wrap self.textilize_without_paragraph, marking output trusted safe
+  def self.textilize_without_paragraph_safe(str, do_object_links = false,
```
JoeCohen: Does this method, and `textilize` safe belong here, or in StringExtensions?

app/extensions/string_extensions.rb
```diff
@@ -363,39 +365,46 @@ class String
...
+  # making it all safe for output
   def tpl(sanitize = true)
-    '<div class="textile">'.html_safe +
-      Textile.textilize(self, true, sanitize).html_safe +
-      "</div>".html_safe
+    textile_div_safe { Textile.textilize(self, true, sanitize) }
+  end
+
+  def textile_div_safe
+    %(<div class="textile">#{yield}</div>).
```
JoeCohen: This avoids calling `html_safe` on the div tags.

app/classes/textile.rb
mo-nathan: Because it's about textile, this seems like the right place for it.

app/extensions/symbol_extensions.rb
```diff
@@ -156,7 +157,9 @@ def self.test_localize(val, args = {}, level = []) # :nodoc:
     :test.localize_postprocessing(val, args, level)
   end
 
-  def localize_postprocessing(val, args, level, capitalize_result = false) # :nodoc:
+  def localize_postprocessing( # :nodoc:
```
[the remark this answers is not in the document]
JoeCohen: 1. Eliminated. 
2. I don't use rdoc, but do sometimes generate it by accident. Your comment prompted me to finally fix that. I'll do that in a separate PR.

app/extensions/string_extensions.rb
JoeCohen: Done.

app/extensions/string_extensions.rb
JoeCohen: Done. Thanks for the pointer; I also just applied it in another brach. And it made me understand Rails' escaping a bit better.

app/extensions/string_extensions.rb
mo-nathan: Seems like this should go into textile.rb or maybe if there was a textile_helper.rb it could get access to content_tag.

mo-nathan:
Thanks for the continued work on this.  It's looking great!  I had one last comment for you to consider, then I think it can be merged.

app/extensions/string_extensions.rb
JoeCohen: I'll give both of them a try. 
I've been wondering how to both give it access to content_tag, yet have it accessible to everything that now calls it.

mo-nathan: Yeah, since starting to do a lot more Rails at work, I haven't gotten used to Rail's opinions about what is available where or figured out how to work around them consistently.  Python is very explicit about this sort of thing which I'm getting more convinced is the right way to go.

JoeCohen: I moved `textile_div_safe` to Textile, and added a test.  I couldn't find a a way to provide access - as indicated above - to `content_tag`.

mo-nathan: Yeah, I think to resolve it we might have to move these methods to no longer be on String and into a proper helper.  However, that would mean changing all calls from something like `s.tp` to `tp s` which would be a real pain for very little gain.

PR #468, opened 2019-03-14
Fix LineLength offenses in helpers

JoeCohen:


JoeCohen:
I have reviewed this to remove extra spaces which caused LineLength to be > 80.

app/helpers/pagination_helper.rb
```diff
@@ -35,13 +35,14 @@ def pagination_letters(pages, args = {})
...
     args[:params] = (args[:params] || {}).dup
     args[:params][pages.number_arg] = nil
-    str = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z].map do |letter|
-      if !pages.used_letters || pages.used_letters.include?(letter)
-        pagination_link(letter, letter, pages.letter_arg, args)
-      else
-        content_tag(:li, content_tag(:span, letter), class: "disabled")
-      end
-    end.safe_join(" ")
+    str = %w[A B C D E F G H I J K L M N O P Q R S T U V W X Y Z].
```
mo-nathan: I'd move the `%w[A...Z]` to a constant.

app/helpers/object_link_helper.rb
```diff
@@ -146,9 +146,11 @@ def location_search_links(name)
     [
       link_to("Google Maps", "http://maps.google.com/maps?q=" + search_string),
       link_to("Yahoo Maps", "http://maps.yahoo.com/#mvt=m&q1=" + search_string),
-      link_to("Wikipedia",
-              "http://en.wikipedia.org/w/index.php?title=Special:Search&search=" +
-              search_string),
+      link_to(
```
mo-nathan: I'd make a little helper function, say `search_link` that gets called as `search_link("Google Maps", search_string)` or even `search_link(GOOGLE_MAPS, search_string)` and then have the URLs as part of a constant dictionary where the key is the site name.

app/helpers/pagination_helper.rb
JoeCohen: I'll have a go at this.

app/helpers/object_link_helper.rb
JoeCohen: I like it.

app/helpers/object_link_helper.rb
```diff
@@ -140,20 +140,26 @@ def description_link(desc)
     link_with_query(result, desc.show_link_args)
   end
 
+  # Array of links to searches on external sites;
```
JoeCohen: @mc-nathan: I hope this is what you had in mind when suggesting constant dictionary plus small helper method.

app/helpers/pagination_helper.rb
```diff
@@ -55,6 +53,8 @@ def need_letter_pagination_links?(pages)
       (!pages.used_letters || pages.used_letters.length > 1)
   end
 
+  LETTERS = ("A".."Z").freeze
```
JoeCohen: @mo-nathan / @pellaea : What's the best/most conventional place for constants in these kinds of files:
- top of module
- top of the section (i.e., here just above `def pagination_letters`)
- order of use (which is where I ended up sticking it)

JoeCohen: See prior comment about placement of constants.

app/helpers/object_link_helper.rb
mo-nathan: Looks good!  As a really minor point, I'm used to seeing constants like LOCATION_SEARCH_URLS defined before they get used.

app/helpers/pagination_helper.rb
mo-nathan: If I understand your terms, I'm most used to "top of module".  "Top of the section" is also reasonable.  "Order of use" (meaning after they get referenced) is surprising to me.

app/helpers/object_link_helper.rb
JoeCohen: Moved. Thanks for the direction here and in response to my other question about location of constants.

PR #471, opened 2019-03-16
Fix LineLength in models Nxx-Txx

JoeCohen:


JoeCohen:

app/models/name/merge.rb
```diff
@@ -29,23 +29,25 @@ def merge(old_name)
...
-      if name == self
-        name.correct_spelling = nil
-      else
-        name.correct_spelling = self
-      end
+      name.correct_spelling = if name == self
+                                nil
+                              else
+                                self
+                              end
```
mo-nathan: This seems like a legit use for the `? :` operator:
`name.correct_spelling = (name == self ? nil : self)`

app/models/notification.rb
```diff
@@ -55,56 +53,51 @@ def self.all_flavors
...
-      when :name
-        result = Name.find(obj_id)
-      end
-      @target = result
-    end
-    result
+    # target already known
+    return @target if @target
+
+    @target = case flavor
```
mo-nathan: Again this seems like a legit use for `? :`.
`(flavor == :name ? Name.find(obj_id) : nil)`

Case statements with a single case strike me as odd and I don't like not expressing the default value.

app/models/notification.rb
```diff
@@ -55,56 +53,51 @@ def self.all_flavors
...
+              when :name then Name.find(obj_id)
+              end
   end
 
   # Return a string summarizing what this Notification is about.
   def summary
-    result = "Unrecognized notification flavor"
     case flavor
     when :name
```
mo-nathan: I'd switch this to an `if` rather than `case`.

app/models/name/merge.rb
JoeCohen: Quite right.

app/models/notification.rb
JoeCohen: **Ternary operator**. Agree.

**Single-case `case`**. I also scratched my head about this, but decided to leave it alone:
1. I generally tried to minimize changes so that there was less to review;
2. The `case` statement makes it easier to add more notification flavors. 
3. There are 2 other single-case `case` expressions in this file. I think that if one changes form `case` to `if`, the others should also change. 
I'll change all three.
If, on reflection, you'd prefer the single-case `case`, just let me know.

**Unexpressed default**.  There's a special case when the default is `nil`. RuboCop complains  "Style/EmptyElse: Redundant `else`-clause" when it sees:
```ruby
if cond
  xyz
else
  nil
end
```
See https://rubocop.readthedocs.io/en/latest/cops_style/#styleemptyelse
So I'll have to do something like:
```ruby
result = nil
if cond
  result = xyz
end
result
``` 
Alternatively I could change the cop's EnforcedStyle to Empty (versus Both, which is the default).

To move things forward, I'll do the first (`result = nil`). If you change your mind, let me know and I'll go back to the unexpressed `nil`.

app/models/notification.rb
```diff
@@ -55,56 +53,51 @@ def self.all_flavors
...
+    return unless (template = note_template)
 
-        template.
-          gsub(":observer", observer.login).
-          gsub(":observation", "#{MO.http_domain}/#{naming.observation_id}").
-          gsub(":mailing_address", tracker.mailing_address || "").
-          gsub(":location", naming.observation.place_name).
-          gsub(":name", naming.format_name)
+    case flavor.to_sym
+    when :name
```
mo-nathan: It's not at all clear to my why there are all these single case `case` statements.  I would convert all of these to `if`s.

test/models/notification_test.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+# test the Notification model
+class NotificationTest < UnitTestCase
+  def test_summary
```
mo-nathan: What about notifications is this testing?  It seems like it's just validating the fixtures.

test/controllers/account_controller_test.rb
```diff
@@ -141,6 +141,22 @@ def test_invalid_login
     assert(@request.session["user_id"])
   end
 
+  def test_email_new_password
```
mo-nathan: I assume this is just providing some coverage for the AccountController.  Not sure why it's buried in this PR, but I don't see any harm in it.

app/models/notification.rb
JoeCohen: My guess:
There are 4 defined flavors, see line 35, so it was written as `case`. But only 1 flavor is available, so each `case` has just the single case.
I'll re-write as `if`.

app/models/notification.rb
mo-nathan: Not a big deal.  I'd rather get rid of the single case `case` expressions.  I'm fine with an implicit default of `nil` if rubocop insists.

JoeCohen: I forgot to say that I was able to shorten this to be a one-line ternary if, with an expressed `nil`.
```ruby
  def target
    @target ||= flavor == :name ? Name.find(obj_id) : nil
  end
 ```

File: test/models/notification_test.rb (hunk 5 above)
Comment (JoeCohen) [reply test/models/notification_test.rb]: (Grr. I accidentally clobbered my explanation.)
- I don't think the fixtures say or imply that `notification.summary` for a Name includes the display_name.
- I don't think the fixtures say or imply that the `summary` for a bad flavor is the expected string.
Background: I was trying to add some coverage to stop Coveralls from failing. The second assertion covers an allegedly "newly-missed" line. The first assertion is there for balance; it looks silly to just cover the unhappy path.

Questions: 
- Is there a better way to test `summary`?
- Are you OK with the above explanation.

File: app/models/notification.rb (hunk 2 above)
Comment (JoeCohen) [reply app/models/notification.rb]: My above comment is unclear.
I **did** get rid of the single case `case`, replacing 11 lines of code with this single line (including an expressed `nil`).
```ruby
@target ||= flavor == :name ? Name.find(obj_id) : nil
```
Is this too obscure?

mo-nathan: Works for me!

test/models/notification_test.rb
mo-nathan: Got it!  I didn't realize summary was a method and I didn't realize it was to make Coveralls happy.

PR #473, opened 2019-03-17
Fix LineLength offenses in config files

JoeCohen:


JoeCohen:
I reviewed this for possible removal of extra spaces which caused LineLength to be > 80.

PR #474, opened 2019-03-18
Fix LineLength offenses in all tasks

JoeCohen:
This needs review more than other LineLength PRs, as I assume these files are not tested.


JoeCohen:

lib/tasks/jason.rake
```diff
@@ -99,6 +101,9 @@ namespace :jason do
 
   ##############################################################################
 
+  BROWSER_MATCH =
```
mo-nathan: +1

PR #476, opened 2019-03-18
Fix LineLength offenses in /test

JoeCohen:


JoeCohen:

test/general_extensions.rb
```diff
@@ -460,9 +464,11 @@ def assert_string_equal_file(str, *files)
         result = true
         break
       elsif !msg
-        # Write out expected (old) and received (new) files for debugging purposes.
-        File.open(filename + ".old", "w:#{encoding}") { |fh| fh.write(template) }
-        File.open(filename + ".new", "w:#{encoding}") { |fh| fh.write(str) }
+        # Write out expected (old) and received (new) files for debugging.
+        File.
+          open(filename + ".old", "w:#{encoding}") { |fh| fh.write(template) }
```
mo-nathan: I'd change these to
```
... do |fh|
  ...
end
```

mo-nathan:
One minor comment.

test/general_extensions.rb
JoeCohen: Yes, that makes more sense. Will fix.

PR #483, opened 2019-04-06
Include preferred Name in Observation title

JoeCohen:
Include the Preferred Name in the Observation title.
Omit author(s) in the Observation title top line.

The Preferred Name is included as a step toward demotivating needless nomenclatural voting and arguments.
Authors are omitted to save space and improve the data-to-ink ratio. 

I had considered making the Preferred Name a link to it's Name page. I decided against that because such a link -- author included -- already appears twice: in the About  and in Proposed Names.

### DIFF
--- a/app/helpers/observer_helper.rb
+++ b/app/helpers/observer_helper.rb
@@ -1,37 +1,59 @@
 # frozen_string_literal: true
 
 # helpers for show Observation view
-module ObservationHelper
+module ObserverHelper
+  ##### Portion of page title that includes consensus (site id) ################
+
   def show_obs_title(obs)
-    @owner_id ? show_obs_title_site_id(obs) : show_obs_title_num_after_name(obs)
+    @owner_id ? obs_title_with_name_and_site_id(obs) : obs_title_with_name(obs)
   end
 
-  # Observation: Agaricus (5)
-  def show_obs_title_num_after_name(obs)
+  # Observation nnn: Hydnum washingtonianum (Site ID)
+  #   or, if name is deprecated, include preferred name in parentheses
+  # Observation nnn: Hydnum neorepandum (Hydnum washingtonianum) (Site ID)
+  def obs_title_with_name_and_site_id(obs)
     capture do
-      concat(:show_observation_header.t)
-      concat(" #{obs.id || "?"}: ")
-      concat(obs.name.format_name.t)
-    end
-  end
-
-  # Observation 5: Agaricus (Site ID)
-  def show_obs_title_site_id(obs)
-    capture do
-      concat(:show_observation_header.t)
-      concat(" #{obs.id || "?"}: ")
-      concat(obs.name.format_name.t)
+      concat(obs_title_with_name(obs))
+      # "(Site ID)" differentiates this Name from Observer Preference
       concat(" (#{:show_observation_site_id.t})")
     end
   end
 
-  # Observer Preference: Agaricus L.
+  # Same as above, sans (Site ID)
+  # Used when user prefers not to display Observer Preference
+  def obs_title_with_name(obs)
+    capture do
+      concat(:show_observation_header.t)
+      concat(" #{obs.id || "?"}: ")
+      concat(obs_title_consensus_id(obs.name))
+    end
+  end
+
+  ##### Portion of page title that includes Observer Preference ################
+
+  # Observer Preference: Hydnum repandum
   def owner_id_line(obs)
     return unless User.view_owner_id_on?
 
     capture do
       concat(:show_observation_owner_id.t + ": ")
       concat(owner_favorite_or_explanation(obs).t)
+    end
+  end
+
+  private ######################################################################
+
+  # name portion of Observation title
+  def obs_title_consensus_id(name)
+    capture do
+      concat(name.short_display_name.t)
+      # Disable cop because a guard clause (or break) inside a `capture` block
+      # doesn't return correct value for the method
+      if name.deprecated #rubocop:disable GuardClause
+        # concat leading space separately because `.t` would strip it
+        concat(" ")
+        concat("(#{name.best_preferred_synonym.short_display_name})".t)
+      end
     end
   end
 
--- a/app/helpers/observer_helper.rb
+++ b/app/helpers/observer_helper.rb
@@ -46,12 +46,12 @@ module ObserverHelper
   # name portion of Observation title
   def obs_title_consensus_id(name)
     capture do
-      concat(name.short_display_name.t)
+      concat(name.short_authors_display_name.t)
       if name.deprecated &&
          (current_name = name.best_preferred_synonym).present?
         # concat leading space separately because `.t` would strip it
         concat(" ")
-        concat("(#{current_name.short_display_name})".t)
+        concat("(#{current_name.short_authors_display_name})".t)
       end
     end
   end
--- a/app/helpers/observer_helper.rb
+++ b/app/helpers/observer_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # helpers for show Observation view
-module ShowObservationHelper
+module ObservationHelper
   def show_obs_title(obs)
     @owner_id ? show_obs_title_site_id(obs) : show_obs_title_num_after_name(obs)
   end
--- a/app/helpers/observer_helper.rb
+++ b/app/helpers/observer_helper.rb
@@ -40,7 +40,7 @@ module ObserverHelper
     if name.deprecated &&
        (current_name = name.best_preferred_synonym).present?
       capture do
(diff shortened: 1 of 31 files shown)


JoeCohen:
The net result of the commits is fine. But they're out of order because I didn't understand the fine points of doing squash commits in my Git gui.

JoeCohen:
After this PR is merged (eventually) the next step is changing the default search for Observations of Name. See https://www.pivotaltracker.com/story/show/164957447

app/helpers/observer_helper.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
```
JoeCohen: This file was renamed from ShowObservationHelper in order to comply with Rails conventions. The new name makes testing much easier.

app/models/name/format.rb
```diff
@@ -1,14 +1,23 @@
+# frozen_string_literal: true
```
JoeCohen: There is only one substantive change in this file: I added `short_display_name`.
Everything else is simply reorganization: I had a hard time figuring out where to put that method, so I moved some things around.

app/models/name/synonymy.rb
```diff
@@ -215,4 +215,28 @@ def time_of_last_naming
       last_use || created_at
     end
   end
+
+  # "Best" preferred synonym of a deprecated name.
+  def best_preferred_synonym
+    most_recently_updated(preferred_synonyms_with_most_observations)
```
JoeCohen: The algorithm here is slightly different from that in `more_popular` (used by EOL?).
`more_popular` picks the Name which has more Observations or was last used or was last created.
`best_preferred_synonym` picks the Name which has more Observations or was last updated. That's simpler to code, and doesn't hit the db.

test/models/name_test.rb
```diff
@@ -2195,6 +2195,36 @@ def test_changing_author_of_autonym
...
+      user: users(:rolf)
+    )
+    assert_equal("**__Russula__** sect. **__Russula__**",
+                 autonym.short_display_name)
+
+    # group without author
+    assert_equal(names(:unauthored_group).display_name,
+                 names(:unauthored_group).short_display_name)
+
+    # group with author
```
JoeCohen: Groups are different from other Ranks because the author (if any) does not appear immediately after formatting markup, but after an unformatted "Group"/"Clade"/etc.

test/models/name_test.rb
```diff
@@ -2503,6 +2533,80 @@ def test_other_approved_synonyms
     assert_empty(names(:lactarius_alpinus).other_approved_synonyms)
   end
 
+  def test_best_preferred_synonym
```
JoeCohen: 70 lines here to test 30 lines in the app.
But it helped me a bunch in writing the app code, and I think I have all the edge cases covered.

test/models/name_test.rb
```diff
@@ -2503,6 +2533,80 @@ def test_other_approved_synonyms
...
+    # be tied in # of Observations
+    revised_best_synonym = names(:chlorophyllum_rhacodes)
+    Observation.create(
+      name: revised_best_synonym,
+      user: users(:rolf), when: Time.current, location: locations(:albion)
+    )
+    # other_approved_synonyms.name.observations is cached by Rails, so
+    # it didn't change when we created the Observation above.
+    # So reload it
+    deprecated_name.other_approved_synonyms.
```
JoeCohen: Before I added the `reload`, I was getting mysterious behavior which completely stumped me. @pellaea talked me through it.
I hope the code comments explain the `reload` adequately.

mo-nathan:
I'm pretty sure this PR relates to https://www.pivotaltracker.com/n/projects/224629/stories/164957640.  It would help me if the initial description of PR has a link to any issue related to the work.

mo-nathan:
Personally I don't love losing all author information from the observation name.  If I were doing it, I'd probably leave the author information for the current consensus and remove it from the preferred name.  I also like the idea of making the names links.  Even though I see there are two other links, this is the biggest most obvious name and I kind of naturally want to click on it.  I honestly don't remember if there was a compelling reason not to make the observation name a link, but it sure seems like it would be handy.  I'd even consider eliminating the "About <name>" link if the title name were a link.

JoeCohen:
> I'm pretty sure this PR relates to https://www.pivotaltracker.com/n/projects/224629/stories/164957640. It would help me if the initial description of PR has a link to any issue related to the work.

You're right on both counts:  the Story involved, and including a link to the issue. I simply forgot to do that. Sorry.

JoeCohen:
> Personally I don't love losing all author information from the observation name. If I were doing it, I'd probably leave the author information for the current consensus and remove it from the preferred name. I also like the idea of making the names links. Even though I see there are two other links, this is the biggest most obvious name and I kind of naturally want to click on it. I honestly don't remember if there was a compelling reason not to make the observation name a link, but it sure seems like it would be handy. I'd even consider eliminating the "About " link if the title name were a link.

In the title, I'll:
- add author to the consensus name
- make both names into links

Are you okay with

JoeCohen:
Are you okay with following ICN Recommendation 46C.2?

> After a name published jointly by more than two authors, the citation should be restricted to the first author followed by “& al.” or “et al.”, except in the original publication.

If so, I may do this on other places on the Observation page (About, Proposed Names).

mo-nathan:
Yeah, I'm cool with "& al." or "et al.".  I would love it the nomenclature stuff was refactored in the Name object.  It would be better if we had something like `full_author` ("(J.L. Russell) G. Wu, Kuan Zhao & Zhu L. Yang"), `brief_author` ("(J.L. Russel) G. Wu et al."), rather than have these functions that hack the RedCloth markup representation to get the right thing.  If I were to do things now, I would remove all the markup stuff from the name class and put it into some sort of helper class that gets used in the view, but that would be a significant amount of work and a bit slower (but the helper could do some caching).  It would better separate the real data from the presentation.  Maybe someday...

JoeCohen:
> ..."I would love it the nomenclature stuff was refactored in the Name object. It would be better if we had something like full_author ("(J.L. Russell) G. Wu, Kuan Zhao & Zhu L. Yang"), brief_author ("(J.L. Russel) G. Wu et al."), rather than have these functions that hack the RedCloth markup representation to get the right thing. If I were to do things now, I would remove all the markup stuff from the name class and put it into some sort of helper class that gets used in the view, but that would be a significant amount of work and a bit slower (but the helper could do some caching). It would better separate the real data from the presentation. Maybe someday..."

Did you intend **that this PR include** refactoring the nomenclature stuff in the Name object instead of using a modified version of the `display_name` (the RedCloth markup)? 
I don't see an easy way to do this because creating the markup looks complicated to me. The only thing that occurs to me is to parse the name all over again. Something like:
- `str = name.text_name`;
- `str <<` `full_author` or `brief_author` as appropriate; 
-  `parsed_name = ParseName.parse_name(str, rank:` as appropriate, `deprecated:` as appropriate`)`;
-  `new_display_name = parsed_name.display_name`

Feels like a different, more complicated hack.

JoeCohen:
Related Story: https://www.pivotaltracker.com/story/show/164957640

Manual Testing Script:
- Make sure that User Preference "View Observer’s Preferred ID?" is checked.
- Simply view Observation 99708 noting:
    - In the Title
      - Site ID (consensus) shortens the authors (Redhead et al.)
      - Current name is now included. It's in parens, and without authors, after the Site ID
      - Site ID and current name link to their respective Name page
    - Elsewhere on the page, only one original author (in parens) and only one of the latest authors. This includes: Name Info section (RH side), About; Name Info, Preferred Name, Name Info, "Observations of"; Proposed Names

Issues: 
- I used the shortened (et al.)  name everywhere I could  on the Observation page. 
- ICN has no explicit examples of parenthetical (basionym) author citations with 3 authors. I applied ICN Recommendation 46C.2 separately to the basionym authors and to the authors of the new combination.   
- I'm unsure if I responded fully to @mo-nathan's  suggestions. See my Comment above. https://github.com/MushroomObserver/mushroom-observer/pull/483#issuecomment-483468494 
- What is the best place for `display_name_brief_authors` and `display_name_without_authors`? I decided to keep them in name/format so as not to separate them from `display_name`, `format_name`, etc. 
- Shortening the authors where there are > 3 relies on presence of a comma. Is there a better way?

mo-nathan:
Finally got the chance to read through all of this.  I'm ready to merge it.  There are several things going on here that are outside the stated goal of the PR, but they all seem reasonable.  My only worry is that we're going to create a kerfluffle among some of the users and we might need to effectively back this out.  All the extra stuff will make this a bit trickier, but I think it will be easy enough to just fail forward and remove some or all of the calls to `display_name_brief_authors`.

JoeCohen:
Much thanks!
*Joe Cohen*

PR #484, opened 2019-04-07
[Finshes #165119868] Update gems and related files

JoeCohen:
- `bundle update` (catalyzed by security update to Rails)
- Add, install, and require rubocop-performance gem in response to deprecation warning
- Update Codeclimate's rubocop channel in response to updating of RuboCop
- Update link to info about rubocop channel
- Reorganize Gemfile


JoeCohen:
I'll work on this. It will probably take some experimenting, as I cannot run Code Climate locally.

app/classes/pivotal/story.rb
```diff
@@ -52,7 +52,7 @@ def initialize(json)
                   end
     end
 
-    def to_json
+    def to_json(*_args)
```
mo-nathan: How is this related to updating the gems?

app/classes/query/modules/sequence_operators.rb
```diff
@@ -60,18 +59,23 @@ def prev
         new_self.current_id = result_ids[index - 1]
       end
     elsif has_outer?
-      while new_self = new_self.outer_prev
-        if new_new_self = new_self.last(:skip_outer)
-          new_self = new_new_self
-          break
-        end
-      end
+      new_self = prev_inner(new_self)
```
mo-nathan: How is this related to updating the gems?

app/controllers/observer_controller/search_controller.rb
```diff
@@ -25,12 +25,7 @@ def pattern_search # :nologin: :norobots:
       :name, :project, :species_list, :herbarium_record
       ctrlr = type
     when :google
-      if pattern.blank?
-        redirect_to(action: :list_rss_logs)
-      else
-        search = URI.escape("site:#{MO.domain} #{pattern}")
-        redirect_to("https://google.com/search?q=#{search}")
-      end
+      site_google_search(pattern)
```
mo-nathan: How is this related to updating the gems?

test/controllers/observer_controller_test.rb
```diff
@@ -554,7 +554,7 @@ def test_pattern_search
     pattern =  "hexiexiva"
     params = { search: { pattern: pattern, type: :google } }
     target =
-      "https://google.com/search?q=site:mushroomobserver.org%20#{pattern}"
+      "https://google.com/search?q=site%3Amushroomobserver.org+#{pattern}"
```
mo-nathan: Why is this change happening?  It seems kind of surprising to me.

mo-nathan:
Just read the comments on the individual commits and I think I understand the why better.  I still find the one in test/controllers/observer_controller_test.rb to be a bit surprising, but I guess I'm willing to roll with it.

app/classes/query/modules/sequence_operators.rb
JoeCohen: See below.

app/controllers/observer_controller/search_controller.rb
JoeCohen: Codeclimate had 2 flavors of complaints about this method:
- ABC/Too Many Lines/Perceived Complexity etc.  To fix those, I extracted `site_google_search`.
- Then, the follow-on complaint about URI.escape. See https://rubocop.readthedocs.io/en/latest/cops_lint/#linturiescapeunescape. I fixed that by using `URI.encode1

JoeCohen:
> Just read the comments on the individual commits and I think I understand the why better. I still find the one in test/controllers/observer_controller_test.rb to be a bit surprising, but I guess I'm willing to roll with it.

1. If you're referring to the RuboCop complaint about [Lint/UriEscapeUnescape](https://rubocop.readthedocs.io/en/latest/cops_lint/#linturiescapeunescape), see this warning about `UR.escape` in [the Ruby 2.6.2 docs](https://ruby-doc.org/stdlib-2.6.2/libdoc/uri/rdoc/URI/Escape.html#method-i-escape):
>This method is obsolete and should not be used. Instead, use CGI.escape, URI.encode_www_form or URI.encode_www_form_component depending on your specific use case.

2. How can I structure my Commits and my Comments on the PR so as to make it easier for you to review? I greatly appreciate the time you're spending on this; it's a big help for me; and I'd like to make it less work for you.

mo-nathan:
What helps me the most is if the first thing I see is a link to any relevant tickets, followed by a description of how to functionally test the work.  In this case that reasonably obvious, but in your other recent PR something like: "Find or create an observation with a deprecated name (e.g., Boletus frostii).  Note the change in the title name between the master branch and this branch." would have been a help.  Finally, anything unexpected about this PR or areas where you want special focus in the review.  So in this case, "Did some Rubocop cleanup to make CodeClimate happy.".  Personally, I only rarely dig into the individual commits to find notes there since there are often tens of commits and it's just not worth the effort.  In this case I did since I was puzzled by some of the changes.  At MassChallenge we referred to this type of stuff as "Putting the mint on the pillow".  The basic observation is that the person who creates the PR has a lot more context about the work than anyone coming and reviewing the code.  If they can quickly share some of that context it avoids a fair amount of head scratching and back and forth.  I'm also not a huge fan of squash commits.  They don't seem to me to be really worth the effort.  I'd rather have the detail if I want it and most of the time they just seem to be the equivalent of looking at the PR overview.  I expect on really big active projects like the Linux kernel they make more of a difference, but I've never worked anywhere that uses them to a significant degree.

mo-nathan:
Regarding test/controllers/observer_controller_test.rb, it just seems weird that you encoded ":", but unencoding "+" on a line that appears to have been working.  However, I subsequently found https://perishablepress.com/stop-using-unsafe-characters-in-urls/ which supports the changes, so I'm good with them.

JoeCohen:
1. Thanks for the guidance on PR Comments. It's always helpful to know how the professionals do it. Much better than my amateur improvisation. 

**Question:
Would it also help to move/make more Github comments within the code?**

2. Thanks also for the link on unsafe characters in urls. I hadn't dug into things that far.

PR #487, opened 2019-04-23
Search help

JoeCohen:
Provide easy access to pattern search help via a link in the search bar. See [Pivotal #165481629](https://www.pivotaltracker.com/story/show/165481629).

Manual Test:
- All pages with a search bar should have a "Search Help" link between the Search button and Advanced Search.
- Click on the link to display the help, which describes all available pattern search terms.

Notes:
As part of fixing RuboCop offenses I deleted all :nologin: comments in InfoController; one of them caused a new offense.

### DIFF
--- a/app/views/observer/search_bar_help.html.erb
+++ b/app/views/observer/search_bar_help.html.erb
@@ -2,9 +2,9 @@
 
 <div class="max-width-text">
   <%= :pattern_search_terms_help.tp %>
-  <%= content_tag(:p, "#{:NAMES.t} #{:SEARCHES.t}", { class: "bold" }) %>
-  <%= PatternSearch::Name.terms_help.tp %>
-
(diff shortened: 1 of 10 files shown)


app/views/observer/search_bar_help.html.erb
```diff
@@ -0,0 +1,10 @@
+<% @title = :search_bar_help_title .t %>
+
+<div class="max-width-text">
+  <%= :pattern_search_terms_help.tp %>
+  <%= content_tag(:p, "#{:NAMES.t} #{:SEARCHES.t}", { class: "bold" }) %>
```
mo-nathan: I'd consider putting the observations section first since that's the default search option and the explanation of how to do date searches is only associated with the `date` option which is only available for observation searches.  Alternatively, adding the format explanation to `created` and `modified` text would resolve that issue.

JoeCohen: Done.

mo-nathan:
Looks good!  Thanks.

PR #488, opened 2019-04-26
Tweaked Location#force_valid_lat_longs! 

pellaea:
to have it ensure locations are never a point, because that breaks the create_ and edit_location form.


pellaea:
Probably would have been better to figure out how to fix make_editable_map to create a box even when the location is just a point, thereby allowing locations to be a point, but this was *way* easier to do.

pellaea:
Sorry, don't have time to deal with size metrics.

PR #491, opened 2019-06-13
Refactoring of Query#initialize_flavor

pellaea:
At least a step in a better direction.  Mostly amounts to improving the names and transparency of the many various old obscure initialize_model_do_xxx methods.

Also, fixed image_with_observations_xxx, location_with_observations_xxx and name_with_observations_xxx to accept all the standard filter parameters defined in observation_base.  This allows any customized observation searches to be coerced successfully into image, location and name queries (the "Show Images", "Show Locations" and "Show Names" links on the observation index results page).

Lastly, I tweaked Query#coercable? so that it actually performs the coercion instead of guessing.  If it catches any exceptions, it assumes the query is not coercable. There is presumably some loss of performance, but it's not called frequently, so I believe it is safe.  And it will be much more robust.


pellaea:
Current state: passes all tests, but haven't checked rubocop.  Also, need to write many new tests for the new parameters to the xxx_with_observations_xxx queries.

JoeCohen:
Wow! That's a ton of stuff.
How can I help?

pellaea:
Can you check if this is going to conflict badly with what you've already
written?

I'm in the middle of fixing all the ABC and method length offenses now.
Later will add tests for new parameters.

I think that's all I need to do for my part(?)  If so, then you can start
incorporating it immediately.  When I'm done with my part just merge any
changes I've made into yours.

If there are more changes to Query required I can make those, just to help
avoid conflicts.

pellaea:
Hey Joe, finished with rubocop fixes.  Surprisingly, only two methods where
I had to turn off a cop.  All tests passing again.

I say it's all yours: merge it and run with it.  I'll write tests later.
Really not feeling the joy for writing tests just now.

(I pushed my changes out to my repo, I think they'll be incorporated in the
PR automatically.  Just make sure, if you wouldn't mind, that
app/classes/queries/modules/google_search.rb exists.  If not, then that's
not my latest...)

pellaea:
Oh good, it's all here.  Github often doesn't work for me, just assumed I wouldn't be able to check myself...

JoeCohen:
I will get to work on this shortly. (jJust got home from dentist.)
I can't imagine that this will have any effect on my work (other than getting some now-failing tests to pass). All I'm doing is:
For Observation and Name pattern searches,
If there's a Pattern which lacks a variable, then pattern = synonym_of:"_variable_" (and save the new pattern to the session).

pellaea:
Okay, thanks, Joe.  I'll get back to this tonight hopefully.

JoeCohen:
No conflicts between this branch and what I've done.
- I will need to fix one test in my branch
- Probably for a separate story: It would be nice if some of the page titles of pattern searches were more expressive.  E.g., If I do a pattern search for synonym_of:"Boletus edulis", the page title of the results is: 
> Observation Index.

That's the way it's always been. But It would be nice if were instead something like:
>Observations of Synonyms of Boletus edulis Bull.

That's what you get by going to https://mushroomobserver.org/name/show_name/344, then clicking "Other Taxa, this name proposed"

pellaea:
Oooh, that would be fun.  Turning a random set of parameters into something
sort of Englishy for the title. :)

I've also thought for a while it would be neat to turn all the various
indexes in MO into pattern searches.  That is, if you click on "Your
Observations", for example, it would change the search bar to "user:252".
If you clicked on observations for a given name, it would change the search
bar to "synonyms_of:Blah" or whatever is appropriate.  And so on.  It would
probably really help people learn how to use the more powerful aspects of
the search bar.

JoeCohen:
And the url would be re-usable.

JoeCohen:
What kind of tests did you have in mind? E.g.:
- Merely whether it's redirected to the right page, sort of like
```ruby
    params = { search: { pattern: 'synonym_of:"56"', type: :name } }
    get_with_dump(:pattern_search, params)
    assert_redirected_to(controller: :name, action: :name_search,
                         pattern: 'synonym_of:"56"')
```
- The search finds the correct objects?
- Something else?

Maybe I can help.

pellaea:
I wanted to add tests for all the new parameters for
images/locations/names_with_observations_xxx queries.  I started last
night, but got side-tracked when I discovered my backup drive appears to no
longer mount.  But no error messages, so I wasn't able to learn anything
googling it.  Just says it can't find an ext4 filesystem on it any more.
Eek!  That is sort of critical, so it took precedence.  Bad news is there
is no longer any way to create a full backup from scratch, since half the
original images are in object store, and I don't want MO to pay for the
bandwidth to download them all.  Oh well.  It was always extremely
difficult for me to keep up, anyway, because my personal bandwidth is so
poor.  Probably better if it is more clear that I do *not* have a backup
any more.  That was more or less the case anyway before, but it would come
as a surprise if we ever actually needed it, now it's official, and we know
not to trust it.

-Jason

JoeCohen:
This PR has been combined into PR #494

PR #493, opened 2019-06-19
[166802401] rubocop update

JoeCohen:
Delivers [Pivotal #166802401](https://www.pivotaltracker.com/story/show/166802401)
- Updates RuboCop (Adds new extensions split from main gem; Updates RuboCop configuration)
- Updates CodeClimate configuration to use highest available RuboCop channel

There is no manual test script, though I suppose one could simply RuboCop any file.
Note: when running RuboCop, one should no longer use the `-R/--rails` option.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -118,8 +118,9 @@ plugins:
   rubocop:
     enabled: true
     # Specify RuboCop version so that we're not stuck with Code Climate's
-    # default, which can be different than what's in Gemfile.lock.
-    channel: rubocop-0-66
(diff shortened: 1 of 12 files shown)


JoeCohen:
Will fix the Codeclimate issues. They all offend the [Naming/RescuedExceptionsVariableName](https://rubocop.readthedocs.io/en/latest/cops_naming/#namingrescuedexceptionsvariablename) cop.
It's a new cop on the block, which we are encountering because this PR bumps (by 4) the RuboCop version  used by Codeclimate. Mildly annoying to have a new cop added, but easy to fix (especially as it's supposed to support autocorrection.)

PR #494, opened 2019-07-03
166590668 query coercion

JoeCohen:
Delivers [Pivotal #166590668](https://www.pivotaltracker.com/story/show/166590668) (That Story is creating quite a logjam, currently blocking 6 other stories, some of which have been partially coded.)
 
This PR is built on top of (and incorporates) @pellaea's PR #491, whi:
- Improves the names and transparency of the `initialize_model_do_xxx` methods.
- Fixes `xxx_with_observations_xxx` to accept all standard filter parameters defined in observation_base, thereby allowing any customized Observation search to be coerced into Image, Location, or Name queries.

Added in this PR:
- Add missing tests so that all relevant flavors of `xxx_with_observations_xxx` are tested.
- Changed order of assertions in the test methods. (This made it easier for me to keep track of things.)
- A couple of bug fixes.
- Removed `herbarium_record` parameter for `xxx_with_observations_xxx` searches. See https://www.pivotaltracker.com/story/show/166818177

**Manual Test Script**
- pattern Search for Observations, using Pattern: `synonym_of:"Boletus edulis"`
- on resulting Observation Index page, click "Show Locations"
Result: Lists Locations, does not throw Error (which master currently does).

PR #496, opened 2019-07-04
Change default Pattern Search for Names and Observations

JoeCohen:
Delivers [Pivotal #164957447](https://www.pivotaltracker.com/story/show/164957447)
**This is a draft PR because it cannot be merged until #494 is merged**. (Test suite will not pass until it's merged.)

Manual tests
- Pattern Search for Names, pattern "Hydnum neorepandum".
Result should include _H. neorepandum_, _H. washingtonianum_, and exclude _H._ subgenus _Hydnum_ and _H._ sect. _Hydnum_ 
- Pattern Search for Observations, pattern "Boletus edulis".
Result should include Observations only of _B. edulis_, and exclude _Caloboletus inedulis_, _Boletus austroedulis_, _Amanita muscaria_ subsp. _flavivolvata_, etc.

Note: The new default Observation pattern search does not return Observations of subtaxa of the pattern. It ideally would, but that would require a new Query parameter. IMO, this PR is an improvement over master, so it should be merged when ready, and perfected later.

### DIFF
--- a/app/controllers/observer_controller/search_controller.rb
+++ b/app/controllers/observer_controller/search_controller.rb
@@ -79,13 +79,17 @@ class ObserverController
 
   ##### Advanced Search #####
 
+  ADVANCED_SEARCH_CONDITIONS = [
+    :content, :location, :name, :user
+  ].freeze
+
   # Advanced search form.  When it posts it just redirects to one of several
   # "foreign" search actions:
   #   image/advanced_search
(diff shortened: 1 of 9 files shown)


JoeCohen:
Closing this PR because its approach is obsolete. See https://www.pivotaltracker.com/story/show/164957447/comments/204469528

PR #498, opened 2019-07-10
167134181 fix hash alignment

JoeCohen:
Delivers [Pivotal #167134181](https://www.pivotaltracker.com/story/show/167134181)

- Changes alignment of all hashes to "key". See Slack discussion starting at https://mushroomobserver.slack.com/archives/C040TH9FV/p1562555355004800
- Update configuration so that RuboCop will complain about new AlighHash offenses and a few other offenses that are normally suppressed by .rubocop_todo.yml.


JoeCohen:
Codeclimate found 57 offenses. They (almost) all relate to extra spacing **within** hash values. In order to completely correct them with autocorrect, I need to temporarily, locally change Layout/ExtraSpacing AllowForAlignment: false.

JoeCohen:
Wrong. In almost every case, the extra space is to align other things. RuboCop allows that by default.
I think the only time extra spaces for alignment are not the default is alignment of hash values. Feels weird.

PR #500, opened 2019-07-13
Glossary rubocop cleanup

mo-nathan:
Should be no functional changes, but the files related to the glossary functionality should now pass rubocop.

### DIFF
--- a/app/models/glossary_term.rb
+++ b/app/models/glossary_term.rb
@@ -1,10 +1,15 @@
 class GlossaryTerm < AbstractModel
   require "acts_as_versioned"
 
-  belongs_to :thumb_image, class_name: "Image", foreign_key: "thumb_image_id"
+  belongs_to(:thumb_image,
+             class_name: "Image",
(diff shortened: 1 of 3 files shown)


app/models/glossary_term.rb
```diff
@@ -1,10 +1,18 @@
...
+  belongs_to(:thumb_image,
+             class_name: "Image",
+             foreign_key: "thumb_image_id",
+             inverse_of: :best_glossary_terms)
   belongs_to :user
   belongs_to :rss_log
+  # Rubocop would rather this was a has_many :through, but that
+  # requires a migration process as detailed here:
+  # chrisrolle.com/en/blog/migration-path-from-habtm-to-has_many-through
+  # rubocop:disable Rails/HasAndBelongsToMany
```
JoeCohen: Should this be disabled in .rubocop.yml?

mo-nathan: Yes.

PR #508, opened 2019-08-01
[Fixes #161497126] Country List

JoeCohen:
Delivers (Pivotal #161497126)[https://www.pivotaltracker.com/story/show/161497126]
- Clearer headings on /location/list_countries
- Add United Kingdom (so that we can deal with Observations in British Overseas Territories, like Cayman Islands).
- Refactor CountryCounter to use ActiveRecord query instead of SQL
- Refactor CountryCounterTest per refactor of CountryCounter and to remove RuboCop offenses

Manual Test
- http://localhost:3000/location/list_countries Result: Headings differ from https://mushroomobserver.org/location/list_countries
- now click on Cayman Islands, Edit Location, change Where to "Cayman Islands, United Kingdom". Result: Successfully updated location
- Go back to http://localhost:3000/location/list_countries, and reload. Result: United Kingdom listed under Countries with Observations, Cayman Island not listed under of Other Localities with Observations.

### DIFF
--- a/app/models/name/synonymy.rb
+++ b/app/models/name/synonymy.rb
@@ -13,9 +13,7 @@ class Name < AbstractModel
       if @synonyms
         @synonyms.map(&:id)
       elsif synonym_id
-        Name.connection.select_values(%(
-          SELECT id FROM names WHERE synonym_id = #{synonym_id}
-        )).map(&:to_i)
+        Name.where(synonym_id: synonym_id).pluck(:id)
       else
         [id]
       end
--- a/test/fixtures/names.yml
+++ b/test/fixtures/names.yml
@@ -153,6 +153,7 @@ conocybe_filaris: # 11
   display_name: '**__Conocybe filaris__**'
   search_name: Conocybe filaris
   sort_name: Conocybe filaris
+  synonym:                        # has no synonym
   created_at: 2007-03-19 15:51:00
   updated_at: 2007-03-19 15:51:00
 
--- a/test/models/name_test.rb
+++ b/test/models/name_test.rb
@@ -1,5 +1,8 @@
+# frozen_string_literal: true
+
 require "test_helper"
 
+# Tests for methods in models/name.rb and models/name/xxx.rb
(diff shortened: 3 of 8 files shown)


app/models/name/synonymy.rb
```diff
@@ -13,9 +13,7 @@ def synonym_ids
       if @synonyms
         @synonyms.map(&:id)
       elsif synonym_id
-        Name.connection.select_values(%(
-          SELECT id FROM names WHERE synonym_id = #{synonym_id}
-        )).map(&:to_i)
+        Name.where(synonym_id: synonym_id).pluck(:id)
```
JoeCohen: Although not related to the subject of this PR, refactored in connection with adding some coverage in order to get Coveralls to pass. See Comment on test/name_test.rb.

This ActiveRecord query is simpler and produces the same SQL as the old version.
`.map(&:to_i)` is unnecessary because `id` is the primary key and ActiveRecord primary keys **are** integers. https://edgeguides.rubyonrails.org/active_record_basics.html#schema-conventions

test/fixtures/names.yml
```diff
@@ -153,6 +153,7 @@ conocybe_filaris: # 11
   display_name: '**__Conocybe filaris__**'
   search_name: Conocybe filaris
   sort_name: Conocybe filaris
+  synonym:                        # has no synonym
```
JoeCohen: Affirmatively have no synonyms to ensure durability of test which is added in test/models/name_test.rb

test/models/name_test.rb
```diff
@@ -2639,6 +2643,32 @@ def test_suggest_alternate_spelling
                            Name.suggest_alternate_spellings("Lecanoa grandis"))
   end
 
+  def test_synonym_ids
```
JoeCohen: A convenient place to add coverage to fix Coveralls failure.
Coveralls failed in Commit #8c963439 although there were no newly uncovered lines.
So I picked some existing uncovered lines (though not related to the subject of this PR) and covered them.

PR #509, opened 2019-08-01
[Fixes #167646466] Upgrade Travis Ubuntu to Xenial

JoeCohen:
See [Pivotal #167646466](https://www.pivotaltracker.com/story/show/167646466)


JoeCohen:
#583489a didn't help. Is the problem that mysql is already running? How can I start it after copying the database.yml from db to config/db

PR #516, opened 2019-09-16
Change tooltip location

bertschingert:
This change moves the location of the tooltip for icons in the navigation bar at the top of the screen. 

Previously, they pop left which covers up other icons. I moved them to pop to the bottom so that all icons remain visible. 

Since this is a simple UI change, testing just involves going to any MO screen and confirming the location of the tooltip. 

See this discussion in the development google group: https://groups.google.com/forum/#!topic/mo-developers/PQm8_g29KQY


pellaea:
Looks good to me.  Any reason not to merge this and push it out immediately?  If I don't hear any objections, I'll take care of it before turning in tonight.

JoeCohen:
Thanks Thomas!
Looks good, runs fine.

pellaea:
I'll merge and push out now.

pellaea:
Arg, never mind, github doesn't work for me right now, times out.  Maybe
later...

pellaea:
Thanks, Joe.  I pushed it out.  Yup, there it is, working just fine.

PR #520, opened 2019-11-18
Added admin util to block / unblock IP at rails level.

pellaea:
Not terribly elegant, but just an admin utility, and probably temporary.

Note that this does not block IPs at the firewall.  Instead it uses the pre-existing config/blocked_ips.txt mechanism to send a message to the user along the lines of:

"we've noticed a lot of traffic from this IP, there is almost certainly a better and more respectful way of doing whatever you're doing that won't strain our server's resources, please contact the webmaster to discuss what you're doing"

This PR is just intended to make it easier for me to copy IPs from the script/identify_robots cronjob emails into the production server's copy of blocked_ips.txt.

I'll leave the PR up for a day or two in case anyone has the time or interest in reviewing it.  But I'm assuming no one is interested, and I'd rather have this utility sooner than later, because there is someone really hammering MO right now and switching IPs regularly, so I'm having to log into MO and add IPs multiple times a day, a real pain with my bandwidth the way it is.

### DIFF
--- a/app/classes/robots.rb
+++ b/app/classes/robots.rb
@@ -13,6 +13,26 @@ class Robots
       @@blocked_ips.include?(ip)
     end
 
+    def blocked_ips
+      populate_blocked_ips unless blocked_ips_current?
+      @@blocked_ips
+    end
+
+    def add_blocked_ip(ip)
+      file = MO.blocked_ips_file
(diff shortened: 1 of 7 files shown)


pellaea:
Sorry, I need this.  I can fix issues later.  Just let me know.

PR #522, opened 2019-11-27
First draft of new machanism for keeping stats on requests and blocki…

pellaea:
…ng ones exhibiting excessive traffic.

Totally untested.  Just working out how I want to approach the problem.


pellaea:
Summary:

1. New file: log/ip_stats.txt.  Every request is logged:
    -- time started, IP, user id, server time used, controller, action
2. Rails continues to block ips in config/blocked_ips.txt - config/okay_ips.txt.  Blocked IPs are sent this plain text message:
    -- We have noticed an excessive amount of server-intensive traffic from this IP address.  Please log in or contact the webmaster (webmaster@mushroomobserver.org.
3. Cron job runs very regularly and updates blocked_ips.txt based on stats in log/ip_stats.txt.
    -- Parameters are to be fine-tuned when we go live.  Right now:
        a. number of requests per minute limited
        b. total server time per minute limited
        c. logged-in users are reported by email to webmasters, but allowed
        d. non-logged-in IPs are blocked for at least 24 hours past their last infraction
        e. IPs are removed from blocked_ips.txt after 24 hours if number of requests drop below limit
4. admin page has been removed

This keeps nginx out of the loop completely => very simple and stable.

Amount of server resources is minimal for normal requests: just appends a single line to ip_stats.txt and each rails instance keeps a local copy of blocked_ips.txt up to date, refreshing automatically whenever the modification date changes.

All of the logic is handled by the cron job: see script/update_ip_stats.rb, particularly the method bad_ip?

Most of the code is now in a separate module: app/classes/ip_stats.rb.  None in app/classes/robots.rb now.  Only ApplicationController refers to IpStats: once to log stats, once to check if a request is to be blocked.

I'll work on writing tests now.

We should probably think about diagnostics we'd like to add (at least at first) to ensure that it is actually working as desired.

-Jason

app/classes/ip_stats.rb
```diff
@@ -0,0 +1,124 @@
...
+  class << self
+    def log_stats(stats)
+      file = MO.ip_stats_file
+      now = Time.current
+      File.open(file, "a") do |fh|
+        fh.puts [
+          now,
+          stats[:ip],
+          User.current_id,
+          now - start[:time],
```
mo-nathan: Should this be `stats[:time]`?

pellaea:
Yes, it should.  I haven't even tried to run it yet, let alone test it!  I
just want to know whether the general approach sounds reasonable first.

PR #526, opened 2020-01-01
hide-lichens

JoeCohen:
Fix user content Lichen filter [#170488100]

With user preferences set to "Hide Lichens", MO incorrectly showed **only** Lichen Observations. See https://www.pivotaltracker.com/story/show/170488100 and https://www.facebook.com/groups/mushroomobserver/permalink/801921940268757/. The problem was an incorrect conditional (in lichen.rb), which returned `true` when the filter was set to "no".
- Write test (with fixtures) to prevent reversion
- Broaden conditional to include values created by both Content Filter and Pattern Search
- Minor RuboCopping of touched files


JoeCohen:
@pellaea says (in email): "Yep, looks good.  I like "antilichenologist". :)"

PR #535, opened 2020-05-17
Update name of explicitly disabled cop

mo-nathan:
Minor change to a cop name.  Not clear why AlignHash no longer works, but it doesn't.

### DIFF
--- a/app/extensions/string_extensions.rb
+++ b/app/extensions/string_extensions.rb
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 #  = Extensions to String
 #  == Class Methods
--- a/app/extensions/string_extensions.rb
+++ b/app/extensions/string_extensions.rb
@@ -36,7 +36,7 @@ class String
 
   # :stopdoc:
   # Disable cop in order to make code more readable
-  # rubocop:disable Layout/HashAlignment
+  # rubocop:disable Layout/HashAlignment,Layout/AlignHash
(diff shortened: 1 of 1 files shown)


app/extensions/string_extensions.rb
```diff
@@ -36,7 +36,7 @@ class String
 
   # :stopdoc:
   # Disable cop in order to make code more readable
-  # rubocop:disable Layout/AlignHash
+  # rubocop:disable Layout/HashAlignment,Layout/AlignHash
```
JoeCohen, line 39: I suggest simply
```ruby
rubocop:disable Layout/HashAlignment
```
RuboCop has recently made a bunch of changes to namespaces. I don't know when they changed this. I think it was a long time ago, and I simply missed it.

app/extensions/string_extensions.rb
```diff
@@ -365,7 +365,7 @@ class String
       "nbsp"  => " "
     }.freeze
   end
-  # rubocop:enable Layout/AlignHash
+  # rubocop:enable Layout/HashAlignment,Layout/AlignHash
```
JoeCohen, line 368: I suggest simply
```ruby
rubocop:disable Layout/HashAlignment
```

app/extensions/string_extensions.rb
mo-nathan: That's what it was and it didn't work.  Both are needed I believe due to a mismatch between our current Rubocop version and the version used by CodeClimate.

PR #537, opened 2020-05-19
Use File.truncate rather than File::TRUNC

mo-nathan:
File::TRUNC was causing tests to fail on the new Bionic Beaver build and based on some Google searches I tried File.truncate instead and it worked like a charm.


JoeCohen:
Seems like this should be merged.

PR #538, opened 2020-05-22
Rubocop 0.83

JoeCohen:
- Upgrade RuboCop to 0.83
- Lock version (temporarily, until completion of 1st pass of autocorrections)
- Force Codeclimate to use RuboCop 0.83
- Update .rubocop.yml to enable cops added by 0.83
- Update .rubocop_todo.yml to include cops added by 0.83

PR #539, opened 2020-05-22
Hash align cop name

JoeCohen:
Tests if syncing Codeclimate channel to RuboCop version 
stops CodeClimate from choking on `Layout/HashAlignment`


JoeCohen:
Combined with https://github.com/MushroomObserver/mushroom-observer/pull/538

PR #540, opened 2020-05-23
Auto correct 2020 05 23

JoeCohen:
Starts 1st daily PR per https://www.pivotaltracker.com/story/show/172890204
This is a Work In Progress, and therefore a Draft PR. 
It should be changed to a regular PR after:
- 200 files are changed, see protocol at https://www.pivotaltracker.com/story/show/172890204; 
- Rubocop 0.83 #538 is merged; and
- Any conflicts with #538 are resolved.


JoeCohen:
Questions about work flow when fixing easily auto-correctable offenses. ([Pivotal #172890204](https://www.pivotaltracker.com/story/show/172890204)).
- **Multiple Layers of Auto-Correctable Offenses**: Auto-correcting an offense can sometimes create new _auto-correctable_ offenses, and that process can iterate. I have followed that process iteratively thinking that the object is to fix auto-correctable offenses. **Should I have done that?**
- **Increased numbers of some non-auto-correctable offenses**: Auto-correcting occasionally creates non-auto-correctable offenses. I (almost) never fixed those, thinking that the object is to fix the auto-correctable ones ASAP, leaving the rest till later. **Should I have addressed new no-auto-correctable offenses?**
- **Coverage decrease**: Coveralls reports a drop in Coverage. I ignored it for two reasons: (1) all the "new missed lines" are old uncovered lines which have changed; (2) the decrease is trivial (-0.007%). **Should I try to add some tests so that coverage does not drop at all.**
- **Regenerating `.rubocop_todo.yml`**: I did this as the last step in this PR. **Is that how it should work?**

JoeCohen:
PS: Scorecard
Total Offenses -- Before: 3134, After: 2909. About 7% decrease
Increased number of these non-auto-correctable offenses:
-  Lint/UselessAssignment 54 => 55

app/models/vote.rb
```diff
@@ -221,7 +221,8 @@ def user_weight
   # Now we are free to change the implementation later.
   def anonymous?
     (user.votes_anonymous == :no) ||
-      (user.votes_anonymous == :old && updated_at > Time.parse(MO.vote_cutoff))
+      (user.votes_anonymous == :old &&
+                               updated_at > Time.zone.parse(MO.vote_cutoff))
```
mo-nathan, line 225: This indentation is misleading.  I'm pretty sure `==` has higher precedence that `&&`, so I believe this is interpreted as:
`((user.votes_anonymous == :old) && (updated_at > Time.zone.parse(MO.vote_cutoff)))` but this indentation implies that it's interpreted as `(user.votes_anonymous == (:old && update_at > Time.zone.parse(MO.vote_cutoff)))` which I pretty sure is not what is intended.

JoeCohen: Quite right. I'll fix it.

PR #544, opened 2020-05-24
Consensus refactor

mo-nathan:
This PR refactors the method that had the highest AbcSize such that the violation is addressed and no new violations are reported.  It has full coverage except for the `add_debug_message` which I'm tempted to simply remove at this point.

Personally, this experience makes me continue to prefer a lower limit for the AbcSize metric than we have because I think some of the resulting functions are a bit more complex than they should be.  However, it is nice to at least have reached this goal.


pellaea:
Hi Nathan, I don't have any spare time these days for MO stuff.  If it
passes tests, etc. then I am fine with what you've done!  Also, please feel
free to get rid of the debug stuff.  We haven't had any problems with
calc_consensus in years.

PR #549, opened 2020-05-30
Master to bootstrap-4

mo-nathan:
This merges all the commits on master into bootstrap-4 with strong preference for whatever is on bootstrap-4.


JoeCohen:
All look trivial (added comment, line length adjustments, etc.), except `all.each` to `find_each`.

PR #551, opened 2020-05-31
Fix FrozenStringLiteralComment offenses

JoeCohen:
When complete, this PR will fix all RuboCop FrozenStringLiteralComment offenses.
Auto-correction of those offenses is not safe, and in fact throws errors in the MO test suite
because in some files we do modify Strings.
The PR therefore involves manual intervention for those files (and auto-correction in the others).

The PR will deliver [Pivotal #173095705](https://www.pivotaltracker.com/story/show/173095705)


app/helpers/object_link_helper.rb
```diff
@@ -160,7 +162,7 @@ def location_search_links(name)
   def search_link_to(site_symbol, search_string)
     return unless (url = LOCATION_SEARCH_URLS[site_symbol])
 
-    link_to(site_symbol.to_s.titlecase, url << search_string)
+    link_to(site_symbol.to_s.titlecase, "#{url}#{search_string}")
```
JoeCohen, line 165: Uses a frozen string. An alternative, with a mutable string:
```ruby
link_to(site_symbol.to_s.titlecase, url + search_string)
```

app/models/observation.rb
```diff
@@ -575,7 +577,7 @@ def self.export_formatted(notes, markup = nil)
     return "" if notes.blank?
     return notes[other_notes_key] if notes.keys == [other_notes_key]
 
-    result = notes.each_with_object("") do |(key, value), str|
+    result = notes.each_with_object(+ "") do |(key, value), str|
```
JoeCohen, line 580: Uses a mutable string. I don't see an easy way to use a frozen string, and that also doesn't make sense to me.

app/models/sequence.rb
```diff
@@ -138,7 +139,7 @@ def deposit?
 
   # url of a search for accession
   def accession_url
-    WebSequenceArchive.search_prefix(archive) << accession
+    "#{WebSequenceArchive.search_prefix(archive)}#{accession}"
```
JoeCohen, line 142: Uses a frozen string. An alternative using a mutable string:
```ruby
WebSequenceArchive.search_prefix(archive) + accession
```

app/models/species_list.rb
```diff
@@ -403,7 +405,7 @@ def construct_observation(name, args = {})
 
   # id of view textarea for a member notes heading
   def self.notes_part_id(part)
-    notes_area_id_prefix << part.tr(" ", "_")
+    "#{notes_area_id_prefix}#{part.tr(" ", "_")}"
```
JoeCohen, line 408: Uses a frozen string. An alternative using mutable string:
```ruby
notes_area_id_prefix + part.tr(" ", "_")
```

test/models/pattern_search_test.rb
```diff
@@ -1,15 +1,18 @@
...
     parser = PatternSearch::Parser.new("")
-    x = 'test name:blah two:1,2,3 foo:"quote" bar:\'a\',"b" slash:\\,,"\\""'
-    assert_equal([:pattern, "test"], parser.parse_next_term!(x))
-    assert_equal([:name, "blah"], parser.parse_next_term!(x))
-    assert_equal([:two, "1,2,3"], parser.parse_next_term!(x))
-    assert_equal([:foo, '"quote"'], parser.parse_next_term!(x))
-    assert_equal([:bar, '\'a\',"b"'], parser.parse_next_term!(x))
-    assert_equal([:slash, '\\,,"\\""'], parser.parse_next_term!(x))
+    # make str mutuable because it is modified by parse_next_term
+    str = + 'test name:blah two:1,2,3 foo:"quote" bar:\'a\',"b" slash:\\,,"\\""'
```
JoeCohen, line 9: I don't see an easy alternative to using a mutable string.

app/models/observation.rb
```diff
@@ -497,7 +499,7 @@ def other_notes=(val)
 
   # id of view textarea for a Notes heading
   def self.notes_part_id(part)
-    notes_area_id_prefix << part.tr(" ", "_")
+    "#{notes_area_id_prefix}#{part.tr(" ", "_")}"
```
JoeCohen, line 502: Uses a frozen string. An alternative using a mutable string:
```ruby
notes_area_id_prefix + part.tr(" ", "_")
```

JoeCohen:
1. While the purpose of the PR is to fix RuboCop violations, I think the bigger issue is whether we want to go with the (in-progress) Ruby program -- and current RuboCop default -- of freezing string literals. My (naive?) opinion is that it's best to follow the conventions. **What do others think?**

JoeCohen:
1. While the purpose of the PR is to fix RuboCop violations, I think the bigger issue is whether we want to go with the (in-progress) Ruby program -- and current RuboCop default -- of freezing string literals. My (naive?) opinion is that it's best to follow the conventions. 
**What do others think?**
2. This PR should be much easier to review than suggested by the number of files changed (555). There are just 7 spots that really need review, each of which is highlighted by a Review Comment.
- Almost all of those changes simply add the magic comment `# frozen_string_literal: true` to the top of the file, and -- in only about 6 files -- remove now-redundant `freeze` calls. (The removal was done by auto-correction in order to avoid manual typos.) 
- There are only 7 files really needing review. In those I made changes to avoid a Ruby `FrozenError`. I also had a question about autocorrect of FrozenStringLiteralComment. 
**In all those cases, I left a Review Comment, so all that's needed is reviewing those comments.**
3. If this PR is merged to master, my plan to also combine it with bootstrap-4 is: after resolving conflicts, go through the whole process again on that branch. See the Tasks in https://www.pivotaltracker.com/story/show/173095705

.rubocop.yml
```diff
@@ -116,7 +113,7 @@ Style/FrozenStringLiteralComment:
   # Frozen string literals may be default in Ruby 3.0.
   # The comment will be added below a shebang and encoding comment.
   # The frozen string literal comment is only valid in Ruby 2.3+.
-  AutoCorrect: false
+  AutoCorrect: true
```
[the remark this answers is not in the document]
mo-nathan: Yeah, I agree it's not safe, but it's great if you've resolved all the current cases so we can turn it on going forward.  It's one of those things that's much easier to do at the time than to retrofit.

app/models/observation.rb
mo-nathan: This seems reasonable to me.  Does rubocop have an opinion between having a space between `+` and the first quote or not?  Personally I lean towards no space.  I think of it kind of like `-5`.  I assume it has very high precedence since unary operators usually do, so I like to keep them closer.

app/helpers/object_link_helper.rb
mo-nathan: Conceptually I prefer this to be frozen if it can.  I would be surprised if we actually later modify this string.

mo-nathan:
Looks good.  I wouldn't be surprised if we bounce against some other issues that the tests missed later down the road.  I know we did at work.  It's pretty easy to miss code that modifies "constant" strings, but it's much better to be explicitly aware of them in the code.  The comments are really helpful.

Regarding getting it to bootstrap-4, I would just make sure you first merge master again into bootstrap-4 before doing a Rubocop driven pass.  That will avoid git merge pain in the future.  I'd also wait for all the tests to be working since it's the only way we have to find the problem cases.

config/initializers/backtrace_silencers.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 # Be sure to restart your server when you modify this file.
```
mo-nathan, line 2: I'd expect the blank line to be right after the # frozen... line.

JoeCohen: I'll move the blank line as you suggest.

PR #554, opened 2020-06-01
FrozenStringLiteralComment offenses

JoeCohen:
Merge changes from master https://github.com/MushroomObserver/mushroom-observer/pull/551 and subsequent master commits

- Copies master fixes of FrozenStringLiteralComment offenses


JoeCohen:
Note:
- This PR does not fix all FrozenStringLiteralComment offenses, but only those that existed in master.

JoeCohen:
My only question:
- Should I fix **all* FrozenStringLiteralComment offenses (i.e., add the magic comment to the top of all files)? 

As of now, this PR fixes only the FrozenStringLiteralComment offenses that existed in master. There are bootstrap-4 files that lack the magic comment. In those files, mutable strings are the default. This is not the case in master, where frozen strings are the default in every file.

If I fix all the bootstrap-4 FrozenStringLiteralComment offenses (i.e., add the magic comment to the top of all files), then
- We'll see Ruby `FrozenError` multiple times when running tests, which might slow us down, but
- We can fix those errors sooner rather than later.

nimmolo:
Hi Joe - Responding to your question from slack: Yes, matrix_box_presenter.rb is currently log_item_presenter.rb. 

Sorry for the brevity, im on the phone here. Just picked up some take-out Indian food :)

JoeCohen:
I'll merge this now, to avoid conflicts and potential duplicate commits in the Git History. 
See https://github.com/MushroomObserver/mushroom-observer/pull/557#issuecomment-636987748
(I'll deal later with missing magic comments.)

PR #557, opened 2020-06-01
Geocoder no method

JoeCohen:
Fix test/models Errors caused by Geocoder name change

- Changes method calls, test Class name, and test file name.

This PR fixes Errors caused by the change of Class name from `Geocoder` to `MOGeocoder` in Commit https://github.com/MushroomObserver/mushroom-observer/commit/81ddb7bb

Finishes https://www.pivotaltracker.com/story/show/173110049

### DIFF
--- a/test/models/mo_geocoder_test.rb
+++ b/test/models/mo_geocoder_test.rb
@@ -2,15 +2,15 @@
 
 require "test_helper"
 
-class GeocoderTest < UnitTestCase
+class MOGeocoderTest < UnitTestCase
   def test_unknown_place_name
-    obj = Geocoder.new("Somewhere Out There")
+    obj = MOGeocoder.new("Somewhere Out There")
     assert_not(obj.valid)
     assert_nil(obj.north)
   end
 
   def test_falmouth
-    obj = Geocoder.new("North Falmouth, Massachusetts, USA")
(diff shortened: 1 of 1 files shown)


JoeCohen:
@pellaea & @nimmolo: **Are changes also needed in js?** (`edit_location.js` and `observation_form_map.js`)

JoeCohen:
1. There's only one file changed by this PR, and the changes are tiny. See https://github.com/MushroomObserver/mushroom-observer/pull/557/commits/8aeb07af. 
(I don't know why Git/my GitGUI/GitHub wants to re-commit 21 previous commits to bootstrap-4. I'll have to re-examine my work flow to see if I can figure that out.)
2. @nimmolo & @pellaea: **Are changes also needed to js?** (`edit_location.js` and `observation_form_map.js`)

JoeCohen:
1. Thanks Jason!
2. Re your question about Geocoder. I think that namespace is used by the Geocoder gem, https://github.com/alexreisner/geocoder.
3. The reason for all the commits and file here is that they come from #554, which isn't merged. DOH! I'll figure out a way to avoid littering the Git history with duplicates. If worse comes to worse, I can simply close this PR and re-create it.

PR #558, opened 2020-06-01
Fix model test failures

JoeCohen:
Fix the 3 current test/models Failures.

### DIFF
--- a/app/classes/textile.rb
+++ b/app/classes/textile.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-# frozen_string_literal: true.
-
 require "cgi"
 require "redcloth"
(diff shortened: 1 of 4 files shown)


app/classes/textile.rb
```diff
@@ -198,10 +196,10 @@ def self.clear_textile_cache
 
   NAME_LINK_PATTERN = /
     (^|\W) (?:\**_+) ([^_]+) (?:_+\**) (?= (?:s|ish|like)? (?:\W|\Z) )
-  /x.freeze
+  /x
```
JoeCohen, line 199: `freeze` is redundant in light of `# frozen_string_literal: true`

PR #566, opened 2020-06-05
Merge recent form & controller refactor work

nimmolo:
Note: this is ongoing, the form HTML is still a mess.


app/controllers/sequences_controller.rb
```diff
@@ -73,7 +73,7 @@ def observation_index # :norobots:
     show_selected_sequences(query, always_index: true)
   end
 
-  def show
+  def show # :norobots:
```
JoeCohen, line 76: I think you can actually get rid of this magic comment per this email:
>Jason Hollinger <pellaea@gmail.com> | 23 de abril de 2019 a las 05:17
_Para:_ Joseph Cohen <jdcohenesq@gmail.com>
CC: Nathan Wilson <nathan@collectivesource.com>
Get rid of them all, I think.  I feel like I was following some pre-existing convention... maybe it was :nodoc:.  But I don't think any, even the one(s) I didn't make up, are important.  :nodoc: is effectively unused since we don't use the autodocumentation stuff.  :norobots: and :prefetch: don't appear to be used either.  Oh look, I actually even wrote something (script/old/refresh_controller.rb) that looks at these directives, but never apparently used it.-Jason

pellaea:
Yes, it looks like script/refresh_sitemap does NOT pay attention to
":norobots:" comments.  I still think it is okay to remove these magic
comments.

app/views/images/_image_thumbnail.html.erb
```diff
@@ -47,16 +47,16 @@
             }
   %>
 
-  <div class="vote-section position-absolute ab-bottom zindex-2 text-center">
-    <% if User.current and votes and image %>
+  <% if User.current and votes and image %>
```
JoeCohen, line 50:
```ruby
<% if User.current && votes && image %>
```

JoeCohen:
I looked at this quickly, but lack time to understand the forms (in general). Sorry not to be abel to contribute more.

app/views/images/_image_thumbnail.html.erb
nimmolo: Done! Thanks :)

nimmolo:
Removed all the  # :norobots: Thanks for figuring that out.

PR #573, opened 2020-06-18
Replace show_link_arg with path helpers

JoeCohen:
First commit of [Pivotal Story #173166977](https://www.pivotaltracker.com/story/show/173166977).

My text editor says:
>"168 occurrences of `show_link_arg` in found in 83 files."[1]

My plan: 
- Convert this to a regular PR by the end of the day Thu 18 Jun, even if there are still some occurrences remaining to be fixed. 
- Deal with any remaining occurrences will be dealt in a new branch and PR.
_____
[1] Many of these occurrences are in comments.


app/views/names/map.html.erb
```diff
@@ -3,7 +3,7 @@
 
   tabs = [
     link_to(:name_map_about.t(name: @name.display_name),
-            name_path(@name.show_link_args, q: get_query_param),
+            name_path(@name, q: get_query_param),
```
JoeCohen, line 6: Original appears to be a typo. I changed it to what I think was intended.

app/views/species_lists/bulk_editor.html.erb
```diff
@@ -128,7 +128,7 @@
   <div id="row_controls" class="row_controls">
     <%= :OBSERVATION.t %>:
     <%=
-      args = Observation.show_link_args("XXX")
+      args = observation_path("XXX")
```
JoeCohen, line 131: I think this is an accurate "translation" of the original.
But I have no idea how the original worked or whether the new version will work.

nimmolo: Yes this is odd. It may be used by a script... I'll check it out now.

nimmolo: Yes. JS checks for the XXX and replaces it with current_id (selected by the user)

app/views/names/map.html.erb
nimmolo: I believe you're right!

app/models/external_link.rb
```diff
@@ -61,5 +61,9 @@ def self.edit_action; end
   def self.index_action; end
   def self.show_url(id); end
   def self.show_link_args(id); end
+  def self.external_link_path(*args); end
```
JoeCohen, line 64: Define ExternalLink#external_link_path methods
Bootstrap4 uses `x_path` helpers in place of `show_link_args`.
ExternalLink has its own definition of `show_link_args`.
So **should ExternalLink have its own, corresponding definition of `external_link_path`**?

JoeCohen:
I think the changes are generally straightforward. But I left Comments on a few.
Although the changes aren't rocket science, it still bothers me to do PRs like this; there are so many test errors and failures, and this PR affects so many files, that I have no practical way to test whether the PR has broken (some) things.

app/models/external_link.rb
nimmolo: I believe this stretch of definitions is just defining these methods as "nil" for an external link, basically. I don't believe we ever call show_link_args (or would ever call object_path) on an external link, so it probably does not need its own path method. The only place i could think of that happening is in rss_logs, but i don't believe they get logged that way either.

nimmolo:
I hear you Joe. We are making sweeping updates to the code syntax and as you can see, even one syntax change consistently applied breaks tons of things. The alternative, though, seems to be increasingly complicated workarounds to keep the old code working. I have to believe this will prove to be worth it, but we'll see. If you prefer I'm happy to handle all of this rough work, and let you know when it's closer to passing. I never intended to impose this mess on anyone else!

The great thing is that the MO code is already copiously commented by everyone to date. This helps me understand how it's held together, and in figuring out how to fix tests i've broken. I haven't found anything yet that made no sense. My approach to maintaining that sanity has been keeping comments and commented-out code wherever i'm not sure of implications. 

What you're doing is helpful to me, but i don't want to stress you out :) If you prefer to take a break for awhile i'm pretty sure i can come up with less of a mess for further work.

app/views/species_lists/bulk_editor.html.erb
JoeCohen: Thanks for checking this out!

app/models/external_link.rb
JoeCohen: I'll ask @pellaea to glance at these two lines of code; I think he wrote ExternalLink.

JoeCohen:
> I hear you Joe. We are making sweeping updates to the code syntax and as you can see, even one syntax change consistently applied breaks tons of things. The alternative, though, seems to be increasingly complicated workarounds to keep the old code working. I have to believe this will prove to be worth it, but we'll see. If you prefer I'm happy to handle all of this rough work, and let you know when it's closer to passing. I never intended to impose this mess on anyone else!
> 
> The great thing is that the MO code is already copiously commented by everyone to date. This helps me understand how it's held together, and in figuring out how to fix tests i've broken. I haven't found anything yet that made no sense. My approach to maintaining that sanity has been keeping comments and commented-out code wherever i'm not sure of implications.
> 
> What you're doing is helpful to me, but i don't want to stress you out :) If you prefer to take a break for awhile i'm pretty sure i can come back in a week or so with less of a mess for further work.

I'll be okay. I'd like to do what I can to try getting the branch in better shape by Aug 1. 
This branch was a boring slog, and I didn't learn much. But working on PR #571 kept my interest. So I'll pick another controller/view/test and try to work my way through it. 

Any suggestions?

nimmolo:
Herbaria, herbarium_records, projects, sequences, and species lists (iirc maybe you did some of these already? I'm pretty tired at the moment!) would be good. Api, Account, Articles, Authors should go pretty well, but i know there are snafus in the views (especially tabs) that are tripping up all the tests, and i want to get those ironed out.

Name and Name Descriptions you can leave for me. My next PR is going to be getting Names::Descriptions in line with Locations::Descriptions, should be much quicker since I bashed it out with L::D. Then i will start fixing those controller tests, and then join you with the others.

app/models/external_link.rb
JoeCohen: As you've seen from the email, he doesn't remember enough to comment.
All these empty methods were added 4 years ago by @pellaea. The commit message:
> "Oops! Forgot to tell API how to render external links!"
I will merge the PR **with** the new methods. My thought: 
If presence/absence of those methods causes a problem, it's easier to figure out and debug if the methods are present than if they are absent.
But feel free to remove them fi you think they are a problem.

JoeCohen:
@nimmolo: Thanks for all the suggestions! I haven't done any of them.
I will work on `sequences`. (It's the only one of those which I wrote, so it might go easier.)

JoeCohen:
Now merging #573, including explicitly defined `extenal_link_path` methods. See https://github.com/MushroomObserver/mushroom-observer/pull/573#discussion_r442849546

PR #579, opened 2020-06-19
Remove `get_with_dump`

JoeCohen:
- Change non-comment `get_with_dump` to `get`
- Fix all resulting argument alignment issues
- Delete definition of `get_with_dump`


JoeCohen:
When working on #578, I decided to do this as a batch because:
- I'm confident it won't break anything
- It's easy and semi-automatic via a regex

PR #581, opened 2020-06-21
New naming

mo-nathan:
This enables the ability for anyone at any time to simply suggest a name with no vote rather than being forced to vote.


JoeCohen:
Code looks good. 
(I haven't actually tried it in dev mode because of the difficulty of switching between branches based on bootstrap4 and master.)

JoeCohen:
Tried it out in dev mode. Looks fine. (I had to switch branches for other reasons. But I'm finding it painful to move from bootstrap4 to master. I end up having to rebuild the VM.)

PR #587, opened 2020-06-24
bootstrap4 HerbariumRecordsController

JoeCohen:
Turn herbarium_records_controller_test green (when the PR is ready for review).

### DIFF
--- a/app/controllers/herbarium_records_controller.rb
+++ b/app/controllers/herbarium_records_controller.rb
@@ -113,13 +113,18 @@ class HerbariumRecordsController < ApplicationController
   alias_method :create_herbarium_record, :new
 
   def create
+    store_location
+    pass_query_params
+    @layout      = calc_layout_params
+    @observation = find_or_goto_index(Observation, params[:id])
+    return unless @observation
+
     @herbarium_record =
       HerbariumRecord.new(whitelisted_herbarium_record_params)
     normalize_parameters
-    if !validate_herbarium_name! ||
-       !can_add_record_to_herbarium?
-      return
-    elsif herbarium_label_free?
+    return if !validate_herbarium_name! || !can_add_record_to_herbarium?
+
+    if herbarium_label_free?
       @herbarium_record.save
       @herbarium_record.add_observation(@observation)
     elsif @other_record.can_edit?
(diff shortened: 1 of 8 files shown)


JoeCohen:
All tests green, but I need to do some cleanup.

app/controllers/herbarium_records_controller.rb
```diff
@@ -154,28 +131,37 @@ def edit
...
+
     old_herbarium = @herbarium_record.herbarium
     @herbarium_record.attributes = whitelisted_herbarium_record_params
     normalize_parameters
-    if !validate_herbarium_name! ||
-       !can_add_record_to_herbarium?
-      return
-    elsif herbarium_label_free?
+
+    if !validate_herbarium_name! || !can_add_record_to_herbarium?
```
JoeCohen, line 151: This part of `:update` should be refactored, but I'm leaving that for now. There's a story that includes that refactoring.
https://www.pivotaltracker.com/story/show/173515917

It's complex and messy because the `if` includes bang methods, and the happy path is in the middle of the if/end code.

app/controllers/herbarium_records_controller.rb
```diff
@@ -319,42 +343,12 @@ def can_add_record_to_herbarium?
...
     else
-      # redirect_with_query(
-      #   action: :index_herbarium_record,
-      #   id: @herbarium_record.id
-      # )
-      redirect_to herbarium_record_index_herbarium_record_path(
-        q: get_query_param
-      )
+      redirect_with_query(action: :index_herbarium_record,
+                          id: @herbarium_record.id)
```
JoeCohen, line 351: I'd like to use the path helper `herbarium_records_index_herbarium_record` but that doesn't do the right thing with `:id`:
```ruby
(byebug) herbarium_records_index_herbarium_record_path
"/herbarium_records/index_herbarium_record"
(byebug) herbarium_records_index_herbarium_record_path(123)
"/herbarium_records/index_herbarium_record.123"
(byebug) herbarium_records_index_herbarium_record_path(id: 123)
"/herbarium_records/index_herbarium_record?id=123"
```
But I want 
```ruby
/herbarium_records/index_herbarium_record/123
```
Something needs to be added/fixed in the routes. I included that in a follow-up story in order to avoid delaying merger of the changes in this PR.

nimmolo:
This looks good, Joe. Definitely mergeable

app/controllers/herbarium_records_controller.rb
nimmolo: Noted. My rake routes shows that herbarium_records_index_herbarium_record_path should be able to take an :id, but byebug probably isn't lying to you. Hmm. 

Emailing you about it separately.

PR #594, opened 2020-06-29
Fix ApplicationController#coerced_query_link

JoeCohen:
1. The action that lists the results of a search has changed between `master` to `bootstrap4`
2. `add_query_param` no longer works because
- There's no route to `.../index_action` + params id: id. So I've appended `/id` to the path.
- And I have to append the query to that.

Note also that there's no helper path to index_action(/:id)

(cherry picked from commit 955070ba984802a79d2dfc6832291273b4e6f9f1)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1052,8 +1052,10 @@ class ApplicationController < ActionController::Base
 
     [
       :show_objects.t(type: model.type_tag),
-      add_query_param({ controller: "/#{model.show_controller}",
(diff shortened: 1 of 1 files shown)


nimmolo:
Hi Joe - yes, add_query_param should be replaced that way.
Those are exactly the methods i'm going to be revising tomorrow. :)

JoeCohen:
Thanks for the feedback!
I will merge this (and other hotfixes) today so that I don't have to do the
same fix on multiple branches.

PR #595, opened 2020-07-04
bootstrap4 ArticlesController

JoeCohen:
Fix errors and some failures in ArticlesControllerTest
- Does **not** fix failures caused by incomplete UI changes.


app/views/articles/_object.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class="list-group-item">
-  <%= tag.span class: "text-lg", link_to(object.title.t,
-                  object_path(object, q: get_query_param)) %>
+  <%= tag.span link_to(object.title.t, object_path(object, q: get_query_param)),
```
JoeCohen, line 2: `tag` parameters are partly reversed from `content_tag`. In particular content comes before options.
See https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag

test/controllers/articles_controller_test.rb
```diff
@@ -4,18 +4,65 @@
...
+    assert(:success)
+  end
+
+  def test_index_article_ui
+    # Prove privileged users get link to create an article
+    login(users(:article_writer).login)
+    get(:index_article)
+    assert_select("a", { text: :create_article_title.l },
+                  "Page is missing a link to :create")
+  end
```
JoeCohen, line 27: This test fails.
I've changed my mind about how to deal with assertions that fail because the UI is incomplete. Instead of commenting them out and adding a TODO, I'll isolate them from passing assertions by putting them into separate methods.
That will make it easier to find out where the UI needs help (or the assertions need to be improved).

JoeCohen:
Ready for review and merger, although 2 tests are failing.
The failures are both due to the incomplete UI.
I've changed my mind about how to deal with such failures. See https://github.com/MushroomObserver/mushroom-observer/pull/595/files?file-filters%5B%5D=.rb#r449828700

PR #597, opened 2020-07-06
Merge master to bootstrap 4

mo-nathan:
Adds the ability for all users to merely suggest a name without expressing how strongly they support it.


JoeCohen:
Looks good!

PR #603, opened 2020-07-09
bootstrap4 UsersController

JoeCohen:
Turn bootstrap4 UsersControllerTest green

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -20,8 +20,6 @@
 #  checklist
 #
 class UsersController < ApplicationController
-# TODO: Review routes, actions: simplify & conform to other controllers,
-# including aliases, e.g.: index, show, search, search_by_contribution, _by_name
   before_action :login_required, except: [
     :checklist,
     # :lookup_user, # in LookupController, redirected to :show
@@ -32,7 +30,7 @@ class UsersController < ApplicationController
     :show_prev,
     :show_user, # aliased
     :user_search,
-    :users_by_contribution,
+    :users_by_contribution
   ]
 
   before_action :disable_link_prefetching, except: [
@@ -135,7 +133,7 @@ class UsersController < ApplicationController
     )
   end
 
-  alias_method :next_user, :show_next
+  alias next_user show_next
 
   # Go to previous user: redirects to show_user.
   def show_prev
@@ -146,7 +144,7 @@ class UsersController < ApplicationController
     )
   end
 
-  alias_method :prev_user, :show_prev
+  alias prev_user show_prev
 
   # Display a checklist of species seen by a User, Project,
   # SpeciesList or the entire site.
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,12 +1,31 @@
 # frozen_string_literal: true
 
+#  = User Controller
+#
+#  ==== Index
+#  index  == index_user ==list_users
+#  users_by_contribution
+#  users_by_name (admin only)
+#  user_search
+#
+#  ==== Show, Create, Edit
+#  show::
+#  show_next:: == next_user
+#  show_prev:: == prev_user
+#
+#  ==== Manage
+#  change_user_bonuses
+#
+#  ==== Other
+#  checklist
+#
 class UsersController < ApplicationController
 
   before_action :login_required, except: [
     :checklist,
-    :lookup_user,
-    :next_user,
-    :prev_user,
+    # :lookup_user, # in LookupController, redirected to :show
+    :next_user, # aliased
+    :prev_user, # aliased
     :show,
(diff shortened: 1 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -30,6 +49,8 @@ def index
     end
   end
 
+  alias index_user index
+
   # People guess this page name frequently for whatever reason, and
   # since there is a view with this name, it crashes each time.
   alias list_users index
```
JoeCohen, line 56: In `master` `: list_users ` was aliased.

app/views/application/_context_navbar.erb
```diff
@@ -11,13 +11,17 @@
...
-    <%= link_to({ controller: controller.controller_name, action: :new },
-                { class: "mx-4 mr-lg-5 btn nav-item nav-link btn-link",
-                  id: "create_link" }) do -%>
-          <%= tag.span class: "text-nowrap" do -%>
-            <%= tag.span :CREATE.t, class: "mr-2" -%>
-            <%= tag.span class: "fas fa-edit" -%>
-          <% end %>
+    <% if Rails.application.routes.url_helpers.method_defined?(
+            "new_#{controller.controller_name.singularize}_path"
+    ) %>
```
JoeCohen, line 16: This fixes a bug. The partial throws an error for Users, as there is no `:new` action. 
I wonder if -- in the long run -- there should be one; seems simpler to have `users` be a resource with the full set of actions, rather than calling methods like `new` and `destroy` from AccountController. But that means we have to take extra steps to hide those actions and/or make them unavailable to non-admins.

test/controllers/users_controller_test.rb
```diff
@@ -1,41 +1,28 @@
...
+  def test_index_user
+    login("rolf")
+    get(:index_user)
 
-      login("rolf")
-      get(page, params: params)
-      assert_redirected_to(controller: :rss_logs, action: :index)
-      assert_flash_text(/denied|only.*admin/i)
+    # assert_redirected_to(:root)
+    assert_redirected_to(:rss_logs)
```
JoeCohen, line 19: This fixes a bug caused by changing `:root` between `master` and `bootstrap-4`; it was `:rss_logs`, now is `observations#index`.
I'm concerned about that change. `master` gives users the option of seeing updates for all objects or only some, e.g., just Observations (though it's a little hard to notice that option). `bootstrap-4` seems to remove those options. (I, and some other users, want to see them all.)

nimmolo: Hi Joe - I anticipated that and discussed a bit with Jason a few months ago.

Note that in routes.rb, :root actually has two behaviors: it's observations#index for non-logged-in users, and rss_logs#index for logged-in users. 

To me this makes more sense - a lurker likely wants the observation log only, and either will be confused by or have no interest in the rss_log. In any case they cannot do much with anything in rss_log, unless logged in. Because the site remembers logins, regular users should notice no difference - they will be directed to rss_logs#index as the :root

Of course I want to run that by everyone, but I think people will agree that as a workflow it feels "friendlier" to approach newbies with a page that is intelligible - rather than our full "kitchen sink".

nimmolo: Therefore, to the question of your bugfix: Logged in users should probably be redirected to :root after all, for this action. 
Let's fix routes with a small PR now?

JoeCohen: @nimmolo: Thanks. I hadn't noticed the different behavior for non-logged in vs logged-in. (I was lazy, and looked at the text generated by `rails routes`. rather than the routes themselves.)
It makes good sense.
Take your time in uncommenting that line. It might be counter-productive to uncomment it until you've finished templating the changes for rss_logs#index.

app/controllers/users_controller.rb
```diff
@@ -1,12 +1,31 @@
 # frozen_string_literal: true
 
+#  = User Controller
+#
+#  ==== Index
+#  index  == index_user ==list_users
+#  users_by_contribution
+#  users_by_name (admin only)
+#  user_search
```
JoeCohen, line 9: I need to review the routes, paths, action names, and aliases to make sure I've got it all correct.
I'd also like to keep it simple and conform to other controllers e.g., 
`:search` rather than `:user_search` (with appropriate changes in the routes, aliases, tests)

nimmolo: I agree about :search action names. 
Will check through these PRs this weekend. I'm just glancing via the phone now, but hopefully we'll implement uniform index params across all the controllers too (for search and other index-refining and -sorting actions)

JoeCohen:
d7ef329623c42c224eaa70144a80362f6b6d2215 finishes the first pass. 
All tests are green, but this PR needs lots more work before it's ready for merger.

PR #605, opened 2020-07-11
bootstrap4 home pages

JoeCohen:
Different home page for logged-in users

If not logged in, home is Observation Index
If logged in, it's rss feed

See https://github.com/MushroomObserver/mushroom-observer/pull/603#discussion_r452480856, et seq.


JoeCohen:
I will close this, redo with a different branch name.

PR #609, opened 2020-07-11
Account Login hotfix

nimmolo:
This fixes login form errors occurring in IntegrationTests, where params were not passed/parsed the same way

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -212,10 +212,11 @@ class AccountController < ApplicationController
   end
 
   def login_post
-    user_params = params[:user] || {}
-    @login    = user_params[:login].to_s
-    @password = user_params[:password].to_s
-    @remember = user_params[:remember_me] == "1"
+    # New login form params are not nested under user - AN 07/20
+    # user_params = params[:user] || {}
+    @login    = params[:login].to_s # was user_params
+    @password = params[:password].to_s # was user_params
+    @remember = params[:remember_me] == "1" # was user_params
     user = User.authenticate(@login, @password)
     user ||= User.authenticate(@login, @password.strip)
 
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -191,6 +191,8 @@ class AccountController < ApplicationController
     request.method == "POST" ? login_post : login_get
   end
 
+  alias :index :login
+
   def email_new_password
     request.method == "POST" ? email_new_password_post : email_new_password_get
   end
@@ -330,6 +332,8 @@ class AccountController < ApplicationController
     update_prefs_from_form
     update_copyright_holder if prefs_changed_successfully
   end
+
+  # alias :show :prefs # maybe? or should profile be a :show? - AN
 
   def update_password
     return unless (password = params["user"]["password"])
--- a/app/helpers/thumbnail_helper.rb
+++ b/app/helpers/thumbnail_helper.rb
@@ -82,7 +82,7 @@ module ThumbnailHelper
       notes: notes
     }.merge(args)
 
-    render "images/image_thumbnail", locals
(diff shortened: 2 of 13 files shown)


nimmolo:
Second commit addresses several errors in views/shared/log_item, presenters/log_item_presenter, helpers/thumbnail_helper, and observations/index and rss_logs/index that prevented indexes from rendering.  Error-free index view rendering is necessary for many tests and to verify correct account login. (Logged-in users get rss_logs/index, not-logged-in users get observations/index.)

nimmolo:
OK, i believe this is ready for merge with bootstrap-4. 
This should make many tests pass with less trouble: account/login and observations/rss_logs/index rendering error free.

nimmolo:
Several more fixes to make rudimentary front-end views work.
Note: object_path would be very useful in the log_item_presenter because of its versatility, but it is not available, even (seemingly) when explicitly included. How to make it available?

app/controllers/account_controller.rb
```diff
@@ -330,6 +333,8 @@ def prefs
     update_copyright_holder if prefs_changed_successfully
   end
 
+  # alias :show :prefs # maybe? or should profile be a :show? - AN
```
JoeCohen, line 336: I've always been confused by the organization of `account` vs `user`. I wish I knew how other applications typically handle this stuff.
To me, `profile` feels like it belongs to `user` ... something like `user/show` (with some attributes displayed only if in admin_mode or if viewing one's own profile).
But that's probably a set of changes that should wait until all the tests are passing with the current functionality.

app/helpers/thumbnail_helper.rb
```diff
@@ -82,7 +82,7 @@ def thumbnail(image, args = {})
       notes: notes
     }.merge(args)
 
-    render "images/image_thumbnail", locals
+    render partial: "images/image_thumbnail", locals: locals
```
JoeCohen, line 85: Does the old line work., and would you prefer to use it?
```
    render "images/image_thumbnail", locals
```
My understanding `partial` is necessary only when passing something other than `locals` or when `locals` has >1 entry.

JoeCohen:
Sorry, I'm not fluid enough in Rails to be able to comment on these changes in real time. 
Nothing jumped out at me as bad.

app/helpers/thumbnail_helper.rb
nimmolo: That must be why it couldn't find the locals unless i added the keywords. There are a lot of them.

app/controllers/account_controller.rb
nimmolo: Yes - i'm not proposing changing that division of function across those two user controllers, although we can revisit later of course. The way i'm understanding the current division is: 
Account: "manage your own stuff"
User: "manage other people's stuff" (for admins)

The reason i aliased Account#index to :login was to get the navbar "context home" title link working quickly, but i believe this will need to be fixed properly to get the navbar's intended UI working. What i'm proposing there is a bit like the use of "model.show_controller" - 

The navbar as i've currently got it designed needs the controller to provide a link (with a translatable title) so the user can re-orient themselves in the current context of their activity. In Account, it should probably be login, or their profile. In Observations, it would probably be the index. Here's a screenshot of how it appears in the navbar.

Screen Shot 2020-07-12 at 8 43 47 AM

nimmolo:
Thanks for checking it over, Joe.

JoeCohen:
Thanks for the update! The screenshot looks great. (I wish I could load
a db snapshot with bootstrap4 so I could see it on my VM. But nothing
happened. I now have screwed up my VM and will have to rebuild.)
Have a great time in the mountains. (i assume the Sierras.)
I'll keep plugging away. Is there someplace I should focus my efforts?

PR #612, opened 2020-07-19
Account for login state when asserting search controller redirects

nimmolo:
This attempts to clear up confusion with the new dual :root behavior (observations_path being the new root for not-logged-in users).

The unexpected Rails behavior, to be researched another time, results from :root === observations_path.

Inside a controller, `redirect_to(observations_path)` still generates `"/observations"`, not `"/"` -- because `observations_path` is a stored object, independent of and apparently not yet processed by the routes.rb ":root" setting, within the controller. 

So in a test for not-logged-in activity, `assert_redirected_to("/observations")` will work, but `assert_redirected_to(controller: :observations, action: :index)` will not. 

When you use the controller/action hash, Rails expects the :root path ("/"), because it seems to process that route through all rules in routes.rb.


nimmolo:
Wrong branch

PR #613, opened 2020-07-19
Account for login state when asserting search controller redirects

nimmolo:
This attempts to clear up confusion with the new dual :root behavior (observations_path being the new root for not-logged-in users).

The somewhat confusing Rails behavior can be researched another time, but I hope this is a practical workaround. It only applies to the observations controller index.

Inside a controller, `redirect_to(observations_path)` apparently still generates `"/observations"`, not `"/"`. I _believe_ this is because `observations_path` is itself a stored object, independent of and not yet processed by the routes.rb ":root" setting, within the controller.

So in a test for not-logged-in activity, `assert_redirected_to("/observations")` or `assert_redirected_to observations_path` will work, _even though this is the root path_, but `assert_redirected_to(controller: :observations, action: :index)` will not. I _believe_ this is because when you use the controller/action hash, Rails seems to process that route through all rules in routes.rb, and generates the :root path ("/").

For logged in users, on the contrary, you apparently _should_ use the controller/action hash, because Rails knows you're logged in at that point, and will correctly generate the path. 

See comments also in the test.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -38,7 +38,7 @@ class SearchController < ApplicationController
       return
     else
       flash_error(:runtime_invalid.t(type: :search, value: type.inspect))
-      redirect_back_or_default(:root)
+      redirect_back_or_default(observations_path)
       return
     end
     # If pattern is blank, this would devolve into a very expensive index.
@@ -58,7 +58,7 @@ class SearchController < ApplicationController
 
   def site_google_search(pattern)
     if pattern.blank?
-      redirect_to(:root)
+      redirect_to observations_path
     else
       search = URI.encode_www_form(q: "site:#{MO.domain} #{pattern}")
       redirect_to("https://google.com/search?#{search}")
--- a/test/controllers/search_controller_test.rb
+++ b/test/controllers/search_controller_test.rb
@@ -69,19 +69,21 @@ class SearchControllerTest < FunctionalTestCase
     get(:pattern_search, params)
     assert_redirected_to(target)
 
+    # note, due to :root for non-logged-in users being "observations#index",
+    # the path literal works, but not the controller/action hash! Mysterious -AN
     params = { search: { pattern: "", type: :google } }
     get(:pattern_search, params)
-    assert_redirected_to(:root)
+    assert_redirected_to("/observations")
 
     params = { search: { pattern: "x", type: :nonexistent_type } }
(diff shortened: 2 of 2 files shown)


nimmolo:
Just updated the path strings to `observations_path` because that also works.

app/controllers/search_controller.rb
```diff
@@ -38,7 +38,7 @@ def pattern_search
       return
     else
       flash_error(:runtime_invalid.t(type: :search, value: type.inspect))
-      redirect_back_or_default(:root)
+      redirect_back_or_default(observations_path)
```
JoeCohen, line 41: The point of re-directing to `:root` in `bootstrap-4` was to change behavior for those not logged in. I'm not sure if this is intentional, but  this looks like a change in behavior from `master` for logged-in users. The corresponding `master` line is:
```ruby
redirect_back_or_default(action: :list_rss_logs)
```
(FWIW, I think the `master` behavior is right (at least for logged-in users). If I somehow mess up the search type, I want to go back to my home page.)

app/controllers/search_controller.rb
```diff
@@ -58,7 +58,7 @@ def pattern_search
 
   def site_google_search(pattern)
     if pattern.blank?
-      redirect_to(:root)
+      redirect_to observations_path
```
JoeCohen, line 61: See above.

app/controllers/search_controller.rb
nimmolo: Got it. Reverting and re-testing!

app/controllers/search_controller.rb
nimmolo: LOL! 

I had interpreted your earlier comments to mean that you had worked out a compromise that was not what you wanted.
Looks like it was working as intended though, and this PR is irrelevant.

app/controllers/search_controller.rb
nimmolo: Yes, this is definitely the whole idea behind dual-root .Logged in users in bootstrap-4 should be returned to their familiar Activity Log; not-logged-in users should never be redirected there, unless they navigate to it expressly. 

The working logic is: only people with some investment in the site are likely to understand what the Activity Log is recording, to have any interest in it, or to have any business with it. Casual visitors can safely be assumed to be on MO almost exclusively to browse observations, or maybe names, although they can of course browse wherever they like.

PR #620, opened 2020-07-25
GlossaryTerms cleanup

JoeCohen:
Some simple cleanup to follow up on PR #611:
- Eliminate funky `:action` flag in views
- Delete `GlossaryControllerTest`. (I neglected to do that in PR #611.)


app/views/glossary_terms/_form.html.erb
```diff
@@ -10,50 +10,10 @@
...
-      <%= f.select :license_id, @licenses,
-                   { selected: @upload_license_id },
-                   { class: "custom-select" } %>
-      <p class="help_block">(<%= :glossary_term_copyright_warning.t %>)</p>
-    </div>
-  <% end %>
+  <%# Let view add image input form if needed %>
+  <%# Currently, an image is added only when creating, not editing, term.
+      See /config/locales/en.txt :glossary_term_index_intro %>
+  <%= yield f %>
```
JoeCohen, line 16: This lets us eliminate the conditional and the `action` flag.

app/views/glossary_terms/edit.html.erb
```diff
@@ -16,6 +16,5 @@
 
 <div class="form-max-width mt-4">
   <%= tag.h1 @title.humanize, class: "display-1" %>
-  <%= render partial: "form",
-             locals: { button_name: :edit_glossary_term_save, action: :edit } %>
+  <%= render "form", button_name: :edit_glossary_term_save %>
```
JoeCohen, line 19: In contrast to the `new` form a block is not needed here because `edit` will ignore the form when the `form` partial yields to `edit`.

app/views/glossary_terms/_form.html.erb
nimmolo: Thanks for figuring this out. I will study what you did and try to apply as I go through!

Right now, i'm doing a whirlwind tour to add the @navbar to all views, so we can start to pass all those tests that relied on the tabset links.

test/controllers/glossary_terms_controller_test.rb
```diff
@@ -73,6 +73,8 @@ def test_new
     login
     get(:new)
     assert_template(:new)
+    assert_select("input#upload_image", true,
```
JoeCohen, line 76: Somewhat brittle. Is there something less brittle we could test for?

PR #621, opened 2020-07-25
Nimmo navbar views 072520

nimmolo:
Restore `@navbar` title and tabs to views in
Users
Votes
Translations
Support
Species Lists
Sequences
(first pass)

### DIFF
--- a/app/views/application/_navbar_link.erb
+++ b/app/views/application/_navbar_link.erb
@@ -6,6 +6,7 @@ args = {
 }
 args[:method] = link[:method].to_sym if link[:method]
 args[:confirm] = link[:confirm].to_s if link[:confirm]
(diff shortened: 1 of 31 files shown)


nimmolo:
@JoeCohen - I'm trying to get these all in quickly just to get the views _roughly_ in shape, so we have more test coverage.

JoeCohen:
I don't have anything useful to say, so merge if you you think that's best.
(I haven't looked at the navbar code, and it would take me a while to understand it. 
I don't want to hold things up. It would be good to merge this if it reduces test errors and/or increases coverage.)

nimmolo:
By the time ive got it in all the views, i dearly hope it reduces errors!
Note, i added some new tests for a couple of the new controllers split off from observer_controller

PR #623, opened 2020-07-26
Update _context_navbar.erb

nimmolo:
This just reorders the links in the navbar

### DIFF
--- a/app/views/application/_context_navbar.erb
+++ b/app/views/application/_context_navbar.erb
@@ -6,14 +6,14 @@
   <%= render "application/navbar_title",
              title: @navbar[:title] if @navbar&.key?(:title) %>
 
+  <%= render "application/navbar_filter" if @navbar&.key?(:filter) %>
+  <%= render "application/navbar_sort" if @sorts %>
+
   <% if @navbar&.key?(:links) && @navbar[:links].kind_of?(Array)
         @navbar[:links].reject(&:nil?).each do |link| %>
       <%= render "application/navbar_link", link: link %>
     <% end %>
   <% end %>
-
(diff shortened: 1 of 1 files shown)


nimmolo:
@JoeCohen  - i'm out today, but will email you a note about navbar in case this helps.

nimmolo:
On second thought I'll leave the note here for reference.

The top (context) navbar gets `@navbar` instance variable from the view. 

This should contain:
`@navbar[:title] = { title: :CONTROLLER_TITLE(USUALLY).t, url: controller_index_path_usually }`
This almost always links to the index, so any identical index links can be eliminated from the below.

Optional attributes:
`@navbar[:filter]` - only used for Activity Log, check implementation there. Basically a dropdown with checkboxes
`@navbar[:sorts]` - takes @sorts instance if present and builds a dropdown of sortby links, used in several indexes
`@navbar[:dropdowns]` - array of hashes, each with a `title:` and an array of `links:` -- can use pre-existing `tabs` array, or `create_links(@links)` if `@links` variable set by controller.
`@navbar[:links]` - an array of hashes of top-level navbar links. Use these if there are 2 or fewer links, or maybe 4 and no searchbar. Note that these can contain an `icon:` key, see preliminary examples in views now (icons TBD later, should probably set in a central place as constants.). Generally, `Edit`, `Create`, `Destroy`, `Cancel and go back` links should probably get the same consistent icons, can copy from existing.
`@navbar[:search]` - set to any value to make the searchbar appear. Note the idea is to get this out of the way in irrelevant contexts, because :root is always only a click a way. Trying to only implement in contexts where directly useful: observations, rss_logs, images, names, locations. Otherwise people can always click on the home link and search from there.

PR #624, opened 2020-07-27
fix bootstrap4 Description Query

JoeCohen:
Makes Query work with namespaced models: `Location::Description`, `Name::Description`

Delivers https://www.pivotaltracker.com/story/show/173999735

### DIFF
--- a/app/classes/query/modules/titles.rb
+++ b/app/classes/query/modules/titles.rb
@@ -9,12 +9,9 @@ module Query
 
       def initialize_title
         @title_tag = "query_title_#{flavor}".to_sym
-        @title_args = {
-          type: model.to_s.underscore.
-                # for namespaced models, convert paths to underscore separation
-                gsub("/", "_").
-                to_sym
-        }
+        @title_args = { type: model.to_s.underscore.
+                      tr("/", "_"). # for namespaced models
+                      to_sym }
       end
 
       # translated title, used by views
--- a/app/classes/query/modules/titles.rb
+++ b/app/classes/query/modules/titles.rb
@@ -9,9 +9,15 @@ module Query
 
       def initialize_title
         @title_tag = "query_title_#{flavor}".to_sym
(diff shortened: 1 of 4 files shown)


app/classes/query/modules/titles.rb
```diff
@@ -9,9 +9,12 @@ module Titles
 
       def initialize_title
         @title_tag = "query_title_#{flavor}".to_sym
-        @title_args = { type: model.to_s.underscore.to_sym }
+        @title_args = { type: model.to_s.underscore.
+                      tr("/", "_"). # for namespaced models
+                      to_sym }
```
JoeCohen, line 14: This would create an ambiguity if we ever had models `A::B` and `A_B`.
But I think that's very unlikely, and don't want to spend the time to engage in Speculative Generality.

pellaea: Agreed.

JoeCohen:
@pellaea looked at this and commented by email

app/classes/query/modules/titles.rb
nimmolo: Late to the game, but looks good to me. :)

JoeCohen:
> @JoeCohen For `class_name_minus_model_name`, is [`demodulize` or maybe `deconstantize`](https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-demodulize) what you want?

Good point. `demodulize` is more Rai-lish (but less Ruby-ish) than `sub(/^.*::/, "")`. 
Bonus: you made me think about this further: it can be more transparent by also using `delete_prefix`:
`self.class.to_s.demodulize.delete_prefix(model.to_s.remove("::"))`
instead of
`self.class.to_s.sub(/^.*::/, "")[model.to_s.remove("::").length..-1]`
I'll create a Pivotal story as a reminder. (I might not get to it for a while.)

PR #627, opened 2020-07-29
bootstrap4 SupportController

JoeCohen:
Turn SupportControllerTest green.

Delivers https://www.pivotaltracker.com/story/show/174070984
This controller could use more work to make it look more like routine controller. 
But that's part of another Pivotal story, https://www.pivotaltracker.com/story/show/174077354.

### DIFF
--- a/app/helpers/support_helper.rb
+++ b/app/helpers/support_helper.rb
@@ -2,12 +2,13 @@
 
 # helpers for Support controller and views
 module SupportHelper
-  def init_navbar(links = nil )
+  def init_navbar(links = nil)
     { title: { title: :SUPPORT.t, url: support_donate_path }, links: links }
   end
 
   def create_donation_navbar
-    init_navbar(default_links <<
+    init_navbar(
+      default_links <<
         { title: :review_donations_tab.t, url: support_review_donations_path,
           icon: "fa-clipboard-list" }
     )
--- a/app/helpers/support_helper.rb
+++ b/app/helpers/support_helper.rb
@@ -6,7 +6,18 @@ module SupportHelper
     { title: { title: :SUPPORT.t, url: support_donate_path }, links: links }
   end
 
-  def donors_navbar
+  def create_donation_navbar
+    init_navbar([
+      { title: :donors_tab.t, url: support_donors_path,
+        icon: "fa-user-friends" },
+      { title: :donate_tab.t, url: support_donate_path,
+        icon: "fa-hands-helping" },
+      { title: :review_donations_tab.t, url: support_review_donations_path,
+        icon: "fa-clipboard-list" }
+    ])
+  end
+
+  def donate_navbar
     links = [
       { title: :donors_tab.t, url: support_donors_path,
         icon: "fa-user-friends" }
@@ -20,6 +31,10 @@ module SupportHelper
       ]
     end
     init_navbar(links)
+  end
+
+  def donors_navbar
+    donate_navbar
   end
 
   def default_links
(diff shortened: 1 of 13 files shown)


JoeCohen:
> @JoeCohen This is great!! Thank you for writing it!
> 
> Question: Do you think the ivars `@title` and `@navbar` should be defined by the controller action, instead of at the top of the view? It seems more in sync with separation of concerns, maybe.

Very good question.
It's logical to set all the ivars in the controller, and I was thinking about doing that with `@navbar`.
The things that made me hesitate were:
- I wanted to keep the controllers as thin as possible.
- These particular ivars felt like they were focused on the views, as opposed to something which depends on something in the model.

I don't know which way to go on this. I wish I had experience with other Rails apps. How do they handle things like this?

nimmolo:
I see. What if it's set in the helper?
Each view-helper method could end in an action that sets those ivars.

This is another motivation to prune redundant routes, so we don't have to make so many of these.

JoeCohen:
I'm probably just being dense, but I don't understand what you mean by:
>Each view-helper method could end in an action that sets those ivars.

Could you give an example?

nimmolo:
I can add an example to this PR when i get home this evening, if you like!

JoeCohen:
Super!

nimmolo:
Sorry Joe, my idea was an unnecessary detour. It doesn't work. Neither the `@navbar` nor the `@title` can be supplied by the controller. (It's a bit confusing to me why it can be supplied by the view, and not the controller, which is supposedly upstream of the view.) 

However the main thing is, with your helper pattern the view is decluttered, and much improved. I'll need to modify the navbar layout so that it looks better with just a few links, later. I vote to merge your work.

The only thing I did is remove a bug i had introduced in the donate title (it doesn't need to be `humanize`d).

JoeCohen:
Thanks for digging into this, and for the donate title fix.

I'm also surprised by the fact that the ivars cannot be supplied by the
controller. I may look at that some more.

JoeCohen:
Looks to me like the problem with defining @navbar & @title in the controller:
- Happens only if the controller is doing an implicit render. I.e., if the controller explicitly renders the view, no problem.
This makes me think it's a Rails bug.
- Happens only when there's a partial with a form && the partials is called with
 `render "<partial>", <key>: <value>` e.g., `render "form", button_name: :SAVE`
as opposed to the longer:
`partial: <partial>, locals: {  <key>: <value> }`, e.g., `render partial: "form", locals: { button_name: :SAVE }`

So if you really want to do this in the controller, I think I need to be careful how I call partials.

To make sure that I'm correct about the above, I'm going to try it in the PublicationsController in https://github.com/MushroomObserver/mushroom-observer/pull/628

JoeCohen:
1. This will confirm that @title and @navbar **can** be assigned in the controller. I've done it that way in PR #628.
2. I _slightly_ prefer assigning in the template and computing the `navbar` in a helper, mostly because that lightens the controller. Also:
- My old edition of AWDR (4th ed. 2011, for RoR 3.1) has an example where `@title` is used in the layout and assigned in the template. 
- Maybe in the long run the navbar should be computed as a PORO, rather than in the controller or a helper.) 

But my preferences aren't strong, so take your pick.

JoeCohen:
I will merge both this and PR #628 -- despite the different ways of handling the ivars. 
The merger will make bootstrap4 a slightly better blueprint for changes to `master` and will let me accept the Pivotal stories and delete my local branches.

PR #628, opened 2020-07-31
Fix PublicationsController

JoeCohen:
- Turn PublicationsControllerTest green
- Other improvements, particularly decluttering the view by moving the setting of ivars `@title` and `@navbar` to the controller.


JoeCohen:
In this PR the controller computes and assigns `@title` and `@navbar`.
Contrast to PR #627, where computation is in a helper, and assignment is in the templates.
We should be consistent about this, so one of these PRs should be changed.

PR #629, opened 2020-08-02
Normalize ArticleController

JoeCohen:
Updates controller to Rails 5 standards. See https://www.pivotaltracker.com/epic/show/4622322
Delivers https://www.pivotaltracker.com/story/show/174134084


JoeCohen:
Thanks Nimmo!

1. It's great to have this input now; it informs an issue that I've just
begun to think about for this controller -- action and route pruning. Of
course, they're both things which you raised in connection with
bootstrap-4, but which I generally ignored. I should have paid attention
earlier.

2. This controller assigns from only one parameter at a time. Therefore no
mass assignment; we don't use of the private method that encapsulates a
list of mass-assignable params; the method is not, and cannot be, tested;
Coveralls complains that the method is uncovered.

pellaea:
Nimmo -- Not sure if Joe's explanation cleared everything up for you
concerning whitelisting parameters.  My understanding (and please correct
me if I'm wrong or missing something important, Joe!) is that Rails at
least used to encourage "lazy" programmers to do things like

def update_observation
  @observation = Observation.find(params[:id])
  @observation.attributes = params[:observation]
  @observation.save
end

This lets malicious agents add unexpected parameters to
params[:observation], which then get saved.  You could potentially change
ownership of an observation, or maybe worse.

Whitelisting is Rails's way to ensure that only the parameters you expect
are coming through from the form.

-Jason

nimmolo:
@pellaea Thanks Jason--- however i think i'm still confused!!  

From what you said, it still seems like maybe `:id`, `:by` and `:q` need to be whitelisted...?? or are those general params whitelisted by Rails? I couldn't find them whitelisted in MO anywhere.

> Nimmo -- Not sure if Joe's explanation cleared everything up for you concerning whitelisting parameters. My understanding (and please correct me if I'm wrong or missing something important, Joe!) is that Rails at least used to encourage "lazy" programmers to do things like def update_observation @observation = Observation.find(params[:id]) @observation.attributes = params[:observation] @observation.save end This lets malicious agents add unexpected parameters to params[:observation], which then get saved. You could potentially change ownership of an observation, or maybe worse. Whitelisting is Rails's way to ensure that only the parameters you expect are coming through from the form.
> […](#)

pellaea:
I don't think whitelisting is automagical.  We call the method ourselves
where it is needed.  Here's an example rom collection_number_controller:

@collection_number.attributes = whitelisted_collection_number_params

The parameters "id", "by", "q" are not used to set attributes of
@collection_number, so don't need to "whitelist" them.

app/controllers/application_controller.rb
```diff
@@ -1014,17 +1014,28 @@ def query_params(query = nil)
...
+    if !browser.bot?
+      params[:q] = get_query_param(query)
+    end
+    params
+  end
+  helper_method :add_query_param
+
+  # Allows us to add query to a path helper:
+  #   object_path(@object, q: get_query_param)
+  def get_query_param(query = nil)
```
JoeCohen, line 1026: `ApplicationController#get_query_param` is a bit of @nimmolo's magic from `bootstrap-4`. 
He has some other magic there which probably is worth porting -- e.g., methods to literally use `object_path(@object, q: get_query_param)`. 
I'm not sure where all of them are.
I also would be more comfortable if there were a reasonable way to test  ApplicationController.

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,196 @@
...
+#
+#
+#  Callbacks and Methods
+#
+#    ignore_request_unless_permitted:: Unless user permitted to perform request,
+#                       then index_articles
+#    permitted?         boolean: permitted to create/update/destroy Articles
+#
+class ArticlesController < ApplicationController
+  before_action :login_required, except: [
```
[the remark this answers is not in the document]
JoeCohen: This offense may go away on its own: 
- I may get rid of aliases;
- As @nwilson-eol suggested, it may be unnecessary to includes the aliases in the callbacks.

app/views/articles/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_with model: @article do |f| %>
+
+  <div class="form-group push-down">
+    <%= f.label(:article_title, :article_title.t + ":") %>
```
JoeCohen, line 4:
```suggestion
    <%= f.label(:article_title, "#{:article_title.t}:") %>
```

JoeCohen: There's one less assignment with the suggestion.

app/views/articles/_form.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <%= f.text_field(:title, class: "form-control") %>
+    <div class="help-block">
+      <%= :form_article_title_help.t %>
+      <%= :field_textile_link.t %>
+    </div>
+  </div>
+
+  <div class="form-group push-down">
+    <%= f.label(:article_body, :article_body.t + ":") %>
```
JoeCohen, line 13:
```suggestion
    <%= f.label(:article_body, "#{:article_body.t}:") %>
```

app/views/articles/_object.html.erb
```diff
@@ -1,6 +1,8 @@
 <div class="list-group-item">
-  <%= content_tag(:big, link_with_query(object.title.t,
-                  object.show_link_args)) %>
+  <%= content_tag(
+        :big,
+        link_to(object.title.t, article_path(object, q: get_query_param))
```
JoeCohen, line 4: I think that in `bootstrap-4` @nimmolo modified `link_with_query` to work with helper paths. 
I think he also made it work with `object_path` (as opposed to `article_path`). 
If so, should those be ported to `master`? But see my comments above above not being able to test ApplicationController.

app/views/shared/_nav_left.html.erb
```diff
@@ -73,7 +73,7 @@
   <%= link_to(:HERBARIA.t, {controller: :herbarium, action: :index}, {class: "list-group-item indent"}) %>
   <%= link_to(:LOCATIONS.t, {controller: :location, action: :list_locations}, {class: "list-group-item indent"}) %>
   <%= link_to(:NAMES.t, {controller: :name, action: :observation_index}, {class: "list-group-item indent"}) %>
-  <%= link_to(:NEWS.t, {controller: :article, action: :list_articles}, {class: "list-group-item indent"}) %>
+  <%= link_to(:NEWS.t, articles_path, {class: "list-group-item indent"}) %>
```
JoeCohen, line 76: Nice example of the decluttering effect of helper paths.

app/views/shared/_nav_left.html.erb
```diff
@@ -73,7 +73,7 @@
...
   <%= link_to(:PROJECTS.t, {controller: :project, action: :list_projects}, {class: "list-group-item indent"}) %>
 
   <div class="list-group-item disabled bold">
```
JoeCohen, line 79: Can this be further DRYed/de-cluttered by iterating over a list of [title, href] pairs?

routes_list.txt
```diff
@@ -0,0 +1,874 @@
+                                            Prefix Verb     URI Pattern                                                                              Controller#Action
```
JoeCohen, line 1: Mistakenly added to version control and committed.
Revert this.

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,199 @@
+# frozen_string_literal: true
```
JoeCohen, line 1: I should have renamed `article_controller` instead of creating a new file.
- Renaming allows a better diff (at least in GitHub). 
- Also one less chore (deleting the old file).

test/controllers/articles_controller_test.rb
```diff
@@ -3,28 +3,86 @@
...
+  def test_list_articles
+    get(:list_articles)
+    assert(:success)
+  end
+
+  def test_show
+    # Prove that an actual article gets shown
+    get(:show, id: articles(:premier_article).id)
+    assert_response(:success)
+    assert_template(:show)
```
JoeCohen, line 29: Add a test of content here, instead of, or in addition to the template test.
Probably test that <body> include the full article text.

test/models/abstract_model_test.rb
```diff
@@ -395,4 +395,37 @@ def test_species_list_rss_log_life_cycle
...
   end
+
+  # -------------------------------------------------------------------
+  #  Test the methods that generate controller and action names
+  #  based on the Query model
+  # -------------------------------------------------------------------
+
+  def test_show_controller
+    assert_equal("articles", Article.show_controller)
+    assert_equal("#{self.class.name.underscore}/phony", Phony.show_controller)
```
JoeCohen, line 406: Is there a better way to test for a controller with a singular name?
I didn't want to use any of the current actual controllers because they may be normalized, which would break the test.

app/views/shared/_nav_left.html.erb
nimmolo: Definitely. bootstrap-4 calls a partial for each item, but here you could just make an array and then iterate right in the view.

app/views/articles/_object.html.erb
nimmolo: @JoeCohen I did add a module, in `models/concerns/routeable.rb` that makes the path helpers available anywhere when included. But I believe `object_link_helper` ended up not needing it. 

The module contains comments with links to articles explaining how to use it. (let's keep those comments for now!)

app/controllers/application_controller.rb
nimmolo: Those were added in `helpers/object_link_helpers`.
See below for reference to the `routeable` module.

I feel like testing the methods that depend on methods in application_controller might be coverage enough...

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,199 @@
...
+        title: params[:article][:title],
+        body: params[:article][:body],
+        user_id: @user.id
+      )
+      @article.save
+      redirect_to article_path(@article.id)
+    end
+  end
+
+  def update
```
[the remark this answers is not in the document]
nimmolo: @JoeCohen -- What do you do when codeclimate gets picky about ABC? I'm baffled by these, even though i guess i _vaguely_ get what it means. But it would be great if i understood...

app/views/shared/_nav_left.html.erb
JoeCohen: From a coding standpoint, I'd greatly prefer using partials `_nav_left`. Just looking at the current code hurts my eyes. :-) 
But there's a performance question, especially with stuff that appears on nearly every page. I don't have the knowledge and time to investigate that now. And the question I raised is really outside the scope of this PR and story.
I created a new story. https://www.pivotaltracker.com/story/show/174142891
We should at least iterate over an array.

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,199 @@
...
+#    index::           List all articles
+#    index_articles::  List selected (based on last search) articles
+#    new::             Display form for new article
+#    show::            Show one article
+#    update::          Update article from "edit" form
+#
+#
+#  Callbacks and Methods
+#
+#    ignore_request_unless_permitted:: Unless user permitted to perform request,
```
JoeCohen, line 19: Remove this line; it's (now) a private method.
```suggestion
```

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,199 @@
...
+  before_action :store_location
+  before_action :ignore_request_unless_permitted, except: [
+    :index,
+    :index_articles,
+    :list_articles, # aliased
+    :show,
+    :show_article # aliased
+  ]
+
+  ############ - Actions to Display data (index, show, etc.)
```
JoeCohen, line 40: I suggest that all controllers uniformly order actions and methods. (Makes it easier to find things and read the code.)
There isn't an official Rails default ordering. (Generated controllers are empty. See [here](https://guides.rubyonrails.org/getting_started.html#laying-down-the-groundwork).) So I used the same order (indicted by the comments in the file) as AWDR:
1. Actions to Display Data (index, show, etc.)
2. Actions to Present Forms -- (new, edit, etc.)
3. Actions to Modify data: (create, update, destroy, etc.)
4. Public methods (unrouted)
5. private methods
This may also suggest a way to divide the controller into separate files, if that's desirable.

test/controllers/articles_controller_test.rb
```diff
@@ -3,28 +3,86 @@
 require "test_helper"
 
 # Controller tests for news articles
-class ArticleControllerTest < FunctionalTestCase
-  def test_create_article_get
+class ArticlesControllerTest < FunctionalTestCase
+  ############ test Actions that Display data (index, show, etc.)
```
JoeCohen, line 7: As with the controller, I suggest ordering test methods uniformly, and in a way that mirrors the order of controller methods.

test/controllers/articles_controller_test.rb
```diff
@@ -3,28 +3,86 @@
...
+
+  def test_index
+    # Prove any user can see article index
+    get(:index_articles)
+    assert(:success)
+
+    # Prove privileged user get link to create an article
+    login(users(:article_writer).login)
+    get(:index_articles)
+    assert_select("a", text: :create_article_title.l)
```
JoeCohen, line 17: Replace comment with a message. The message provides information both when reading the code and if the test fails. (I don't regard this suggestion as part of controller normalization.)
```suggestion
    login(users(:article_writer).login)
    get(:index_articles)
    assert_select("a", text: :create_article_title.l, "Privileged user should have link to create and article")
```

app/controllers/articles_controller.rb
JoeCohen, line 106: Should we comment each actions with the path, AWDR style? E.g.:
```ruby
  # PATCH/PUT /articles/1
  def update
```

test/controllers/articles_controller_test.rb
```diff
@@ -3,28 +3,86 @@
...
+  end
+
+  def test_list_articles
```
JoeCohen, line 20: `test_list_articles` will probably go away.
The aliases will be replaced by routes which re-direct, and there will be a test. It will test the routes. It might also test that redirects actually happen, but my instinct is that this is testing Rails, rather than MO.

test/controllers/articles_controller_test.rb
```diff
@@ -3,28 +3,86 @@
...
 # Controller tests for news articles
-class ArticleControllerTest < FunctionalTestCase
-  def test_create_article_get
+class ArticlesControllerTest < FunctionalTestCase
+  ############ test Actions that Display data (index, show, etc.)
+
+  def test_index
+    # Prove any user can see article index
+    get(:index_articles)
+    assert(:success)
```
JoeCohen, line 12: 1. There's a bug here and in line 23.
```suggestion
    assert_response(:success)
```
2. Should have a test for content.
Contrary to the comment in the code, this does not prove that the user sees anything. There's a similar misleading comment in the [RailsGuide](https://guides.rubyonrails.org/testing.html#what-to-include-in-your-functional-tests) about `assert_response(:success)`. ("ensur[es]that the right response body has been generated.")

test/controllers/articles_controller_test.rb
```diff
@@ -35,53 +93,36 @@ def test_create_article_post
 
     # Prove unauthorized user cannot create Article
     login(users(:zero_user).login)
-    post(:create_article, params)
+    post(:create, params)
     assert_flash_text(:permission_denied.l)
     assert_equal(old_count, Article.count)
```
JoeCohen, line 98:
```suggestion
    assert_difference("Article.count") do
      post(:create, params)
    end
```
Here and in similar assertions.

test/controllers/articles_controller_test.rb
```diff
@@ -91,86 +132,53 @@ def test_edit_article_post
...
-    assert_redirected_to(action: :show_article, id: article.id)
+    assert_redirected_to(article_path(article.id))
 
     # Prove removing title provokes warning
     params[:article][:title] = ""
-    post(:edit_article, params)
+    post(:update, params)
     assert_flash_text(:article_title_required.l)
-    assert_template(:edit_article)
+    assert_template(:edit)
```
JoeCohen, line 162: Replace with `assert_select` that tests for presence of form that posts to `:edit`.

app/views/articles/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_with model: @article do |f| %>
+
+  <div class="form-group push-down">
+    <%= f.label(:article_title, :article_title.t + ":") %>
+    <%= f.text_field(:title, class: "form-control") %>
```
JoeCohen, line 5: Should we parenthesize/unparenthesize some/all arguments in the views? Do we care?
- We generally parenthesize.
- No parens is less cluttered, and various Rails documentation has plenty of examples without parens. 
- RuboCop's default for `.rb` files is parens. [Style/MethodCallWithArgsParentheses](https://docs.rubocop.org/rubocop/0.85/cops_style.html#stylemethodcallwithargsparentheses). RuboCop doesn't check `.erb` files. 
- In `bootstrap-4`, @nimmolo did not use parens in views (also in many places in the tests and the app). In `bootstrap4` I tried avoiding parens when I edited the views, but often forgot about that. It's hard to switch back and forth.

pellaea:
I personally think we should be consistent.

I think the ruby / rails community consensus is no parens.  But we have a
*lot* of parens throughout our code.

I personally prefer parens because I think it makes it easier to read.  For
example, some constructs are ambiguous without parens:

method1 arg1, arg2 && method2

Which of these two is it equivalent to?

1. method1 arg1, (arg2 && method2)
2. method1(arg1, arg2) && method2

I think rubocop has us using (2) frequently, necessitating parens at least
some of the time.

I also personally like to be able to use "%" in vi to jump between opening
and closing parents, brackets, etc.  This is obviously not possible if you
don't use parens.

JoeCohen:
> @JoeCohen (This may be a premature comment, since i realize you're just trying to get coverage to 100% at this point.)
> 
> I realize I don't really understand when parameters need to be whitelisted or not. But in the process of CRUDification that i was working on with Sequences in bootstrap-4, removing non-CRUD redundant routes meant relying on params more - there were new params for `:prev` and `:next` for example, in that controller, to route from the `show` action. But not here of course.
> 
> In any case, for that style of CRUDified index action in articles, there wouldn't be any new params:
> 
> ```
>   def index
>     store_location
>     if params[:q] || params[:by]
>       index_article
>     else
>       query = create_query(:Article, :all)
>       show_selected_articles(query)
>     end
>   end
> 
>   alias :list_articles :index
> ```
> 
> All this does is allow you to remove the route `articles_index_article_path` -- it is redundant as long as you can use `articles_path(q: get_query_param)` or `articles_path(by: :user)`.
> 
> The difference is that `articles_path` with no params gets you a fresh index, while with the param it sends you to the `index_article` action.

Re eliminating `Index_article` action.
@nimmolo's suggestion should work for `articles` but **Is it an example how to do things in other controllers?**
Could the user get into a loop (particularly with Query-able models).
I.e., user creates a Query, then wants to see the entire index.
Will the suggested conditional, `if params[:q] || params[:by]` always prevent MO from returning the full index? 
If so, could we work around it? E.g, with links that explicitly set :q and :by to `nil`?

nimmolo:
Maybe we need to expressly set q to nil for the index action (this should work without deleting the stored query!)

JoeCohen:
> ~Maybe we need to expressly set q to nil for the index action (this should work without deleting the stored query!)~
> 
> Actually @JoeCohen I'm almost certain this pattern will work for all controllers.
> 
> It's a question of how the link is built in the view. A `link_to articles_path` or an analagous path in any other controller, with no params, will definitely take you through to the full index (with the conditional quoted above in the code block), because **any param not explicitly included in a path helper will not be included**. The `:index` controller action will not find a `q:` or a `by:` param.
> 
> Other links are intended to take you to an index with a query, and the links are built that way in the views -- for example, many views have something like `link_with_query({ controller: :foo, action: :index_foo })`.
> 
> For that case, if we're using paths, we've got `q: get_query_param`, so we can add the query back in to the path as a param, such as `link_to articles_path(q: get_query_param)` or `link_to articles_path(by: sort_by)`. Only with that kind of syntax will the path send the param.
> 
> So in observations, for example, the `:index` action could have many more param checks sending the request to any of the different indexing actions, through the same `observations_path` route. `link_to({ controller: :observations; action: :observations_by_user })` could be refactored as `link_to observations_path(by_user: user_id)`. This seems much better than `link_to observations_observations_by_user_path(user_id: user_id)`.
> 
> Params, as i picture it, basically should allow us to sort requests to multiple actions through a tightly pruned network of routes.

Thank you Nimmo. This is extremely helpful. 
It made me realized that some links are the production version are wrong (IMO), and it explains how to fix them.

JoeCohen:
Thanks! Kind of follows naturally from what you suggested.

nimmolo:
Whoa. Zero issues????!!

app/views/shared/_nav_left.html.erb
nimmolo: I'd agree with @pellaea that if partials are more pleasing to code, we should not worry about the relative performance of them. I would agree, just going from analagous situations in other languages, that the performance hit is almost certainly pretty trivial. But we could benchmark it via a separate PR - why not?

(The biggest hits to MO performance that I found working on Bootstrap 4 are the expensive queries for name, naming and location that are run on observation#show. I believe having the production server running Ubuntu 16 is a considerable factor too).

config/routes.rb
```diff
@@ -493,6 +495,60 @@
   lookup_user
 ].freeze
 
+def redirect_old_crud_actions(
+  old_controller: "",
+  new_controller: old_controller&.pluralize,
+  model: new_controller&.singularize,
+  actions: [:create, :destroy, :edit, :index, :show]
+)
+  return if old_controller.blank?
```
JoeCohen, line 504: Might be better to throw an error than silently return.

test/controllers/articles_controller_test.rb
```diff
@@ -3,85 +3,143 @@
...
+    article = Article.first
+    query = Query.lookup(:Article, :in_set, ids: [article.id])
+    params = @controller.query_params(query)
+    get(:index, params: params)
+    assert_select("a:match('href',?)", %r{/articles/\d+}, { count: 1 },
+                  "filtered Article Index has wrong number of entries")
+    assert_select(
+      "a[href *= '#{article_path(article.id)}']", true,
+      "filtered Article Index missing link to #{article.title} (##{article.id})"
+    )
```
JoeCohen, line 28: Are these `assert_select`s un-performant? If so, what should I do?

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+  # articles  GET /articles(.:format)
+  def index
+    filter_index? ? index_filtered : index_full
+  end
+
+  # article  GET /articles/:id(.:format)
+  def show
+    return false unless (@article = find_or_goto_index(Article, params[:id]))
+
+    @canonical_url = article_url(@article.id)
```
JoeCohen, line 41: By coincidence, it's the same string as the old canonical_url. 
That's not necessarily the case with other controllers, and -- as @pellaea oints out -- might get the bots all excited.

test/models/query_test.rb
```diff
@@ -1340,8 +1340,13 @@ def test_article_pattern_search
...
                  :Article, :pattern_search, pattern: "premier_article")
     # body
-    assert_query(Article.where("title LIKE '%Body of Article%'
-                               OR body LIKE '%Body of Article%'"),
+    expect = Article.where("title LIKE ? AND title LIKE ? AND title LIKE ?",
+                           "%Body%", "%of%", "%Article%").
+             or(
+               Article.where("body LIKE ? AND body LIKE ? AND body LIKE ?",
+                             "%Body%", "%of%", "%Article%")
+             )
```
JoeCohen, line 1348: Fixes a bug (of mine) which was exposed when I changed the articles fixtures.

JoeCohen:
I think the PR is ready for review, though I'd like to do more work (E.g., grabbing @nimmolo's `object_link` module.)
I'd appreciate any comments, but may be out of commission until Friday.

JoeCohen:
That's phony. It means 0 new issues.
There are hundreds of RuboCop issues hiding in .rubocop_todo.yml, many of
them difficult ABC or other metric issues.
And there likely are issues that we've completely suppressed from other
engines used by Codeclimate.

But the point is not to merge into master if there are any new issues.

config/routes.rb
```diff
@@ -98,13 +98,15 @@
...
+  articles: {
+    # resources
+    # create: {},
+    # destroy: {},
+    # edit: {},
+    # index: {},
+    # new: {},
+    # show: {},
+    # update: {}
   },
```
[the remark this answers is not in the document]
nimmolo: Agree. That was my intent with that list.

config/routes.rb
```diff
@@ -493,6 +495,60 @@
   lookup_user
 ].freeze
 
+def redirect_old_crud_actions(
```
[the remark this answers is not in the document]
nimmolo: Is this method intended to redirect any request that gets replaced by the new resource/CRUD routes?

I would suggest adding `:search_#{model}` routes to the `:index` case - they are essentially a filtered index route that can be easily parameterized.

JoeCohen:
Thanks for the suggestion. I'll have a look.

config/routes.rb
```diff
@@ -493,6 +495,60 @@
...
+            via: [:get, :post] # new, create
+    when :edit
+      match "/#{old_controller}/edit_#{model}/:id",
+            to: redirect("/#{new_controller}/%{id}/edit"),
+            via: [:get, :post] # edit, update
+    when :destroy
+      match "/#{old_controller}/destroy_#{model}",
+            to: redirect("#{new_controller}/%{id}"),
+            via: [:patch, :post, :put]
+    end
```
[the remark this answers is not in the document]
mo-nathan: Are you familiar with Martin Fowler’s book “Refactoring”?  He’s pretty much the de facto authority on refactoring.  He has this to say about switch/case statements in the current edition:

“ Repeated Switches

Talk to a true object-oriented evangelist and they’ll soon get onto the evils of switch statements. They’ll argue that any switch statement you see is begging for Replace Conditional with Polymorphism (272). We’ve even heard some people argue that all conditional logic should be replaced with polymorphism, tossing most ifs into the dustbin of history. Even in our more wild-eyed youth, we were never unconditionally opposed to the conditional. Indeed, the first edition of this book had a smell entitled “switch statements.” The smell was there because in the late 90’s we found polymorphism sadly underappreciated, and saw benefit in getting people to switch over. These days there is more polymorphism about, and it isn’t the simple red flag that it often was fifteen years ago. Furthermore, many languages support more sophisticated forms of switch statements that use more than some primitive code as their base. So we now focus on the repeated switch, where the same conditional switching logic (either in a switch/case statement or in a cascade of if/else statements) pops up in different places. The problem with such duplicate switches is that, whenever you add a clause, you have to find all the switches and update them. Against the dark forces of such repetition, polymorphism provides an elegant weapon for a more civilized codebase.”

However, in this case using polymorphism seems a bit crazy.  If there was more symmetry between the cases such that data was the only difference I’d consider using a hash, but I don’t see a simple way to do that so you might be stuck with a case statement.

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+  before_action :pass_query_params, except: :index
+  before_action :ignore_request_unless_permitted, except: [
+    :index,
+    :show
+  ]
+
+  # ---------- Actions to Display data (index, show, etc.) ---------------------
+
+  # List selected articles, filtered by current Query.
+  # articles  GET /articles(.:format)
```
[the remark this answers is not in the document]
mo-nathan: I don't know what you mean by AWDR style comments and Google is not being helpful.  Personally I find the `GET /articles(.:format)` bit helpful, but the rest strikes me as noise.

app/controllers/articles_controller.rb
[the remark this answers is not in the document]
mo-nathan: This appears to be resolved.

app/models/abstract_model.rb
```diff
@@ -306,28 +306,59 @@ def formatted_errors
...
   #
-  #   Name.index_action => "index_name"
+  #   Article.index_action => "index" # normalized controller
+  #
+  #   Name.index_action => "index_name" #unormalized
   #   name.index_action => "index_name"
   #
   def self.index_action
-    "index_" + name.underscore
+    if controller_normalized?(name)
```
mo-nathan, line 357: Not a blocker, but I think I'd write this as:
```
def self.index_action
  return "index" if controller_normalized?(name) # Rails standard

  "index_" + name.underscore # Old MO style
end
```

app/models/abstract_model.rb
```diff
@@ -337,11 +368,17 @@ def index_action
...
   #
-  #   Name.show_action => "show_name"
+  #   Article.show_action => "show" # normalized controller
+  #
+  #   Name.show_action => "show_name" #unormalized
   #   name.show_action => "show_name"
   #
   def self.show_action
-    "show_" + name.underscore
+    if controller_normalized?(name)
```
mo-nathan, line 377: Same comment as above.

app/models/abstract_model.rb
```diff
@@ -350,11 +387,19 @@ def show_action
...
+  #   # normalized controller
+  #   Article.index_action => "http://mushroomobserver.org/article/12"
+  #
+  #   # unnormalized controller
   #   Name.show_url(12) => "http://mushroomobserver.org/name/show_name/12"
   #   name.show_url     => "http://mushroomobserver.org/name/show_name/12"
   #
   def self.show_url(id)
-    "#{MO.http_domain}/#{show_controller}/#{show_action}/#{id}"
+    if controller_normalized?(name)
```
mo-nathan, line 398: And again.

app/models/abstract_model.rb
```diff
@@ -306,28 +306,59 @@ def formatted_errors
...
   #
-  #   Name.show_controller => "name"
+  #   Article.show_controller => "articles" # for normalized controller
+  #
+  #   Name.show_controller => "name" # unnormalized controller and special cases
   #   name.show_controller => "name"
   #
   def self.show_controller
-    name.underscore
+    if controller_normalized?(name)
```
mo-nathan, line 324: Again, I'd use the `return X if foo; return Y` idiom.

config/routes.rb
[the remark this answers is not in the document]
mo-nathan: Looking at this more closely, I think this could be done with a hash.  That maps an action to a function that takes new_controller, old_controller, model, and id.  The code of each `when` are the different functions that go in the hash.

test/controllers/articles_controller_test.rb
mo-nathan: I'd time it with the assert_selects and without.  If the different is less than a second or less than 10% of the run time for the rest of the test I wouldn't sweat it.  I'm guessing the `get` dominates the time it takes to run the test.

test/controllers/articles_controller_test.rb
```diff
@@ -3,85 +3,143 @@
 require "test_helper"
 
 # Controller tests for news articles
-class ArticleControllerTest < FunctionalTestCase
-  def test_create_article_get
+class ArticlesControllerTest < FunctionalTestCase
+  ############ test Actions that Display data (index, show, etc.)
+
+  def test_index
```
mo-nathan, line 9: I would split this test up into three separate tests - one for each `get`.

app/controllers/articles_controller.rb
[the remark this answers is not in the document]
JoeCohen: Thanks for the feedback. I'll ditch them all as either noise or only marginally helpful.
(AWDR = [Agile Web Development with Rails](https://pragprog.com/titles/rails51/).)

config/routes.rb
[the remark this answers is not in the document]
JoeCohen: I'll try doing this with a hash.
(I own Refactoring Ruby Edition (co-authored by Fowler) and consulted it when I ran into this problem. Same result: it also suggests polymorphism, which also seemed extreme to me (and I wasn't sure how to do it.)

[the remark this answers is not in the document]
JoeCohen: I don't understand what the hash key would do. I should be able to define a method (`redirect_<action>`) for each action, then iterate over the existing hash of actions.

config/routes.rb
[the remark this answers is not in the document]
JoeCohen: I'm doing the redirection for backward compatibility. I don't think it can replace the dynamically built links. We'll have to catch those by passing in the new controller.

I like the idea of automatically redirecting the `:search<model>` routes. 
It will require a separate `when` (or equivalent -- I've now replaced the `case` statement. 
I'd rather add this with the first controller that has an explicit `search` action, as it will be possible to at least test manually there.

app/models/abstract_model.rb
JoeCohen: I changed the other ones, but left this as-is because
`return "#{MO.http_domain}/#{show_controller}/#{id}" if controller_normalized?(name)`
won't fit on one line (and would offend RuboCop).

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+    args = {
+      action: :index,
+      letters: "articles.title",
+      num_per_page: 50
+    }.merge(args)
+
+    @links ||= []
+
+    # Add some alternate sorting criteria.
+    args[:sorting_links] = [
```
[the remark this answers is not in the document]
JoeCohen: I like putting them into `args =`. 
I don't remember why I moved them out of their own function.

app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,180 @@
...
+  def create
+    return render(:new) if flash_missing_title?
+
+    @article = Article.new(
+      title: params[:article][:title],
+      body: params[:article][:body],
+      user_id: @user.id
+    )
+    @article.save
+    redirect_to article_path(@article.id)
```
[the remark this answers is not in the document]
JoeCohen: I agree, and thanks for noticing the problem. 
I thought things worked that way, but I misread the RuboCop documentation. Parens around method calls is the default style, but I didn't notice that [Style/MethodCallWithArgsParentheses](https://docs.rubocop.org/rubocop/0.86/cops_style.html#stylemethodcallwithargsparentheses) is not enabled by default. 
I will fix.

JoeCohen:
Re `assert_select` speed. See https://github.com/MushroomObserver/mushroom-observer/pull/629#discussion_r467488824
I timed one test method with and without `assert_select`s. (The method has 2 `assert_select`s.) (I hope I'm doing it right.)
I don't see a difference. 
```sh
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # with assert_select's
real	0m19.005s
user	0m6.705s
sys	0m3.690s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # with assert_select's
real	0m14.645s
user	0m6.599s
sys	0m3.450s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # with assert_select's
real	0m14.453s
user	0m6.542s
sys	0m3.426s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # with assert_select's
real	0m14.693s
user	0m6.581s
sys	0m3.587s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # without
real	0m14.362s
user	0m6.629s
sys	0m3.402s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # without
real	0m13.961s
user	0m6.655s
sys	0m3.408s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # without
real	0m15.471s
user	0m6.618s
sys	0m3.347s
vagrant@vagrant /vagrant/mushroom-observer (174134084-normalize-ArticleController)
$ time rails t test/controllers/articles_controller_test.rb:20 > /dev/null # without
real	0m14.847s
user	0m6.506s
sys	0m3.478s
```

pellaea:
Looks right.  Note that the first test should be thrown away.  See how it
is significantly longer than the rest?  Something about system caching,
presumably.

JoeCohen:
Jason: Thanks for the fast feedback

test/controllers/articles_controller_test.rb
mo-nathan: I don't see `test_index` getting broken up.  Was this intentional?

JoeCohen: Thanks for taking the time to review this again.
I think GitHub's incomplete file display obscures some changes.
I did break the former `test_index` into 3 tests, 1 get/test:
`test_index`, `test_index_filtered`, `test_index_links_to_creates`
It might be easiest to see here: https://github.com/MushroomObserver/mushroom-observer/pull/629/commits/0fe65e28ebb4c4d3ccfc6c8f22671576803281b1

mo-nathan: Ah, I see it now.  Sorry about that.

config/routes.rb
```diff
@@ -493,6 +485,82 @@
...
+
+  actions.each do |action|
+    send("redirect_#{action}", old_controller, new_controller, model)
+  end
+end
+
+# rubocop:disable Style/FormatStringToken
+# For consistency with Rails Guide "Rails Routing from the Outside In"
+# (And I don't know if RuboCop's suggestion works here)
+def redirect_create(old_controller, new_controller, model)
```
mo-nathan, line 512: This isn't quite what I was going for.  I'm working on it and I'll have a more concrete example shortly.

config/routes.rb
```diff
@@ -493,6 +485,82 @@
...
+end
+
+def redirect_edit(old_controller, new_controller, model)
+  match("/#{old_controller}/edit_#{model}/:id",
+        to: redirect("/#{new_controller}/%{id}/edit"),
+        via: [:get, :post]) # edit, update
+end
+
+def redirect_destroy(old_controller, new_controller, model)
+  match("/#{old_controller}/destroy_#{model}",
```
mo-nathan, line 525: Why doesn't this have `/:id`?

test/integration/redirects_test.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+class RedirectsTest < IntegrationTestCase
```
JoeCohen, line 5: @mo-nathan:
1. I assume you prefer 1 get one per method.
2. Have I overdone the number of tests: 1 test  / old action / request method? (Exception: destroy)

test/integration/redirects_test.rb
```diff
@@ -0,0 +1,58 @@
...
+  def test_controller
+    get("/article")
+    assert_equal("/articles", @response.request.fullpath)
```
JoeCohen, line 8: 1. Should I use literal paths or instead use path helpers here and elsewhere. I.e.
` assert_equal("/articles", @response.request.fullpath)`
vs
` assert_equal(articles_path, @response.request.fullpath)`
I think the literal path is faster, but I couldn't see a difference in a quick-and-dirty `time` test.
I think the path helper is more idiomatic.
2. `fullpath` or `original_fullpath`? (I don't understand the difference.)

test/integration/redirects_test.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def test_show_article
+    get("/article/show_article/#{Article.first.id}")
+    assert_equal("/articles/#{Article.first.id}", @response.request.fullpath)
+  end
+
+  def test_create_article_get
+    login!("article_writer", "testpassword", true)
```
JoeCohen, line 27: I don't see a way to avoid `login!` 
Maybe @mo-nathan can pull another rabbit out of his hat.
I could avoid multiple calls to `login!` by combining all the methods that call it.

test/integration/redirects_test.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def test_edit_article_post
+    login!("article_writer", "testpassword", true)
+    post("/article/edit_article/#{Article.first.id}")
+    assert_equal("/articles/#{Article.first.id}/edit",
+                 @response.request.fullpath)
+  end
+
+  def test_destroy_article_post
```
JoeCohen, line 52: Should we also have `test_destroy_article_patch` and `test_destroy_article_patch`?

test/integration/redirects_test.rb
JoeCohen: PS:
I wonder if a better approach is:
- Run `rails routes`
- Check the output for the expected routes.

That would test only whether `routes.rb` generates the expected redirects, which is what I originally set out to do. It's not dependent on what Rails does with one of the redirected requests. (That can be a little funky, e.g., PATCH|PUT `destroy` vs POST 'destroy`.) No http requests are needed.
On the other hand, it's dependent on how Rails formats the routes.

JoeCohen: I've now implemented that proposal as `test/unit/redirects_test.rb`.
Bonus: on my old MacBook Pro, the unit test is faster by ~20% (4 seconds).

test/unit/redirects_test.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+# Test existence of redirect routes
+class RedirectsTest < UnitTestCase
+  # Test existence of correct routes from MO old-style CRUDISH actions
+  # to normalized, resourceful Crud
+  def test_redirect_old_crud_actions
```
JoeCohen, line 9: Similar (but nonidentical) functionality to `test/integration/redirects_test.rb`
See discussion at https://github.com/MushroomObserver/mushroom-observer/pull/629/files?file-filters%5B%5D=.rb&hide-deleted-files=true#r468752266
I prefer this unit test to the integration test.

test/unit/redirect_routes_test.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+# Test existence of redirect routes
+class RedirectRoutesTest < UnitTestCase
+  # Test existence of correct routes from MO old-style CRUDISH actions
+  # to normalized, resourceful Crud
+  def test_redirect_old_crud_actions
```
JoeCohen, line 9: Similar (but nonidentical) functionality to test/integration/redirects_test.rb
See discussion at https://github.com/MushroomObserver/mushroom-observer/pull/629/files?file-filters%5B%5D=.rb&hide-deleted-files=true#r468752266
I prefer this unit test to the integration test.

test/integration/redirects_test.rb
mo-nathan: Is 4 secs/test or 4 secs overall?

Using the output of a Rails command that is intended for human readability seems a bit fragile to me.  For example, I know when I run `rails routes` at work, I get one line per verb and the Prefix is not repeated (e.g., https://backend.turing.io/module2/lessons/archive/advanced_routing_rails).  I haven't figured out why we aren't seeing that, but it seems equally likely that it's because we're using a more recent release and we're using a more recent release at work.  In any case, I think the point remains.  It also just feels weird to me to be testing the results of a shell command (although I think we're doing that in other places).

JoeCohen: Good point about fragility. I'll stick with the integration test.
(And it's 4 sec. overall. Probably less on other machines.)

PR #631, opened 2020-08-07
Update Travis configuration

JoeCohen:
- Explicitly specify os. (This avoids one Travis warning.)
- Remove `sudo` key. (It is obsolete.)
- Change Ubuntu distribution from Xenial to Bionic.
At the moment, we use Xenial (16.04) in production, Bionic (18.04) in development.
So Travis is guaranteed to be out of sync with one or the other.
I opted to have Travis use Bionic,  as we are planning to upgrade the production environment.
(It also might make Travis run slightly faster.)

There is no manual test script. (But perhaps look at the Travis build file to see if it's as expected.)

Delivers https://www.pivotaltracker.com/story/show/173998390

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
-dist: xenial # use Ubuntu 16.04
+os: linux
+dist: bionic # Ubuntu 18.04 (Bionic Beaver)
 services:
   - mysql
 
 language: ruby
 rvm:
   - "2.6.6"
-
-# make Travis use its legacy infrastructure
-sudo: true
 
 # uncomment this line if your project needs to run something other than `rake`:
 # script: bundle exec rake lang:update


mo-nathan:
Is there a more specific motivation for this change currently?  I don't see one in the ticket.  Personally I'd rather make this change when we are actively working on getting the server to bionic.  I think there is some risk involved in having our CI/CD system using a different OS than our production system.  While I'm in favor of upgrading the production system, I'd rather stay focused on the controller work if there isn't a pressing need.

JoeCohen:
It's not pressing. It was just on my mind, and I was looking for
easy things to do while PR #629 is awaiting review and finalization.
There's also risk in having CI/CD use a different OS than the development
machine.
Let's just leave this unmerged for now.
Would you like me to change it to a Draft PR in order to prevent accidental
merger?

JoeCohen:
I will turn this PR into a Draft so it doesn't get merged until the production server is upgraded to Bionic.

JoeCohen:
Closing PR because the relevant parts are included, or suggested for inclusion, in #693.

PR #634, opened 2020-08-12
Changed translations of "herbarium" to "fungarium" throughout.

pellaea:
But left the database table and classes and such as "herbarium".

### DIFF
--- a/README_API_HELP_MESSAGES.txt
+++ b/README_API_HELP_MESSAGES.txt
@@ -333,7 +333,7 @@ GET observation
   west:	longitude (min longitude)
 
 POST observation
-  accession_number:	string (unique herbarium id)
+  accession_number:	string (unique fungarium id)
   altitude:	altitude
   collection_number:	string
   collectors_name:	string
--- a/config/locales/en.txt
+++ b/config/locales/en.txt
@@ -267,16 +267,16 @@
   glossary_term: glossary term
   GLOSSARY_TERMS: Glossary Terms
   glossary_terms: glossary terms
-  HERBARIUM_RECORD: Herbarium Record
-  herbarium_record: herbarium record
-  HERBARIUM_RECORDS: Herbarium Records
-  herbarium_records: herbarium records
-  herbaria: herbaria
-  HERBARIA: Herbaria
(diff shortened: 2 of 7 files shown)


JoeCohen:
1. Thanks for doing all this. What a pain.
2. I haven't looked at this in detail. If you want me to, I can do that tomorrow.
3. If we want to make an immediate change, this is a good approach. 
In the long run, should we make all those other changes: db table, classes, migration. 
And how about routes, controllers, views? 
Would it work to do model changes be in a separate PR from controller/view?
Would it work to do `herbarium` changes separately from `herbarium_record`?

pellaea:
Yeah, this is intended to be the minimal work required to address FunDiS's
requests.  I just don't have time to make more fundamental changes.  It
probably does make sense to split it as much as possible.  But there *is*
an advantage to doing both herbarium and herbarium_record at the same time:
In principle we can just do a global search and replace if we do both!  But
there are at least three things that we need to do simultaneously:

db migration to rename tables and columns:
  herbaria
  herbarium_curators
  herbarium_curators.herbarium_id
  herbarium_records
  herbarium_records.herbarium_id
  herbarium_records_observations
  herbarium_records_observations.herbarium_record_id

alias / redirect old routes for:
  /herbarium/*
  /herbarium_record/*
  /api/herbarium
  /api/herbarium_record

copy over any translations that have already been done for:
  HERBARIUM_RECORD
  herbarium_record
  HERBARIUM_RECORDS
  herbarium_records
  HERBARIA
  herbaria
  HERBARIUM
  herbarium
  HERBARIUMS
  herbariums
  etc.

There are too many tranlsation tags to list.  What I've done doesn't change
the *tag*, so it won't result in the throwing away of any translations.
But when we change the tags, we need to make sure the old translation for
"herbarium", etc. gets transferred over to "fungarium".  Obviously, we'd
prefer for the original translator to fix their translations, too.  But in
the meantime, having it correctly translated to "herbarium" in their
language is preferable to leaving it "fungarium" in English!  FunDiS
doesn't concern itself with other languages so far as I know.)

We're talking a fair amount of work.

-Jason

PR #635, opened 2020-08-14
Auto-parenthesize method arguments

JoeCohen:
- Uses RuboCop autocorrect to parenthesize all method arguments.
- Fixes the offenses created when I enabled `Style/MethodCallWithArgsParentheses` in  #629, https://github.com/MushroomObserver/mushroom-observer/commit/6cf282066f6b73cc253782acb7a953b43f99bc17.
(In #629, @mo-nathan observed that I was inconsistent in parenthesizing method arguments, said he preferred parens -- as do I and @pellaea; @nimmolo does not -- asked that the cop be turned on.)

**Issue**
Should we whitelist some methods to not require parens? (RuboCop allows that. See https://docs.rubocop.org/rubocop/0.86/cops_style.html#stylemethodcallwithargsparentheses.)
Some candidates:
- `gem`
- `require`
- `raise`
- [ActiveRecord callbacks](https://guides.rubyonrails.org/v5.2/active_record_callbacks.html#available-callbacks)
- [filters (controller callbacks)](https://api.rubyonrails.org/classes/AbstractController/Callbacks/ClassMethods.html#method-i-after_action)


pellaea:
While I generally prefer the nonambiguity and C-ishness of parens, I see
that there are cases where it looks better without.  Some of these methods
that you mention, like active record filters, things like require and alias
-- generally speaking, "directive-like" methods, things that don't actually
"feel" like methods -- look better to me without parens.  But constructs
like

   raise "blah de blah" unless blah

look (slightly) too ambiguous for my tastes.  Maybe you can snip a single
character out of those error messages to make the lines fit! :)

Just to be clear, though, these are just aesthetic preferences. I don't
think any rule, no matter how sophisticated, will work in every case.  My
personal preference would be to use rubocop as a guide, not a "cop".  All
of its metrics and everything else it enforces -- they're all great ideas,
but I have run into exceptions (in my opinion) to every single rule.

Summary: I like consistency, and would prefer greater consistency either
direction -- with or without parens -- to randomness.  (Consistency makes
search-and-replace, and other automated tasks and searches and stuff
easier.)  And if we have to be strict, then so be it, let's be strict.
I'll just trust that there is some long-term dividend that outweighs
aesthetics in those (hopefully rare) cases where the rules don't make
sense.  I can live with that, too.

JoeCohen:
Thanks Jason!
TLDR: 
- I'd like @mo-nathan to weigh in; 
- I'm personally ok with anything (including disabling the cop); 
- Easiest to administer is simply accepting RuboCop's defaults. 
- I like your approach, but we'd need our own list of declarative methods.

RuboCop agrees with you about "directive" methods. [RuboCop says](https://rubystyle.guide/#method-invocation-parens): 
1. Omit parens methods that are part of an internal DSL (e.g., Rake, Rails). Ex: `validates :name, presence: true`;
2. Omit parens declarative methods (a.k.a. DSL methods or macro methods) that have "keyword" status in Ruby (e.g., various Module instance methods). Ex: `attr_reader :name, :age`
3. For non-declarative methods with "keyword" status (e.g., various Kernel instance methods, `puts`) either: (a) never parens, exor (b) always parents. "Never" is more popular. 
Problem: there's no agreement which methods are in this category, and RuboCop simply punts. So with the defaults, RuboCop complains about `puts "foo"`. So we'd need to build our own list. 

I'll a look at yesterdays' list of Codeclimate issues.

app/classes/api/name_api.rb
```diff
@@ -104,7 +104,9 @@ def validate_create_parameters!(params)
     def validate_update_params!(params)
       @classification = params[:classification]
       validate_classification!(params)
-      raise(MissingSetParameters.new) if params.empty? && no_other_update_params?
+      return unless params.empty? && no_other_update_params?
+
+      raise(MissingSetParameters.new)
```
JoeCohen, line 109: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/classes/gmaps.rb
```diff
@@ -181,9 +181,10 @@ def event_init(obj, event, code)
     def div(args)
       width = height = nil
       args.each do |key, val|
-        if key == :width
+        case key
+        when :width
           width = val
-        elsif key == :height
+        when :height
```
JoeCohen, line 187: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/classes/gmaps.rb
```diff
@@ -263,9 +264,10 @@ def initialize(lat_long, opts)
       self.draggable   = false
       self.info_window = nil
       opts.each_key do |key|
-        if key == :draggable
+        case key
+        when :draggable
           self.draggable = !!opts[key]
-        elsif key == :title
+        when :title
```
JoeCohen, line 270: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/classes/pattern_search/term.rb
```diff
@@ -245,6 +245,9 @@ def parse_date_range
       raise(TooManyValuesError.new(var: var)) if vals.length > 1
 
       val = vals.first
+      # rubocop:disable Style/CaseLikeIf
```
JoeCohen, line 248: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/controllers/location_controller.rb
```diff
@@ -889,10 +889,10 @@ def update_observations_by_where(location, given_where)
...
       o.location_id = location.id
       o.where = nil
-      unless o.save
-        flash_error(:runtime_location_merge_failed.t(name: o.unique_format_name))
-        success = false
-      end
+      next if o.save
+
+      flash_error(:runtime_location_merge_failed.t(name: o.unique_format_name))
+      success = false
```
JoeCohen, line 895: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/models/name_sorter.rb
```diff
@@ -129,14 +129,15 @@ def only_single_names
   end
 
   def push_synonym(arg)
-    if arg.is_a?(Integer)
+    case arg
+    when Integer
       @approved_synonyms.push(Name.find(arg))
-    elsif arg.is_a?(ActiveRecord::Base)
+    when ActiveRecord::Base
```
JoeCohen, line 135: Fixes offense created by autocorrecting `MethodCallWithArgsParentheses`

app/models/name_sorter.rb
```diff
@@ -129,14 +129,15 @@ def only_single_names
...
       @approved_synonyms.push(arg)
     else
       raise(TypeError.new(
         "NameSorter synonyms must be Integer or ActiveRecord::Base, "\
-        "not #{arg.clasS}."
+        "not #{arg.class}."
```
JoeCohen, line 140: Fixes unrelated bug I spotted. 
Part of the reason this bug is here is that this line is uncovered.
So I plan to: revert this fix, add a new PR which makes the fix and covers this line.

JoeCohen: Story created: https://www.pivotaltracker.com/story/show/174349504

mo-nathan:
In my heart of hearts I'm old Lisp guy.  Ruby's optional parens have always struck me as trouble waiting to happen.  The mere fact that parens have now become the standard (with a few arguable exceptions) I think proves my point.  Heck personally, I'd prefer to get rid of precedence rules and force parens everywhere, but I recognize that this is my inner Lisp fanboy speaking.  That being said, I can kind of understand making exceptions for `require` and `gem`, so I'm cool with a small list of exceptions if someone feels strongly about it and wants to propose that list, but in my heart I think parens are a very good thing.  At work we use RSpec which tries to really embrace the "DSL" thing so it generally plays fast and loose with parens, but in my experience it just creates more bugs and confusion and we've essentially gone back to using parens.

JoeCohen:
> In my heart of hearts I'm old Lisp guy. Ruby's optional parens have always struck me as trouble waiting to happen. The mere fact that parens have now become the standard (with a few arguable exceptions) I think proves my point. Heck personally, I'd prefer to get rid of precedence rules and force parens everywhere, but I recognize that this is my inner Lisp fanboy speaking. That being said, I can kind of understand making exceptions for `require` and `gem`, so I'm cool with a small list of exceptions if someone feels strongly about it and wants to propose that list, but in my heart I think parens are a very good thing. At work we use RSpec which tries to really embrace the "DSL" thing so it generally plays fast and loose with parens, but in my experience it just creates more bugs and confusion and we've essentially gone back to using parens.

I'm OK with parens everywhere.
@pellaea or @nimmolo: Are you interested in proposing a short exception list?

pellaea:
I can live with parens everywhere.

PR #636, opened 2020-08-15
NameSorter coverage

JoeCohen:
Completes coverage of NameSorter, fixing a bug in a previously uncovered line.

Delivers https://www.pivotaltracker.com/story/show/174349504

(There is no manual test.)

### DIFF
--- a/app/models/name_sorter.rb
+++ b/app/models/name_sorter.rb
@@ -136,7 +136,7 @@ class NameSorter
     else
       raise TypeError.new(
         "NameSorter synonyms must be Integer or ActiveRecord::Base, "\
-        "not #{arg.clasS}."
+        "not #{arg.class}."
       )
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/name_sorter.rb
```diff
@@ -129,15 +129,16 @@ def only_single_names
   end
 
   def push_synonym(arg)
-    if arg.is_a?(Integer)
+    case arg
```
mo-nathan, line 132: This is better than the if/elsif/else, but personally I'd do:
```
  def push_synonym(arg)
    return @approved_synonyms.push(Name.find(arg)) if arg.is_a?(Integer)
    return @approved_synonyms.push(arg) if arg.is_a?(ActiveRecord::Base)

    raise(TypeError.new(
            "NameSorter synonyms must be Integer or ActiveRecord::Base, "\
            "not #{arg.class}."
          ))
  end
```

JoeCohen: Yes, that's better.

PR #641, opened 2020-08-19
Fix article bug introduced in #629

JoeCohen:
Fix `RssLog#url` for articles

- Test `rss_log.url` for  normalized controllers
- Add `AbstractArticle#show_path`
- Correct `rss_log.url` for `article`.  (This was overlooked in #629.)
- Add rss_log article fixture (just in case).

Delivers https://www.pivotaltracker.com/story/show/174394837
There is no manual test.


app/models/rss_log.rb
```diff
@@ -287,7 +287,7 @@ def url
       format("/glossary/show_glossary_term/%d?time=%d",
              glossary_term_id, updated_at.tv_sec)
     elsif article_id
-      format("/article/show_article/%d?time=%d", article_id, updated_at.tv_sec)
+      format("/articles/%d?time=%d", article_id, updated_at.tv_sec)
```
JoeCohen, line 290: Is the `url` method dead code? 
Is it called in `rss.xml.builder`?
If it's dead, then this PR should be closed and ignored.

test/models/rss_log_test.rb
```diff
@@ -2,5 +2,44 @@
...
+      assert_match(type_normalized_show_path(type, id), rss_log.url,
+                   "rss_log.url incorrect for #{model(type)}")
+    end
+  end
+
+  # ---------- helpers ---------------------------------------------------------
+
+  def normalized_rss_log_types
+    RssLog.all_types.each_with_object([]) do |type, ary|
+      ary << type if model(type).controller_normalized?(model(type).name)
```
pellaea, line 20: Why does model(type).name have to be passed as a parameter?

test/models/rss_log_test.rb
```diff
@@ -2,5 +2,44 @@
...
+    end
+  end
+
+  def model(type)
+    type.camelize.constantize
+  end
+
+  # rss_log factory
+  def rss_log_for_type(type)
```
pellaea, line 29: It would help me to emphasize that this is *creating* a sample rss_log object.  That would save me from having to parse this method to find out what it's doing.  Or a comment would work just as well, but those are frowned on I guess?  What about something like "create_test_rss_log" (since type is being passed as a parameter, the "for_type" is sorta implicit).

test/models/rss_log_test.rb
```diff
@@ -2,5 +2,44 @@
...
+    # Target must have id; use an existing object to avoid hitting db
+    target = model(type).first
+    rss_log = RssLog.new
+    rss_log["#{type}_id".to_sym] = target.id
+    rss_log.updated_at = Time.zone.now
+    rss_log
+  end
+
+  def target_id(rss_log)
```
pellaea, line 38: This method doesn't appear to be used(?)  And I was going to question how it helps, anyway, since "rss_log.target_id" is actually one character shorter and arguably easier to understand (doesn't require checking to make sure "target_id(rss_log)" doesn't do anything fishy in addition to the obvious.

test/models/rss_log_test.rb
```diff
@@ -2,5 +2,44 @@
...
+    rss_log.target_id
+  end
+
+  def type_normalized_show_path(type, id)
```
pellaea, line 42: This seems a bit circular actually.  If I'm understanding this test correctly, the whole point is to test whether the model "knows" its own path correctly.  So you test that "model.show_path" = "model.show_controller/id"... but surely model.show_path uses model.show_controller internally(?)  (If not, then there is clearly room for DRY improvements.)  It seems to me this would be a more meaningful test if the test derived the correct path itself independently.  That is, convert type to the appropriate path.  Or perhaps better, just write in the expectation explicitly, then we're left with no question about whether it's making the correct, independent assertion.

test/models/rss_log_test.rb
JoeCohen: You're right! Good catch.
I wrote this before this before realizing that `rss_log.target_id` existed.
Then I forgot to delete it.

test/models/rss_log_test.rb
JoeCohen: I know that code looks (and probably is) ridiculous. But`AbstractModel#controller_normalized?` takes an argument.
```ruby
  def self.controller_normalized?(name)
    class_defined?("#{name.pluralize}Controller")
  end
```
I no longer remember why I wrote `controller_normalized?` that way.

test/models/rss_log_test.rb
JoeCohen: Thank you!
Your name for that method is much better. I'll fix it.

pellaea:
Ah, right, I remember running across that and being very confused by that
method, too.  Fair enough, save it for some other time.

test/models/rss_log_test.rb
JoeCohen: I don't think it's circular.
It compares
- the normalized show path for that type (`type_normalized_show_path(type, id)`)
to
- whatever is returned by RssLog.url (`rss_log.url`), which is hard coded and could be incorrect. 
For `article`, RssLog.url was returning `/article/show_article/...` which is wrong (and is fixed in this PR); it should be `/articles/...`

pellaea:
Fair enough, I trust you.  I just couldn't tell what was going on there.

mo-nathan:
What's going on with this PR?  I see there are conflicts.  It would be really helpful to summarize the state of things in a comment on the PR when a PR is lingering like this.  I vaguely remember some dialog about this PR, but I don't remember where we left it.

JoeCohen:
Just noticed this comment. I'll have a look shortly.

pellaea:
I feel a little bad about this, Joe.  I feel like we kinda dropped the ball
on it.  But you are more likely to be able to reconstruct where we left it
than I would be, because I've been extremely distracted the last several
months and never really paid as much attention to it as I should have.

JoeCohen:
No problem. It's not critical.
I'll have a look now.

JoeCohen:
Turning this back to a draft, as my fix of the conflicts created many test errors.

JoeCohen:
I am:
- closing this PR
- renaming the branch to `old..."
- create a new branch with a similar name, starting form the current master, then cherry picking the commits (there are only 5) from this branch.
I screwed things up in the merger, did not undo it properly on my VM, and am afraid I'll break other things by continuing with this branch.

pellaea:
I'm so sorry!  Would it help if we promised to actually review it in a
timely manner this time around?...

JoeCohen:
You did review it in a timely manner. 
Everything else is a problem I created or could have avoided by: (a) keeping the branch up to date, and (b) being more attentive when merging.

PR #642, opened 2020-08-19
Api region

pellaea:
Give support for "region" Query parameter to observations API queries.  And add observation ids to detail=low output of images API queries.

### DIFF
--- a/app/views/api/_image.xml.builder
+++ b/app/views/api/_image.xml.builder
@@ -13,16 +13,17 @@ xml.tag!(tag,
   xml_datetime(xml, :last_viewed, object.last_view)
   xml_boolean(xml, :ok_for_export, true) if object.ok_for_export
   xml_string(xml, :original_name, object.original_name) if check_permission(object)
-  xml_detailed_object(xml, :license, object.license)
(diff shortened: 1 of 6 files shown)


app/views/api/_image.xml.builder
```diff
@@ -13,16 +13,17 @@ xml.tag!(tag,
   xml_datetime(xml, :last_viewed, object.last_view)
   xml_boolean(xml, :ok_for_export, true) if object.ok_for_export
   xml_string(xml, :original_name, object.original_name) if check_permission(object)
-  xml_detailed_object(xml, :license, object.license)
+  xml.observations(number: object.observations.length) do
+    for id in object.observation_ids
```
JoeCohen, line 17: You would offend RuboCop 😭  if RuboCop had looked at this page. 😁 
RuboCop would like:
```suggestion
    object.observation_ids.each do |id|
```
I wonder if I could get RuboCop to look at .xml.builder pages. (Not likely.)

pellaea:
Sorry, for quick fixes, I don't worry about fixing pre-existing problems.
Just don't have the time. :(

PR #647, opened 2020-08-24
Two small tweaks

pellaea:
The "Show More" and "Edit" links for the Brief Description on show_name page were not showing up, not sure why.  Anyway, I moved them to beside the section header to be like all the rest of the sections.  Works now.

Also, some error messages on the observation page were white-on-white and thus invisible in the Amanita theme.  Overrode this and made the text black for that one specific case.


JoeCohen:
Thank you! 
I never realized those links were there.

PR #648, opened 2020-08-27
Cleaning up show_species_list page.

pellaea:
1. Removed bulk_editor, as it has not been used since January 2019 and no longer works.

2. Added "download" page to combine all the various ways of downloading names and observations (make_report x 3, print_labels and download_observations).

3. Added "Clear Species List" to allow users to quickly remove all observations from a list.  (Useful for making a general-purpose "clipboard"-type list that can be used to accumulate selected observations, then be cleared so another list of selected observations can be built.)

4. Renamed and rearranged the links in the upper right to be more self-explanatory and easy to understand.

5. Added the remaining functionality from the regular equivalent observation index (the old "Show Observations") 
to this page so that you don't have to click through to "Show Observations" in order to show a map, for example.

6. Fixed and tested the "Set Source" / "Save For Later" feature which has been broken, apparently, for a long time.

7. Added "Add To / Remove From Another List" (from regular observation index) to complete the set of things one can do with the observations in the list: "set source", "clone list", "add/remove from another list", "clear list".

8. Extended SiteData.update_contribution slightly to accept a number of contributions > 1 so that we can continue to use this method with bulk operations.


app/models/species_list.rb
```diff
@@ -109,6 +109,18 @@ def remove_obs_callback(_obs)
...
 
+  def clear
+    num = SpeciesList.connection.select_value(%(
+      SELECT COUNT(observation_id) FROM observations_species_lists
+      WHERE species_list_id = #{id}
+    ))
+    SpeciesList.connection.delete(%(
+      DELETE FROM observations_species_lists
+      WHERE species_list_id = #{id}
+    ))
```
JoeCohen, line 120: Will this work:
```suggestion
(SpeciesList.find(id)).observations.delete_all
```
It generates sql like this:
```sql
DELETE FROM `observations_species_lists` 
WHERE `observations_species_lists`.`species_list_id` = 75375684 
AND `observations_species_lists`.`observation_id` IN (573429226, 1046914674)
```

app/views/species_list/_form_species_lists.html.erb
```diff
@@ -11,7 +11,7 @@
   <% end %>
 
   <% if @checklist && @checklist.any? %>
```
JoeCohen, line 13:
```suggestion
<% if @checklist&.any? %>
```

test/controllers/species_list_controller_test.rb
```diff
@@ -1116,6 +1116,26 @@ def test_make_report
     end
   end
 
+  def test_download
+    spl = species_lists(:one_genus_three_species_list)
+    query = Query.lookup_and_save(:Observation, :in_species_list,
+                                  species_list: spl)
+
+    get_with_dump(:download, id: spl.id)
```
JoeCohen, line 1124:
```suggestion
 get(:download, id: spl.id)
```
My memory is that it was okay to get rid of the `with_dump` methods (because we were't checking the validity of the generated html)

pellaea:
Great, thanks, Joe!  Yes, I was specifically interested in whether I should
be using the path helpers.

I like all but one of your suggestions:

spl.observations.delete_all -- Exactly what I wanted!... Oh wait, but will
it try to destroy the observations? or call any callbacks on every single
observation?  I better check!  That could get super ugly.

get instead of get_with_dump -- Sounds right, I think I remember that
conversation.

obj&.method -- Ah, that's what that means, yes, I like.

# :norobots: no longer used -- I thought so, but they are still there, so
maybe I misremembered... If not used we should actually remove them.  I
tend to just copy what's around in the interests of consistency and
changing as little as possible.  I'll create a small PR that strips out
those # :norobots: comments (and I think there are a couple of others like
:admin: and :login: or something similar).

I don't like the new way to count observations, since it requires two joins
where none are required.  Admittedly this is not a common action, so maybe
"who cares?"  Maybe a compromise, since we already have the species list
(self!):

self.observations.count

I'll check what kind of SQL that executes.  One join would be okay I
suppose.  Certainly easier to read.  I just have to hold my nose since I
don't like not having any idea what's happening under the hood.  But that's
not the Rails Way!  You'd think I'd have gotten over it by now. :)

Thanks again for looking it over, Joe.

JoeCohen:
Thanks Jason! I appreciate the feedback.

1. Re sql vs ActiveRecord. I didn't completely understand the intent, as I
looked only at the diffs, rather than the entirety of the changed files.
Sorry about that.
Maybe

self.observations.delete_all(:nullify)

It shouldn't destroy the observations. And I think it doesn't trigger
callbacks, but the documentation <http://Link> isn't clear about that.

2. Re paths/helpers
With current routing,

redirect_to(action: "show_species_list", id: @species_list)

would be replaced by

redirect_to(species_list_show_species_list_path(@species_list))

which strikes me as hard to understand and ugly.
When SpeciesListController is normalized with new routes, it will end up as

redirect_to(species_list*s*_path(@species_list))

3. Contribution. Given where Nathan is headed with revising Namings,
Contribution would matter only for bragging rights. So maybe in the long
run, Contribution could go away. So maybe you shouldn't worry too much
about it.

On the other hand: (a) Who knows exactly how Namings will end up or when
that work will be finished, and (b) in any event, bragging rights can
motivate some people. (But maybe the wrong people and people game the
system.)

pellaea:
Good point re: contribution points going away.  And amen!  Good riddance.

re: routing explanation -- Thanks!  That's not complicated at all.  Good.

re: SQL queries -- Mostly I was too lazy to look up the Rails way of doing
it and  to make sure it was doing what I intended.  I can take the time
today.  I used to know these things, but I've forgotten most of what I know
about Rails. :(

pellaea:
I bet the decrease in coverage has to do with removing bulk_editor.  I'm pretty sure I've covered every change I made.

JoeCohen:
Maybe we need a new framework: Ruby on SQL

nimmolo:
Sorry it's taken me so long to get to this. I'm overloaded on work right now.

I defer to Joe's very sharp review here.

JoeCohen:
> I bet the decrease in coverage has to do with removing bulk_editor. I'm pretty sure I've covered every change I made.

In general you can find out the details of Coverage change by following the Details link on the GitHub page for the PR.
Screen Shot 2020-08-28 at 16 02 06

That takes you to the Coveralls build page, which should list total # of "new" missed lines with a link to their files.
Screen Shot 2020-08-28 at 16 03 13

Then follow the file links and scroll down to see the "new" missed lines, e.g.:
Screen Shot 2020-08-28 at 16 12 07

"New" means that **the line** is different. I.e., it apply to lines which were previously uncovered and are changed.

pellaea:
Thanks, Joe.  That makes sense.  I moved that method.  Didn't change it,
but it would *look* new.  It's only there for backwards compatibility, but
I'll cover it real quick anyway.  So easy to.

JoeCohen:
To clarify: there are 11 "new" uncovered lines; I included only a few in
the screen shot. Sorry for the confusion.
I'm not implying that you have to cover all of them (some/all were
uncovered before you started),
But if you're interested see
https://coveralls.io/builds/33053468/source?filename=app%2Fcontrollers%2Fspecies_list_controller.rb#L208
and
scroll down.

mo-nathan:
Let me know if you want me to help out with getting the coverage back up.

pellaea:
Thanks, Nathan, that's okay.  I was just starting to look into this
branch.  I'll fix the conflicts, cover new and existing uncovered lines so
I can merge it in the next few days.

I take it there are no objections to the changes I made?

app/controllers/species_list_controller.rb
```diff
@@ -995,44 +926,60 @@ def find_chosen_name(id, alternatives)
...
-                    join: { observations: :names },
-                    where: "rss_logs.observation_id > 0",
-                    limit: 1000
-                  )
-                else
-                  []
-                end
+    return unless query ||= query_from_session
+
+    case query.model.name
```
mo-nathan, line 931: Only the Observation case is covered.  Given the refactor it would be good to get these other cases covered since code varies from case to case.

mo-nathan:
Seem like good changes and the implementation makes sense.  There's a bit more coverage I'd like to see.  Fixed a couple of stray Rubocop whitespace issues.  I haven't seen it working yet, so can't comment on the actual functionality.  However, I'd be comfortable having it merged.

app/models/species_list.rb
pellaea: I'm looking into this.  I'm concerned that it needs to include the full list of observation_ids explicitly.  At very least this will involve an additional query to look up those observation_ids (even thought it is completely unnecessary).  But I'm also concerned that this will trigger multiple callbacks.  I don't feel like this action should result in any callbacks.

pellaea: Doesn't look like it calls any callbacks, which is good.  But I'm still concerned that it has to load all the observations -- and it loads the *entire* observations, not just the ids.  Some species lists are huge.  It would be easy, in fact, to create a list with every observation on MO in it!

It's admittedly less "Railsy" but the SQL is so simple -- and it's a lot clearer to figure out what it actually does!  I don't have to spin up my VM and carefully test it to prove whether it calls callbacks or makes unnecessary expensive queries -- that I personally see no compelling reason not to use it.  It's not like some upgrade to Rails will ever invalidate basic low-level SQL!

JoeCohen: The SQL is best. (I can't find an efficient way to do this with an ActiveRecord query.)

pellaea: Including the rails version in a comment together with summary of this discussion seems to get us the best of both worlds.

PR #649, opened 2020-08-28
Removed :norobots:, :login:, :root: and :prefetch: "directives".

pellaea:
I was going to write a test to use these directives to verify that the login_required filter and others were maintained correctly. But since I didn't and we therefore the directives fell into disuse, they are now more of a liability than a help.


JoeCohen:
Thank you! Looks great.
And you've timed it so that Travis CI passes!

PR #651, opened 2020-08-28
Relax api restrictions

pellaea:
1. Reduces weight given to API requests when calculating rate and load stats for an IP.
2. Treats any request including an API key as logged in (and therefore even more relaxed rate and load restrictions apply).


pellaea:
I don't know what the alleged two codeclimate complexity issues are.  Undoubtedly must be pre-existing problems if they are real.  I'm going to ignore so that I can get this into production sooner than later.

PR #653, opened 2020-08-29
Fix Rails/Date and /TimeZone issues in views

JoeCohen:
RuboCop can't find or fix these because it doesn't examine view files. 

Details at https://www.pivotaltracker.com/story/show/174562904


pellaea:
Looks good to me.  Shall I merge it?

JoeCohen:
Thanks so much for reviewing it!
If you feel like pushing it, please go ahead.
Otherwise, I'll do it in the a.m. (I'm heading to sleep in a few seconds.)

PR #654, opened 2020-08-30
Fixed tiny bug in name description stuff found in the wild.

pellaea:


mo-nathan (approved):
After switching from a case statement to a series of guard clauses, I confirmed that the test fails on current master and succeeds on the branch.  Also checked that there aren't any new rubocop violations added.

pellaea:
Thanks, Nathan.  Not a fan of switch statements? :)

mo-nathan:
> Thanks, Nathan. Not a fan of switch statements? :)

Yeah.  A lot of folks don't like them.  They are kind of verbose for what they do and they tend to lead to more complexity than they are worth.  Guard clauses are much more Ruby-ish and result in both tighter and in my view more readable code.

pellaea:
Ah well, so it goes.  My grand purpose on this earth is apparently to
ensure greater diversity in opinion!  It just sucks always being in the
minority. :)

PR #657, opened 2020-09-02
Fix Articles destroy links

JoeCohen:
Change `Articles` destroy links to buttons

`destroy` links don't work because they require js,
so use button instead.

- Promote and refactor `destroy_button` method
- Use button instead of link to destroy `article`
- Revise GlossaryTermsHelper and view to work with promoted method

Delivers https://www.pivotaltracker.com/story/show/174524961

Suggested manual test script:
- Go into admin mode
- Create an article. 
  Result: The show_article page should have a "Destroy" link
- Click it.
Result: The article should be destroyed.
- Create a GlossaryTerm
  Result: The show_glossary_term page should have a "Destroy Glossary Term" link
- Click it
  Result: The term should be destroyed.

### DIFF
--- a/app/helpers/glossary_terms_helper.rb
+++ b/app/helpers/glossary_terms_helper.rb
@@ -2,11 +2,7 @@
 
 # View Helpers for GlossaryTerms
 module GlossaryTermsHelper
-  def destroy_button(term)
-    button_to(
-      :destroy_glossary_term.t,
-      { action: "destroy", id: term.id },
-      method: :delete, data: { confirm: "Are you sure?" }
-    )
+  def glossary_term_destroy_button(term)
+    destroy_button(object: term, name: :destroy_glossary_term)
   end
 end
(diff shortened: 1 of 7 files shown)


app/helpers/glossary_terms_helper.rb
```diff
@@ -2,11 +2,7 @@
...
 # View Helpers for GlossaryTerms
 module GlossaryTermsHelper
-  def destroy_button(term)
-    button_to(
-      :destroy_glossary_term.t,
-      { action: "destroy", id: term.id },
-      method: :delete, data: { confirm: "Are you sure?" }
-    )
+  def glossary_term_destroy_button(term)
+    destroy_button(object: term, name: :destroy_glossary_term)
```
mo-nathan, line 6: Surprisingly not covered by current tests.

mo-nathan (changes requested):
Functionality and rubocop check out.  Coverage missing for glossary_term_destroy_button.

JoeCohen:
Thanks for the review. I'll take a look at coverage. I'm also surprised
it's not covered.

mo-nathan (approved):
Looks good.  Thanks!

PR #658, opened 2020-09-04
Validate glossary terms

JoeCohen:
Prevents saving a GlossaryTerm if
- the Name is blank
- the Name is not unique
- the Term lacks both a Description and an Image.

**Suggested Manual Test**
Try to create (and/or edit) a GlossaryTerm with:
1. a blank Name;
2. a name that duplicates the Name of an existing term (e.g. "Acrid", "Lichen");
3. neither Description nor Image.
Expected result: 
You are returned to the form, with appropriate red flash error(s).

Delivers https://www.pivotaltracker.com/story/show/174606044
Also incidentally delivers:
- https://www.pivotaltracker.com/story/show/174608605 (Use single form for `new` and `edit` forms)
- https://www.pivotaltracker.com/story/show/124382351 (Edit and Create GlossaryTerm should have link to Textile sandbox)

### DIFF
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -37,19 +37,19 @@ class GlossaryTermsController < ApplicationController
   # ---------- Actions to Modify data: (create, update, destroy, etc.) ---------
 
   def create
-    @glossary_term = \
-      GlossaryTerm.new(user: @user, name: params[:glossary_term][:name],
-                       description: params[:glossary_term][:description])
+    @glossary_term = GlossaryTerm.new(
+      user: @user,
+      name: params[:glossary_term][:name],
+      description: params[:glossary_term][:description]
+  )
 
     if params[:glossary_term][:upload_image]
       @glossary_term.add_image(process_image(image_args))
     end
 
-    unless @glossary_term.save
-      return reload_form("new")
-    else
-      redirect_to(glossary_term_path(@glossary_term.id))
-    end
+    return reload_form("new") unless @glossary_term.save
+
+    redirect_to(glossary_term_path(@glossary_term.id))
   end
 
   def update
@@ -58,11 +58,9 @@ class GlossaryTermsController < ApplicationController
                                 permit(:name, :description)
     @glossary_term.user = @user
 
-    unless @glossary_term.save
-      return reload_form("edit")
-    else
-      redirect_to(glossary_term_path(@glossary_term.id))
-    end
+    return reload_form("edit") unless @glossary_term.save
+
+    redirect_to(glossary_term_path(@glossary_term.id))
   end
 
   def destroy
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -136,7 +136,9 @@ class GlossaryTermsController < ApplicationController
     return if @glossary_term.save # happy path
 
     # term failed, so clean up the orphaned (unassociated) image
-    added_image.try(:destroy)
+    # and its flash notice ("Successfully uploaded image ...")
+    saved_image.try(:destroy)
+    flash_clear
     false
   end
 
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -26,10 +26,7 @@ class GlossaryTermsController < ApplicationController
   # ---------- Actions to Display forms -- (new, edit, etc.) -------------------
 
   def new
-    @copyright_holder = @user.name
-    @copyright_year = Time.now.utc.year
-    @upload_license_id = @user.license_id
-    @licenses = License.current_names_and_ids(@user.license)
+    init_image_form_instance_variables
   end
 
   def edit
@@ -46,8 +43,36 @@ class GlossaryTermsController < ApplicationController
     if params[:glossary_term][:upload_image]
       glossary_term.add_image(process_image(image_args))
     end
-    glossary_term.save
-    redirect_to(glossary_term_path(glossary_term.id))
+
+    unless glossary_term.save
+      return reload_form(term: glossary_term, form: "new")
+    else
+      redirect_to(glossary_term_path(glossary_term.id))
+    end
+  end
+
+  def reload_form(term:, form:)
+    load_flash_errors(term)
+    @name = params[:glossary_term][:name]
+    @description = params[:glossary_term][:description]
+    init_image_form_instance_variables
+    render(form)
+  end
+
+  def load_flash_errors(term)
+    errors = ""
+    term.errors.messages.each_value do |v|
+      errors = errors + "#{v.first}\n"
+    end
+    flash_error(errors)
+    # redirect_back_or_default(glossary_terms_path)
+  end
+
+  def init_image_form_instance_variables
+    @copyright_holder ||= @user.name
+    @copyright_year ||= Time.now.utc.year
+    @upload_license_id ||= @user.license_id
+    @licenses ||= License.current_names_and_ids(@user.license)
   end
 
   def update
@@ -99,6 +124,7 @@ class GlossaryTermsController < ApplicationController
 
   # --------- Other private methods
 
+  # Permit mass assignment of image arguments for testing purposes
   def image_args
     Rails.env.test? ? permit_upload_image_param : strong_upload_image_param
   end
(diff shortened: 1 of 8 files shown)


app/models/glossary_term.rb
```diff
@@ -13,6 +14,14 @@ class GlossaryTerm < AbstractModel
 
   # Add before_save validity check(s)
   # See https://www.pivotaltracker.com/story/show/174606044
+  validates :name, presence: {
+    message: Proc.new { :glossary_error_name_blank.t }
+  }
+  validates :name, uniqueness: {
```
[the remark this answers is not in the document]
JoeCohen: ### Is it worth creating a db index for `name`?

(I didn't think it was worth the trouble. Having two GlossaryTerms with the same Name is awkward and looks silly, but it won't break anything.)

app/controllers/glossary_terms_controller.rb
```diff
@@ -99,6 +103,28 @@ def show_past
...
+    flash_error(model_error_messages)
+  end
+
+  def init_image_form_instance_variables
+    @copyright_holder ||= @user.name
+    @copyright_year ||= Time.now.utc.year
+    @upload_license_id ||= @user.license_id
+    # rubocop:disable Naming/MemoizedInstanceVariableName
+    # RuboCop wants to name the ivar "@update_image_form_instance_variables"
+    @licenses ||= License.current_names_and_ids(@user.license)
```
JoeCohen, line 123: Is there a better way to resolve the RuboCop complaint about line 123:
>Memoized variable @licenses does not match method name init_image_form_instance_variables. Use @init_image_form_instance_variables instead.

(I'm not going to name the ivar `@init_image_form_instance_variables`.)

app/controllers/glossary_terms_controller.rb
```diff
@@ -39,24 +37,30 @@ def edit
...
+    )
 
     if params[:glossary_term][:upload_image]
-      glossary_term.add_image(process_image(image_args))
+      @glossary_term.add_image(process_image(image_args))
     end
-    glossary_term.save
-    redirect_to(glossary_term_path(glossary_term.id))
+
+    return reload_form("new") unless @glossary_term.save
```
JoeCohen, line 50: There's a slight (7-year old) issue here, exacerbated by validation:
An image can be saved but the `@glossary_term.save` fails.
**Should I wrap them in a transaction?**
But I thinks that's not the right thing to do. In particular, if the image.save fails, then the term should nonetheless save (without the image).
Maybe (pseudocode
```
Save the term.
If that fails, 
  reload the form
else 
  something like`@glossary_term.thumb_image = image`
  redirect to the show page
end
```

JoeCohen: That won't work. If there's an image but no description, then the "save the term" will fail a validity check.

JoeCohen:
Temporarily converting to Draft PR while I think through one issue.

app/controllers/glossary_terms_controller.rb
```diff
@@ -99,28 +98,55 @@ def show_past
...
-    args = strong_upload_image_param
-    # Remove "permitted: false" from this param so that model can mass assign it
-    args[:image] = params[:glossary_term][:upload_image][:image]
-    args
+  def add_glossary_term_error_messages_to_flash
+    @glossary_term.errors.messages.each_value do |val|
+      # flash_error takes a string; val is an array of size 1, e.g. ["message"]
+      flash_error(val.first)
+    end
+  end
```
JoeCohen, line 122: ### Is there a better way to handle validation?

I want to:
- Have the model (not the controller) do the validation as closely as possible to the "Rails way". See https://guides.rubyonrails.org/v5.2/active_record_validations.html. It seems to me that this is "business logic" that should be handled by the model.
- Have the model provide an error message (sensu lato). Seems to me again to be business logic, though not as clear as the above.
- Have the error message be translatable by MO;
- Display the error message in the view.

My solution:
- GlossaryTerm uses `validates presence`, etc. See `glossary_terms.rb`
- GlossaryTerm provides an error message using a proc. Ex: `message: proc { :glossary_error_name_blank.t }` 
- Controller finds any error messages and puts them into the flash. I'm not in love with this, but I think it's required by the application layout. I think that normally, the view would handle this without intervention by the controller.

JoeCohen:
### Status:

- Is there a better way (than disabling the cop) to resolve the [RuboCop complaint](https://github.com/MushroomObserver/mushroom-observer/pull/658/files#r483861394) about `init_image_form_instance_variables`? See https://github.com/MushroomObserver/mushroom-observer/pull/658/files#r483869099.
- Is there a better way to handle validation? See https://github.com/MushroomObserver/mushroom-observer/pull/658/files#r484570362
- Should GlossaryTerm `name` be indexed to guarantee uniqueness of `name` in the db? I don't think it's worth the trouble. See https://github.com/MushroomObserver/mushroom-observer/pull/658/files#r483861385. When this question is resolved, I can turn the Codeclimate check green by either disabling the relevant cop or indexing `name`.

app/controllers/glossary_terms_controller.rb
```diff
@@ -26,10 +26,8 @@ def show
   # ---------- Actions to Display forms -- (new, edit, etc.) -------------------
 
   def new
-    @copyright_holder = @user.name
-    @copyright_year = Time.now.utc.year
-    @upload_license_id = @user.license_id
-    @licenses = License.current_names_and_ids(@user.license)
+    @glossary_term = GlossaryTerm.new
+    init_image_form_instance_variables
```
pellaea, line 30: Do you really mean "image" here and not "term"?

pellaea: Oh, sorry, just confused, ignore me!

app/controllers/glossary_terms_controller.rb
pellaea: Sounds exactly right to me.  I think we should follow that pattern everywhere.

app/controllers/glossary_terms_controller.rb
```diff
@@ -99,28 +98,55 @@ def show_past
...
+    return if @glossary_term.save # happy path
+
+    # term failed, so clean up the orphaned (unassociated) image
+    # and its flash notice ("Successfully uploaded image ...")
+    saved_image.try(:destroy)
+    flash_clear
+    false
+  end
+
+  def upload?
```
pellaea, line 145: Very minor point: but would you mind changing the name to "upload_present?" or equivalent so that it is more readily apparent that this method is not a verb (meaning an action actually occurs)?

app/models/glossary_term.rb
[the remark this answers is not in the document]
pellaea: I'd say ignore codeclimate here, I like what you did.  Duplicate terms does seem odd, but it's not such a big deal that we would create in index in the DB table (really?!) to make it physically impossible.  There are so many *far* better places to add indexes if we're going there!  (Like all the glue tables.)

pellaea:
This looks good to me.  re: "memoized variable" -- (what horrible terminology) I don't think it really applies in this case.  You aren't caching @licenses, you are supply a default value for it.  It so happens that's all the method does so it looks superficially like a method which caches a derived value.  But that's not at all what the method's purpose is.  (This would have been more obvious and wouldn't have wasted 15 minutes of my time if people used "cached derived variable" instead of the nonsensical "memoized variable" terminology.  Funny that a cop whose sole purpose is to police names would be itself so poorly named!)

app/controllers/glossary_terms_controller.rb
JoeCohen: Yep. It's real confusing. I wish I could have found a better name.

app/models/glossary_term.rb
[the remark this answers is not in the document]
JoeCohen: Thanks! I'll disable the cop for that one line.

app/controllers/glossary_terms_controller.rb
JoeCohen: Very good point! I will use `upload_specified?`

app/controllers/glossary_terms_controller.rb
mo-nathan: Why are you memoizing these values anyway?  Specifically what is the `||=` getting you that a simple `=` wouldn't?  I don't see anywhere obvious that these values are getting setting twice.  If you can remove the `||` from any of these and then put it at the end, I believe the violation will go away.

JoeCohen: Thank you.
You're absolutely right.
I un-memoized the ivars. Also improved them by re-using the values input by the user (rather than initializing) when reloading the form.

mo-nathan (approved):
Looks great to me!

JoeCohen:
Thanks for the multiple reviews and improvements.
I'll merge it.

PR #659, opened 2020-09-06
API partials

pellaea:
Revised API in several ways.
    1. Removed all the "sub-partials" and replaced them with simple inline helpers.
    2. Removed show_urls from XML objects; for some reason that was really slowing things down.
    3. Ensured that all low-detail responses do not involve any "includes" (except low-detail images include the license name, but that's okay since there are like 3 licenses!).
    4. Moved observation_ids from low-detail image response to high-detail image response; now high-detail responses aren't expensive so it's okay to force users to use them to get image-observation link, either from observations side or images side.
    5. "Flattened" the high-detail observation response a bit by placing votes at the same level as namings.
    6. Fixed several spurious bugs here and there that were encountered during the revision.
    7. Changed main results partial to render a *collection* of partials.

Removing partials sped up a high-detail observations query by factor of 3.5 locally (7.7 seconds to 2.2 seconds).  Changing to render collection of partials gained an additional 10% (just under 2 seconds; not much but really no reason not to do it).

I will probably do a sanity check on low- and high-detail observations queries before and after just to make sure nothing important has changed.  I don't want to break anyone's code that uses our API.


app/views/api/_external_site.xml.builder
```diff
@@ -5,7 +5,7 @@ xml.tag!(tag,
 ) do
   xml_string(xml, :name, object.name)
   if !detail
-    xml_minimal_object(xml, :project, Project, object.project_id)
+    xml_minimal_object(xml, :project, :project, object.project_id)
```
JoeCohen, line 8: `... :project, :project ...` looks redundant.
I don'r understand this well enough to understand the intent of `Project` in the original file.

app/views/api/_collection_number.xml.builder
```diff
@@ -8,12 +8,14 @@ xml.tag!(tag,
...
   else
     xml_detailed_object(xml, :user, object.user)
-    xml.observations(number: object.observations.length) do
-      object.observations.each do |observation|
-        xml_detailed_object(xml, :observation, observation)
+    if object.observation_ids.any?
+      xml.observations(number: object.observation_ids.count) do
+        object.observation_ids.each do |observation_id|
+          xml_minimal_object(xml, :observation, :observation, observation_id)
+        end
```
JoeCohen, line 18: Is `:observation, :observation` redundant?

app/views/api/_herbarium_record.xml.builder
```diff
@@ -3,20 +3,22 @@ xml.tag!(tag,
...
+    xml_minimal_object(xml, :herbarium, :herbarium, object.herbarium_id)
+    xml_minimal_object(xml, :user, :user, object.user_id)
   else
     xml_detailed_object(xml, :herbarium, object.herbarium)
     xml_detailed_object(xml, :user, object.user)
-    xml.observations(number: object.observations.length) do
-      object.observations.each do |observation|
-        xml_detailed_object(xml, :observation, observation)
+    if object.observation_ids.any?
+      xml.observations(number: object.observation_ids.length) do
```
JoeCohen, line 18: For (mindless) consistency:
```
      xml.observations(number: object.observation_ids.count) do
```

app/views/api/_herbarium_record.xml.builder
```diff
@@ -3,20 +3,22 @@ xml.tag!(tag,
...
+        object.observation_ids.each do |observation_id|
+          xml_minimal_object(xml, :observation, :observation, observation_id)
```
JoeCohen, line 20: `:observation, :observation` redundant?

app/views/api/_location.json.builder
```diff
@@ -1,15 +1,18 @@
...
-json.name             object.text_name
-json.latitude_north   object.north
-json.latitude_south   object.south
-json.longitude_east   object.east
-json.longitude_west   object.west
-json.altitude_maximum object.high
-json.altitude_minimum object.low
-json.notes            object.notes.to_s.tpl_nodiv
+json.name             object.text_name.to_s
+json.latitude_north   object.north if object.north.present?
```
JoeCohen, line 4: Would this work & be more efficient? (No reply necessary)
```suggestion
json.latitude_north   object.try(:north)
```

JoeCohen:
I'm ignorant about what's going on with this code.
I left a few comments which probably simply drive home that point. :-)
No reply necessary for any of them.

app/views/api/_external_site.xml.builder
pellaea: I needed to pass in the class before, because I was including the "show" url.  But since I'm not any more, I no longer need the class, so it's easier all around (especially in one particular case associated with comments) to switch to just the "type tag", which is all it needs now.

app/views/api/_collection_number.xml.builder
pellaea: Yeah, I noticed that most calls to xml_minimal_object have these two parameters identical now.  But not *all*.  The first is the field name, the second is the type.  In most cases those are the same, but in some like "owner" type "user" they are not.  Probably could refactor them all to use named params and provide a default field name, but it was late and I've already spent way too much time on this "small bug fix".

app/views/api/_herbarium_record.xml.builder
pellaea: Ah yes, forgot that we've switched from length to count.  Will do.

app/views/api/_herbarium_record.xml.builder
pellaea: See earlier comment.

app/views/api/_location.json.builder
pellaea: Probably more efficient, but specifically my point in adding all of these "if object.xxx.present?" constructs was to reduce the length of the output.  There is no reason to include nulls or empty strings in the JSON or XML output.  As the vast majority of observations do not include all fields (e.g., obscure things like external_links, or even sequences, herbarium records, etc. etc.) this reduces the overall length of the response our server needs to send to the client radically.  When you're dealing with data on 1000 observations, that adds up significantly.

pellaea:
Thanks, Joe.  I was most interested to hear whether you and Nathan and Nimmo could swallow in-lining the "sub-partials".  For example, the observation partial used to render also sub-partials for owner, images, namings, votes (and their corresponding users and names!), comments, sequences, etc. etc.  There were zillions of partials!!  In-lining them all speeds up the render by a factor of up to 3.5x.  I think we've maxed out on the benefits of in-lining, because changing from rendering each of the 1000 observations individually is scarcely slower than in-lining them in bulk using the "collection" mechanism (only 10%).  That's why I stopped short of in-lining the main object partials.

I think I will split api_helper into two pieces before merging it.  The extra helpers caused api_helper module to be too long.  It may seem to make sense to split it on json / xml lines, but I'd really like to keep each pair of xml/json_<object> helpers together so that if we change one we don't forget to change the other identically.  I sure wish there were a way to DRY those two versions of each "in-lined partial" (without adding so much more complexity that it defeats the purpose).

app/classes/api/observation_api.rb
```diff
@@ -6,27 +6,23 @@ class API
...
     self.delete_page_length      = 1000
 
     self.high_detail_includes = [
-      :comments,
-      { images: :license },
-      :location,
-      :name,
-      { namings: [:name, { votes: :user }] },
       :collection_numbers,
+      { comments: :user },
```
mo-nathan, line 16: Does this mean that only the users who commented are returned or that both the comment and now the user who created that comment is included?

app/classes/api/observation_api.rb
```diff
@@ -6,27 +6,23 @@ class API
...
+    self.low_detail_page_length  = 1000
     self.put_page_length         = 1000
     self.delete_page_length      = 1000
 
     self.high_detail_includes = [
-      :comments,
-      { images: :license },
-      :location,
-      :name,
-      { namings: [:name, { votes: :user }] },
```
mo-nathan, line 19: This looks like we are no longer providing the vote information in this API call.  That makes me a bit worried that we might break a client somewhere, but I don't know much about how the API is actually used so maybe not.  I'd be more confident if they changes were only additive, but even then some folks code API input stuff very strictly so even additive changes can cause breakage.  The traditional thing to do is to actually version the API, but that's not a simple change.  It's not clear to me why you are changing the results of the API call in this ticket which seems like it's mostly about easy performance wins.  Makes it seem a lot more risky to me.  Or maybe I don't understand what these changes mean and the results aren't actually being changed by this.

app/classes/api/observation_api.rb
```diff
@@ -6,27 +6,23 @@ class API
...
       :collection_numbers,
+      { comments: :user },
+      :external_links,
       { herbarium_records: :herbarium },
-      :sequences,
-      :user
-    ]
-
-    self.low_detail_includes = [
```
mo-nathan, line 26: I'm surprised this is going away essentially for the same reasons as mentioned above.

app/classes/api/sequence_api.rb
```diff
@@ -11,7 +11,6 @@ class SequenceAPI < ModelAPI
     self.delete_page_length      = 1000
 
     self.high_detail_includes = [
-      :observation,
```
mo-nathan, line 14: Again it's not clear to me if this is removing information from an API call that clients might be expecting.

app/helpers/api_helper.rb
```diff
@@ -8,127 +8,94 @@
...
-    str = begin
-            "%d" % val
-          rescue StandardError
-            ""
-          end
-    xml.tag!(tag, str, type: "integer")
-  rescue StandardError
+    val.present? &&
+      (str = val.try(:to_i)) &&
+      xml.tag!(tag, str.to_s, type: "integer")
```
mo-nathan, line 21: It would be easier for me to understand this if it were written as:
```
return false if val.blank?

str = val.try(:to_i)
xml.tag!(tag, str.to_s, type: "integer") if str
```
Essentially the same amount of code with less dependencies on understanding the parens and chaining.  Also gets rid of an assignment inside a statement which can be dangerous and confusing.  Should be just as efficient.

mo-nathan: Other similar cases below.

app/helpers/api_helper.rb
```diff
@@ -8,127 +8,94 @@
...
+      xml.tag!(tag, str.to_s, type: "float", units: "degrees east")
   end
 
   def xml_altitude(xml, tag, val)
-    str = "%d" % val
-    xml.tag!(tag, str, type: "integer", units: "meters")
-  rescue StandardError
-  end
-
-  def xml_undefined_location(xml, tag, val)
```
mo-nathan, line 105: Don't understand why this is going away.  Again it seems like a potential change to the output of the API call.  I *think* this might be redundant given that the `:location` is specified at least for observations, but I'm not sure.  Again I don't know how the API is really being used, but I would caution against changing the output.

mo-nathan:
> Thanks, Joe. I was most interested to hear whether you and Nathan and Nimmo could swallow in-lining the "sub-partials". For example, the observation partial used to render also sub-partials for owner, images, namings, votes (and their corresponding users and names!), comments, sequences, etc. etc. There were zillions of partials!! In-lining them all speeds up the render by a factor of up to 3.5x. I think we've maxed out on the benefits of in-lining, because changing from rendering each of the 1000 observations individually is scarcely slower than in-lining them in bulk using the "collection" mechanism (only 10%). That's why I stopped short of in-lining the main object partials.
> 
> I think I will split api_helper into two pieces before merging it. The extra helpers caused api_helper module to be too long. It may seem to make sense to split it on json / xml lines, but I'd really like to keep each pair of xml/json_ helpers together so that if we change one we don't forget to change the other identically. I sure wish there were a way to DRY those two versions of each "in-lined partial" (without adding so much more complexity that it defeats the purpose).

Short story: This looks like a great improvement!  Keep going!
Longer story: Honestly it seems kind of weird to me to use the partial mechanism at all for deep rendering of XML and JSON for an API.  I guess it's a kind of clever reuse of existing infrastructure, but I don't think I've ever seen it done that way before.  I can see it for the top level node so you get the right web stuff going on, but underneath that I'm used to just seeing calls to to_json (or to_xml if we have to support XML).

I agree that there's a lot of common code between the XML and JSON that I think could be DRYed pretty easily.  I'd have to play with it, but I would think if you essentially assembe hashes like you are for the JSON and then pass those to a fairly simple mechanism that can generate correct JSON or XML for a given hash and then invoke that mechanism from the to_xml and to_json methods, I think you might be done.  Does that make any sense?  I'd have to play with a small example to see if it pans out, but it sure seems like there should be a way.

I do notice that you're not always rendering a given object the same way depending on a requested level of detail.  To me that suggests having a PORO that points to the Rails object and knows how to generate the desired hash and encode it correctly using to_json or to_xml.  I suspect this is kind of a major refactor essentially turning what you've done inside out, but I think that's how I would approach it.  That being said, what you have seems better than what we currently have so I'm cool with it.

app/classes/api/observation_api.rb
pellaea: No, I must've confused the diff, because it still includes comments and the users who made the comments like always.

app/classes/api/observation_api.rb
pellaea: Yes, the diff is very confused.  I did only add, not subtract.  The one change I made that could possibly break existing API users was that I moved the votes "up".  That is, I include the namings exactly like before, but instead of including the corresponding votes under each naming, I show all of the votes in one array at the same level as the namings.  To my knowledge no one has used any of that information yet.  I know specifically that iNat does not, even they they wanted the information.  I think the votes were too "hidden", that's why they didn't see them.  This change will make them more visible.  (iNat wants to give users the ability to import MO observations using the owner's preferred name instead of the consensus name.)

app/classes/api/observation_api.rb
pellaea: Low detail queries were never supposed to include related objects.  The stated rule of thumb was that low-detail queries returned only what was in that table.  If you want more (e.g., the owner's name instead of just the owner's user_id), then you should be asking for the high-detail response.  I was including the name and location before because I couldn't imagine any use case where a low-detail response without name and location would be useful.  But now, thanks to the location name and consensus name (text_name) being cached in the observations table, we can still include those in the low-detail responses.  What that means is you can get pretty much everything you need for simple uses from low-detail: who, when, where, what and even the thumbnail image.

It is worth noting that as far as the DB is concerned, both the low-detail and high-detail queries are the same order.  The way the includes (eager loading) works is that it loads the main table, e.g., observations first, then in a second query it loads all the owners, then in a third it loads all the consensus names, then namings, then votes, then images, then comments, etc. etc. each table one at a time.  So it's all still O(N).  The big difference in run time is from the render, not the DB.  And now the render time is drastically reduced, so it's okay to request high-detail responses as much as you want.

So, yes, there are some minor changes that could potentially break existing API users' code, but the functionality is all still there.  It should be very easy to adapt to the new (more consistent and cleaner) format.

app/classes/api/sequence_api.rb
pellaea: Yes, I am removing details of the observation from this query.  I couldn't imagine any plausible use-case where it would ever happen.  The typical case will be to go the other direction: Ask for observations that match your criteria and that have sequences.  And in fact, high-detail responses for the observations end point include the basic sequence info (locus, bases, etc.)  An alternative way to go would be to ask just for the ids of observations matching your criteria (detail = none), and then ask for sequences for those observations.  This returns only the information you want if you are only interested in sequences.  In neither case does it make any sense to return a bunch of details about the observation when querying sequences.

I think there are a few other cases like this, too.  Collection numbers and herbarium records and external links come to mind.  These are all very specialized things that users will probably rarely query directly, or will ask for once they already have the observations.  By doing this I'm subtly encouraging people to use observations as the entry point for just about everything.  Note that, for example, the interface for querying sequences and collection numbers and comments and even images does not allow you to filter by species name or location or anything helpful.  If you want to do that you must go through observations.  I know Query allows some of that, but those queries in general involve joins and some become quite expensive (especially images joined to anything via observations and images_observations).  For the website UI it made sense to allow that to a limited degree.  But for the API, by definition, users will be programmers, and as such will be perfectly capable of doing these joins themselves.  So they are encouraged to do their fancy queries on observations first, and either get the info about the attached objects like sequences and comments and images via high detail response of the original observations query, or by asking for the matching sequences, comments, images, etc. in subsequent queries to each of those endpoints.

You could argue "what's the harm in keeping the observation details in there", but observations are *huge*, there are so many fields associated with observations.  So in the spirit of pairing things down to just what's actually used and speeding everything up, out went the observation details.

app/helpers/api_helper.rb
pellaea: It was superceded by xml_minimal_location and xml_detailed_location.  I forget what the issue was that required the change.

pellaea:
Thanks for the careful review, Nathan.  Yes, I realize that rails encourages a much more automagical approach.  I forget what happened at the beginning when I first wrote this stuff.  It's probably lost to mists of time at this point.  In principle, I provided a mechanism for specifying an API version number.  Meaning we could rewrite the rendering stuff completely, and leave the old code there for a year or whatever to give users time to port over.  Feel free to make that a pivotal ticket.

I'm concerned I got carried away with changing the output as well.  I still want to compare before and after responses to see how much it actually changed, and maybe back-pedal on some of the changes where possible.  I think it looks like more changes than it actually is.

app/helpers/api_helper.rb
pellaea: Must be another case like the hated switch statement.  This in particular is a consequence of rubocop preventing what I'd really like to do:

  return if val.blank?
  return unless str = val.try(:to_i)
  xml.tag!(tag, str.to_s, type: "integer")

But rubocop requires a blank line in there that really grates on me.  (I like to keep small pieces of code together so that I can use [ and ] to move through the code one method at a time in vi.)

Since rubocop loves these screwy "this && that && do_something()" constructs, I thought I'd try that.  Alas, that I didn't realize this also is not acceptable!

pellaea:
Okay, I have a summary of the changes I made.  I don't feel comfortable reverting any of them, unfortunately.

api_keys -- field names were wrong (created_at, last_used, num_uses)

herbarium_records -- field names were wrong (initial_determination, accession_number)

images -- moved observation_ids to high-detail response

observations -- full consensus, location and owner details moved to high detail response; votes moved up to top level of the structure for high detail response (used to be buried within namings)

projects -- removed admin and member ids (hard to eager-load, see note below)

species_lists -- removed project_ids; full location details only available in high detail response now

collection_numbers, comments, external_links, herbarium_records -- only includes observation ids now not full details


I examined API use for 2020, and no one has used projects, comments, collection_numbers or herbarium_records.  Scarcely any use of species_lists.  (There was very extensive use of external_links, though.  In fact there was more use of that endpoint than any other!  Weird.)

It's observations and images we have to worry about.  Those responses only changed by moving things around: moving info from low to high detail, or moving the location of votes.  So I'm comfortable with the changes.

re: Including list of ids of has_and_belongs_to_many relationships: It is impossible to eager-load just the ids of HABTM relationships.  This affects the ability to list the users in the project admin and member user groups; the projects and species lists that observations belong to; projects that species lists and images belong to.  Presently we have to eager load the entire user, project and species list just to get their ids.  Especially in the case of observations, this is a significant performance issue.

I recommend (and apparently the community broadly agrees) that we consider eventually switching our HABTM relationships to "has_many :through"  relationships.  This would allow us to directly eager-load just the ids of these relationships.  And it would let us supply a direct endpoint for the images_observations table, something which has come up repeatedly in the past, but that I haven't been able to accommodate without serious work.

Use case: Many users have asked to use the API to grab image ids for various sets of observations, but they don't have any need for the extreme level of detail provided by high-detail responses on the observations endpoint.  Indeed they should be able to get by with no-detail responses!  Query a list of observation ids, then query the images_observations table for a list of corresponding image ids, done.  But the current implementation of the API requires high-detail queries to either observations or images endpoints.  The latter is definitely better, but it's still not optimal.

app/classes/api/observation_api.rb
mo-nathan: I guess it's ok with me.  The API stuff is really your thing (as is probably obvious from my comments) so it's up to you.  I know I've had to deal with grumpy users whenever I've just made changes to an API pretty much no matter how small without versioning the API.  APIs are very different from human readable UIs.  The problem is that if anything breaks in their code due to some assumption they've made, they probably don't have the time to fix it at the moment so this often means the just stop using the API rather than finding time to fix it.  This is one of the reasons that old versions of APIs linger on often for many years.  In some ways the value of an API is related to how stable it is.  That being said if there aren't many users and you feel confident that either their code won't break with these change or they'll have the time and inclination to get things fixed if they do, then I'm fine with moving forward.

app/helpers/api_helper.rb
mo-nathan: Ha!  I guess it all depends on what you like best and what your tools work best with.  I've gotten really used to that blank line after the last guard and I kind of like it now, but I can see it being a problem as well.  I'm ok with the code as written, I was just saying what would be more familiar and easier for me to read.

mo-nathan:
@pellaea Your call on whether to merge this as is or not.  It still seems risky to me given all the changes and if I were doing, I would take a really hard look at versioning the API and try to go that way.  I recognize that leaving the old stuff there can continue to cause performance issues, but if it's a big concern, then I would look at deprecating the API over time with some sort of notification to the users.

The HABTM vs "has_many :through" argument kind of makes sense to me, but I haven't ever been through such a conversion.  I suspect it is related to some of the Rubocop Rails cops that have bitten me in the past so I don't have a clear vision of what that change would entail, but I'd love to know more.

JoeCohen:
Our RuboCop config file has a link to to how convert from HABTM to has_many
:through

> # MO uses HABTM extensively; RuboCop prefers the newer has_many through
> # Switching would require a migration process, detailed here:
> #
> http://chrisrolle.com/en/blog/migration-path-from-habtm-to-has_many-through
> # It's not worth the effort
> Rails/HasAndBelongsToMany:
>   Enabled: false

pellaea:
Thanks for sharing your thoughts on APIs, Nathan.  That's very helpful, and
I agree completely with what you're saying.  I was thinking last night that
I can write a script to scrape API users from the last couple of years.
That way I can write to everyone to see if my proposed changes will break
anything, and if so, I can work with them on it.  We're small and have very
few API users, let's make that work for us.

I think I will roll out a much smaller PR in the meanwhile that fixes the
most egregious bug (there are some missing includes in the current version
that cause high-detail observations requests to take forever).

pellaea:
Okay, what I've done is clone the API.  There are now two almost completely independent versions, API and API2.

This will be much easier to do again in the future, as there are lots of occurrences of API which refer to the concept of an API rather than the specific class API.  In the future we can just search and replace API2 with API3 and so on.

The one point of overlap is in helpers because rails doesn't allow you to separate out helpers.  However, fortunately, the only helpers used both in the new and old versions are xml_minimal_object and xml_detailed_object, so I just renamed the old ones to xml_minimal_object_old and xml_detailed_object_old.

The existing version (/api) is now completely unchanged (although it does incorporate the recent minimal fixes we pushed out to production the other week).

I will write a script to harvest the ids of API users from the logs so that we can notify them that there is a new version and that they should upgrade some time in the next year.  Fortunately, the changes on the users' end should be very minimal in this case.  And the new API is much faster (up to 3-4x faster!), so there is a good incentive to do so.

Then sometime next year we can simply remove API altogether.

I still apparently need to do some rubocop work, although this is made a bit challenging because rubocop no longer runs locally (complains about an invalid configuration).  I'll start hacking away at that soon as I can convince codeclimate to let me *see* the damned errors!  (grumble)

JoeCohen:
The Codeclimate stuff is mostly my fault. (It's almost all due to the RuboCop version upgrade. I probably should have held off on that.)
So I will take care of it tonight or tomorrow.

pellaea:
Actually, it's mostly because I cloned the API.  Now all existing
violations are considered "new".  If you could tell me how to fix the
config file locally, though, that would be a help.  Internet here is very
painfully slow today.

JoeCohen:
I don´t understand why RuboCop is complaining about configuration.
Are you running it on the VM or the host?

pellaea:
On the VM.  I get this error message:

Error: The `Style/MethodMissingSuper` cop has been removed since it has
been superseded by `Lint/MissingSuper`. Please use `Lint/MissingSuper`
instead. (obsolete configuration found in .rubocop_todo.yml, please update
it)

Thanks!

JoeCohen:
I think the following will work:
in .rubocop_todo.yml,
change Style/MethodMissingSuper to Lint/MissingSuper

pellaea:
It's in .rubocop_todo.yml.  That worked, thanks!

pellaea:
The last three issues apparently are pre-existing TODOs, and I have no intention of fixing those now.

JoeCohen:
Would it be okay if I replaced each of the preexisting TODOS with:
- A Pivotal Tracker ticket, plus
- A comment in the code referencing the ticket?

pellaea:
Sure, no difference to me.

JoeCohen:
thanks. Will do.

PR #661, opened 2020-09-06
Fix bug introduced when normalizing ArticlesController

JoeCohen:
Fix RssLog#url for articles

- Test rss_log.url for normalized controllers
- Add AbstractArticle#show_path
- Correct rss_log.url for article. (This was overlooked in #629.)
- Add rss_log article fixture (just in case).
- Delivers https://www.pivotaltracker.com/story/show/174394837

**This is a re-do of closed PR #641. See conversation there for some Comments.**

There is no manual test.

### DIFF
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -381,26 +381,6 @@ class AbstractModel < ApplicationRecord
     self.class.show_action
   end
 
-  # Return the path of the "show_<object>" action
-  #
-  #   # normalized controller
-  #   Article.index_action => "articles/12"
-  #
-  #   # unnormalized controller
-  #   Name.show_url(12) => "http://mushroomobserver.org/name/show_name/12"
-  #   name.show_url     => "http://mushroomobserver.org/name/show_name/12"	  #
-  def self.show_path(id)
-    if controller_normalized?(name)
-      "/#{show_controller}/#{id}"
-    else
-      "/#{show_controller}/#{show_action}/#{id}"
-    end
-  end
-
-  def show_path
-    self.class.show_url(id)
-  end
-
   # Return the URL of the "show_<object>" action
   #
   #   # normalized controller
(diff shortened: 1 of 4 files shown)


JoeCohen:
Current open issues:
1. Is `RssLog#url` used anywhere? If not, then it's dead and it and this PR should bite the dust. See https://github.com/MushroomObserver/mushroom-observer/pull/641#discussion_r473230626
2. Is the test method circular. See https://github.com/MushroomObserver/mushroom-observer/pull/641#discussion_r473247402

JoeCohen:
One more question:
**Should I delete `AbstractModel`#show_path` (class & instance methods)**
I don't remember why I added then. Probably just superstitious coding. Maybe I was thinking about refactoring `show_url`.
They are untested (that's what causes the Coveralls failure) and appear unused. (All tests pass without them.)

mo-nathan:
> One more question:
> **Should I delete `AbstractModel`#show_path`(class & instance methods)** I don't remember why I added then. Probably just superstitious coding. Maybe I was thinking about refactoring`show_url`.
> They are untested (that's what causes the Coveralls failure) and appear unused. (All tests pass without them.)

Yes.  As far as I can tell it is dead code.

mo-nathan:
When trying to see what this was actually fixing, I finally traced it down to the ancient RSS feed: http://mushroomobserver.org/observer/rss.  I sincerely doubt that anyone is actually using this these days.  However, to see the bug you can do the following:

1) Create a Glossary Term (click 'Glossary' in the left-hand nav, followed by 'Create New Glossary Term')
2) Go to observer/rss.  This will generate an XML RSS feed.  Embedded in that feed you will see the URL for the glossary term.

Current behavior:
URL shows up with '/glossary_term/' in it.

Fixed behavior:
URL shows up with '/glossary_terms/' in it.

mo-nathan:
Rubocop and spec coverage check out.  Confirmed fixed functionality.  This can merge as far as I'm concerned.

PR #664, opened 2020-09-10
Fix Useless Assignments

JoeCohen:
Re-enables the Lint/UselessAssignment cop without creating a backlog of offenses
- Manually fix all current offenses;
- Remove the cop from the todo list.

Notes:
- I have only one particular question. See https://github.com/MushroomObserver/mushroom-observer/pull/664#discussion_r486648331
- This is an older cop.
- I think we'd all agree it's a desirable cop. Useless assignments are at best dead code and may signal a problem elsewhere in the code, so we'd like to get notice of potential problems.
- The cop was silently disabled, thereby hiding and accumulating new offenses.
- I re-enabled the cop by removing it from the todo list. 
- I also fixed all the outstanding offenses so that we wouldn't suddenly see a batch of 60 new things to fix.
- The fixes are manual because there's no autocorrect feature. (Indeed there were a few false positives, which I "fixed" by disabling the cop in the source code.)

There is no manual test.
Delivers https://www.pivotaltracker.com/story/show/174567330

### DIFF
--- a/app/classes/query/modules/lookup_names.rb
+++ b/app/classes/query/modules/lookup_names.rb
@@ -80,10 +80,7 @@ module Query
           if /^\d+$/.match?(val.to_s)
             result << minimal_name_data(Name.safe_find(val))
           else
-            # rubocop:disable Lint/UselessAssignment
-            # cop generates a false positive
-            result += find_matching_names(val)
-            # rubocop:enable Lint/UselessAssignment
+            result + find_matching_names(val)
           end
         end.uniq.reject(&:nil?)
       end
--- a/app/classes/query/modules/lookup_names.rb
+++ b/app/classes/query/modules/lookup_names.rb
@@ -80,8 +80,10 @@ module Query
           if /^\d+$/.match?(val.to_s)
             result << minimal_name_data(Name.safe_find(val))
           else
-            # Cop appears to generate false positive
+            # rubocop:disable Lint/UselessAssignment
+            # cop generates a false positive
             result += find_matching_names(val)
+            # rubocop:enable Lint/UselessAssignment
           end
         end.uniq.reject(&:nil?)
       end
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -39,10 +39,7 @@ class ApiKey < AbstractModel
 
   def self.new_key
     result = String.random(KEY_LENGTH)
-    # rubocop:disable Lint/UselessAssignment:
-    # RuboCop gives false positive
-    key = String.random(KEY_LENGTH) while find_by(key: result)
-    # rubocop:enable Lint/UselessAssignment:
+    result = String.random(KEY_LENGTH) while find_by(key: result)
     result
   end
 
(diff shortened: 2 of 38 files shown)


app/models/api_key.rb
```diff
@@ -39,7 +39,10 @@ def verify!
 
   def self.new_key
     result = String.random(KEY_LENGTH)
-    key = String.random(KEY_LENGTH) while find_by_key(result)
+    # rubocop:disable Lint/UselessAssignment:
+    # RuboCop gives false positive
+    key = String.random(KEY_LENGTH) while find_by(key: result)
+    # rubocop:enable Lint/UselessAssignment:
     result
```
JoeCohen, line 46: I don't completely understand this. (My fault, not yours; I'm just not as fluent in Ruby as the rest of you.)
`key = String.random(KEY_LENGTH) while find_by(key: result)` seems quite elegant (though it took me a while to understand it).
**But is the return value correct?** 
In particular, I don't understand why it returns `result` rather than `key` when find_by(key: result) finds something.

app/classes/query/modules/lookup_names.rb
```diff
@@ -80,8 +80,10 @@ def find_exact_name_matches(vals)
           if /^\d+$/.match?(val.to_s)
             result << minimal_name_data(Name.safe_find(val))
           else
-            # Cop appears to generate false positive
+            # rubocop:disable Lint/UselessAssignment
+            # cop generates a false positive
             result += find_matching_names(val)
```
mo-nathan, line 85: I don't think this is a false positive, but is a bit subtle. += does two things.  In this case only the '+' is used not the '=' bit.  I believe you will get the exact same result if you change this to `result + find_matching_names(val)`.

JoeCohen:
Thanks Nathan.
I will have a look tonight or tomorrow a.m., if time and the fires allow.

app/models/api_key.rb
mo-nathan: I don't think this is correct.  I think the intended code is `result = String.random(KEY_LENGTH) while find_by(key: result)`.  I suspect we've been getting away it because `find_by_key(String.random(KEY_LENGTH)` is extremely unlikely to return true.

JoeCohen:
Thanks. That makes sense.

test/models/name_test.rb
```diff
@@ -2401,13 +2401,15 @@ def test_format_autonym
   def test_mysql_sort_order
     return unless sql_collates_accents?
 
-    # RuboCop gives false positives
+    # rubocop:disable Lint/UselessAssignment
+    # cop gives false positives
```
mo-nathan, line 2405: If it's complaining about n2, n3, and n4, then I would not consider these to be false positives.  I recommend changing this to something like:
```
names = [
    create_test_name("Agaricus Aehou"),
    create_test_name("Agaricus Aeiou"),
    create_test_name("Agaricus Aeiøu"),
    create_test_name("Agaricus Aëiou"),
    create_test_name("Agaricus Aéiou"),
    create_test_name("Agaricus Aejou")]
names[4].update(author: "aÉIOU"))
x = Name.where(id: names.map(&:id)).order(:author).pluck(:author)
assert_equal(%w[Aehou Aeiou Aëiou aÉIOU Aeiøu Aejou], x)
```
(Note I haven't tested the above, but hopefully you get the idea)

JoeCohen:
Yes, that's what it's complaining about. I will have a look.

PR #668, opened 2020-09-24
Softened block message, and made blocked IPs permanent.

pellaea:
This is in response to the jokers who we finally defeated a few months ago after weeks and weeks of battle, and who have now returned now that their IPs have been automatically cleared after a period of no use.  Apparently, we need to keep those IPs permanently on the block list.  Too bad I didn't keep a copy of the original list.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -163,7 +163,10 @@ class ApplicationController < ActionController::Base
     logger.warn("BLOCKED #{request.remote_ip}")
     msg = "We have noticed an excessive amount of server-intensive " \
           "traffic from this IP address (#{request.remote_ip}). " \
-          "Please contact the webmaster (#{MO.webmaster_email_address})."
+          "Please contact the webmaster (#{MO.webmaster_email_address}) " \
+          "so that we can see if there is a better way to do what you are " \
+          "trying to do.  If you believe you have been mistakenly blocked, " \
+          "please contact us so that we can remove the block."
     render(plain: msg,
            status: :too_many_requests,
            layout: false)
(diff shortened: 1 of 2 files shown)


JoeCohen:
If it's going to be permanent, should it go in the IP tables?

pellaea:
My thoughts are:

1. Automating changes to the firewall is really scary.  If you screw it up,
you can wind up permanently blocked from your own server and you basically
have to rebuild it from scratch.  Very scary.  I've seen Alan do something
clever where he makes a simple script that will automatically restore the
previous version after several seconds if he doesn't press control-C.  If
the change is okay, no problem, he can press control-C right away and the
change is permanent.  But if it goes wrong and accidentally blocks you,
then you cannot press control-C (you're blocked!) so it will automatically
revert in a few seconds.  Very clever.  But I'm way too lazy to do this
myself! :)

2. Anyone blocked by the firewall gets no useful error message; but when
rails blocks a request it sends that "we've noticed excessive traffic...
please contact us" message.  I still naively hold out hope that at least
some people who exceed our limits are legitimate and will actually contact
us.  Actually this has happened repeatedly over this summer, and it has led
to quite a number of people getting the data they need far more efficiently
and satisfactorily.

Of course, you could make an argument that any IP address that stays on the
block list for over a year, say, or some other long period of time, should
be transferred to the firewall permanently.  The problem right now is that
it would take additional effort to discover which of these IP addresses I
just restored to the block list are actually ones we can safely move to the
firewall.  I may well have accidentally re-added some IPs that were later
successfully resolved.  I definitely don't want to accidentally further
annoy them by adding them to the firewall!

3. The rails blocked ips list is easily manageable via the website
(/account/blocked_ips?  or something like that).  The firewall is a pain to
modify by hand.

4. I really doubt that traffic blocked at rails is causing any significant
drain on resources.  Feel free to prove me wrong, and you know my
preoccupation with performance, I would be very interested in addressing
the problem if it turned out that it *is* a serious performance issue.  I
just doubt it because the block list is cached per server instance and it
is one of the first before_filters that executes.  No database use at all,
and no rendering time at all.  Can't be that bad, surely.

-Jason

AlanRockefeller:
Jason is right on all points.   It's nice to be able to give a message
rather than silently blocking people.    The web server is never very
taxed, it's always the DB that slows down, and requests blocked by the web
server never make it to the database.

JoeCohen:
Thank you both for the explanation.

PR #669, opened 2020-09-24
Add nomenclature identifier field to Names

JoeCohen:
Adds a field for the identifier -- the registration number in fungal nomenclature databases.

- Adds an ICN Identifier field to the Name create/edit form
- Displays the results of that field on the Name page
- If the Name lacks an identifier, the Name page 
  - notes that it's missing and includes links to Index Fungorum and MycoBank searches, or
  - if the Name is clearly unregistrable (e.g., a plant) does not display anything related to the ICN number.

Delivers [PIvotal Story #167748352](https://www.pivotaltracker.com/story/show/167748352)

Suggested manual test:
- View a fungal name that does not have an identifier.
Result: Nomenclature section should display an Identifier missing message, links to IF and MB searches
- Edit that Name, adding the correct Identifier
Result: Nomenclature section should display the identifier, links to the IF and MB records. (MB actually links to the results of a search for the identifier.)


app/controllers/application_controller.rb
```diff
@@ -797,7 +797,7 @@ def flash_error(*strs)
   def flash_object_errors(obj)
     return unless obj&.errors && !obj.errors.empty?
 
-    flash_error(obj.formatted_errors)
+    obj.formatted_errors.each { |error| flash_error(error) }
```
JoeCohen, line 800: Fixes a bug. `object.formatted.errors` is an array of strings, rather than a simple string.

pellaea:
Cool, thanks, Joe!  I'll take a look at this soon as I can.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -139,30 +143,42 @@ def perform_merge_names(new_name)
...
+    unless minor_change?
+      email_admin_name_change(
+        old_identifier: Name.find(@name.id).icn_id
+      )
+    end
     @name.attributes = @parse.params
   end
 
-  def set_locked_citation_and_notes
+  def set_unparsed_attrs
```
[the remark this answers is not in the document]
JoeCohen: I know of no reasonable way to make `set_unparsed_attrs` simpler so that there's no ABC offense.

**Any suggestions?**

app/helpers/object_link_helper.rb
```diff
@@ -51,47 +51,49 @@ def name_link(name, str = nil)
...
-  # input is I18n language abbreviation
-  # return html parameter of official Mycobank translation,
-  # if such translation exists, else return pseudo-English parameter
-  # Although MycoBank doesn't recognize &Lang=Eng, this (or another language
-  # parameter else which MycoBank does not recognize) must be be included when
-  # switching to the default MycoBank language (English); otherwise MycoBank
-  # keeps using the last language it did recognize.
-  def mycobank_language_suffix(lang)
-    "&Lang=" + i18n_to_mycobank_language.fetch(lang, "Eng")
+  def mycobank_search_path
```
JoeCohen, line 87: MycoBank has a new UI. The paths have changed, and it ignores a language query.

app/views/name/_name.html.erb
```diff
@@ -1,67 +0,0 @@
-<p><%= :RANK.t %>: <%= name.rank ? rank_as_string(name.rank) : :unknown.t %></p>
```
JoeCohen, line 1: This partial was renamed to `_nomenclature`

app/views/name/show_name.html.erb
```diff
@@ -84,9 +84,6 @@
             <%= content_tag(:p, link_to(
                   "MyCoPortal", mycoportal_url(@name), target: :_blank
                 )) %>
-            <%= content_tag(:p, link_to(
```
JoeCohen, line 87: This link was moved to `_nomenclature`

JoeCohen (commented):
I will be away most of Thu Sep 24

JoeCohen:
I'd welcome any suggestions on the code.
I don't have a reasonable way to fix the Codeclimate issues, so I'm planning to ignore them.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -128,8 +129,11 @@ def merge_names(new_name)
 
   def perform_merge_names(new_name)
```
JoeCohen, line 130: **Do you a 10,000 ft. view how to refactor `perform_merge_names`?** 
(It badly needs refactoring. I've tried a few times, but always break the tests. I may try again tomorrow.)
**Would it make sense to move it to a separate file because of this offense: `(Metrics/ClassLength: Class has too many lines. [263/250]`?**

app/controllers/application_controller.rb
pellaea: Good catch.  Another way to solve this would be just to add a star:

    flash_error(*obj.formatted_errors)

Although, honestly, your way is more clearly going to work.

app/controllers/name_controller/create_and_edit_name.rb
pellaea: While it's well-known that I have a far greater tolerance for long and/or complex methods and files (and therefore think it's fine the way it is -- "if it ain't broke don't fix it"), I accept that others feel differently.

The short answer, I think, is to notice that there are three comments in perform_merge_names: if it needs a comment, maybe the bits of code that require a comment should be moved into their own methods.  Hell you could even automate it by turning the comment into a method name!  And yeah, move the methods that are exclusively used by the merge to a new file.

But yuck, we already started that process with "set_locked_citation_and_notes".  Already I have to jump around the code to understand how this method works.  At least in that case, that trivial method is used in two other places as well, so it satisfies a DRY itch.

No, I think the long answer involves recognizing the many bad smells coming off of this code and therefore realizing that a major refactor is probably required to make it more natural and elegant.  And that means expanding the realm of refactor to include Name and revisiting where various logic belongs.  We have a long history of randomly throwing logic in wherever we happen to be working, without asking ourselves "where does this really belong?"  Which logic belongs in the controller?  Which in the model?  (View isn't a concern in this case at least.)

So for example, the actual merge, as I'd call it, is buried hidden toward the bottom of perform_merge_names -- "new_name.merge(@name)".  That's a model method.  Why aren't things like deprecation status, author, order, etc. also handled in the model?  The superficial reasoning is clear enough: everything requiring access to the parser or params is being handled by the controller, the rest is left for the model.  But in all honesty, this action doesn't actually require -- or shouldn't! -- either.

Maybe, therefore, the "earliest" problem is simply that there is no dedicated request for merging two names, and instead is handled as a screwy side-effect of changing the name of an existing name!  (I admit full responsibility for this state of affairs.)  Then, that removes the parameters and parser BS in this perform_name_merge method, allowing most of this to move into the model where (it seems to me) it really belongs.  When I think about the general problem -- how do you merge two names? -- I can't think of any compelling reason there should be any user options!  Either it can happen or it can't.  Let the model decide which to keep and which to delete, which data to keep in the final merged name, and so on.

But is this really what this PR is about?  Of course not.  We're talking about a *major* refactor at this point, none of which has anything to do with Mycobank identifiers.

So I repeat: "don't fix what ain't broke".  Of course, that's just my opinion on the subject.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -139,32 +143,52 @@ def perform_merge_names(new_name)
...
     new_name.change_deprecated(change_deprecated) unless change_deprecated.nil?
-    @name.display_name = old_display_name_for_log
+
+    # move associations from @name to new_name and destroy @name
     new_name.merge(@name)
     args = { this: @name.real_search_name, that: new_name.real_search_name }
     flash_notice(:runtime_edit_name_merge_success.t(args))
+
     @name = new_name
+    if new_name.icn_id != old_identifier
```
pellaea, line 160: Should this also make sure old_identifier is not blank?  Do we want an email notification if the merge results in a missing ICN id being filled in?

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -139,32 +143,52 @@ def perform_merge_names(new_name)
...
     @name.save
   end
 
   def set_name_author_and_rank
     return unless name_unlocked?
 
-    email_admin_name_change unless minor_change? || just_adding_author?
+    unless minor_change?
+      email_admin_name_change(
+        old_identifier: Name.find(@name.id).icn_id
```
pellaea, line 171: I feel like there should be a better way to do this than querying the database.  Doesn't ActiveRecord give access to what the old state of a parameter is up until the change is saved?  Maybe something like @name.icn_id_was or something like that?

Also, I feel like ICN id is somewhat equivalent to author in this sense.  If the user is just adding an ICN id, I don't see that as worth an admin notification, the same way adding an author doesn't warrant a notification.  (The logic is that I want to be made aware of potentially damaging changes, but adding data where there was none before -- adding an author to a name which had no author, or adding an ICN id to a name without one --  is nondestructive, so I don't care about keeping an eye open for potential abuse there.)

So I would personally recommend reverting this and doing it more like it was before:

email_admin_name_change unless minor_change? || just_adding_author_or_icn_id?

If that can be done.

pellaea: Sorry, I see that minor_change? now includes just_adding_author?, and that probably makes sense.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -253,19 +284,28 @@ def name_unlocked?
...
+    return true if just_adding_author?
+
     old_name = @name.real_search_name
     new_name = @parse.real_search_name
     new_name.percent_match(old_name) > 0.9
   end
 
+  def icn_id_changed?
+    params[:name][:icn_id] != @name.icn_id
+  end
```
pellaea, line 297: I would call changing the ICN id a nontrivial change.  That could be very damaging for any agents that use the ICN id and expect it to be accurate.  I would only consider adding an ICN id where there was none before "minor".

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -253,19 +284,28 @@ def name_unlocked?
...
+
   def just_adding_author?
     @name.author.blank? && @parse.text_name == @name.text_name
   end
 
-  def email_admin_name_change
+  def email_admin_name_change(old_identifier: nil)
```
pellaea, line 303: It might be possible to avoid passing old_identifier around.  If email_admin_name_change is only ever called before the changes are saved, ActiveRecord provides access to the old value of icn_id.

app/models/name.rb
```diff
@@ -397,4 +411,33 @@ def self.count_observations(names)
...
+  def icn_id_unique
+    return if icn_id.nil?
+    return if (conflicting_name = other_names.first).blank?
+
+    errors[:base] << :name_error_icn_id_in_use.t(
+      number: icn_id, name: conflicting_name.real_search_name
+    )
+  end
+
+  def other_names
```
pellaea, line 440: Please elaborate name better, e.g., "other_names_with_same_icn_id".

app/helpers/object_link_helper.rb
```diff
@@ -51,47 +51,49 @@ def name_link(name, str = nil)
...
   end
 
+  # ----- links to names and records at external websites ----------------------
+
   # Create link for name to MyCoPortal website.
   def mycoportal_url(name)
-    "http://mycoportal.org/portal/taxa/index.php?taxauthid=1&taxon=" +
-      name.text_name.tr(" ", "+")
+    "http://mycoportal.org/portal/taxa/index.php?taxauthid=1&taxon=" \
+      "#{name.stripped_text_name.tr(" ", "+")}"
```
pellaea, line 59: I might not understand this all correctly, but I'm concerned that the stripped part here might be wrong.  Example from lichens is "Lecidea atrobrunnea group".  This should never be given an ICN id, nor should it ever have an entry at Mycoportal.  At least I don't think so.  In any case it should in no circumstances take you to the entry for "Lecidea atrobrunnea" at Mycoportal!

app/models/name/format.rb
```diff
@@ -71,13 +71,17 @@ def real_search_name
...
 
+  def stripped_text_name
+    text_name.gsub(/([."]|group|clade)/, "").strip.squeeze(" ")
+  end
+
   def self.display_to_real_text(name)
-    name.display_name.gsub(/ ^\*?\*?__ | __\*?\*?[^_\*]*$ /x, "").
-      gsub(/__\*?\*? [^_\*]* \s (#{ANY_NAME_ABBR}) \s \*?\*?__/x, ' \1 ').
+    name.display_name.gsub(/ ^\*?\*?__ | __\*?\*?[^_*]*$ /x, "").
+      gsub(/__\*?\*? [^_*]* \s (#{ANY_NAME_ABBR}) \s \*?\*?__/x, ' \1 ').
```
pellaea, line 80: Backslash for star is just not necessary inside square brackets?

app/models/name/merge.rb
```diff
@@ -106,6 +112,9 @@ def merge(old_name)
       self.citation = old_name.citation.strip_squeeze
     end
 
+    # Update the identifier if user provided one
+    self.icn_id = old_name.icn_id if old_name.icn_id
```
pellaea, line 116: I seem to remember the general logic here is that where a data field is provided by both names, it prefers self over old_name.  That's because usually self has more references / uses in the database, and that old_name is often a one-off typo or mistake.  Self should definitely be the more reliable source.

Therefore I recommend:

    self.icn_id = old_name.icn_id if self.icn_id.blank?

app/models/name/taxonomy.rb
```diff
@@ -83,6 +83,66 @@ def has_eol_data?
...
+  def unpublished?
+    /\b(nom prov|comb prov|sensu lato|ined)\b/i =~ author&.delete(".")
+  end
+
+  # Kingdom as a string, e.g., "Fungi", or nil if no Kingdom
+  def kingdom
+    parse_classification.find { |rank| rank.first == :Kingdom }&.last
+  end
+
+  public
```
pellaea, line 142: Impressive bit of logic!  I'll trust you that it works in most cases.  And for the record I am fine with erring on the side of calling things registrable / searchable even when they aren't.

The other way around could be quite annoying.  That is, omitting a search link where it is perfectly possible to include one, but we just don't think it would work because it's a screwy case.  I'd much rather erroneously include a broken link in some or even all cases instead of omitting valid links for unusual names.

Of course, I could understrand people disagreeing with me, though.  (For example, I can imagine some users complaining "MO is broken!" because it includes a broken link for some screwy name.  Ugh.)

app/views/name/_form_name.html.erb
```diff
@@ -10,14 +10,27 @@
...
+      <%= form.select(
+            :rank, Name.all_ranks.map {|r| [rank_as_string(r), r]}, {},
+            {class: "form-control"}) %>
     </div>
+
     <div class="form-group indent">
       <label for="name_deprecated"><%= :Status.t %>:</label>
-      <%= form.select(:deprecated, [[:ACCEPTED.t, false], [:Deprecated.t, true]],{},{class: "form-control"}) %>
+      <%= form.select(
+            :deprecated, [[:ACCEPTED.t, false], [:Deprecated.t, true]],{},{
```
pellaea, line 32: Not that it makes a difference in this case sense all translations of "accepted" and "deprecated" are probably single words, but... it's very strange that it would request all caps for "accepted" but not for "deprecated"!  We should choose one or the other and be consistent.  (Probably my fault, but since you're fixing so many preexisting stylistic issues already, might as well nail this one, too!)

app/views/name/_form_name.html.erb
```diff
@@ -71,6 +82,7 @@
 
 <p>
   <label for="notes"><%= :form_names_taxonomic_notes.t %>:</label><br/>
+  <span class="mark""><%= :form_names_taxonomic_notes_warning.t %></span>
```
pellaea, line 85: The extra double-quote must be a typo here.

test/models/name_test.rb
```diff
@@ -2881,4 +2888,78 @@ def test_make_sure_names_are_bolded_correctly
...
+    name = Name.new(text_name: "Agaricus pinyonensis",
+                    author: "Isaacs nom. prov.")
+    assert(name.searchable_in_registry?,
+           "Provisional names can be searchable in registry")
+
+    name = Name.new(text_name: "Myxomycetes", rank: :Class,
+                    classification: "Kingdom: Protozoa")
+    assert(name.searchable_in_registry?,
+           "Protozoa should be searchable in registry")
+  end
```
pellaea, line 2964: Huge respect for actually providing useful failure messages!  I wish we all would do that.  Would save so much time parsing tests to figure out what it's supposed to be doing and why it's failing!

pellaea:
This is good work, Joe!  Any changes to Name at this point are going to be unnecessarily complex and difficult (as you just discovered!)  This is some of the oldest code in MO.  But you did a great job surgically adding just the bits that you needed to without affecting anything else.  Not a trivial job.

JoeCohen:
Thanks for the extensive, thoughtful, and helpful comments.
I'll try to dig into them later today or tomorrow.

app/controllers/application_controller.rb
JoeCohen: Thanks! I hadn't thought about adding a star. 
Per your comment I'll leave it as explicitly iterating over the object's formatted_errors

app/controllers/name_controller/create_and_edit_name.rb
JoeCohen: You're right. I will re-work the code so that there's no admin notification when merge results merely filling in a missing ICN id.

app/controllers/name_controller/create_and_edit_name.rb
JoeCohen: 1. I'll look into how to avoid hitting the db. It did bother me that I was doing this, but I was too lazy/tired to figure out how to avoid it.
2. I won't notify admin if the only change is filling in a blank ICN id

app/controllers/name_controller/create_and_edit_name.rb
JoeCohen: Thank you for the suggestion.
(I disliked adding the `old_identifier` flag: I don't like flags in the first place, and this method is already too complicated.)

app/models/name.rb
JoeCohen: Will do.

app/helpers/object_link_helper.rb
JoeCohen: You are correct on both counts. I will simplify things so that it will work as follows:
- "_Lecidea atrobrunnea_ group" can never be given an `icn_id`. 
(It does currently work this way. Names are validiated before_save for registrability. Anything with "group" (or with a bunch of other issues) is unregistrable, and will throw an object error.)
- I'll get rid of `stripped_name`. 
  - The MyCoPortal entry point (if any) will be the text_name, if the name is `searchable`?. (There will be no MyCoPortal link If the name is `unsearchable?`, e.g., _Pseudotsuga menziesii_.)
  - The IF and MB entry points will be:
    - The record, if there's an ICN number
    - A search for the `text_name` if the name is registrable, e.g., _Lecidea atrobrunnea_
    - The general IF or MB search if the name is searchable, but not registrable. _Lecidea atrobrunnea_ group
    - (No IF or MB link if the name is unsearchable.)
I might move some of the link stuff to NameHelper.

app/controllers/name_controller/create_and_edit_name.rb
mo-nathan: In case you decide you want to refactor it anyway, it strikes me that there there are three sections to this code.  1) the lines before the potential swap, 2) the swap, and 3) the cleanup afterwords.  Reading the code, it's not clear to me why there is so much happening before the potential swap and if that is really the expected behavior.  If I were going after this method I would try to closely follow the process outlined in Fowler & Beck's book "Refactoring" (https://martinfowler.com/books/refactoring.html) where you make really small changes and then verify that the tests continue to pass.  I might start with seeing if you can move the swap logic earlier.

app/models/name/format.rb
mo-nathan: Wasn't aware of that either, but it makes sense to me when I think about it.  I've known for a long time that `-` only has a special meaning in character classes if it is actually between two characters so if I want to include '-' in an explicit character class, then I just throw it at the end of the list (e.g., [a-z_-]).  I found all of this behavior described here: https://www.regular-expressions.info/charclass.html, in the Metacharacters Inside Character Classes section.  It also describes some interesting behavior regarding ']' that I wasn't aware of.

JoeCohen: And I hadn't been aware of it. But RuboCop has known about it for a little while. https://docs.rubocop.org/rubocop/0.85/cops_style.html#styleredundantregexpescape

app/controllers/name_controller/create_and_edit_name.rb
JoeCohen: Thank you both for the comments and the 10,000' views.
It's not broken in the sense that all the tests pass. But it's quite complicated -- it took me multiple hours and re-reads, and some experimentation to understand it to some extent. And I made it worse.
But my current thinking is to make the refactor a separate, follow-up story. 
In addition to the ideas that each of you suggest, it would help me to change/add more expressive variable names, e.g., things like `surviving name` `name_to_be_merged`, name_being_edited` `found_oame`

app/controllers/name_controller/create_and_edit_name.rb
JoeCohen: Agree. See above.

app/views/name/_form_name.html.erb
JoeCohen: I agree. I will make them consistent.

app/views/name/_form_name.html.erb
JoeCohen: Good catch. I will fix.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -223,73 +229,131 @@ def parse_name
...
+      perform_merge_names(new_name)
+      redirect_to_show_name
     else
-      redirect_with_query(action: :approve_name, id: @name.id)
+      redirect_to_merge_request(new_name)
+    end
+  end
+
+  def perform_merge_names(new_name)
+    old_display_name_for_log = @name[:display_name]
```
JoeCohen, line 303: I don't understand the 3 lines involving @name[:display_name] and @name.display_name: this line, line 315, and line 317 --
- Why is the attribute being accessed via a hash key here and in line 315?
- Why do we need these lines at all?

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -223,73 +229,131 @@ def parse_name
...
+    @name.attributes = @parse.params
+    set_unparsed_attrs
+    # Only change deprecation status if user explicity requested it.
+    if @name.deprecated != (params[:name][:deprecated] == "true")
+      change_deprecated = !@name.deprecated
+    end
+    # Automatically swap names if that's a safer merge.
+    if !@name.mergeable? && new_name.mergeable?
+      @name, new_name = new_name, @name
+      old_display_name_for_log = @name[:display_name]
```
JoeCohen, line 315: See comment above.

app/controllers/name_controller/create_and_edit_name.rb
```diff
@@ -223,73 +229,131 @@ def parse_name
...
+    end
+    @name.display_name = old_display_name_for_log
```
JoeCohen, line 317: See comment above.

app/controllers/name_controller/create_and_edit_name.rb
pellaea: Check out app/models/name/format.rb.  Apparently we override display_name.

JoeCohen:
Thanks Jason!
I'll check it out this evening.
Joe

JoeCohen:
I think I've responded to all the suggestions. Thanks again!
I don't plan to deal with [the remaining Codeclimate issue](https://codeclimate.com/repos/58cfc16d1035280287001d93/pull/669) (too many lines in class `NameController`).
I did refactor `perform_merge_names` to fix the ABC offense for that method.  
**Question: Does the refactor work?** Tests pass, but I'm concerned about my elimination of the `old_display_name_for_log` flag; that change could use another set of eyes.

pellaea:
I'll take a look at this tomorrow.

app/views/name/_form_name.html.erb
pellaea: I'm so confused.  I still see the extra double-quote.  Did you forget to fix it, or is git still showing the state before the last comment?  If the latter, that's scary, because I've been assuming this diff page on github shows the latest!

JoeCohen:
You're not confused; I forgot to fix it.

pellaea:
Ah!  I get confused easily when picking up loose threads more than a couple
of days old! :)

pellaea:
This all looks good to me.  Ready to merge whenever you are.

JoeCohen:
Super! Thanks so much.
I'll get it merged.

PR #670, opened 2020-09-30
Tweaked excessive traffic block message.

pellaea:
I am trying to word this better so that innocent bystanders are not put on the defensive.


JoeCohen:
That looks great. 
Thank you, Nastassja!

PR #671, opened 2020-10-03
Fix publication_citation typo

JoeCohen:
- Change "sited" to "cited"
- Also update and tighten wording.

Manual test:
- Go to https://mushroomobserver.org/publications
Result: 
It should display (emphasis added) "... websites can be **cited**.", not "... websites can be **sited**."

Delivers https://www.pivotaltracker.com/story/show/174119049.


JoeCohen:
My real motive for this PR: 
See if commit signing is working for me and integrated with my Git GUI.
Screen Shot 2020-10-03 at 12 38 30

pellaea:
I don't understand this verified business, but the change in language looks good to me.

JoeCohen:
Thanks for taking the time to looki at this.
Verified == has a verified PGP signature.
Not important. But I noticed that Nathan was doing it sometimes, and I
wanted to be just like the pros. ☺

PR #672, opened 2020-10-04
Fixed broken link in show_species_list.

pellaea:
No tests or review, just a quick bug fix.


JoeCohen:
Can you also please fix the User who complained about this? :-(

pellaea:
Way beyond the scope!  He was right in this case, though.

JoeCohen:
He is right occasionally. Thanks for the fix!

PR #673, opened 2020-10-04
Fix small synonym-related bug

pellaea:
Somehow or another we have gotten a bunch of name with a synonym_id but no corresponding entry in the synonym table.  Even though this shouldn't matter at all, it causes name.synonyms to return nil when name.synonym_id does not.

I have fixed the database (removed all unused synonyms, added missing synonyms).

This fixes a case in the API which calls name.synonym.names to make it more robust for this sort of failure.

No review or anything should be required here.  Very trivial change.

### DIFF
--- a/app/views/api/_name.json.builder
+++ b/app/views/api/_name.json.builder
@@ -15,7 +15,7 @@ json.ok_for_export   object.ok_for_export ? true : false
 if !detail
   json.synonym_id object.synonym_id if object.synonym_id
 else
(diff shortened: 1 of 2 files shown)


pellaea:
I checked the code for anywhere else we actually use the synonym object.  There is only one case, in app/models/name/synonymy.rb, and it explicitly ensures that synonym exists before doing so.  (It is the bit of code responsible for deleting unused synonyms.)

PR #674, opened 2020-10-04
Cover HerbariumController

JoeCohen:
- Achieve 100% coverage of HerbariumController;
- Also fixes all RuboCop offenses in HerbariumController.

Manual Test: check Coveralls or SimpleCov to confirm 100% coverage of `app/controllers/herbarium_controller.rb`

This is the first step in "normalizing" HerbariumController. See https://www.pivotaltracker.com/story/show/174310149. 
I thought I should get this step into `master` because it's independently useful and I don't know how long that Story will take.

### DIFF
--- a/test/controllers/herbarium_controller_test.rb
+++ b/test/controllers/herbarium_controller_test.rb
@@ -16,6 +16,18 @@ class HerbariumControllerTest < FunctionalTestCase
     }
   end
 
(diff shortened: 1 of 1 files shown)


pellaea:
This all looks fine to me, I think.

JoeCohen:
Thank you! I'll merge it.

PR #675, opened 2020-10-05
Fix bug in pattern search.

pellaea:
If the pattern string is empty it redirects to list_xxx to get full listing of all objects. But it was not pluralizing "herbarium" correctly and failing for that object type.

Again, just a trivial bug fix and simple corresponding test.  Review not really required.  But I am going to wait for Travis to run the tests for me.  My laptop is too busy with other intensive jobs right now to run them myself.

### DIFF
--- a/test/controllers/observer_controller_test.rb
+++ b/test/controllers/observer_controller_test.rb
@@ -590,6 +590,11 @@ class ObserverControllerTest < FunctionalTestCase
     params = { search: { pattern: "", type: :observation } }
     get_with_dump(:pattern_search, params)
     assert_redirected_to(controller: :observer, action: :list_observations)
+
+    # Make sure this redirects correctly to list_herbaria not list_herariums.
+    params = { search: { pattern: "", type: :herbarium } }
+    get(:pattern_search, params)
+    assert_redirected_to(controller: :herbarium, action: :list_herbaria)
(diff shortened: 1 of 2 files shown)


test/controllers/observer_controller_test.rb
```diff
@@ -590,6 +590,11 @@ def test_pattern_search
     params = { search: { pattern: "", type: :observation } }
     get_with_dump(:pattern_search, params)
     assert_redirected_to(controller: :observer, action: :list_observations)
+
+    # Make sure this redirects correctly to list_herbaria not list_herariums.
+    params = { search: { pattern: "", type: :herbarium } }
+    get(:pattern_search, params)
```
JoeCohen, line 596: Looks good!
Thanks for catching this and reducing the number of messages we get.
Suggestion if you feel like it:
```suggestion
    get(:pattern_search, params: params)
```
(See https://docs.rubocop.org/rubocop-rails/2.7/cops_rails.html#railshttppositionalarguments. This is one of the many kinds of offenses that's hidden because we have so many.)

pellaea:
Thanks, Joe.  I meant to do that, but forgot.  There are a zillion others
that need to be fixed in that file already, though, so I'm not too worried
about it! :)

PR #680, opened 2020-10-10
No admin email when ICN id unchanged

JoeCohen:
When comparing existing `icn_id` to `params[:name][;icn_id]`, convert existing ID to string.

Otherwise ICN ID always appears to change, and an admin email is sent (to webmaster).
That's what happens when you compare 5 apples to 5 oranges and expect them to be equal.

Delivers https://www.pivotaltracker.com/story/show/175158610


pellaea:
Good catch, thanks for fixing this, Joe.

PR #682, opened 2020-10-10
Changed add_comment layout so that the observation notes and comments…

pellaea:
… fit in the column below the comment text area to the left of the images if there are a lot of images.  Before they would get pushed down below the images leaving a lot of unnecessary dead space.  Also, instead of hiding the images for "xs" browsers, it moves them to the end.


JoeCohen:
Looks good to me.

PR #686, opened 2020-10-12
Added table to keep track of the last time each user saw each observation.

pellaea:
It is maintained solely by show_observation.  No automagic callbacks or anything.

Also fixed bug in footer of show_observation page to make it display the *previous* time the page was displayed, not *this* time.  That must have been there for years!

I'll do queries and other follow-on features as separate PRs.

I might also attempt to populate the table using the logs.  I can go back to beginning of 2019 on the server.  If it is worthwhile, maybe Nathan would be willing to run a script for me on the archived logs from < 2019?  Not sure when we started logging the user id with requests.  Maybe > 2019?

Probably need to do some rubocop work.  This should be considered a draft for the moment.

### DIFF
--- a/app/models/observation_view.rb
+++ b/app/models/observation_view.rb
@@ -8,7 +8,7 @@ class ObservationView < AbstractModel
   def self.update_view_stats(observation, user)
     return if observation.blank? || user.blank?
 
-    if view = find_by(observation: observation, user: user)
+    if (view = find_by(observation: observation, user: user))
       view.update!(last_view: Time.zone.now)
     else
       create!(observation: observation, user: user, last_view: Time.zone.now)
--- a/app/models/observation_view.rb
+++ b/app/models/observation_view.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Keep track of when each user last viewed each observation.
+class ObservationView < AbstractModel
+  belongs_to :observation
+  belongs_to :user
+
+  def self.touch(observation, user)
+    return if observation.blank? || user.blank?
+
+    if view = where(observation: observation, user: user).first
+      view.update_attributes!(last_view: Time.zone.now)
+    else
+      create!(observation: observation, user: user, last_view: Time.zone.now)
+    end
+  end
+end
--- a/app/models/observation_view.rb
+++ b/app/models/observation_view.rb
@@ -5,7 +5,7 @@ class ObservationView < AbstractModel
   belongs_to :observation
   belongs_to :user
 
-  def self.touch(observation, user)
+  def self.update_view_stats(observation, user)
     return if observation.blank? || user.blank?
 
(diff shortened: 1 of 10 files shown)


pellaea:
Okay, this is ready for review.  Too many more-important things to work on.  I'd like to get this merged ASAP.

There are no new controllers, so there's nothing to worry about re: conforming to the new normalized CRUD standards.  There *is*, however, a new (minimal) model (ObservationView).  Anything I need to look out for there?  I think this is our very first use of a "has_many through" relationship on MO, whee!

db/migrate/20201011132400_create_observation_views.rb
```diff
@@ -0,0 +1,13 @@
...
+    create_table :observation_views, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+      t.integer  :observation_id
+      t.integer  :user_id
+      t.datetime :last_view
+    end
+  end
+
+  def down
+    drop_table :observation_views
+  end
```
JoeCohen, line 12: This is a little shorter:
```suggestion
  def change
    create_table :observation_views, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
      t.integer  :observation_id
      t.integer  :user_id
      t.datetime :last_view
    end
  end
```

app/models/observation_view.rb
```diff
@@ -0,0 +1,17 @@
...
+
+# Keep track of when each user last viewed each observation.
+class ObservationView < AbstractModel
+  belongs_to :observation
+  belongs_to :user
+
+  def self.update_view_stats(observation, user)
+    return if observation.blank? || user.blank?
+
+    if view = find_by(observation: observation, user: user)
```
JoeCohen, line 11: To avoid RuboCop complaint:
```suggestion
    if (view = find_by(observation: observation, user: user))
```

JoeCohen (commented):
Looks good.
I don't have any substantive comments.

app/models/observation_view.rb
JoeCohen: PS: The cop is [Lint/AssignmentInCondition](https://docs.rubocop.org/rubocop/0.85/cops_lint.html#lintassignmentincondition)
You're not seeing a complaint because .rubocop_todo.yml is squelching the cop because we so many offenses (50).

db/migrate/20201011132400_create_observation_views.rb
pellaea: Thanks, Joe!  I wanted to do this, but after an hour of trying to access the docs for the correct syntax, I finally gave up and went with the old style. :)

app/models/observation_view.rb
pellaea: Thanks, didn't know, will do.

PR #687, opened 2020-10-13
Added an admin notification for new name tracking notifications with nonblank template.

pellaea:
This includes both new notifications with non-blank template, and updating existing notifications where the template changes from blank to non-blank.

I didn't add any new tests, because this really isn't important, and I don't expect it to stick around for long.

### DIFF
--- a/app/controllers/name_controller.rb
+++ b/app/controllers/name_controller.rb
@@ -1165,6 +1165,7 @@ class NameController < ApplicationController
         @notification.note_template = note_template
         flash_notice(:email_tracking_updated_messages.t)
       end
+      notify_admins_of_notification(@notification)
       @notification.save
     when :DISABLE.l
       @notification.destroy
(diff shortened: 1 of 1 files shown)


app/controllers/name_controller.rb
```diff
@@ -1175,6 +1176,21 @@ def submit_tracking_form(name_id)
     redirect_with_query(action: "show_name", id: name_id)
   end
 
+  def notify_admins_of_notification(notification)
+    return if notification.note_template.blank?
+    return if !notification.new_record? &&
+              notification.note_template_before_last_save.blank?
```
JoeCohen, line 1182: Does the method `note_template_before_last_save` already exist?

JoeCohen:
Looks good. Just the one question I added above.

app/controllers/name_controller.rb
pellaea: Yes, it's an automagic method.  Before you save an ActiveRecord instance, you can access the original value of any attribute "attr" by tacking on "_before_last_save", that is "instance.attr_before_last_save".  (It used to be "attr_was", but that is now deprecated.)

JoeCohen:
Thanks! I'm glad to learn that.

PR #688, opened 2020-10-13
Hide map on obs form

pellaea:
Replace map with a button.  When user presses the button, the old map appears.  Google maps API is not loaded until the button is pressed.

Let's hear it for jQuery magic!  Couldn't have been easier.

Also, let's hear it for PRs that touch only views and javascript -- no rubocop or tests to contend with at all! :)


JoeCohen:
I can fix that -- in part.
Codeclimate has a javascript linter which we can enable. :-)

pellaea:
Bite your tongue, you bad, bad man!

PR #689, opened 2020-10-14
Update MB record url 

JoeCohen:
Updates the url for links to MB by RecordNumber

MycoBank restored this form of link today. 
The restored url goes directly to the record, rather than to a search for the record.

###Manual Test
- Add a ICN Registration Identifier to a Name.
- On the resulting show_name page, follow the link to the MB record (it looks like `[#nnnnnn] MycoBank`)
Result: You land on the MycoBank record for that name.

Delivers https://www.pivotaltracker.com/story/show/175247423

### DIFF
--- a/app/helpers/object_link_helper.rb
+++ b/app/helpers/object_link_helper.rb
@@ -65,9 +65,8 @@ module ObjectLinkHelper
   end
 
   # url for MB record by number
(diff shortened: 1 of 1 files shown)


JoeCohen:
This is a one-liner, simply changing a string that's the url for an MB record.

PR #690, opened 2020-10-14
Name merger log

JoeCohen:
Fixes a bug in how Name mergers were logged:
- Shows the destroyed Name (rather than the surviving Name) In the "Name Destroyed" log entry.

Delivers https://www.pivotaltracker.com/story/show/175219034


pellaea:
Looks good to me.

JoeCohen:
Jason:
Thanks for looking at this.
I'll hold up on merger until the sudo branch is straightened out.
Joe

PR #692, opened 2020-10-15
Revising some stuff in project controller

pellaea:
Project add_members was listing all verified users -- *way* too many.  Limited it to the last 100 to login, and added a write-in autocompleted field.

Added flash notices for both add_members and change_status.

(Added autocomplete and flash notices for Switch Users, too.)

Should be pretty by-the-numbers, review probably not necessary, but welcome if anyone has time.

### DIFF
--- a/app/controllers/project_controller.rb
+++ b/app/controllers/project_controller.rb
@@ -294,15 +294,27 @@ class ProjectController < ApplicationController
   # "Posts" to the same action.  Stays on this view until done.
   def add_members
     pass_query_params
-    if @project = find_or_goto_index(Project, params[:id].to_s)
-      @users = User.where.not(verified: nil).order("login, name").to_a
-      if !@project.is_admin?(@user)
-        redirect_with_query(action: "show_project", id: @project.id)
-      elsif params[:candidate].present?
-        @candidate = User.find(params[:candidate])
-        set_status(@project, :member, @candidate, :add)
-      end
+    return unless @project = find_or_goto_index(Project, params[:id].to_s)
+    return must_be_project_admin!(@project.id) unless @project.is_admin?(@user)
+
+    @users = User.order("last_login desc").limit(100).to_a
+    @candidate = params[:candidate]
+    add_member(@candidate, @project) if @candidate.present?
+  end
+
+  def add_member(str, project)
+    if (user = find_member(str))
+      set_status(project, :member, user, :add)
+      @candidate = nil
+    else
+      flash_error(:add_members_not_found.t(str))
     end
+  end
+
+  def find_member(str)
+    return User.safe_find(str) if str.to_s.match?(/^\d+$/)
+
+    User.find_by_login(str.to_s.sub(/ <.*>$/, ""))
   end
 
   # Form to make a given User either a member or an admin.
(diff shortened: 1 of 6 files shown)


app/controllers/project_controller.rb
```diff
@@ -294,17 +294,29 @@ def admin_request
...
     pass_query_params
-    if @project = find_or_goto_index(Project, params[:id].to_s)
-      @users = User.where.not(verified: nil).order("login, name").to_a
-      if !@project.is_admin?(@user)
-        redirect_with_query(action: "show_project", id: @project.id)
-      elsif params[:candidate].present?
-        @candidate = User.find(params[:candidate])
-        set_status(@project, :member, @candidate, :add)
-      end
+    return unless @project = find_or_goto_index(Project, params[:id].to_s)
```
JoeCohen, line 297: Sorry I didn't notice this earlier
```suggestion
    return unless (@project = find_or_goto_index(Project, params[:id].to_s))
```

app/controllers/project_controller.rb
```diff
@@ -294,17 +294,29 @@ def admin_request
...
+    return must_be_project_admin!(@project.id) unless @project.is_admin?(@user)
+
+    @users = User.order("last_login desc").limit(100).to_a
+    @candidate = params[:candidate]
+    add_member(@candidate, @project) if @candidate.present?
```
JoeCohen, line 302:
```suggestion
    return unless (@candidate = params[:candidate]).present?
    
    add_member(@candidate, @project)
```

test/controllers/project_controller_test.rb
```diff
@@ -403,6 +403,35 @@ def test_add_member_member
...
   end
 
+  def test_add_member_writein
+    eol_project = projects(:eol_project)
+    target_user = dick
+    assert_not(target_user.in_group?(eol_project.admin_group.name))
+    assert_not(target_user.in_group?(eol_project.user_group.name))
+    params = {
+      id: eol_project.id,
+      candidate: "dick <Tricky Dick>"
```
JoeCohen, line 413:
```suggestion
      candidate: "#{target_user.login} <#{target_user.name}>"
```
would make it easier to choose a different fixture if that were ever needed. I think also a bit less magical.

test/controllers/project_controller_test.rb
```diff
@@ -403,6 +403,35 @@ def test_add_member_member
...
+    }
+    requires_login(:add_members, params, mary.login)
+    assert_response(:success)
+    target_user = User.find(target_user.id)
```
JoeCohen, line 417: Is this in order to reload `target_user`?

JoeCohen:
Still looks good. Sorry that all I can offer is just picking around the edges. 
If that annoys you, please let me know.

test/controllers/project_controller_test.rb
```diff
@@ -410,12 +410,12 @@ def test_add_member_writein
...
       id: eol_project.id,
-      candidate: "dick <Tricky Dick>"
+      candidate: "#{target_user.login} <Should Ignore This>"
     }
     requires_login(:add_members, params, mary.login)
     assert_response(:success)
-    target_user = User.find(target_user.id)
-    assert_not(target_user.in_group?(eol_project.admin_group.name))
+    target_user.reload
+    assert_not(target_user.reload.in_group?(eol_project.admin_group.name))
```
JoeCohen, line 418: Do we need to reload twice?
(I think you could get rid of line 417.)

pellaea: Oops!  Will fix that.

PR #694, opened 2020-10-20
Handle improper add_comment request without crashing.

pellaea:
We got a whole string of error messages from someone requesting add_comment without an id.  Simple patch to handle this case without crashing.  Just have it redirect back or to the main page.

### DIFF
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -191,7 +191,9 @@ class CommentController < ApplicationController
   def add_comment
     pass_query_params
     @target = Comment.find_object(params[:type], params[:id].to_s)
-    if !allowed_to_see!(@target)
+    if !@target
+      redirect_back_or_default("/")
+    elsif !allowed_to_see!(@target)
       # redirected already
     elsif request.method == "GET"
       @comment = Comment.new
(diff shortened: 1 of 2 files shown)


app/controllers/comment_controller.rb
```diff
@@ -191,7 +191,9 @@ def prev_comment
   def add_comment
     pass_query_params
     @target = Comment.find_object(params[:type], params[:id].to_s)
-    if !allowed_to_see!(@target)
+    if !@target
+      redirect_back_or_default("/")
+    elsif !allowed_to_see!(@target)
```
JoeCohen, line 196: Would this work:
```suggestion
    redirect_back_or_default("/") unless @target
    
    if !allowed_to_see!(@target)
```

JoeCohen:
Thanks doing this. I've seen a bunch of strange errors lately, and wish I knew how to run them down and fix them.
Looks good! You ought to merge and put into production when you're ready.

app/controllers/comment_controller.rb
pellaea: Not quite, we'd need to return, too.  But look at the whole method, it's badly in need of renormalizing.  I figured it was better to stick with the style of the rest of the method for such a tiny, inconsequential bug fix.  Sorry, I should have explicitly stated that this PR is not worth wasting time on!

JoeCohen:
How about "return redirect_or_back unless ..." ?

PR #695, opened 2020-10-20
Multivalue search

pellaea:
Add support for multiple values in search bar.  It was mostly already done, but had never fully "hooked it up".

Also add support for multiple regions in query.

Joanne wanted this for a FunDiS initiative.

Example search strings:

name:"Genus species1","Genus species2"
region:"California, USA","Oregon, USA","Washington, USA"


JoeCohen:
This sounds great. Thanks Jason (and Joanne).
I won't be able to look at this tonite. I'll try for tomorrow afternoon.

pellaea:
Thanks, Joe.  Content filters are so confusing!  I will be out tomorrow,
but I *would* like to get this out sooner than later for Joanne.  I'm not
expecting the average user to make use of this feature (which was why I
never bothered to support it from the start).  But it is useful as a "saved
search".  We need to engineer a complicated search query for the
investigators on this FunDiS project with 10 different names and 4
different states and provinces.  Way too much to type by hand, but as a
tinyurl it should be just fine.  Later on we can do "real" saved searches,
but for now this gets the project on its feet.

JoeCohen:
Looks great. Go for it. I don't have anything to add.

pellaea:
Thanks, Joe.

PR #698, opened 2020-11-01
Fixes for the clean.sql script

mo-nathan:
To test:

   `mysql -u root -p mo_development < db/clean.sql`

should not report any errors.  This is the script that takes a production database to a point where it doesn't have any publicly available identifying information.  There will still be some emails addresses in things like comments, but since these are already publicly available I don't think we need to worry about them.

### DIFF
--- a/db/clean.sql
+++ b/db/clean.sql
@@ -4,12 +4,13 @@ update donations set email = 'webmaster@mushroomobserver.org';
 update donations set who = 'anonymous' where anonymous = true;
 
 update herbaria set email = 'webmaster@mushroomobserver.org';
+update collection_numbers set name = 'xxx';
 
 update image_votes set user_id = 0 where anonymous = true;
 
 update images set original_name = 'xxx';
 
-update observations set lat = null, long = null where gps_hidden = true;
+update observations set `lat` = null, `long` = null where `gps_hidden` = true;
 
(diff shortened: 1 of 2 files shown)


db/clean.sql
```diff
@@ -4,12 +4,13 @@ update donations set email = 'webmaster@mushroomobserver.org';
 update donations set who = 'anonymous' where anonymous = true;
 
 update herbaria set email = 'webmaster@mushroomobserver.org';
+update collection_numbers set name = 'xxx';
```
JoeCohen, line 7: Is this necessary? Isn't publicly visible? See partial screenshot from ttps://mushroomobserver.org/collection_number/show_collection_number/7939
Screen Shot 2020-11-01 at 15 43 09

JoeCohen:
Thanks for doing this and everything else associated with the snapshot.

db/clean.sql
pellaea: I agree with Joe.  Is there a use-case you have in mind, Nathan?

mo-nathan: I spotted some email addresses there, but I guess it's not critical that it be emptied.

pellaea:
Looks good to me.  Thanks for taking care of this, Nathan.

PR #703, opened 2020-11-10
Fixes bug in create/edit_naming.

pellaea:
The bug is that if you create or edit a naming, and enter an existing genus that currently has a single version, but fill in a different author, it will change the author.  It should only fill in the author if the existing name has no author.  For example, propose the name "Psilocybe Azurescens".  Psilocybe already has an author, "(Fr.) Kumm.".  After the naming is created, it will change the author to "Azurescens".  (Even if the name is locked!)

So far, all I've done is write a test that reproduces the bug.

### DIFF
--- a/app/models/name/create.rb
+++ b/app/models/name/create.rb
@@ -52,7 +52,7 @@ class Name < AbstractModel
   end
 
   def self.set_author(names, author, fill_in_authors)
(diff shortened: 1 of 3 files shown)


pellaea:
Okay, this is ready for review now.

PR #707, opened 2020-11-13
Changed association.count to association.size throughout API.

pellaea:
Very trivial.

### DIFF
--- a/app/views/api2/_image.xml.builder
+++ b/app/views/api2/_image.xml.builder
@@ -25,13 +25,13 @@ xml.tag!(
     xml_minimal_object(xml, :owner, :user, object.user_id)
   else
     xml_detailed_object(xml, :owner, object.user)
-    xml.files(number: Image.all_sizes.size + 1) do
+    xml.files(number: Image.all_sizes.length + 1) do
       (Image.all_sizes + [:original]).each do |size|
         xml_image_file(xml, object, size)
       end
     end
     if object.observations.any?
-      xml.observations(number: object.observations.size) do
+      xml.observations(number: object.observations.length) do
         object.observations.each do |observation|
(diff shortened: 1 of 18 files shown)


app/views/api2/_image.xml.builder
```diff
@@ -25,13 +25,13 @@ xml.tag!(
     xml_minimal_object(xml, :owner, :user, object.user_id)
   else
     xml_detailed_object(xml, :owner, object.user)
-    xml.files(number: Image.all_sizes.count + 1) do
+    xml.files(number: Image.all_sizes.size + 1) do
```
JoeCohen, line 28: Is `length` more performant here? 
>If the collection has been already loaded `size` and `length` are equivalent. If not and you are going to need the records anyway `length` will take one less query. Otherwise `size` is more efficient.

https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-size

app/views/api2/_name.xml.builder
```diff
@@ -24,15 +24,15 @@ xml.tag!(
...
+      xml.synonyms(number: object.synonyms.size - 1) do
         (object.synonyms - [object]).each do |synonym|
           xml_detailed_object(xml, :synonym, synonym)
         end
       end
     end
     if object.classification.present?
       parse = Name.parse_classification(object.classification)
-      xml.parents(number: parse.count) do
+      xml.parents(number: parse.size) do
```
JoeCohen, line 35: See above re `length`

pellaea:
Oh geez, really?  I'm sorry, I thought size did that.  Who named these
silly things, anyway???
Okay, I've changed them all from .size to .length.
Gah, and now I've got to test 18 API calls *again* to make absolutely sure
there are no N+1 queries...

PR #709, opened 2020-11-15
Update site stats

JoeCohen:
Update site statistics page:
- Add Contributing Users to Site Stats
- Display Contribution Users instead of Users
- Add Sequences and Sequenced Observation to stats, and display them.

See [Pivotal #175685466](https://www.pivotaltracker.com/story/show/175685466) for rationale.

### Manual Test
- Go too observer/show_site_stats
Result:
Screen Shot 2020-11-15 at 08 09 06

Delivers https://www.pivotaltracker.com/story/show/175685466.


app/classes/site_data.rb
```diff
@@ -35,8 +35,7 @@
 #      },
 #    }
 #
-################################################################################
-
+#
 class SiteData
```
JoeCohen, line 39: Half-baked idea: I don't have time to do it now, but I wonder if this old class can be refactored by replacing the sql queries with chained ActiveRecord scopes.

app/views/observer/show_user.html.erb
```diff
@@ -61,7 +61,11 @@
 
   # These are roughly in decreasing order of importance.
   rows = []
-  for field in SiteData::ALL_FIELDS - [:users]
+  (SiteData::ALL_FIELDS -
+   SiteData::SITE_WIDE_FIELDS -
+   # omit sequence stuff because it has no weight
+   [:sequences, :sequenced_observations]).
```
JoeCohen, line 67: I also think that showing these in the User Summary is more likely to trigger an argument about the Contributions algorithm. I.e., that it should include sequences. 
(I would favor including sequences, but coding it is not a priority.)

app/classes/site_data.rb
```diff
@@ -164,29 +178,30 @@ def self.update_contribution(mode, obj, user_id = nil, num = 1)
       user_id ||= User.current_id
     end
     weight = FIELD_WEIGHTS[field]
-    if weight&.positive? && user_id&.positive?
-      update_weight(calc_impact(weight * num, mode, obj, field), user_id)
-    end
+    return unless weight&.positive? && user_id&.positive?
```
JoeCohen, line 181: Lots of easy RuboCopping here & in other files

test/models/site_data_test.rb
```diff
@@ -3,11 +3,42 @@
 require("test_helper")
 
 class SiteDataTest < UnitTestCase
-  def test_create
-    obj = SiteData.new
-    obj.get_site_data
-    obj.get_user_data(rolf.id)
-    obj.get_all_user_data
+  def test_site_data
```
JoeCohen, line 6: Any suggestions for more robust tests?

pellaea:
Looks good to me, although all of the rubocopping does admittedly make it a bit challenging to find substantive changes in the haystack!  This is some extremely old code, and yes, it could undoubtedly be refactored dramatically.  I just don't think it makes the priority cut-off for any of us. :(  These pages are so rarely used.

JoeCohen:
> Looks good to me, although all of the rubocopping does admittedly make it a bit challenging to find substantive changes in the haystack! This is some extremely old code, and yes, it could undoubtedly be refactored dramatically. I just don't think it makes the priority cut-off for any of us. :( These pages are so rarely used.

I appreciate your looking this over and digging through the haystack to find the needles. 
It certainly is rarely used. I did it in response to an email thread triggered by an inquiry from Anne Pringle. I wanted to deal with it while it was still fresh in my mind.

pellaea:
Yes exactly, we are probably the most frequent users of that page! :)

PR #710, opened 2020-11-16
Non-numeric ICN ID

JoeCohen:
Fixes a bug which set ICN ID to zero when user entered non-numeric ID (e.g., "MB28460"). [1]

### Manual Test Script
- Edit any fungal name, entering a non-numeric ID. 
Expected Result:
- Form is reloaded
- Flash Error displayed:
>Unable to save changes.
Name ICN Identifier is not a number. 

Delivers https://www.pivotaltracker.com/story/show/175730994

---

[1] The bug is actually more complex.
The first time any user saves a non-numeric `icn_id`, MO will:
- flash success
- save the ICN id as 0 in the db
- show_name page will say "ICN Identifier: _missing_"

If someone saves a non-numeric `icn_id` for a second Name:
- MO tries to save `icn_id` as 0, but fails 
- flashes error "An ICN identifier number may be used by only one Name. ‘0’ is already used by ‘\<first Name>’."


config/locales/en.txt
```diff
@@ -1507,6 +1507,7 @@
   form_names_refs_help: Each line is considered a separate reference, but no other formatting is enforced.
   name_error_unregistrable: "The [rank] '[name]' is not eligible for an ICN identifier."
   name_error_icn_id_in_use: "An ICN identifier number may be used by only one Name. '[number]' is already used by '[name]'."
+  name_error_icn_id_non_numeric: "ICN ID must be numeric"
```
pellaea, line 1510: Minor I know, but I feel like "must be numeric" is programmer-talk.  Maybe better to spell it out in English: "ICN ID must be a number".

pellaea:
Looks good to me!

config/locales/en.txt
JoeCohen: Thanks! 
I think I ended up not using this string. Will double check, and either delete it or turn it to English.

PR #713, opened 2020-11-20
Fix "Locate on Map" button

pellaea:
Needs to open the map first if not already open.


pellaea:
I will push this immediately because I have no way to test in thoroughly in dev since gmaps doesn't work in dev.

PR #715, opened 2020-11-22
Change CSV separator and column ordering for Darwin Core CSV export

mo-nathan:
This allows the CSV to be included in a complete Darwin Core Archive including the attached meta.xml file.  Note: GitHub required that it end with .txt.
[meta.xml.txt](https://github.com/MushroomObserver/mushroom-observer/files/5580147/meta.xml.txt)

To test:
- Download a Darwin Core CSV (simplest way I've found is to go to a species list, then click 'Downloads, Reports, Print Labels'.  Select 'Darwin Core'.  Click 'Download'.
- Rename the downloaded file 'observations.txt'
- Download the attached meta.xml.txt and rename to just 'meta.xml'.
- Create a zip file with `zip dwca.zip meta.xml observations.txt`.
- This should create a file called `dwca.zip`.
- Go to the GBIF validator: https://www.gbif.org/tools/data-validator
- Click 'SELECT FILE' and verify that the dwca.zip file is considered valid.

### DIFF
--- a/app/classes/observation_report/base.rb
+++ b/app/classes/observation_report/base.rb
@@ -3,13 +3,13 @@
 module ObservationReport
   # base class
   class Base
-    attr_accessor :query
-    attr_accessor :encoding
+    attr_accessor :encoding, :query
 
     class_attribute :default_encoding
     class_attribute :mime_type
     class_attribute :extension
     class_attribute :header
+    class_attribute :separator
 
     def initialize(args)
       self.query    = args[:query]
--- a/app/classes/observation_report/csv.rb
+++ b/app/classes/observation_report/csv.rb
@@ -9,9 +9,10 @@ module ObservationReport
     self.mime_type = "text/csv"
     self.extension = "csv"
     self.header = { header: :present }
+    self.separator = ","
 
     def render
-      ::CSV.generate do |csv|
+      ::CSV.generate(col_sep: separator) do |csv|
         csv << labels
         formatted_rows.each { |row| csv << row }
       end.force_encoding("UTF-8")
--- a/app/classes/observation_report/darwin.rb
+++ b/app/classes/observation_report/darwin.rb
@@ -3,12 +3,16 @@
 module ObservationReport
   # Darwin format.
   class Darwin < ObservationReport::CSV
+    self.separator = "\t"
+
     def labels
       %w[
+        CatalogNumber
+        OccurrenceID
+        BasisOfRecord
         DateLastModified
         InstitutionCode
         CollectionCode
-        CatalogNumber
         ScientificName
         ScientificNameAuthor
         ScientificNameRank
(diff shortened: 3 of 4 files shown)


pellaea:
This looks great to me.  Nice and simple and sweet.  Let me just check to see if there are any other columns potentially with wacky stuff in them...

Yeah, that really *should* be it, although the paranoid voice (way) in the back of my head warns that users could do wacky things even with location names and user names.  How bad is the mode of failure?  The "ultimate" solution would be to add a post-processing hook that we could use to strip those things out of each line right before it is sent, but after they are formatted.  Or alternatively, a hook that applies to every single value sent in to the formatter (although I would also start to worry about performance costs of that).  Maybe just additional clean_value() calls on user_name_or_login and locality?  Or forget about it and wait to see if it ever fails and deal with it then?  Or... yet another option, add validation on the model side for User#login, User#legal_name, Location#name, and just do a quick one-time check of whatever's currently in the DB?  Somehow validation never seems to be foolproof, though.  (For example, I recently saw a name_versions row with null updated_at timestamp.  WTF?  Should be impossible!  Actually I was too lazy to fix it, should still be there.)

JoeCohen:
1. I ran the manual test on an export of my Observations.
The file was validated with 
> Record Structure: Column mismatch
GBIF Occurrence Interpretation: Taxon match none, Basis of record invalid.
2. Question: Would it be better to create a new observation report instead of modifying the existing one. My concern is whether changing the structure or labels in an existing report will break something for an existing consumer of the existing report. (I think what you're doing is better because it accurately reflects the report name.
3. Is `Notes` (the final label) a DarwinCore field? If not, how about `occurenceRemarks`? https://dwc.tdwg.org/terms/#dwc:occurrenceRemarks

JoeCohen:
Scratch the above. I was in the wrong branch.

pellaea:
I really don't think anyone has ever used this report format.  And Nathan's
improved version is definitely more consistent with the standard.  The
standard, incidentally, at least as I understand it, is insensitive to
column order.  GBIF's implementation is a bit cheap-and-dirty if it cares.
I'm just saying, if changing column order is breaking someone's code,
that's their problem!! :)  And lastly, we tend to keep things backwards
compatible to a fault, saddling us with unnecessary past baggage.  I'd
really rather not have two Darwin formats, and all the confusion that will
necessarily result in.  Although...

That said, I could see a strong case being made for calling this format
Nathan is working on the "GBIF" format instead.  Given that GBIF is
unhealthily inflexible about the format, this makes some sense.  What if
some years from now someone decides to add additional stuff to the Darwin
export that should be totally acceptable according to the Darwin Core
specs, but which nevertheless breaks GBIF?  Totally understandable
"mistake".  Not really even a mistake, honestly...

Either way, I'm fine: change the existing format, or make it a new format
but call it "GBIF" not "Darwin Core".

JoeCohen:
Validates file exported from correct branch.
FYI:
Screen Shot 2020-11-22 at 13 02 02

JoeCohen:
Other suggestions (maybe to be implemented in additional PRs) here: https://www.pivotaltracker.com/story/show/175834487

PR #716, opened 2020-11-27
Create Zip file for DwCA export option

mo-nathan:
To test:
- Go to somewhere you can do an observation list export (e.g., go to a species list page then click 'Downloads, Reports, Print Labels')
- Select 'Darwin Core' and click Download

Expect result: A zip archive that can be successfully loaded into the GBIF data validator (https://www.gbif.org/tools/data-validator) without any modification.

Previous result: A CSV file that had to be combined with a meta.xml file and then bundled into a zip archive.

Changes: Added rubyzip gem.  Refactored ObservationReport:Base to separate out the table export specific stuff (BaseTable) from file export stuff (Base).  BaseTable inherits from Base and CSV now inherits from BaseTable.  Created DarwinCsv which is essentially the same as the old Darwin class.  Introduced ZipReport class which inherits from Base for creating Zip Archives.  Changed the Darwin class to inherit from ZipReport and uses DarwinCsv to build a real DwCA file.  Added a static meta.xml file.  Added a test and fixed up the existing tests.


JoeCohen:
Very nice! It validates with Issues. zip file attached if it's any use.
[dwca.zip](https://github.com/MushroomObserver/mushroom-observer/files/5609637/dwca.zip)


As far as I can see, all the Issues make sense: 
- Validator detects missing minus side for longitude ("Presumed negated longitude"). Very cool. (And makes me wonder about the possibility of using a DWC export and the validator as a tool for detecting issues in MO Observations/Names.)
- Validator is matching at a higher rank ("Taxon match higherrank") when the scientific name isn't (yet) in the GBIF db. It's doing this for prov names; Groups; newer scientific names.
- **Validator says "Taxon match none" when dwc:specificEpithet is populated with an infrageneric rank.** (That's an issue that we've already flagged, and is probably in another ticket.)

Question: I can see only a limited number of records for each Issue. Is there a way to see all of them?)

JoeCohen:
I looked at the code and it's almost all beyond me without spending a lot more time.
My only question is whether it's possible and would make sense to have a consolidated dictionary that includes:
index, term, label, format_row

pellaea:
Sorry, might not be able to get to this tonight.  Please don't hold
anything up for me.

mo-nathan:
Thanks for the thoughts on this.  I'm going to merge this and my other PR in the morning unless anyone raises a concern between now and then.  Tomorrow I'm hoping to go after the author and taxon files.  Maybe even the media file.  If I'm feeling even more ambitious I might even go after Joe's suggested changes.

I'm not sure what Joe is suggesting with "a consolidated dictionary that includes: index, term, label, format_row".

Once all that is squared away, the next challenge will be creating a dynamic list of observations for export which may require a background job as previously suggested.  I'm debating explicitly marking observations as "appropriate for export" to avoid doing too much thinking about  which observations we want to export.  Then we could go through all the historical observations over time and fine tune the export in various ways.

JoeCohen:
I meant a constant that defines (i.e. a lookup table), by column:

   - Dawc column index
   - Dawc term
   - csv column label
   - value of the column [now defined in format_row]

(I thought that by collecting all that info in one place, it would be
easier to keep things in sync and add new columns.)

pellaea:
Cleverly done.  Looks good.  I did not test it myself.

PR #717, opened 2020-12-01
Add multimedia.csv to Darwin Core export

mo-nathan:
To test:
- Generate a Darwin Core export (e.g., go to a species list and click the download link and select Darwin Core)
- Unzip the results
- Checkout the new multimedia.csv file.
- Validate the entire zip file at GBIF
- Profit!


app/classes/observation_report/darwin/observations.rb
```diff
@@ -0,0 +1,96 @@
...
+          Country
+          StateProvince
+          County
+          Locality
+          Latitude
+          Longitude
+          MinimumElevation
+          MaximumElevation
+          Notes
+        ]
```
JoeCohen, line 45:
```suggestion
        ].freeze

File: app/classes/observation_report/darwin/taxa.rb [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+      def formatted_rows
+        sort_after(observations.taxa)
+      end
+
+      def labels
+        %w[
+          taxonID
+          scientificName
+        ]
```
JoeCohen, line 24:
```suggestion
        ].freeze
```

test/models/observation_report_test.rb
```diff
@@ -114,12 +114,31 @@ def test_darwin_csv
...
+    expect = ["meta.xml", "observations.csv", "multimedia.csv"]
+    do_zip_test(ObservationReport::Dwca, expect)
+  end
+
+  def test_taxa_report
+    taxa_report = build_taxa_report
+    report_content = taxa_report.body
+    assert_not_empty(report_content)
+    table = CSV.parse(report_content, col_sep: taxa_report.separator)
+    assert_equal(Observation.all.pluck(:name_id).uniq.length + 1, table.count)
```
JoeCohen, line 130:
```suggestion
    assert_equal(Observation.select(:name_id).distinct.count + 1, table.count)

**Comment by JoeCohen** [comment]
The multimedia file is cool. I'm waiting for the $ to roll in.
When I feed the zip to the GBIF validator, I get:
>The file can be indexed by GBIF

with issues in `meta.xml`:
>Validation Issues
Resource Structure
Unknown term
Simple Multimedia | http://rs.gbif.org/terms/1.0/gbifID
Simple Multimedia | http://rs.tdwg.org/ac/terms/accessURI

No issues in `multimedia.csv`
Same old stuff (expected) in `observations.csv`

**Comment by mo-nathan** [comment]
I saw the same issues as Joe with the multimedia.csv.  These terms came directly from the SvampeAtlas and EOL DwCAs.  I was hoping to be able to generate a single export for both so the multimedia.csv and meta.xml would need to be the same.  However, it is now clear that we're going to need separate meta.xml files.  So I've revised things for this PR to more closely follow the SvampeAtlas example.

File: app/classes/observation_report/darwin/images.rb [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      def sort_after(rows)
+        rows.sort_by { |row| row[0].to_i }
+      end
+
+      private
+
+      def image_url(id)
+        "https://images.mushroomobserver.org/1280/#{id}.jpg"
```
pellaea, line 88: The "correct" URL is https://mushroomobserver.org/images/$size/$id.jpg.  For 1280 that is currently identical, but we cannot guarantee it long-term.  Also, this corrected URL will automatically server the image locally if it has not been transferred to the image server yet.  Yes, it generally only takes a few seconds to transfer images, but on rare occasions owing to glitches, etc., it can take up to a day before the nightly script/verify_images job catches any missed transfers.

app/classes/observation_report/darwin/images.rb
```diff
@@ -0,0 +1,103 @@
...
+          images: Image.arel_table,
+          images_observations: Arel::Table.new(:images_observations),
+          observations: Observation.arel_table,
+          users: User.arel_table,
+          licenses: License.arel_table
+        }
+        self.query = tables[:images]
+        join_table(:images_observations, :image_id, attribute(:images, :id))
+        io_attribute = attribute(:images_observations, :observation_id)
+        join_table(:observations, :id, io_attribute)
```
pellaea, line 33: I always worry whenever we join observations to images.  And I think in this case you don't even need to.  The only column you apparently need from observations is id, which is in images_observations.

app/classes/observation_report/darwin/images.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      def add_joins
+        self.tables = {
+          images: Image.arel_table,
+          images_observations: Arel::Table.new(:images_observations),
+          observations: Observation.arel_table,
+          users: User.arel_table,
+          licenses: License.arel_table
+        }
```
pellaea, line 29: These arel_tables are pretty slick!  I think I've finally gotten my head around what's going on here.

Question: Did you consider making self.tables a method instead?  It's a perfect opportunity to use that screwy-named technique... that's it: memoization! :)

def tables
  @tables ||= {
    images: Image.arel_table,
    etc.
  }
end

Really just serves to remove this assignment to self.tables from add_joins, but that seems worthwhile.  Initializing the arel_tables seems to be sort of foundational to this style of building a query, so burying that step inside "add_joins" seems a bit strange to me.  I'm just saying, I guess, that it would have saved me a great deal of head-scratching if the arel_tables were instead explicitly "initialized" or at least have them implicitly spring into existence (via a memoized accessor).  Then I would have known that they were not just something needed by the joins.  For example, I was mightily confused when you seemingly arbitrarily assigned tables[:images] to self.query!

Formal suggestion:

1. move the line self.query = tables(:images) to the top of initialize_query
2. move self.tables = { ... } into a memoized method "tables"

I think the flow and structure would be a little bit clearer.  (To me at least!)

app/classes/observation_report/darwin/images.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      def join_table(join_name, join_field, attribute)
+        table = tables[join_name]
+        join_attribute = table[join_field]
+        self.query = query.join(table).on(join_attribute.eq(attribute))
+      end
+
+      def add_project
+        query.project(attribute(:images, :id),
+                      attribute(:observations, :id).as("obs_id"),
```
pellaea, line 46: Can you change this to the following?

    attribute(:images_observations, :observation_id).as("obs_id")

pellaea:
Very cleverly done!

mo-nathan:
@pellaea I like all your suggestions.  Thanks!  I may not be able to implement them until the weekend, but we'll see if I can find some time before that.

app/classes/observation_report/darwin/images.rb
mo-nathan: I understand your worry.  I think it's avoidable in this case as you suggest, but I think the EOL export is going to require it since it will need to connect images to taxa and I'm pretty sure that means going through observations.  Assuming the underlying issue is total memory usage, it might be a time to use the `find_each` approach and limit the queries to 1000 rows or so and then iterate through them.  I expect these will ultimately be happening in a separate job/thread so I'm not terribly worried about how long it takes assuming it doesn't create some sort of table lock.

app/classes/observation_report/darwin/images.rb
mo-nathan: Yep.  Of course I started from observation ids which is why it didn't occur to me to let it go, but it seems better.  It would be interesting to see if MySQL 8 is smart enough that it doesn't really make any difference.

app/classes/observation_report/darwin/images.rb
mo-nathan: This is the first time I've used the Arel stuff.  I kind of knew about it in the back of my head and it definitely took some head scratching to figure it out, but I ended up liking it pretty well.  Happy to clean it up some now that it's working.

Ha, I thought about going the memoization route, but I thought you didn't like it and I couldn't remember how it works off the top of my head.  Happy to do it that way.

pellaea:
Haha, I love memoization, I hate the name "memoize"! :)  Or should I say, I
can't for the life of me figure out its origin, so it makes no sense to
me.  Is it somehow a play on "memorize"??

pellaea:
from wikipedia: "The term "memoization" was coined by Donald Michie
<https://en.wikipedia.org/wiki/Donald_Michie> in 1968[3]
<https://en.wikipedia.org/wiki/Memoization#cite_note-Michie1968-3> and is
derived from the Latin <https://en.wikipedia.org/wiki/Latin> word "
memorandum <https://en.wikipedia.org/wiki/Memorandum>" ("to be
remembered"), usually truncated as "memo" in American English, and thus
carries the meaning of "turning [the results of] a function into something
to be remembered". While "memoization" might be confused with "memorization
<https://en.wikipedia.org/wiki/Memorization>" (because they are
etymological cognates <https://en.wikipedia.org/wiki/Cognate>),
"memoization" has a specialized meaning in computing."  Not terribly
helpful, I guess because "memo" means something totally different to me
now... but wow, it goes back to K&R days!  I'll get over it eventually.  If
you let me gripe about it a few more times. :)

app/classes/observation_report/darwin/images.rb
mo-nathan: Will do.  I was just worried about putting that much pressure on mushroomobserver.org since I expect folks will grab the DwCA and then immediately start downloading everything.  I wonder if there would be a way to create a tarball somewhere that folks could download instead of hitting all though URLs.  I was also figuring that we'd be generating these things on a slow enough time scale that I wouldn't have to worry about a lag of even a day.  I think the current idea is to generate one of these something like once a quarter.  I'm also debating what the image size should be.  I think Alan Celestino used 320s.  That seems a bit small these days, but maybe that's preferred.  I'm hoping to get Jess and Alan's opinions on that.

pellaea:
I wouldn't worry about hitting the webserver.  Nginx rewrites it, so it
never hits rails.  Should be really low-impact.  Famous last words, right?
Zipping won't reduce the total download size, as jpegs are already
compressed.  Besides, I believe that all the preceding users only
downloaded a small fraction of the total images.  I'm not sure there's any
way to get around letting users download just the images they want.

PR #718, opened 2020-12-02
Location name suggestions

pellaea:
Improves observation form's handling of locations.

1. Officially changes policy concerning including counties back to what it was at the beginning: counties are now encouraged again.

2. When lat/long is updated (pin dropped or dragged, gps taken from image header, user enters it by hand), it uses google to reverse geolocate, and it sends the resulting info to the observer controller: country, state, county, city.  (It makes the submit buttons wait for google to respond.)

3. "Earth" is no longer considered acceptable.  It will complain "please enter at least country!" and force the user to approve "Earth" before accepting it.  If geolocation data is available, it will suggest existing locations which contain the lat/long, or which match google's reverse geolocation, or if nothing matches it will suggest a fairly reasonable name for a new location based on google's reverse geolocation.  This latter will require some editing in some cases, but it will at least be better than nothing.

4. If the lat/long is way outside of the bounding box for the chosen location, it will complain, and it will suggest some alternative existing locations which contain the lat/long.

5. If a smaller location contains the lat/long it will suggest this location.

6. If the location name entered does not match an existing location, it will go to some rather extreme efforts to correct a number of common mistakes, including but not limited to correcting spelling mistakes, filling in omitted country, state, county, unabbreviating states.  It will use info from google's reverse geolocation in some cases to fill in missing terms.  But importantly, this process will only ever suggest *existing* locations.  All the usual mistakes are still going to happen when creating new locations.

7. If any of the above cause MO to suggest one or more alternative location names, it will reload the form and display the suggestions as a set of radio boxes, along with at least some explanation of why we're suggesting the alternative locations.  The user can then click any of the radio boxes and re-submit.  Or the user can simply resubmit the form to "approve" whatever data they already have entered.

All tests are passing, but I feel like a good deal more testing is required before merging this.  Also, two of the primary location name correction methods (in app/models/location/suggestions.rb) violate ABC complexity and should probably be refactored a bit.  I will take a look at this later when I'm fresher to see if any clever ideas arise.

Meanwhile, this is a working draft.


app/controllers/observer_controller/create_and_edit_observation.rb
```diff
@@ -792,4 +863,15 @@ def whitelisted_observation_params
 
     params[:observation].permit(whitelisted_observation_args)
   end
+
+  def geolocation(params)
+    {
+      country:   params[:country],
+      state:     params[:state],
+      county:    params[:county],
```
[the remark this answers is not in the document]
pellaea: Errr... anyone care to explain to me how these aren't already aligned??

JoeCohen:
Wow! This sounds great!
Thanks so much.
But I won't have a chance for a detailed look until this evening
or tomorrow morning.

app/controllers/observer_controller/create_and_edit_observation.rb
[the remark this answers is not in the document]
JoeCohen: I suggest either:
- ignore this for now, and when you're ready to merge, tell Codeclimate to Accept (I think that's the right word) it;
- Tell CodeClimate `wontfix` for **each** allegedly misaligned line
- `rubocop:disable Layout/HashAlignment` (and later `enable`) around this Hash

The problem is that we use
```yml
Layout/HashAlignment:
  EnforcedColonStyle: key
```
which demands "left align keys, **one space before ... values**" which @mo-nathan likes, 
rather than
```yml
  EnforcedColonStyle: table
```
which demands "left align keys ... and values", which you like.
See generally https://docs.rubocop.org/rubocop/1.1/cops_layout.html#layouthashalignment
Unfortunately, there doesn't seem to be a way to get RuboCop to accept multiple styles.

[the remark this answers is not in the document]
pellaea: Oh!  Okay, no, that's fine.  The error message said only that the *keys* need to be aligned, so it didn't occur to me that the values should *not* be aligned.  I'm fine with that.  I literally just didn't understand what it wanted me to do. :)

JoeCohen:
This will be super in practice: better exports, less work for admins, more consistency.
But I'd like to understand if it would likely to cause out-of-pocket expenses for MO, Inc. I think we'll be okay, but I'd like to double check. 
**What's your best guess how many geocode requests/mo would be added by this PR?**
(Back of the envelope numbers: we could do 20,000 additional GeoCode requests/month without incurring out-of-pocket costs.)

Background:
- We get $450/mo free use of GoogleMaps ($200 free tier, plus annually, manually renewed $250 nonprofit credit).
- For the last 12 months we've averaged ~$210/mo. Of course, there are months where we go higher. In October 2020, we hit $293.45. So we have about $150 to spare without incurring out-of-pocket expenses. Both the average and max monthly costs are likely to increase. 

Geocoding costs 0.005 USD per each request (if one makes 0–100,000 requests/month). https://developers.google.com/maps/documentation/geocoding/usage-and-billing

At that rate we can do 30,000 requests in a month like Oct 2020. With a safety margin of 10K requests/month, we could do 20,000 requests/month.

JoeCohen:
1. **Suggested locations will give some screwy-looking answers because the boundaries of many current locations are screwed up.** For instance: with `Where` field empty, attached photo, and accepting lat/lon (38.8495/-123.6387) from photo, suggested places are

> Please enter at least the country! Here are some existing locations. Will one of these work? If you submit the form again without any changes we will accept your observation as-is.   
  >>Mono Creek Overflow, Sierra NF, Fresno Co, California, USA 
Santa Rosa Memorial Park Cemetery, Santa Rosa, California, USA
Odd Fellows Cemetery, Santa Rosa, California, USA
Chesterfield Post Office, Chesterfield, Missouri, USA
Robert J. Henke Nature Preserve, Lake Grove, New York, USA
Iversen Dr., south of Point Arenas, Mendocino Co., California, USA
Meeting of the Pines Natural Area, Michaux State Forest, Pennsylvania, USA

IMG_0701

Two ideas:
- Exclude from suggested locations anything in a different country, state, or county than that obtained by a reverse geocode of the photo coordinates. (Here, that would have left Iversen Drive, which is perfect.)
- Exclude anything with an area that's "too large". (It would be interesting to know how many Locations are too large.)

2. I haven't dug into the code, but I assume you are deliberately not suggesting Locations which have parts of the Location name missing, e.g., in the above case: "California, USA" and "Mendocino Co., California, USA". What do you think about including counties in the suggestions?

JoeCohen:
> This will be super in practice: better exports, less work for admins, more consistency.
> But I'd like to understand if it would likely to cause out-of-pocket expenses for MO, Inc. I think we'll be okay, but I'd like to double check.
> **What's your best guess how many geocode requests/mo would be added by this PR?**
> (Back of the envelope numbers: we could do 20,000 additional GeoCode requests/month without incurring out-of-pocket costs.)
> 
> Background:
> 
> * We get $450/mo free use of GoogleMaps ($200 free tier, plus annually, manually renewed $250 nonprofit credit).
> * For the last 12 months we've averaged ~$210/mo. Of course, there are months where we go higher. In October 2020, we hit $293.45. So we have about $150 to spare without incurring out-of-pocket expenses. Both the average and max monthly costs are likely to increase.
> 
> Geocoding costs 0.005 USD per each request (if one makes 0–100,000 requests/month). https://developers.google.com/maps/documentation/geocoding/usage-and-billing
> 
> At that rate we can do 30,000 requests in a month like Oct 2020. With a safety margin of 10K requests/month, we could do 20,000 requests/month.

PS: Some numbers to think about. 
Observations created 2020 YTD: 42,695 (including deleted Observations). So if geocode were called for each one, we'd pay ~$145 out of pocket.
Locations created or updated 2020 YTD: 2,809. Probably a better count would be # of LocationVersions updated_at like 2020%.

pellaea:
Excellent questions, Joe.

re: numbers -- Is there a typo?  You said we get 20,000 requests per month,
and we created 40,000 observation this year, so that would be 3333
observations per month, allowing up to 6 geocoding requests per
observation.  There are ways to mitigate it, as well.  For example, do the
reverse geocode request only when the form is submitted, reducing it to 1
per observation, and only for obs with lat/long.  Maybe further restricting
it to obs that don't already have a valid location containing the lat/long.

re: bogus location suggestions -- Yikes.  That is a mess.  I like your idea
of leveraging reverse geocode info better where it is available.  (And it
should always be available in cases like the one you mention.)

re: not including California, Earth -- Yes, it is only suggesting the most
"accurate" locations which contain the given lat/long.  Accuracy measured
very roughly by the number of "terms" in the location name.  Therefore if
"Burbank, Los Angeles Co., California, USA" contains the point, then only
locations with 4 terms will be suggested.  If there is a 5-term location
containing that point, then Burbank won't even be suggested, either.  It
also only suggests *existing* locations.  And since most counties are not
actually defined locations (towns within a county may exist, but the county
itself doesn't have an existing location all to itself), that means most
counties won't even be considered.  It started to get prohibitively
confusing when I started trying to simultaneously *autocorrect* location
names and *suggest* existing location names.  Since the primary goal was to
save poor Igor (and me!) the trouble of repeatedly cleaning up bad versions
of existing locations, I focused on that aspect.  (The one exception is if
it can't suggest any existing location and no location is written in at
all.  Then it will suggest a non-existing location name for you based on
google reverse geocode info.)

Thanks for taking it for a test drive!  I hadn't actually taken the time to
run it live (on the dev VM) yet myself.  I didn't realize how far short of
my optimistic expectations it fell! :)

Both this PR and my other are likely to fester a bit, because I promised to
devote December to catching up on "professional" lichen work.  But I will
try to merge master into both periodically to make sure they don't go
stale, at least, because I am still very interested in both.

JoeCohen:
* The numbers look OK to me, so it looks like the PR won't cause financial problems.
* Those are great ideas if we run into problems in the future. Another thing to keep in mind is to stop using the GoogleMaps Elevation API, which accounts for about 20% of our costs.
> Excellent questions, Joe. re: numbers -- Is there a typo? You said we get 20,000 requests per month, and we created 40,000 observation this year, so that would be 3333 observations per month, allowing up to 6 geocoding requests per observation. There are ways to mitigate it, as well. For example, do the reverse geocode request only when the form is submitted, reducing it to 1 per observation, and only for obs with lat/long. Maybe further restricting it to obs that don't already have a valid location containing the lat/long.

pellaea:
Yes, good idea re: elevation api.  Hey, how about this?  Instead of doing
it on the form, do it on the server after the obs is created, and obviously
only if elevation is not already given.  Normally we don't do it that way
because we don't want to tie up a server thread waiting for google api to
return, but in that case we can just launch a background job to do it.

JoeCohen:
Re elevation API, form vs server. Seems complicated and easy to lose track of. I'd forget about it for now given that we currently seem to have Google API requests to spare.

JoeCohen:
### Codeclimate/Brakeman issues.
#### Potentially unsafe model attribute in link_to href NEW
>31 `link_to(link_string, location.show_link_args)`
Severity: MinorFound in app/helpers/object_link_helper.rb by brakeman

brakeman (latest version) does complain when run locally. This is a XSS issue. See https://brakemanscanner.org/docs/warning_types/link_to_href/. The local brakeman report has an expanded version of the possibly unsafe line:
>Potentially unsafe model attribute in link_to href near line 32: 
```ruby
link_to((where_string(Location.find(location).display_name, count) + " [#{:click_for_map.t}]"), Location.find(location).show_link_args)
```

#### Possible SQL injection NEW
>223 `WHERE id IN (#{locations.keys.sort.map(&:to_s).join(",")})`
Severity: MinorFound in app/controllers/observer_controller/indexes.rb by brakeman

Local report (with line expanded):
SQL Injection | Possible SQL injection near line 223: 
```ruby
Location.connection.select_rows(
 SELECT id, name, north, south, east, west FROM locations
 WHERE id IN (#{{ loc_id.to_i => nil }.keys.sort.map(&:to_s).join(",")})
)
```

config/brakeman.yml
```diff
@@ -0,0 +1,3 @@
+---
+:url_safe_methods:
+- :show_link_args
```
JoeCohen, line 3: Tells brakeman that `show_link_args` generates a safe url. See https://brakemanscanner.org/docs/warning_types/link_to_href/

This is one way to deal with the Codeclimate/brakeman issue [Potentially unsafe model attribute in link_to href](https://github.com/MushroomObserver/mushroom-observer/pull/718#issuecomment-743769794). But **it is a valid fix only if the brakeman issue is a false positive**.

Two other ideas for dealing with this issue:
1. Change ObjectLinkHelper#location_link line 31 as follows. (This does not assume that the issue was a false positive.)
From:
```ruby
      link_to(link_string, location.show_link_args)
```
To:
```ruby
      link_to(link_string, "#{location_show_location_path}/#{location.id}")
```

2. Just tell Codeclimate "Invalid" or "Wontfix"

JoeCohen:
Grr. I broke this. I will try to fix it tonite or more probably tomorrow.

JoeCohen:
This branch works through #ac591e983d03b45bcd948ad77567fd5918873010
It breaks with #a401a3aa172322a878b17e7a2fb151f118d5c99f

app/controllers/observer_controller/create_and_edit_observation.rb
```diff
@@ -42,6 +50,10 @@ def create_observation
     end
   end
 
+  ##############################################################################
+
+  private
```
JoeCohen, line 55: `private` sections added in order to clarify which methods are (not) part of the public interface

test/controllers/observer_controller_test.rb
```diff
@@ -1350,6 +1350,19 @@ def test_destroy_observation
     end
   end
 
+  def test_destroy_observation_failure
+    obs = observations(:minimal_unknown_obs)
+    params = { id: obs.id.to_s }
+    login(obs.user.login)
+
+    Observation.any_instance.stubs(:destroy).returns(false)
```
JoeCohen, line 1358: **Question: Is there a way to make `obs.destroy` return `false` without stubbing?**
For instance, is there something I can do to `obs` so that `destroy` blows up?
The object here is to cover these lines in `CreateAndEditObservation`:
```ruby
426     elsif !@observation.destroy
427       flash_error(:runtime_destroy_observation_failed.t(id: obs_id))
428       redirect_to(add_query_param({ action: "show_observation", id: obs_id },
429                                   this_state))
```

**Question: If stubbing is best, is there a better way to stub `obs.destroy`?**
I tried `obs.stub(:destroy, false) do ...` but it didn't work. Do I have the syntax wrong?

app/controllers/observer_controller/indexes.rb
```diff
@@ -218,12 +218,11 @@ def map_observations
 
     unless locations.empty?
       # Eager-load corresponding locations.
-      @locations = Location.connection.select_rows(%(
-        SELECT id, name, north, south, east, west FROM locations
-        WHERE id IN (#{locations.keys.sort.map(&:to_s).join(",")})
-      )).map do |id, *the_rest|
-        locations[id.to_i] = MinimalMapLocation.new(id, *the_rest)
-      end
+      @locations = Location.where(id: locations.keys.sort).
```
JoeCohen, line 221: I think the catalyst for changing this was a brakeman complaint about possible XSS.

test/controllers/observer_controller_test.rb
```diff
@@ -2613,6 +2741,25 @@ def test_create_observation_strip_images
...
+        specimen: "0",
+        thumb_image_id: "0"
+      },
+      name: {},
+      vote: { value: "3" }
+    }
+    login("rolf")
+
+    ObserverController.any_instance.stubs(:unshown_notifications?).returns(true)
+    post(:create_observation, params: params)
```
JoeCohen, line 2758: **Same questions about stubbing**
(I couldn't figure out another way to make `unshown_notifications?` return `true`.)
The object is to cover line 372 of CreateAndEditObservations:
```ruby
365   def redirect_to_next_page
366     if @observation.location.nil?
...
371     elsif unshown_notifications?(@user, :naming)
372       redirect_to(action: "show_notifications", id: @observation.id)
```

test/controllers/observer_controller_test.rb
```diff
@@ -2880,6 +3027,76 @@ def test_edit_observation_strip_images
...
+        "when(2i)" => "05",
+        "when(3i)" => "11",
+        specimen: obs.specimen,
+        thumb_image_id: obs.thumb_image_id
+      },
+      log_change: { checked: "0" }
+    }
+
+    login(obs.user.name)
+    Observation.any_instance.stubs(:save).returns(false)
```
JoeCohen, line 3047: **Same questions about stubbing**
The point is to cover line 452:
```ruby
446           if save_observation(@observation)
...
451           else
452             any_errors = true
```

test/controllers/observer_controller_test.rb
```diff
@@ -2880,6 +3027,76 @@ def test_edit_observation_strip_images
...
+      },
+      log_change: { checked: "0" }
+    }
+
+    login(obs.user.name)
+    Image.any_instance.stubs(:save).returns(false)
+    post(:edit_observation, params: params)
+
+    assert_redirected_to(/#{obs.id}$/)
+    assert_equal(img_notes, img.reload.notes)
```
JoeCohen, line 3097: There is also a problem with my test (or maybe with the code being tested).
While I am covering the line, the assertion isn't what I want: 
I'd like to assert that the flash has an Image error. But when I look at the flash here, there's no Image error. 
(And I think the assertion I wrote is actually testing whether `stubs`. So I probably should get rid of it.)

JoeCohen:
With #b97d0b5, CreateAndEditObservation is covered 100%.
So I will now poke at location suggestions (maybe starting tomorrow).

pellaea:
Maybe it isn't supposed to fail without somehow raising an error or
something like that.  So if you just stub it and have it return false, no
error message is actually produced?

JoeCohen:
Doh! That makes sense.

JoeCohen:
Summary of some ideas in random order. (You're probably already doing some of these):
- Demand that the location name include the county, If the geolocation is in the USA and a county is available;
(county is "available" if the user provides one or if Google returns one). 
(I'm just guessing at what a reverse geocode provides. I need to review Google's documentation to see if I can begin to understand this.) 
- Demand that new/edited Observations in the USA locations include the state (not just the country). (We shouldn't have more Observations located in placed like "(Pacific Northwest, USA)[https://mushroomobserver.org/location/show_location/3876]".
- Exclude from suggestions: 
  - anything in a different country, state, or county than that obtained by a reverse geocode of the photo coordinates;
  - anything in a state different from that provided by Google or the User;
  - anything in a county different from that provided by Google or the User;
  - locations that are "too large".

pellaea:
I think I try to require county if one is present.  Google's API is highly
nonspecific as to what the various administration levels represent or if
they will be present.  I made a stab at it, at least.

I think I require state for countries in which we have the states listed.

I do NOT try to filter out nonsensical suggestions.

If google provides country/state/county and user does not, I think I do try
to suggest them(?) but maybe only if an existing location matches.

If google provides AND user provides AND they are different... I have no
idea what I do!  I wouldn't trust the user.  I might not always trust
google either, though.  Personally, I do trust google in the US (I use
reverse geolocate all the time in my private database system).  But things
get funny in other countries, so I don't want to try to force the user to
use what google provides.

There are times when I filter locations which are too large.  For example,
if the user provides gps but no location, it will find all existing
locations which contain that gps, then suggest only those with the largest
number of "terms" (= number of commas essentially).  I was far from
thorough about this, though.  There may be additional cases where "Earth"
will be suggested, and that will obviously be a mistake(!)

Does the structure of the code make sense at all?  I rewrote it from
scratch several times because the whole thing made my brain hurt so much.
But I think the latest incarnation has a certain compelling logic to it.
But there are empirically some cases not handled correctly, esp. as
concerns incorrectly-defined existing locations, a case I never considered
at any point in the process.

A few cases I know I need to investigate more are:
1. bogus existing locations
2. google and user disagree at country, state or county level
3. existing locations are only very vague but user has provided plausible
more accurate location

(A common mistake that might be worth checking for explicitly is forgetting
the minus sign on either longitude or latitude, or accidentally swapping
latitude and longitude.  I just did this myself a few weeks ago! I don't
recall handling it in the existing PR.)

But I'm sure there are more cases to consider.

JoeCohen:
Thanks for the feedback. I haven't yet dug into the code. That's the next.
I'm still thinking about what to suggest, and once we have all the suggestions, how to order them.
Ex: If user doesn't provide GPS, _maybe_ suggestions should: 
- if possible include at least one suggestion at administrative level 2,
- be ordered by area, ascending, with the default being the smallest.
I'm not sure. I'm still thinking about all of this, especially verbal specificity vs size.

JoeCohen:
The user can create an Observation with Location "Earth". Did you intend that? (I think not.) Ex:
- Create Observation
  - Leave "Where" blank. 
  - Add a photo that has geolocation. [This, and some of the following, steps might be unnecessary to reproduce the problem.]
  - Set Observation Date to image date
  - Set Observation Latitude/Longitude to image lat/lon
  - Click Create
- Result: Create Observation form reloaded with Location problem form, and with `Where: Earth`
  - Click Create
- Result: Create Observation form again reloaded with Location problem form, and with `Where: Earth`
  - Again Click Create 
- **Result: Observation finally created with location: Earth**

JoeCohen:
Should we ever fill in "Earth" in the Observation form place_name field?
(That might fix the above issue.)

pellaea:
Actually I'm not sure I have a problem with that.  We sure make it
difficult to use "Earth", but at least it never gets set to "NULL" or utter
garbage.

JoeCohen:
Two cases where user doesn't fill in Where:
- We have no lat/lon from the user (via either exif or manually entered). We don't have a way to supply a reasonable location, so our choices are to force them to supply something or accept Earth. I'm ok with Earth, but would prefer to send them in a loop until they give us a location. (I know that risks have nonsense supplied.)
- We have lat/lon from user (either exif or manual). I think we can do better than Earth. Perhaps we fill in Where with the results of geocode or with the "best" LocationSuggestion? (I know that the geocode can be untrustworthy.) Or maybe we do this just for collections in the USA?

pellaea:
In the second case (user gives gps but no location) we do suggest a
rudimentary location even if it doesn't exist yet.  I think for the US it
gives park or city, county, state and country.  Try it out and see what you
think.  It's partly coded in javascript, and partly coded in ruby.  JS
interprets return from google in rough strokes, then passes it on to ruby,
who further refines based on whether we have a list of states for that
country or not.

There may be legitimate uses of "Earth" that we haven't foreseen, therefore
I would vote for accepting "Earth" in absence of any better alternative,
and after asking user to do better.  I was only expecting to ask the user
once, not twice.  But if it works that way currently, I'm not averse to
leaving it as-is.

JoeCohen:
Thanks! I won't tinker with Earth.
I'll dig into the code some more tomorrow.

app/models/location/suggestions.rb
```diff
@@ -0,0 +1,281 @@
...
+# Return array of suggested Locations for a given user input (name or lat/lon)
+class Location < AbstractModel
+  # SQL string to calculate db grid size in lat/lon squares
+  # This is the db counterpart of pseudoarea, differing by rounding error
+  GRID_SQUARES = "IF(east >= west, east - west, 360 + east - west) " \
+                 "* (north - south)"
+
+  # max suggested size in lat/lon "squares"
+  # suggestions should not include a Location with a larger pseudoarea
+  MAX_SUGGESTED_SIZE = 6800
```
JoeCohen, line 12: #### What's a reasonable number to use here?
I have no idea. 6800 is huge.
There are now 4 locations larger than this: North America, Russia, Earth, Latin America. (There used to be 57, but I fixed the rest yesterday.)

app/models/location/suggestions.rb
```diff
@@ -0,0 +1,281 @@
...
+                           "IF(east >= west, west <= ? AND east >= ?, " \
+                           "west <= ? OR east >= ?)",
+                           lat, lat, long, long, long, long).
+            # exclude locations that are too large
+            where("#{GRID_SQUARES} <= #{MAX_SUGGESTED_SIZE}").
+            to_a
+      # Return only most "precise" Locations.
+      [3, 2, 1, 0].each do |n|
+        locs = all.select { |loc| loc.name.split(",").length > n }
+        return locs.sort_by(&:pseudoarea) if locs.any?
```
JoeCohen, line 48: #### Is there a reasonable way to get the db to sort the results by size ascending?
I'd rather have that done by the db than by Rails. But I couldn't figure out the right syntax. Maybe it's so complex that we should just leave it to Rails. There won't be a noticeable performance difference if there aren't many "precise" hits.

I found it helpful to have the suggestions sorted by size. Compare unsorted (good suggestions highlighted):
Screen Shot 2020-12-27 at 16 01 06
to sorted:
Screen Shot 2020-12-27 at 16 00 11

JoeCohen:
This is so complex (probably necessarily so) that I can't keep it in my head.
I'm going to do some more manual testing, hoping that will help me.
I haven't yet looked at the geocoding you describe.

Thinking out loud, but these shouldn't delay the rollout. 
- Would it make sense to have `suggestions` be PORO?
- Would it make sense to provide `suggestions` with all the user input, letting `suggestions` do the geocoding and some of the other work now performed by CreateAndEditObservation?

pellaea:
Good grief!  Missouri??  Sigh.

I like the sorted results, too.

I personally would vote for doing the sorting in rails.  The whole instance
should be available already.  We're only talking a small number of
suggestions, should not be too much of a performance such to sort by any
heuristics you can think up, no matter how convoluted!

pellaea:
Agreed that it is too complex.  Did I mention I rewrote it from scratch
several times before even getting it this "simple"??

re: PORO = stand alone class or something like that?  If so, yes,
definitely.  I was struggling to pass all the necessary data around.  Would
be so much easier to have all that stuff just be owned by the instance of
some class.  Sorry, I really struggle with the activation energy to create
new classes for things.  Always feels like this huge extra overhead.  But
in this case, by the end, I definitely regretted not doing so.

re: division of labor between suggestions and controller -- By all means
move things around.  Just note that feedback occurs in multiple ways:

1. The vast majority of the time it's just a list of suggestions, and a
little text explaining why it's making these suggestions.
2. It can also auto-fill the "where" field if it was empty and we have
geocoding from google

Not that complicated, and now that I think of it, may help simplify things
if we separate out these two cases in the controller and call different
entry points into the suggestions public interface.

JoeCohen:
Some Geocoding questions:
- Can I do it directly from Rails. i.e., something like: 
```ruby
results = Geocoder.search([38.84952222222223, -123.63869444444445])
```
which would return something similar to:
```sh
 => [#<Geocoder::Result::Nominatim:0x0000563376a6b2d8 @data={"place_id"=>265241413, "licence"=>"Data © OpenStreetMap contributors, ODbL 1.0. https://osm.org/copyright", "osm_type"=>"way", "osm_id"=>10313835, "lat"=>"38.849467", "lon"=>"-123.638783", "display_name"=>"46599, Iversen Drive, Mendocino County, California, 95445, United States", "address"=>{"house_number"=>"46599", "road"=>"Iversen Drive", "county"=>"Mendocino County", "state"=>"California", "postcode"=>"95445", "country"=>"United States", "country_code"=>"us"}, "boundingbox"=>["38.849417", "38.849517", "-123.638833", "-123.638733"]}, @cache_hit=nil>] 
```
- If so, what (if any) configuration/setup so I need in order to have it work from my dev VM? 
-

JoeCohen:
I Closed this PR by accident (and immediately re-opened). Sorry.

pellaea:
The point of doing it in JS is to have the client do it, not the server.
Maybe it doesn't take long, though?  And it doesn't hit the DB, which is
our weak link these days.

JoeCohen:
Makes sense. 
Is there a way to force it to happen when `Where` is blank but lat/lon is completed (ideally via photo or manually, but if it only works for photo, that's a start)?
The point is to be able to use the geocode results to reject some of the Suggestions, maybe also fill in `Where`.
Ex: User leaves `Where` blank, provides photo with lat/lon [38.84952222222223, -123.63869444444445]. The geocode results should let us
Reject:
  - Mono Creek Overflow, Sierra NF, **Fresno Co**, California, USA
  - Meeting of the Pines Natural Area, Michaux State Forest, **Pennsylvania,** USA
  - Chesterfield Post Office, Chesterfield, **Missouri,** USA
Fill in `Where` with: 
 - Mendocino Co., California, USA

pellaea:
It is supposed to initiate a geocode request every time the lat/long is
changed.  It should additionally prevent submission until the result is
returned. It then stores the result in hidden form fields so that rails can
see the results.

pellaea:
Another idea, if you're worried about excessive google API requests, is to
actually wait until the user submits the form, then when they click "send",
it sends the google api request, waits for it, then submits.  Does that
work better for you?

JoeCohen:
That seems reasonable, but wasn't on my mind.
Rather, I'm not seeing geocode data when I expect to see it. (I probably don't understand things.)
Here's what I'm seeing:
- Start server
- Create Observation
- enter only Latitude: `38.84952`, Longitude: `-123.6387`
- click Create
Result: 
With a breakpoint in `CreateAndEditObservation` like this:
```ruby
   182:   def location_missing
   183:     @location_suggestion_reasons << :form_observations_location_missing.t
   184:     geo = geolocation(params)
   185: byebug
```
At this breakpoint I see:
```ruby
   182:   def location_missing
   183:     @location_suggestion_reasons << :form_observations_location_missing.t
   184:     geo = geolocation(params)
   185: byebug
=> 186:     if geo[:country].present?
   187:       @place_name = Location.geolocation_to_name(geo)
   188:       @location_suggestions = Location.suggestions(@place_name, geo)
   189:       @place_name = Location.user_name(@user, @place_name)
   190:     end
(byebug) geo
{:country=>"", :state=>"", :county=>"", :city=>"", :latitude=>0.388495e2, :longitude=>-0.1236387e3}
```

pellaea:
You might check javascript console / errors.  Do you have a working google
API key?  Maybe something else is going wrong in javascript on your local
server.  You can add console.log("blah") to javascript to track progress
and state inside javascript.  Syntax is, as you probably know, very C-like.

pellaea:
How about try out a test script on the server to see how long a google
geocoder API request takes.  If it's on order tenths of a second to a
second, then this isn't worth screwing with.  Just move it all onto the
server.

JoeCohen:
I thought I had a working Google API key.
Screen Shot 2020-12-28 at 14 38 13

But it apparently doesn't work, either for MO development or elsewhere. 
Screen Shot 2020-12-28 at 14 40 48

I must be doing something wrong on the Google side. I'll try to figure it out.

pellaea:
I had trouble with this, too.  I eventually copied over my personal key
that I use for my own lichen stuff, and that worked.  Just make sure not to
accidentally check in whatever you do! :)

JoeCohen:
I had already put my key in `config/gmaps_api_key.yml`, under `development` and `test`.
Is that the right place?

pellaea:
I seem to remember needing to put it in two places?  Or was that the DB
password?

JoeCohen:
I don't know.
I'm going to put it to the side for now. It's too bewildering for me. 😢

JoeCohen:
@pellaea: As I'm a js complete novice, could you point me at the place where a geocode request every time the lat/long is changed.

pellaea:
Sorry, Joe, I just haven't had the time to follow all of this as closely as
I would like.  You're right, it does not do a geocode request each time the
lat/long is changed currently.  And in fact, if the map isn't open, the
code hasn't even been *loaded* yet, so there would be no way for it to do
so.

Look for updateFields() in
app/assets/javascripts/observation_form_map.js.erb.  It looks like it is
only called whenever someone... clicks or drags the pin on the map (via
addGmapsListener).

When the user uploads an image that has a lat/long in the EXIF, it creates
a radio button of some sort, and when the user clicks on it, it will popup
it's very own (separate) map showing where that location is on the map (but
does nothing with this!).  There is *another* button somehow associated
with these radio buttons (it's very hard to picture what the layout looks
like when dynamically generated with jQuery code!) that, when pressed, will
set observation_lat, observation_long, observation_alt and then hide the
EXIF radio button controls.  But as there are no callbacks associated with
those fields, nothing further happens.  The map shouldn't even be updated
if open, if I'm reading this right.

If we do what I originally suggested, we could pull the geocode stuff out
of updateFields(), and attach it to the observation_lat and
observation_long fields instead.  Then no matter how those fields are
changed -- by hand, by clicking on dragging the pin, or by selecting one of
the EXIF lat/long controls -- it will trigger the geocode.

Alternatively, move the geocode request from updateFields() to the submit
button.  Have it wait until the user submits the form, then, if the user
has changed either the lat/long or where fields, have it postpone
submission and first make a geocode request, then submit the form when that
request returns or after a short timeout period, with a twirly thing
displayed in the meantime to let the user know it's thinking.  (That is
already mostly in place, because I already have it delay submission until
the geocode request is finished.  But the geocode request is initiated only
by clicking/dragging the pin on the map, apparently.)

Lastly, we can do away with the geocode on the server instead and avoid the
whole javascript problem.  But again, I would advise at least testing how
quickly google responds.  It would be poor form to have server instances
held up for seconds twiddling their thumbs while waiting for google!
Hmmm.  Although... If this is a significant problem, note that waiting
takes no server resources (just memory which I think is no longer
limiting), so we could potentially offset the average "waiting for google"
time by adding additional server instances.  That bears consideration...

Phew, that's a mouthful.  Sorry.  And sorry I didn't dive into this more
deeply earlier.  Good news is I just finally finished and sent off last
night a big paper revising part of genus Anaptychia, a paper I've been
working on (initially with Alan's help at CCL actually! :) for over 5
years.  Of course, there are two additional papers that are high
priority...  It never ends!

JoeCohen:
Thanks so much! That's a huge help for me. 
Now I have potential paths forward to help with this branch; before I felt I was at a dead end.
And congratulation on the _Anaptychia_ paper. I'll have to figure out what _Anaptychia_ are; it's another way for me to procrastinate :-)

pellaea:
Nastassja wants to entitle the paper "Musical Chairs: Two new species of
Anaptychia from the Great Basin Desert, U.S.A."  That's how messed up the
taxonomy is in NA right now!

JoeCohen:
Manually resolved (in Codeclimate) as Invalid.
These are both false positives due to Codeclimate running an older version of Brakeman. 
They are not issues in Brakemen 5.2.3

> Code Climate has analyzed commit [65e57ee](https://github.com/MushroomObserver/mushroom-observer/commit/65e57ee9c8d5c28bd27d2b45eda90d6c70a24a42) and detected **2 issues** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Security	2
> Note: there is **1 critical issue.**
> 
> View more on [Code Climate](https://codeclimate.com/github/MushroomObserver/mushroom-observer/pull/718).

nimmolo:
Is this PR worth reviving?  It sure seems like it.

JoeCohen:
Definitely. This would improve the quality of many Observations and avoid
some correction of Observations by admins and regular users.

pellaea:
Yes and no.  I learned a lot while doing it, but it was way
over-engineered.  I think if I were to pick this back up I would start over
from scratch and do something much simpler... i.e., not incurring so much
technical obligation.  That's my opinion.

Is the PR description clear?  The whole point is to help people not
misspell location names.  But the fuzzy match I did for species names
doesn't really work for locations, because leaving a single word out is
"minor" for location names but causes the Levenstein distance to be huge.
"Close" matches according to Levenstein distance are just off by a few
letters, not entire words.  And things like spelling of "county" shouldn't
count against the match.  And we know all the country names and even state
names for many countries.  And so on.  So I went a little crazy.  There
must be a better way to solve this general problem.

nimmolo:
I get it. One option people seem to use is to have Google Maps API validate the place name. I'm not crazy about tying our data to Google but i guess open to it for expedience if others feel like it's a good idea. On May 10, 2023, at 7:40 PM, Jason Hollinger ***@***.***> wrote:﻿
Yes and no.  I learned a lot while doing it, but it was way
over-engineered.  I think if I were to pick this back up I would start over
from scratch and do something much simpler... i.e., not incurring so much
technical obligation.  That's my opinion.

Is the PR description clear?  The whole point is to help people not
misspell location names.  But the fuzzy match I did for species names
doesn't really work for locations, because leaving a single word out is
"minor" for location names but causes the Levenstein distance to be huge.
"Close" matches according to Levenstein distance are just off by a few
letters, not entire words.  And things like spelling of "county" shouldn't
count against the match.  And we know all the country names and even state
names for many countries.  And so on.  So I went a little crazy.  There
must be a better way to solve this general problem.

PR #719, opened 2020-12-06
Rubocop 2020-12-05

JoeCohen:
- Fix some RuboCop Lint, Style, Performance and Rails offenses, mostly using autocorrect.
- Fix follow-on Codeclimate/Rubocop issues (many of which were new in `master,` but hidden from Codeclimate by `.rubocop_todo.yml`)


lib/tasks/jason.rake
```diff
@@ -300,19 +300,20 @@ namespace :jason do
       lines = [line]
 
       # Skip blank lines and comments.
-      if /^\s*(#|$)/.match?(line)
+      case line
```
JoeCohen, line 303: Did I correctly convert the `if-elsif` to `case-when`?

JoeCohen:
This is almost all straightforward, with much of it being auto-corrected (and then manually inspected).
I am most concerned with the changes to `lib/tasks/jason.rake`, in particular the change from `if-elsif` to `case-when`. See https://github.com/MushroomObserver/mushroom-observer/pull/719#discussion_r536946826 above.

pellaea:
Please delete jason.rake.  I'm so sorry you wasted time on it.  It's been
ages since I've used that, no idea what it even does. :(

JoeCohen:
Thanks! And not a waste of my time. I learn, or reinforce, something with
each change.
I'll get rid of it.

app/models/name/spelling.rb
```diff
@@ -37,8 +37,6 @@ def self.suggest_alternate_spellings(str)
     results
   end
 
-  private
```
mo-nathan, line 40: Why the methods before the explicit call to public, methods public?  I can certainly see moving the private section to the end of the file so you don't have an explicit `public`, but it's surprising to me to make methods that were private public without any new calls to the method.  I see that the tests call it which might be causing an issue, but my general opinion is that private methods should not be directly tested.

JoeCohen: TLDR: Good catch! I will make these methods, and other methods related to #a9374baced13cc952a1bb056bac32248d48fc3de (truly) private.

Longer story: Thanks for taking the time to look at this PR.
In #a9374baced13cc952a1bb056bac32248d48fc3de, I applied the auto-corrections of the [Lint/UselessAccessModifier Cop](https://docs.rubocop.org/rubocop/1.0/cops_lint.html#lintuselessaccessmodifier). For this line RuboCop said:
>`Line #40` – Warning: Lint/UselessAccessModifier: Useless `private` access modifier.
  `private`

RuboCop's point is that `private` is useless for methods that are explicitly defined on `self` (or any other singleton methods). 
>`Line #43` – Warning: Lint/IneffectiveAccessModifier: `private` (on line 40) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self block` instead.
```ruby
  def self.guess_word(prefix, word) # :nodoc:
```

See also, e.g., https://jakeyesbeck.com/2016/01/24/ruby-private-class-methods/?utm_source=rubyweekly&utm_medium=email; https://stackoverflow.com/questions/4952980/how-to-create-a-private-class-method. 

So -- in addition to applying the autocorrect -- I will make these methods truly private. (And I agree about not testing private methods.)

JoeCohen: OH! Now I see -- these methods are tested directly.
I'll think about what to do with them.
PS: I don't understand what you mean by "Why the methods before the explicit call to public, methods public?" I can't find an explicit call to `public` in this file.

JoeCohen: Please ignore my question about explicit `public`; I'm not functioning very well this a.m.
I'll make stuff really private, and move them to the end of the file.

JoeCohen: I made some methods private, but left `Name.guess_word` and `guess_error` public.
It would take me way too long to rewrite the tests<sup>[1]</sup>, and I don't think that the other changes in this PR should wait for the solution of this pre-exisiting problem.
What do you think?

-----
1. I.e., tests that don't call those methods but do check that Name correctly handles everything correctly.

mo-nathan: That's fine.  I was just surprised by the change.  Thanks for pursuing it.  I had forgotten about the whole `class << self` thing which apparently is needed to actually make a class methods private.

PR #722, opened 2020-12-07
163189614 versioned enums

JoeCohen:
Uses a patched version of `acts_as_versioned` which stores enum attributes as integers in Version tables

Delivers https://www.pivotaltracker.com/story/show/163189614

### Suggested Manual Test
`bundle install`
* Create Name - Rank: `Genus`, `Candy Cap` 
* Edit Name to be Rank: `Species`, Scientific Name: `Candy cap Mushroom` 
* Edit Name to be Rank: `Subspecies`, Scientific Name: `Candy cap subsp. maple` Author: `DeShazer`
* Click `Previous Version`
Expected Result: Displays "Version 2: **_Candy cap_** Mushroom" with `Rank: Species`
* Click `Version 1`
Expected Result: Displays "Version 1: **_Candy_** Cap" with `Rank: Genus`


JoeCohen:
This is simple and it works, though it's not the [solution from the 2020-12-07 technical meeting](https://www.pivotaltracker.com/story/show/163189614/comments/220291172).
It should at least prevent further data loss pending a longer-range solution.

Questions:
- Does MO's `Gemfile` specify the gem in the best way?
- In the patched gem itself (https://github.com/MushroomObserver/acts_as_versioned):
  - Although MO's tests pass, I couldn't run the gems' tests. So I could have broken a gem feature that we don't use. 
  - Although my changes work, I didn't know how to do things the official RubyGems way. 
    - Should the gem be renamed?
    - Should it be in a different branch?
    - It would be nice to update the gem version, date, etc., the official RubyGems way, but I'm confused about how to do that. `rake gemspec` updates the date, but how do I update the version to 0.6.4? What else should be changed?

Gemfile
```diff
@@ -68,10 +68,13 @@ gem("rtf")
 gem("xmlrpc")
 
 # Simple versioning
-gem("cure_acts_as_versioned")
+# Use our own fork, which stores enum attrs as integers in the db
+gem("cure_acts_as_versioned",
+    git: "https://github.com/MushroomObserver/acts_as_versioned/")
```
JoeCohen, line 73: Is there a better way to do this?

mo-nathan: This seems reasonable to me until/unless you can get your fixes committed to the original gem and registered with ruby gems.

mo-nathan:
Regarding updating the version, I'm guessing you tried modifying the acts_as_versioned.gemspec file.  If that didn't work, then I'm not sure.  That's clearly what they did in this commit: https://github.com/cure/acts_as_versioned/commit/0481a03d5fd2f8670901853f0e66a18aafe9582b

Given that this thing is 7 years old, I'd make a pull request against the repo, but I wouldn't expect it to get noticed.  Having a local copy is fine.

JoeCohen:
Thanks for both suggestions. I'll make a PR, but expect the same as you.
I didn't manually modify the gemspec file because the comments in the file
said don't do that because the versioning because `rake gemspec` would
handle the versioning. But `rake gemspec` simply restored the version to
0.6.0.
So I will manually change the version.

JoeCohen:
I made a PR. But there were already 6 PRs outstanding, the first of which
is from 2011.
So I won't hold my breath.

PR #723, opened 2020-12-13
Add python

mo-nathan:
Adding python 2 to the Travis build to see if it resolves the current build issue with libv8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo: true
 # script: bundle exec rake lang:update
 
 before_install:
-  - sudo apt-get install exiftool
+  - sudo apt-get install exiftool python
 
 install: ./script/travis_setup
 


JoeCohen:
Thanks for trying this. I didn't know how to do it, and I was afraid that I
would break something.
Thinking out loud:

   - Is this installing Python3 or Python2?
   - Does libv8 need Python2?
   - Is order important, i.e., should it be sudo apt-get install python
   exiftool  instead of sudo apt-get install exiftool python
   - Could we replace therubyracer? Maybe with miniracer?
   https://github.com/rubyjs/mini_racer. Are there issues installing
   miniracer on the VM?

JoeCohen:
I will try python2. 
(I'll create a new branch. It may take ~1 hr because I'm having some issue with my laptop.)

JoeCohen:
Add Pivotal ticket. https://www.pivotaltracker.com/story/show/176132861

JoeCohen:
Same/similar issue installing `python2` on Travis. See https://travis-ci.org/github/MushroomObserver/mushroom-observer/builds/749441904#L50659
Next I will try substituting `mini_racer` (note the underscore) for `therubyracer`.

mo-nathan:
The original PR did install python2 and got a lot further during the libv8 install process, but there was still a segfault.

mo-nathan:
mini_racer seems like a better approach.

PR #724, opened 2020-12-13
Use mini_racer instead of therubyracer

JoeCohen:
- Hopefully solves Travis `libv8` install failure. See https://www.pivotaltracker.com/story/show/176132861
- Uses a more modern gem. In particular note it's using liv8 which is 5 major versions ahead of that used by `therubyracer`.
- And is more lightweight than `therubyracer`.

I have manually tested create_observation, but don't know how else, or what else to do, to test js.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,9 +23,12 @@ gem("sassc-rails")
 # Use jquery as the JavaScript library
 gem("jquery-rails")
 
-# Use thebuyracer as JavaScript runtime for ExecJS
+# Use therubyracer as JavaScript runtime for ExecJS
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
-gem("therubyracer", platforms: :ruby)
+# gem("therubyracer", platforms: :ruby)
+
+# Use mini_racer as a substitute for therubyracer
+gem("mini_racer")
 
 # Use CoffeeScript for .js.coffee assets and views
 gem("coffee-rails")
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,7 +122,7 @@ GEM
       railties (>= 3.1)
     json (2.3.0)
     kgio (2.11.3)
-    libv8 (3.16.14.19)
+    libv8 (8.4.255.0)
     loofah (2.7.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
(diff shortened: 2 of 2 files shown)


JoeCohen:
**Question: How can I be confident that our `js` functions work properly with this.` 
`mini_racer`:
- fixes the Travis Error problem;
-  is up-to-date gem vs. `therubyracer`.
BUT
**How can we test whether it's compatible with our js functions?**

(I created an Obs, and the photo stuff worked; typeahead works; I don't know if all the Google Maps stuff works, as we can't use MO's GoogleMaps API key from the VM. I don't know what else to test manually.)

JoeCohen:
Can the pivotal stuff be tested manually from the VM in dev mode?

mo-nathan:
Thanks for digging into this.  This looks like a reasonable solution to me.  I don't believe we have any real JS tests, so all you can do is manual testing which it sounds like you've done.  If some of it needs to deployed to work, then we'll probably just need to do that and see what happens.  I'd be most nervous about the JS stuff that Ray did for the image viewing.  I'm not terribly worried about the Google Maps stuff.

JoeCohen:
Thanks! In create & edit Obs, the image stuff seems to work OK.
But I'll play around with it some more right now.
If I don't see any problems, I will merge it.

JoeCohen:
In create obs, I can drag or select one or multiple images; they are added, the right info appears for Set observation date, Set image date, Set observation Latitude/Longitude.

JoeCohen:
Here's something that's broken in this PR:
- Create Observation
- In Where type-ahead window, begin typing until some Locations appear. E.g. type `Jackson Demon`
- Highlight one of the names, e.g., `Jackson Demonstration State Forest ...`
- Click on the highlighted name. 
Result: On the production server, the complete highlighted Location name appears in Where,
But in this branch, clicking the highlighted Name has no effect (at least in Google Chrome 86.0.4240.198 (Official Build) (x86_64)). [There's a work-around: hit `tab` key instead of clicking on the line. But I don't think our users will know about that.]

JoeCohen:
Never mind!! It's now working Okay. 
So I wil merge.

JoeCohen:
As you've probably seen I merged mini_racer (and some older branches that
were green).
I will re-start Travis builds on the branches where Travis currently Errors
to make sure that things work.
Travis takes its sweet time, so it will be a while before I can confirm
this.

mo-nathan:
Thanks Joe!  Sounds like a better solution.

PR #725, opened 2020-12-17
Cover Synonymy

JoeCohen:
- Gets 100% coverage of `synonymy.rb` 
- Fixes bugs exposed by additional coverage
- Minor refactoring 
- RuboCopping

Delivers https://www.pivotaltracker.com/story/show/176197454

There is no manual test


app/models/name/synonymy.rb
```diff
@@ -47,25 +47,21 @@ def sort_synonyms
 
   # Same as +other_authors+, but returns ids.
   def other_author_ids
-    @other_author_ids ||= begin
-      if @other_authors
-        @other_authors.map(&:id)
-      else
-        Name.pluck(:id).where(text_name: text_name).map(&:to_i)
-      end
-    end
```
JoeCohen, line 56: Question: **Is the `begin ... end` block necessary?**
(In removing it, I might have missed something basic/subtle/deep.)

app/models/name/synonymy.rb
```diff
@@ -47,25 +47,21 @@ def sort_synonyms
...
   # Returns an Array of Name's, including itself, which differ only in author.
   def other_authors
-    @other_authors ||= begin
-      if @other_author_ids
-        # Slightly faster since id is primary index.
-        Name.where(id: @other_author_ids).to_a
-      else
-        Name.where(text_name: text_name).to_a
-      end
-    end
```
JoeCohen, line 68: Same question about `begin ... end`.

app/models/name/synonymy.rb
```diff
@@ -180,26 +176,19 @@ def observation_count
   # Returns either self or name,
   # whichever has more observations or was last used.
   def more_popular(name)
-    result = self
-    unless name.deprecated
-      if deprecated
-        result = name
-      elsif observation_count < name.observation_count
-        result = name
```
JoeCohen, line 188: **Question: Did I correctly understand the intent of this method?** (It was not previously tested.):
I think there was a bug:
It did not account for the possibility that `self` had more observations than `name`; instead it immediately fell back on which had the most recent `naming`.

test/models/name_test.rb
```diff
@@ -2230,9 +2233,210 @@ def test_changing_author_of_autonym
     assert_equal("Ach.", name.author)
   end
 
+  # --------------------------------------
+  #  Synonymy
+  # --------------------------------------
+
+  def test_synonym_ids
```
JoeCohen, line 2240: This method is unchanged; It's simply been moved.

test/models/name_test.rb
```diff
@@ -2399,6 +2603,14 @@ def test_format_autonym
     )
   end
 
+  def test_make_sure_names_are_bolded_correctly
```
JoeCohen, line 2606: This method is unchanged; It's simply been moved.

PR #726, opened 2020-12-20
This happened after rebuilding my linux box from scratch

mo-nathan:
Not clear on why exactly.  It is clearly related to the recent stuff with libv8.  The explicit reference to linux makes me a bit nervous.  Definitely worth testing at least on a Mac before merging.

To test:
- Switch to this branch
- Run `bundle install`
- Look over the output to make sure there aren't any errors
- Make sure the tests all pass

Please add a comment to this PR with the host OS you're using and whether you are currently using the Focal VM.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -128,6 +128,7 @@ GEM
     json (2.3.0)
     kgio (2.11.3)
     libv8 (8.4.255.0)
+    libv8 (8.4.255.0-x86_64-linux)
     loofah (2.7.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
@@ -271,6 +272,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-linux
 
 DEPENDENCIES
   RedCloth
@@ -313,4 +315,4 @@ DEPENDENCIES
   xmlrpc
 
 BUNDLED WITH
-   2.1.4
+   2.2.2


JoeCohen:
Nothing blew up when I ran your manual test script.
Host OS: macOS Catalina v 10.15.7 
```sh
vagrant@ubuntu-focal /vagrant/mushroom-observer (gemfile-lock-update)
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```

Questions/Comments:
* Is there something I should do to avoid overwriting this change to `Gemfile.lock` in the future?
* Does the Vagrant Box need a change for the same reason. (If I had to rebuild my VM, would I be in danger of overwriting `Gemfile.lock`?)

JoeCohen:
P.S.: Both when I bundled and ran the tests, I got this warning:
>Warning: the running version of Bundler (2.1.4) is older than the version that created the lockfile (2.2.2). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.2.2`.

Should I follow those directions?

> Nothing blew up when I ran your manual test script.
> Host OS: macOS Catalina v 10.15.7
> 
> ```shell
> vagrant@ubuntu-focal /vagrant/mushroom-observer (gemfile-lock-update)
> $ lsb_release -a
> No LSB modules are available.
> Distributor ID:	Ubuntu
> Description:	Ubuntu 20.04.1 LTS
> Release:	20.04
> Codename:	focal
> ```
> 
> Questions/Comments:
> 
> * Is there something I should do to avoid overwriting this change to `Gemfile.lock` in the future?
> * Does the Vagrant Box need a change for the same reason. (If I had to rebuild my VM, would I be in danger of overwriting `Gemfile.lock`?)

mo-nathan:
@JoeCohen I'd recommend upgrading bundler.  All I did was run the `mo-dev` script after I brought up the current VM using vagrant.  It's conceivable to me that the new version of bundler is the source of this issue and I don't have much else to point at for possible sources of the issue.  However, as long as running `bundle install` didn't change the Gemfile.lock for you I don't think it's urgent.  The only situation where I would expect it might revert would be if you were to run `bundle update` which we don't really do.  I've thought about advocating for doing it more often, but it creates a bit more chaos that I think we can effectively deal with.

PR #728, opened 2020-12-21
Cover Description model

JoeCohen:
- Improves coverage of `description.rb` to 100%
- Fixes most `description.rb` RuboCop offenses, including changing other files as needed

Comment: In some cases, the added coverage is minimal; it hits the line and includes a very basic assertion. So there still may be bugs in the model.

Delivers https://www.pivotaltracker.com/story/show/176223877


app/models/description.rb
```diff
@@ -152,7 +154,7 @@ def unique_format_name
 
   # Descriptive title without parent name, in plain text.
   def partial_text_name
-    build_name(:part).t.html_to_ascii
+    put_together_name(:part).t.html_to_ascii
```
JoeCohen, line 157: Fixes a bug exposed by new tests:
```sh
ArgumentError: When assigning attributes, you must pass a hash as an argument.
```

app/models/description.rb
```diff
@@ -170,38 +172,16 @@ def unique_partial_format_name
     string_with_id(partial_format_name)
   end
 
-  # Descriptive subtitle for this description (when it is not necessary to
```
JoeCohen, line 173: I moved this to the end to a `private` section

app/models/description.rb
```diff
@@ -296,42 +276,42 @@ def self.admins_join_table
     "#{table_name}_admins".to_sym
   end
 
-  # Name of the join table used to keep admin groups.
+  # Wrapper around class method of same name
```
JoeCohen, line 279: DRYer; also simplifies test because it's unnecessary to explicitly test the class method if the instance method is covered.

app/models/description.rb
```diff
@@ -296,42 +276,42 @@ def self.admins_join_table
...
 
-  # Name of the join table used to keep reader groups.
+  # Wrapper around class method of same name
   def readers_join_table
-    "#{self.class.table_name}_readers".to_sym
+    self.class.readers_join_table
   end
 
-  # List of all the admins for this description, as ids.
+  # List of all the admins for this description
```
JoeCohen, line 304: Fixes mistake in the comment

app/models/description.rb
```diff
@@ -396,18 +376,6 @@ def remove_reader(arg)
     chg_permission(readers, arg, :remove)
   end
 
-  # Change a given User's or UserGroup's privileges.
-  def chg_permission(groups, arg, mode)
```
JoeCohen, line 400: Moved to private method at the end of file

app/models/description.rb
```diff
@@ -566,10 +522,61 @@ def update_users_and_parent
...
+    if (mode == :add) &&
+       !groups.include?(arg)
+      groups.push(arg)
+    elsif (mode == :remove) &&
+          groups.include?(arg)
+      groups.delete(arg)
+    end
+  end
+
+  def user_made_a_change?(user)
```
JoeCohen, line 570: **Questions**: 
* Would `EXISTS` make the SQL more efficient? (This method just needs to return a truthy value; we don't need to load an `id`.) 
* If so, what's the syntax? (I couldn't figure it out.)

test/fixtures/name_descriptions.yml
```diff
@@ -65,6 +65,7 @@ suillus_desc:
   <<: *DEFAULTS
   name: suillus
   user: dick
+  source_type: <%= NameDescription.source_types[:public] %>
```
JoeCohen, line 68: Repeat the defaults to emphasize that tests depend on these values in these fixtures.

app/models/description.rb
```diff
@@ -566,10 +522,61 @@ def update_users_and_parent
...
+    self.class.connection.select_value(%(
+      SELECT id FROM #{versioned_table_name}
```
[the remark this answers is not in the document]
JoeCohen: #### Question: Is there a way to change this so that brakeman won't re-raise the complaint if this method is tweaked?
I could tell CodeClimate that this is a false positive.
brakeman is highlighting `versioned_table_name`, see attached jpg. That's not user input. 
Screen Shot 2020-12-21 at 07 52 59
But if the SQL gets changed, see, e.g., https://github.com/MushroomObserver/mushroom-observer/pull/728#discussion_r546760871, it would be nice if brakeman doesn't complain again. 
So can the SQL be changed to guarantee that brakeman won't complain in the future? 
What about a prepared statement (or is that not worth the effort)?

app/controllers/description_controller_helpers.rb
```diff
@@ -111,7 +111,7 @@ def do_merge_description(src, dest, delete_after)
     src_title = src.format_name
 
     # Doesn't have permission to edit destination.
-    if !in_admin_mode? && !dest.is_writer?(@user)
+    if !in_admin_mode? && !dest.writer?(@user)
```
JoeCohen, line 114: This and similar changes fix [Naming/PredicateName offenses](https://docs.rubocop.org/rubocop/1.5/cops_naming.html#namingmethodname).

app/models/description.rb
```diff
@@ -237,14 +217,14 @@ def all_notes=(notes)
   #   num_fields, total_length = name.note_status
   #
   def note_status
-    fieldCount = sizeCount = 0
+    field_count = size_count = 0
```
JoeCohen, line 220: See https://docs.rubocop.org/rubocop/1.2/cops_naming.html#namingvariablename

app/models/description.rb
```diff
@@ -425,14 +393,7 @@ def permitted?(table, user)
...
-    return @group_users[table] if @group_users[table]
-
-    ids = group_user_ids(table)
-    ids = ["-1"] if ids.empty?
-    id_list = ids.map(&:to_s).join(",")
-    @group_users[table] = User.find_by_sql(%(
-      SELECT * FROM users
-      WHERE id IN (#{id_list})))
+    @group_users[table] ||= User.where(id: group_user_ids(table)).to_a
   end
```
JoeCohen, line 397: This change could use a second set of eyes.

app/models/description.rb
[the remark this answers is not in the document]
JoeCohen: I will use the simplest solution: tell CodeClimate this is a false positive.

PR #729, opened 2020-12-22
Delete bad route

JoeCohen:
`update_whitelisted_observation_attributes` was part of a skeleton #01276fcce2893daf675191d6b9b326550e19f787
but never caught, so it found its way into the final routes


mo-nathan:
Looks good! Thanks!

PR #732, opened 2021-01-02
Fix Location pattern_search with id

JoeCohen:
- Redirect to Location when a Location `pattern_search` pattern is Location id.
- Makes Location pattern search work like other types of pattern_search

Delivers https://www.pivotaltracker.com/story/show/176286532

#### Manual test
- Do a Location-type pattern search for `14656` (or the id of any other Location)
- Expected result: Displays that Location


mo-nathan:
Looks good.  Thanks!

PR #734, opened 2021-01-05
Improved flash for expired Advanced Search

JoeCohen:
"Search expired or cannot be found. Please re-enter search criteria."
instead of 
"undefined method 'id' for nil:NilClass"

Delivers https://www.pivotaltracker.com/story/show/24489297

#### Suggest manual test
- Go to http://localhost:3000/observer/advanced_search?q=xxxxx
- Result: Advanced Search form with flash: "Search expired or cannot be found. Please re-enter search criteria."

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1100,6 +1100,17 @@ class ApplicationController < ActionController::Base
     result
   end
 
+  # Handle advanced_search actions with an invalid q param,
+  # so that they get just one flash msg if the query has expired.
+  # This method avoids a call to find_safe, which would add
+  # "undefined method `id' for nil:NilClass" if there's no QueryRecord for q
+  def handle_advanced_search_invalid_q_param?
+    return unless invalid_q_param?
+
+    flash_error(:advanced_search_bad_q_error.t)
+    redirect_to(controller: "observer", action: "advanced_search")
+  end
+
   private ##########
 
   def map_past_bys(args)
@@ -1178,6 +1189,11 @@ class ApplicationController < ActionController::Base
 
   def query_needs_update?(new_args, query)
     new_args.any? { |_arg, val| query.params[:arg] != val }
+  end
+
+  def invalid_q_param?
+    params && params[:q] &&
+      !QueryRecord.where(id: params[:q].dealphabetize).exists?
   end
 
   public ##########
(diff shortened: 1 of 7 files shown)


JoeCohen:
#### Questions
- Was this ticket still open? [Pivotal #24489297](https://www.pivotaltracker.com/story/show/24489297) says (emphasis added):
- Should I simlple throw out this PR and close the ticket. 

The ticket says:
> **Crashes** in advanced_search when q=XXX expires?
Not absolutely sure, but it looks like http://mo.org/image/advanced_search?q=XXX causes "undefined method `id' for nil:NilClass" error when the query associated with XXX expires.

It does cause that error, but the error is rescued. So maybe the issue was fixed by rescuing the error. (Jason opened this ticket: 6 Feb 2012. It's our oldest Bug ticket.)

Thus, what this PR does is: provide a message that makes more sense to non-technical users. But the fix strikes me a somewhat smelly. (I can't think of a better way to fix it without substantial re-writing.) So maybe we should just throw out this PR and close the ticket. 

(This is something I started looking at prior to yesterday's technical meeting. I wanted to address it while it was still on my mind.)

app/controllers/application_controller.rb
```diff
@@ -1100,6 +1100,17 @@ def find_query(model = nil, update = !browser.bot?)
...
+  # so that they get just one flash msg if the query has expired.
+  # This method avoids a call to find_safe, which would add
+  # "undefined method `id' for nil:NilClass" if there's no QueryRecord for q
+  def handle_advanced_search_invalid_q_param?
+    return unless invalid_q_param?
+
+    flash_error(:advanced_search_bad_q_error.t)
+    redirect_to(observer_advanced_search_form_path)
+  end
+
```
JoeCohen, line 1113: Is there a better place for this method?

mo-nathan:
This looks reasonable to me.  I don't see any harm in merging this.  I believe the deeper desire is to get rid of the q parameter entirely and actually track the query parameters in the URL, but that's a lot of work.

pellaea:
Just so it's said... it might not actually be so hard to replace the q
mechanism!  A first draft would be simply replace the value of q parameter,
currently = alphabetize(query_record.id), with the actual parameters =
query_record.descritpion.  Then, do away with query_record altogether.  I
think query_record.description is just the query.params hash serialized in
some way.  It will certainly result in larger URLs, *way* longer in some
cases presumably.  But that is after all what we are talking about doing.
Nice and RESTful.  Well mostly: there will still be a minimal session.

PR #738, opened 2021-01-11
Replace coveralls gem with coveralls_reborn

JoeCohen:
Makes Travis CI-Coveralls integration work again.
A side effect is that a few other germs are updated.

Delivers https://www.pivotaltracker.com/story/show/175663051

There is no manual test script.


mo-nathan:
Look good!  Thanks for hunting this down.

PR #740, opened 2021-01-25
HerbariaController

JoeCohen:
Modernize HerbariaController, including:

- Standardize names & routes for basic CRUD actions. (`index, show, new/create, edit/update, destroy`) 
- Move nonstandard actions into namespaced controllers as standard CRUD actions. 
Ex: `get Herbarium#curator_request` => `get Herbaria::CuratorRequest#new`

Delivers https://www.pivotaltracker.com/story/show/174310149


JoeCohen:
This is an incomplete work-in-progress, so I'm not requesting general review at this time. 
(I don't mind them but I don't think it's a good use of other's time to generally review this PR now.)
I'm creating the draft PR because I will have specific questions, and this is a good place to ask them.
(I also like the diff UI.)

JoeCohen:
This is in good enough shape that I'd like on two things:
- Overall controller/action organization;
- Organizing the actions that call `ApplicationController#show_index_of_objects`
###  Overall controller/action organization
Controllers are pluralized. (`HerbariumController` => `HerbariaController`)
The only actions are `index, show, new/create, edit/update, destroy`. Any other action
has been:
- renamed. Ex: `list` => `HerbariaController#index`, or
- put in a namespaced controller and renamed. 
Ex: `request_to_be_curator (get)` =>  `Herbaria::CuratorRequest#new`
- (The exceptions are `next_herbarium` and `prev_herbarium`, which I merely renamed. I did not namespace them because of issues with ApplicationController and `_prev_next_pager.html.erb`
```
  # Old MO Action (method)        New "Normalized" Action (method)
  # ----------------------        --------------------------------
  # create_herbarium (get)        new
  # create_herbarium (post)       create
  # delete_curator (delete)       Herbaria::Curators#destroy
  # destroy_herbarium             destroy
  # edit_herbarium (get)          edit
  # edit_herbarium (post)         update
  # herbarium_search (get)        Herbaria::Searches#index (get)
  # index (get)                   Herbaria::Nonpersonals#index (get)
  # index_herbarium (get)         Herbaria::Filtereds#index (get)
  # list_herbaria (get)           index
  # merge_herbaria (get)          Herbaria::Merges#new (get)
  # next_herbarium (get)          next
  # prev_herbarium (get)          prev
  # request_to_be_curator (get)   Herbaria::CuratorRequest#new
  # request_to_be_curator (post)  Herbaria::CuratorRequest#create
  # show_herbarium (get)          show
  # show_herbarium (post)         Herbaria::Curators#create
```

pellaea:
Thanks for the summary of what the old actions become in the new namespaced
controller edifice!

JoeCohen:
I hit Comment too soon. I'm still working on the comment. It will be done
in a few minutes.

JoeCohen:
I am about to make this Ready for Review. 
I'm looking for advice/comments/help in three areas. 
- Overall controller pattern(s). I'll post a separate Comment about this.
- Creating routes to redirect the old actions to the new ones. I need help and will post a separate Comment.
- Miscellaneous questions about implementation details. I will comment in the code.

JoeCohen:
### Controller patterns
Any advice, comments, etc. on:
#### Using only the 7 Rails CRUD actions (`index, show, new, create, edit, update, destroy`). 
Espoused by David Heinemeier Hanson, see [How DHH Organizes His Rails Controllers
](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/), but isn't a Rails standard.
Results in fewer actions, skinnier controllers (sometimes anorexic?), and more controllers.
Requires namespacing controllers and/or combining actions.
(It isn't how I did `ArticlesController` or `GlossaryTermsController`. But it would be little work to conform them to this pattern.)
#### Namespacing vs Combining actions with flags
I ended up namespacing practically everything in order to use only the 7 actions. For example, I have
`Herbaria#index`, `Herbaria::Nonpersonals#index`, `Herbaria::Alls#index`. 
The namespacing is different from what we discussed earlier. I ended up disliking the use of parameters plus a test in the action; they seem kind of smelly ... sort of case statements or switches, that complicate each action. They also make the views a bit more complicated.
The only exception I made was combining the old `Herbarium#next_herbarium` and `Herbarium#prev_herbarium` into `Herbaria::Nexts#show` with a new `nexts` parameter. The actions have identical filters, are almost identical in substance, and there's only on place in the views that needed changing,

JoeCohen:
### Redirecting legacy actions
I'm looking for help with a few of these. In particular, redirecting

| old | new |
| --- | --- |
| `post Herbarium#delete_curator` | `post Herbaria::Curators#destroy` |
| `post Herbarium#request_to_be_curator` | `post Herbaria::Curator_Requests#create` |
| `post Herbarium#show_herbarium` | `post Herbaria::Curators#create` |

I can route the old ones to the new ones. (At least that's what the tests seem to prove, but I'm not 100% confident of that.) But it takes me two lines of routing, and Rails seems to follow the redirect. I hope I'm making some newbie mistake that results from a basic misunderstanding of routing to redirects.

For details, see `routes.rb` and `test/integration/redirects_test.rb`

app/controllers/application_controller.rb
```diff
@@ -1018,11 +1018,28 @@ def query_params(query = nil)
...
-    params
+    return params if browser.bot?
+
+    query_param = get_query_param(query)
+    if params.is_a?(String) # i.e., if params is a path
+      append_query_param_to_path(params, query_param)
+    else
+      params[:q] = query_param
+      params
+    end
```
JoeCohen, line 1029: These changes make `add_query_param` work with Rails path helpers (in addition to controller/action hashes)

app/controllers/herbaria/alls_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    # list_herbaria (get)           Herbaria::All#index (get)
+
+    # ---------- Actions to Display data (index, show, etc.) -------------------
+
+    # Display all Herbaria
+    # linked (conditionally) from HerbariaIndex
+    def index
+      query = create_query(:Herbarium, :all, by: :name)
+      show_selected_herbaria(query, always_index: true)
+    end
```
JoeCohen, line 20: Questions:
- **Should I combine all the index-type actions as `Herbaria#index`** by: 
  - always calling Herbaria#index with a parameter (e.g., `index-type`)
  - using a `case` statement that depends on that parameter.
- **Should I interchange `Herbaria#index` and `Herbaria::Alls#index`?**  
This action is the former `Herbarium#list`, and lists _all_ Herbaria. I think the action that lists all model instances would typically be `#index`. It turned out to be _much_ simpler to translate the old actions as follows:

| old | new |
| --- | ---|
| Herbarium#index | Herbaria::Nonpersonals#index |
| Herbarium#index_herbarium | Herbaria#index - lists query results |
| Herbarium#list_herbaria | Herbaria::Alls#index - all herbaria |

app/controllers/herbaria/nexts_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Herbaria
+  # redirect to next/prev herbarium
+  class NextsController < ApplicationController
```
JoeCohen, line 5: is there a short, less awkward name for this controller?

app/controllers/herbaria/shared_private_methods.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
+
+# private methods shared by HerbariaController and subcontrollers
+module Herbaria::SharedPrivateMethods
```
JoeCohen, line 4: Is there a better or more typical way to share private actions among `HerbariaController` and namespaced controllers?

app/controllers/herbaria_controller.rb
```diff
@@ -0,0 +1,273 @@
...
+    @herbarium.description = @herbarium.description.to_s.strip
+    @herbarium.code = "" if @herbarium.personal_user_id
+  end
+
+  def validate_herbarium!
+    validate_name! &&
+      validate_location! &&
+      validate_personal_herbarium! &&
+      validate_admin_personal_user!
+  end
```
JoeCohen, line 132: IMO. If possible, the validation should be done in the model. I'm not sure that's possible here, so I've made it into a separate story so as not to delay this PR.

app/controllers/observer_controller/search_controller.rb
```diff
@@ -23,9 +24,16 @@ def pattern_search
...
+    when :comment, :image, :location, :name, :project, :species_list,
+      :herbarium_record
       ctrlr = type
+    when :herbarium
+      redirect_to_search_or_index(
+        pattern: pattern,
+        search_path: herbaria_searches_path(pattern: pattern),
+        index_path: herbaria_alls_path
+      )
+      return
```
JoeCohen, line 36: As more of our legacy controllers get updated, these lines should be generalized. I'd prefer to wait on that. 
**Question: Should I instead generalize now, based on the one example?**

app/helpers/application_helper.rb
```diff
@@ -3,10 +3,41 @@
 #
 #  = Application Helpers
 #
-#  These methods are available to all templates in the application:
+#  Methods available to all templates in the application:
```
JoeCohen, line 6: Documents a list of methods for convenience. I've found that handier than looking through the whole file.

app/helpers/application_helper.rb
```diff
@@ -53,13 +86,57 @@ def link_to_coerced_query(query, model)
...
+  def destroy_button(target:, name: :DESTROY.t)
+    options = if target.is_a?(String)
+                target
+              else
+                { action: "destroy", id: target.id }
+              end
+    button_to(
+      name, options, method: :delete, data: { confirm: :are_you_sure.t }
+    )
+  end
```
JoeCohen, line 129: Generalizes parameters to accept a path or object as `:target`.
See #b579c28

app/helpers/herbaria_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# helper methods for Herbaria views
+module HerbariaHelper
+  def herbarium_top_users(herbarium)
+    Herbarium.connection.select_rows(%(
+      SELECT u.name, u.login, COUNT(u.id)
+      FROM herbarium_records hr JOIN users u ON u.id = hr.user_id
+      WHERE hr.herbarium_id = #{herbarium.id}
+      GROUP BY u.id ORDER BY COUNT(u.id) DESC LIMIT 5
+    ))
```
JoeCohen, line 11: I just wanted to get this SQL query out of the template.
I like to turn it into an ActiveRecord query, but couldn't figure out how. (And maybe it's not possible.)

app/helpers/object_link_helper.rb
```diff
@@ -115,10 +115,10 @@ def species_fungorum_sf_synonymy(record_id)
   def user_link(user, name = nil)
     if user.is_a?(Integer)
       name ||= "#{:USER.t} ##{user}"
-      link_to(name, User.show_link_args(user))
+      link_to(name, "#{observer_show_user_path}/#{user}")
     elsif user
       name ||= user.unique_text_name
-      link_to(name, user.show_link_args)
+      link_to(name, "#{observer_show_user_path}/#{user.id}")
```
JoeCohen, line 121: This prevents namespacing the link to `show_user`.
See #a0db86e472ec0d561181fcfc7e646a8996bef886

config/routes.rb
```diff
@@ -668,9 +678,66 @@ def route_actions_hash
...
+       to: redirect(path: "/herbaria/curator_requests?id=%{id}"))
+  get("/herbaria/curator_requests",
+      to: "herbaria/curator_requests#create", id: /\d+/)
+  # The next post and get combine to redirect the legacy
+  #   POST /herbarium/show_herbarium/:id to
+  #   POST herbaria/curators#create
+  post("/herbarium/show_herbarium", to: redirect(path: "herbaria/curators"))
+  get("/herbaria/curators", to: "herbaria/curators#create", id: /\d+/)
+  get("/herbarium", to: redirect(path: "herbaria/nonpersonals#index"))
+  # rubocop:enable Style/FormatStringToken
```
JoeCohen, line 737: Huge and ugly.
Questions:
- Can it be simplified?
- Can it be extracted in order to reduce the size of the containing block (`MushroomObserver::Application.routes.draw do`)?
- How can I fix the routing of the problematic legacy actions? See https://github.com/MushroomObserver/mushroom-observer/pull/740#issuecomment-782729221

test/controllers/articles_controller_test.rb
```diff
@@ -22,7 +22,9 @@ def test_index_filtered
     query = Query.lookup(:Article, :in_set, ids: [article.id])
     params = @controller.query_params(query)
     get(:index, params: params)
-    assert_select("a:match('href',?)", %r{/articles/\d+}, { count: 1 },
+
+    assert_select("div#content a:match('href',?)", %r{/articles/\d+},
```
JoeCohen, line 26: Fixes a failure caused by adding an article link to the banner.

test/integration/curator_test.rb
```diff
@@ -151,36 +201,104 @@ def test_herbarium_change_code
...
+      form.assert_value("[code]", "")
+      form.assert_value("[place_name]", "")
+      form.assert_value("[email]", "")
+      form.assert_value("[mailing_address]", "")
+      form.assert_value("[description]", "")
+      form.assert_unchecked("[personal]")
+
+      form.change("[name]", "Mary's Herbarium")
+      form.check("[personal]")
+      form.submit(:CREATE.l)
```
JoeCohen, line 240: New style forms (using [Model Object Helpers](https://guides.rubyonrails.org/v5.2/form_helpers.html#dealing-with-model-objects-model-object-helpers) instead of `_tag` do not create `id`s. 
So I tweaked/hacked `test/session_form_extensions.rb` to accept pseudo-`name`s.

test/integration/redirects_test.rb
```diff
@@ -142,4 +162,197 @@ def test_show_past_glossary_term
...
+  # request_to_be_curator (post)  Herbaria::CuratorRequest#create (post)
+  def test_request_to_be_herbarium_curator_post
+    nybg = herbaria(:nybg_herbarium)
+    email_count = ActionMailer::Base.deliveries.count
+    login("mary")
+    post("/herbarium/request_to_be_curator/#{nybg.id}")
+
+    # Rails seems to follow the redirect, instead of just redirecting
+    assert_equal(herbarium_path(nybg), @response.request.fullpath)
+    assert_equal(email_count + 1, ActionMailer::Base.deliveries.count)
```
JoeCohen, line 336: I don't understand why the response is :success rather than a 301. 
Obviously a consequence of my routing. 
How can I fix this?

test/integration/redirects_test.rb
```diff
@@ -142,4 +162,197 @@ def test_show_past_glossary_term
...
+
+  # show_herbarium (post)         Herbaria::Curators#create (post)
+  def test_show_herbarium_post
+    nybg = herbaria(:nybg_herbarium)
+    assert(nybg.curators.include?(rolf))
+    curator_count = nybg.curators.count
+    login("rolf")
+    post("/herbarium/show_herbarium?id=#{nybg.id}&add_curator=#{mary.login}")
+
+    assert_equal(herbarium_path(nybg), @response.request.fullpath)
```
JoeCohen, line 355: I don't understand why the response is :success rather than a 301. 
Obviously a consequence of my routing. 
How can I fix this?

test/integration/redirects_test.rb
```diff
@@ -142,4 +162,197 @@ def test_show_past_glossary_term
...
+  # merge_herbaria (get)          Herbaria::Merges#new (get)
+  def test_merge_herbaria
+    fundis = herbaria(:fundis_herbarium)
+    assert_true(fundis.owns_all_records?(mary))
+    marys = mary.create_personal_herbarium
+    login("mary")
+
+    get("/herbarium/merge_herbaria?this=#{fundis.id}&that=#{marys.id}")
+    # fundis ends up being the destination because it is older.
+    assert_equal(herbaria_path(id: fundis.id), @response.request.fullpath)
```
JoeCohen, line 282: I don't understand why the response is :success rather than a 301. 
Obviously a consequence of my routing. 
How can I fix this?

test/session_form_extensions.rb
```diff
@@ -219,6 +221,67 @@ def get_field!(id)
       get_field(id, :strict)
     end
 
+    # Find the field whose name ends in the given String or matches the given
```
JoeCohen, line 224: New style forms (using [Model Object Helpers](https://guides.rubyonrails.org/v5.2/form_helpers.html#dealing-with-model-objects-model-object-helpers) instead of `_tag` do not create `id`s. 
So I tweaked/hacked `test/session_form_extensions.rb` to accept pseudo-`name`s.

config/routes.rb
```diff
@@ -668,9 +678,66 @@ def route_actions_hash
...
+  #   GET("/herbarium/delete_curator/nnn?user=uuu") and
+  #   POST("/herbarium/delete_curator/nnn?user=uuu")
+  # to
+  #   GET("/herbaria/curators/nnn?user=uuu")
+  # Therefore we need the following "get" to prevent
+  #   No route matches [GET] "/herbaria/curators/nnnnn"
+  match("/herbarium/delete_curator/:id",
+        to: redirect(path: "/herbaria/curators/%{id}"),
+        via: [:get, :post])
+  get("/herbaria/curators/:id", to: "herbaria/curators#destroy", id: /\d+/)
```
JoeCohen, line 711: I am concerned about the security implications of:
- the `match` because it routes both `get` and `post` to the same action; and
- the `get` because it writes to the db.

See https://guides.rubyonrails.org/v5.2/routing.html#http-verb-constraints

config/routes.rb
```diff
@@ -668,9 +678,66 @@ def route_actions_hash
...
+  #   POST herbaria/curator_requests#create
+  post("/herbarium/request_to_be_curator/:id",
+       to: redirect(path: "/herbaria/curator_requests?id=%{id}"))
+  get("/herbaria/curator_requests",
+      to: "herbaria/curator_requests#create", id: /\d+/)
+  # The next post and get combine to redirect the legacy
+  #   POST /herbarium/show_herbarium/:id to
+  #   POST herbaria/curators#create
+  post("/herbarium/show_herbarium", to: redirect(path: "herbaria/curators"))
+  get("/herbaria/curators", to: "herbaria/curators#create", id: /\d+/)
```
JoeCohen, line 735: I am concerned about both of these routings because the `get` writes to the db. See https://guides.rubyonrails.org/v5.2/routing.html#http-verb-constraints

app/controllers/observer_controller/search_controller.rb
nimmolo: I do like the idea of generalizing, if it can be done. The action seems pretty standard.

app/helpers/herbaria_helper.rb
nimmolo: Nice

app/helpers/object_link_helper.rb
nimmolo: I'm guessing the "observer_show_user_path" will get a revision when observer controller is broken up

config/routes.rb
nimmolo: Got it. Since i'm relatively out of the loop lately, let me ask: 

at this point how usable are the named path helpers like "herbaria_curators_path"... or are we still planning to move towards them? 

I believe with default routing the destroy action should work with a DELETE request to "herbaria_curator_path(:id)". But maybe the legacy requests are GETs, so you need to route it this way.

test/session_form_extensions.rb
nimmolo: hmmm. Looking into it...

app/controllers/herbaria/alls_controller.rb
nimmolo: My initial reaction is, "yes" to question 1.  I feel like just on principle, the "Alls" controller seems like it should be unnecessary.

But that seems easy to accomplish if we follow the pattern where Herbaria#index (or any controller's index action) would either list all herbaria, or list *selected* herbaria if it's passed a q param. 

fwiw this is something i had been experimenting with, but i haven't poked around enough to see if you've got an example of that available elsewhere here.

app/controllers/herbaria/nexts_controller.rb
nimmolo: How about just enabling the herbaria#show action to parse next/prev params with a case statement? Or is that not possible. I feel like I've read through examples of simple CRUD controllers in the wild that handle next/previous, without another controller.

app/controllers/herbaria/shared_private_methods.rb
nimmolo: Hm. Seems like the first ones (`keep_track_of_referrer` and `redirect_to_referrer`) could maybe be inherited from Application_Controller.

test/integration/redirects_test.rb
nimmolo: Should the action at Herbaria::Merges#new be rather Herbaria::Merges#create, and then instead of `get("/herbarium/merge_herbaria?this=#{fundis.id}&that=#{marys.id}")
` you'd do a `post` to `merge_herbaria`?

test/integration/redirects_test.rb
nimmolo: Oof, i guess i was wrong in my comment below, suggesting that a `post` action would give you the 301 response. If i recall correctly I never resolved this; I don't think Rails ever gives the 301 as a response, maybe it's the way we have it configured. I feel like my tests ALWAYS followed the redirect and gave a fairly useless `:success` response message. Ugh. There must be a standard way to test this kind of thing.

JoeCohen:
@nimmolo: Thanks for the review. I will make time to look at it this weekend.

app/controllers/observer_controller/search_controller.rb
JoeCohen: Thanks for the input @nimmolo.
I will re-think this (both generalization and whether to route searches to #index.

app/helpers/object_link_helper.rb
JoeCohen: Good point. Yes, that will need revision when ObserverController is updated.

test/integration/redirects_test.rb
JoeCohen: Good point. 
I was (and still am) confused about what should be the right action.
Under some circumstances it actually performs the merger, which is both an update of one herbarium and a destroy of the other. In other cases, it doesn't change any herbarium, but merely sends an email to the webmaster.

config/routes.rb
JoeCohen: I'd like to move toward the path helpers for the updated controllers.
I think the legacy requests are GETs. (I put a breakpoint at the top of Herbarium#destroy_herbarium, clicked `Destroy Fungarium` from an appropriate `show_herbarium` page, and -- when MO got to the breakpoint, called `request.method`:
```ruby
   141:   def destroy_herbarium
   142: byebug
...
(byebug) request.method
"GET"
```

JoeCohen:
Converted to draft. I intend to combine some actions into `#index`, then use a Case statement or other switch to differentiate them. Will also be looking more carefully at @nimmolo's other suggestions.

test/integration/redirects_test.rb
JoeCohen: Interesting. Almost all the methods in this file respond with a 301. This is one of the few exceptions.

config/routes.rb
```diff
@@ -668,6 +653,73 @@ def route_actions_hash
...
+
+  # Herbaria: broken redirects of legacy Herbarium actions
+  get("/herbarium/next_herbarium/:id",
+      to: redirect(path: "herbaria/%{id}?flow=next"))
+  get("/herbarium/prev_herbarium/:id",
+      to: redirect(path: "herbaria/%{id}?flow=prev"))
+  get("/herbarium/merge_herbaria",
+      to: redirect(path: "herbaria/merges#create"), that: /\d+/, this: /\d+/)
```
JoeCohen, line 715: ### These redirects are broken. See test output below.
### `next` and `prev`
I can't figure out how to add a new, static parameter while retaining the `q` param.
Potential solutions:
- My preferred solution: Don't redirect the legacy action. I don't think this will affect anyone. There shouldn't be old links in the cloud that point to the legacy `next` and `prev` actions
- 2nd choice: A namespaced controller with a `show` action and a param to control direction. Somehow the analogous routes work for a separate controller, e.g.
```ruby
  get("/herbarium/next_herbarium/:id",
      to: redirect(path: "herbaria/nexts/%{id}?next=next"))
```
- 3rd choice: Separate non-CRUD `next` and `prev` actions in HerbariaController. 
### merge
I can't figure out how to route the legacy GET action to a new POST action.
Potential solutions:
- My preferred solution: Don't redirect the legacy action. There shouldn't be a link in the cloud that points to a merger.
- 2nd choice: Use `new` instead of `create`. This works, but has security implications; Rails advises _never_ writing to the db from a GET. See https://guides.rubyonrails.org/v5.2/routing.html#http-verb-constraints
-----
```sh
 FAIL["test_next_herbarium", #<Minitest::Reporters::Suite:0x000055613a15f5b8 @name="RedirectsTest">, 8.719991796999238]
 test_next_herbarium#RedirectsTest (8.72s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"/herbaria/1022913654?flow=next&q=1Wi"
        +"/herbaria/1022913654?flow=next?q=1Wi"
        test/integration/redirects_test.rb:22:in `assert_old_url_redirects_to_new_path'
        test/integration/redirects_test.rb:291:in `test_next_herbarium'

 FAIL["test_prev_herbarium", #<Minitest::Reporters::Suite:0x00007fecec57de48 @name="RedirectsTest">, 12.29093096699944]
 test_prev_herbarium#RedirectsTest (12.29s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"/herbaria/1022913654?flow=prev&q=1Wl"
        +"/herbaria/735604515?flow=prev?q=1Wl"
        test/integration/redirects_test.rb:22:in `assert_old_url_redirects_to_new_path'
        test/integration/redirects_test.rb:308:in `test_prev_herbarium'
```

mo-nathan: I think it's perfectly reasonable in these cases not to redirect the legacy action.  As you say, it's very unlikely that these have been saved as links any where and even if they are I doubt anyone will complain about it.

config/routes.rb
```diff
@@ -668,6 +653,66 @@ def route_actions_hash
...
+        via: [:get, :post])
+  get("/herbaria/curators/:id", to: "herbaria/curators#destroy", id: /\d+/)
+
+  # The next post and get combine to redirect the legacy
+  #   POST /herbarium/request_to_be_curator to
+  #   POST herbaria/curator_requests#create
+  post("/herbarium/request_to_be_curator/:id",
+       to: redirect(path: "/herbaria/curator_requests?id=%{id}"))
+  get("/herbaria/curator_requests",
+      to: "herbaria/curator_requests#create", id: /\d+/)
```
JoeCohen, line 702: ### Should I also remove these redirect routes?
Although they work, they raise issues similar to the routes removed in [Commit #0069792](https://github.com/MushroomObserver/mushroom-observer/pull/740/commits/006979285243a1086b89f880ce590864123b3272): using a GET to write data, unlikely that there are links that point to the legacy actions, and even if such links existed, it's unlikely anyone would complain.

config/routes.rb
```diff
@@ -668,6 +653,66 @@ def route_actions_hash
...
+
+  # The next post and get combine to redirect
+  #   POST /herbarium/show_herbarium/:id to
+  #   POST herbaria/curators#create
+  post("/herbarium/show_herbarium", to: redirect(path: "herbaria/curators"))
+  get("/herbaria/curators", to: "herbaria/curators#create", id: /\d+/)
```
JoeCohen, line 708: ### Should I also remove these redirect routes?
Although they work, they raise issues similar to the routes removed in [Commit #0069792](https://github.com/MushroomObserver/mushroom-observer/pull/740/commits/006979285243a1086b89f880ce590864123b3272): using a GET to write data, unlikely that there are links that point to the legacy actions, and even if such links existed, it's unlikely anyone would complain.

PR #745, opened 2021-03-07
Logo contest

mo-nathan:
This implements the logo contest collection interfaces.  Got to:

http://localhost:3000/contest_entries/new

to create new logo contest entries.  You must be in admin mode.

Go to:

http://localhost:3000/contest_entries

to see all the current entries (again you must be an admin)

Go to:

http://localhost:3000/contest_ballots

To see the current user's ballot.  Should be available to any registered user.


JoeCohen:
Thanks for doing this! Sorry I haven't been able to contribute/
I will look @ Logo Contest PR tomorrow (Weds) a.m.

app/views/contest_ballots/index.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  <%= form_for(:contest_ballot, url: {action: :create}) do |form| %>
+    <div class="hidden">
+      <%= submit_tag "Submit" %>
+    </div>
+    <div><h3>
+      <%= form.check_box(:confirmed,
+                         { checked: @votes.first.confirmed,
+                           onclick: 'this.form.submit()' }) %>
+      <%= :contest_ballot_check_text.t %>
+    </h3></div>
```
JoeCohen, line 19: I suggest moving this to the bottom of the form, per the instructions ("When you are ready for your ballot to count, check the box at the _bottom_ of the ballot." Emphasis added.)
```suggestion
```

app/views/contest_ballots/index.html.erb
```diff
@@ -0,0 +1,47 @@
...
+            <%= content_tag(:big, thumbnail(entry.image_id, {votes: false})) %>
+          </td>
+          <td style="padding-top:40px;padding-bottom:20px">
+            <% if entry.alternate_image_id %>
+              <%= content_tag(:big, thumbnail(entry.alternate_image_id, {votes: false})) %>
+            <% end %>
+          </td>
+        </tr>
+      <% end %>
+    </table>
```
JoeCohen, line 45:
```suggestion
   <div><h3>
      <%= form.check_box(:confirmed,
                         { checked: @votes.first.confirmed,
                           onclick: 'this.form.submit()' }) %>
      <%= :contest_ballot_check_text.t %>
    </h3></div>
```
See above comment

JoeCohen:
Looks good so far. (Obviously there's more to come in terms of weighting and tallying votes.)
I played around a tiny bit with it in dev mode, and glanced through the code. Made one suggestion above.
**Suggestion: Maybe add `ContestEntries#edit`, `update`, and `destroy` in case you make a mistake when creating an entry?**
(If not, maybe prune the routes, i.e. `resources :contest_entries, only: [:new, :create, :index, :show], id: /\d+/`)

app/views/contest_ballots/index.html.erb
mo-nathan: Thanks for catching this inconsistency.  I actually want to revise the wording rather than move the box.  It was at the bottom and I moved it since it was not obvious.

mo-nathan:
As you suggest I would like to added the edit, update and destroy, but figured it wasn't high priority.  That's why I didn't take it out of the routes.  Given that we're presumably going to get rid of this stuff once the contest is over, I figured it wasn't critical to resolve this issue.

PR #747, opened 2021-03-17
Adjust fancy_time for dst-to-standard change

JoeCohen:
Fixes this bug:
```sh
 FAIL["test_fancy_time", #<Minitest::Reporters::Suite:0x000055db6acfa940 @name="TimeExtensionsTest">, 2.8948509150000064]
 test_fancy_time#TimeExtensionsTest (2.89s)
        Expected: "last week, 2021-03-10"
          Actual: "6 days ago, 2021-03-10"
        test/models/time_extensions_test.rb:98:in `assert_fancy_time'
        test/models/time_extensions_test.rb:87:in `test_fancy_time'
```

Delivers https://www.pivotaltracker.com/story/show/177378817

### DIFF
--- a/app/extensions/time_extensions.rb
+++ b/app/extensions/time_extensions.rb
@@ -90,13 +90,10 @@ class ActiveSupport::TimeWithZone
   private
 
   def diff_corrected_for_dst_change(ref)
-    if dst? && !ref.dst?
-      ref - self - 1.hour
-    elsif !dst? && ref.dst?
-      ref - self + 1.hour
-    else
-      ref - self
-    end
+    return ref - self - 1.hour if dst? && !ref.dst?
+    return ref - self + 1.hour if !dst? && ref.dst?
+
+    ref - self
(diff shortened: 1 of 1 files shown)


JoeCohen:
`master` breaks for one week after a change from standard to daylight time.
The case that breaks things is `test/models/time_extensions_test.rb:87:in `test_fancy_time'`:
```ruby
    assert_fancy_time(7.days, :time_one_week_ago)
```
This PR fixes that, and other potential breakages.

Another possible "fix" is not testing the above edge case.

Comment for future reference (I don't have time to deal with it now.)
I personally dislike fancy_time because the static output can quickly become inaccurate. For instance "5 minutes ago" becomes inaccurate in a minute. I'd prefer to get rid of fancy_time and just have yyyy mm dd hh mm ss
But if we need a "fancy" output, why not use `distance_of_time_in_words` instead of rolling our own.

PR #755, opened 2021-04-11
SpeciesList year

JoeCohen:
Prevent users from creating a SpeciesList with a nil `when`.

- Uses Observation validation logic to validate `SpeciesList.when`
- Extract that logic to a Concern, using the concern in both Observation and SpeciesList
- Delivers https://www.pivotaltracker.com/story/show/177682975

### Manual Test 
- Create a SpeciesList, blanking out the Year in When.
Expected Result: It should create and display the list, with When set to the date the list was created


JoeCohen:
### Question:
Is a Concern an appropriate/the best way for models to share a custom validation?
If not, how?

mo-nathan:
> ### Question:
> Is a Concern an appropriate/the best way for models to share a custom validation?
> If not, how?

Personally I'm not a fan of concerns.  They are too much like multiple inheritance and it's just way too easy to create overly complicated and subtly buggy code.  I would probably implement this as a simple module that takes a 'when' and an 'errors'.  It doesn't really have any state that we care about.

JoeCohen:
Thanks you. That's exactly the kind of feedback I was seeking.

JoeCohen:
> > ### Question:
> > Is a Concern an appropriate/the best way for models to share a custom validation?
> > If not, how?
> 
> Personally I'm not a fan of concerns. They are too much like multiple inheritance and it's just way too easy to create overly complicated and subtly buggy code. I would probably implement this as a simple module that takes a 'when' and an 'errors'. It doesn't really have any state that we care about.

I made it a simple module, rather than a Concern.
I either don't understand your comment about taking a `when` and `errors`, or I don't see the advantage of doing it that way or how it leads to less complicated code.

app/models/validations.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+# validations used by multiple objects
+module Validations
+  # Validates obj.when.
+  # Includes setting it to current date if user set it to nil.
+  # used by: Observation, SpeciesList
+  def validate_when
+    self.when ||= Time.zone.now
```
mo-nathan, line 9: I believe this line is causing this to act effectively as a concern.  Any reference to "self" means you are adding on to a class which is what I am concerned about.  It makes it really hard to track down where things are happening.  Similarly, looking at this file it is really unclear what errors without looking at where this code is called from.  Instead, I would write this as something like:

def time_validation(when, errors)
  check_date(when, errors) && check_time(when, errors) && check_year(when, errors)
end

and then move the update of self.when back into the class where it belongs.

JoeCohen: @mo-nathan:
Thanks for taking the time to explain this. Sorry to be slow on the uptake.
I just pushed something that I think is fully responsive to your Comments.
It's less dry and wordier than a Concern: 
- It repeatedly passes arguments instead of relying on `self.where` 
- Adds a little duplication between `observation.rb` and `species_list.rb` for the same reason. 

But that looks to me like that's a necessary tradeoff for avoiding multiple inheritance. Is there a better way?

mo-nathan (approved):
Looks good!  Thanks for the doing the refactoring.

JoeCohen:
And thank you for getting to this so quickly and for the repeated reviews!

PR #761, opened 2021-04-25
Add new logo to top of left hand nav menu

mo-nathan:
Would love to have some else try this out before it goes live.  Especially on OSes other than Ubuntu and browsers other than Chrome and Firefox since it seems to work fine for me on those two.


pellaea:
Code-wise this looks straightforward enough.  I haven't booted the dev VM in ages.  I'll give it a shot to see how it looks "live" and to ensure that it autosizes, etc. correctly for me, too.

mo-nathan:
Thanks!

PR #772, opened 2021-07-26
Change icn_id dependencies

JoeCohen:
Enable sending admin an email when user changes identifier of a Name that has dependencies.
- Delivers https://www.pivotaltracker.com/story/show/178233771
- Fixes a bug which occurs when user tries to change `icn_id` of a Name having dependencies:
MO would simply re-display the Edit form without any changes.

### Suggested Manual Test ###
- Edit a Name that has both dependencies and an icn_id. [_Hydnum_
](https://mushroomobserver.org/name/show_past_name/805) should work.
- Change the ICN id
- Save changes
**Expected result**: Send Name Change Request form
- Send
**Expected result**: redirected to show_name page with flash success message

### DIFF
--- a/app/controllers/name_controller/create_and_edit_name.rb
+++ b/app/controllers/name_controller/create_and_edit_name.rb
@@ -119,7 +119,8 @@ class NameController
         params: {
           name_id: @name.id,
           # Auricularia Bull. [#17132]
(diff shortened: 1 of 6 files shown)


JoeCohen:
Closing this:
- I found a bug in a display
- Need to fix coverage decrease

PR #774, opened 2021-08-01
Herbaria #Records reversed

JoeCohen:
Fixes a bug in Herbaria by #Records reversed
- Delivers https://www.pivotaltracker.com/story/show/176833509
- The bug was that sorting links were not displayed.
- The links were not displayed because `Query#num_results` returned the wrong value (1) for #Records reversed. (It also returned the wrong value for #Records not reversed.

#### Suggested Manual Test ###
- http://localhost:3000/herbarium or http://localhost:3000/herbaria?flavor=all
- Sort by: #Records
- Sort by: Reverse Order
**Expected Result**: Page includes sort tabs

This works, but seems like a kluge.
It feels like the correction should be made somewhere in one of the Query Herbaria modules.
But I can't figure out where or how.

### DIFF
--- a/app/classes/query/image_advanced_search.rb
+++ b/app/classes/query/image_advanced_search.rb
@@ -31,12 +31,8 @@ class Query::ImageAdvancedSearch < Query::ImageBase
     # [Sorry, yes, this is a mess. But I don't expect this type of search
     # to survive much longer. Image searches are in desperate need of
     # critical revision for performance concerns, anyway. -JPH 20210809]
-    args2 = args.dup
-    args2.delete(:select)
-    args2.delete(:order)
-    args2.delete(:group)
-    params2 = params.dup
-    params2.delete(:by)
+    args2 = args.except(:select, :order, :group)
+    params2 = params.except(:by)
     ids = Query.lookup(:Observation, flavor, params2).result_ids(args2)
     ids = clean_id_set(ids)
     args2 = args.dup
--- a/app/classes/query/image_advanced_search.rb
+++ b/app/classes/query/image_advanced_search.rb
@@ -28,8 +28,13 @@ class Query::ImageAdvancedSearch < Query::ImageBase
   end
 
   def execute_content_search(args)
+    # [Sorry, yes, this is a mess. But I don't expect this type of search
+    # to survive much longer. Image searches are in desperate need of
+    # critical revision for performance concerns, anyway. -JPH 20210809]
     args2 = args.dup
     args2.delete(:select)
+    args2.delete(:order)
+    args2.delete(:group)
     params2 = params.dup
     params2.delete(:by)
     ids = Query.lookup(:Observation, flavor, params2).result_ids(args2)
--- a/app/classes/query/modules/high_level_queries.rb
+++ b/app/classes/query/modules/high_level_queries.rb
@@ -110,13 +110,15 @@ module Query::Modules::HighLevelQueries
 
   # Make sure we requery if we change the letter field.
   def need_letters=(letters)
-    if !letters.is_a?(String)
+    unless letters.is_a?(String)
       raise("You must pass a SQL expression to 'need_letters'.")
-    elsif need_letters != letters
-      @result_ids = nil
-      @num_results = nil
-      @need_letters = letters
     end
+
+    return if need_letters == letters
+
+    @result_ids = nil
+    @num_results = nil
+    @need_letters = letters
   end
 
   # Returns a subset of the results (as ids).  Optional arguments:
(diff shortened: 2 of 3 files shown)


JoeCohen:
This is low priority. It should probably wait until @pellaea can eyeball it.

pellaea:
You're right, Joe, I don't like this solution.  The underlying problem is the interaction of selecting COUNT(*) from a query which has a GROUP BY clause in it.  Ultimately, this was a bit of sloppiness / laziness on my part.  I knew that it was unnecessary to include the sorting criteria which simply counting the number of results, but I thought (incorrectly) that it would be harmless and therefore I left the code simple.  But of course, I hadn't anticipated a complex case like sorting by number of herbarium_records when I first wrote Query.  So now I should finally deal with it correctly.  I think it will be a simple fix.  Let me play with it for a minute, see if the naive approach works right...  More in a minute...

pellaea:
You're right, Joe, I don't like this solution.  The underlying problem is the interaction of selecting COUNT(*) from a query which has a GROUP BY clause in it.  Ultimately, this was a bit of sloppiness / laziness on my part.  I knew that it was unnecessary to include the sorting criteria which simply counting the number of results, but I thought (incorrectly) that it would be harmless and therefore I left the code simple.  But of course, I hadn't anticipated a complex case like sorting by number of herbarium_records when I first wrote Query.  So now I should finally deal with it correctly.  I think it will be a simple fix.  Let me play with it for a minute, see if the naive approach works right...  More in a minute...

[By the way, something similar could potentially occur when sorting images by name... at least it also introduces a GROUP BY clause, so should therefore also mess up the results of the COUNT(*) query.]

JoeCohen:
>Jason Hollinger <pellaea@gmail.com> Mon, Aug 9, 2021 at 11:56 AM
To: mo-developers <mo-developers@googlegroups.com>, Joseph Cohen <jdcohenesq@gmail.com>
Sorry, Joe, I can't get github to work for me any more today, so I'll just write it in a comment. Please revert your changes to `num_results() `in high_level_queries.rb, and instead use this fix:
```ruby
 # Number of results the query returns.
 def num_results(args = {})
   @num_results ||=
     if @result_ids
       @result_ids.count
     else
       # Explicitly disable GROUP BY and ORDER clauses for the purposes of
       # simply counting the number of results.  This is important because
       # GROUP BY in particular will mess up the COUNT(*) spec.  Passing in
       # empty strings for group and order will tell it to explicitly override
       # anything in self.group and self.order.
       rows = select_rows(args.merge(select: "COUNT(*)", group: "", order: ""))
       begin
         rows[0][0].to_i
       rescue StandardError
         0
       end
     end
 end
```

>I'm still running the tests, but it looks like it does what it is supposed to do. And this fix should guard against similar gotchas in the future (this is the "correct" solution I should have been doing all along)... within the context, of course, of a ginormous complex class like Query which we would never write nowadays!!. And further, it should improve performance: it will no longer unnecessarily sort queries which are simply requesting the number of results.
Uh oh, there is now one failure associated with listing images of an advanced search. I'll look into it. Hopefully this just caught another bug we hadn't noticed and that's all that's going on..

JoeCohen:
>Jason Hollinger <pellaea@gmail.com> Mon, Aug 9, 2021 at 12:22 PM
To: mo-developers <mo-developers@googlegroups.com>, Joseph Cohen <jdcohenesq@gmail.com>
Okay, no worries, the fix just interacted badly with an earlier kludge in the image advanced search query (intended to make advanced image searches with nonempty "content" field not crash the server!) The solution is trivial. Change `execute_content_search()` in app/classes/query/image_advanced_search.rb to:

```ruby
  def execute_content_search(args)
+    # [Sorry, yes, this is a mess.  But I don't expect this type of search
+    # to survive much longer.  Image searches are in desperate need of
+    # critical revision for performance concerns, anyway. -JPH 20210809]
    args2 = args.dup
    args2.delete(:select)
+    args2.delete(:order)
+    args2.delete(:group)
    params2 = params.dup
    params2.delete(:by)
```

app/classes/query/image_advanced_search.rb
```diff
@@ -28,10 +28,11 @@ def handle_content_search!
...
 
   def execute_content_search(args)
-    args2 = args.dup
-    args2.delete(:select)
-    params2 = params.dup
-    params2.delete(:by)
+    # [Sorry, yes, this is a mess. But I don't expect this type of search
+    # to survive much longer. Image searches are in desperate need of
+    # critical revision for performance concerns, anyway. -JPH 20210809]
+    args2 = args.except(:select, :order, :group)
```
pellaea, line 34: Excellent!  I tried to do this but couldn't figure out the syntax.  Thank you.

PR #775, opened 2021-08-03
Herbaria by Code

JoeCohen:
Keep all hits when sorting Herbaria by Code
- Delivers https://www.pivotaltracker.com/story/show/176751952

### Suggested Manual Test and Comparison###
- For comparison purposes, start with the current behavior in the production server
  - https://mushroomobserver.org/herbaria?flavor=nonpersonal
  - click the Code sort tab
**Result** The sort makes some hits disappear (i.e., the non-personal herbaria with a blank Code).
- Now do the same thing in this branch:
  - http://localhost:3000/herbaria?flavor=nonpersonal
  - click the Code sort tab
 **Result** The sort includes all hits from the prior search.

### DIFF
--- a/app/classes/query/modules/ordering.rb
+++ b/app/classes/query/modules/ordering.rb
@@ -148,7 +148,6 @@ module Query::Modules::Ordering
       "collection_numbers.name ASC, collection_numbers.number ASC"
 
     when "code"
-      where << "herbaria.code != ''"
       "herbaria.code ASC"
 
     when "code_then_name"
--- a/test/models/query_test.rb
+++ b/test/models/query_test.rb
@@ -1450,6 +1450,11 @@ class QueryTest < UnitTestCase
     assert_query(expect, :Herbarium, :all, by: :records)
   end
 
+  def test_herbarium_by_code
+    expect = Herbarium.all.sort_by(&:code)
+    assert_query(expect, :Herbarium, :all, by: :code)
+  end
+
   def test_herbarium_in_set
     expect = [
       herbaria(:dick_herbarium),


JoeCohen:
- I know that the current behavior is deliberate, so it's not fair to call it a bug. But it always surprises me when I see it, and there's no obvious way around it.
- If we really want the ability to list only the herbaria with Codes, I suggest: 
  - add a new query variable `code?` Now we can do a Fungarium pattern search for `code?: yes` which returns only the herbaria with `code.present?`

pellaea:
In principle, I have nothing wrong with continuing to include personal herbaria when sorting the index by herbarium code, but I believe the proposed implementation would put those herbaria at the top.  I suggest that this is non-intuitive.  If I do a search for herbaria that happens to return multiple pages of results -- is this an unrealistic use case? what about a listing of all herbaria? -- and want to look for a particular code, so I tell it to sort by code, I would find it obnoxious to have to page down multiple pages to get to herbaria that actually have codes.  Okay, here's a reasonable use case: I know the herbarium code more or less, but it's one of those that totally doesn't resemble the institution name, and I'm misremembering the code slightly so my search for the code failed.  How do I look this herbarium up?  I tried searching, that failed.  I can't search by the institution name, because I have no idea what it is and it doesn't resemble or can't be guessed at from the code.  So I'm forced to look at a listing of all herbaria.  It is probably possible to use some search magic to request only hits that begin with a given letter, or to explicitly ask for herbaria with non-empty codes, but not every user will know how to do that.  I suggest that they would be annoyed with having to wade through code-less herbaria first.  "Why in the world would you want to see herbaria without codes when you've asked it to sort by code??"  (Of course the answer is that another use case would be explicitly to get a listing of herbaria which do not have a code!)  I just think it makes all kinds of sense to put the codeless herbaria at the end.

Phew.  Okay, so I suggest adding a sort term along these lines:

    IF(herbaria.code IS NULL OR herbaria.code = "", 0, 1) DSC

Would that work?  Sort specs may need to be columns these days.  If so, I don't know how to solve this problem off the top of my head.  Maybe you can assign a local variable in the select clause?

app/classes/query/modules/ordering.rb
```diff
@@ -148,7 +148,6 @@ def initialize_order_specs(by)
       "collection_numbers.name ASC, collection_numbers.number ASC"
 
     when "code"
-      where << "herbaria.code != ''"
       "herbaria.code ASC"
```
pellaea, line 151: Proposal:

"IF(COALESCE(herbaria.code, '') = '', 1, 0) ASC, herbaria.code ASC"

Looks like this works as desired.

JoeCohen:
I'll give a try. (I might be able to get to it until tomorrow a.m.)
Note that the herbaria that disappear when sorting by code include herbaria that are institutional, but not registered with Index Herbariorum.

pellaea:
Fair enough, but I still think it makes more sense to sort them at the end.

JoeCohen:
I agree.

PR #776, opened 2021-08-12
Propagate generic classifications

pellaea:
This script finally completes the suite of maintenance tasks to be performed regularly (nightly?) that will keep our database of names, classifications, and mirrored columns in observations table accurate and complete.

Initially I was going to write this as a class method of Name, but it is sufficiently complex and specialized, that I changed my mind and have retained it in scripts/refresh_caches, the only place it will likely ever be used.  Feel free to convince me otherwise.

It basically follows from the argument that our users really shouldn't be forced to fill in the classification strings of taxa below genus, since the parent is clearly present right there in the name(!)  Of course, choosing which particular version of the genus to grab the classification from, especially when dealing with synonymy and multiple versions of genera, can be challenging, so I've written this to be very conservative, and with lots of feedback so we can keep an eye on how many errors it is catching and correcting and exactly what those corrections are.

The process isn't as complicated as I expected, so I think the code is solid.  But the first time it is run it catches *a lot* of errors.  It will be some time before I go through and spot check enough of those errors to be fully confident that I haven't missed any strange pathological cases.

However, once this goes live, it is exciting, because it means the ability to search by higher classification units (e.g., search for everything in Pilocarpaceae) will finally work fully.  (Right now, many members of Pilocarpaceae do not have their classification string filled in at all, and many more are placed in the wrong family, and often there is a great deal of conflict within the species within the same genus.)

This will also make it possible for the new version of the FunDiS report, which contains phylum, class, order and family, to work properly.  Right now, those columns are mostly blank.  (The new FunDiS report is in a separate PR, and I'm having trouble creating it because of internet issues, so don't be concerned if you do not see it.)

And just in general, this will ensure that the mirrored columns observations.text_name, life_form and classification are kept accurate.  Redundant data inevitably drifts, as Nathan has argued many times, and this is no exception.  My hope is that this nightly cronjob will limit and minimize that drift to the point where it is essentially no longer an issue.

### DIFF
--- a/app/classes/query/modules/ordering.rb
+++ b/app/classes/query/modules/ordering.rb
@@ -148,7 +148,6 @@ module Query::Modules::Ordering
       "collection_numbers.name ASC, collection_numbers.number ASC"
 
     when "code"
-      where << "herbaria.code != ''"
       "herbaria.code ASC"
 
     when "code_then_name"
--- a/app/models/name/spelling.rb
+++ b/app/models/name/spelling.rb
@@ -165,4 +165,23 @@ class Name < AbstractModel
   end
 
   private_class_method :guess_pattern
+
+  # This catches cases where correct_spelling_id = id and just clears it.
+  # Not sure why this would ever happen, but empirically there are presently
+  # three cases of it in the database.  Presumably something to do with
+  # name merges?  Whatever.  This fixes it and will run nightly. -JPH 20210812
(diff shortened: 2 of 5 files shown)


pellaea:
Oops, I don't know how that commit from one of Joe's PRs ended up in there.  I'm going to try to delete this PR and create it afresh.

PR #777, opened 2021-08-13
Propagate generic classifications2

pellaea:
This script finally completes the suite of maintenance tasks to be performed regularly (nightly?) that will keep our database of names, classifications, and mirrored columns in observations table accurate and complete.

Initially I was going to write this as a class method of Name, but it is sufficiently complex and specialized, that I changed my mind and have retained it in scripts/refresh_caches, the only place it will likely ever be used.  Feel free to convince me otherwise.

It basically follows from the argument that our users really shouldn't be forced to fill in the classification strings of taxa below genus, since the parent is clearly present right there in the name(!)  Of course, choosing which particular version of the genus to grab the classification from, especially when dealing with synonymy and multiple versions of genera, can be challenging, so I've written this to be very conservative, and with lots of feedback so we can keep an eye on how many errors it is catching and correcting and exactly what those corrections are.

The process isn't as complicated as I expected, so I think the code is solid.  But the first time it is run it catches *a lot* of errors.  It will be some time before I go through and spot check enough of those errors to be fully confident that I haven't missed any strange pathological cases.

However, once this goes live, it is exciting, because it means the ability to search by higher classification units (e.g., search for everything in Pilocarpaceae) will finally work fully.  (Right now, many members of Pilocarpaceae do not have their classification string filled in at all, and many more are placed in the wrong family, and often there is a great deal of conflict within the species within the same genus.)

This will also make it possible for the new version of the FunDiS report, which contains phylum, class, order and family, to work properly.  Right now, those columns are mostly blank.  (The new FunDiS report is in a separate PR, and I'm having trouble creating it because of internet issues, so don't be concerned if you do not see it.)

And just in general, this will ensure that the mirrored columns observations.text_name, life_form and classification are kept accurate.  Redundant data inevitably drifts, as Nathan has argued many times, and this is no exception.  My hope is that this nightly cronjob will limit and minimize that drift to the point where it is essentially no longer an issue.

### DIFF
--- a/app/models/name/propagate_generic_classifications.rb
+++ b/app/models/name/propagate_generic_classifications.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+class Name < AbstractModel
+  class << self
+    # This is used only by script/refresh_caches.  I have placed it here in
+    # order to make it easily accessible to unit testing.  As a separate file,
+    # it should ever be loaded by the web server, so it's safe from causing
+    # even more unnecessary bloat.  I think. -JPH 20210814
+    def propagate_generic_classifications(dry_run: false)
+      fixes = []
+      accepted_names = build_accepted_names_lookup_table
+      classifications = accepted_generic_classification_strings
+      Name.select(:id, :synonym_id, :text_name, :classification).
+        where(rank: 0..(Name.ranks[:Genus] - 1)).
+        each do |name|
+          old_class = old_classification(name)
+          new_class = new_classification(name, accepted_names, classifications)
+          next if old_class == new_class
+
+          fixes << [name, old_class, new_class]
+        end
+      execute_propagation_fixes(fixes, dry_run)
+    end
+
+    private
+
+    def old_classification(name)
+      str = name.classification
+      str.present? && str.strip
+    end
+
+    def new_classification(name, accepted_names, classifications)
+      accepted_text_name = accepted_names[name.synonym_id] || name.text_name
+      genus = accepted_text_name.split.first
+      str = classifications[genus]
+      str.present? && str.strip
+    end
+
+    def build_accepted_names_lookup_table
+      accepted_names = {}
+      Name.select(:synonym_id, :text_name).
+        where(rank: 0..Name.ranks[:Genus], deprecated: false).
+        where.not(synonym_id: nil).
+        each do |name|
+          accepted_names[name.synonym_id] = name.text_name
+        end
+      accepted_names
+    end
+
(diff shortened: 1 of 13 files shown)


JoeCohen (commented):
I continued to be bothered by the possibility that the genus of _Coprinus comatus_ could be _Stereum_.  See `test_genus_of_synonym`.
But you're the one doing the work, so it's your call.
Maybe consider running this by some key users or the Facebook group to see if you get a lot of pushback?

app/models/name/propagate_generic_classifications.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        msgs << "Setting classifications for #{ids.join(",")}"
+        next if dry_run
+
+        # Deliberately skip validations
+        # rubocop:disable Rails/SkipsModelValidations
+        Name.where(id: ids).update_all(classification: classification)
+        # rubocop:enable Rails/SkipsModelValidations
+      end
+      msgs
```
JoeCohen:
```suggestion
      bundles.each_with_object([]) do |bundle, msgs|
        # bundle => [<classification string>, [array of Name ids]]
        ids = bundle[1]
        msgs << "Setting classifications for #{bundle[1].join(",")}"
        next if dry_run

        # Deliberately skip validations
        # rubocop:disable Rails/SkipsModelValidations
        Name.where(id: bundle[1]).
          update_all(classification: bundle[0])
        # rubocop:enable Rails/SkipsModelValidations
```
See above re `each_with_object`. The syntax here is awkward enough so that I added a comment.

app/models/name/propagate_generic_classifications.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    def changes_in_classification_string(old_class, new_class)
+      msgs = []
+      Name.all_ranks.each do |rank|
+        old_name = grab_name_from_classification_string(old_class, rank)
+        new_name = grab_name_from_classification_string(new_class, rank)
+        msgs << "#{old_name} => #{new_name}" if old_name != new_name
+      end
+      msgs.join(", ")
```
JoeCohen, line 120:
```suggestion
      Name.all_ranks.each_with_object([]) do |rank, msgs|
        old_name = grab_name_from_classification_string(old_class, rank)
        new_name = grab_name_from_classification_string(new_class, rank)
        msgs << "#{old_name} => #{new_name}" if old_name != new_name
      end.join(", ")
```
- Ditto re each_with_object
It reads better here than in the other places.

PR #785, opened 2021-09-10
Glossary image fix

pellaea:
The old way glossary terms worked, the first time you went to add an image, it would set the thumbnail but not actually add the image to the glossary term.  This was creating a lot of confusion.  The fix was, well, just to fix that.

It probably wasn't noticed at first because the show page shows the thumbnail separately, then shows the images attached to the term later.  So you would see the first image attached right away in the thumbnail section, even though it wasn't showing it in the full images section.  But ever since I cleared out the corrupted images from the database, there are now a bunch of glossary terms with images but no thumbnail selected.  This shouldn't be a problem, but what happens when someone goes to add an additional image to a glossary term without a thumbnail is that it tries to set the thumbnail but doesn't actually either save the change or add the image to the term, so nothing happened.  (See email complaint from user on 2021-09-10.)

(Actually, come to think of it, I'm not sure how image/reuse_image_for_glossary_term ever worked for the first image added to a glossary term, because of this problem.  Maybe it was actually saving the change to the thumbnail somewhere automagically...  Well, whatever, it now saves it explicitly, and even checks for an error if it fails.)

### DIFF
--- a/test/controllers/image_controller_test.rb
+++ b/test/controllers/image_controller_test.rb
@@ -572,7 +572,7 @@ class ImageControllerTest < FunctionalTestCase
       img_id: image.id.to_s
     }
     login("mary")
-    get_with_dump(:reuse_image_for_glossary_term, params)
+    get(:reuse_image_for_glossary_term, params)
     assert_redirected_to(glossary_term_path(glossary_term.id))
     assert(glossary_term.reload.images.member?(image))
     assert_objs_equal(image, glossary_term.thumb_image)
--- a/test/controllers/image_controller_test.rb
+++ b/test/controllers/image_controller_test.rb
@@ -562,6 +562,22 @@ class ImageControllerTest < FunctionalTestCase
     assert(glossary_term.reload.images.member?(image))
   end
 
+  def test_reuse_image_for_glossary_term_post_without_thumbnail
+    glossary_term = glossary_terms(:convex_glossary_term)
+    image = images(:commercial_inquiry_image)
+    assert_empty(glossary_term.images)
+    assert_nil(glossary_term.thumb_image)
+    params = {
+      id: glossary_term.id.to_s,
+      img_id: image.id.to_s
+    }
+    login("mary")
+    get_with_dump(:reuse_image_for_glossary_term, params)
+    assert_redirected_to(glossary_term_path(glossary_term.id))
+    assert(glossary_term.reload.images.member?(image))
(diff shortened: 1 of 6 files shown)


pellaea:
I think coverage is decreasing infinitesimally because there is now a short clause that occurs when term.add_image or term.save fails.  How does one simulate either of those?  I'm having trouble imagining a case in which they would fail.

JoeCohen:
Jason:
Thanks for jumping on this issue. I probably created the problem I redid
GlossaryTermsController.
The image stuff there always made my head swim.
I'll try to look at the PR later today, but doubt that I can help.

In response to your question, how about stubbing term.add_image and/or
term.save.
See GlossaryTermsControllerTest#test_create_image_save_failure

BUT I suspect that the coverage is decreasing infinitesimally because you
removed some code that was formerly covered.
Coveralls says that the only coverage decreases are in AbstractModel and
GlossaryTerm. But Coveralls isn't flagging any lines in those files as
*newly* uncovered. See https://coveralls.io/builds/42791233.
So maybe just ignore the coverage decrease.
Joe

pellaea:
Ah, okay, yes, I removed a method from abstract model.  Not sure why
glossary term would have decreased, though...

JoeCohen:
GlossaryTerm#add_image is now 2 lines shorter.

JoeCohen:
This looks great. The image stuff now makes sense (as opposed to what I did.)
Here, Coveralls essentially is punishing you for simplifying the code. I sometimes ignore these or find a random untested line anywhere in the codebase that I can easily cover.

test/controllers/image_controller_test.rb
```diff
@@ -562,6 +562,22 @@ def test_reuse_image_for_glossary_term_post
...
+    glossary_term = glossary_terms(:convex_glossary_term)
+    image = images(:commercial_inquiry_image)
+    assert_empty(glossary_term.images)
+    assert_nil(glossary_term.thumb_image)
+    params = {
+      id: glossary_term.id.to_s,
+      img_id: image.id.to_s
+    }
+    login("mary")
+    get_with_dump(:reuse_image_for_glossary_term, params)
```
JoeCohen, line 575:
```suggestion
    get(:reuse_image_for_glossary_term, params)
```
https://groups.google.com/g/mo-developers/c/Qi8N2TVSp6c
https://www.pivotaltracker.com/story/show/173424826

pellaea: Done.

PR #782, opened 2021-09-05
Abbreviate subgenus

JoeCohen:
Abbreviate "subgenus" to "subg."

- Delivers https://www.pivotaltracker.com/story/show/86389186
- Stores and displays subgenera as _Xxx_ subg. _Yyy_.

### Suggested Manual Test ###
* CreateName Aaa subg. Bbb
Expected result: "Successfully created name ‘Aaa subg. Bbb’."
* CreateName Bbb subgenus Ccc
Expected result: "Successfully created name ‘Bbb subg. Ccc’." (saves using the abbreviation)
* Edit any old Name that still uses "subgenus",  Save **without making any changes**.
Expected result: "Successfully updated name ‘_X_. subg. _Y_’." (i.e. automatically updates when the Name is touched)

* Pattern Search for "rank:subgenus"
Expected result: List includes both "subgenus" and  "subg." Names

PR #787, opened 2021-09-13
Subg follow-up

JoeCohen:
Resolves TODO created in #782 in app/classes/rank_matcher.rb
Delivers https://www.pivotaltracker.com/story/show/179551893

- Update regex used to map `text_name` to  the rank Subgenus
- Update test

 After #86389186 (including a manual update to the db),  only "subg." in text_name maps to Subgenus.

PS: I've confirmed that there are no Names in the db matching "sub.":
```ruby
2.6.6 :002 > Name.where("text_name LIKE '%sub.%'").count
 => 0
2.6.6 :003 > Name.where("search_name LIKE '%sub.%'").count
 => 0
2.6.6 :004 > Name.where("display_name LIKE '%sub.%'").count
 => 0
2.6.6 :005 > Name.where("sort_name LIKE '%sub.%'").count
 => 0
```

### DIFF
--- a/app/classes/rank_matcher.rb
+++ b/app/classes/rank_matcher.rb
@@ -32,9 +32,7 @@ TEXT_NAME_MATCHERS = [
   RankMatcher.new(:Stirps,     / stirps /),
   RankMatcher.new(:Subsection, / subsect\. /),
   RankMatcher.new(:Section,    / sect\. /),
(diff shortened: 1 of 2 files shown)


JoeCohen:
I will merge and deploy now; I just noticed this fixes a possible bug.

PR #789, opened 2021-09-15
Api name and location primers

pellaea:


app/controllers/ajax_controller/primers.rb
```diff
@@ -0,0 +1,35 @@
...
+    synonyms = names.to_a.map(&:synonym_id).reject(&:nil?).uniq
+    names |= Name.where(deprecated: false, synonym_id: synonyms).select(*fields)
+    names.map do |name|
+      [name.id, name.text_name, name.author, name.deprecated ? 1 : 0,
+       name.synonym_id]
+    end
+  end
+
+  def location_list
+    Observation.where("location_id IS NOT NULL").
```
JoeCohen, line 30:
```suggestion
    Observation.where.not(location: nil).
```
 A little more Rail-sy and a little shorter

app/controllers/ajax_controller/primers.rb
```diff
@@ -0,0 +1,35 @@
...
+  def location_primer
+    render(json: location_list)
+  end
+
+  private
+
+  def name_list
+    name_ids = Observation.select(:name_id).distinct
+    fields = [:id, :text_name, :author, :deprecated, :synonym_id]
+    names = Name.where(id: name_ids).select(*fields)
```
JoeCohen, line 20: If you feel like it, you could eliminate one call to the db (and one assignment:
```suggestion
    fields = [:id, :text_name, :author, :deprecated, :synonym_id]
    names = Name.where(id: Observation.select(:name_id).distinct).select(*fields)
```

JoeCohen:
Jason: 
Thanks for doing all this!
Sorry to only pick at the edges.  I'm out of time for today, and am traveling Thu & Fri.
Joe

pellaea:
Thanks, Joe!  You read my mind re: where.not(location: nil) -- I figured
you would know the correct syntax better than I do and would fix it for me!
:)

But the other one -- nesting the query within the query -- I will resist.
Just like turning complex run-on sentences into multiple short simple
sentences is almost always better in papers, so, too, simple feels better
in programming.... unlike this sentence!! :)

-Jason

JoeCohen:
I figured you might like the second one. I was on the fence a bit myself,
and thought I would mention it for your consideration.

JoeCohen:
I meant might *not* like the second.

app/controllers/ajax_controller/primers.rb
```diff
@@ -0,0 +1,35 @@
...
+    synonyms = names.to_a.map(&:synonym_id).reject(&:nil?).uniq
+    names |= Name.where(deprecated: false, synonym_id: synonyms).select(*fields)
+    names.map do |name|
```
oliviacpu, line 23: I may not have context for why we're returning arrays instead of objects, but if we omit lines 23-26 the app will have the field names returned with the values, which would be helpful for the mobile app.

`[{"id":139785830,"text_name":"Chlorophyllum rachodes","deprecated":false,"synonym_id":402632615,"author":"(Vittad.) Vellinga"},...]`

I've tried the same for the location list (omitting the `map`) and get a `null` id which is expected for what it looks like that query is doing, but `location_id` and `where` come through fine.

`[{"id":null,"location_id":547147019,"where":"Burbank, California, USA"},...]`

pellaea: We can definitely do that... but two thoughts, and these are just personal opinions:

1. I always cringe to expose internals to the outside like that.  Granted, in practice, it is extremely unlikely that the column names in our tables would ever change.  But it's not impossible.  For example, rails made some change that required that we change the content of the rank column of names, and that would have required the changing of everything downstream.  I could imagine a different version of mysql or if we were to change to postgresql then maybe "true" and "false" might change to "0" and "1" or maybe "null" or who knows.  I realize this mapping requires extra code and processing power, but it allows us to guarantee consumers that this is what the result will look like, period, and will never change.

2. I'm also concerned about size, believe it or not.  We're talking 100s of thousands of names.  Adding ... 52 characters? per line, almost doubling the download size, is not something to be dismissed out of hand.  Of course, if we had nginx configured right (maybe we do?) this is compressed before transmission, and the difference in bandwidth would probably be negligible in practice...

Just to throw it out there, sort of a solution that might make everyone happy: Can I just explicitly encode the output as an array of hashes instead?

    ...map do |obs|
      {id: obs.location_id, name: obs.where}
    end

I think this would age better.  I agree that the map for names would be entirely superfluous at the moment. If our internals change, we can always add a map *then* that would preserve the old key names, etc.

(The reason we aren't dealing with objects is that the instantiation of that many Name and Location objects in ActiveRecord makes this query take way longer.  I actually got impatient and never let the query finish when I tried it.  It may even time out for all I know.  Using selects like this is pretty efficient, at least last I tried it.)

oliviacpu: Thanks for the context! I think a hash is what I meant, in the JSON sense, rather than the Ruby objects? I can work with whatever is available, ultimately. I hadn't considered the size of adding names for values once we were dealing with the production level of names in the database, and that's a fair concern.

pellaea:
Olivia, should we accept and merge this PR the way it is, too?  I've made the changes you requested.  I'm fine with waiting to see how well it works before making a final decision about compressing the response.

oliviacpu:
@pellaea I'm fine with merging it if you are, and we can make a revision if needed!

PR #790, opened 2021-09-17
Add token-based authentication to API

pellaea:
In support of mobile app.

First draft, as discussed in general email, is simply to allow app to pass in the user's password when creating an API key.  If the password is correct, it verifies the API key immediately, and it functions -- so far as I can tell -- identically to a normal token.

We can discuss additional features, such as having API keys automatically expire after a certain amount of time.  And whether we want certain apps to be considered "trusted" and exactly what that means.  (At the moment, I can't think of anything a "trusted" app would be able to do that I'm not comfortable giving every app the ability to do!)

Consider this the opening of a dialog about implementation details...

### DIFF
--- a/README_API_HELP_MESSAGES.txt
+++ b/README_API_HELP_MESSAGES.txt
@@ -1,6 +1,7 @@
 POST api_key
   app:	string (identifier used to help user distinguish which api key belongs to which app)
   for_user:	user (user you are creating api key for)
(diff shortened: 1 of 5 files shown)


oliviacpu (commented):
Just tried this after resetting my dev environment. It is working for me and should fit what we need! Thanks so much

pellaea:
Awesome!  I'll go ahead and merge it, then.

PR #791, opened 2021-09-18
Nontrivial location name change notifications

pellaea:
This adds yet another admin notification, this time for nontrivial changes to the name of locations.  This happens from time to time when a user accidentally uses the wrong location for an observation, and instead of changing the observation to fix it, they edit the location (despite the clear warning on the location form not to do this!).  Hopefully this will allow me to keep a better tabs on that kind of error.

Also, I added links direct to the edit action (in addition to the show action link which is already there) to all of these name and location admin notifications.


JoeCohen:
Looks good. I think you should merge it.

pellaea:
Thanks, Joe.  Done.  Trivial stuff, but will save me a click frequently...

PR #792, opened 2021-09-19
Tweaks to API POST observation

pellaea:
Added ability to set naming reasons when POSTing observations.

Added user's custom notes_template to the output of GET users so that the mobile app can know which custom notes fields to provide.


JoeCohen:
Thanks for doing this. I have no suggestions other than: merge & deploy; we can try it in the mobile app to see if there are any issues.

pellaea:
Thanks, Joe.

PR #795, opened 2021-10-03
Whitelist robots instead of blacklisting

JoeCohen:
Block non-whitelisted robots that obey `robots.txt`.

Delivers https://www.pivotaltracker.com/story/show/179791971
See https://groups.google.com/g/mo-developers/c/qh3r9h5nXoU/m/oEgrj4IjAgAJ.

### DIFF
--- a/script/refresh_sitemap
+++ b/script/refresh_sitemap
@@ -93,30 +93,32 @@ LOCALES = Language.connection.select_values(%(
 ))
 
 ROBOTS_HEADER = <<~"TXT"
+  # Control behavior of crowlers that follow robots.txt rules
   # For documentation, see:
   # https://developers.google.com/search/docs/advanced/robots/robots_txt
   # http://www.robotstxt.org
+
+  # Other ways for MO to block/control robots include:
+  # - Manually block ip address ranges at the firewall
+  # - Automatically block in Rails an anonymous user (including robots)
+  #   that's hogging the webserver. See script/update_ip_stats.rb
+  # - Automatically block in Rails a fake GoogleBot.
+  #    See script/update_googlebots.rb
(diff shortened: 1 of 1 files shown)


script/refresh_sitemap
```diff
@@ -92,31 +92,39 @@ LOCALES = Language.connection.select_values(%(
...
-  User-agent: SpiderLing
-  User-agent: spbot
+  User-agent: AdsBot-Google-Mobile
+  User-agent: AdsBot-Google-Mobile-Apps
   Disallow: /
 
-  # Rules applying to some allowed bots
+  # Whitelisted crawlers
+  User-agent: bingbot
   User-agent: CCBot
```
JoeCohen, line 125: CCBot is combined with other whitelisted bots so that ROBOTS_FOOTER directives will also apply to it.
Is that right?
A side-effect is that there is now a shorter delay for CCBot (15 sec vs 30 sec). I think we can live with that.

script/refresh_sitemap
```diff
@@ -128,7 +136,10 @@ ROBOTS_HEADER = <<~"TXT"
   Disallow: */lookup*user_locale=*
   Disallow: /*?*q= # block urls having the q parameter
   Disallow: /name/show_name
```
JoeCohen, line 138: This was blocked 6 months ago for performance reasons. See c6ee56e
In the long run, it should be unblocked. But I favor keeping it this way for the moment until we block all unregistered users.
What do others think?

JoeCohen:
I ran some manual local tests with the [Google Robots.txt Parser and Matcher Library](https://github.com/google/robotstxt/blob/master/README.md).
**Is there a better way to test this?**

PR #797, opened 2021-10-10
Api post user with nonalphanumerics in email address

pellaea:
Expanded permittivity of regex in email address parser of API.  Pretty trivial PR, although it may be worth commenting that the new regex is still not fully permissive in that it does not bother to permit a bizarre double-quoted form of email address which is supposedly allowed but which I've never seen in practice.  At the same time, it is overly permissive (allows email addresses which start with a dot, for example), but I opted to keep the regex simple over building a perfect but hideously complex regex that would accept all valid email addresses and only valid email addresses.  Honestly, it's just not our job to do so!  The point is to give useful feedback to the app using the API.  Therefore, it accepts the vast majority of email addresses we will ever encounter, but will raise error messages for any of the common typos -- missing "@", strange characters, etc.  And if some perverse case ever comes up, it will be very easy to find a more complicated regex on-line and drop it into place.  (I saw several during my research.)


JoeCohen:
Strange that this is showing only 1 check.

pellaea:
Bizarre!  Search me.  I only work here.

JoeCohen:
> Strange that this is showing only 1 check.

Travis CI is not running. It says:
>Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

It might be time to move to a different CI platform. That won't be trivial. What a pain.

pellaea:
Ah, I see.  My two cents: we can still run the tests just fine
ourselves...  And don't we all do that already anyway??  Having Travis run
them for us was harmless of course, but it never really told me anything I
didn't already know.  Code Climate, however, does provide some value.  At
least, it will catch things that merely running rubocop does not, and
besides it is better about only testing code that has changed.  Travis was
less useful in my opinion.

JoeCohen:
I agree in part about the relative utility of Travis vs CodeClimate. 
But:
- Travis was automatic and saved me when I neglected to run the tests;
- On a few occasions Travis caught a problem which was not caught by local testing;
- I think Travis was the thing that was responsible for running Coveralls.

pellaea:
Ah yes, coveralls was useful, I admit.  And more checks are better than
fewer.

PR #802, opened 2021-10-18
Fix GTM tag to find no-js users

nimmolo:
Was missing an edited param!

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -53,7 +53,7 @@
 <body>
   <% if Rails.env == "production" %>
     <!-- Google Tag Manager (noscript) -->
-    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-PJKJR59"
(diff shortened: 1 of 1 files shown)


nimmolo:
Good question.

OK I have one idea to rule out Safari content blocking:
put a little snippet of JS on the page directly in a script tag, that adds a class to the body.
That way if js is on in the browser, it will definitely apply the class to the element, because it’s not an offsite script.
Or, alternatively, put an element on the page within noscript tags, and give it a class and an ID.

Hopefully GTM can test for the presence of a unique element having a class...
But would this still give false positives to bots?
Agh. More research...

Nimmo

nimmolo:
I just called radware for a quote, I believe i’ll get a call back today.
They offer a free analysis of our server logs and tell us exactly what % is scrapers. Can we provide?

Even if we wind up back where we started with growing traffic, I believe it’s a waste of MO donations to pay for bandwidth for scrapers.
People are most likely donating to make the site available to humans, and hey, more users might mean more donations.

Nimmo

PR #803, opened 2021-10-19
Convert model method User.authenticate to use named params with nil defaults

nimmolo:
Very small tweak, let's see if tests pass


nimmolo:
Here's what's failing minitest:

FAIL["test_blocked_ips", #<Minitest::Reporters::Suite:0x00005635ac91a400 @name="AccountControllerTest">, 35.00275371099997]
 test_blocked_ips#AccountControllerTest (35.00s)
        Expected false to be true.
        test/general_extensions.rb:118:in `assert_true'
        test/controllers/account_controller_test.rb:792:in `test_blocked_ips

PR #806, opened 2021-10-24
Created encrypted credentials and master key

nimmolo:
**Note** -- Along with this ~~draft~~ PR, I've also generated the master.key that unlocks these credentials.
If anyone wants to check out (and update) the credentials in this PR, I’ll send them the master key.

For the APIs:
I couldn’t find our AWS keys - probably a good thing.
I followed `/config/gmaps_api_key.yml-template` for gmaps. I don’t know if those are the actual keys.
Gmaps seems a little trickier, because there are three different production URLs with api_keys (?)

I’m not familiar with where the api ids/keys are used in the site code, but these classes are part of it:
```
classes/geocoder.rb
classes/gmap.rb
classes/image_s3.rb
```
To test if the keys are working, you can change those classes to access the keys like this:

`access_key_id = Rails.application.credentials[Rails.env.to_sym][:aws][:access_key_id]`

For general Rails Credentials reference you can use this command:

`bin/rails credentials:help`

and get the following:

=== Storing Encrypted Credentials in Source Control

The Rails `credentials` commands provide access to encrypted credentials,
so you can safely store access tokens, database passwords, and the like
safely inside the app without relying on a mess of ENVs.

This also allows for atomic deploys: no need to coordinate key changes
to get everything working as the keys are shipped with the code.

=== Setup

Applications after Rails 5.2 automatically have a basic credentials file generated
that just contains the secret_key_base used by MessageVerifiers/MessageEncryptors, like the ones
signing and encrypting cookies.

For applications created prior to Rails 5.2, we'll automatically generate a new
credentials file in `config/credentials.yml.enc` the first time you run `bin/rails credentials:edit`.
If you didn't have a master key saved in `config/master.key`, that'll be created too.

Don't lose this master key! Put it in a password manager your team can access.
Should you lose it no one, including you, will be able to access any encrypted
credentials.

Don't commit the key! Add `config/master.key` to your source control's
ignore file. If you use Git, Rails handles this for you.

Rails also looks for the master key in `ENV["RAILS_MASTER_KEY"]`, if that's easier to manage.

You could prepend that to your server's start command like this:

 RAILS_MASTER_KEY="very-secret-and-secure" server.start

=== Editing Credentials

This will open a temporary file in `$EDITOR` with the decrypted contents to edit
the encrypted credentials.

When the temporary file is next saved the contents are encrypted and written to
`config/credentials.yml.enc` while the file itself is destroyed to prevent credentials
from leaking.


nimmolo:
Revising the description of this PR's scope: 
It will leave **existing** credentials the way they are, and add the new Rails 5.2 credentials as an "**alongside**" option. I'd like to use these for the new GraphQL API, and would like to merge this soonish if possible. Just have to go back through it to be sure i didn't try to replace credential use anywhere (I'm pretty sure i didn't, i just stubbed out where the keys would go).

PR #810, opened 2021-11-06
Initial docker attempt

oliviacpu:
An initial attempt to move to Docker after VirtualBox issues on Mac

Notes in DOCKER_NOTES


oliviacpu:
The `database-config.yml` file is in `.gitignore` currently, but I created one based on the example article. My instinct is that we may be fine adding that to the repository as-is.

Depending on the environment there may be a "correct" way to supply those environment variables to `docker-compose.yml`, as well as creating the Google Maps API config which is needed by the app.

Finally there are some things in the `Vagrantfile` in the `developer-startup` repository like setting the locale and others that I'm not sure how to convert to the `Dockerfile` yet.

nimmolo:
My recent PR moving us to Rails-native “Credentials” and secret storage may be relevant here.

Nimmo

pellaea:
Hmmm, maybe I'm misunderstanding you, but we definitely don't want the
production database configuration public!  That has password and all for
the production database, giving anyone access to anything in the database.

i suspect the way vagrant is doing it is by creating a dev version of
config/database.yml itself.

And Nimmo's recent PR probably is applicable, yet another PR I haven't had
a chance to look into, because I want to have time to really understand it
since it will have such direct effects on production!

oliviacpu:
Closing this as we decided to approach Dockerizing the app from a different direction

PR #812, opened 2021-11-07
GitHub Actions

JoeCohen:
Uses GitActions for Continuous Integration

- Sets up and runs test suite
- Submits test results to Coveralls
- Codeclimate runs independently of GitHub Actions
- I'm still working on having Coveralls send a notification of its results. But that is handled in GitHub rather than in the code.

There is no manual test.
- I suppose one  could push a commit to this branch and confirm that it works.
- You could also check out the branch and confirm that Coveralls still works locally.


.github/workflows/ci_rails.yml
```diff
@@ -0,0 +1,77 @@
...
+
+    # MySQL is installed but does not run by default
+    # https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#mysql
+    - name: Start mySQL
+      run: sudo systemctl start mysql.service
+      
+    - name: Create and configure db
+      run: |
+        mysql -u root -proot < db/initialize.sql
+        cp db/vagrant/database.yml config
```
JoeCohen, line 39: From `script/travis_setup`.  Questions:
- **Is it correct?** (It works, but I don't understand what it does.)
- I got rid of a bunch of conditionals from `script/travis_setup`. **Should I instead retain them?**

Gemfile
```diff
@@ -108,11 +108,12 @@ gem("byebug", group: [:development, :test])
...
+# Use built-in Ruby coverage to generate html coverage file
+gem("simplecov", require: false)
+# generate lcov file to send to Coveralls by Github Actions
+gem("simplecov-lcov", require: false)
 # Automatically track code test coverage
-# Use coveralls_reborn gem instead of coveralls gem
-# With `coveralls` Travis CI runnning with Ubuntu focal gets an SSLError
-# when Travis submits the coverage report to Coveralls
-gem("coveralls_reborn", "~> 0.20.0", require: false)
+gem("coveralls_reborn", "~> 0.23.0", require: false)
```
JoeCohen, line 116: - I don't completely understand how these 3 gems interact. 
- I don't know why I have to explicitly include `simplecov`, but doing so fixed some strange errors. 
See #fd2e4385a4c4a21dae11a04882db41e27698a309

JoeCohen:
It would help me if people would eyeball this quickly, particularly the questions I posted in the comments.

A quick overview of GitHub Actions:
The workflow file is a .yml file that defines a bunch of "steps" taken by GitHub Actions
Each Step has:
- a Name for convenience
- `uses` prebuilt actions (Don't worry about these.) or it
-  `run` shell commands
-  `with` ~ environment

.github/workflows/ci_rails.yml
pellaea: This looks right to me.  It creates and sets up the test database.

re: conditionals -- Is there any chance it will retain the test environment from one build to another?  If not, if it starts from scratch with every single build, then yeah, there's no point in the conditionals.

JoeCohen:
**Question**:
It looks like the two test runs are a push and an integration run. 
> Continuous Integration / test (pull_request) Successful in 4m
Continuous Integration / test (push) Successful in 4m

**Is there an independent way to confirm that?**

pellaea:
Fantastic, Joe!  (I've got no answer for your question, though -- this is
all way above my pay grade. :)

.github/workflows/ci_rails.yml
```diff
@@ -0,0 +1,77 @@
...
+    # check-out repo under $GITHUB_WORKSPACE, so that workflow can access it.
+    # https://github.com/actions/checkout
+    - name: Checkout code
+      uses: actions/checkout@v2  
+
+    # https://github.com/actions/setup-ruby
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        bundler-cache: true # runs bundle install, caches gems
```
JoeCohen, line 26: I just noticed how slick this is. 
The 1st run on any branch takes 7-8 minutes -- about the same as Travis.
But (due to gem caching) subsequent runs are only about 4 minutes -- about 2x faster than Travis.

nimmolo: This seems great! Thanks for doing it Joe.
I'm in the woods today... Can check later

JoeCohen:
Thanks Nimmo!
My goal is to merge this no later than Mon.
(I'll be out of town for 2 weeks starting Tues. with my Monterey laptop. So
I won't have access to the dev environment, assuming that it hasn't been
Dockerized.)
- Joe

nimmolo:
Joe - Following your suggestion in the first comment, I pushed a minor commit to this branch (adding a file to gitignore). Will check back on progress.

nimmolo:
Hi Joe - 
The only thing i noted is, `ci_rails.yml` will be a place where the new Rails 5 credentials can be used. That new credentials system can run alongside the old secrets system, so there is no need to change it over, but maybe let's make a mental note to revisit the config and try to set that up when the credentials PR is merged, it will be a good and simple test case.

nimmolo (approved):
Comments on main thread

JoeCohen:
@nimmolo: Thanks for looking at this and for the improvement.

PR #813, opened 2021-11-13
Remove coveralls_reborn

JoeCohen:
See https://github.com/tagliala/coveralls-ruby-reborn#github-actions
- remove gem
- bundle install
- remove all references to gem

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-require("simplecov")
-require("coveralls")
-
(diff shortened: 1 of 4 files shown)


JoeCohen:
Remove coveralls_reborn

To be doubly safe, I:
- created this branch
- ran it locally, and confirmed that it created `.coverage/index.html`
- pushed it, and confirmed that it created lcov file and submitted it to Coveralls

So now merging it should remove coveralls_reborn in PR #812

PR #817, opened 2021-12-08
Accept DELETE requests for API.

pellaea:
I thought there was more wrong, but it's just a routing issue.


pellaea:
The current solution is actually deprecated.  Hopefully Joe will come to my rescue and tell me how to fix this correctly!

JoeCohen:
A big improvement. 
IMO you should merge it.
And sometime in the future we should look seriously at route pruning for al controllers.

pellaea:
Thanks, Joe.

PR #819, opened 2022-01-16
Use favicon based on new (2021) logo

JoeCohen:
### Manual Test
- Start the rails server in a browser
Expected result: New Icon appears in the tab for localhost::3000. (This happened instantly for Chrome, Firefox. For Safari it took longer and several tries.)


app/views/layouts/application.html.erb
```diff
@@ -17,7 +17,7 @@
     :rss, observer_rss_path, { title: :app_rss.l }
   ) %>
   <title><%= :app_title.l%>: <%= title_tag_contents(controller.action_name)%></title>
-  <link rel="SHORTCUT ICON" href="https://mushroomobserver.org/favicon.ico"/>
+  <link rel="SHORTCUT ICON" href="/favicon.ico?20220116"/>
```
JoeCohen, line 20: * I removed the protocol and domain in order to make this work locally. I think it will also work on the webserver. 
**Is there a reason why it won't work on the webserver?**

* I used an ico instead of the png that Jeff sent me. (I converted Jeff's file to an ico.):
  * I couldn't figure out how to make the png work. (I tried to make appropriate changes from the png example in https://en.wikipedia.org/wiki/Favicon#How_to_use, but couldn't get any browser to show the favicon.)
  * Maybe that's for the best: IE < v11 won't work with a png. 
* The query forces the browser to refresh the favicon. See https://stackoverflow.com/questions/2208933/how-do-i-force-a-favicon-refresh

JoeCohen:
Finally ready for review (after repeated typo fixes).
Please see Question and comments at https://github.com/MushroomObserver/mushroom-observer/pull/819#discussion_r785533924

pellaea:
This looks just right, Joe.  I think it does have to be in *.ico format, for historical reasons no doubt.  The query is the standard technique for handling caches.  (Works for stylesheets and javascript files, too, but fortunately Rails handles that for us!)  And I see no reason removing the domain from the URL shouldn't work.

I just downloaded the new favicon.ico, and Gimp can't open it, complaining about a zero-width icon.  I'm a little concerned about that.  Wow, even ImageMagick doesn't know what to do with it.  ... Yes, and my Firefox cannot display it either.  Would you mind trying a different tool for converting the png to an ico?  Maybe imagemagick, for example?

JoeCohen:
> This looks just right, Joe. I think it does have to be in *.ico format, for historical reasons no doubt. The query is the standard technique for handling caches. (Works for stylesheets and javascript files, too, but fortunately Rails handles that for us!) And I see no reason removing the domain from the URL shouldn't work.
> 
> I just downloaded the new favicon.ico, and Gimp can't open it, complaining about a zero-width icon. I'm a little concerned about that. Wow, even ImageMagick doesn't know what to do with it. ... Yes, and my Firefox cannot display it either. Would you mind trying a different tool for converting the png to an ico? Maybe imagemagick, for example?

Thanks Jason! I'll try a different conversion tool.
I greatly appreciate the feedback because it works fine in my Firefox (96.0.1 (64-bit) on macOS 12.1). Weird.
Hope the snowstorm hasn't/won't hit you.

pellaea:
Oh we did (are getting) hit!  It's gorgeous! :)  I had to de-ice the dish
to get internet back earlier, though...

nimmolo:
Joe - I suggest one of the online favicon generators. They’re great. You actually “need” quite a bit more formats than you think. They can all go in the same directory.

https://realfavicongenerator.net  — This is the one I’ve used the most (mannny times).
https://favicon.io

Nimmo

JoeCohen:
TLDR: I think it's okay now.
Long story: I downloaded and installed Imagemagick and Gimp, got called to dinner, ate, did the dishes and pots. Then finally figured out that the favicon on GitHub was different -- or at least behaved differently from -- the one in my local repo.
Deleted it and pushed the change. 
Used Imagemagick to convert the png to an ico. Added it to my local repo. 
And finally pushed again.
Now the GitHub favicon is/behaves the same as my local version. So I think it's good to go.

(I will also have a look at the generators mentioned by @nimmolo. They will likely come in handy for future changes.)

Issue #830, opened 2022-01-26
Incorrect rss_log messages for Observation create and Name create


JoeCohen:
1. When an Observation is created with an Image, the message is "Image added", but it should be "Observation Created".
2. When a species is created, the message is "Name updated" but it should be "Name Created".

JoeCohen:
This is apparently related to the deletion of a clause in `app/presenters/matrix_box_presenter.rb` that checked to see if the object was just created (thanks @pellaea), and if so, it overrode the rss_log with the "object created" message, without creating something similar in `app/models/rss_log.rb`.

The deleted code `app/presenters/matrix_box_presenter.rb` lines 135-143 is:
```ruby
    elsif !time || time < target.created_at + 1.minute
      self.detail = :rss_created_at.t(type: target_type)
      unless [:observation, :species_list].include?(target_type)
        begin
          self.detail += " ".html_safe + :rss_by.t(user: target.user.legal_name)
        rescue StandardError
          nil
        end
      end
```
I think the equivalent in `app/models/rss_log.rb`, line-385-386 is:
```ruby
    elsif !time
      :rss_created_at.t(type: target_type)
```

nimmolo:
That sounds correct @JoeCohen. So is it that too many messages are getting logged about a single observation with images being created, or is `detail` retrieving the wrong message? 

If it IS considered useful to have each added image logged to the associated obs, maybe the `detail` retrieval method could iterate back through the array of log messages, comparing timestamps with less than a >1minute difference and taking the first one.

I'm imagining for an observation created with images, the log messages would be like this:
image added 12:21:59
image added 12:21:21
image added 12:21:04
image added 12:20:40
observation created 12:20:25

pellaea:
I think neither: I wouldn't say strictly speaking too many msgs are getting
logged.  And RssLog#detail is faithfully retrieving the latest.  I wouldn't
say either behavior is "wrong".  But we can definitely improve the behavior
for specific cases.

I love that you've simplified the code.  Now let's take a minute to
"re-fix" it correctly this time around.  I assume either Nathan or I wrote
the old code in the "whatever works" school of coding.  (Which admittedly
sounds much more like me than Nathan! :)

pellaea:
Good point.  Excellent, in fact.  So why not change it on the other end?
That is, when RssLog actually logs the name, images, etc. after the
observation is created, have it check to see if the observation was just
created -- it definitely has access to that information! --  and silence
[...]

PR #831, opened 2022-01-26
Restore expected rss_log messages without causing n+1 queries in RssLog#detail

nimmolo:
In models/rss_log.rb#detail, restores check: If the most recent message is within one minute of target.created_at time, just print the "Object Created" message.

This PR should fix issue #830

### DIFF
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -379,10 +379,10 @@ class RssLog < AbstractModel
     end
     if !target_type
       :rss_destroyed.t(type: :object)
-    elsif !target ||
+    elsif !target_id ||
           tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
       :rss_destroyed.t(type: target_type)
-    elsif !time
+    elsif !time || time < created_at + 1.minute
       :rss_created_at.t(type: target_type)
     else
       tag.t(args)
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -459,7 +459,7 @@ class RssLog < AbstractModel
     !time || time < created_at + 1.minute
   end
 
-  def creation_message(tag, args, time)
+  def creation_message(tag, args)
     if [:observation, :species_list].include?(target_type)
       :rss_created_at.t(type: target_type) # user would be redundant
     else
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -372,14 +372,14 @@ class RssLog < AbstractModel
 
   # Figure out a message for most recent update.
   def detail
-    parse = parse_log
-    latest_tag, latest_args, latest_time = parse.first
+    log = parse_log
+    latest_tag, latest_args, latest_time = log.first
     if target_simply_destroyed?
       :rss_destroyed.t(type: :object)
     elsif target_combined?(latest_tag)
       :rss_destroyed.t(type: target_type)
     elsif target_recently_created?(latest_time)
-      creation_message(parse.first)
+      creation_message(log)
     else
       latest_tag.t(latest_args)
     end
@@ -459,10 +459,11 @@ class RssLog < AbstractModel
     !time || time < created_at + 1.minute
   end
 
-  def creation_message(tag, args)
+  def creation_message(log)
     if [:observation, :species_list].include?(target_type)
       :rss_created_at.t(type: target_type) # user would be redundant
     else
+      tag, args = log.last
       tag.t(args)
     end
   end
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -378,13 +378,10 @@ class RssLog < AbstractModel
       :rss_destroyed.t(type: :object)
     elsif target_combined?(latest_tag)
       :rss_destroyed.t(type: target_type)
-    elsif !target_recently_created?(latest_time)
-      latest_tag.t(latest_args)
-    elsif [:observation, :species_list].include?(target_type)
-      :rss_created_at.t(type: target_type) # user would be redundant
+    elsif target_recently_created?(latest_time)
+      creation_message(parse.first)
     else
-      first_tag, first_args, first_time = parse.last
-      first_tag.t(first_args)
+      latest_tag.t(latest_args)
     end
   rescue StandardError
     ""
@@ -450,12 +447,6 @@ class RssLog < AbstractModel
 
   private
 
-  def number_of_objects_changed?(tag, time)
-    target_simply_destroyed? ||
-      target_combined?(tag) ||
-      target_recently_created?(time)
-  end
-
   def target_simply_destroyed?
     !target_type
   end
@@ -466,5 +457,13 @@ class RssLog < AbstractModel
 
   def target_recently_created?(time)
     !time || time < created_at + 1.minute
+  end
+
+  def creation_message(tag, args, time)
+    if [:observation, :species_list].include?(target_type)
+      :rss_created_at.t(type: target_type) # user would be redundant
+    else
+      tag.t(args)
+    end
   end
 end
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -372,20 +372,22 @@ class RssLog < AbstractModel
 
   # Figure out a message for most recent update.
   def detail
-    begin
-      tag, args, time = parse_log.first
-    rescue StandardError
-      []
-    end
-    return tag.t(args) unless number_of_objects_changed?(tag, time)
-
+    parse = parse_log
+    latest_tag, latest_args, latest_time = parse.first
     if target_simply_destroyed?
       :rss_destroyed.t(type: :object)
-    elsif target_combined?(tag)
+    elsif target_combined?(latest_tag)
       :rss_destroyed.t(type: target_type)
-    else # it must have been recently created
-      creation_detail(args)
+    elsif !target_recently_created?(latest_time)
+      latest_tag.t(latest_args)
+    elsif [:observation, :species_list].include?(target_type)
+      :rss_created_at.t(type: target_type) # user would be redundant
+    else
+      first_tag, first_args, first_time = parse.last
+      first_tag.t(first_args)
     end
+  rescue StandardError
+    ""
   end
 
   ##############################################################################
@@ -464,14 +466,5 @@ class RssLog < AbstractModel
 
   def target_recently_created?(time)
     !time || time < created_at + 1.minute
-  end
-
-  def creation_detail(args)
-    if [:observation, :species_list].include?(target_type) ||
-       args[:user].blank?
-      :rss_created_at.t(type: target_type) # user would be redundant
-    else
-      "#{:rss_created_at.t(type: target_type)} by #{args[:user]}"
-    end
   end
 end
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -118,6 +118,8 @@
 #  == Attributes
 #
 #  id::                 Locally unique numerical id, starting at 1.
+#  created_at::         Date/time log or object was created.
+#                       See AbstractModel#init_rss_log
 #  updated_at::         Date/time it was last updated.
 #  notes::              Log of changes.
 #  location::           Owning Location (or nil).
@@ -149,8 +151,6 @@
 #
 #  None.
 #
-################################################################################
-
 class RssLog < AbstractModel
   belongs_to :location
   belongs_to :name
@@ -315,7 +315,7 @@ class RssLog < AbstractModel
                           relevant_args(args),
                           args[:time] || Time.zone.now)
     RssLog.record_timestamps = false if args.key?(:touch) && !args[:touch]
-    self.notes = entry + "\n" + notes.to_s
+    self.notes = "#{entry}\n#{notes}"
     # self.updated_at = args[:time] if args[:touch]
     save_without_our_callbacks unless args.key?(:save) && !args[:save]
     RssLog.record_timestamps = true
@@ -340,7 +340,7 @@ class RssLog < AbstractModel
   def orphan(title, key, args = {})
     args = args.merge(save: false)
     add_with_date(key, args)
-    self.notes = RssLog.escape(title) + "\n" + notes.to_s
+    self.notes = "#{RssLog.escape(title)}\n#{notes}"
     save_without_our_callbacks
   end
 
@@ -354,7 +354,7 @@ class RssLog < AbstractModel
   def parse_log(cutoff_time = nil)
     first = true
     results = []
-    for line in notes.to_s.split("\n")
+    notes.to_s.split("\n").each do |line|
       if first && !line.match(/^\d{14}/)
         tag  = :log_orphan
         args = { title: self.class.unescape(line) }
@@ -377,15 +377,14 @@ class RssLog < AbstractModel
     rescue StandardError
       []
     end
-    if !target_type
+    return tag.t(args) unless number_of_objects_changed?(tag, time)
+
+    if target_simply_destroyed?
       :rss_destroyed.t(type: :object)
-    elsif !target_id ||
-          tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
+    elsif target_combined?(tag)
       :rss_destroyed.t(type: target_type)
-    elsif !time || time < created_at + 1.minute
-      :rss_created_at.t(type: target_type)
-    else
-      tag.t(args)
+    else # it must have been recently created
+      creation_detail(args)
     end
   end
 
@@ -443,5 +442,36 @@ class RssLog < AbstractModel
   # Reverse protection of special characters in string for log encoder/decoder.
   def self.unescape(str)
     str.to_s.gsub(/%(..)/) { Regexp.last_match(1).hex.chr }
+  end
+
+  #############################################################################
+
+  private
+
+  def number_of_objects_changed?(tag, time)
+    target_simply_destroyed? ||
+      target_combined?(tag) ||
+      target_recently_created?(time)
+  end
+
+  def target_simply_destroyed?
+    !target_type
+  end
+
+  def target_combined?(tag)
+    !target_id || tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
+  end
+
+  def target_recently_created?(time)
+    !time || time < created_at + 1.minute
+  end
+
+  def creation_detail(args)
+    if [:observation, :species_list].include?(target_type) ||
+       args[:user].blank?
+      :rss_created_at.t(type: target_type) # user would be redundant
+    else
+      "#{:rss_created_at.t(type: target_type)} by #{args[:user]}"
+    end
   end
 end
--- a/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
+++ b/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
@@ -1,5 +1,5 @@
 class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
   def change
-    add_column(:rss_logs, :created_at, :datetime, null: false)
+    add_column(:rss_logs, :created_at, :datetime)
   end
 end
--- a/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
+++ b/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
@@ -1,0 +1,11 @@
+class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
+
+  def up
+    add_column :rss_logs, :created_at, :datetime
+  end
+
+  def down
+    remove_column :rss_logs, :created_at
+  end
+
+end
--- a/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
+++ b/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
@@ -1,11 +1,5 @@
 class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
-
-  def up
-    add_column :rss_logs, :created_at, :datetime
+  def change
+    add_column(:rss_logs, :created_at, :datetime, null: false)
   end
-
-  def down
-    remove_column :rss_logs, :created_at
-  end
-
 end
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -762,6 +762,7 @@ class AbstractModel < ApplicationRecord
       result = rss_log
     else
       rss_log = RssLog.new
+      rss_log.created_at = created_at unless new_record?
       # Don't attach to object if about to destroy.
       if !orphan
         rss_log.send("#{type_tag}_id=", id) if id
--- a/db/migrate/20220128202651_change_column_null_rss_logs_created_at.rb
+++ b/db/migrate/20220128202651_change_column_null_rss_logs_created_at.rb
@@ -1,0 +1,5 @@
+class ChangeColumnNullRssLogsCreatedAt < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null(:rss_logs, :created_at, false)
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_04_17_194852) do
+ActiveRecord::Schema.define(version: 2022_01_26_211212) do
 
   create_table "api_keys", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
     t.datetime "created_at"
@@ -565,6 +565,7 @@ ActiveRecord::Schema.define(version: 2021_04_17_194852) do
     t.integer "project_id"
     t.integer "glossary_term_id"
     t.integer "article_id"
+    t.datetime "created_at"
   end
 
   create_table "sequences", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=latin1", force: :cascade do |t|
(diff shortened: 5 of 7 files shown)


app/models/rss_log.rb
```diff
@@ -382,7 +382,7 @@ def detail
     elsif !target ||
           tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
       :rss_destroyed.t(type: target_type)
-    elsif !time
+    elsif !time || time < target.created_at + 1.minute
```
pellaea, line 385: We definitely want to check the logs for the activity log to make sure this isn't causing N extra queries.  (Anything else use RssLog#detail we care about?)

nimmolo: I believe only the matrix box presenter calls RssLog#detail.

From inspecting my GraphQL/GraphiQL queries, I realized even line 382 above (`elsif !target`) queries the target. Neither seem to affect the presenter though, due to eager loading, and this additional check in line 385 doesn't seem to cause any further queries.

It seems the eager loading needs to happen via the use of some gems in my GraphQL setup. (It is recursively re-querying the same observations when it goes to get their rss_log.detail!)

pellaea:
Okay, cool.  I thought that would be the only real customer of this
method.  You should be able to check target_id instead of target.  But if
the eager-loading is working right (I'm suspicious of this), then I have no
problem with it.  Check especially observation indexes.  In order for this
to work, it would have to load the observations then eager load rss_logs
then eager load observations again!  That seems really unnecessary.  Maybe
your GraphQL handles this okay somehow (again, though, I can't imagine how,
because you've got to start with observations, then just tracing the way
rails works, it would have to go observation to rss log then back to the
observation, and the way that works in rails when you go obs.rss_log.target
it will result in a different instance of the obs, not the original
instance, this means at very least eager loading the observations twice),
but I'll guarantee our code does not handle it right.

nimmolo:
OK - I'd like to tinker with it a bit more then, to see if I can get the rss_log to only pull from its own properties for the detail.

It's certainly not impossible.

pellaea:
I am serious about considering adding "created_at" column to the table.  It
doesn't matter if it's populated correctly for existing objects, just new
ones.  It would be very little extra space.  And it would make this problem
very easy to solve.

Alternatively, a single new method which just parses out the very last
(chronologically first) date in the log would probably also work.  Just a
little less elegant to my eye.

nimmolo:
Just saw this. Agree about making a created_at column! 

I feel like it would be desirable to try to accurately populate it by grabbing the last/first date, just so we have a way to inspect old logs in the db by created_at, in case we ever wanted to.

pellaea:
If you have the time and patience to do so, aboslutely!  But fair warning,
it will definitely make the migration (relatively speaking) nontrivial.
But maybe worthwhile in the long term(?)

Or... Leave the migration trivial (populate it with null as a first pass).
Then I would be happy to populate it by hand myself using vi macros and
duct-tape and baling wire.  Save you a big hassle.  No real compelling
reason to test or "institutionalize" the population of this column as
official code in a migration.  I could hammer that out in 5 minutes, no
prob.

nimmolo:
Ah, riiiight it’s a rails g migration.

It seems like i could write the code to do a first pass at populating the column, within the migration class, no?

I could just write it to pop the last/first log's date, and if that comes up empty then it’s null.

Nimmo

nimmolo:
Considering this advice:

https://jakeyesbeck.com/2021/04/10/avoid-models-in-migrations/

I’m going with “Use temporary models”, seems more workable.
I'll try to cook up a migration that actually pulls the date, and put it in the draft PR for you to evaluate.

Nimmo

pellaea:
If that sounds fun, go right ahead.  Again, doing it by hand outside the
code base sounds *a lot* easier to me, personally, so I'll keep the offer
on the table.  Beware that such procedures can take a long time to run.
And note that the live server needs to be effectively down while it runs on
the live server.  I can create a temporary table by hand then the update
happens lickety-split.  If I do it by hand.  Just something to consider.

-Jason

nimmolo:
Yep it’s difficult, and your considerations about tying up the live server seem more important.
(I ran into trouble handling nil time in the log notes.)

Ok, so seems like the migration (and column creation, and likely population) must happen prior to trying to call “created_at” in the model.

Should the migration be a separate PR from this one?

Nimmo

pellaea:
I would do it all at once myself.

nimmolo:
It's failing on ABC for `detail`, but iirc we can disregard (unless anyone can think of a way to reduce the complexity). Should I maybe mark it "wontfix" in CodeClimate?

pellaea:
It definitely _can_ be fixed.  But I would personally recommend maintaining
focus on the original goal... which I've completely forgotten at this
point!  But it had nothing to do with rss logs. :)  Let's just say, you
didn't _create_ the complexity.  In fact you improved it.  It's just not
perfect yet.  Dude, that's a win in my book.

nimmolo:
Test and fixture are updated to work the new details method case when the log_update is <1m after the created_at time.

I believe the PR is ready for review.

app/models/rss_log.rb
```diff
@@ -379,10 +379,10 @@ def detail
     end
     if !target_type
       :rss_destroyed.t(type: :object)
-    elsif !target ||
+    elsif !target_id ||
           tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
       :rss_destroyed.t(type: target_type)
-    elsif !time
+    elsif !time || time < created_at + 1.minute
```
pellaea, line 385: Arg, I just realized that this is not quite right.  It used to be (equivalent to) time < target.created_at.  But note that names are auto-created without a log, and their log doesn't get created until the first time they are modified.  The result is, for example, when Joe just updated Macrolepiota procerus your present code will claim that he just *created* it.  When in fact it is name #3702, so it was created way back in 2007.

Consider the following quick fix.  When initializing the log (line 765 of app/models/abstract_model.rb) have it set rss_log.created_at explicitly if the target isn't new:

    rss_log.created_at = created_at unless new_record?

This will ensure that rss_log.created_at really does reflect when the target was created (as your code expects).  Perhaps this is misleading, because the created_at column generally would refer to the time the rss_log was created, but in this case it will instead refer to the time when the *target* was created, not when the rss_log was created.  But I think we can make an exception in this case, no.  What do you think?

nimmolo: I like it 👍

db/migrate/20220126211212_add_created_at_to_rss_logs.rb
```diff
@@ -0,0 +1,11 @@
+class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
+
+  def up
+    add_column :rss_logs, :created_at, :datetime
+  end
+
+  def down
+    remove_column :rss_logs, :created_at
+  end
```
JoeCohen, line 9:
```suggestion
  def change
    add_column(:names, :created_at, :datetime)
  end
```

JoeCohen:
In app/models/rss_log.rb, change lines 33-40 to:
```ruby
#         def change
#           add_column(:rss_logs, :model_id, :integer)
#           add_column(:models, :rss_log_id, :integer)
#         end
```
Sorry I don't know how to do this as a GitHub suggestion

db/migrate/20220126211212_add_created_at_to_rss_logs.rb
JoeCohen, line 9:
```suggestion
  def change
    add_column(:rss_logs, :created_at, :datetime)
  end
```

JoeCohen:
In app/models/rss_log.rb, around line 121, consider adding something like:
```ruby
#  created_at::         Date/time log or object was created. 
#                       See AbstractModel#init_rss_log
```

JoeCohen:
What the best way to propose these changes so that you can easily accept or reject them? 
(It seems like I cannot make a "suggestion" in areas of a file that are not changed in the PR.)

Some easy rubocop changes. 
Line #318 – Convention: Style/StringConcatenation: Prefer string interpolation to string concatenation.
    self.notes = entry + "\n" + notes.to_s
```ruby
    self.notes = "#{entry}\n#{notes.to_s}"
```
Line #343 – Convention: Style/StringConcatenation: Prefer string interpolation to string concatenation.
    self.notes = RssLog.escape(title) + "\n" + notes.to_s
```ruby
    self.notes = "#{RssLog.escape(title) }\n#{notes.to_s}"
```

JoeCohen:
In app/models/rss_log.rb, how about something like this: (complete file attached in next comment).
- The guard clause fixes the ABC issue. 
- The rest helps me understand it better. 
- `private` just clarifies which methods are not part of class's public interface.

```ruby
  def detail
    begin
      tag, args, time = parse_log.first
    rescue StandardError
      []
    end
    return tag.t(args) unless number_of_objects_changed?(tag, time)

    if target_simply_destroyed?
      :rss_destroyed.t(type: :object)
    elsif target_combined?(tag)
      :rss_destroyed.t(type: target_type)
    else # it must have been recently created
      :rss_created_at.t(type: target_type)
    end
  end


 ##############################################################################

  private
```

```ruby
  def number_of_objects_changed?(tag, time)
    target_simply_destroyed? ||
      target_combined?(tag) ||
      target_recently_created?(time)
  end

  def target_simply_destroyed?
    !target_type
  end

  def target_combined?(tag)
    !target_id || tag.to_s.match?(/^log_#{target_type}_(merged|destroyed)/)
  end

  def target_recently_created?(time)
    !time || time < created_at + 1.minute
  end
```

JoeCohen:
Thanks @nimmolo for doing all the hard work on this. I'm just picking around the edges. Please feel free to ignore any/all of my comments.

nimmolo:
@JoeCohen These are great, thanks. 

One thing i'd add is that the new column`created_at` should be declared non-nullable in the migration, with `add_column(:rss_logs, :created_at, :datetime, null: false)` but I couldn't make the migration work adding that at the outset - the fields were there, and they were null. I believe @pellaea has filled in `created_at` in all rss_logs now, though. He uploaded a new checkpoint_stripped.tar.gz to google drive last night that has them.

I would be **very** happy for you to commit those changes on this branch, if you like — i believe that means they'll be included in the PR. Thanks for fixing the ABC!

(I'm about to go to an orthopedic appointment so i'll be out for awhile. Otherwise i can do them when i get back.)

JoeCohen:
I'm out 9-early afternoon PST.

JoeCohen:
I'm trying this with the new @pellaea's db snapshot.
I noticed that the message no longer says who created the object. 
That's fine for Observations and Species Lists, where that's redundant. But we should include the creator for other objects.
I'll take a crack at it. 
`matrix_box_presenter.rb` used to handle this with:
```ruby
      self.detail = :rss_created_at.t(type: target_type)
      unless [:observation, :species_list].include?(target_type)
        begin
          self.detail += " ".html_safe + :rss_by.t(user: target.user.legal_name)
        rescue StandardError
          nil
        end
      end
```

pellaea:
Joe, the challenge is to make the rss_log stand alone, and to get
everything it needs to build its detail message *without* looking at the
target.

That's the whole reason behind either adding RssLog#created_at column (as
we did) or to give RssLog a way to parse it out of the log (not possible in
the case of names).

In the case of observations that were just created, maybe we can grab that
information from the last entry in the log?

I'm feeling like we've missed the boat on this one.  It is increasingly
looking like the "correct" behavior is to "group" log messages and show the
first log message of the latest grouping.  In the case of creating an
observation with secondary flurry of image_created and name_proposed and
votes and all, the first message would presumably be "observation
created".  Let's just make sure the log message includes the information we
would want to see on the Activity Log page.  Then by design we would never
need to access the target in order to display a log message.

Maybe there are additional cases where an action (voting on a name?
proposing a new name?) causes secondary actions (consensus changed).  Do we
want to treat these all the same way -- that is, log the first of the
related group of actions, not the last?

Example log from most recently created observation:

20220127225034 log_image_created name Bild%20#1438038 user Atlides
20220127225033 log_image_created name Bild%20#1438037 user Atlides
20220127225032 log_consensus_changed new
**__Stereum__**%20Hill%20ex%20Pers. old **__Fungi__**%20Bartl. user Atlides
20220127225029 log_observation_created user Atlides

Since all of these actions are within a few seconds of each other (< 1
minute), we group them and display the first: "observation created by
Atlides".

-Jason

nimmolo:
@JoeCohen Here's what i had started stubbing out in `RssLogs`, to find the first message in a "group". It doesn't work yet!

```ruby
  # Creating an observation can generate a flurry of log messages 
  # (names, images, sequences, etc). The first one in the flurry
  # is the most relevant (eg, "Observation Created") 
  # Note: messages are in reverse chronological order, so we're going back
  def relevant_entry
    if parse_log.length < 2 
      parse_log.first
    else
      # NOTE: This is where i don't know what i'm doing in Ruby
      parse_log.each_with_index {|e, i| 
        # if !parse_log[i+1].exists || parse_log[i][time] - parse_log[i+1][time] > 1.minute or no e+1,
        #   parse_log[i] 
        # end
      }
  end
```

JoeCohen:
I'll push something up in a second that just restores the old behavior without calling `target.user`.

nimmolo:
Oh, right - the `args`!  
Looks pretty sharp to me.

JoeCohen:
I'm content with just restoring the old behavior (without causing n+1 problems).
But if someone else wanted to run with [@pellaea's idea](https://github.com/MushroomObserver/mushroom-observer/pull/831#issuecomment-1023738970), I wouldn't object.

app/models/rss_log.rb
```diff
@@ -444,4 +443,38 @@ def self.escape(str)
...
+  def target_recently_created?(time)
+    !time || time < created_at + 1.minute
+  end
+
+  def creation_detail(args)
+    if [:observation, :species_list].include?(target_type)
+      :rss_created_at.t(type: target_type) # user would be redundant
+    else
+      begin
+        "#{:rss_created_at.t(type: target_type)} by #{args[:user]}"
```
pellaea: Can we really expect args[:user] to be the object's creator?  I suppose it usually will be true.  But is this reliable?  Actually the more likely case would be args[:user] not being set at all for whatever the latest log entry is.  Maybe I'm creating bugs where bugs don't exist...

Dude!  Just got a flash of inspiration.  We are already parsing the whole log!  But we're throwing away everything but the latest entry.  Then we discover that we actually want the last entry, but instead of just keeping the entire parsed log, we choose to kludge the hell out of it.  This makes no sense.  How about this instead:

  def detail
    parse = parse_log
    latest_tag, latest_args, latest_time = parse.first
    if !target_type
      :rss_destroyed.t(type: :object)
    elsif target_combined?(latest_tag)
      :rss_destroyed.t(type: target_type)
    elsif target_created_recently?(latest_time)
      log_message(*parse.last)
    else
      log_message(*parse.first)
    end
  rescue StandardError
    ""
  end

  def log_message(tag, args, time)
    tag.t(args)
  end

JoeCohen:
Thanks @pellaea. 
I'll give it a try later tonite (maybe in a couple of hours) unless @nimmolo wants to work on it first.

pellaea:
I can try it out myself.  Just a minute (or 50)...

JoeCohen:
I'm fooling around with it. It looks good, except user appears 2x in
Observation and Species List Created. Not the end of the world, but nice to
have.
Also:
Change the name/call of target_created_recently? to be consistent. (I like
your name, target_created_recently?, better than mine,
target_recently_created?)
Rubocop issues
Use underscore to indicate variable is unused:
latest_tag, _latest_args, latest_time = parse.first
...
def log_message(tag, args, _time)

pellaea:
I was sloppy in my suggestions because I wasn't looking at the latest
code.  I'm testing a cleaner version now...  It passes tests.  How do I
contribute this to the PR?  Feel free to reject it.

JoeCohen:
Just push it to origin.
I'm now really away for 2 hours.

pellaea:
I've pushed it to a branch in my repo called the same thing
(nimmo-restore-log-created-messages).  Can I push directly to nimmo's repo,
is that how this should work??

nimmolo:
Yes just commit directly to this branch

nimmolo:
You guys!
I shudder to think what you'll find if you check out my Svelte code. 
But... I'm looking forward to it!

I guess it's safe for now due to Javascript aversion...

pellaea:
Oops!  I think I see -- it's only origin/master we cannot push to?

See what you think of the new version.  I think it simplifies the code a
bit.

nimmolo:
Correct, it's only origin/master we can't push to. 
In other words this is "my" branch in name only, it's really just a working branch for the PR.

pellaea:
Oops, I forgot to clean up a few things.  This should be better.

JoeCohen:
That's a nice simplification.
It looks good to me, **except for my screwup in `db/migrate/20220126211212_add_created_at_to_rss_logs.rb`**.
I'm too tired now to fix that. I'm off to bed.
But if you want to fix it, merge it, and deploy it, that's fine. (It is - after all -- a bug fix.)
Or I can try to do that tomorrow a.m., and maybe add some tests to avoid the epsilon drop in coverge.
Ironically, I think the coverage may have decreased because of the code simplifications removed previously covered lines.

pellaea:
Thanks, Joe.  Actually I feel like this is pretty poorly covered right now,
and it really should be covered better.  But I really don't have time,
either.  Can we classify this as an improvement and call it a day?

(I have no idea what you're talking about concerning the migration!)

JoeCohen:
Migration: I added `null: false` to `db/migrate/20220126211212_add_created_at_to_rss_logs.rb`. Now the migration is out of sync with the schema. So just delete `, null: false`. It can be added as a separate migration as part of this PR or in a separate PR.
I agree about the testing and calling it a day. 
I'm too tired to do anything more today; I'm afraid I'll screw it up.

pellaea:
Ah!  Okay, tomorrow we can fix these things.

JoeCohen:
Question re `null: false`
I'm concerned that this was already done directly in the db, with the result that the Rails schema is different from the db schema in this regard. I don't have time to check it out now. (I have some errands to run for about 1 hour, then a ISP tech is coming to work on our DSL.)

pellaea:
Maybe the thing to do, just to be certain, is to do another migration in
which we create a temporary column to hold created_at's current value, then
delete created_at and re-created it from scratch correctly, then restore it
from the temporary column and delete the temporary column.  Does that make
any sense??

I'm struggling to find time, too.  But I'll try to get to this in a few
hours.

JoeCohen:
I think the first thing to do is check whether the schemas are really out of sync.
I can't do that right now because I blew up the dev environment on my desktop and I'm having issues rebuilding. 
(I can still use my laptop.)
I will try to check this in a few minutes.

JoeCohen:
I'm dead in the water on the dev environment on both machines. Will try to fix that.

nimmolo:
@JoeCohen Hold on — if you edited the migration class (to add `null: false`) and also manually edited schema.rb, and then removed both those edits but **did not** re-run the migration, I believe the should still be in sync with the schema. 

It appears that the schema gets updated with the timestamp of the migration only when it's run. In other words, at the top of schema.rb,
```ruby
ActiveRecord::Schema.define(version: 2022_01_26_211212) do
```
The version matches the filename of the most recent migration,
```
20220126211212_add_created_at_to_rss_logs.rb
```

When that migration was first run, there was no null: false, so removing it in both files, we should be good. 

In other words I strongly suspect they are presently in sync. If anyone has a second to educate me on the dangers of them being out of sync, i'm all ears - I don't understand what's at stake here. 

For the non-nullable, you could just make a separate PR with that tiny "migration" that adds `change_column_null`.

JoeCohen:
I didn't manually edit the Rails schema.

nimmolo:
Sounds even better. 
(I did at first, locally! Then had to revert the change of course)

JoeCohen:
@nimmo: Since I'm dead in the water right now, could you use the nifty new
rails_db gem to check MySQL's schema for rss_logs (in the new snapshot)?
In particular, what does it say about created_at?
Does it look like this:
[image: Screen Shot 2022-01-28 at 6.27.06 AM.png]

nimmolo:
Will do. The screen shot didn't come through... try emailing me directly?

nimmolo:
Good idea. Here's what i've got directly from the rails db schema, new snapshot

```
name       | sql_type | null | limit | precision | scale | type
created_at | datetime | true |       | 0         |       | datetime
```


<br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">

JoeCohen:
Thanks!
Based on that, the Rails schema and the MySQL schema *are* in sync: both
allow created_at to be nil.
So I think all that needs to be done is

   - delete ", null: false" from
   db/migrate/20220126211212_add_created_at_to_rss_logs.rb
   - create a new migration (change_column_null(:rss_logs, :created_at,
   false))
   - commit
   - rake db:migrate

I suppose to be absolutely certain I should check the live db first. It
will take me a while to figure out how to do that.

nimmolo:
New migration added, migrated and committed

JoeCohen:
I think that fixes the current state. (I can't figure out how to get the
MySQL schema on the live db; it won't let me sudo because it doesn't
recognize my pw. But the snapshot should be the same as the live db.)

Nimmo:
I'm sorry, I think you've contacted the ring person here, I don't think I
recognise this project.

From looking back thorough the messages you were talking to an Andrew Nimmo
rather than myself.

JoeCohen:
Thanks for letting me know about the wrong contact.

JoeCohen:
I've seen some new issues in the rss log messages. I will try to dig into
them this morning.
Sample screenshots attached. The first might be an issue in the db rather
than the code.
[image: Screen Shot 2022-01-28 at 6.59.52 PM.png]
The second appears to be a consistent issue with deletion of Observations.
Probably code and I hope easy to fix.
[image: Screen Shot 2022-01-29 at 6.08.15 AM.png]

nimmolo:
@JoeCohen - I'm not getting the attached screenshots in any comments submitted via email, either reading them on email or in GitHub.

JoeCohen:
Screenshots that didn't upload properly in https://github.com/MushroomObserver/mushroom-observer/pull/831#issuecomment-1024943092:
Screen Shot 2022-01-28 at 6 59 52 PM

Screen Shot 2022-01-29 at 6 08 15 AM

pellaea:
Great, thanks, Joe.

I'm guessing the "Lichen Lichen" thing is a destroyed object.  In that case
the log is correct, so we need to fix how we render it.

In the Noah Siegel observation there is no "observation_created" log, so
the created_at date is wrong.  This is something I plan to fix in the
database once we've settled the migration issue -- is that settled yet, by
the way?  I wasn't able to understand the string of emails associated with
it.  Anyway, this is not the code, this is the data, and I will fix it.
(That is, I will go back and fix all the created_at dates to be the date
the *target* was created, not the when the first activity was logged for
it.)

-Jason

pellaea:
To test my proposed fix to the database, run this on the dev VM:

update rss_logs r join names          x on r.name_id          = x.id set
r.created_at = x.created_at where r.name_id          is not null and
r.created_at != x.created_at;
update rss_logs r join observations   x on r.observation_id   = x.id set
r.created_at = x.created_at where r.observation_id   is not null and
r.created_at != x.created_at;
update rss_logs r join species_lists  x on r.species_list_id  = x.id set
r.created_at = x.created_at where r.species_list_id  is not null and
r.created_at != x.created_at;
update rss_logs r join locations      x on r.location_id      = x.id set
r.created_at = x.created_at where r.location_id      is not null and
r.created_at != x.created_at;
update rss_logs r join projects       x on r.project_id       = x.id set
r.created_at = x.created_at where r.project_id       is not null and
r.created_at != x.created_at;
update rss_logs r join glossary_terms x on r.glossary_term_id = x.id set
r.created_at = x.created_at where r.glossary_term_id is not null and
r.created_at != x.created_at;
update rss_logs r join articles       x on r.article_id       = x.id set
r.created_at = x.created_at where r.article_id       is not null and
r.created_at != x.created_at;

JoeCohen:
- Thanks Jason.
- The Lichen/Lichen thing (and other similar ones) are indeed destructions.
- The migration thing should be fixed. But could you triple that in the live db by checking whether the schema reflects that the the `rss_logs created_at` column has `NULL false`? (I tried to check it, but I can't sudo in the db server because I have the wrong pw or something.)
- I'll try your script tomorrow. Too many other things today.

pellaea:
I confirm that the live DB does not allow nulls in rss_logs.created_at:

mysql> show columns from rss_logs;
+------------------+----------+------+-----+---------+----------------+
| Field            | Type     | Null | Key | Default | Extra          |
+------------------+----------+------+-----+---------+----------------+
| id               | int      | NO   | PRI | NULL    | auto_increment |
| observation_id   | int      | YES  |     | NULL    |                |
| species_list_id  | int      | YES  |     | NULL    |                |
| updated_at       | datetime | YES  |     | NULL    |                |
| notes            | text     | YES  |     | NULL    |                |
| name_id          | int      | YES  |     | NULL    |                |
| location_id      | int      | YES  |     | NULL    |                |
| project_id       | int      | YES  |     | NULL    |                |
| glossary_term_id | int      | YES  |     | NULL    |                |
| article_id       | int      | YES  |     | NULL    |                |
| created_at       | datetime | NO   |     | NULL    |                |
+------------------+----------+------+-----+---------+----------------+

pellaea:
PS. Joe, when you log in to the production server and change over to user
"mo", there should be an alias:

mo> mosql
# connects to production mysql server in interactive mode
mo> mosqle blah.sql
# runs a sql script

I never remember the password either -- it's huge and ugly!

JoeCohen:
Thanks. I didn't know about connecting via the web server. Very useful!

PR #832, opened 2022-01-26
Create 20220126211212_add_created_at_to_rss_logs.rb

nimmolo:
Note: This migration doesn't seem to work yet. When i run `rake db:migrate` for the simple migration i get

Mysql2::Error: Duplicate column name 'created_at': ALTER TABLE `rss_logs` ADD `created_at` datetime

Here’s my migration:
```ruby
class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]

  def up
    add_column :rss_logs, :created_at, :datetime
    change_column_null :rss_logs, :created_at, false
  end

  def down
    remove_column :rss_logs, :created_at
  end

end
```

### DIFF
--- a/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
+++ b/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
@@ -1,0 +1,12 @@
+class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
+
+  def up
+    add_column :rss_logs, :created_at, :datetime
+    change_column_null :rss_logs, :created_at, false
+  end
+
+  def down
+    remove_column :rss_logs, :created_at
+  end
+
+end
--- a/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
+++ b/db/migrate/20220126211212_add_created_at_to_rss_logs.rb
@@ -2,7 +2,6 @@ class AddCreatedAtToRssLogs < ActiveRecord::Migration[5.2]
 
   def up
     add_column :rss_logs, :created_at, :datetime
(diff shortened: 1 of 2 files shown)


nimmolo:
I think this PR needs to go in front of #831, when it's ready.

pellaea:
I suspect you already did a partial migration at some point?  If so, you
can end up with a mess sometimes.  But in this case, just try running "rake
db:migrate:down" first.  If that doesn't work, then throw a temporary line
into the "up" migration which removes the column first.  This can happen if
you tried to do a migration in which it only half-succeeded: it created the
column but failed when, say, populating it.  Rails might consider the
migration to have failed, so that if you migrate down it won't undo this
migration (it thinks it never did this migration, so no need to undo it).
But there's still that lingering half-baked state in the db.  In this case,
super eassy to fix, just delete the column first.  OF course, you need to
remove that from the final version!  But to fix this half-baked state, it
works just fine.  Does that make sense?

nimmolo:
Yep, done.

It also didn’t like my change_column_null line, presumably because they’re all null, so i removed it.

Nimmo

nimmolo:
I think this super-simple migration is ready to go.

In the other PR, the revised `RssLog#detail` method will throw an error unless the `created_at` column exists, so i believe this migration needs to be executed first?

pellaea:
Sure, let me take care of that now... assuming I can load github!

pellaea:
You should also, presumably, be able to simply merge this PR into your
other PR, right?

pellaea:
Looks perfect.  Will merge immediately.  Good thing to learn how to do!

PR #838, opened 2022-01-29
Scaffold GraphQL API

nimmolo:
This PR

- Adds several graphql gems: `graphql`, `graphql-connections`, `graphql-batch`, `rubocop-graphql`
- Adds`/app/controllers/graphql_controller.rb`, `/app/graphql/...` type definitions (Ruby classes) and some config initializers and rake tasks
- Adds gem `graphiql-rails`, the rails port of the GraphiQL browser tool (note the "i") for performing live queries of the API in dev mode, available at `localhost:3000/graphiql`
- **Only enables the `/graphql` route, the API endpoint, in development**
- Adds gem `rack-cors` so a local frontend app (not included) can access the API locally, in dev mode, from port 3001 only

The goal of this PR is to get a work-in-progress version committed...

- so that graphql work doesn't keep getting out of sync with master, and 
- so the PRs for graphql work going forward can be smaller. 
 
However, I don't want this to create bugs, coverage decreases or constant codeclimate annoyance for other committers. I'm very open to suggestions about how to do this PR better, fix codeclimate complaints, etc.

One thing, though, there are going to be TODOs all over the /app/graphql/ directory, for sure - I'd like to keep those! The coverage reduction is something i'd like to fix as soon as possible, however. 

I need to read up on [testing](https://graphql-ruby.org/testing/schema_structure) [graphql](https://graphql-ruby.org/testing/integration_tests), and maybe for example I should comment out the queries that are stubbed out but not working yet.


pellaea:
I definitely support writing tests at all stages. They can be really basic
at first: for example just run each end point and don't even bother to test
the result!  Then progressively add to the tests as you firm up behavior
and expectations.

I've got no problem with TODOs.  I never understood that cop.  It just
encourages people to remove the word "TODO"!  (Perhaps I should just speak
for myself... ;)

nimmolo:
I'm closing this PR, and opening another one with vastly reduced scope.

PR #839, opened 2022-01-29
GraphQL gems, configs and rake task

nimmolo:
This PR should not alter any current behavior of the production server.

- Adds several graphql gems: `graphql`, `graphql-connections`, `graphql-batch`, `rubocop-graphql`
- Adds two config initializers needed to run graphql in development 
- Adds a rake task to generate a graphql schema dump
- **Only enables the `/graphql` API endpoint route in development**
- Adds gem `rack-cors` so a local frontend app (not included) can access the API, from localhost:3001 only
- Adds gem `graphiql-rails`, a rails port of the GraphiQL browser tool (note the "i") for performing live queries of the API in dev mode, available at `localhost:3000/graphiql`. It doesn't do anything here yet, though, because the graphql_controller and app/graphql are removed from this PR.

The goal of this PR is to get the base gems committed to master so that:

- gemfile, config, etc doesn't keep getting out of sync with graphql working branch, necessitating painful rebases 
- the PRs for graphql work going forward can be easier to understand. 
 
I don't want this PR to create any issue for other committers, so I've removed all the graphql app code. I'm very open to suggestions about how to do this PR better.

Note, there are many commented-out gems in gemfile with notes - things I may use, but am not using yet in development. I'm open to deleting those, but need to save the notes somewhere, they represent a lot of research. I could put them in a graphql working README. What would you guys prefer?


JoeCohen:
I get the same 2 test Errors that are reported by the GitHub CI. https://github.com/MushroomObserver/mushroom-observer/runs/4992819475?check_suite_focus=true (expand Run Tests).
Are they anything I should worry about?

nimmolo:
@JoeCohen Hmm.  I'm seeing two errors "Unable to upload image", putting them here for people checking in email:
```
Progress: |=================grep: /home/runner/work/mushroom-observer/mushroom-observer/log/test.log: No such file or directory
=================E, [2022-01-29T19:28:55.245010 #3173] ERROR -- : Unable to upload image
===========================E, [2022-01-29T19:29:51.400751 #3173] ERROR -- : Unable to upload image
=======|
```
No idea, but could it be anything with cors?

nimmolo:
Ah - I bet i know what it is. 
Check `config/environments/test.rb`, lines 85-92.   
```ruby
config.log_level = :ERROR
```

JoeCohen:
I don't have a problem with this -- to the extent I understand it. (But almost all of it is over my head.)
I'm OK with adding the commented out lines to Gemfile. But maybe group them all together and add a comment explaining why they're commented out; that should avoid accidental deletion.

JoeCohen:
> Ah - I bet i know what it is. Check `config/environments/test.rb`, lines 85-92.
> 
> ```ruby
> config.log_level = :ERROR
> ```

What is it trying to upload?

nimmolo:
@JoeCohen - 

I think those lines in test.rb are just making rails log a higher level of error detail on all existing tests? 
GraphQL isn't trying to upload anything, that must be from an existing image upload test.

Re: gem comments
I tried to do it logically, open to other suggestions.  
The graphql-ruby extension gems are grouped together, but they're sub-grouped by the functionality they perform. There are duplicate gem candidates for each functionality.

```
# Comment about functionality 1
# Top candidate gem
gem graphql-function
# 
# Runner-up unused candidate gem
# gem graphql-similar-function
# 
# Second runner-up
# gem graphql-alternative-function
```

config/environments/test.rb
```diff
@@ -82,6 +82,14 @@
   # To control the debugger turing testing
   config.activate_debugger = false
 
+  # Enable stdout logger
+  config.logger = Logger.new($stdout)
+
+  # Set log level
```
pellaea, line 88: Would you mind including more informative comments.  One can clearly see that it is setting something called log_level without your comment... however that tells me nothing about what it does! :)

config/initializers/cors.rb
```diff
@@ -0,0 +1,26 @@
+# Be sure to restart your server when you modify this file.
+
+# frozen_string_literal: true
+
+# Avoid CORS issues when API is called from the frontend app.
+# Handle Cross-Origin Resource Sharing (CORS) in order to accept
+# cross-origin AJAX requests. A different port is considered cross-origin.
+
```
pellaea, line 8: Exactly!  See, this is extremely informative.

config/environments/test.rb
nimmolo: @pellaea I'll try to track down why i added these - i've been trying to comment everything copiously but missed this one. They may have been generated by a graphql gem, on the other hand.  🤔

lib/tasks/graphql.rake
```diff
@@ -0,0 +1,16 @@
...
+
+# Note: Make dumps of schema for different environments - public/private!
+
+namespace :graphql do
+  desc "Write a dump of the entire GraphQL Schema"
+  task(dump_schema: :environment) do
+    # Get a string containing the definition in GraphQL IDL:
+    schema_defn = MushroomObserverSchema.to_definition
+    # Choose a place to write the schema dump:
+    schema_path = "app/graphql/schema.graphql"
```
pellaea, line 11: This may be picky, but can I assume that schema.graphql is more data than code?  I feel like data does not belong in the app tree.  Maybe this belongs in public or ... beats me! somewhere else ... even the rails root would be preferable (I already have a bunch of crud in my rails root, for example, but I keep everything else but public clean).

pellaea:
Ah!  That would make sense.  Automated coding is a wonderful dream, but we
are so totally not there yet, as this would prove...

pellaea:
I love how simple this PR is, Nimmo!  Perfect.

So, I know you've talked about this a lot, especially in private emails with me, but naturally I wasn't paying attention or have forgotten.  It might be helpful to give us a hint how to get started with this new fangled machinery.  Like and example query we can test to see the lights blink or a link to a brief intro or something?

I'll try my usual "push random buttons until something happens", but I suspect that might not cut it with this thing! :)

nimmolo:
Glad to hear it. I will address the issues you just raised.

You won’t be able to actually query anything till the graphql controller and quite a few of those types are in. To even get the observations query returning anything good, i had to write a shit ton of files and type definitions, more than 90.

I’ll try to commit as few as possible at a time, but the next PR will be more involved. There are base types, queries, mutations, etc. that app-specific analogues inherit from, and then very detailed definitions of all the specific data types and queries themselves. And they will all need test coverage and careful vetting.

Next PR - the core definitions.

Nimmo

pellaea:
Ah!  Understood.  Makes perfect sense.

BTW, all tests passed just fine for me.  (I commented out the log_level
thing in the config file.)

JoeCohen:
I think that once this is merged, we'll see those log entries when we run the tests locally. 
Is there a way to avoid that (other than doing what Jason did).

pellaea:
I'm still confused about what they're for to start with.  Debugging?  More
feedback while writing tests?  In either case, shouldn't these settings be
turned on only when actively working on coding?

JoeCohen:
PS: @nimmolo is correct about those errors already existing in master.
(I added the lines to master, ran tests and saw:
```ruby
E, [2022-01-30T01:29:18.352572 #81900] ERROR -- : Unable to upload image
E, [2022-01-30T01:29:34.113522 #81900] ERROR -- : Unable to upload image
```

pellaea:
Oh!  Okay, great, that probably is uncovering a pre-existing bug in the
tests, then, and therefore we should leave the log_level set to error.  I
guess I'll try to figure out why that's happening at some point.

JoeCohen:
I sort of remember a test that required an image upload to fail.
I'll have a quick look to see if memory is correct, and -- if so -- whether
that's causing the log entries.

JoeCohen:
One of the errors is logged when executing the POST request in `test/controllers/observer_controller_test.rb#test_image_upload_when_process_image_fails`.
That comes immediately after
```ruby
    # Simulate process_image failure.
    Image.any_instance.stubs(:process_image).returns(false)
```
I'll try to run down the other one. (It's somewhere in the controller tests.)

JoeCohen:
The other one is in `test/controllers/glossary_terms_controller_test.rb#test_create_process_image_failure`:
```ruby
  def test_create_process_image_failure
    login
    # Simulate process_image failure.
    Image.any_instance.stubs(:process_image).returns(false)
    post(:create, params: term_with_image_params)

    assert_flash_error
  end
```

pellaea:
Thanks, Joe.  We probably just need to remove the error log.  I certainly
don't pay attention to that in the production log.

JoeCohen:
So it's not a bug.  Some options:
- Don't change config/environments/test.rb;
- Rewrite the tests. (I remember puzzling over whether there was a better way to assure that the image processing would fail, and not finding one.)
- Delete the tests. (I don't like this option.)
- Live with the display of the error log when running the tests. (I also don't like this. It's confusing and implies that there is a bug.)

pellaea:
You missed the option of removing the error message!

pellaea:
There are three places we log the failure to upload an image.  Just remove
each of them:

    logger.error("Unable to upload image")

They serve no purpose to my knowledge.  No one pays any attention to those
messages.

JoeCohen:
I don't understand how that would would work. For me the issue is that the logged errors go to $stdout so get displayed during testing. That confused both of us.

JoeCohen:
Sorry, I was typing when you explained how to do that.

pellaea:
You are of course welcome to fix it better as well.  Maybe some sort of
stub?  That just seemed the simplest, and probably should be done in any
case.

JoeCohen:
I see 4 places. (And 3 of them could be DRYed a bit.)

JoeCohen:
I don't know how the fix the tests. 
So I think that to move forward we should use your suggestion -- don't log that error in the first place. (And maybe DRY 3/4 places that were logging it.
I'll sleep on it.
But anyone else is welcome to run with it.

pellaea:
Oops, sorry, I was being too sloppy.  Five if you include the one in the
unused image_s3 class.  They can all go so far as I care.

pellaea:
My take on this is that it's not necessary for GraphQL, and it's certainly
not critical, so go with the easiest solution, since it is a step in a good
direction anyway.

JoeCohen:
Just to clarify: What you mean is that logging that error is the the right way to go.

pellaea:
My opinion is we should remove all the logger.error("Falied to upload
image") lines because it is the easiest thing to do that fixes the present
problem, and those error messages are useless and ignored anyway.  But it's
just my opinion, and I'd be happy to support better solutions if someone
else has time to implement them!  I'd be happy to take care of this trivial
fix right now, if it would help.

JoeCohen:
I agree. It would be great if you could fix them. (I'm looking at the Activity Log delete issues right now.)

pellaea:
Okay, will do.

JoeCohen (approved):
I'm fine with merging this.

lib/tasks/graphql.rake
nimmolo: Moving it to /public/graphql :)

PR #840, opened 2022-01-30
Stage graphql_controller, generated classes and User type

nimmolo:
This is just the graphql_controller and several base object classes (in app/graphql) generated as scaffolding for a graphql API. Also, an app/classes/token class to deal with token encryption and decryption.

The User type and query, ~login input type and mutation~ have been added to app/graphql. Note the User type has all its associations commented out.

There are two basic "context queries" to figure out `current_user` and `in_admin_mode?` from the token: Visitor and Admin. 

There's also an "ImageUrls" interface that generates url fields for the appropriate `Types::Models` (currently just `UserType`). These are derived for all image sizes via `Image.url` from the appropriate `image_id` db column for each model.


pellaea:
Excellent, I'm looking forward to seeing this.  I appreciate this slow
introduction.  It will give me time to see what is auto-generated and not,
what the basic functionality is, etc.

nimmolo:
You'll notice the `types/models` (currently, only `User`) parallel Rails model definitions and db schema.
For `User`, I just commented out all the associations because the types don't exist yet in this PR.

**Question**
Because GraphQL is all about data typing, I believe this is a good time to `create_column_null` wherever relevant in the User model --`created_at` is a candidate. What do you guys think?

In subsequent PRs, as each model is added to graphql's types, we can do the same, and we'll have [@pellaea 's Pivotal story](https://www.pivotaltracker.com/story/show/153682526) solved.

nimmolo:
The types/models files are generated by a command line, but that only gets you the fields that come directly from **table columns**. IIRC i manually added the nullability params, checking against schema.rb.

I definitely manually added all the associations and field resolvers.

app/graphql/types/models/user_type.rb
```diff
@@ -0,0 +1,138 @@
...
+
+    # has many
+    # field :api_keys, [Types::Models::ApiKeyType], null: true
+    # field :comments, [Types::Models::CommentType], null: true
+    # field :donations, [Types::Models::DonationType], null: true
+    # field :external_links, [Types::Models::ExternalLinkType], null: true
+    # field :images, [Types::Models::ImageType], null: true
+    # field :interests, [Types::Models::InterestType], null: true
+    # field :locations, [Types::Models::LocationType], null: true
+    # field :location_descriptions, [Types::Models::LocationDescriptionType],
```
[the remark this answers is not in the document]
nimmolo: I'm trying to do Rubocop-friendly line breaks between params here, but i don't know what it wants.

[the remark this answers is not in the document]
pellaea: codeclimate is complaining about invisible white space at the end of the line, not the blank line

app/graphql/types/models/user_type.rb
```diff
@@ -0,0 +1,138 @@
...
+    # field :authored_names, [Types::Models::NameDescriptionType], null: true
+    # field :edited_names, [Types::Models::NameDescriptionType], null: true
+    # field :authored_locations, [Types::Models::LocationDescriptionType], 
+    #                            null: true
+    # field :edited_locations, [Types::Models::LocationDescriptionType], 
+    #                          null: true
+    # field :curated_herbaria, [Types::Models::HerbariumType], null: true
+
+    # custom fields
+    # field :img_src_thumb, String, null: true
```
pellaea, line 105: What are all of these img_xxx fields going to be for?  They might make sense in the image model, but not here in the user... or am I missing something?

nimmolo: They all just call the Image.url method. I guess those URLs are created programatically in Rails - and they could be created programatically on the front end from the Image id, but in case the Rails method changes in the future, i thought it made sense to have this on the Rails side. (Hmmm. Maybe it should be one field returning an array...)

View wise, they deliver a `src` for every size we have (using HTML `srcset` allows a modern browser to download the smallest image necessary for the space occupied by the img at current screen size.) For a "matrix box" carousel with a fullscreen mode, if that's desirable, for example, you'd encode all those sizes in the `srcset`

nimmolo: Oh - the reason they're here is just so we don't have to query the (user profile) image from the image table. Calling Image.url on the User.image_id gives the urls without hitting the image table.

But i think i understand what you mean, they could still all be in the Image model as methods!

pellaea:
Right!  That's exactly what I mean, I think.  Thank you for thinking about
avoiding the unnecessary hit on the images table.  Same thing will apply to
the observation thumbnail image, too.

nimmolo:
That's a cool thing about that Graph**i**QL "API explorer" - you can run a query and observe every db hit, in a terminal window, as it's happening. n+1's really show up - that's how I noticed it, and why i started working on a way to avoid the hit.

I think Glossary terms and another model have images too, maybe herbarium. I will make a helper method for all these types when i figure out how to do it. The "featured" images have different column names for different models, though. It makes accessing the stored image id a bit trickier.

pellaea:
Something which might help.  Some or all of the polymorphism of image
rendering is encapsulated in Image::Url.  Perhaps there's a way to use it
as-is, or tweak it / beef it up so that you could reduce the repetition in
each graphql object type to a minimum.  Honestly, all it needs is an id!
For some purposes you also need, perhaps, size and/or transferred bit,
too.  But as a first pass / default, id alone should be enough for
everything.

nimmolo:
One prob, i believe in Image.url - it doesnt fill in the domain in some cases (idk which! Seems random.) Only returns a relative url. 

That works on the live site but not in the separate frontend, so maybe we could add a check that it's returning a full Url.?

pellaea:
Right, that might be worth fixing so that it always returns a fully
qualified URL.  It might have something to do with the image
configuration.  Check config/image_config.yml.  If development local files
are served relative, remote files are fully-qualified.  Maybe you want to
change the local "read" spec to "http://localhost:3000/images"?  Is there
any reason to worry that this may not be same on mac versus linux, or
vagrant versus docker?  The relative address is by contrast very likely to
work on all of the above, no matter what port, etc.  Presumably that's why
I wrote it that way.  All of the production specs are fully-qualified.

nimmolo:
Sorry - nevermind! Just dug a bit deeper and realize the observations in question simply did not have a `thumb_image_id`, and i forgot to check before sending a null id to Image.url

pellaea:
That would do it!

app/models/user.rb
```diff
@@ -512,6 +515,33 @@ def verify
...
+    # else
+    #   raise "bogus user"
+    # end
+  end
+
+  # param is an http_auth_header
+  def self.get_from_token(auth_header)
+    user_id = User.crypt.decrypt_and_verify(auth_header).to_i
+
+    user = safe_find(user_id)
```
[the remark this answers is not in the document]
nimmolo: I'm using the variable immediately below, in the method. 
How should i rephrase?

test/models/user_test.rb
```diff
@@ -327,4 +327,17 @@ def test_notes_template_validation
...
+
+  def test_graphql_token
+    token = rolf.create_graphql_token
+    user = User.get_from_token(token)
+    assert_equal(rolf.id, user.id, "Token for user created and decrypted")
+  end
+
+  def test_unverified_graphql_token
+    assert_raises("User not verified") do
+      token = users(:unverified).create_graphql_token
```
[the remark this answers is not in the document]
nimmolo: I'm using the variable immediately below, in the method. 
How should i rephrase?

test/graphql/http_requests/http_request_test.rb
```diff
@@ -0,0 +1,34 @@
+require("test_helper")
```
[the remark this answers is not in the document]
nimmolo: it's there...

nimmolo:
OK - This PR is the first to make use of the new Rails > 5.2 credentials added in #806, and it's the reason I proposed adding them. The new class generates the tokens.

However, they won't work unless the new **master key**, which I emailed around at the time but can easily resend, is added to the production environment. If this hasn't been added already, I believe that's why the test is failing, where it passes locally for me.

The master key is intentionally not part of the repo, so it needs to be [uploaded to the production server](https://medium.com/cedarcode/rails-5-2-credentials-9b3324851336) and required in config/environments/production.rb to work on the live site.

```ruby
config.require_master_key = true
```

I guess the master key also needs to be somehow added to codeclimate's environment so it can test using the class. I have no clue how to do this!

test/models/user_test.rb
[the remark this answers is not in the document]
JoeCohen: I'm not not sure. Perhaps this is what you mean
```suggestion
  def test_unverified_graphql_token
    assert_raises("User not verified") do
      token = users(:unverified).create_graphql_token
      User.get_from_token(token) 
    end
  end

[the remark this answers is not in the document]
Comment (JoeCohen) [reply test/models/user_test.rb:339]: GitHub didn't format that right, but I think you'll figure it out.

**Comment by nimmolo** [comment]
@JoeCohen It looks like our [ci_rails.yml](https://github.com/MushroomObserver/mushroom-observer/blob/master/.github/workflows/ci_rails.yml) may be the place to somehow set the master key without putting it in the repo? I'm still researching, since there must be a standard way to do this.

**Comment by nimmolo** [comment]
*Update* I found [various proposals for adding the Rails master.key to GitHub CI.](https://stackoverflow.com/questions/61444542/github-actions-and-rails-master-key). 

@JoeCohen @pellaea Do you guys like any of these, keeping in mind that of course the secret cannot live in the repo? [This one](https://stackoverflow.com/a/64520213/3357635) looks right to me, but I think maybe only you two and Nathan can add a secret or environment variable.

**Comment by JoeCohen** [comment]
My memory is that the key itself does not go into ci_rails.yml
<https://github.com/MushroomObserver/mushroom-observer/blob/master/.github/workflows/ci_rails.yml>
(which
is in the repo and under version control).
Rather there's a way to refer to the key so that it gets passed to
Coveralls.
It might already be there.
# https://github.com/marketplace/actions/coveralls-github-action
- name: Coveralls
uses: ***@***.***
with:
github-token: ${{ secrets.GITHUB_TOKEN }}
path-to-lcov: ./coverage/lcov/lcov.info


But I don't have time to dig into this until Mon or Tues.

**Comment by nimmolo** [comment]
@JoeCohen Thanks for checking it out, no need to respond til Monday. 

I think there is an option in GitHub (for admins, but not for me) to add secrets **to** the repo that are not **in** the repo, and those secrets can get passed to CI, I believe by setting an environment variable in ci_rails.yml
 
As far as I can tell, ci_rails.yml is also not under version control (?? or at least not in the mushroom-observer i'm working with)

**Comment by nimmolo** [comment]
OK - I added some more test coverage (for User Query) and I believe this PR is good to go.

Now that master.key is on the production server, though we're not using it on the live site yet, I  add this comment to config/environments/production.rb so we don't forget it's necessary later. 

```ruby
# config.require_master_key = true
```

app/graphql/mushroom_observer_schema.rb
```diff
@@ -0,0 +1,43 @@
...
+  def self.resolve_type(_abstract_type, _obj, _ctx)
+    # NOTE: Implement this function to return the correct object type for `obj`
+    # raise(GraphQL::RequiredImplementationMissingError)
+  end
+
+  # Relay-style Object Identification:
+
+  # Return a string UUID for `object`
+  def self.id_from_object(object, type_definition, query_ctx)
+    # Here's a simple implementation which:
```
pellaea, line 28: Just plain old object.id is guaranteed to be "unique within type", no reason to encode it.

test/graphql/http_requests/http_request_test.rb
```diff
@@ -0,0 +1,49 @@
...
+         params: { query: query_string },
+         headers: { "User-Agent" => "iPadApp" })
+
+    json_response = JSON.parse(@response.body)
+
+    assert_nil(json_response["data"],
+               "Unauthenticated requests have no current_user")
+
+    # This time, add some authentication to the HTTP request.
+    rolf = User.find_by(login: "rolf")
```
pellaea, line 33: Do you have access to the standard fixture helpers like "users(:rolf)"?

nimmolo: Yes... commit coming!

app/models/user.rb
```diff
@@ -512,6 +515,29 @@ def verify
     save
   end
 
+  # Create a Rails native token (used for authentication in GraphQL API)
+  def create_graphql_token
+    raise("User not verified") unless verified
```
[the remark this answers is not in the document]
nimmolo: Name proposal welcome. The situation i'm trying to guard against is spam/bot registration - quick bot signup with fake email, and then sign in with costly queries. There's no way i can imagine an unverified user should get a token... When they first create an account, that shouldn't give them a token either, i believe. Even on the current site, they sign up, but must verify their email, and only then can they login. right?

app/graphql/mushroom_observer_schema.rb
nimmolo: This is part of a group of necessary hooks stubbed out by the graphql-ruby gem author as examples - I think it's typically used for something specific to graphql, to do with giving "nodes" unique IDs, even when there are possible duplicate nested records, or ensuring some other "uniquity" within the query object, maybe because different records are mixed in the query. I commented out because I don't understand it yet!

I should just comment out the whole method, but the docs say it's necessary. They're not implemented anywhere yet as far as I can tell. Here's a partial [explanation](https://graphql-ruby.org/schema/definition#object-identification-hooks) from Robert Mosolgo.

nimmolo: For example within the nodes of a query there could be multiple instances of the same user. I believe graphql wants each node to have a UUID (or already automatically gives them one).

```
observation {
  name
  user
  location
  species_list {
    user
  }
  sequence {
    user
  }
}

**Comment by nimmolo** [comment]
Regarding decreased coverage:

The scaffolded and yet unused base types for mutation, inputObject, interface, connection, etc are a lot of the coverage decrease. I can remove them and re-add with later PRs as necessary, it'll just be more confusing to re-require and include them as parameters where they go in other types.

Any suggestions for the controller? By the way, I remembered why the string `user-id:` was encoded in the token:  we may also want to put `autologin:` and `admin-mode:` in the token. (Does that sound right? The token is the only source of session state between the front end and Rails.)

**Comment by pellaea** [comment]
I get the part about the unlikely but real possibility of a bad agent
trying to sneak a token for an unverified user in there, and definitely
agree that we need to test for that.  But I'm questioning whether to have
it happen automagically *here* where you lookup the user instance.  I say
we let this go for the moment and I'll wait until I see how it is
eventually used within the context of GraphQL before making any
suggestions.  I'm still also privately holding out for the possibility of
moving these methods out of User some day.  Fortunately, the method names
are distinctive and unique, so it will be easy to change our minds on this
later.  (We will be able to search for all uses of User.find_by_token and
User#create_graphql_token and modify as required easily.)  But again, I
need to understand the broader picture before I can even form an opinion
let alone make a coherent argument for any concrete change.  Just some
things that bother me and that I'm going to keep an eye on.

**Comment by pellaea** [comment]
Right, I figured that the user of "user_id:" implied there would eventually
be other data included.  But in that case, we need to fix the parser.  I
think it was something like token.gsub("user_id:", "")  but that will leave
all the other data in there.  And writing a new token becomes complicated:
do we want to write a fresh new token with just one bit of data, or add a
bit of data to an existing token, or change the value of user_id in an
existing token?  If this hasn't already been anticipated and solved by
rails or GraphQL, we probably need to write a simple Token class that can
handle these sorts of requests.  Ah, but I see, doing it this way will
automatically be forward-compatible!  Bare user_id would not, the string
"user_id:1234" is.  Okay, again, I'm understanding why you've got the
quick-and-dirty gsub parser in there.  You already know that's temporary,
so no reason to sweat over it...  Okay, yet again, ignore me and proceed.
Hopefully I'll remember all these things.  (Ha, what a joke!)

**Comment by nimmolo** [comment]
It's awesome that you know enough to be opinionated on these things!!

Okay, i'm feeling like i should treat coverage as sacrosanct. I'm pulling out the yet-unused base classes. 

I think i'd like to try to figure out `admin_mode` and `autologin` for the token now, so i can have your keen insight on it while you remember the whole arrangement. This can be our "auth" PR.

**Comment by pellaea** [comment]
I was hesitating to recommend this, because it sounds like a pain in the
butt for you.  But honestly, it would make the PR much easier to review.
There is so much unused stuff in there that I had trouble figuring out what
actually *was* used!

**Comment by nimmolo** [comment]
I reckon the token should be built as a hash, converted to a string and encrypted, and then decrypted as a string and parsed to a hash for manipulations? Seems more convenient than string parsing.

**Comment by pellaea** [comment]
Definitely.

**Comment by nimmolo** [comment]
Ok - the methods are modified (certainly starting to look like a candidate for a Token class!) and the hash is parsed ok. 

But i can't seem to get `token_in_admin_mode?` to work. It's in there commented out, but when i call it from `graphql_controller` it's giving 

`ActiveSupport::MessageVerifier::InvalidSignature` in `::User.decrypt_token_hash`. 

I can't imagine what's wrong -  i modeled `self.token_in_admin_mode?(http_auth_header)` on `self.get_from_token(http_auth_header)` and it's passed the same header param to parse in `graphql_controller`. Does the signature expire once it's used?

**Comment by nimmolo** [comment]
Sorry, got it. Commit coming

**Comment by nimmolo** [comment]
Although coverage is now increased (first time i've been a part of that!), I should add some tests for invalid queries and variables to exercise the error handling side of the graphql_controller

File: app/models/user.rb [hunk 9]
```diff
@@ -512,6 +515,54 @@ def verify
...
+
+    token_hash = { user_id: id, in_admin_mode: false }
+    User.encrypted_token(token_hash)
+  end
+
+  def create_admin_token
+    # if they're not an admin, just return false and keep the token
+    token_hash = { user_id: id, in_admin_mode: admin }
```
pellaea, line 528: Note that admins are rarely in admin mode.  Admins are the only users who can enter admin mode.  But there's no reason to store whether the user is an admin in the token, since it's in the user.  But whether or not the admin has turned on admin mode is a state, and should be either stored in the session (I forget whether GraphQL uses a session), or in the token.  Very little is kept in the session generally, so I see no reason we can't put it all in the token, personally, and this has the added advantage of keeping GraphQL state-less.

nimmolo: I think i'm finally one step ahead of you (it is, and will remain rare, let me enjoy!) This is correctly stored in the token - there's no session. The way it works:

**Login** - `create_graphql_token`. No one's in admin mode at login, `in_admin_mode: false`

**An admin user turns admin mode on** - calls `create_admin_token`. `in_admin_mode: true`

**A non admin tries to turn admin mode on, somehow** - `create_admin_token` simply silently keeps admin mode off. `in_admin_mode: false`

**An admin user turns admin mode off** - calls `create_graphql_token` which again sets `in_admin_mode: false`

That's why it's `in_admin_mode: admin` - it's pulling directly from the user's admin prop.

app/classes/token.rb
```diff
@@ -0,0 +1,28 @@
...
+  def self.decrypt_from_header(auth_header)
+    token = Token.crypt.decrypt_and_verify(auth_header)
+    hash = JSON.parse(token)
+    new(hash)
+  rescue ActiveSupport::MessageVerifier::InvalidSignature
+    nil
+  end
+
+  def self.crypt
+    @crypt ||= ActiveSupport::MessageEncryptor.new(
```
pellaea, line 24: I guessed at random about whether to use @ (class instance var) or @@ (class variable).  I suspect I was wrong, and this should be @@.  Joe, do you remember?  I looked it up for a while, and the thing that resonated with me is that class instance variables (@) are in a sense an "accident" of how ruby creates classes.   But class variables respect inheritance and behave how most people would expect class variables to act.  So I think think @@ is recommended over @...  I think...

app/controllers/graphql_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+    handle_error_in_development(e)
+  end
+
+  private
+
+  # https://www.howtographql.com/graphql-ruby/4-authentication/
+  # Decrypt the current user from token stored in the header (not the session)
+  # Note: There may not be a token!
+  def current_user
+    return unless token.is_a?(Hash)
```
pellaea, line 40: I would consider making these ... oh shoot there's a term for it, but I'm forgetting the lingo...  ah! "memo"... I would consider memoizing this method and a few of the others below.  All the ones you can reasonably expect to need multiple times for a given query and which involve significant processing.  Basically, fix it to do all this processing once, and then cache the result so it doesn't have to do it again on subsequent calls.  (Things that can return nil may require some tweaking.)

app/controllers/graphql_controller.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def in_admin_mode?
+    return false unless token.is_a?(Hash)
+
+    token["in_admin_mode"]
+  end
+
+  def token
+    ::Token.decrypt_from_header(http_auth_header).hash
```
pellaea, line 58: Definitely memoize this one, too.

nimmolo: Just attempted this - 
It works if I add `@token ||= ::Token.decrypt_from_header(http_auth_header).hash` but not if i'm replacing `token` with `@token` everywhere in the other methods. I don't understand how it's helping if i don't use the instance var everywhere?

pellaea:
This PR looks great now.  I've made a couple of recommendations.  See if you think they're worth it.

nimmolo:
Memoizations in! But not sure if i did them right...

pellaea:
No, that's perfect.  It will only evaluate the right hand side once if it
resolve to something non-nil.

nimmolo:
Oh - but should `@token ||= blah` be `@@token ||= blah` also?

In this case it seems like the "class instance" part is appropriate. Each token is evaluated per instance. (For instance!)

app/controllers/graphql_controller.rb
pellaea: This works for me.  The main thing to avoid having to do every time is User.safe_find.  Thanks!

nimmolo:
Oh... One outstanding question (beyond this pr's scope, also):

> Because GraphQL is all about data typing, maybe this is a good time to `create_column_null` where it's missing and relevant in the User model --`created_at` is a candidate?

Sooner or later having a mismatch on field nullability may create problems with GraphQL. I don't know what the situation is in the db currently, it just seemed worth raising the question.

pellaea:
Ah, no, because that's an instance method, not a class method.  The token
is going to be different for each GraphqlController instance (that is, for
each request).  But the encrypter in Token.crypt is going to be constant
for all requests, as it depends only on the global credentials.  I think
this is exactly right.

pellaea:
Personally, I don't think this is so important.  For one, it should be
impossible to alter the state of users.created_at.  And second, the null
thing is just for error-checking, so it's not critical.  The worst case
would be if the database won't accept null but graphql does, then the
database will throw an unexpected error.  But what's so bad about the db
throwing the error instead of graphql?  I mean, sure, if it's convenient,
fix these things as you come across them.  But I wouldn't waste much time
on it.

nimmolo:
What brought it to mind was those last migrations making `rss_logs.created_at`, and making it non-nullable -- it was an extra step, but easy. I don't know what it affects, it just seems to make sense to tidy it up, as long as it's a simple migration... a basic thing that should be safe to assume, and could cause unexpected headaches in the future?

However there are many other tables in the db where `created_at` is nullable. Luckily (in the db snapshot I am working with) I just checked them, and I didn't find any actual null records in any of those tables. This implies maybe that the migration is not necessary -- but also that doing it might not disrupt anything.

Incidentally, reading through the schema just now i'm noticing lots of tables where there is an `updated_at` column but no `created_at` - this was the rss_logs situation too. That seems more involved to do, if it's even necessary. The nullability seems simple though. Am i missing a step?

pellaea:
Can we do it in a separate PR, at least?  Then if there is some unexpected
side effect or complication it will be better isolated.

nimmolo:
Oh yeah. I didn't mean this one. I figure any db migration should be its own PR in any case. 

sorry I should have put all these questions in a separate email thread!

pellaea:
No worries!  Migrations do not always belong in a separate PR.  But in this
case, I agree, they can live in their very own private PR without anything
else.

app/classes/token.rb
JoeCohen: I think the relevant difference is that class variables can be modified by subclasses, which in many circumstances can lead to unpleasant surprises.  Good examples under "What is the difference between class variables and class instance variables?" at https://www.ruby-lang.org/en/documentation/faq/8/

app/graphql/queries/user.rb
```diff
@@ -0,0 +1,26 @@
...
+    type Types::Models::UserType, null: false
+    argument :id, Integer, required: false
+    argument :login, String, required: false
+    argument :email, String, required: false
+    argument :name, String, required: false
+
+    def resolve(args)
+      return {} unless args
+
+      if args.key?(:id)
```
JoeCohen, line 15: Consider replacing these with a `case` statement or a series of `return`

app/graphql/queries/user.rb
```diff
@@ -0,0 +1,26 @@
...
+        ::User.find(args[:id])
+      elsif args.key?(:login)
+        ::User.find_by(login: args[:login])
+      elsif args.key?(:email)
+        ::User.find_by(email: args[:email].sub(/ <.*>$/, ""))
```
JoeCohen, line 20: I'm too dense to understand this. Is this how users who are not in admin mode are prevented from finding a user by email?

test/graphql/user/user_query_test.rb
```diff
@@ -0,0 +1,115 @@
...
+          }
+        }
+      GRAPHQL
+
+      user = rolf
+
+      result = MushroomObserverSchema.execute(
+        query_string, variables: { id: user.id }
+      )
+      user_result = result["data"]["user"]
```
JoeCohen, line 30: Is there any way that someone who is not in admin mode could discover the email through this query?

JoeCohen (commented):
I looked at this, but can barely contribute anything.

pellaea:
I think I recognize this.  You're talking about the bit of code removing
the "<blah>" at the end of email addresses?  The autocompleter
auto-completes not only the email address, but also the user's name in
<angle brackets>.  Instead of removing that on the client in javascript
before submitting the form, I chose to strip it out on the server in the
controller.  I don't know the details of what's going on here, but maybe it
will be similar in the end.

pellaea:
Haha, thanks, Joe!  See this is why we keep you around.  I even tried to
look this up, but there are conflicting opinions out there.  The issue does
seem to revolve around what people expect.  In this case there is unlikely
to be a subclass of Token.  If there were, what would it be?  Maybe we
would have multiple subclasses each with a different kind of encryption?
If so, then class instance variable is definitely the way to go!

Okay, I'll try to drill it into my thick skull: "never use @@"...  As if
I'll remember...

app/classes/token.rb
nimmolo: Ok so i'm going back to `@crypt` here then?

JoeCohen:
I think you're right, unless you think that there will be a subclass which
will need to change it.

app/graphql/queries/user.rb
nimmolo: I couldn't figure out a way to phrase this with `case`, please let me know if something jumps out. It's intended to be a versatile query that can be used with one arg only.

The processing advantage of `case` as i understand it would be that the case predicate is only evaluated once, and then it's compared to the `when`s. 

But here it's looking for a different key in a small array each time, and i couldn't think of a way to process the `args` in a way that the `when`s wouldn't just be looking in some other derived array.

pellaea:
Am I missing something... this is already @crypt, right?  But yes, I defer
to Joseph on this -- he's so much better at finding out what the consensus
/ current best practices in the ruby/rails community.

test/graphql/user/user_query_test.rb
nimmolo: Yes - hahaha! This is not really over your head is it Joe?! 

As of this PR nothing is exposed to the internet by the graphql api, but you're right on. Field authorization (more granular than, but in addition to, query and mutation auth) is one of the main things we need to build in on the Rails side. 

Next PR will introduce protected fields, and we'll need to implement them on User. Or if you guys prefer i can add them to this PR once I figure it out. It's pretty simple.

JoeCohen:
You're right about case. I wasn't paying attention.

pellaea:
Yes, I thought so, thanks for clarifying that, Nimmo!

nimmolo:
I just demoted `@@crypt` to `@crypt` in the latest commit. Should be good to go, if you guys approve.

(I've been playing with field authorization, but it's not working yet.)

pellaea:
Done, thanks, Nimmo!

pellaea:
Done, thanks, Nimmo!

PR #442, opened 2019-01-31
Rubocop Style/R...

JoeCohen:

PR #841, opened 2022-01-31
Changed RssLog#target_simply_deleted? to be a more generally useful R…

pellaea:
…ssLog#orphan? and fixed it to account for more pathological cases found in the wild of improperly-orphaned logs -- that is, orphaned logs that still point to a nonexistent target and for which target_type would mistakenly return non-nil.

### DIFF
--- a/app/models/rss_log.rb
+++ b/app/models/rss_log.rb
@@ -13,45 +13,47 @@
 #  associations for the different types of owners.  The resulting performance
 #  hit was significant.
 #
-#  Possible owners are currently:
+#  List of models which log activity are presently:
 #
+#  * Article
+#  * GlossaryTerm
 #  * Location
 #  * Name
 #  * Observation
 #  * Project
 #  * SpeciesList
-#  * GlossaryTerm
-#  * Article
 #
 #  == Adding RssLog to Model
 #
-#  I think this is relatively easy.  Try following these steps:
+#  This might be out of date, but start with these steps:
 #
-#  1) Add columns to rss_logs and new model tables via migration:
+#  1) Database: Add columns to rss_logs and new model tables via migration.
 #
 #       class AddRssLogToModel < ActiveRecord::Migration
-#         def self.up
+#         def self.change
 #           add_column(:rss_logs, :model_id, :integer)
 #           add_column(:models, :rss_log_id, :integer)
 #         end
-#         def self.down
-#           remove_column(:rss_logs, :model_id)
-#           remove_column(:models, :rss_log_id)
-#         end
 #       end
 #
-#  2) Inform model of the new association: (automatically inherits +log+
-#     method from AbstractModel)
+#  2) New model: Create association, tell it which actions (if any) to
+#     automatically log, provide standard title methods.  (Note that
+#     AbstractModel will provide the +log+ method. The optional exclamation
+#     points tell it to pop the object to the top of the Activity Log.)
 #
 #       belongs_to :rss_log
+#       self.autolog_events = [:created!, :updated!, :destroyed!]
+#       def text_name
+#       def format_name
+#       def unique_text_name
+#       def unique_format_name
 #
-#  3) Inform RssLog of the new association:
+#  3) RssLog: Create association, add type to RssLog.all_types.
 #
-#       (just search for "location" in this file)
+#       belongs_to :model
+#       self.all_types # add "model"
 #
-#  4) Add partial view for +list_rss_logs+:
-#
-#       (just clone, e.g., app/views/observer/_location.rhtml)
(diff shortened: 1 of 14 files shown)


app/models/rss_log.rb
```diff
@@ -387,6 +387,11 @@ def detail
     ""
   end
 
+  # Has target been destroyed (orphaning this log)?
```
JoeCohen, line 390: Is `orphan?` intended to be part of the class's public interface?  (I see that it's called by the tests.)
- If no, consider either:
  - Moving it to the private section, and in the tests accessing it with `send`
  - At least moving it the next section
- If yes, document it with other methods (after line 136).

app/models/rss_log.rb
```diff
@@ -387,6 +387,11 @@ def detail
...
+  def orphan?
+    !target_type || !notes.match?(/\A\d{14}/)
```
JoeCohen, line 392: Does this deserve some comment?

JoeCohen (commented):
Looks good.
IMO you should do whatever  you want in response to my (2) comments about particular code changes and the question below. Then merge and deploy.
(I may follow up in a separate PR with mostly RuboCoppish stuff.)

**Question**: In the documentation of `orphan`, should the following be deleted?
`clear references to associated object`. (See lines 334-335).

JoeCohen (commented):
I had a quick look, and it looks good.
Thanks for all the work on this. What a project it turn out to be.

JoeCohen:
Thanks again @pellaea for doing this, including digging through the logs in the db to find the problems.

JoeCohen:
Sorry. Just noticed this from RuboCop.  But I don't understand the code or how to fix the alleged offense.
If it can't/shouldn't be fixed, maybe turn off the cops,, adding a comment why.
>app/models/rss_log.rb - 2 offenses
Line #442 – Convention: Style/FormatStringToken: Prefer annotated tokens (like %s) over unannotated tokens (like %s).
    str.to_s.gsub(/[%\s]/) { |m| "%%%02X" % m.ord }
Line #442 – Convention: Style/FormatString: Favor format over String#%.
    str.to_s.gsub(/[%\s]/) { |m| "%%%02X" % m.ord }

pellaea:
There are three settings for that cop.  Has anyone actually looked into
which we prefer?  It's easy enough to fix, though.

JoeCohen:
For the [Style/FormatString](https://docs.rubocop.org/rubocop/cops_style.html#styleformatstring) and [Style/FormatStringToken](https://docs.rubocop.org/rubocop/cops_style.html#styleformatstringtoken) cops, we've used the default settings ("format" and "annotated", respectively). 
We've never had a conversation about which format we should use. Maybe we should. I don't know the best place to do that. Maybe we can follow up in the 5-Feb tech meeting.

pellaea:
If you like those settings, that's fine.  I just saw that they were the
defaults, so I wasn't sure anyone had actually looked or cared.

app/models/rss_log.rb
pellaea: Yes, I did intend to make it public even though we have to potential uses for it yet.  If I had to articulate my feelings on the subject... I guess I prefer to make public, potentially useful methods over private, highly-specialized methods.  Yes, it's the same exact method!  But what was target_simply_deleted? doing?  It was in effect checking to see if the log is an orphan.  Well, hell, that's a really useful piece of information lots of people might want access to.  But no one would think to use "target_simply_deleted?"  That makes sense -- and it makes great sense, I admit! -- in the context of RssLog#detail, but outside of that context I would personally wonder about the choice of wording.  I would assume it was doing something highly specific and customized for its one caller.  It's not, but I would have to look up the code to figure that out.  Then, imagine code elsewhere that calls rss_log.target_simply_deleted? in a different context -- say, a script which culls stale orphan logs.  That method name would make no sense in that context.  But if we rename it orphan?, yes it requires a bit more thought inside RssLog#detail, but you can figure out what it does easily enough, and you would quickly understand why you were using it there.  *And* it would work well in other contexts like the orphan log culler.

Oh I got it!  Another way to look at it that might make more sense is that you were naming the method according to what the *caller* wants it to do, not according to what the *method* is doing.  Does that make sense?  The caller wants a method that tells if if the target was simply deleted.  That happens to work out to checking if the log is an orphan.  The method doesn't have the context, all it knows that it's doing is checking if its an orphan.  "Simply deleted" implies that there are alternatives such as complicatedly deleted, or "deleted and something else", or ... who knows?  It's really kind of vague on what "simply" means!  "Orphan?" on the other hand is easily understood out of context.  It describes perfectly what the *method* is doing, not what the first caller happened to need it for.

JoeCohen:
When I first named target_simply_deleted? I intended to differentiate a
target that had been both merged and deleted, giving the latter a different
message to display.

PR #843, opened 2022-01-31
Fix small bug that was creating a bunch of "ghost" rss_logs.

pellaea:
Name#save_with_log would log the action before checking that it could actually save the changes (or in this case, the new record).


JoeCohen (approved):
Super work tracking this down.
Should it be merged and deployed ASAP to avoid any more ghost logs?

pellaea:
Yeah, I see no reason not to.  I just created another PR to keep an eye on
this and two other cases and auto-correct them, too.

pellaea:
Thanks, Joe.

PR #846, opened 2022-02-01
Changed RssLog#detail to display better messages for destroyed and me…

pellaea:
…rged objects.


pellaea:
In order to test this properly you might need to run script/check_rss_logs first.  That script clears target_id for orphans.  That has some effect on RssLog behavior.

JoeCohen:
Nothing bad jumps out at me. And it runs fine locally.
Can we merge and deploy it?

pellaea:
Will do.  Thanks for looking at it, Joe.

PR #848, opened 2022-02-03
Update RuboCop todo

JoeCohen:
This PR simply updates the list of current repressed RuboCop offenses.

I'd like to again chip away at those offenses, so I'd like a current list.


JoeCohen:
Merging because the changes affect nothing (other than the list itself).
(There are currently 2652 repressed offenses.)

PR #849, opened 2022-02-04
Fix RuboCop GuardClause offenses

JoeCohen:
This PR:
- Fixes all existing Style/GuardClause offenses and re-enables that cop.
Because there were some many of those offenses, `.rubocop-todo.yml` had hid them and disabled the cop.
- Fixes offenses that were created, or exposed, by  fixing the Style/GuardClause.
- Delivers https://www.pivotaltracker.com/story/show/181143205 which
Is part of https://www.pivotaltracker.com/epic/show/3884358

GuardClause offenses generally are not challenging (contrast Metrics offenses) but not auto-correctable by RuboCop, so those changes have been made manually. 

No manual test is suggested.


pellaea:
Nice work, Joe!

nimmolo:
Nice work Joe!

.rubocop_todo.yml
```diff
@@ -104,14 +104,15 @@ Lint/UnreachableLoop:
...
-# Offense count: 9
+# Offense count: 10
 # Configuration parameters: CheckForMethodsWithNoSideEffects.
 Lint/Void:
   Exclude:
     - 'app/classes/mo_paginator.rb'
     - 'app/classes/query/modules/high_level_queries.rb'
     - 'app/classes/query/modules/nested_queries.rb'
     - 'app/classes/query/modules/sequence_operators.rb'
+    - 'app/controllers/description_controller_helpers.rb'
```
JoeCohen, line 115: This represents a new offense I'd like to get rid of. 
RuboCop says
>Line #658 – Warning: Lint/Void: Operator - used in void context.
    `new_groups - old_groups.each do |group`

I think this is a bug in RuboCop, as it's not flagging similar Line 662:
`old_groups - new_groups.each do |group|`

app/controllers/translation_controller.rb
```diff
@@ -74,15 +74,15 @@ def get_language_and_authorize_user
...
+    raise(:edit_translations_login_required.t) unless @user
+
+    unless @user.is_successful_contributor?
       raise(:unsuccessful_contributor_warning.t)
-    elsif lang.official && !reviewer?
-      raise(:edit_translations_reviewer_required.t)
     end
+    return unless lang.official && !reviewer?
+
+    raise(:edit_translations_reviewer_required.t)
```
JoeCohen, line 85: Is this right? 
(I actually don't understand what `raise` does. Does it interrupt execution and return to one level up?)

app/models/name/change.rb
```diff
@@ -19,9 +19,9 @@ def change_text_name(in_text_name, in_author, in_rank, save_parents = false)
       parents = Name.find_or_create_name_and_parents(parse.parent_name)
       if parents.last.nil?
         raise(:runtime_unable_to_create_name.t(name: parse.parent_name))
-      elsif save_parents
-        parents.each { |n| n.save if n&.new_record? }
       end
+
+      parents.each { |n| n.save if n&.new_record? } if save_parents
```
JoeCohen, line 24: Does this work? (Same issue with `raise`)

JoeCohen (commented):
There are just a few areas where I have questions.
There's a lot here, and it's an imposition to ask anyone to review the whole thing.
We can discuss that in the tech meeting Sat.

app/controllers/translation_controller.rb
pellaea: Raise interrupts all the way to the top and exits the program, unless it is caught somewhere above.  That's what rescue does: it catches some or all raised exceptions.  I don't remember where these exceptions are being caught off the top of my head.

app/models/name/change.rb
pellaea: I suspect so.  Name should be well-tested, so I would trust this if it passes test.

pellaea:
I think I'm going to have to see more context around those two proposed changes to convince myself that they are the same, logically speaking.  In a minute...

.rubocop_todo.yml
JoeCohen: I think this exposes a bug in our code (in addition to a bug in RuboCop).
Here is the whole method. (app/controllers/description_controller_helpers.rb lines 657-666).
```ruby
  # Throw up some flash notices to reassure user that we did in fact make the
  # changes they wanted us to make.
  def flash_description_changes(old_groups, new_groups, type)
    new_groups - old_groups.each do |group|
      name = group_name(group)
      flash_notice(:"runtime_description_added_#{type}".t(name: name))
    end
    old_groups - new_groups.each do |group|
      name = group_name(group)
      flash_notice(:"runtime_description_removed_#{type}".t(name: name))
    end
  end
```
I think we need to parenthesize `new_groups - old_groups` (and later, `old_groups - new_groups`).
As currently written, I think the block
```ruby
    new_groups - old_groups.each do |group|
      name = group_name(group)
      flash_notice(:"runtime_description_added_#{type}".t(name: name))
    end
```
will flash each member of `old_groups`, and return `new_groups - old_groups`.  See example below.
But I suspect that what was intended was to flash the members of ` new_groups - old_groups`.
Example:
```ruby
> a_e = %w[a b c d e]
 => ["a", "b", "c", "d", "e"] 
> d_e = %w[d e]
 => ["d", "e"] 
> a_e - d_e.each { |l| puts l}
d
e
 => ["a", "b", "c"] 
```
(RuboCop is buggy in saying that `-` is invalid in this context.)

app/controllers/translation_controller.rb
JoeCohen: Thanks @pellaea. The changes should work fine.

app/models/name/change.rb
JoeCohen: Thanks again @pellaea. The code will work. (I was unsure about this because I didn't understand `raise`, which you have now explained.)

.rubocop_todo.yml
pellaea: You've convinced me!  I'm sure this is something I wrote... but I don't typically use the each construct.  I bet this was originally a for loop and it just got converted hastily.  Obviously not a very important piece of code since no one has complained!

JoeCohen:
Aha!
You probably wrote it without bugs, and RuboCop and/or I incorrectly turned
it from `for` to `each`.

pellaea:
Exactly!  Couldn't possibly have been my fault! :)  After all, I test
everything so thoroughly...

JoeCohen:
Yes. Before RuboCop and I auto"corrected" it, it was:
```
    for group in new_groups - old_groups
```

JoeCohen:
I lost three ten-thousandths of a percent of coverage because I added some parentheses?
The newly uncovered line is line 21 in app/models/name/change.rb.
I don't know how to cover it, so I plan to ignore the epsilon coverage decrease.
```ruby
10:   def change_text_name(in_text_name, in_author, in_rank, save_parents = false)
11:     in_str = Name.clean_incoming_string("#{in_text_name} #{in_author}")
12:     parse = Name.parse_name(in_str, rank: in_rank, deprecated: deprecated)
13:     if !parse || parse.rank != in_rank
14:       raise(:runtime_invalid_for_rank.t(rank: :"rank_#{in_rank.to_s.downcase}",
15:                                         name: in_str))
16:     end
17:     if parse.parent_name &&
18:        !Name.find_by_text_name(parse.parent_name)
19:       parents = Name.find_or_create_name_and_parents(parse.parent_name)
20:       if parents.last.nil?
21:         raise(:runtime_unable_to_create_name.t(name: parse.parent_name))
22:       elsif save_parents
23:         parents.each { |n| n.save if n&.new_record? }
24:       end
25:     end
26:     self.attributes = parse.params
27:   end
```

pellaea:
I can live with 0.0003% decrease in coverage!

And that was really sweet that you tracked down the origin of that problem
with new_groups - old_groups.each.  I do find it reassuring that I didn't
make that mistake.  Although, it does need to be pointed out that it
doesn't really clear my muddied name: if I had written proper tests the
botched autocorrect would have been caught immediately.  <head hanging in
shame>

app/controllers/translation_controller.rb
pellaea: I wonder if the following would be easier to read?  I like that it ensures all of the criteria do the same thing: they are all about cases which raise exceptions.

  def validate_language_and_user(locale, lang)
    raise(:edit_translations_bad_locale.t(locale: locale)) unless lang
    raise(:edit_translations_login_required.t) unless @user
    raise(:unsuccessful_contributor_warning.t) \
      unless @user.is_successful_contributor?
    raise(:edit_translations_reviewer_required.t) \
      if lang.official && !reviewer?
  end

pellaea:
Except for that one minor point, everything looks really good.  Thanks for taking this on, Joe!  The -w flag on 'git diff' really made this a cinch to review. :)

app/controllers/translation_controller.rb
JoeCohen: I agree it would be easier to read if each line started consistently. 
I did it this way because of line-length issues. 
I'll see if continuing the line solves the problem.

pellaea:
I checked earlier, and rubocop seems happy with this version.

JoeCohen:
Strange. It tells me:
>Line #79 – Convention: Style/IfUnlessModifier: Modifier form of unless makes the line too long.
    raise(:unsuccessful_contributor_warning.t) unless @user.is_successful_contributor?
Line #79 – Convention: Layout/LineLength: Line is too long. [86/80]
    raise(:unsuccessful_contributor_warning.t) unless @user.is_successful_contributor?

I think continuing the line will fix that, but I need to check something.

JoeCohen:
It works.
And thanks for taking the time to slog through all the changes. I hope they didn't put you to sleep. :-) I also had reviewed the changes one last time and @mo-nathan's suggestion really helped.

PR #851, opened 2022-02-07
Whitelist bots

JoeCohen:
This PR completely[1] kicks off[2] the site any bot[3] that is not whitelisted.  The whitelisted bots are:
- Googlebot
- Bingbot

This is a step toward realizing the objective -- discussed in MO Technical meetings -- of improving website response by limiting site access to whitelisted robots and logged-in users.

The PR delivers [Pivotal #181166642](https://www.pivotaltracker.com/story/show/181166642).

No manual test is suggested.[4] 

The PR also has a very small amount of incidental RuboCopping.

---

[1] Absolutely no actions or pages allowed.
[2] Returns html status code 403 Forbidden.
[3] bot == user_agent that is recognized as a bot by the `browser` gem.
[4] There might be a way to manually test with `curl`, but I don't have the time to figure it out. If you're interested and can specify the user_agent in curl, then GET `http://localhost:3000/observer/intro` with user_agent `Baidu`.


app/classes/robots.rb
```diff
@@ -1,18 +1,24 @@
...
-      return false if args[:ua].downcase.include?("yandex")
+    # Is the robot authorized to be on the site?
+    def authorized?(user_agent)
+      # Googlebot not followed by a hyphen excludes Googlebot-Image, etc.
+      /Googlebot(?!-)|bingbot/.match?(user_agent)
+    end
+
+    def action_allowed?(args)
+      populate_allowed_robot_actions unless defined?(@allowed_robot_actions)
+      return true if args[:controller].start_with?("api")
```
JoeCohen, line 14: Allows anonymous API use. 
Do we want to require a key?

pellaea:
re: class variables -- I admit I never remember the difference since I use
them so frequently.  I perfectly happy to use whatever rubocop, etc. prefer.

re: anonymous API use -- We should check if robots ever use the API.  I
don't think so.  If they do, let's revisit this.  If we were to change this
we might break an unknown number of API users out there.  That's a whole
"thing".  We'd have to figure out how to alert them of the up-coming
change, but without an API key we have no way to link the use to an account
or email address, so... I don't know, I guess we might have to institute
some sort of temporary thing where we check for anonymous API use and
somehow send them a message in the response... but of course they aren't
going to know to look for a message in the response... so... I have no idea
how to do it!  Site banner, I suppose?  As I said, it's a big thing.  I'd
rather continue to avoid it as long as possible! :)

app/classes/robots.rb
nimmolo: This is a question that seems to relate to the next GraphQL PR, too. 

Somehow my initial API test `HttpRequestTest` was getting identified as a robot, and I could only get it to run by modifying line 14 above with something like `.start_with?("api", "graphql")`. 

On the other hand, per consultation with Jason, we may want to set up the graphql api to require an API key regardless - and the SvelteKit front end would use its own API key. 

I don't know if the last part is strictly necessary, for example if the eventual front end is on another port of the same URL, as it is when I run the front end locally.

pellaea:
Note that you will still need to let /graphql/user/login through without a
token/key, Nimmo.  For the old API this is not an issue because you create
the API key in person on-line first, not through the API.

nimmolo:
Message understood - although for the record there won't be any routes like that. 

Everything just goes as JSON to /graphql/. The user and token are parsed from the http request header, then sent as part of a "context" object to the request parsing classes in /app/graphql/. Those classes determine whether the user in "context" has a right to do anything. 

If we were to require an API key for graphQL requesta, it would need to be part of the http request header for every request. Since API keys are per-user, it obviously wouldn't map to the user in therequest - the "owner" of that api key would be SvelteKit.

JoeCohen:
@pellaea I agree that we should leave the API stuff as is, at least for now. I don't know how much the API is used without a key. If that becomes a problem, we can deal with it in the future.

app/classes/robots.rb
JoeCohen: Thanks @nimmolo. 
I'd like to look into this, in particular to understand it thought that request was coming from a robot.
Can you tell me how to replicate the issue?

pellaea:
Thanks for the clarification, Nimmo.  I keep forgetting how different
GraphQL is.

re: SvelteKit having its own API key that it would always use -- Ah, right,
of course, that makes sense.  Then yes we can require an API key to use
GraphQL for all requests. ... Which is good, because as you point out,
there's no way to know what request is being made without parsing the
context!

app/classes/robots.rb
nimmolo: Hi Joe, 
I can only reproduce it on my PR branch, because it's a graphql controller test that is blocked as a robot. I can show you how over there.

But - on the other hand maybe i can resolve it another way - like giving the test request an arbitrary user agent string? Is that how we detecting robots in the first place?

JoeCohen:
@nimmolo:
We use the browser gem to detect robots.
The relevant code is in ApplicationController#kick_out_robots (line 180):
return true unless browser.bot?

The rules that browser uses to detect bots are here:
https://github.com/fnando/browser#bots.
If you have an empty user_agent, it thinks you are a bot.
If that's the problem, then -- as you suggest -- simply create a user agent
string that browser won't regard as a robot.

If that doesn't work, please let me know so we can figure it out.

nimmolo:
Thanks Joe! I'll do that, and if Rails will let me set and fake that header, I bet that clears it up.

nimmolo:
@JoeCohen That worked. 

I'd like to figure out a way to set that user agent by default for all the graphql tests, [possibly using a Capybara "driver".](https://stackoverflow.com/questions/39096779/set-custom-user-agent-on-rails-testing)

JoeCohen:
@nimmolo: Can you use [setup](https://api.rubyonrails.org/v5.2.4.4/) to do that? 
At the top of the testing class:
```ruby
def setup
  @request.user_agent = "Whatever you're using as the UA string"
end
```

JoeCohen:
Merged and deployed to webserver

pellaea:
Thanks!

PR #857, opened 2022-02-15
GraphQL - Login, field visibility and authorization

nimmolo:
This PR will implement visibility and authorization checks on query and mutation fields, and test them.

Also, User Login and token generation via the API.


nimmolo:
I believe this is ready for review. 
There's still only one db model involved at this point, Users. 

This PR:
- Introduces field-level authorization for viewing object properties (i.e. columns in the db), and tests them
- Adds and tests UserLogin, tests the token, and the authorization it provides for users' own fields vs public fields
- Breaks out helpers to simplify the **two main routes of testing this API**
  - full **http_requests** from a front end
  - direct **execution of graphql** in the backend
- Breaks out all Queries into a third file, `graphql_queries.rb`, included in both of the above helpers. Queries are basically JSON objects that must follow a specific syntax paralleling the query definitions in `/graphql`. So, to facilitate this, and use it going forward we'd:
  - manually update the GraphQL schema as it's developed, with `rails graphql:dump_schema` to `public/schema.graphql`
  - copy/paste query syntax needed in a test from the dump into the `graphql_queries.rb` file
Although this reduces query definition duplication in individual tests, it still introduces a "test-side" duplication of the schema that I don't like. 

The schema itself should ideally serve as a single source of truth for graphql definitions. I'm researching how to use the schema directly in tests and will introduce that (and a test that the schema is up-to-date), in the next PR. If you prefer, though, i'll try to land it on this one.

JoeCohen:
Suggestion:
In `app/controllers/graphql_controller.rb` and other classes and modules, 
add a comment immediately before the start of the class module.  See example below.
This avoids adding to the concealed backlog of [Style/Documentation](https://docs.rubocop.org/rubocop/1.25/cops_style.html#styledocumentation) offenses. 
```ruby
# Parses an incoming request header to build a Context hash (authentication)
# and forwards that to the graphql Query, w/ Variables and Operation Name.
# Possible operation names: Query, Mutation, Union etc
class GraphqlController < ApplicationController
  # Note 22/02/10 - Nimmo
  # This controller is short but important.
```

JoeCohen:
FYI. 
The developers should probably have a conversation about [nested module/class definitions instead of compact style](https://docs.rubocop.org/rubocop/1.25/cops_style.html#styleclassandmodulechildren). Compact style saves space, but does have some risks. https://rubystyle.guide/#namespace-definition
```ruby
app/graphql/inputs/user/login.rb - 2 offenses
Line #3 – Convention: Style/ClassAndModuleChildren: Use nested module/class definitions instead of compact style.
module Inputs::User
Line #4 – Convention: Style/Documentation: Missing top-level class documentation comment.
  class Login < Inputs::BaseInputObject
```

app/graphql/inputs/user/login.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module Inputs::User
+  class Login < Inputs::BaseInputObject
+    description "Credentials necessary for user login"
+    # the name is usually inferred by class name but can be overwritten
+    graphql_name "UserLoginInput"
+
+    argument :login, String, required: true
+    argument :password, String, required: true
+    argument :remember_me, Boolean, required: false
```
JoeCohen, line 11: Do you want/does this need a default?

nimmolo: Still researching this. I believe it defaults to null in any case... is there a more desirable default? `required: nullable` is a possibility. 

The [argument docs](https://graphql-ruby.org/fields/arguments.html) lay out the options, but i haven't found a login input example out there with default values yet.

nimmolo:
For module nesting, I'm happy to switch. I think here in some cases, we evade the risks cited in your linked article (thanks for that, it's key!) because the inner module will stay empty. It's only there to facilitate folder organization and autoloading. 

So for example currently in `graphql/inputs/user/login.rb`
```ruby
module Inputs::User
  class Login < Inputs::BaseInputObject
    description "Credentials necessary for user login"
    # the name is usually inferred by class name but can be overwritten
    graphql_name "UserLoginInput"

    argument :login, String, required: true
    argument :password, String, required: true
    argument :remember_me, Boolean, required: false
  end
end
```
...the `module User` doesn't add any code to its descendants. 
But i like the idea of rewriting it, just for style consistency. Is this correct?
```ruby
module Inputs
  module User
    class Login < Inputs::BaseInputObject
      description "Credentials necessary for user login"
      # the name is usually inferred by class name but can be overwritten
      graphql_name "UserLoginInput"

      argument :login, String, required: true
      argument :password, String, required: true
      argument :remember_me, Boolean, required: false
    end
  end
end
```

JoeCohen:
1. I think that's right.
2. I do want to learn more about GraphQL, and actually started looking at https://graphql.org/ yesterday. But it's going to be about 6 or 7 weeks before I can dig into it.
3. I'm out of steam for today, and don't think I can contribute anything more on this PR. So unless other people have comments, I thin you should move ahead.

app/graphql/mutations/user/login.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        return {} unless user
+
+        raise(GraphQL::ExecutionError.new("User not verified")) unless
+          user.verified
+
+        now = Time.zone.now
+        args = {
+          last_login: now,
+          updated_at: now
```
pellaea, line 26: updated_at should automagically be updated for you.

app/graphql/types/base_field.rb
```diff
@@ -3,5 +3,45 @@
...
+    #   super && (@only_show_owner ? owner_or_admin(obj, ctx) : true)
+    # end
+
+    # Field #authorized? methods are called before resolving a field
+    # Note this effectively blocks resolution of a query field, eg user.email
+    # But we need to enable unauthorized search for user by email in login
+    def authorized?(obj, args, ctx)
+      # if `require_admin:` was given, require current user to be an admin
+      super && (@require_admin ? ctx[:current_user]&.admin? : true)
+      super && (@require_owner ? owner_or_admin(obj, ctx) : true)
```
pellaea, line 44: The first line is basically ignored in this.  I think you mean:

  super &&
    (@require_admin ? ctx[:current_user]&.admin? : true) &&
    (@require_owner ? owner_or_admin(obj, ctx) : true)

Or maybe this is easier to read?

  return false unless super
  return false if @require_admin && !ctx[:current_user]&.admin?
  return false if @require_owner && !owner_or_admin(obj, ctx)
  return true

app/graphql/inputs/user/login.rb
```diff
@@ -0,0 +1,16 @@
...
+module Inputs
+  module User
+    class Login < Inputs::BaseInputObject
+      description "Credentials necessary for user login"
+      # the name is usually inferred by class name but can be overwritten
+      graphql_name "UserLoginInput"
+
+      argument :login, String, required: true
+      argument :password, String, required: true
+      argument :remember_me, Boolean, required: false
```
pellaea, line 13: I'm having trouble figuring out what "remember_me" does in this context!

In the old UI, it causes rails to set a cookie on the client browser.  If that cookie is present it automatically logs the user in if session_user is cleared.  That way, if you come back tomorrow after the session has expired, it will log you back in without you even noticing.

How would this look from a svelte perspective?  Would you have Svelte store the token in a cookie after logging in?  Then when you return to the site the next day and it doesn't have a token, it checks for a token in the "remember me" cookie and checks to see if that's valid, if so, you're automatically essentially logged in.  If that's right, then you would never actually send "remember_me" to the server at all.  It's all handled by svelte, and graphql doesn't need to know anything about it.

But if... something else... maybe graphql does need to know.  But as I see it right now, remember_me is not actually used anywhere.

app/graphql/types/base_field.rb
nimmolo: I went for your second option - because it's the only one I understand, and that's including the one I wrote ;)

pellaea:
This is pretty slick.  I couldn't follow very well by looking at the changes, but when I finally just downloaded your PR and browsed through all the files in the app/graphql directory tree... nice, very simple and powerful.  I like.

pellaea:
This is pretty slick.  I couldn't follow very well by looking at the changes, but when I finally just downloaded your PR and browsed through all the files in the app/graphql directory tree... nice, very simple and powerful.  I like.

app/graphql/inputs/user/login.rb
nimmolo: You're right, it doesn't do anything right now. And yes, Sveltekit stores the token in a local storage. 

I was thinking Rails would encode that param in the token as a flag to the frontend to either delete the token after x hours if the user forgets to log out (as in the case of a public or shared computer) -- or, if "remember me" is true, just keep using the same one. I believe the token is good for a while, unless we set the duration shorter. 

Am I still thinking garbled about tokens? Highly likely.... hopefully i'm starting to get it though...

nimmolo:
If you've got it downloaded, a good way to kick the tires is to start up Rails, go to `http://localhost:3000/graphiql` and try a query. I'm going to try it right now though first! If it's still OK after my commits today i'll give you an example query to play with.

nimmolo:
Whew still good.

Try pasting this in the left column of the `graphiql` screen, and then click the "play" button (selecting one or the other query):
```graphql
mutation userLogin {
  userLogin(input: { login: "nathan", password: "password" }) {
    user {
      id
      name
      email
    }
    token
  }
}

query user {
  user(id: 1) {
    id
    name
    email
  }
}
```
You can of course change any of those params.

pellaea:
Sounds like you understand it better than me.  I'm convinced.  It is
confusing when multiple layers and systems interact like this.

pellaea:
I'll give it a shot soon as the dev VM fires up...

So, do I understand correctly that you could essentially flesh out the
read-only part of the interface pretty rapidly at this point?  It looks
like you have all the basic underlying functionality worked out.

I realize mutations will require a bit more work.  It's a bit more involved
than simply listing column types and taking a bow. :)

What about the associations?  Is there more infrastructure work there?

pellaea:
I don't seem to be able to access the owner-only fields.  It's like it
doesn't realize I've logged in.  Is there some way to add the token I got
back from the userLogin mutation  to the user query?

nimmolo:
Re: owner only fields... GraphiQL has, on the bottom left, a REQUEST HEADERS editor, so you can paste in the token you get from a login mutation right there in GraphiQL, and use it to authorize a request to see Jason Hollinger's email address. Muhahahahahaa. 
(Oddly, it's the same as Rolf Singer's - "webmaster@mushroomobserver.org" Are they the same person?!)
Screen Shot 2022-02-22 at 8 05 00 PM

Yes it does seem that the read-only part will be quicker to flesh out. This foundation is already waay better than what i'd been playing with before getting review from you and Joe. 

Associations get added directly to the type, as you probably noticed in the long commented-out section at the bottom of `types/models/user.rb`. That's where the batch/eager loading comes in to play, we'll need to pay closer attention to that at some point.

nimmolo:
Using an auth header will also allow you to get something from checking the `visitor` query. Be sure to ask for user fields in the return... any user field should work.
```graphql
query visitor {
  visitor {
    id
    login
    name
    email
  }
}
```

Note for the build-out: Unions are a separate Type from Associations, in GraphQL.

app/graphql/mutations/user/login.rb
nimmolo: ok. Do i understand right, that if I just do `User.update({ last_login: now })` this is all I need to do, and will update `updated_at`? If so I will delete that second arg.

pellaea:
Yes, that can be reduced to a single line!

  user.update(last_login: Time.zone.now)

pellaea:
Yes, that worked!  Thanks.  I would never have been able to do all that
without copying and pasting from your examples, but that's what svelte is
gonna be for, right? :)

nimmolo:
I know what you mean and I was kind of mystified by how to use graphiql at first. But all the available correct syntax for the current schema is available in the Documentation Explorer on the upper right corner of the graphiql screen.  (The "Documentation" this refers to is _our own_ -- whatever we've written in our Graphql type descriptions.)

Screen Shot 2022-02-22 at 9 05 39 PM
In this screenshot above, the words Query and Mutation are links to the available Queries and Mutations on MO, and their syntax.

Screen Shot 2022-02-22 at 9 10 36 PM
But even easier than that, you can just type slowly... graphiql's autocomplete suggests everything available, in correct context too, while you're typing. For example here, I'm naming a query `findJason`, and as soon as I tell graphiql it's a "user" query, it suggests params. The same suggestions are made for the return hash.

In the SvelteKit app, I do have to maintain TypeScript definitions of the "queries" and "mutations" i'm using in the front end. These get called by a GraphQL client JS library (@apollo/client)

One of the libraries (graphql-codegen) also generates its own copy of the schema by doing an "introspection", a special graphQL query that says "Give me the whole schema structure", that should definitely be disabled in production!

pellaea:
Good point, I did notice it autocompleting handily.  The autodocumentation
is slick, too.  Thanks.

nimmolo:
Updated userLogin to handle the "remember_me" arg properly.
Also adds a "remember_me" boolean to the token, with a "created_at" time. 
TBD how the front end will handle these.

nimmolo:
LOL, hold up.  Am I obligating the front end to decrypt the token to figure out `remember_me` and the `created_at` time? Hahahaha that ain't gon work... 

I should pull this all out and make "remember_me" be a param of the front end's local storage ONLY. Nothing to do with graphql, as @pellaea wisely intuited!!

pellaea:
Ha!

nimmolo:
Maybe it would be wise to pass the value of "remember_me" straight back to the front end via the return hash -- the form values might be cleared at this point, or otherwise inaccessible? Then the front end has everything it needs in one place to store the token, user, and remember preference locally...

pellaea:
That makes a lot of sense.  Come to think of it, I'm not sure how else it
could work.  These sorts of things mess with my mind.

nimmolo:
I believe this will work and is preferable. It keeps the submitted form data together (so the remember_me checkbox doesn't have to live outside the form element), and keeps it handy in the return hash, with all the other goodies that go into local storage. 

If that's backwards, i'll remove it later.

nimmolo:
Thank you for that very patient sanity check!

nimmolo:
Still altering test helpers. They weren't posting JSON, they were posting ruby hashes, and booleans/integers were converted to strings.

nimmolo:
Merged recent commits to master into this branch. I believe it's good to go.

pellaea:
Okay, I'll merge it.  Thanks, Nimmo!

PR #862, opened 2022-02-19
Fix Style/RedundantRegexpEscape offenses

JoeCohen:
Removes unnecessary escapes inside Regexp literals.
Shortens and improves regex readability.


pellaea:
Look good to me!

PR #864, opened 2022-02-20
Remove S3 code

JoeCohen:
This PR removes obsolete, unused code that was intended for access to images on AWS/dreamhost S3.

- Delivers https://www.pivotaltracker.com/story/show/181320270
- No manual test suggested


script/bash_images
```diff
@@ -103,11 +103,6 @@ function copy_file_to_server {
     path=${image_server_data[$server,path]}
     scp $image_root/$file $path/$file || die "Failed to transfer $file to $path."
     ;;
-  "s3")
-    cmd=$app_root/script/s3cp.rb
-    wait_for $cmd
-    $cmd $image_root/$file $server $file || die "Failed to transfer $file to $path."
-    ;;
```
pellaea, line 110: I see one additional "s3" case in this file for "read_server_directory".

pellaea:
Aside from that one thing it looks like you missed -- or maybe that was deliberate? -- this looks good.  I can't find any other references to "s3" anywhere in the code.  And I can't think of any reason that tearing any of this out will affect us at all.

script/bash_images
JoeCohen: Good catch, thanks. I'll take care of it. It wasn't deliberate

PR #865, opened 2022-02-21
Add tests to ensure objects with rss_logs are logging destruction and…

pellaea:
… orphaning the log appropriately.


JoeCohen:
Looks good to me.
I'd say merger it. The worst that can happen is some funny log entries; there's no change in the underlying objects.

pellaea:
Thanks, Joe, done.

PR #866, opened 2022-02-21
Fix column null migration

mo-nathan:
Provide a default value when making created_at column non-null and then populate created_at based on existing updated_at data.


JoeCohen:
Thanks Nathan. I was just about to hit the Merge button.

PR #868, opened 2022-02-22
Add Bullet gem

pellaea:
This is the bare miminum activation of the Bullet gem.  I have turned on only the n_plus_one_query check, and only in test mode.  Furthermore, it only seems to operate in integration tests.  But it's better than nothing.

I have it configured to raise an exception whenever it detects that eager loading would help.  I was able to fix all but three cases:

1. Bullet wants us to eager load details of the targets of comments in comment indexes, but I can't find any way to eager load polymorphic associations.  I have disabled Bullet for all of these actions.

2. We're supposed to eager load synonym names, but I can't get it to do so, and I'm not sure why.  This affects only edit_species_list, so I just disabled Bullet for that action.

3. Similarly, we're supposed to eager load something project-related in edit_observation, but again I couldn't figure out how to, so I just disabled Bullet for that action.

I have probably introduced a bunch of rubocop violations.  I'll try to fix them as they're uncovered.

This was way more work than it was probably worth.  But maybe it will catch actual useful issues someday.

It would probably be worthwhile running all the basic "list everything" indexes for every controller in an integration test just to make sure the most important eager loading is done correctly.  I'll do that in a future PR.  (Or more likely put it on my "I'll never get to this" todo list...)


pellaea:
The two codeclimate issues are both pre-existing and nontrivial to fix.  I'm leaving them for later.  I feel like I've fixed enough pre-existing rubocop issues already.  It's ready for review if anyone cares to check it out.  It's pretty dull stuff probably.

JoeCohen:
@pellaea: Thanks for adding the gem, getting rid of n+1's, and RuboCoppiing so many things that weren't created by this PR.
I will try to finish looking at this PR (& @nimmolo's today. (I have some other neglected chores I have to attend to this a.m.)
**Question**: Is there any particular reason you choose to not activate bullet in dev mode?

pellaea:
Yeah, just one step at a time!  Please feel free to.  There are a number of
ways to do so -- it can have it create a page footer if there's a problem,
popup an alert, log to various files, etc.  I personally hardly ever run
the server in dev mode anymore, so it wasn't a big priority for me.  I
figure if I ever get a bee in my bonnet about performance, I can turn it on
locally and play around.  But I didn't feel the need to subject everyone to
it right now.  I leave it up to you!

JoeCohen:
To integrate `bullet` with `minitest` the solution at the following link worked for me:
https://github.com/flyerhzm/bullet/issues/442

(In order to avoid making `test/test_helper.rb` any longer, I:
- put the new module in a `test/bullet_helper.rb`, and 
- required that file in `test/test_helper.rb`)

Running the tests then yielded 124 errors. I haven't figured out how many are unique.
Because of the number of errors, I didn't want to push the change.

pellaea:
Brilliant, Joe!  Maybe we can add that as an issue/story and punt it to the
future.  Fixing these things can be hard, tedious work, and the vast
majority are niggling things that aren't going to have any performance
benefit at all.

Example: removing an observation from a species list.  It has to check if
the observation is in the species list, but the species list is itself the
result of a query.  At the time we load the species list we have no idea if
we will be removing an observation from one or more of them.  In fact we
can never remove an observation from more than one at a time.  But Bullet
sees us accessing the species_list.observations association from the result
of a species_list query, and concludes that eager-loading would help.  It's
wrong.  But what are our options for silencing this false positive?  Have
edit_observation (the action that performs the removal) eager load
observations for all your species lists anyway?  That's moronic.  It will
almost never need any of those eager-loaded observations.  Tell Bullet to
ignore the species_list.observations association everywhere -- again, this
makes no sense, because it really could make a big difference someday for
another unrelated action.  Tell Bullet to ignore the edit_observation
action completely (I think this is what I elected to do) -- but this would
mask actual legitimate missing eager loading problems.  Or massive
refactor?  For example, move the remove_observation function to a small
dedicated action.  Yes, that's a great idea, and I regret my laziness for
lumping all types of mutations on observations into a single massive
action.  But wow, that's a lot of work.  And again, it won't actually
improve performance, which is what this is all about!  Actually, the
refactor will turn a single request into two requests, so it will decrease
performance if anything.  Is this really what we want to spend our time
doing?

I'm considering instead that we might want to approach this whole thing
differently.  Turn on Bullet only for one or two test suites.  For example,
write a bullet_integration_test.rb and have it test all the important
indexes and other queries (site_data) that do massive database queries.
Then leave Bullet turned off elsewhere.

Thoughts?

JoeCohen:
For the reasons you suggest, I like that strategy:
- Turn on `bullet` a bit at a time so that we can address the n+1's in small batches.
- The ultimate goal is to turn it on everywhere, including dev mode.

pellaea:
Thanks for simplifying update_project.  Probably wasn't hard to do, but man
was I beat!

nimmolo:
Reading along with interest. What I learn from Jason’s description of poking into these corners of the controllers is also going to be applicable to the setup of graphql queries.

My perspective is probably obvious but I guess it’s worth stating - If time is a limited resource I’d vote for learning graphql. The buildout of that API involves thinking about all the stuff you’re confronting in the controllers: query optimization, authorization, caching - and it's a chunk of work.

To me, work on the controllers beyond the minimum feels like a blind alley (that I spent hundreds of hours in). Getting to refactored controllers and views that are Rails 6-ready would not get the web app to what I’d call a contemporary standard of UX and maintainability, if it were even realistic. It will be enough just getting models and classes ready for 6. (5.2 is EOL this summer)

I’ve kind of turned the page and maybe not everybody’s there yet. I notice out there on the web that a separate front-end on a graphql-rails backend is an achievable goal, in a reasonable amount of time and code - what do you guys think?

I feel it’s a better way to get MO on a versatile technical footing going forward. And if it’s done carefully, and people can help building it out (review is already very helpful), development will move more quickly. And at the end, we’ll all be familiar with how graphql works with Rails.

I hope all this reads as enthusiasm rather than vehemence. I am happy to keep writing the API by myself, but if anyone has an appetite for scrutinizing query efficiency, i’d love company in the graphql directory. It’s not that hard to learn, and I feel building this will get us to a better app more efficiently.

Nimmo

pellaea:
Thanks for your reflections on this, Nimmo.  I'm hearing a little
frustration underneath that you aren't being more supported in your
efforts.  Or maybe not unsupported, but that you could use some more hands
to help the process go faster.

I'd like specifically to acknowledge your comment: "work on the controllers
beyond the minimum feels like a blind alley".  I think you're right, both
in that statement, and that I at least am not quite there.  I'm still at:
"let's let Nimmo demo graphql and svelte and see how it works in practice,
and then discuss if this is where we want to focus all of our development
energy".  So, if your goal is to persuade us to lend a hand, I would
recommend focusing on getting a demo up and functional.  And I think that's
where you are.

(And this message comes as a timely reminder that you're waiting for a
review of your latest PR which is an important step toward that end.  Thank
you!  I'll try to get to that as soon as I can today.)

I apologize for tending to be overly defensive, but I will offer this one
defense of what I just did re: Bullet -- I thought it was going to be
trivial.  It was very much not what I thought it would be.  But I needed to
put in the (only one) night of hard work to understand the problems
involved.

My role at MO is 1) to provide support for other developers. 2) keep the
lights on, 3) occasionally when the mood strikes take on trivial PRs that
will only take an hour or two.  I am not interested any more in diving into
long term projects.  In practice I do not have the capacity for a variety
of reasons.  (Look at the last 2-3 nontrivial PRs I took on: they are still
sitting there, now over a year old, and with no plans to do anything with
them.  If it takes more than a night, I will not get it done.  I'm through
fighting it, I just accept it.)

I totally support what you're doing.  I can empathize if you feel some
frustration at not getting more hands-on help.  But that has definitely
been the nature of working for MO.  Even so, I think real meaningful
progress can and is being made.  And I think what you're doing especially
is a perfect example of that.  You are essentially completely rewriting the
interface of MO, but you're doing it piecemeal in small bites that don't
require tearing everything down or starting over from scratch (neither of
which ever amounts to anything in my experience, even when I had the
boundless energy of youth).  That is exactly what MO needs.  As you get
more done, and can demonstrate the power of graphql + svelte, I think
you'll find increasing support.  Please be patient!  This is the most
challenging time: you're starting to tire, but it hasn't started to catch
on yet.  It will, and then synergy will hopefully reinvigorate and
accelerate progress.

My two cents...

-Jason

JoeCohen:
I will probably ignore this, but just in case.
The thing that's making Brakeman unhappy is
```ruby
projects_member.map(&:id).join(",")
```

pellaea:
Are you talking about the SQL injection?  That was definitely pre-existing,
and there are probably dozens of cases just like it throughout the code.
It is actually safe, because id can only be an integer, so it is safe to
put off dealing with it.  The solution is probably to figure out how to
rewrite the complicated query using arel, but there is not-insignificant
chance of breaking or creating more problems by fixing it.

That was just my take, and why I wanted to leave it for another day and
another PR.

nimmolo:
:[]  -  I responded via email and didn't realize this was a Github thread! Sorry for diverting attention from the PR. 

Appreciate the response, Jason! Actually, where i was coming from was my perception that this PR signified the start of a new initiative to overhaul the controllers. Considering the work you had put into it (you get a lot more done in a day than i do!) it looked like the grand opening a big new trench into the mountain of technical debt, and i just wanted to say "guys, if you're going to dig that much, i smell more truffles over here."

Sorry I missed the meeting - that would have cleared it up.

PR #871, opened 2022-02-23
Fix test_observations_by_known_user

JoeCohen:
This test was broken by the merger of PR #858
(Post merger, users must login in order to GET `observations_by_user`.)


pellaea:
Thanks, Joe.

PR #873, opened 2022-02-23
Fixed minor eager-loading problem in show_observation.

pellaea:
This popped up during the flurry of PR merges.


JoeCohen:
Thanks Jason.

PR #877, opened 2022-02-25
redirect_anonymous_users after autologin

JoeCohen:
This PR changes the order in which the `redirect_anonymous_users` and  `autologin` filters are applied.
That's the correct order. And it might prevent an issue that @pellaea ran into. 
See https://groups.google.com/g/mo-developers/c/tSg60WpAKOY/m/lNEZJ6uYAgAJ?utm_medium=email&utm_source=footer
It also
- Moves `redirect_anonymous_users` after  `autologin` in the file
- Deletes the now-unneeded method `verified_user_logged_in?`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -106,11 +106,11 @@ class ApplicationController < ActionController::Base
   around_action :catch_errors_and_log_request_stats
   before_action :kick_out_excessive_traffic
   before_action :kick_out_robots
-  before_action :redirect_anonymous_users
   before_action :create_view_instance_variable
   before_action :verify_authenticity_token
   before_action :fix_bad_domains
   before_action :autologin
+  before_action :redirect_anonymous_users
   before_action :set_locale
   before_action :set_timezone
   before_action :refresh_translations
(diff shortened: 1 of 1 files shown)


JoeCohen:
I'm not going to worry about the trivial coverage decrease; it results from the deletion of one covered line rather than any newly uncovered lines.

pellaea:
Of course, I agree.  I'll look at this in a minute, but I can't imagine not
approving it.  Merge at your discretion.

pellaea:
Yep, looks fine.  I like that you can simplify it a little now.

JoeCohen:
Thanks Jason.
I wish I were certain about this being the cause of the CSRF-like hiccup.
If so, I would try to write a test to prevent reversion.
Maybe some odd combination of expired cookie, autologin, & save_location
being set to login, resulting in a loop that sends you to login after you
log in?

pellaea:
Yeah, I don't fully understand it, either.  Certainly, if the session
expires but autologin cookie is correct, it would give odd behavior.  Soon
as the session expires, this filter would punt you to the login page, but
you would find yourself logged in automatically on arrival at the login
page via the autologin cookie... but you wouldn't necessarily notice it
because you would be on the login page asking for login credentials.  Not
sure where CSRF comes in.

PR #884, opened 2022-03-01
Fixed translations of the Activity Log checkbox items.

pellaea:
The checkboxes at the top of the Activity Log had gotten all messed up.  It was doing

  :observation.classify.constantize.to_s.pluralized_title

This does:

  :observation -> "Observation" -> Observation -> "Observation" -> "Observations" (!!) 

This does not get translated.  Instead do:

  :observation.to_s.pluralize.upcase.to_sym.t

This does:

  :observation -> "observation" -> "observations" -> "OBSERVATIONS" -> :OBSERVATIONS -> "Observaciones"

Or better yet, let the translator deal with it!

  :"rss_one_#{:observation}"
    -> :rss_one_observation -> [:OBSERVATIONS] -> "Observaciones"

This latter approach requires a bunch of aliases in the translation file, but it let us override individual choices. For example, we can choose to change "Glossary Terms" to just "Glossary" and "Articles" to "News".  And it makes the code much more readable.

Note that these extra translations are all just aliases, so there is no need for translators to do anything.  But it does let them override our choices. What works best in English isn't necessarily what's best in Urdu. :)

### DIFF
--- a/app/extensions/string_extensions.rb
+++ b/app/extensions/string_extensions.rb
@@ -517,15 +517,6 @@ class String
     strip.squeeze(" ")
   end
 
-  # Uncamelizes and converts string to pluralized title. E.g.:
-  # "Observation"  => "Observations"
-  # "SpeciesList"  => "Species Lists"
-  # "GlossaryTerm" => "Glossary Terms"
-  # "good dog"     => "Good Dogs"
-  def pluralized_title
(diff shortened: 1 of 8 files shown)


pellaea:
I believe the minute decrease in coverage is from removing a covered method.  Everything I touched should be covered.

PR #886, opened 2022-03-08
Fix unstable language test

pellaea:
Capybara doesn't reset the session between tests.  This can be problematic if a test expects the session to start in the blank state.  All you need to do in such cases is explicitly tell Capybara to do so:

  Capybara.reset_sessions!

Nice and easy.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,12 @@ gem("mysql2")
 # Use sqlite3 as the database for Active Record
 # gem "sqlite3"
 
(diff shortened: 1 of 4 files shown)


JoeCohen:
Sorry to be missing all this work.
(This email caught my eye.)
Is there a place to put the reset so that it doesn't have to be called
individually in each Capybara test or Capybara test file? Maybe in the
Capybara helper file?

PR #890, opened 2022-03-08
Convert other SQL strings in app/models to Arel

nimmolo:
Ready for review, and advice for Abc violations


app/models/collection_number.rb
```diff
@@ -102,25 +107,68 @@ def log_destroy
   end
 
   def destroy_without_callbacks
```
pellaea, line 109: Actually, there might be a *much* better way to do this.  It's possible collection_number.delete does this for us!

app/models/collection_number.rb
```diff
@@ -102,25 +107,68 @@ def log_destroy
...
+
+  # UPDATE collection_numbers_observations cno,
+  #   herbarium_records_observations hro,
+  #   herbarium_records hr
+  # SET hr.accession_number = #{new_format_name}
+  #   WHERE cno.collection_number_id = #{id}
+  #   AND cno.observation_id = hro.observation_id
+  #   AND hro.herbarium_record_id = hr.id
+  #   AND hr.accession_number = #{old_format_name}
+  def arel_update_corresponding_herbarium_records(old_format_name)
```
pellaea, line 144: Yikes!  This method cannot be called very often.  It was kinda ugly to start with, but now it's insane and impossible to fix the ABC complexity.  What would you think about rewriting this entirely in ActiveRecord?

  observaton.herbarium_records.where(accession_number: old_format_name).update_all(accession_number: format_name)

(I'm trying to look up the API dock for update_all to make sure I'm remember the syntax right, but it won't load.  Sorry.  But update_all definitely doesn't fire callbacks, so it should be the right one to use, at least.)

nimmolo: Was just remembering this!
I don't believe what i have here can even work, since it is also an UpdateManager on a join.
`update_all` it is!

app/models/description.rb
```diff
@@ -400,25 +400,46 @@ def group_users(table)
...
+  private
+
+  # SELECT DISTINCT u.user_id FROM #{table} t, user_groups_users u
+  # WHERE t.#{type_tag}_id = #{id}
+  #   AND t.user_group_id = u.user_group_id
+  # ORDER BY u.user_id ASC
+  def arel_select_group_user_ids(table)
+    table = Arel::Table.new(table.to_sym)
+    ugu = Arel::Table.new(:user_groups_users)
+    table.join(ugu).on(table["#{type_tag}_id".to_sym].eq(id).
```
pellaea, line 427: "#{blah}".to_sym is equivalent to :"#{blah}"... any chance that would help with the ABC violation??

app/models/description.rb
```diff
@@ -400,25 +400,46 @@ def group_users(table)
...
+                          select_values(select_manager.to_sql).map(&:to_i)
   end
 
+  private
+
+  # SELECT DISTINCT u.user_id FROM #{table} t, user_groups_users u
+  # WHERE t.#{type_tag}_id = #{id}
+  #   AND t.user_group_id = u.user_group_id
+  # ORDER BY u.user_id ASC
+  def arel_select_group_user_ids(table)
```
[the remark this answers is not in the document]
pellaea: I don't see any way to make this shorter. Disable the stupid ABC cop, I guess.

app/models/donation.rb
```diff
@@ -5,11 +5,20 @@ class Donation < ApplicationRecord
...
+  # "SELECT who, SUM(amount) total, MAX(created_at) most_recent
+  # FROM donations WHERE anonymous = 0 and reviewed = 1
+  # GROUP BY who, email ORDER BY total DESC, most_recent DESC"
+  private_class_method def self.arel_select_donor_list
+    d = Donation.arel_table
+    d.where(d[:anonymous].eq(0).and(d[:reviewed].eq(1))).
+      project(d[:who], d[:amount].sum, d[:created_at].maximum).
+      group(d[:who], d[:email]).
+      order(d[:amount].sum.desc, d[:created_at].maximum.desc)
   end
```
pellaea, line 22: Nice, this looks good.

app/models/herbarium.rb
```diff
@@ -168,19 +168,31 @@ def merge_associated_records(src)
...
+    h = Herbarium.arel_table
+    c = Arel::Table.new(:herbaria_curators)
+    # user_id = 252
+
+    h.join(s).on(s[:herbarium_id].eq(h[:id])).join(c).on(
+      h[:id].eq(c[:herbarium_id]).and(c[:user_id].eq(user_id))
+    ).project(h[:name]).distinct.order(s[:updated_at]).take(100)
+  end
+
   def self.find_by_code_with_wildcards(str)
```
pellaea, line 196: Again, this looks good, disable the ABC cop.

pellaea:
Wow, that's hard work!  You must be ninja in Arel by now! :)

I have to go troubleshoot a lichenologist's external drive for them.
That's all I can do tonight.  I'm leaving off at image.rb...

app/models/collection_number.rb
nimmolo: Here's what API dock has for examples:
```
# Update all customers with the given attributes
[Customer](https://apidock.com/rails/Customer).[update_all](https://apidock.com/rails/ActiveRecord/Relation/update_all) wants_email: true

# Update all books with 'Rails' in their title
Book.where('title LIKE ?', '%Rails%').[update_all](https://apidock.com/rails/ActiveRecord/Relation/update_all)(author: 'David')

# Update all books that match conditions, but limit it to 5 ordered by date
Book.where('title LIKE ?', '%Rails%').order(:created_at).limit(5).[update_all](https://apidock.com/rails/ActiveRecord/Relation/update_all)(author: 'David')

# Update all invoices and set the number column to its id value.
Invoice.[update_all](https://apidock.com/rails/ActiveRecord/Relation/update_all)('number = id')

Model.where(:foo => 'bar').where(:attr => 1).update_all("author = 'David'")
```

app/models/collection_number.rb
nimmolo: Hahaha - I'm not an Arel ninja, i'm more like a jarhead. I have a hammer and everything seems like a nail...! Now that i'm realizing that, there will probably be a bunch of cases where my Arel could be simplified with AR, and i've started on a few. 

The good thing about the Arel of course is that it's pretty explicit, less magical, so it's more evident if it's building something moronically overcomplicated. (Hopefully!)

app/models/collection_number.rb
nimmolo: Completely rewrote it.

app/models/observation.rb
```diff
@@ -225,62 +229,56 @@ def self.refresh_content_filter_caches
...
-  def self.refresh_cached_column_fix_errors(type, foreign, local)
-    Observation.connection.execute(%(
-      UPDATE observations o, #{type}s x
-      SET o.#{local} = x.#{foreign}
-      WHERE o.#{type}_id = x.id
-        AND o.#{local} != x.#{foreign}
-    ))
+    # Refresh the mirror of a foreign table's column in the observations table.
+    broken_caches.update_all(
+      "`observations`.`#{local}` = `#{type.pluralize}`.`#{foreign}`"
```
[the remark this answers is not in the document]
nimmolo: Trying to fix this with `.column_name.includes?`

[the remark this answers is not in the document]
nimmolo: Could use help here. I don't understand why my guard clause does not work.

app/models/herbarium.rb
```diff
@@ -168,19 +168,33 @@ def merge_associated_records(src)
...
-        AND h.id = c.herbarium_id
-        AND c.user_id = #{user_id}
-        ORDER BY s.updated_at DESC
-        LIMIT 100
-      )).sort
+      select_manager = arel_select_herbarium_primer
+      # puts(select_manager.to_sql)
+      result = connection.select_values(select_manager).sort
     end
     result
```
JoeCohen, line 175:
```suggestion
return "" unless User.current

select_manager = arel_select_herbarium_primer
# puts(select_manager.to_sql)
connection.select_values(select_manager).sort
```

app/models/herbarium.rb
```diff
@@ -168,19 +168,33 @@ def merge_associated_records(src)
...
+    s = HerbariumRecord.arel_table
+    h = Herbarium.arel_table
+    c = Arel::Table.new(:herbaria_curators)
+    # user_id = 252 # for testing
+
+    h.join(s).on(s[:herbarium_id].eq(h[:id])).join(c).on(
+      h[:id].eq(c[:herbarium_id]).and(c[:user_id].eq(user_id))
+    ).project(h[:name]).distinct.order(s[:updated_at]).take(100)
+  end
+  # rubocop:enable Metrics/AbcSize
```
JoeCohen, line 196: How about the following? 
It's just ActiveRecord. (I probably should have figure this out a long time ago.) Is there an advantage to using AREL if ActiveRecord does the trick?
@pellaea: Is it correct to return a an empty string (as opposed to an empty list) if there's no current user?
```suggestion
  def self.primer
    return "" unless (user_id = User.current_id)

    Herbarium.joins(:herbaria_curators, :herbarium_records).
      where(herbaria_curators: {user_id: user_id}).distinct.
      order(updated_at: :desc).limit(100).pluck(:name)
end
```

app/models/description.rb
```diff
@@ -400,25 +400,46 @@ def group_users(table)
...
-    @group_user_ids[table] ||= self.class.connection.select_values(%(
-      SELECT DISTINCT u.user_id FROM #{table} t, user_groups_users u
-      WHERE t.#{type_tag}_id = #{id}
-        AND t.user_group_id = u.user_group_id
-      ORDER BY u.user_id ASC
-    )).map(&:to_i)
+    select_manager = arel_select_group_user_ids(table)
+    @group_user_ids[table] ||= self.class.connection.
+                               select_values(select_manager.to_sql).map(&:to_i)
   end
```
pellaea, line 406: I wonder if we might not be better off using AR here, too.  It will instantiate a bunch of UserGroups and Users that won't ever be used, but maybe the performance hit would be minimal, a reasonable price to pay for reducing this entire mess to simply:

  old --> name_description.group_user_ids(admins_join_table)
  new --> name_description.admin_groups.map(&:users).flatten.map(&:id)

  old --> name_description.group_ids(admins_join_table)
  new --> name_description.admin_groups.ids

And then we remove these two methods (and all their ABC-violating helpers) completely!

app/models/description.rb
```diff
@@ -561,11 +580,18 @@ def chg_permission(groups, arg, mode)
...
+
+  # SELECT id FROM #{versioned_table_name}
+  # WHERE #{type_tag}_id = #{id} AND user_id = #{user.id}
+  # LIMIT 1
+  def arel_select_user_made_a_change(user)
+    vtn = Arel::Table.new(versioned_table_name.to_sym)
+    vtn.where(vtn["#{type_tag}_id".to_sym].eq(id).
+              and(vtn[:user_id].eq(user.id))).
+      project(vtn[:id]).take(1)
   end
```
pellaea, line 595: Do the version tables have a corresponding AR model?  There is an association, so presumably it must:

  Name#versions

Could this also be rewritten:

  versions.where(user: user).any?

(Note that this method is used only once, and its return value is treated as a simple boolean -- as the question mark suggests it should.)

app/models/donation.rb
pellaea: I think it still makes sense to have mysql take care of the sum, etc. here.  Rewriting this in AR probably wouldn't buy us anything(?)

JoeCohen:
It's really great that nimmo has worked his way through all this stuff and
provided AREL, along with documentation.
I hadn't yet looked at collection_number (or anything other than herbarium).
I won't have a good chunk of time till next Weds or so.
One thing I will try to figure out is whether there's an AR alternative. I
don't know if it will always be better.

app/models/herbarium.rb
```diff
@@ -168,19 +168,33 @@ def merge_associated_records(src)
...
-        AND c.user_id = #{user_id}
-        ORDER BY s.updated_at DESC
-        LIMIT 100
-      )).sort
+      select_manager = arel_select_herbarium_primer
+      # puts(select_manager.to_sql)
+      result = connection.select_values(select_manager).sort
     end
     result
   end
```
pellaea, line 177: This silly thing isn't even used by anyone!  Rip it out, please.

app/models/image.rb
```diff
@@ -972,26 +972,81 @@ def self.update_copyright_holder(old_name, new_name, user)
...
     new_name = Image.connection.quote(new_name)
-    data = Image.connection.select_rows(%(
-      SELECT id, YEAR(`when`), license_id FROM images
-      WHERE user_id = #{user.id} AND copyright_holder = #{old_name}
-    ))
+
+    select_manager = arel_select_copyright_holder(old_name, user)
+    # puts(select_manager.to_sql)
+    data = Image.connection.select_rows(select_manager.to_sql)
     return unless data.any?
```
pellaea, line 979: Notwithstanding my comment above, I think this part at very least would be equivalent in AR:

  data = Image.where(user: user, copyright_holder: old_name).pluck(:id, :when, :license_id)

app/models/image.rb
```diff
@@ -972,26 +972,81 @@ def self.update_copyright_holder(old_name, new_name, user)
...
+      manager.columns << cc[:user_id]
+      manager.columns << cc[:updated_at]
+      manager.columns << cc[:target_type]
+      manager.columns << cc[:target_id]
+      manager.columns << cc[:year]
+      manager.columns << cc[:name]
+      manager.columns << cc[:license_id]
+      manager.values = manager.create_values_list(values_list)
+    end
+  end
```
pellaea, line 1023: Yikes, what a mouthful.  I would love to see if Joe can find an AR equivalent that will bulk-add records to a table.  Unfortunately, adding these one at a time is not an option.  That's what the "orders of magnitude faster" comment in the original code refers to.

pellaea: Not out of the box, at least not until Rails 6.  See this thread:

  https://stackoverflow.com/questions/10022178/insert-multiple-records-using-ruby-on-rails-active-record

app/models/language.rb
```diff
@@ -58,55 +61,116 @@ def self.menu_options
...
-        FROM users
-        WHERE id IN (#{user_ids.join(",")})
-        ORDER BY FIND_IN_SET(id, '#{user_ids.join(",")}')
-      ))
+      user_select_manager = arel_select_top_contributors_in_order(user_ids)
+      user_ids = self.class.connection.select_rows(user_select_manager.to_sql)
+      # puts(user_ids.inspect)
     end
     user_ids
   end
```
pellaea, line 74: Consider rewriting this as AR, too:

  TranslationString.where(language: self).and.not(user_id: 0).group(:user_id).order("count(id)").limit(num).pluck(:user_id)

I would also consider refactoring the caller in order to avoid the FIND_IN_SET thing.  Instead, return just the user ids, let the caller grab the logins, either itself or via a private helper.

app/models/collection_number.rb
```diff
@@ -101,26 +106,36 @@ def log_destroy
...
+    hr = Arel::Table.new(:herbarium_records)
+    new_format_name = Observation.connection.quote_string(format_name)
+    old_format_name = Observation.connection.quote_string(old_format_name)
+
+    HerbariumRecord.joins(observations: :collection_numbers).where(
+      hr[:accession_number].eq(old_format_name).
+      and(hro[:herbarium_record_id].eq(hr[:id])).
+      and(cno[:collection_number_id].eq(id)).
+      and(cno[:observation_id].eq(hro[:observation_id]))
+    ).update_all(accession_number: new_format_name)
```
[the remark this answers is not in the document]
JoeCohen: I think you're correct about re-writing this in AR. But there's an extra set of brackets. I think it should be
```ruby
...where(accession_number: old_format_name, collection_number: { id: id }).update_all.

File: app/models/language.rb [hunk 17]
```diff
@@ -58,55 +61,116 @@ def self.menu_options
...
+                            and(v[:user_id].eq(user.id))).join_sources).
+      group(t[:id]).select(self.class.arel_function_group_concat_strings(v))
+  end
+
+  public
+
+  # JASON QUESTION - Is there a way to rewrite these without the group_concat?
+  # I was thinking maybe just select the :text, and append the newlines in Ruby
+  # via map. But this gives a different score, because of the grouping.
+  # The command in PostgresQL is "STRING_AGG"
```
pellaea, line 162: I hear you, Nimmo.  And I am twitching to massively simplify this in some way, too.  I feel like we could still do the grouping in ruby...

Okay, I'm understanding the scoring a bit better.  The tricky part is that we don't want to count every line of long translation strings for each tiny change.  Say, for example, you have 100 line help message as a single translation string.  Now you add a single word.  We want to count that as only a single line change, not 100!  That's what the hashifying business is in Language.score_lines.

Okay, so what if we returned pairs of translation_string_id and text instead of just text, without the grouping.  Then score_lines could include the translation_string_id in the hash key, and I think it might turn out equivalent?

Does this look right?  I wonder how it would perform, mashing everything into one enormous hash like that?

```suggestion
  # This one is not used, rip it out!
  # def self.calculate_users_contribution(user)
  # end

  def calculate_users_contribution(user)
    hash = {}
    TranslationString.joins(:versions).where(id: id, versions: { user: user }).pluck(:id, "versions.text").each do |id, text|
      text.split("\n").each { |str| hash[id, str] = true if str.present? }
    end
    score_lines(hash)
  end

  def score_lines(hash)
    score = 0
    hash.each_key do |id, str|
      score += (key.length.to_f / CHARACTERS_PER_LINE).truncate + 1
    end
    score
  end
```

app/models/location.rb
```diff
@@ -126,15 +130,26 @@ class Location < AbstractModel
...
-         SELECT COUNT(*) FROM locations_versions
-         WHERE location_id = #{ver.location_id} AND user_id = #{ver.user_id}
-       )).to_s == "0"
+       Location.connection.select_value(count_versions.to_sql).to_s == "0"
+      # if (ver.version != 1) &&
+      #    Location.connection.select_value(%(
+      #      SELECT COUNT(*) FROM locations_versions
+      #      WHERE location_id = #{ver.location_id} AND user_id = #{ver.user_id}
+      #    )).to_s == "0"
       SiteData.update_contribution(:add, :locations_versions)
```
pellaea, line 142: Why can't we just use the ActsAsVersioned relation?

```suggestion
  Location::Version.where(location_id: ver.location_id, user_id: ver.user_id).none?
```

pellaea: (Sorry, I'm still learning how to use the suggestions feature!!)

app/models/location.rb
```diff
@@ -467,18 +483,31 @@ def self.countries_by_count
...
-      )) +
-      Location.connection.select_values(%(
-        SELECT `where` FROM `observations`
-        WHERE `where` is not NULL
-      )) +
-      Location.connection.select_values(%(
-        SELECT `where` FROM `species_lists`
-        WHERE `where` is not NULL
-      ))
+      Location.connection.select_values(
```
pellaea, line 491: In AR:

```suggestion
  Location.pluck(:name) +
    Observation.where.not(where: nil).pluck(:where) +
    SpeciesList.where.not(where: nil).pluck(:where).uniq
```

app/models/name.rb
```diff
@@ -418,16 +431,26 @@ def best_brief_description
...
-        WHERE observations.name_id = names.id
-        AND names.id IN (#{ids.join(", ")}) group by names.id
-    ))
+    select_manager = arel_select_count_observations
+
+    counts_and_ids = Name.connection.select_rows(select_manager.to_sql)
     result = {}
     counts_and_ids.each { |row| result[row[1]] = row[0] }
     result
   end
```
pellaea, line 440: Ooh, this can be translated into a single line!

  Hash[*Observation.group(:name_id).where(name: names).pluck(:name_id, "count(*)").each]

app/models/name/change.rb
```diff
@@ -90,9 +90,21 @@ def clear_misspelled(save = false)
...
+
+  def arel_update_misspelled_consensus_names
+    obs = Observation.arel_table
+
+    #   UPDATE observations SET name_id = #{correct_spelling_id}
+    #   WHERE name_id = #{id}
+    Arel::UpdateManager.new.
+      table(obs).
+      where(obs[:name_id].eq(id)).
+      set([[obs[:name_id], correct_spelling_id]])
```
pellaea, line 108: Again, AR seems much cleaner:

``
Observation.where(name_id: id).update_all(name_id: correct_spelling_id)
``

nimmolo: OK. 
Note that Rubocop flags all uses of `update_all`, that's why i tried to use an `UpdateManager` where possible. I certainly prefer the AR syntax!

app/models/name/format.rb
```diff
@@ -96,10 +96,9 @@ def self.group_suffix(name)
   # Make sure display names are in boldface for accepted names, and not in
   # boldface for deprecated names.
   def self.make_sure_names_are_bolded_correctly
-    Name.connection.select_values(%(
-      SELECT id FROM names
-      WHERE IF(deprecated, display_name LIKE "%*%", display_name NOT LIKE "%*%")
-    )).map do |id|
+    select_manager = arel_select_accepted_and_deprecated
+    # puts(select_manager.to_sql)
+    Name.connection.select_values(select_manager.to_sql).map do |id|
```
pellaea, line 101: I can't improve upon Arel in this case, but since we're going to load each Name instance anyway, why not use find_by_sql?

Actually, I remember you didn't like using named_function (for the IF in this case).  How about doing it as two queries instead:

``
(Name.where(deprecated: true, Name[:display_name].matches("%*%")).to_a + 
 Name.where(deprecated:false).and.not(Name[:display_name].matches("%*%"))).to_a).each do |name|
  name.changed_deprecated(name.deprecated)
  name.save
end
``

Using matches just to check if there is an asterisk in a string always seemed overkill.  There is a function CHARINDEX... would that be better to use?  Are there any db-agnostic equivalents?

app/models/name/lifeform.rb
```diff
@@ -47,33 +47,91 @@ def validate_lifeform
...
+    search_str = Name.connection.quote_string("% #{lifeform} %")
+
+    for type in %w[name observation] do
+      update_manager = arel_update_add_lifeform(
+        type, concat_str, search_str
+      )
+      # puts(all_children.map(&:id).inspect)
+      # puts(update_manager.to_sql)
+      Name.connection.update(update_manager.to_sql)
+    end
```
pellaea, line 60: Yikes, all_children is potentially very expensive and not cached!  Also, can we use AR's update_all instead.  I don't see a way out of using CONCAT, though.

`name_ids = all_children.map(&:id)`
`Name.where(id: name_ids).and.not(Name[:lifeform].matches("% #{lifeform} %")).`
`           update_all(lifeform: Name[:lifeform] + "#{lifeform} ")`
`Observation.where(name_id: name_ids).and.not(Observation[:lifeform].matches("% #{lifeform} %")).`
`           update_all(lifeform: Observation[:lifeform] + "#{lifeform} ")`

Did I shorten "Arel::Nodes::SqlLiteral.new((table[:column] + string).to_sql)" too much?  Does it need the .to_sql?

`...update_all(lifeform: (Observation[:lifeform] + "#{lifeform} ").to_sql)`

Since update_all can take SQL, I assume Arel::Nodes::SqlLiteral must be unnecessary.

app/models/name/lifeform.rb
```diff
@@ -47,33 +47,91 @@ def validate_lifeform
...
-      UPDATE names SET lifeform = REPLACE(lifeform, #{replace_str}, " ")
-      WHERE id IN (#{all_children.map(&:id).join(",")})
-        AND lifeform LIKE #{search_str}
-    ))
-    Name.connection.execute(%(
-      UPDATE observations SET lifeform = REPLACE(lifeform, #{replace_str}, " ")
-      WHERE name_id IN (#{all_children.map(&:id).join(",")})
-        AND lifeform LIKE #{search_str}
-    ))
+    replace_str = Name.connection.quote_string(" #{lifeform} ")
```
pellaea, line 65: This looks equivalent to above, except there is no handy shorthand for REPLACE.  Still, it would seem worthwhile to use update_all.

app/models/name/propagate_generic_classifications.rb
```diff
@@ -44,20 +44,33 @@ def build_accepted_names_lookup_table
...
+    #   where("author NOT LIKE 'sensu lato%'").
+    #   where("LENGTH(classification) > 2").
+    #   pluck(:text_name, :classification).
+    def arel_select_accepted_generic_classification_strings
+      names = Name.arel_table
+      names.where(names[:rank].eq(Name.ranks[:Genus]).
+        and(names[:deprecated].eq(Arel.sql("FALSE"))).
+        and(names[:author].does_not_match("sensu lato%")).
+        and(names[:classification].length > 2)).
+        project(names[:text_name], names[:classification])
```
pellaea, line 69: This is admittedly pretty clean already.  But in the spirit of using ActiveRecord wherever it is at least as concise and efficient as Arel, could we not also do this as:

`Name.where(rank: Name.ranks[:Genus], deprecated: false).`
`          where(Name[:author].does_not_match("sensu lato%")).`
`          where(Name[:classification[.length > 2)).`
`          pluck(:text_name, :classification).`
`          each_with_object({}) do ...`

This stays a single method, no helpers, and it is all db agnostic.

app/models/name/resolve.rb
```diff
@@ -179,7 +179,9 @@ def self.fix_capitalized_species_epithet(str)
     str2 = str.sub(/ [A-Z]/, &:downcase)
 
     # Return corrected name if that name exists, else keep original name.
-    if Name.where("search_name = ? OR text_name = ?", str2, str2).present?
+    # if Name.where("search_name = ? OR text_name = ?", str2, str2).present?
+    if Name.where(Name[:search_name].eq(str2).
+                  or(Name[:text_name].eq(str2))).present?
```
pellaea, line 184: Isn't there an "or" in AR?

`Name.where(search_name: str2).or(text_name: str2)`

app/models/name/spelling.rb
```diff
@@ -75,9 +75,13 @@ def self.names_from_synonymous_genera(str, parent = nil)
...
-        conditions = ["text_name like ? AND correct_spelling_id IS NULL",
-                      "#{synonym.text_name} #{child_pat}"]
-        result += Name.where(conditions).select do |name|
+        # conditions = ["text_name like ? AND correct_spelling_id IS NULL",
+        #               "#{synonym.text_name} #{child_pat}"]
+        # result += Name.where(conditions).select do |name|
+        result += Name.where(
+          Name[:text_name].matches("#{synonym.text_name} #{child_pat}").
+          and(Name[:correct_spelling_id].eq(nil))
+        ).select do |name|
```
pellaea, line 84: Again, more AR and less Arel looks nicer to me:

`Name.where(Name[:text_name].matches(...)).where(correct_spelling: nil).select...`

app/models/name/taxonomy.rb
```diff
@@ -311,19 +313,26 @@ def self.best_match(name)
   #
   def children(all: false)
```
pellaea, line 314: This looks good.

app/models/name/taxonomy.rb
```diff
@@ -523,26 +532,38 @@ def propagate_classification
...
-                         "#{text_name} %").to_a
-    synonyms = Name.where("deprecated IS TRUE AND " \
-                          "synonym_id IN (?) AND " \
-                          "classification != ?",
-                          subtaxa.map(&:synonym_id).reject(&:nil?).uniq,
-                          classification)
+    # subtaxa = Name.where("deprecated IS FALSE AND " \
+    #                      "text_name LIKE ?",
+    #                      "#{text_name} %").to_a
+    subtaxa = Name.where(Name[:deprecated].eq(Arel.sql("FALSE")).
```
pellaea, line 538: Would it be okay to convert the deprecated part to AR?

app/models/name/taxonomy.rb
```diff
@@ -523,26 +532,38 @@ def propagate_classification
...
+                         and(Name[:text_name].matches("#{text_name} %"))).to_a
+    # synonyms = Name.where("deprecated IS TRUE AND " \
+    #                       "synonym_id IN (?) AND " \
+    #                       "classification != ?",
+    #                       subtaxa.map(&:synonym_id).reject(&:nil?).uniq,
+    #                       classification)
+    synonyms = Name.where(Name[:deprecated].eq(Arel.sql("TRUE")).
+                          and(Name[:synonym_id].in(
+                                subtaxa.map(&:synonym_id).reject(&:nil?).uniq
```
pellaea, line 547: Definitely convert the "synonym_id in ..." to AR!!  Also, I feel like converting deprecated to AR is also marginally simpler / cleaner.

app/models/name/taxonomy.rb
```diff
@@ -523,26 +532,38 @@ def propagate_classification
...
+                              )).
+                          and(Name[:classification].not_eq(classification)))
     (subtaxa + synonyms).map(&:id).uniq
   end
 
   # This is meant to be run nightly to ensure that all the classification
   # caches are up to date.  It only pays attention to genera or higher.
   def self.refresh_classification_caches
```
pellaea, line 555: Oh geez, I am about to expire.  That's all I can handle in one sitting, sorry. :(

pellaea:
Okay, then let's turn off that update_all cop globally!!!

app/models/image.rb
nimmolo: This looks right, but the original method only returns the "year" part of `:when`. Does that matter?

pellaea:
Right, I'm glad you asked.  The goal here is to remove mySQL-specific
stuff, and I'm guessing YEAR() is one of those things without any
convenient Arel helper?  If so, why not just return the whole date and pull
the year off in ruby?  Solves the problem handily.

app/models/image.rb
nimmolo: @JoeCohen Here's what I found that works. It uses arel-helpers, which allow you to reference the table directly, using the class name and square brackets: 
```ruby
data = Image.where(user: user, copyright_holder: old_name).
       pluck(Image[:id], Image[:when].year, Image[:license_id])
```

app/models/language.rb
nimmolo: Note : `Language#self.calculate_users_contribution(user)` is used in `language_test.rb`, we need to delete there also, if the class method is deleted.

nimmolo: I believe your line 168 should be 
```ruby
TranslationString.joins(:versions).where(language_id: id, versions: { user_id: user.id }).pluck(etc
```

nimmolo: Ok. I realize the above is just sketched out, but something's off and I can't figure out what it is. 

At least, it's calculating higher scores than before:

Language 4
user 55281 "MO Webmaster"
score before: 18
score after: 65

Now, the webmaster seems to have mostly translated flash messages, which are long. But they all have the same ID. There are 14 strings for ID 14040, like these first two:
```
[14040, "Be a part of Mushroom Observer’s big plans! <br>\r\n<br>\r\nPlease consider \"**+donating+**\":/support/donate before the end of 2020 to help MO determine how many of our \"**+ideas+**\":/articles/16 we can implement over the next year.<br>\r\n<br>\r\nThis year only U.S. taxpayers who won’t be itemizing deductions for 2020 can get a \"**+special deduction+**\":/articles/17 of up to $300 for charitable contributions made before the end of 2020."], 
[14040, "Do you want to design a logo for Mushroom Observer? <br>\r\nWe are running a contest!  Please consider offering an idea. \r\n<a href=\"https://docs.google.com/document/d/1y9OwJhX8eD6FIkrbIXah3gXRMyFYrox0PE-q52R-YHc/edit?usp=sharing\">Click here for contest rules</a>"]
```

pellaea: Definitely delete the test, too.  No reason to have (or test!) code which isn't used.
I need to look at this more carefully, too.  You're right, it was a sketch while very tired!

pellaea: Okay, I think I see the problem.  It's counting multiple versions of the same translation too many times.  Try:

```suggestion
  def calculate_users_contribution(user)
    hash = {}
    TranslationString.joins(:versions).where(id: id, versions: { user: user }).pluck(:id, "versions.text").each do |id, text|
      text.split("\n").each { |str| (hash[id] ||= {}) && (hash[id][str] = true) if str.present? }
    end
    score_lines(hash)
  end
  
  def score_lines(hash)
    score = 0
    hash.each_value do |hash2|
      hash2.each_key do |key|
        score += (key.length.to_f / CHARACTERS_PER_LINE).truncate + 1
      end
    end
    score
  end
```

Now it is truly doing the original "group" entirely in ruby.

nimmolo: Thanks. I tried this, and it actually still produces the same higher score, "65", for MO Webmaster. I also tried grouping the AR result, but that produced a **lower** than original score, "8".

Here's what i'm using, note the small syntax tweaks in case you want to run locally:
```suggestion
  def calculate_users_contribution(user)
    # Note: Rails didn't like "versions.text"
    v = Arel::Table.new(:translation_strings_versions)
    hash = {}
    strings = TranslationString.joins(:versions).
              where(language_id: id,
                    translation_strings_versions: { user_id: user.id }). # full table name neeeded
              group(:id).pluck(:id, v[:text]) # without the `group` it produces much higher score
    strings.each do |id, text|
      text.split("\n").each do |str|
        (hash[id] ||= {}) && (hash[id][str] = true) if str.present?
      end
    end
    score_lines(hash)
  end

  def score_lines(hash)
    score = 0
    hash.each_value do |hash2|
      hash2.each_key do |key|
        score += (key.length.to_f / CHARACTERS_PER_LINE).truncate + 1
      end
    end
    score
  end
```

app/models/description.rb
nimmolo: Seems like this would mean removing three methods here, or am i misreading.
`#group_users(table)`, #group_user_ids(table)`, `#group_ids(table)`

Would this also be a use case for that `deep_pluck` gem?

app/models/description.rb
nimmolo: This works, great idea. It's
```ruby
versions.where(user_id: user.id).any?
```

app/models/donation.rb
nimmolo: By referencing the arel_tables in the pluck and order statements, we could do the sum in AR, the syntax would be almost the same. I think something like 
```suggestion
  Donation.where(anonymous: 0, reviewed: 1).group(:who, :email).
           order(d[:amount].sum.desc, d[:created_at].maximum.desc).
           pluck(:who, d[:amount].sum, d[:created_at].maximum)
  end
```

app/models/description.rb
pellaea: Yes, group_users, too, thanks.  Is there also a groups?

app/models/location.rb
nimmolo: Committing this change. If it's ok i will remove the commented out Arel stuff.

app/models/name/format.rb
nimmolo: I'll stick with `matches` for now. 

AR/Arel mixing notes: We apparently can't stack AR next to Arel inside the `where` - it has to be two separate `where`s. Also, it didn't like joining the arrays for some reason, so I just made it an `or`. This also reduces it back to a single query, instead of the two.

```ruby
Name.where(deprecated: true).where(Name[:display_name.matches("%*%")).
or(Name.where(deprecated: false).where(Name[:display_name].does_not_match("%*%")))
```

app/models/description.rb
nimmolo: Nope there's no `groups`. 
But are you suggesting above that the three methods should be removed, and MO should use `admin_groups.ids` in the codebase instead? Or that these methods should simply get their values directly from `admin_groups` etc. Sorry i'm confused!

pellaea:
Nice!

pellaea:
Actually it may be a good deal cleaner to change the few uses in the
codebase and to delete these methods altogether.  All the callers are in
the same file app/models/description.rb.  For example Description#admins
calls group_users(admins_join_table), change that to admin_groups.users.
Etc.  Does that make sense?

app/models/collection_number.rb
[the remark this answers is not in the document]
JoeCohen: @pellaea: It works with `collection_numbers` (plural). 
The AR and Arel generate identical `explain`s eventhough they create slightly different SQL
See attached file.
[explain.txt](https://github.com/MushroomObserver/mushroom-observer/files/8555839/explain.txt)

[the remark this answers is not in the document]
pellaea: Thanks, Joe!  Yes, then let's go with straight ActiveRecord then as usual, Nimmo.

PS. I'm still completely swamped, and have no idea when I'll have the clarity of mind to properly review the rest of this PR.  Sorry! ... :(

app/models/donation.rb
JoeCohen: 1. I don't see how to do all of this in AR.
2. It would be nice to have a test assertion that tested the return value of donor list.

nimmolo:
Yes i believe this may be an "Arel 💪 Does It" situation. Not sure it can be done in AR. 

I tested this rewrite on a recent db snapshot (in sandbox mode) over and over to get it working and properly formatted, comparing the complete results against the SQL results. 

Agree it needs a better test, but figured that we don't have a complex enough set of fixtures to really test it against. What do you think? 

I believe there are gems that will throw up a larger test fixture set "db" with some parameters, but that might be an imagined feature.

nimmolo:
This mega-PR is now entirely split into smaller PRs, wherein most of the Arel is converted to AR.

pellaea:
Yes, do that!  Sounds good.

pellaea:
Sorry, it finally came up (only took half an hour to load!).  I think I got
it right.

pellaea:
If so, that means the tests didn't catch it failing... that might be worth
at least adding a comment about or something!

PR #908, opened 2022-04-28
Doubled the size of the notes area for species list form.

pellaea:


JoeCohen:
Is this needed by anyone other than Excited Delirium?
Do we want to encourage/allow this misuse/re-purposing of SpeciesLists?
They are mostly pompous essays, which could easily be put elsewhere, like
GoogleDocs.
IMO, we should tell/encourage this user to move his/her/their
"SpeciesLists" to an external site. Then they could add links to the
external sites in Comments in appropriate places. E.g., link to "The Genus
Inocybe Upon The Earth ┃ Part 1" from the Name page for Inocybe.

PR #914, opened 2022-05-06
Fix ip stats

pellaea:
Just rubocop stuff.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -143,7 +143,6 @@ GEM
     kgio (2.11.4)
     libv8-node (16.10.0.0)
     libv8-node (16.10.0.0-x86_64-linux)
-    libv8-node (16.10.0.0-x86_64-linux-musl)
     loofah (2.17.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
@@ -369,4 +368,4 @@ DEPENDENCIES
   xmlrpc
 
 BUNDLED WITH
-   2.3.6
+   2.3.13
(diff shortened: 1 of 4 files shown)


JoeCohen:
Thanks for fixing this up.
Re Commit #77f3667
<https://github.com/MushroomObserver/mushroom-observer/pull/914/commits/77f366719fc486eb17e6d05bf25c96390bca0697#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fR371>
 ("Not wure why this changed for me.")
Probably when you re-installed the VM, it installed the latest version of
`bundle`.
(We might have to update `bundle` on the production server when we deploy
this PR.)

pellaea:
I did, in fact, do a "bundle install" on the production server when I
deployed it.

I wonder: could we be in some screwy state where we are using different
bundlers and Gemfile.lock is going to get subtly tweaked each time a new
developer merges a PR?

JoeCohen:
Good question.
But maybe not a real problem. I think a machine will complain if `bundle
install` plus the various gems are being used with an incompatible version
of `bundler`.
I've seen that before but don't remember the exact circumstances. I think
it was (a) deploying something to production and seeing a complaint on the
production server, and/or (b) pulling master to my vm and running
bundle install..
Note that `bundle install` does not update bundler -- the version on the
production server is still 2.3.5

***@***.***:~$ cd /var/web/mo; sudo su mo

***@***.***:/var/web/mo$ bundler --version

Bundler version 2.3.5
I'm not sure what to do about it, and don't have time to think it through
this week.
At some point, the various machines should complain about having the wrong
version of `bundler` installed, and will suggest an update.

pellaea:
Sounds good to me, thanks, Joe!

PR #916, opened 2022-05-06
SpeciesList + QueryRecord — Rewrite SQL -> Arel

nimmolo:
Tiny PR. See comments for discussions about converting these to AR.
`update_all`-type method easier to convert than `delete_all` method.

Two models affected: `SpeciesList` and `QueryRecord`


nimmolo:
Ok, that's all for this PR. Seems still small enough to fit under the 15-minute mark :)

app/models/query_record.rb
```diff
@@ -15,12 +19,28 @@ def self.cleanup
...
-      DELETE FROM #{table_name}
-      WHERE
-        access_count = 0 AND updated_at < DATE_SUB(NOW(), INTERVAL 6 HOUR) OR
-        access_count > 0 AND updated_at < DATE_SUB(NOW(), INTERVAL 1 DAY)
-    ))
+    delete_manager = arel_delete_cleanup(table_name)
+    # Nimmo note: Reviewers can examine the generated SQL like so:
+    # puts(delete_manager.to_sql)
+    connection.delete(delete_manager.to_sql)
+
```
pellaea, line 26: What would you and Joe think about simplifying the condition -- ignoring access_count and letting everything live for a day?  Would that let you do this in AR instead of Arel?  I never remember the syntax of date comparisons in ActiveRecord.

app/models/query_record.rb
```diff
@@ -15,12 +19,28 @@ def self.cleanup
...
+  #   access_count > 0 AND updated_at < DATE_SUB(NOW(), INTERVAL 1 DAY)
+  private_class_method def self.arel_delete_cleanup(table_name)
+    table = Arel::Table.new(table_name)
+    Arel::DeleteManager.new.
+      from(table).
+      where(table[:access_count].eq(0).
+        and(table[:updated_at].lt(Time.zone.now - 6.hours)).
+        or(table[:access_count].gt(0).
+        and(table[:updated_at].lt(Time.zone.now - 1.day))))
+  end
```
pellaea, line 45: (wow the parentheses are really hard to keep straight in Arel!!)

app/models/species_list.rb
```diff
@@ -201,13 +219,30 @@ def name_included(name)
...
-    ))
+    update_manager = arel_update_defined_location(location, old_name)
+    # puts(update_manager.to_sql)
+    SpeciesList.connection.update(update_manager.to_sql)
+  end
+
+  # Nimmo Note: This can be done in AR with update_all, something like:
+  #   SpeciesList.where(where: old_name).update_all(
+  #     where: new_name, location_id: location.id)
+  # but I'm not sure that is preferable because it skips validations?
```
pellaea, line 230: I think skipping validations is a bonus!

We keep coming up against skipping validations in this work you're doing, Nimmo.  I never really formulated my position on it in so many words, but I guess where I was intuitively coming down on it is: validations are there to protect us from external data, not for double-checking internal code (that's what the unit tests are for).  And since I imagine that validations are relatively expensive -- presumably at very least they must instantiate instances of every record which is changed -- they are to be avoided except where necessary.  In effect: why not run the "validations" once -- i.e., in unit tests -- instead of every single time the code is run -- AR validation callbacks?

So, in this case, I would definitely recommend using SPL.update_all over Arel.

app/models/query_record.rb
nimmolo: I like the idea of giving it a day and simplifying the conditional. 👍

(Incidentally, and beside the point, check this tricky syntax highlighting. I know you're not a fan on principle, but it keeps me sane...)
Screen Shot 2022-05-07 at 5 23 41 PM

app/models/species_list.rb
nimmolo: Thanks for the feedback on validation! This is what i was hoping to hear. It certainly sounds right, but I am too ignorant about the codebase to have a real opinion. 

From now on, i'll do the AR where possible. I would also propose that we add that gem that allows for a Rails 5.2 version of AR insert_all, when needed in a later PR.

app/models/query_record.rb
nimmolo: See below. I like it. Will attempt and commit.

nimmolo: Oops - I don't understand how we can do this in AR, considering we are passing the table name as a variable.

nimmolo: I did however simplify the conditional.

pellaea:
Why is table_name a variable?  It's always going to be "query_records".

JoeCohen:
Nimmo:
Thanks for continuing this series of PRs.
It makes it super easy to review, and I'm learning a lot from them.
. I won't have a chance to review it until Tues.
Joe

nimmolo:
Jason - 
LOL! I had no idea why -- and could not find any definition of the variable. Just assumed it was a utility method built for versatility.  

Joe - 
Thanks!

nimmolo:
Jason - 
In case it was not clear, the original SQL used that variable. 

Maybe have a quick look at the Query class module that calls `#cleanup` ... If you're sure the `table_name` is always going to be `QueryRecord` i will be happy to rewrite!

pellaea:
Yes, I'm positive.  No idea why I wrote it that way.  Obviously it works,
but I can see how it gives the impression that it was intended to be
generalized, but it never was.  I can't imagine ever using it anywhere else
for any other table.  Sorry about that.

nimmolo:
Rewritten.  Still uses a smidge of Arel syntax because in this case, I find it easier to read. 
```ruby
QueryRecord.where(
  QueryRecord[:updated_at] < (Time.zone.now - 1.day)
).delete_all
```
The alternative in "pure" AR would be
```ruby
QueryRecord.where('updated_at < ?', (Time.zone.now - 1.day)).delete_all
```
but I was never crazy about syntax with strings like this, in any language.

nimmolo:
@pellaea Just added a comment block at the head of the model, but I'd love a review of it. 
Tried to follow template of other models.

pellaea:
(I hope this response goes in the right place!)

"Time.zone.now - 1.day" should be equivalent to "1.day.ago".  I think.
Would that make it any more readable?

nimmolo:
Good idea. I switched it to `1.day.ago`.
(It's the `.where('string_parsing > ?', thing)` syntax that i don't like to read. To me the syntax now is very transparent — it says we're checking the value of a table column.)

app/models/species_list.rb
```diff
@@ -200,14 +218,18 @@ def name_included(name)
...
-      SET `where` = #{new_name}, location_id = #{location.id}
-      WHERE `where` = #{old_name}
-    ))
+    # Note: Need to use connection.quote_string here
+    old_name = SpeciesList.connection.quote_string(old_name)
+    new_name = SpeciesList.connection.quote_string(location.name)
+
+    SpeciesList.where(where: old_name).update_all(
+      where: new_name, location_id: location.id
+    )
```
JoeCohen, line 232:
```suggestion
    SpeciesList.where(where: old_name).update_all(
      where: location.name, location_id: location.id
    )
```
I get easily confused by sql and Arel. But I think AR takes care of the quotes, and that `connection.quote_string` adds an extra, and therefore incorrect, level of quotes.
For example:
```sql
2.6.6 :001 > Observation.last.where
...
 => "Secret 'hidey hole', Oregon, USA" 
2.6.6 :002 > Observation.where(where: "Secret 'hidey hole', Oregon, USA")
  Observation Load (302.8ms)  SELECT  `observations`.* FROM `observations` WHERE `observations`.`where` = 'Secret \'hidey hole\', Oregon, USA' LIMIT 11
 => #<ActiveRecord::Relation [#<Observation id: 485741, created_at: "2022-05-09 23:51:49", updated_at: "2022-05-10 00:37:23", when: "2022-05-09", user_id: 4468, specimen: false, notes: {}, thumb_image_id: nil, name_id: 1, location_id: nil, is_collection_location: true, vote_cache: 0.0, num_views: 1, last_view: "2022-05-10 00:36:57", rss_log_id: 594018, lat: nil, long: nil, where: "Secret 'hidey hole', Oregon, USA", alt: nil, lifeform: " ", text_name: "Fungi", classification: "Domain: _Eukarya_", gps_hidden: false>]> 
2.6.6 :003 > Observation.where(where: Observation.connection.quote_string("Secret 'hidey hole', Oregon, USA"))
  Observation Load (288.7ms)  SELECT  `observations`.* FROM `observations` WHERE `observations`.`where` = 'Secret \\\'hidey hole\\\', Oregon, USA' LIMIT 11
 => #<ActiveRecord::Relation []> 
```
I will also shortly push a new test to cover this method. 
Please change/revert it if you don't like it.

JoeCohen (changes requested):
Thanks again for all the great work, digging thru Arel, and dividing the PRs into easily reviewable pieces.

pellaea:
Good catch on the extra quotes, Joe!!

app/models/query_record.rb
```diff
@@ -1,26 +1,53 @@
...
-      WHERE
-        access_count = 0 AND updated_at < DATE_SUB(NOW(), INTERVAL 6 HOUR) OR
-        access_count > 0 AND updated_at < DATE_SUB(NOW(), INTERVAL 1 DAY)
-    ))
+    QueryRecord.where(
+      QueryRecord[:updated_at] < 1.day.ago
+    ).delete_all
+
     @@last_cleanup = Time.zone.now
   end
```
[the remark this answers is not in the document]
nimmolo: I like it. @pellaea does `@last_cleanup` work for you?

app/models/species_list.rb
nimmolo: I like it.

pellaea:
Sure, works for me.

PR #917, opened 2022-05-09
Project — Rewrite SQL -> AR/Arel

nimmolo:
There's only one method to convert. The join seems like it would be complicated to phrase in AR, but if anyone knows the right syntax, let's try it and I'll benchmark. 

The Arel (in the PR currently) produces exactly the same SQL as it's replacing.


app/models/project.rb
```diff
@@ -145,29 +145,51 @@ def add_observation(obs)
...
     imgs.each { |img| images.delete(img) }
     update_attribute(:updated_at, Time.zone.now)
   end
 
+  # SELECT io.image_id FROM images_observations io, observations_projects op
+  # WHERE io.image_id IN (#{img_ids})
+  #   AND io.observation_id != #{obs.id}
+  #   AND io.observation_id = op.observation_id
+  #   AND op.project_id = #{id}
+  def arel_select_leave_these_img_ids(obs, imgs)
```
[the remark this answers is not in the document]
nimmolo: Note:  I believe that's what's throwing the ABC rubocop is the innermost condition`.and(io[:observation_id].eq(op[:observation_id])`. 

It might seem like this condition is implied by the joins...through, and maybe AR would figure out to join them on `observation_id`. But Arel does not produce the correct SQL if I omit that inner condition.

app/models/project.rb
```diff
@@ -145,29 +145,51 @@ def add_observation(obs)
...
 
+  # Nimmo note: I remember this one was a lot of work to get right.
+  # IIRC, the Arel was definitely more efficient than the AR for this join.
+  # But it's been so long, i honestly don't remember. Viva la small PR!
+
   # Remove observation (and its images) from this project. Saves it.
   def remove_observation(obs)
     return unless observations.include?(obs)
 
     imgs = obs.images.select { |img| img.user_id == obs.user_id }
```
pellaea, line 156: This might be better written with ActiveRecord:  obs.images.where(user_id: obs.user_id).  If obs.images has already been loaded, of course, this will cause an unnecessary extra query.  But in the more likely case the relationship has not, then this could save time... in the extremely unlikely case that the observer has attached another user's images to their observation! :)  Okay, maybe you should ignore me...

app/models/project.rb
```diff
@@ -145,29 +145,51 @@ def add_observation(obs)
...
+    io.join(op).on(
+      io[:image_id].in(img_ids).and(
+        io[:observation_id].not_eq(obs.id).and(
+          io[:observation_id].eq(op[:observation_id])
+        ).and(op[:project_id].eq(id))
+      )
+    ).project(io[:image_id])
+  end
+
   # Add species_list to this project if not already done so.  Saves it.
```
pellaea, line 193: I think this actually looks better than straight SQL(!)  Looks great.

app/models/project.rb
nimmolo: I like the simplicity of that... a lot.  I vote in favor.

nimmolo: Didn't work... reverting.

app/models/project.rb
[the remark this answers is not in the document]
nimmolo: Disabled the cop here

app/models/project.rb
```diff
@@ -145,29 +145,47 @@ def add_observation(obs)
...
+        arel_select_leave_these_img_ids(obs, imgs).to_sql
+      ).map(&:to_i)
+
       imgs.reject! { |img| leave_these_img_ids.include?(img.id) }
     end
+
     observations.delete(obs)
     imgs.each { |img| images.delete(img) }
     update_attribute(:updated_at, Time.zone.now)
   end
```
JoeCohen, line 171:
```suggestion
  def remove_observation(obs)
    return unless observations.include?(obs)

    imgs_to_delete(obs).each { |img| images.delete(img) }
    observations.delete(obs)
    update_attribute(:updated_at, Time.zone.now)
  end

  def imgs_to_delete(obs)
    imgs = obs.images.select { |img| img.user_id == obs.user_id }
    return imgs if imgs.none?

    # Do not delete images which are attached to other observations
    # still attached to this project.
    leave_these_img_ids = Image.connection.select_values(
      arel_select_leave_these_img_ids(obs, imgs).to_sql
    ).map(&:to_i)

    imgs.reject { |img| leave_these_img_ids.include?(img.id) }
  end
```
Gets rid of a hidden ABC violation

app/models/project.rb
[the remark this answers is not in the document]
JoeCohen: +1 on disabling the cop.

If it can be done easily (makes **my** head spin) consider writing Arel that returns the images to be deleted (rather than those to be retained) and re-writing the above method accordingly to (among other things) get rid of `reject`.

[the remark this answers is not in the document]
nimmolo: Good idea, but makes my head spin too. I'll pass on the challenge.

app/models/project.rb
nimmolo: Thanks Joe - this is definitely an improvement.

nimmolo: Committed Joe's suggestion. I feel like this PR is ready to merge...?

PR #918, opened 2022-05-09
Observation — Rewrite SQL -> AR/Arel

nimmolo:
Note that `#refresh_cached_column` and `#refresh_cached_column_fix_errors` are here folded into one method with a bit of Arel.

`#update_cache` is here converted to an AR one-liner.

~`#make_sure_no_observations_are_misspelled` is also AR but throws a Rubocop "possible SQL injection" because of the string interpolation, as did the SQL method it replaces. I'm interested in any ideas for rewriting it that don't produce that warning.~


app/models/observation.rb
```diff
@@ -169,9 +169,9 @@ class Observation < AbstractModel
                      through: :observation_views,
                      source: :user
 
+  before_destroy { destroy_orphaned_collection_numbers }
```
nimmolo, line 172: I don't know why this moved! I assumed it was from an earlier commit on master, but it seems not. Feel free to move it back.

nimmolo: Also don't know why it uses bracket syntax instead of symbol. Going to commit the move back to old position now.

app/models/observation.rb
```diff
@@ -225,62 +225,42 @@ def self.refresh_content_filter_caches
...
     end
-    Observation.connection.execute(%(
-      UPDATE observations o, names n
-      SET o.name_id = n.correct_spelling_id
-      WHERE o.name_id = n.id AND n.correct_spelling_id IS NOT NULL
-    ))
-    msgs
+    misspellings.update_all(
+      "`observations`.`name_id` = `names`.`correct_spelling_id`"
+    )
```
JoeCohen, line 263:
```suggestion
    misspellings.update_all("`name` = `names`.`correct_spelling_id`")
```
Slightly shorter; fits on one line. I don't feel strongly about it.

app/models/observation.rb
```diff
@@ -225,62 +225,42 @@ def self.refresh_content_filter_caches
...
   # date.  Fixes and returns a messages for each one that was wrong.
   def self.make_sure_no_observations_are_misspelled
-    msgs = Observation.connection.select_rows(%(
-      SELECT o.id, n.text_name FROM observations o, names n
-      WHERE o.name_id = n.id AND n.correct_spelling_id IS NOT NULL
-    )).map do |id, search_name|
+    misspellings = Observation.joins(:name).
+                   where(Name[:correct_spelling_id].not_eq(nil))
+
+    misspellings.pluck(Observation[:id], Name[:text_name]).map do
```
JoeCohen, line 257:
```suggestion
    misspellings.pluck(:id, Name[:text_name]).map do |id, search_name|
      "Observation ##{id} was misspelled: #{search_name.inspect}"
    end
```

- `:id` is slightly shorter than `Observation[:id]`. I don't feel strongly about this.
- RuboCop complains about args being on a new line:
>Line #258 – Convention: Layout/MultilineBlockLayout: Block argument expression is not on the same line as the block start.

JoeCohen (changes requested):
Great work! 
Sorry it took me so long to review. The other PRs should go faster ... when I have time to get to them.

app/models/observation.rb
```diff
@@ -225,62 +225,42 @@ def self.refresh_content_filter_caches
...
 
   # Used by Name and Location to update the observation cache when a cached
   # field value is changed.
   def self.update_cache(type, field, id, val)
-    Observation.connection.execute(%(
-      UPDATE observations
-      SET `#{field}` = #{Observation.connection.quote(val)}
-      WHERE #{type}_id = #{id}
-    ))
+    Observation.where("#{type}_id": id).update_all("#{field}": val)
```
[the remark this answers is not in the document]
pellaea: Interesting.  Is where(name: id) identical to where(name_id: id) under the hood?  Both avoid in

app/models/name.rb
```diff
@@ -276,6 +276,8 @@
 class Name < AbstractModel
   require "acts_as_versioned"
   require "fileutils"
+  require "arel-helpers"
+  include ArelHelpers::ArelTable
```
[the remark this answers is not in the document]
nimmolo: Thanks for the review! @JoeCohen I believe you may be reviewing an outdated version of the PR... This is already removed. 

I found out the helpers need to be in ApplicationRecord, one level of inheritance up from AbstractModel, in order to catch the QueryRecord model. (QR does not inherit from AbstractModel)

app/models/observation.rb
```diff
@@ -169,9 +172,9 @@ class Observation < AbstractModel
                      through: :observation_views,
                      source: :user
 
+  before_destroy { destroy_orphaned_collection_numbers }
```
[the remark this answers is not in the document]
nimmolo: Try pulling current changes on this branch - this is done.

app/models/observation.rb
nimmolo: @pellaea yes it is straight-up SQL! I tried all sorts of things to get it to assign without a string, nothing worked... maybe because it is a join? Like: `misspellings.update_all(name_id: Name[:correct_spelling_id])`

@JoeCohen - I believe as above this needs to be `name_id` since that's the column name. Could be something (many things!) I don't know about it though.

app/models/observation.rb
nimmolo: Tried that.... Wouldn't work without the table assignment, since `misspellings` is a join.
I'll rebreak the lines though now.

pellaea:
Fascinating.  I trust you!  I'm sure I don't have a chance if you couldn't
figure it out. :)

app/models/observation.rb
```diff
@@ -225,62 +225,42 @@ def self.refresh_content_filter_caches
...
   def self.make_sure_no_observations_are_misspelled
-    msgs = Observation.connection.select_rows(%(
-      SELECT o.id, n.text_name FROM observations o, names n
-      WHERE o.name_id = n.id AND n.correct_spelling_id IS NOT NULL
-    )).map do |id, search_name|
+    misspellings = Observation.joins(:name).
+                   where(Name[:correct_spelling_id].not_eq(nil))
+
+    misspellings.pluck(Observation[:id], Name[:text_name]).map 
+      do |id, search_name|
```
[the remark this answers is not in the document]
nimmolo: Whoops... fixing

nimmolo:
@pellaea I gave up too soon. Found it!! Committing now

app/models/observation.rb
```diff
@@ -225,62 +225,42 @@ def self.refresh_content_filter_caches
...
-      SET o.#{local} = x.#{foreign}
-      WHERE o.#{type}_id = x.id
-        AND o.#{local} != x.#{foreign}
-    ))
+    # Refresh the mirror of a foreign table's column in the observations table.
+    broken_caches.update_all(
+      sanitize_sql_for_assignment(
+        "observations.#{local} = #{type.pluralize}.#{foreign}"
+      )
+    )
```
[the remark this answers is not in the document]
nimmolo: Thanks :) On latest commit, I've finally gotten this assignment out of the SQL string.

app/models/observation.rb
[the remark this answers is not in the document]
nimmolo: For the record, oddly, `update_all(field.to_s: val)` does not work either.

pellaea:
Oh!!  I just got it.  ...  Nope, it's gone.  It made sense for one fleeting
microsecond. :)   Something about the key in the "key: val" syntax can't be
a variable, it has to be an actual string.  Aha, there it is again.  What
if field were actually a string...  "#{field}": val is not the same as
"field: val".  Looks crazy at first glance but it actually makes sense. :)

app/models/observation.rb
nimmolo: I believe this is resolved without resorting to SQL strings now. 

Note that when we're performing an `update_all` on joined tables, we have to specify the table whose column we're updating, as well as the column - Rails apparently won't assume it's the first table. (This situation doesn't come up much, though.) 

In a way, maybe it's nice, in that it eliminates possible ambiguity (to the human reader) in the SQL: The SQL generated by ActiveRecord (via Arel, under the hood) almost always seems to specify the `table_name` ahead of the `column_name`.

pellaea:
Hooray!  Thank you.

PR #920, opened 2022-05-10
Location, Description — Rewrite SQL -> AR/Arel

nimmolo:
~TODO: More description methods maybe can be AR~ 
Ready to go


app/models/description.rb
```diff
@@ -400,25 +400,46 @@ def group_users(table)
...
     @group_ids ||= {}
-    @group_ids[table] ||= self.class.connection.select_values(%(
-      SELECT DISTINCT user_group_id FROM #{table}
-      WHERE #{type_tag}_id = #{id}
-      ORDER BY user_group_id ASC
-    )).map(&:to_i)
+    select_manager = arel_select_group_ids(table)
+    # puts(select_manager.to_sql)
+    @group_ids[table] ||= self.class.connection.
+                          select_values(select_manager.to_sql).map(&:to_i)
```
JoeCohen, line 415: Do we need `.map(&:to_i)`? 
(I'm having trouble concentrating on this, but it looks to me like `arel_select_group_ids` will always return integers.)

app/models/location.rb
```diff
@@ -127,10 +127,8 @@ class Location < AbstractModel
   versioned_class.before_save do |ver|
     ver.user_id = User.current_id || User.admin_id
     if (ver.version != 1) &&
-       Location.connection.select_value(%(
-         SELECT COUNT(*) FROM locations_versions
-         WHERE location_id = #{ver.location_id} AND user_id = #{ver.user_id}
-       )).to_s == "0"
+       Location::Version.where(location_id: ver.location_id,
+                               user_id: ver.user_id).none?
```
JoeCohen, line 131: If you prefer, `count` is more efficient as it fetches only the count from the db.
So the following (or however you prefer to format it).
```suggestion
       Location::Version.where(
         location_id: ver.location_id, user_id: ver.user_id
       ).count.zero?
```

app/models/location.rb
```diff
@@ -245,7 +243,7 @@ def self.official_unknown
   # Get an instance of the Name that means "unknown".
```
JoeCohen, line 243:
```suggestion
  # Get an instance of the Location whose name means "unknown".
```

app/models/location.rb
```diff
@@ -245,7 +243,7 @@ def self.official_unknown
...
   def self.unknown
     names_for_unknown.each do |name|
-      location = Location.find_by("name LIKE ?", name)
+      location = Location.find_by(Location[:name].matches(name))
       return location if location
     end
```
JoeCohen, line 249: 1. Is there an **easy** Arel way to match one of a list? 
I.e., matches 'a' OR matches 'b' ...?
If so, then you wouldn't have to iterate over `names_for_unknown`.
2. @pellaea: Possible preexisting issue. Almost certainly for another PR.
Why `name LIKE ?` (and the Arel equivalent).
Does this risk returning a Location whose name merely includes the name of an unknown location?
(E.g., analogous to "[Black Earth Community Park, Wisconsin, USA](https://mushroomobserver.org/location/show_location/17558?q=1n0zv)")
Why do we allow "translations" of the unknown Location's name?

JoeCohen (commented):
Lppks good with many improvements.
I've gone as far as my understanding of Arel and the prior code allows.
I'd say allow @pellaea some time to comment, then merge.

app/models/location.rb
nimmolo: @JoeCohen  it would be `matches_any(names_for_unknown)` but i believe we need a guard clause to be sure `names_for_unknown` is not empty

nimmolo: @JoeCohen Just tried it locally. The difference between the results for 
```ruby
    Location.where(Location[:name].matches_any(names_for_unknown)).first
```
and going through the list in order, is that "Unknown" is first in the array `names_for_unknown` so the location "Unknown" will always be returned first in the `each` block. 

Hang on, I'm going to try a `find_in_set` instead of `matches_any` — the former checks the placement in the set.

nimmolo: Hmm. Arel extensions find in set (`&`) didn't do it.
The array `names_for_unknown` in English returns
```ruby
["Unknown", "Earth", "World", "Worldwide", "Anywhere", "Everywhere"] 
```
We need a way to match that first one, the `each` block may be the best way.

nimmolo: Should we do
```ruby
    Location.where(Location[:name].matches(names_for_unknown.first)).first
```

JoeCohen: Will this work:
```ruby
    location = Location.find_by(Location[:name].matches_any(names_for_unknown))
    return location if location
```

nimmolo: According to the test, it works even without assigning it to the variable `location`.

Maybe `Earth` is the correct match here? If i try to force it to match `names_for_unknown.first` i.e. "Unknown" it returns `nil`.

nimmolo: Committing your suggestion here, minus the variable.

JoeCohen: I have to take a break, but will look at it further in an hour or so.

JoeCohen: I don't understand how this is used, but:
- in master & in test mode, it returns the Location named "unknown", so it appears correct that it's returning the same thing in this branch;
- I think you should keep the guard clause and the error handling. 
`names_for_unknown` calls `:unknown_locations.l`.
At the moment, all languages  have "translations" for `unknown_locations`. But some random user could delete them, and I think we want to deal with that possibility.

pellaea: I've got no idea how this is used, either.  Has anyone checked where in the code this is called?  It may be out-dated and up for the chop block.

JoeCohen:
Thanks Jason.
I haven't dug into it yet.

app/models/location.rb
pellaea: No, it's definitely still used.  But I think it pre-dates the lock we put on "Earth".  How about we hard-code "Earth" instead.  Well, not *hard-code* hard-code, I mean make it a global constant, perhaps in config/consts.rb?  Would that help streamline this method?  Then it would just be: Location.where(name: MO.unknown_location_name) or the equivalent that returns just the first matching instance.

JoeCohen:
Makes sense to me
The method would then be
Location.find_by(name: MO.unknown_location_name)

app/models/location.rb
nimmolo: Made it so.

nimmolo: Okay, that blew up the tests. I added the constant to config/consts.rb:
```ruby
  config.unknown_location_name = "Earth"
```
This was the revised method I tried, with guard clause checking the constant:
```ruby
  def self.unknown
    raise("There is no \"unknown_location_name\" configured!") if
      MO.unknown_location_name.blank?
    Location.find_by(name: MO.unknown_location_name)
  end
```
Caused [many errors](https://github.com/MushroomObserver/mushroom-observer/runs/6586310302?check_suite_focus=true) in the tests.

pellaea:
I'm sorry, Nimmo.  My fault.  Nothing is ever as simple as I think it will
be, is it?  The problem is there is no "Earth" in the test env.  It's
currently called "Unknown" instead.  Changing the name of the fixture to
"Earth" is sure to break some other tests, unfortunately.  But hopefully it
won't be hard to whack each one of those broken tests in turn.  I hope it
was worth simplifying this method!  :(

pellaea:
Yikes.  Yes, look for "Unknown" in test/controllers/*.rb and
test/models/*.rb -- there are *tons* of cases.  Maybe you could simply
override the configuration in config/environments/test.rb and make it
"Unknown" for test env?  Doesn't seem ideal long-term, but it will save
lots of time for this PR.  ...  Yeah, maybe that is the way to go -- fix
this "properly" in another PR.  Does that sound good?

nimmolo:
Roger - yes there are tons of cases. 

Before this gives anybody a headache, I vote that we postpone any change in the behavior of `Location.unknown` to a separate PR, and merge this one. The version here duplicates the original SQL functionality in Arel without changing anything.

I'll keep the constant added in config/consts.rb so that when I come back to it (maybe tomorrow even), I won't have to reboot vagrant to add the constant again.

nimmolo:
I don't mind the functionality of the method for now, since it's making use of instance variables and not calculating "unknown" every time. However, a constant is certainly preferable - will batter it up properly in the next fish fry.

nimmolo:
Oh wait, I see what you mean - a different const per environment. Will do...

nimmolo:
Agh, i feel like i'm not crazy about doing that... it could get confusing that they are constants but are different, especially once we step away, forget about it and come back. One more thing to learn and memorize. 

If we do move to using a constant, I'd rather just change the fixture and tests from Location "Unknown" to "Earth". If i grep for tests matching the string "Unknown" for a `Location`, I don't actually find so many of them, do you? A lot of the tests seem to be for the `Name` "unknown".

pellaea:
Agreed, that is definitely the better long term solution.  Fair enough,
please feel free to merge it the way it was.

By the way, you shouldn't have to reboot vagrant when you change
config/consts.rb.

nimmolo:
Ok, will strip out the const and merge. I'm going to start that new PR while i'm thinking about it though!
(Idk why, but the const didn't seem available in the console until reboot...)

pellaea:
That's it, thank you, Joe!  How quickly the memory fails when I stop using
these things...

PR #921, opened 2022-05-10
Language — Rewrite SQL -> AR/Arel

nimmolo:
Quite a bit of fancy footwork in this rewrite, thanks to support from the gem author... but the end result is hopefully very simple and readable.

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -77,11 +77,11 @@ class Language < AbstractModel
 
   private_class_method def self.get_user_translation_contributions(user)
     v = Arel::Table.new(:translation_strings_versions)
-    values = TranslationString::Version.
-             where(v[:user_id].eq(user.id)).
-             group(v[:translation_string_id]).
-             select(v[:text].group_concat("\n", order: [v[:text].asc])).
-             pluck(v[:text])
+    TranslationString::Version.
+      where(v[:user_id].eq(user.id)).
+      group(v[:translation_string_id]).
+      select(v[:text].group_concat("\n", order: [v[:text].asc])).
+      pluck(v[:text])
   end
 
   def calculate_users_contribution(user)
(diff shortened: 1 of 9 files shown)


app/models/language.rb
```diff
@@ -58,55 +58,108 @@ def self.menu_options
...
+    ts = Arel::Table.new(:translation_strings)
+    u = Arel::Table.new(:users)
+    TranslationString.where(language: self).where.not(user_id: 0).
+      group(:user_id).order(ts[:id].count).limit(num).
+      joins(:user).pluck(u[:id], u[:login])
   end
 
   # Count the number of lines the user has translated.  Include edits, as well.
   # It counts paragraphs, actually, and weights them according to length.
   def self.calculate_users_contribution(user)
```
nimmolo, line 68: This is one of the gnarliest bits of Arel we have in these PRs, but it does produce the desired result. :)

The way I can imagine simplifying it is to somehow do the `GROUP_CONCAT` in Ruby, rather than SQL, but i can't wrap my head around it — i'm not so sure how the scoring actually works. (I know it's counting lines of versions...)

nimmolo: ohhhh hold on... there's an Arel Extension for that!

nimmolo: Incorporated arel-extension group_concat.

app/models/language.rb
```diff
@@ -127,12 +125,11 @@ def self.score_lines(text)
   def self.update_recent_translations
     cutoff = @@last_update
     @@last_update = Time.zone.now
```
JoeCohen, line 127:
```suggestion
  # We need a class variable here
  @@last_update = 1.minute.ago # rubocop:disable Style/ClassVars

  # Update I18n backend with any recent changes in translations.
  def self.update_recent_translations
    # We need a class variable here
    # rubocop:disable Style/ClassVars
    cutoff = @@last_update
    @@last_update = Time.zone.now
    # rubocop:enable Style/ClassVars
```
Gets rid of some suppressed RuboCop offenses

app/models/language.rb
```diff
@@ -127,12 +125,11 @@ def self.score_lines(text)
...
-      WHERE t.language_id = l.id
-        AND t.updated_at >= #{Language.connection.quote(cutoff)}
-    ))
+    strings = TranslationString.joins(:language).
+              where(TranslationString[:updated_at].gteq(cutoff)).
+              pluck(Language[:locale], :tag, :text)
+
+    for locale, tag, text in strings
       TranslationString.translations(locale.to_sym)[tag.to_sym] = text
     end
```
JoeCohen, line 134:
```suggestion
    strings.each do |locale, tag, text|
      TranslationString.translations(locale.to_sym)[tag.to_sym] = text
    end
```
Another supressed RuboCop offense.

JoeCohen (changes requested):
I don't have much to contribute here -- just the RuboCop stuff.
The SQL and Arel are too much for me to deal with tonight; I'm too tired.
But it certainly reads much better. 
I'm OK with merging it.

PR #922, opened 2022-05-11
Include `arel-helpers` methods in most models

nimmolo:
Note: The `arel-helpers` gem is already added to the Gemfile via a prior PR. 

This PR simply includes the arel-helpers methods in all models whose tables we might want to query in ActiveRecord.  (Some models, like the email models, seem more like utility classes, unlikely to be queried. Correct me if I'm wrong!)

Turns out we have to add the arel-helpers to any table's model that we might want to query using the gem's convenient syntax, no matter which model method is doing the query... the helpers need to be added to the **source** model.

I'd like to ask for this PR to be merged before the other model SQL->AR revisions, because it will enable clearer syntax as we go through the methods. (Model methods often query tables of other models.)

### DIFF
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -67,6 +67,9 @@
 
 class AbstractModel < ApplicationRecord
   self.abstract_class = true
+  require "arel-helpers"
+  include ArelHelpers::ArelTable
+  include ArelHelpers::JoinAssociation
 
   def self.acts_like_model?
     true
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class ApiKey < AbstractModel
+  require "arel-helpers"
+  include ArelHelpers::ArelTable
+
   belongs_to :user
   before_create :provide_defaults
 
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
 class ApiKey < AbstractModel
-  require "arel-helpers"
-  include ArelHelpers::ArelTable
-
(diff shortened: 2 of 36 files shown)


nimmolo:
@JoeCohen  - This PR would be my top priority, before even the Project PR, because it will let me DRY up some of the methods (including in the few PRs already merged). 

Not sure why the CI test is failing, i'll check it later this evening.

nimmolo:
Seems all clear now.

JoeCohen (commented):
My only (probably dumb) question: Would it work to do this just one in `app/models/abstract_model.rb`?

nimmolo:
That makes sense to me, but maybe it would throw an error if there are models without tables. (Is there such a thing?)

But only if we actually try to call the arel helper methods on such a model. Seems unlikely we would.

JoeCohen:
We do have some models that are not backed by tables.
For the reasons you stated, I still think you should do it in AbstractModel, rather than everywhere. 
I think once you make that change, you should merge it.

> That makes sense to me, but maybe it would throw an error if there are models without tables. (Is there such a thing?) But only if we actually try to call the arel helper methods on such a model. Seems unlikely we would.
> […](#)
> On May 14, 2022, at 7:20 AM, Joseph D. Cohen ***@***.***> wrote: @JoeCohen commented on this pull request. My only (probably dumb) question: Would it work to do this just one in app/models/abstract_model.rb? — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.

nimmolo:
Thanks Joe,
I'll make that change when I get back! I don't have merge capability though.

JoeCohen:
Thanks Nimmo.
I will merge it now. Tomorrow I'll figure out how to give you merge
permission.

PR #923, opened 2022-05-11
User — simplify some Arel->AR

nimmolo:
This PR takes advantage of recent suggestions about AR possibilities from Jason and Joe, and better understanding of the Arel extensions gem, to simplify some methods in User from the first attempt to get the queries into Arel.

~The add-arel-helpers PR should merge before this one.~ (done)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -796,7 +796,7 @@ class User < AbstractModel
       arel_last_login_if_recent.desc, User[:contribution].desc
     ).limit(1000).pluck(:login, :name)
 
-    logins = users.map do |login, name|
+    users.map do |login, name|
       name.empty? ? login : "#{login} <#{name}>"
     end.sort
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -529,13 +529,13 @@ class User < AbstractModel
   end
 
   # Return an Array of Project's that this User is an admin for.
-  # Note: Doing this the short way in ActiveRecord produces two extra joins!
   def projects_admin
     # For join tables with no model, need to create an Arel::Table object
     # so we can use Arel methods on it, eg access columns
+    # Note: ActiveRecord joins: through is slower; produces two extra joins
     ugu = Arel::Table.new(:user_groups_users)
 
-    select_manager = Project.arel_table.project(Arel.star).join(ugu).
+    select_manager = Project.arel_table.join(ugu).
                      on(Project[:admin_group_id].eq(ugu[:user_group_id]).
                         and(ugu[:user_id].eq(id)))
 
@@ -566,7 +566,7 @@ class User < AbstractModel
   def preferred_herbarium
     @preferred_herbarium ||= begin
       herbarium_id = HerbariumRecord.where(user_id: id).
-                     order(created_at: :desc).take(1).pluck(:herbarium_id).first
+                     order(created_at: :desc).pluck(:herbarium_id).first
       herbarium_id.blank? ? personal_herbarium : Herbarium.find(herbarium_id)
     end
   end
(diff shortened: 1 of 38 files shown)


nimmolo:
Note — I may be able to also get the four `erase_user` methods into AR, it seems possible to call `update_all` and `delete_all` on  `table.to_s.classify.constantize`

nimmolo:
Nope, cannot use AR in those. I believe it's because some of these update/deletes are on tables (eg versions) not backed by models.

app/models/user.rb
```diff
@@ -602,22 +598,16 @@ def create_personal_herbarium
...
       if projects_member.any?
-        SpeciesList.includes(include).find_by_sql(
-          arel_select_species_lists_by_user_or_project.to_sql
-        )
+        SpeciesList.includes(include).
+          where(SpeciesList[:user_id].eq(id).
+          or(SpeciesList[:id].in(arel_species_list_ids_in_users_projects))).uniq
       else
         species_lists.includes(include)
       end
```
JoeCohen, line 606:
```suggestion
      return species_lists.includes(include) if projects_member.none?

      SpeciesList.includes(include).
        where(SpeciesList[:user_id].eq(id).
        or(SpeciesList[:id].in(arel_species_list_ids_in_users_projects))).uniq
```
gets rid of 3 lines

app/models/user.rb
```diff
@@ -641,25 +631,18 @@ def arel_select_species_list_ids_in_users_projects
   def interest_in(object)
     @interests ||= {}
     @interests["#{object.class.name} #{object.id}"] ||= begin
-      sql = interest_in_sql(object)
-      case Interest.connection.select_value(sql).to_s
-      when "1"
+      i = Interest.where(
+        user_id: id, target_type: object.class.name, target_id: object.id
+      ).limit(1).pluck(:state).first
```
JoeCohen, line 636: How about:
```suggestion
      ).pluck(:state).first
```

app/models/user.rb
```diff
@@ -641,25 +631,18 @@ def arel_select_species_list_ids_in_users_projects
...
+      case i
+      when true
         :watching
-      when "0"
+      when false
         :ignoring
       end
```
JoeCohen, line 642: Ternary operator would shorten this
```suggestion
      i ? :watching : :ignoring
```

app/models/user.rb
```diff
@@ -774,41 +757,23 @@ def self.primer
...
   end
 
-  private_class_method def self.primer_data
-    # How to order - https://stackoverflow.com/a/71282345/3357635
-    users = User.arel_table
-    User.order(arel_function_last_login_if_recent.desc,
-               users[:contribution].desc).
-      limit(1000).pluck(arel_function_login_plus_name).uniq.sort
-  end
+  # NIMMO NOTE: is this method covered by a test?
```
JoeCohen, line 760: It is covered (visited) by a test. 
To see the coverage (without setting a breakpoint and running the tests)
- go to the PR in GitHub
- near the bottom where it shows the CI checks, click Show All Tests
- next to "coverage/coveralls", click Details
- navigate to the file
- go to the method

JoeCohen (changes requested):
Great work. The Arel is much shorter, more readable, and more comprehensible.
My suggestions are just picking around the edges.

nimmolo:
@JoeCohen Thanks for those improvements, I learn more Ruby from you every time.  
I *think* i've got everything in there... i tried doing too much at once and had to resolve a merge, but i believe all your suggestions are in there now

app/models/user.rb
nimmolo: Done

app/models/user.rb
nimmolo: Ok. Just tried `i.to_bool` and `i.to_i` — that didn't cut it either. Merging it with this case statement!

app/models/user.rb
```diff
@@ -601,23 +597,15 @@ def create_personal_herbarium
...
-      if projects_member.any?
-        SpeciesList.includes(include).find_by_sql(
-          arel_select_species_lists_by_user_or_project.to_sql
-        )
-      else
-        species_lists.includes(include)
-      end
-    end
-  end
+      return species_lists.includes(include) if projects_member.none?
```
pellaea, line 600: I am concerned that this will avoid setting @all_editable_species_lists, defeating the memoization.  While I hate unnecessarily splitting methods up, this may be a "necessary" case.  Consider moving the body into something like "calc_all_editable_species_lists", and then make all_editable_species_lists a trivial memoized method: @all_editable_species_lists ||= calc_all_editable_species_lists.  Then when your guard clause here returns early, it will still save the result in @all_editable_species_lists and not have to run species_lists.includes(include) multiple times if all_editable_species_lists is called multiple times.

app/models/user.rb
JoeCohen: Thanks for the clarification., and my apologies:  I kicked off this whole thread by assuming it was a boolean and trying to use a ternary operator.

PR #925, opened 2022-05-12
Donation, Herbarium — SQL -> AR/Arel

nimmolo:
Remove Herbarium.primer (per @pellaea)
Rewrite Donation.donor_list in AR with a touch of Arel


app/models/herbarium.rb
```diff
@@ -161,26 +157,6 @@ def merge_associated_records(src)
...
-        AND h.id = c.herbarium_id
-        AND c.user_id = #{user_id}
-        ORDER BY s.updated_at DESC
-        LIMIT 100
-      )).sort
-    end
-    result
-  end
-
   def self.find_by_code_with_wildcards(str)
```
pellaea, line 160: Interesting, was this code not used anywhere?

nimmolo: There was a guy i was emailing with online, @pellaea, who said "this isn't even used anywhere, rip it out!" I would never be so bold 😇. 

I'm actually in favor of these primers because Sveltekit/GraphQL will do a lot of autocompletes... At this point it's easy to restore it while the SQL is there. How about we restore for now?

pellaea: Ah!  Haha!

Actually, primers may be the wrong way to go in the future, anyway.  The only reason they're in there at all is because I have always had a slow internet connection.  The way every other autocompleter works is they just wait for you to type something then start to make absurd numbers of queries to the server searching based on the partially-entered value.  If we went that round instead we might not need to do a thing: GraphQL probably would handle the whole thing for you automagically.

Case in point: Olivia is using the primer for entering the location on the mobile app.  It's just a really bad way to go.  We ended up having to expand the "primer" from just handling the most common cases for folks with a poor internet connection, to including the entire frickin' database of location names.  That was never its purpose.  If she wants *all* the locations, then she should just ask for them!  That is, we should have given her a separate, dedicated query.  Then we could have included id, lat/long extents, etc.  And perhaps cached the result as a static file.  And so on and on.  But using auto-complete code in an off-line setting like that makes no sense.

So, returning to the case at hand: what purpose exactly would this herbarium primer serve?  The days of internet being so slow that it's worthwhile having it speedily autocomplete the most common choices are over.  I'm fine with leaving the code the way it is, but I question the value of extending this out-dated code into a new (and importantly much-cleaned-up!) version of the UI.

Or put differently: let's be really clear exactly what purpose this code is expected to serve.  I think we have failed at that frequently throughout the development of MO.  Let's start being more deliberate, since we can now.

nimmolo: OK - asking for an SQL check here. 
My AR/Arel is returning no herbaria for user 252 Jason Hollinger, or user 1293 Steve Axford, in the console, so it can't be right.

Original SQL:
```
SELECT DISTINCT h.name AS x
FROM herbarium_records s, herbaria h, herbaria_curators c
WHERE s.herbarium_id = h.id
  AND h.id = c.herbarium_id
  AND c.user_id = #{user_id}
ORDER BY s.updated_at DESC
LIMIT 100
```

The SQL my method generates:
```
SELECT DISTINCT herbaria`.`name`
FROM `herbaria`
INNER JOIN `herbarium_records`
  ON `herbarium_records`.`herbarium_id` = `herbaria`.`id`
INNER JOIN `herbaria_curators`
  ON (`herbaria`.`id` = `herbaria_curators`.`herbarium_id`)
  AND (`herbaria_curators`.`user_id` = #{user_id})
ORDER BY `herbarium_records`.`updated_at` DESC
LIMIT 100
```

nimmolo: Note it's not actually missing the backtick before herbaria, I just accidentally deleted it while pasting
```
SELECT DISTINCT `herbaria`.`name`
```

nimmolo: Whoops! I didn't see your response before I tore into a bunch of Arel!

nimmolo: Well, that was fun. Going to merge without #primer. At least we have a record here of those Arel attempts 😅

nimmolo: Thank you for that considered response, and extended context. Appreciate it!

pellaea:
These pasted snippets return identical results for me on the production
server.  Maybe you "accidentally" corrected something while pasting them
into the email?

PR #927, opened 2022-05-13
AutoComplete — SQL -> AR/Arel

nimmolo:
Rewritten. Question: Maybe move .rough_matches to a scope for these models? Or does that make the logic scattered unnecessarily.
It could be nice for GraphQL — or maybe GraphQL should use this class.


pellaea:
Looks good to me.

PR #930, opened 2022-05-16
AbstractModelTest - be sure that arel-helpers are included

nimmolo:
Test could be valuable in the future, because *not* including the Arel-helpers will throw an error 
```
NoMethodError: undefined method `[]' for #<Class:0x0000562c259b2998>
```
in the innocuous-looking code `User[:login]` — it might be hard to comprehend at first.

Also - removes the require/include arel-helpers from AbstractModel, because that class, in turn, inherits from ApplicationRecord, which already includes the arel-helpers via the previous PR. @JoeCohen Sorry i didn't notice that when committing that PR!

### DIFF
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -67,9 +67,6 @@
 
 class AbstractModel < ApplicationRecord
   self.abstract_class = true
-  require "arel-helpers"
-  include ArelHelpers::ArelTable
-  include ArelHelpers::JoinAssociation
 
   def self.acts_like_model?
     true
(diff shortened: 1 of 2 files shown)


JoeCohen (approved):
Looks good to me. The test idea is great.
Can you try merging it? (That will test whether I've set the permissions correctly.)

nimmolo:
Thanks Joe - Seems the permissions are there.

I'll wait for review on all my outstanding PR's however. You and Jason have caught important stuff, and I'm sure there will be more.

JoeCohen:
Glad that worked.

PR #932, opened 2022-05-18
Api2Test — SQL -> AR/Arel

nimmolo:
Rewrites SQL strings in AR or Arel. Mostly where clauses testing query results.

Note the convenient [arel-extensions methods](https://github.com/Faveod/arel-extensions) `blank` and `not_blank` here  💅 that generate null-coalescing SQL

### DIFF
--- a/test/models/api2_test.rb
+++ b/test/models/api2_test.rb
@@ -550,13 +550,11 @@ class Api2Test < UnitTestCase
   def test_getting_collection_numbers
     params = { method: :get, action: :collection_number }
 
-    # nums = CollectionNumber.where("year(created_at) = 2006")
     nums = CollectionNumber.where(CollectionNumber[:created_at].year == 2006)
     assert_not_empty(nums)
     assert_api_pass(params.merge(created_at: "2006"))
     assert_api_results(nums)
 
-    # nums = CollectionNumber.where("year(updated_at) = 2005")
     nums = CollectionNumber.where(CollectionNumber[:updated_at].year == 2005)
     assert_not_empty(nums)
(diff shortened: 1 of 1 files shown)


test/models/api2_test.rb
```diff
@@ -1747,7 +1777,9 @@ def test_getting_names
     assert_api_pass(params.merge(is_deprecated: "true"))
     assert_api_results(names)
 
-    names = Name.where("date(updated_at) = '2009-10-12'")
+    # rubocop:disable Style/FormatStringToken
+    names = Name.where(Name[:updated_at].format("%Y-%m-%d") == "2009-10-12")
```
pellaea, line 1781: It seems to me that there should be an AR way to do a simple equals condition.

Name.where(updated_at: Date.parse("2009-10-12"))

or something like that?

nimmolo: I'll try it.

nimmolo: That doesn't work, unfortunately. It generates this, which does not pass the parse to the SQL.
```
"SELECT `names`.* FROM `names` WHERE `names`.`updated_at` = '2009-10-12'"
```
The way i've got it was the recommended short syntax from Arel extensions. It generates:
```
"SELECT `names`.* FROM `names` WHERE DATE_FORMAT(`names`.`updated_at`, '%Y-%m-%d') = '2009-10-12'"
```
The extra format seems extraneous, but more foolproof in the event of any future db change.

pellaea: Oh, right, thanks for the explanation.  I think what I was missing is that updated_at is date-time, so this is not actually an "equals condition" like I'd assumed.

PR #934, opened 2022-05-19
Added an extra check in API2 to make sure it doesn't accidentally set…

pellaea:
… the image filename even if the user has set their preferences to not do so.

### DIFF
--- a/app/classes/api2/image_api.rb
+++ b/app/classes/api2/image_api.rb
@@ -62,9 +62,7 @@ class API2
                           user.legal_name,
         # rubocop:enable Layout/MultilineOperationIndentation
         license: parse(:license, :license) || user.license,
-        original_name: parse(
-          :string, :original_name, limit: 120, help: :original_name
-        ),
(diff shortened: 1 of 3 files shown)


app/classes/api2/image_api.rb
```diff
@@ -118,6 +114,19 @@ def parse_create_params!
...
+
+      # This is just a sanity check for the benefit of the mobile app to make
+      # sure it doesn't accidentally explicitly set the original_name even if
+      # the user has requested not to save it.  I'm not sure the mobile app
+      # has access to that preference.
+      return nil if User.current&.keep_filenames == :toss
+
+      val
+    end
+
```
nimmolo, line 129: Out of curiosity, why wouldn't the guard clause go first?

pellaea:
Sorry about that.  Parse has a side-effect of telling the help message
which parameters are expected.  That way the auto-generated help message is
complete.  Feel free to expand the comment to make that more clear.

nimmolo:
Got it. Learning...

PR #941, opened 2022-05-25
Add mass-insert gem for insert_all functionality

nimmolo:
The gem adds a `mass_insert` method, modeled on Rails 6's `insert_all`, that allows for bulk inserts in a single ActiveRecord statement. 

It takes an identically-formatted argument: an *array of hashes containing the values for each row inserted to the table*. (`mass_insert` also has a couple of useful optional arguments that we don't need yet.) 

When we get to Rails 6, it should be easy to swap the method names out and delete the gem.

(I will now remove this commit from the Image SQL PR. Made a separate PR because I also need `mass_insert` to simplify a method in `Synonym`, and get it passing locally for Rails 6.)
If no objections, I'd like to merge this today.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,11 @@ gem("mysql2")
 gem("arel_extensions")
 gem("arel-helpers")
 
+# Add method `mass_insert` for bulk db inserts in ActiveRecord.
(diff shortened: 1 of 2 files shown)


nimmolo:
All opposed, "nay"...

PR #945, opened 2022-05-27
Fix Rails 6.1 deprecation warning for ExternalSite.name validation

nimmolo:
Passes new "case_sensitive" param to the uniqueness validator for ExternalSite.name. 
This PR proposes setting `case-sensitive: false` because domains, at least, are not case sensitive, even if we write them "MycoPortal.org" we don't want to add a second "mycoportal.org"

Please let me know if this seems OK. It's our only deprecation warning(?!)

### DIFF
--- a/app/models/external_site.rb
+++ b/app/models/external_site.rb
@@ -16,7 +16,8 @@ class ExternalSite < AbstractModel
   has_many   :observations, through: :external_links
 
   validates :project, presence: true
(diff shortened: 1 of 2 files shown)


pellaea:
Sounds perfectly good to me.  Is it worth adding a real simple test to demonstrate that this actually works as advertised?

pellaea:
Thanks, Nimmo.  Looks good to me.

PR #946, opened 2022-05-27
Upgrade to Rails 6.1

nimmolo:
**Upgrades Rails 6.0 to Rails 6.1**. There ain't much to it.

**Minor changes in code:**
There's one minor commit to fix a test that breaks in AjaxControllerTest. The assertion seemingly wasn't finding a built hash inside an array of returned hashes, because the returned hash keyword order had changed. (Maybe activerecord changed the order it selects columns? Maybe they were both being evaluated as strings?) Anyway, this solves it.

In addition, this PR has a couple commits resolving two types of deprecation warnings that arise in 6.1. (The fixes could not be backported in PRs to the app in 6.0, because they "fix what wasn't broken".) They're about how ActiveModel::Errors should be handled.

> DEPRECATION WARNING: Enumerating ActiveModel::Errors as a hash has been deprecated.
> In Rails 6.1, errors is an array of Error objects,
> therefore it should be accessed by a block with a single block
> parameter like this:
> 
> person.errors.each do |error|
> attribute = error.attribute
> message = error.message
> end
> 
> You are passing a block expecting two parameters,
> so the old hash behavior is simulated. As this is deprecated,
> this will result in an ArgumentError in Rails 7.0.
> 
> 

> DEPRECATION WARNING: Calling << to an ActiveModel::Errors message array in order to add an error is deprecated. Please call ActiveModel::Errors#add instead.


test/controllers/ajax_controller_test.rb
```diff
@@ -568,9 +568,9 @@ def test_name_primer
   def build_name_primer_item(name)
     { id: name.id,
       text_name: name.text_name,
+      author: name.author,
       deprecated: name.deprecated,
-      synonym_id: name.synonym_id,
-      author: name.author }.to_json
+      synonym_id: name.synonym_id }.to_json
```
nimmolo, line 573: For some reason, Rails 6.1 wants this hash ordered differently now, because the primer is returning results differently.

JoeCohen (commented):
Looks OK to me.

PR #947, opened 2022-05-27
Fix template name deprecations (no file extensions)

nimmolo:
Removes file extensions from two places they aren't needed and don't belong (one that i just added today!)


pellaea:
I'm still stoked that you figured out why the extension used to be necessary in API json results!

PR #948, opened 2022-05-27
Fix deprecation warnings about enumerating or adding to ActiveModel::Errors

nimmolo:
Fixes two deprecation warnings about how ActiveModel::Errors are handled.

DEPRECATION WARNING: Enumerating ActiveModel::Errors as a hash has been deprecated.
In Rails 6.1, `errors` is an array of Error objects,
therefore it should be accessed by a block with a single block
parameter like this:

person.errors.each do |error|
  attribute = error.attribute
  message = error.message
end

You are passing a block expecting two parameters,
so the old hash behavior is simulated. As this is deprecated,
this will result in an ArgumentError in Rails 7.0.

DEPRECATION WARNING: Calling `<<` to an ActiveModel::Errors message array in order to add an error is deprecated. Please call `ActiveModel::Errors#add` instead. (called from must_have_description_or_image at /home/runner/work/mushroom-observer/mushroom-observer/app/models/glossary_term.rb:103)

### DIFF
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -315,13 +315,15 @@ class AbstractModel < ApplicationRecord
   #   obj.errors.add(:attr, "message").
   def formatted_errors
     out = []
-    errors.each do |attr, msg|
-      if /^[A-Z]/.match?(msg)
-        out << msg
+    errors.each do |error|
+      attribute = error.attribute
+      message = error.message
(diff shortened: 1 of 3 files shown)


nimmolo:
Whoops, apparently these changes need to wait for 6.1, cannot be backported to 6.0. Closing this PR.

PR #949, opened 2022-05-27
Language.calculate_users_contribution — Fixes latent code formatting issue for 6.1

nimmolo:
In Language, the plucks in the two `get_contributions` methods were confusing to future Rails 6.1. I just moved them to the `calculate` methods. Very minor refactor.

~In AjaxControllerTest, the method build_name_primer_item built the hash in an order different from expected. (Could have to do with tightened behavior for kwargs?) Anyway this small change solves it.~ Must wait for 6.1, doesn't pass here

### DIFF
--- a/app/models/language.rb
+++ b/app/models/language.rb
@@ -71,7 +71,8 @@ class Language < AbstractModel
   # It counts paragraphs, actually, and weights them according to length.
   def self.calculate_users_contribution(user)
     lines = 0
-    get_user_translation_contributions_overall(user).each do |text|
+    contributions = get_user_translation_contributions_overall(user)
+    contributions.pluck(:text).each do |text|
       lines += score_lines(text)
     end
     lines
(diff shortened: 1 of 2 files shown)


pellaea:
Very minor, fair enough.  But baffling...  Would you mind explaining why rails 6.1 thinks these two versions are any different at all??

nimmolo:
Yes it was baffling. If i remember correctly, it gave a "NoMethodError No method [] for nil:nilClass" on the arel table inside the pluck: `.pluck(v[:column])`

nimmolo:
It also wouldn't accept the usual AR syntax `pluck(:column)` as long as the pluck was in the select method. It only worked when I moved it to the calculate method!

pellaea:
Bizarre.  I'm impressed you were about to figure out that it needed to move to a different method.  I'm not sure I would have ever even considered trying that!

PR #950, opened 2022-05-27
config/environment/test.rb - add config.action_view.cache_template_loading = true

nimmolo:
Seems desirable, in that it may speed up controller or integration tests. 
It's a new Rails 6 config, what do you guys think?

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -34,7 +34,11 @@ MushroomObserver::Application.configure do
   # test suite.  You never need to work with it otherwise.  Remember that
   # your test database is "scratch space" for the test suite and is wiped
   # and recreated between test runs.  Don't rely on the data there!
+  # Rails 6 makes cache_classes default to false, but i'm keeping it true.
+  # Also adds config.action_view.cache_template_loading, seems desirable
+  # [Nimmo 20220526]
   config.cache_classes = true
+  config.action_view.cache_template_loading = true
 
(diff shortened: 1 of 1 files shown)


pellaea:
Why not?  Do we understand how it actually works, though?  I'm concerned that it either won't help at all unless we do something else, or it will attempt to cache things that aren't actually static.

nimmolo:
Will research it and report back.

nimmolo:
> `config.action_view.cache_template_loading` controls whether or not templates should be reloaded on each request. Defaults to whatever is set for config.cache_classes.

Sounds like it is both pretty harmless and won't alter anything, actually, because if the above is true, it already defaults to true in our case. But IIRc, last night I needed it to get the 6.1 branch happy. For that reason I vote to add the config.

nimmolo:
Further research: This config [goes all the way back to Rails 3.1](https://guides.rubyonrails.org/v3.1/configuring.html), and has always defaulted to `true` if `config.cache_classes` was true. So I think we're good.

nimmolo:
Aha, I think I get what was going on, and why it occurred to me to add it. 
What's "new" in Rails 6 is that the `rails update` task now adds this particular config explicitly to `config/environments/test.rb` by default. I presume that's to fix whatever error I was experiencing in tests.

pellaea:
Cool, thanks for looking into it, Nimmo.

PR #951, opened 2022-05-27
Fix a FrozenError: can't modify frozen String in Image.validate_image_name

nimmolo:
This error gets thrown in future Ruby 2.7.6


pellaea:
Good.  Presumably blah.to_s was always returning a copy before.  Thanks.

PR #953, opened 2022-05-27
Upgrade to Ruby 2.7.6

nimmolo:
Steps to try the update, a couple safeguards to prevent a glitch updating the gem `mini-racer` to `0.6.2`.
These can be performed in vagrant console:

**Add the new Ruby version** alongside 2.6.6
- `rvm install 2.7.6`
- `/bin/bash --login`
- `rvm use 2.7.6`
- `ruby -v` (check!)

**Update rubygems and bundler** 
- `gem update --system`
- `gem update bundler`

**Delete your gemfile.lock** and update the gems
- `bundle update`


nimmolo:
Incredible. Nothing but net!

There are many notices of the upcoming kwarg behavior change in our code to fix, but they're just deprecation warnings for now, and only caused by very few methods. `cure_acts_as_versioned` needs a tweak too. 

If I can track them down, I'll fix them in this PR.

pellaea:
Excellent news!  Yeah, I've slowly been fixing the keyword arguments
whenever I come across them.  But I imagine there are still quite a few
left.

I sure hope we can keep acts_as_versioned around.  That's a hard one to
switch out.  We ran up against that a few years ago, and were very thankful
cure_acts_as_versioned came to the rescue!

nimmolo:
I guess you have been fixing them — It's only one method!!!
`NameTest#do_name_parse_test(str, expects, deprecated: false)`
How can I fix? I've been reading up but i'm not sure I understand where to fix it - in the method, or in the calls to the method.
I have all the line numbers that call it in `name_test.rb`

I also think `cure_acts_as_versioned` is fixable, it's our fork isn't it?

nimmolo:
In `acts_as_versioned.rb`, it's here on line 160:

```ruby
        self.version_association_options  = {
                                                    :class_name  => "#{self.to_s}::#{versioned_class_name}",
                                                    :foreign_key => versioned_foreign_key,
        }.merge(options[:association_options] || {})
```
What I don't understand is what the new usage is, for kwargs. I've read a bunch of posts about it and I still don't get it.

For the `NameTest#do_name_parse_test(str, expects, deprecated: false)` it seems like I could move the `deprecated` into the expects, but that's not very DRY.

pellaea:
I think I've misunderstood something.  I'm going to have to look at this
more and get back to you later...

JoeCohen:
So much has been happening, I can't keep up. 
I'm glad you've merged everything without waiting for me.

Re `cure_acts_as_versioned`. 
- It is our fork.
- If it's easily fixable, it should be done.
- IMO we should think about moving to a more current, more widely gem, so that we don't have to maintain it ourselves, even if that means losing all existing old versions. Example [`paper_trail`](https://rubygems.org/gems/paper_trail).

pellaea:
We talked about doing that -- re: using another gem, or the built-in
functionality of recent versions of rails -- in a previous upgrade but it
turned out to be a huge task, so we fell back on the patch.  I forgot that
we did it ourselves.

nimmolo:
@JoeCohen - Thanks! I do feel like the `acts_as_versioned` fix would be easy, I just don't understand the syntax change. But I believe it's in the lines I pointed out. Here's the deprecation warning:
```
/home/vagrant/.rvm/gems/ruby-2.7.6/bundler/gems/acts_as_versioned-59a408816683/lib/acts_as_versioned.rb:243: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```
@pellaea - More likely it is I who misunderstood something! But here is a sample of the warnings:
```
/vagrant/mushroom-observer/test/models/name_test.rb:1086: warning: Passing the keyword argument as the last hash parameter is deprecated:19
/vagrant/mushroom-observer/test/models/name_test.rb:19: warning: The called method `do_name_parse_test' is defined here
/vagrant/mushroom-observer/test/models/name_test.rb:964: warning: Passing the keyword argument as the last hash parameter is deprecated1:18
/vagrant/mushroom-observer/test/models/name_test.rb:19: warning: The called method `do_name_parse_test' is defined here
/vagrant/mushroom-observer/test/models/name_test.rb:1116: warning: Passing the keyword argument as the last hash parameter is deprecated
/vagrant/mushroom-observer/test/models/name_test.rb:19: warning: The called method `do_name_parse_test' is defined here
```
It's basically each time `do_name_parse_test` is called. Here's the code for the test:
```ruby
  # Parse a string, with detailed error message.
  def do_name_parse_test(str, expects, deprecated: false)
    ...
  end

  ...

  def test_name_parse_21
    do_name_parse_test(
      "Amanita “quoted”",
      text_name: 'Amanita "quoted"',
      real_text_name: 'Amanita "quoted"',
      search_name: 'Amanita "quoted"',
      real_search_name: 'Amanita "quoted"',
      sort_name: 'Amanita quoted"',
      display_name: '**__Amanita "quoted"__**',
      parent_name: "Amanita",
      rank: :Species,
      author: ""
    )
  end
```

JoeCohen:
What we discovered when we talked about it was that there's no migration
path from acts_as_versioned to paper_trail;
We'd have to start over with the current model versions as the baseline.
I personally think it's worth it in order to reduce issues of gem
incompatibility.

nimmolo:
I solved the NameTest kwarg issue with a small refactor.

nimmolo:
Will put that NameTest refactor in a separate PR off master tonight. But for the time being, it helps in checking out this PR. 

Note that :

a) this PR bundles Rails 6.1, since it's assuming that PR would get merged first. 

b) it is easy to switch Ruby version back to 2.6.6 after trying this. just `rvm use 2.6.6` - both versions will still be installed on your vagrant box, and 2.6.6 remains the default.

JoeCohen:
@nimmolo. Re `acts_as_versioned`
I'm also not getting it. 
act_as_versioned line 243:
```ruby
has_many :versions, self.version_association_options
```
`self.version_association_options` is a hash
act_as_versioned lines 160-163:
```ruby
self.version_association_options  = {
                                            :class_name  => "#{self.to_s}::#{versioned_class_name}",
                                            :foreign_key => versioned_foreign_key,
}.merge(options[:association_options] || {})
```
So the call to `has_many` does use a hash as its last argument. 
But `has_many` ought to handle that, as it splats the `options` param:
```ruby
# File activerecord/lib/active_record/associations.rb, line 1370
def has_many(name, scope = nil, **options, &extension)
```
I don't know what to try. Maybe add a second positional arg:
```ruby
has_many :versions, nil, self.version_association_options
```
Maybe try an extra splat
```ruby
has_many :versions, **(self.version_association_options)
```
Even worse, I have no idea how to try out these ideas. I forget how I worked with `acts_as_versioned` when I forked it.

nimmolo:
Joe - I'm going to try the extra splat (forking the repo and telling Gemfile to load my local version). Will report back!

nimmolo:
Hi Joe -
It’s not necessary for my local test, but in case it works...
Could you grant me commit privileges to MO’s “acts_as_versioned”?
I seem to have write access to many of the repos, but not this one.

nimmo

nimmolo:
I believe this is good to go:
- Joe has updated `cure_acts_as_versioned` to fix deprecation warnings
- Everything in the app is fixed to avoid deprecation warnings
...but when it's switched, or before, will the vagrantfile need an update too (in developer-startup)?

nimmolo:
Actually, #957 should merge first. But then the dep warnings will be fixed.

nimmolo:
Update: Both #957 and #960 should merge first, to clear Codeclimate warnings.

nimmolo:
Good to merge if you're ready. Rubocop says 👍, Codeclimate says🤙

JoeCohen:
I woke up thinking about this problem.
1. You now have full access (Admin) to that repo.
2. Simply splatting the arg works for me:

          has_many :versions, **self.version_association_options
(I edited acts_as_versioned.rb on my vm.)
So you can simply edit that file in the repo. And also update CHANGELOG.
Or I can do both in the morning. (I'm going back to bed.)
-- Joe

PR #958, opened 2022-05-28
config/application.rb — Only load Rails components we actually use

nimmolo:
This gets rid of a warning in Rails 6.1 that ActiveStorage is not set up. But also eliminates other unused components!
https://andycroll.com/ruby/turn-off-the-bits-of-rails-you-dont-use/

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,8 +1,23 @@
 # frozen_string_literal: true
 
-require(File.expand_path("boot", __dir__))
+require_relative "boot"
 
-require("rails/all")
+# require("rails/rails")
+
+require "rails"
+# Pick the frameworks you want:
+require "active_model/railtie"
+require "active_job/railtie"
+require "active_record/railtie"
+# require "active_storage/engine"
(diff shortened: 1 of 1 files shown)


nimmolo:
Fix parens rubocop

nimmolo:
After further research, I believe this is good to go.

JoeCohen:
Thanks for doing this. It should not only fix the deployment problem, but also slim things down, improving memory use.
I don't know enough about Rails components to spot anything that's missing.
I think that when @pellaea is around, we should merge this, deploy it, and see if it flies. If not, he can do whatever he did last time to revert it on the server.

pellaea:
Cool, shall I go ahead and merge and deploy this?  (I haven't even looked
at it yet, but I'm willing to trust it if the tests pass.)

nimmolo:
Have a quick look, it's a tiny diff!

pellaea:
Yes, I have no doubt... however, it takes half an hour for me to load said
tiny diff on the github site!  Ugh.

PR #959, opened 2022-05-29
Remove Rails gem, cherry pick dependencies (should resolve Unicorn build)

nimmolo:
**Issue**: 
Unicorn build failed for production, because `rails` itself is a gem, and consequently adds all its dependencies in Gemfile.lock. One of these is `activestorage`, which is not set up on master — it requires a db migration to set up its 3 tables. 

Without doing the migrations that set up ActiveStorage, the [other way to solve this dependency](https://blog.sundaycoding.com/blog/2016/08/10/how-to-remove-action-cable-from-a-rails-app/) is by removing the `gem("rails")` from Gemfile, and explicitly including only the dependencies (like `activerecord`) that we want. This also allows us to unload `actionmailbox` and `actioncable`.

But it also means that no other gem can list `rails` as a dependency, or `activestorage` will sneak into the build and the Gemfile.lock.  

The only gem we use that depends on `rails` *per se* is the `rails_db` gem :( so with this PR, i'm removing it (even though i like and use it!)


JoeCohen (approved):
@nimmolo: Looks good.
The following question probably for another PR.
As part of slimming down, can/should all the `simplecov`, `brakeman`, `rubocop`, etc. stuff be moved to inside `group :test, :development`

pellaea:
Good work tracking that down, Nimmo.  I'm impressed.  Elegant solution.  Now let's see if it works(!)...

pellaea:
Oh, wait, then why can't we keep rails_db, too?

JoeCohen:
@nimmolo:  Same question as @pellaea: Would it work to keep `rails_db` in the development group?

pellaea:
I hate to break the news, but this still doesn't fix the problem with unicorn.  Is it possible that unicorn itself requires rails?

pellaea:
Aha, I got something.  If you run "gem dependency activestorage --reverse-dependencies" it shows that actiontext requires activestorage.  What does actiontext do, and do we need it?

pellaea:
Actiontext is not required by any other gems we have.

pellaea:
I don't think we need it.  I will try removing it to see if that at least solves the problem.

JoeCohen:
@pellaea - Dumb question: Are we using the latest stable version of unicorn, and if not, would upgrading help?
On the webserver:
```
$ unicorn --version
unicorn v5.4.0
```
And it looks like unicorn itself is now at 6.1.0
https://yhbt.net/unicorn/LATEST.html

pellaea:
Nope, something is still looking for actiontext/engine...

pellaea:
re: updating unicorn -- Sure, we should try it.  But I say let's wait until we solve this problem, which appears to have nothing to do with unicorn (despite my insistence on bringing up that possibility! :)

JoeCohen:
@pellaea In case you are not seeing it (because of slow connection to GitHub), CI is now failing here:
```
Run bundle exec rake db:schema:load
  bundle exec rake db:schema:load
  bundle exec rake db:fixtures:load
  shell: /usr/bin/bash -e {0}
rake aborted!
LoadError: cannot load such file -- action_text/engine
/home/runner/work/mushroom-observer/mushroom-observer/vendor/bundle/ruby/2.6.0/gems/activesupport-6.[1](https://github.com/MushroomObserver/mushroom-observer/runs/6644725181?check_suite_focus=true#step:10:1).6/lib/active_support/dependencies.rb:33[2](https://github.com/MushroomObserver/mushroom-observer/runs/6644725181?check_suite_focus=true#step:10:2):in `require'
/home/runner/work/mushroom-observer/mushroom-observer/vendor/bundle/ruby/2.6.0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:[3](https://github.com/MushroomObserver/mushroom-observer/runs/6644725181?check_suite_focus=true#step:10:3)32:in `block in require'
/home/runner/work/mushroom-observer/mushroom-observer/vendor/bundle/ruby/2.[6](https://github.com/MushroomObserver/mushroom-observer/runs/6644725181?check_suite_focus=true#step:10:7).0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:299:in `load_dependency'
/home/runner/work/mushroom-observer/mushroom-observer/vendor/bundle/ruby/2.6.0/gems/activesupport-6.1.6/lib/active_support/dependencies.rb:332:in `require'
/home/runner/work/mushroom-observer/mushroom-observer/config/application.rb:16:in `<top (required)>'
/home/runner/work/mushroom-observer/mushroom-observer/Rakefile:6:in `require'
/home/runner/work/mushroom-observer/mushroom-observer/Rakefile:6:in `<top (required)>'
```

pellaea:
Aha!  I've finally gotten unicorn to run on the vagrant box.  Maybe I can finally troubleshoot this properly...

pellaea:
Holy s---... I think I finally got it work already!  config/application.rb was still requiring actiontext.  I just got unicorn to boot on the dev VM.

pellaea:
Hooray!  It works in production, too.

pellaea:
These 4 codeclimate issues are a totally random selection of warnings that have nothing to do with this PR.  It is complaining about the has_many :comments and has_many :interests relationships in LocationDescription and SpeciesList (but not, for example the identical relationships in NameDescription!)  These relationships do, in fact, look hinky to me, but it's outside the scope of this PR.  So I'm going to merge this PR so we can finally deploy master on the production server.  (It's in a detached head state right now, and that bothers me.)

JoeCohen:
Thank you, thank you!

PR #960, opened 2022-05-29
Rubocop 1.30

JoeCohen:
This PR updates RuboCop and associated gems and files.  The purpose is to have the latest RuboCop bug fixes (as well as new cops).
The PR updates:
- `rubocop`, `rubocop_performance`, and `rubocop_rails` gems to the latest versions, removing the now-useless `mry` gem;
- Codeclimate configuration to use a newer version of RuboCop;
- RuboCop configuration to disable new cops by default, simplify directions for updating RuboCop;
- RuboCop todo file (unfortunately adding hundreds of new offenses).

There is no manual test.

PR #962, opened 2022-05-29
AccountControllerTest#test_blocked_ips — Tried setting file times more explicitly so that this test works

pellaea:
on all systems.


nimmolo:
The rubocop `Do not use to_time on Date objects, because they know nothing about the time zone in use.` is a false positive and has been corrected in recent rubocop, according to this issue thread:
https://github.com/rubocop/rubocop-rails/issues/288

nimmolo:
This fixes it for me. THANK YOU @pellaea !

nimmolo:
Added a tweak to match expectation format of calls to get(action, **args) 

`get(:blocked_ips, params: { add_bad: new_ip })`
`get(:blocked_ips, params: { remove_bad: new_ip })`

nimmolo:
Codeclimate checks should clear if #960 merges first (see note above)

JoeCohen:
Jason: Thanks for fixing this. Works like a charm. I was getting so tired
of seeing that test fail.
As Nimmo suggested, Codeclimate is happy with
AccountControllerTest#test_blocked_ips...#962 now that #960 is merged.
I will merge it.

PR #965, opened 2022-05-30
Remove mass_insert gem

JoeCohen:
This PR removes `mass_insert` gem, replacing its `mass_insert` method with Rails' native `insert_all` (which became available in Rails 6.0).

There is no manual test.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,11 +45,6 @@ gem("mysql2")
 gem("arel_extensions")
 gem("arel-helpers")
 
-# Add method `mass_insert` for bulk db inserts in ActiveRecord.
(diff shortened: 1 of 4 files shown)


JoeCohen:
This is a WIP. I should finish it shortly.
Much thanks to @nimmolo for his detailed comments (which I am removing) about the subtle differences between `mass_insert` and `insert all`.

nimmolo:
Yea! Thank you for doing this, Joe.

nimmolo:
You did everything I would have done, I think this can merge.

pellaea:
Yep, looks pretty straightforward to me, too.  Thanks for doing the honors, Joe.

PR #968, opened 2022-05-30
Add Name scope `not_deprecated`

nimmolo:
~Fix a possible false negative — `Name.where(deprecated: false)` does not necessarily catch `nil`, according to the internet.~ Also the scope seemed handy. Implement the new scope in models.

Also does some further AR/Arel in two scripts, hopefully passes tests.

### DIFF
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -405,7 +405,7 @@ class Name < AbstractModel
   scope :with_rank,
         ->(rank) { where(rank: Name.ranks[rank]) if rank }
 
-  scope :not_deprecated, -> { where(deprecated: [false, nil]) }
+  scope :not_deprecated, -> { where(deprecated: false) }
   
   def <=>(other)
     sort_name <=> other.sort_name
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -405,6 +405,8 @@ class Name < AbstractModel
   scope :with_rank,
         ->(rank) { where(rank: Name.ranks[rank]) if rank }
 
+  scope :not_deprecated, -> { where(deprecated: [false, nil]) }
+  
   def <=>(other)
     sort_name <=> other.sort_name
   end
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -406,7 +406,7 @@ class Name < AbstractModel
         ->(rank) { where(rank: Name.ranks[rank]) if rank }
 
   scope :not_deprecated, -> { where(deprecated: false) }
-  
+
   def <=>(other)
     sort_name <=> other.sort_name
   end
--- a/script/mushroom_mapper.rb
+++ b/script/mushroom_mapper.rb
@@ -57,8 +57,8 @@ synonyms = {}
 aliases  = {}
 names    = {}
 ids      = {}
-name_data = Name.pluck(:id, :text_name, :rank, :deprecated, :synonym_id, 
-  :correct_spelling_id)
+name_data = Name.pluck(:id, :text_name, :rank, :deprecated, :synonym_id,
+                       :correct_spelling_id)
 
 # > 5 parameters needed for 2nd name.data block, and it's efficient
 # to use name_data for the 1st block to avoid hitting db twice
@@ -88,7 +88,7 @@ name_data.
 
 # Build table of number of observations per genus.
 observations = {}
-for id in Observation.pluck(:name_id) do
+Observation.pluck(:name_id).each do
   next unless real_id = aliases[id]
 
   text_name, rank, deprecated = names[real_id]
@@ -103,7 +103,7 @@ end
 # Build mapping from genus to famil(ies).
 genus_to_family = {}
 classifications = {}
-for id, genus, classification in 
+for id, genus, classification in
   Name.with_correct_spelling.not_deprecated.with_rank(:Genus).
   pluck(:id, :text_name, :classification) do
   kingdom =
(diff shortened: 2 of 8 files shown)


app/models/name.rb
```diff
@@ -405,6 +405,8 @@ class Name < AbstractModel
   scope :with_rank,
         ->(rank) { where(rank: Name.ranks[rank]) if rank }
 
+  scope :not_deprecated, -> { where(deprecated: [false, nil]) }
+
```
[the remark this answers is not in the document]
pellaea: Have you considered where.not(deprecated: true) instead?  Would that be any cleaner?  (Would it even work??  Null does strange things in SQL...)

script/mushroom_mapper.rb
```diff
@@ -90,10 +88,7 @@
 
 # Build table of number of observations per genus.
 observations = {}
-for id in Name.connection.select_values(%(
-  SELECT name_id
-  FROM observations
-)) do
+for id in Observation.pluck(:name_id) do
```
pellaea, line 91: Would you mind changing the "for x in y" to "y.each do |x|"?  I think it would make rubocop happier.  (All these for loops were written pre-rubocop, sorry.)

script/refresh_sitemap
```diff
@@ -211,9 +209,11 @@ def map_one_object_type(params, comments, last_update)
...
 
 def update_times_of_new_objects(times, table, last_update)
-  Name.connection.select_rows(%(
-    SELECT id, updated_at FROM #{table} WHERE updated_at > "#{last_update}"
-  )).each do |id, time|
+  table = Arel::Table.new(table.to_sym)
+  Name.connection.select_rows(
+    table.where(table[:updated_at] > last_update).
+          project(table[:id], table[:updated_at]).to_sql
+  ).each do |id, time|
```
pellaea, line 216: Would you mind explaining why pluck won't work here?  Oh, is it because table is dynamic?

JoeCohen:
The scope, no matter how defined (see below), is shorter and IMO reads better than pure AR.
Scope definition:
`<name>.deprecated` should never be `nil`.  There are none in the live db.
```ruby
Loading production environment (Rails 6.1.6)
2.7.6 :001 > Name.where(deprecated: nil)
 => #<ActiveRecord::Relation []> 
```
If it is `nil` should we treat it as undeprecated? Probably context dependent. 
So is this scope substituting a possible false positive for a possible false negative?

**@pellaea: What do you think?**

> Fix a possible false negative — `Name.where(deprecated: false)` does not necessarily catch `nil`, according to the internet. Also the scope seemed handy. Implement the new scope in models.

script/refresh_sitemap
pellaea: Oh hey!  Would it feel better to change the callers of update_times_of_new_objects to supply the model class instead of table name?  Then maybe you could use pluck?  With such a distinctive method name, I bet it would be easy to track down all the callers.  (I bet there aren't many.)

nimmolo: Because table is dynamic, yes. You can't do table.pluck on an arel table, only on a model.

app/models/name.rb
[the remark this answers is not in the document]
nimmolo: Remembering that whole AR/Arel experience, this is a known activerecord gotcha. SO says where.not(column: true) doesn't catch the nil. Here's the reference:

https://thoughtbot.com/blog/activerecord-s-where-not-and-nil

JoeCohen:
@nimmolo I should probably not comment further until you're done. Else we may end up tripping over each other.

pellaea:
FWIW, there are no names with deprecated = nil.  I'd rather tweak the database column and explicitly outlaw nil, and thereby keep this code cleaner.  Personally.  Didn't we add nil behavior to a database table column a few months ago?  Was that easy?

script/mushroom_mapper.rb
nimmolo: Done!

script/refresh_sitemap
nimmolo: That sounds good! Lemme see if i can make it work.

nimmolo:
@JoeCohen or @pellaea Can one of you guys tweak the db column to explicitly disallow nil, and commit that migration here or in another PR, and I'll change the scope in Name here to just check for false.

JoeCohen:
@pellaea: I forgot about that db tweak. I think it had to do with `rss_log`. I vaguely remember there there was a problem because there were some records where the field **was** `nil`. But that shouldn't be a problem here.
I'll look for the tweak in the migrations.

> FWIW, there are no names with deprecated = nil. I'd rather tweak the database column and explicitly outlaw nil, and thereby keep this code cleaner. Personally. Didn't we add nil behavior to a database table column a few months ago? Was that easy?

JoeCohen:
Here it is:
```
class ChangeColumnNullRssLogsCreatedAt < ActiveRecord::Migration[5.2]
  def change
    change_column_null(:rss_logs, :created_at, false, Time.now)
    RssLog.update_all("created_at = updated_at")
  end
end
```
But I wonder if the db already prohibits a null value for `deprecated`. I'll try to figure that out.

pellaea:
Ha!  Yes, the db already has "null: false" on that column.  Sorry. :)

nimmolo:
I shoulda checked the schema first!

JoeCohen:
Thanks, Jason! That certainly saves us some trouble.

script/refresh_sitemap
nimmolo: Done

nimmolo:
Any other suggestions? Those were great!

pellaea:
Ha, no, it looks great, merge away.

You guys are getting so much done.  This is a real pleasure to watch! :)

PR #971, opened 2022-05-31
Remove trailing whitespace

JoeCohen:
Fixes RuboCop Layout/TrailingWhitespace hidden in .rubocop_todo.yml


JoeCohen:
Probably the ultimate useless/OCD Cop and PR. 
The offenses are 
- hidden (in .rubocop_todo.ym.)
- invisible (in most editor configurations)
- have no effect (in almost every case)

I will merge it anyway.

PR #973, opened 2022-05-31
Unlock Capybara gem because we're on Ruby 2.7

nimmolo:
Also removes explicit listing of `sprockets` - it's a dependency of `sprockets-rails` and not necessary to include

(there's also a commented out version bump for possible rails_db update)


JoeCohen (commented):
Looks good to me. 
Maybe completely remove the line with `sprockets` and the now commented-out stuff after "capybara"

PR #974, opened 2022-06-01
Update schema.rb kwargs (via Rubocop) from `options` to `charset`

nimmolo:
Rubocop wants to reformat schema.rb like this... is it ok?

I believe the entire file template is a Rubocop autocorrect, updated to be in sync with Rails 6

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -2,17 +2,17 @@
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
 #
-# Note that this schema.rb definition is the authoritative source for your
-# database schema. If you need to create the application database on another
-# system, you should be using db:schema:load, not running all the migrations
-# from scratch. The latter is a flawed and unsustainable approach (the more migrations
-# you'll amass, the slower it'll run and the greater likelihood for issues).
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
+# migrations from scratch. Old migrations may fail to apply correctly if those
+# migrations use external dependencies or application code.
 #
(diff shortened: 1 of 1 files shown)


pellaea:
I don't see as how this makes any meaningful difference.  If this makes someone -- rubocop? -- happier, then by all means...

JoeCohen:
1. When we next migrate, Rails will use whatever format it wants.
So I don't see any harm in reformatting it, but neither do I see an advantage. I think it's easiest to leave it alone.

2. I didn't under "via Rubocop" in the PR title. Everything in the `db` directory, and recursively below, should be excluded from RuboCop by `.rubocop.yml`
```yml
  # Completely ignore the following
  Exclude:
    - .codeclimate.yml
    ...
    - "db/**/*"
```

nimmolo:
Actually, my Activestorage migration commit was completely reverted, so the schema didnt change then

This reformat of schema.rb must have occurred today from rebuilding my Vagrant machine and reinstalling a db checkpoint, then running the migrations. So it was generated by the Rails migration presumably. 

In any case, I suppose you'll find the same change spontaneously popping up on a Vagrant near you, if you rebuild your VM. 

It might be a good idea to merge this reformat separately from the next migration, so we know if it changes anything in itself...?

JoeCohen:
I get it now (maybe).
It's because you ran a migration in a newer version of Rails than existed for our last migration.
>"[R]unning the db:migrate command also invokes the db:schema:dump command, which will update your db/schema.rb file to match the structure of your database."

https://guides.rubyonrails.org/v6.1/active_record_migrations.html#running-migrations
It happens to me if I just `rails db:migrate` in master; the migration changes the format, but nothing else.
So, provided that all of us are on Rail 6.1, it's fine to merge. (Otherwise the format will just bounce back and forth as migrations happen in different Rails version.)

JoeCohen:
@mo-nathan confirms that he's on Rails 6.1, so I'll merge it.

PR #975, opened 2022-06-01
Style/For

JoeCohen:
This PR fixes RuboCop Style/For offenses, including:
- Autocorrect Style/For offenses;
- Manual review of autocorrected changes (because autocorrect is not guaranteed to be correct for Style/For)
- Regenerate RuboCop todo list.

Result: Style/For cop re-enabled; total offenses reduced by ~4%.


JoeCohen:
When this PR is ready for review, I will change it from a Draft to ready-for-review.
I still need to 
- Review the proposed changes from `for` to `each`. 
I used RuboCop's autocorrect feature to make those changes. They aren't guaranteed to be safe because of the subtle difference between `for` and `each`. So I must check whether that matters in any of the places that are changed. (I don't want to simply rely on the tests passing.)
- Fix the "new" codeclimate issues created by the changes from `for` to `each`.
- Regenerate the RuboCop todo file.

pellaea:
Interesting.  What is the subtle difference between for and each?

JoeCohen:
`for` is like `each`, except that local variables constructed in the `for` loop, as well as the iterator itself, are available outside the loop. Here's a link to a [short, but incomplete, practical example.](https://stackoverflow.com/questions/3294509/for-vs-each-in-ruby#answers)

JoeCohen:
So I need to check whether after the loop we use either the iterator or a local variable defined in the loop.
I doubt that we do that, but better to find out now than have some crazy bug pop up in the future.

pellaea:
Yes, I doubt it, too.  I never knew for behaved like that!  I always
thought the iterator and variables defined inside the block were local to
that block only.  No wonder rubocop wants us to change to each.  I used to
think it was silly, but that's a serious gotcha waiting to happen.  Thanks
for the explanation.

app/models/language_tracking.rb
```diff
@@ -104,7 +104,7 @@ def periodically_clean_up
 
     @@last_clean = Time.zone.now
     glob = tag_file("*")
-    for file in Dir.glob(glob)
+    Dir.glob(glob).each do |file|
```
JoeCohen, line 107: RuboCop (via Codeclimate) says 

Redundant begin block detected. NEW
```ruby
 105     begin
```
Severity: MinorFound in [app/models/language_tracking.rb](https://codeclimate.com/github/MushroomObserver/mushroom-observer/app/models/language_tracking.rb?from_sha=bdbb7a48&to_sha=91e2333c) by [rubocop](https://docs.codeclimate.com/docs/rubocop)

Is that correct?
If so, what should I do? Delete both the `begin` and `each`? (Tests do pass if I do that.)

pellaea: Yeah, I probably didn't know for sure whether I needed a begin-end block around the rescue, so just threw it in to be sure.  If rubocop says it's not necessary, then I trust it.

pellaea:
My eyes glazed over by about the 100th for loop!  But I forced them to at least point at every single one.  I didn't see anything that stood out as suspicious in any way.  If the tests pass, I'm fine with merging this.

JoeCohen:
Thanks so much for looking. 
I know the feeling (i.e. glazed eyes). We have some really long methods and I forced myself to read to the bottom of them to make sure we weren't using a variable created in a former `for` loop.)

pellaea:
Oh my god, that's Herculean patience!!  Thank you for being so careful,
Joe.  Did you find anything unexpected?

nimmolo:
That's very striking news to me also, about the `for` bug. Thank you for researching that, Joe.
Seems all good. I'm in favor of merging.

JoeCohen:
Thanks for weighing in @nimmolo. 
I'll merge.

PR #976, opened 2022-06-02
TranslationStrings, Language, I18n — Refactor for Ruby 3

nimmolo:
Goals:
- `TranslationString`: Use current `I18n` native methods ~callbacks `after_create` and `before_update`~  to update the `I18n.translations` hash store using. (These methods merge updates into a `deep_dup` of the translations hash and reassign the hash. Needs to be done that way because hashes without dynamic stuff are automatically frozen in Ruby 3.0.)
- Activate `I18n`'s lazy-loading ~and caching, if we can make them work.~ Lazy-loading ~may be~ is on by default.
- Eliminate the method `Language.update_recent_translations` and its test, since we're updating the hash using the callbacks. (This method checks the database for recent changes in TS, and adds them to the `I18n.translations` hash. No need if we're doing that on create and save.)
- Eliminate `ApplicationController#refresh_translations` which calls the above method
- Fix methods in tests that change the translations hash directly. Add a class method tests can use, `TranslationString.update_localizations(locale, hash_of_tags_and_texts)` to isolate the tricky parts of the necessary syntax


app/models/language.rb
```diff
@@ -100,9 +100,9 @@ def get_user_translation_contributions(user)
     v = Arel::Table.new(:translation_strings_versions)
     TranslationString.
       joins(:versions).
-      where(TranslationString[:language_id].eq(id)).
+      where(language_id: id).
       where(v[:user_id].eq(user.id)).
-      group(TranslationString[:id]).
+      group(:id).
       select(v[:text].group_concat("\n", order: [v[:text].asc]))
```
nimmolo, line 106: Just doin' a bit of AR-izing the Arel here

app/models/translation_string.rb
```diff
@@ -75,7 +76,13 @@ def update_localization
       raise("Localization for :#{tag.to_sym} doesn't exist!")
     end
 
-    data[tag.to_sym] = text
+    store_localization
+  end
```
nimmolo, line 80: This method may be orphaned now. 

Originally I had "after_update" calling this method, which calls "store_localization" after the guard clause. The problem with that is that when the db is being set up during boot or testing, it throws this error and aborts, because the localizations do not at that point exist.

Maybe a `rescue` would solve the problem? I don't understand that method, but suggestions welcome if you feel the guard is necessary.

app/models/translation_string.rb
```diff
@@ -75,7 +76,13 @@ def update_localization
...
+
+  # Update this string in the translations I18n is using.
+  def store_localization
+    I18n.backend.
+      store_translations(language.locale, { mo: { tag.to_sym => text } })
```
nimmolo, line 85: The key here was storing the translation inside the `:mo` hash, not the whole translations hash! (there are non-MO translations in `I18n`)

nimmolo:
OK — I want to limit the scope of this PR and tackle caching, the remaining issue, in another PR.

According to the `I18n` gem itself, it's [easy to set up](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/backend/cache.rb). 

We'd need to have [ActiveSupport::Cache](https://guides.rubyonrails.org/caching_with_rails.html) set up. But fragment caching, in Rails, may be on in production by default.
Anyway, sounds like the scope of a different PR. But it sounds like it could be worth it, according to gem authors.

Here's what the gem has to say:
```
# This module allows you to easily cache all responses from the backend - thus
# speeding up the I18n aspects of your application quite a bit.
#
# To enable caching you can simply include the Cache module to the Simple
# backend - or whatever other backend you are using:
#
#   I18n::Backend::Simple.send(:include, I18n::Backend::Cache)
#
# You will also need to set a cache store implementation that you want to use:
#
#   I18n.cache_store = ActiveSupport::Cache.lookup_store(:memory_store)
# ...continued...
```
...where would these lines go? In `config/application.rb`?
(The other [reference](https://lingohub.com/blog/2013/08/i18n-gem-advanced-features-ruby-rails-internationalization) I found is equally vague about where the config goes.)

JoeCohen:
I don't know a way to turn off a cop on CI. But:
- I'll think about it and see if I can come up with a hack.
- I'll email you directly to see if there's something that could be a work-around for you.

> by the way @JoeCohen - do you know a way to turn off the trailing whitespace and hash syntax cops on CI? Especially the trailing whitespace!

nimmolo:
Ready for review. This PR resolves every remaining error on the ruby 3 branch.

nimmolo:
Hid some comments (including my own off-topic questions) in the GitHub view to make the PR easier to read and understand. 
No offense intended, just trying to reduce noise.

nimmolo:
Summary of this PR for reviewers: 
It no longer invokes the Rails callbacks at all, and in fact doesn't change much. 

It just switches to use the `I18n.backend.store_translations` method for updating the translations hash, in the model, controllers and tests, in methods where these were modifying the hash directly by assignment.

pellaea:
I like how simplified the result is... but I confess that I don't understand the changes.  I'm going to have to trust you.  If you say this works, that's good enough for me!

app/models/translation_string.rb
```diff
@@ -59,11 +59,11 @@ def update_version?
   end
 
   def self.translations(locale)
-    I18n.backend.load_translations if I18n.backend.send(:translations).empty?
-    I18n.backend.send(:translations)[locale.to_sym][MO.locale_namespace.to_sym]
+    # NOTE: This method now checks if it's been initialized, i.e. not empty.
+    I18n.backend.translations[locale.to_sym][MO.locale_namespace.to_sym]
```
nimmolo, line 63: Note: I restored this check in a subsequent PR. I had misread the I18n method — it has a `translations(do_init: boolean)` parameter

PR #977, opened 2022-06-02
Update .codeclimate.yml to use our current rubocop version

nimmolo:


nimmolo:
@JoeCohen I think this might do it. Seems we have to keep this in sync with our rubocop version.

JoeCohen (commented):
You have a good point, and this is a logical change. But unfortunately [Codeclimate says the highest available channel is 1-23-0](https://github.com/codeclimate/codeclimate-rubocop/tree/channel/rubocop-1-23-0).
Possible options:
- Live with the potential incompatibility.
- Downgrade to RuboCop 1.23.
- Try your change and see what happens. Some possibilities:
Codeclimate uses :
  - RuboCop 1.30.  Yay! (But I don't believe this can happen.)
  - highest available channel. (I think that's OK. But see above.)
  - RuboCop 0.52.1. Boo. (I interpret the [Codeclimate documentation](https://docs.codeclimate.com/docs/rubocop#using-rubocops-newer-versions) to mean that that's their default.) 

(I'm going back to bed. May be unavailable until Fri morning.)

nimmolo:
Right! Hmm. 
[Well, someone's thinking the same as us today.](https://github.com/codeclimate/codeclimate-rubocop/issues/315)

PR #978, opened 2022-06-02
Fix Performance/Detect

JoeCohen:
This PR fixes 5 Performance/Detect offenses. 

Result: All Perfomance offenses are now fixed.


JoeCohen:
I will probably merge this once all the CI checks pass:
- I manually confirmed that the changes yield the same results as the old code
- All the changes are in tests, so production is not affected.

PR #980, opened 2022-06-03
Switch to Zeitwerk autoloading, refactor classes and directories

nimmolo:
Zeitwerk ground rules: 

> 1. Class names must be strictly `"snake_case_file_name".camelize.constantize`, filtered via registered inflections.
> 2. Only the first class in each file will be initialized per file, unless that class initializes subsequent classes in that file. Other classes calling, for example, the 15th error class in a single file must require the file.
> 3. The easiest way to avoid unexpectedly not-autoloaded classes is to keep it simple: if it's one class per file, every one is always autoloaded. 

**The Ruby 3.0 PR has been merged into this PR, so it should wait until that merges.**

**This PR:**

Deletes the class `API` and its dependents. The API class will not autoload under Zeitgeist as currently configured, and that class is frozen, not to be updated. **This PR should also wait until the 'Sunset original API' PR merges**, this will bring the number of "Files changed" waay down.

Enforces consistent acronym inflection everywhere for `API` and some other acronyms, so that Zeitwerk doesn't trip over inconsistent acronym inflection (it will not load the wrong capitalization of a class name)
> e.g., ApiKeys > APIKeys
> BadHttpRequest > BadHTTPRequest

Changes filenames to match class name **and directory structure**
> classes/map_collapsible.rb > classes/collapsible_collection_of_mappable_objects.rb
> classes/gmaps.rb > classes/gm/gmap.rb
> classes/api2/errors.rb > classes/api2/error.rb

Refactors classes that are split up via `require_dependency` to `include Module`. Special syntax must be used when the module is intended to create class methods: see notes.

Splits off error classes in `classes/api2` and `classes/pattern_search` into their own files, for reason stated above


app/classes/api2/api_key_api.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class API2
-  # API for ApiKey
-  class ApiKeyAPI < ModelAPI
-    self.model = ApiKey
+  # API for APIKey
+  class APIKeyAPI < ModelAPI
+    self.model = APIKey
```
pellaea, line 7: Haha, that's *awesome*!  And I apparently created the name "ApiKeyAPI" with a completely straight face. :)

app/classes/api2/core/base.rb
```diff
@@ -213,41 +213,37 @@
...
 #  parsers looking for missing error message.
 #
-class API2
-  API_VERSION = 2.0
-
-  def self.version
-    API_VERSION
+module API2::Base
+  # When we `include` a module, `self` refers to the module not the class,
+  # so only instance methods are added. Add class methods like this:
```
pellaea, line 218: This comment is confusing, and I'm afraid it maybe misleading.  Inside a method self still refers to the instance, neither the class nor the module, right?

app/models/name.rb
```diff
@@ -277,20 +277,206 @@ class Name < AbstractModel
...
+  # convert rank start_match to standard form of rank
+  # subspecies is not included because it's the catchall default
+  STANDARD_SECONDARY_RANKS = {
+    f: "f.",
+    sect: "sect.",
+    stirps: "stirps",
+    subg: "subg.",
+    subsect: "subsect.",
+    v: "var."
+  }.freeze
```
pellaea, line 480: Fair enough.  I'm not thrilled with these constants being shoved in name.rb, but it's not too bad.

config/locales/en.txt
```diff
@@ -3926,7 +3926,7 @@
   api_bad_time_range_parameter_value: Invalid time range, "[val]", expect "YYYYMMDDHHMMSS-YYYYMMDDHHMMSS", "YYYYMMDDHHMM-YYYYMMDDHHMM", "YYYYMMDDHH-YYYYMMDDHH", "YYYYMMDD-YYYYMMDD", "YYYYMM-YYYYMM", "YYYY-YYYY", "MM-MM", "YYYYMMHHMMSS", "YYYYMMDDHHMM", "YYYYMMDDHH", "YYYYMMDD", "YYYYMM", "YYYY", "MM".
   api_bad_user_parameter_value: Invalid or unknown user, "[val]", accept numerical id, login or name.
   api_bad_version: Invalid version number "[version]".
-  api_can_only_synonymize_unsynonimized_names: Presently, we're only allowing API clients to synonymize unsynonmized names with other names.  If one of the names has no synonyms, synonymize that name with the others.  But there's no way to merge two sets os synonyms via the API right now.
```
pellaea, line 3929: I'm not too concerned in this case, so it's good the way it is but...  For future cases, be really resistant to changing language tags.  When you do so we lose all the translations for that tag!  In this case, for example, it would be better to leave the tag misspelled, but feel free to correct the spelling of the *value*.

test/models/localization_files_test.rb
```diff
@@ -142,19 +142,21 @@ def source_files(*paths, &block)
     end
   end
 
+  # FIXME: This test needs to search through files
```
pellaea, line 145: Should this comment go away?  Looks like you've done it.

pellaea:
What does "too many classes defined per method" mean?

I think this looks great.  Holy moly, it must have been a *lot* of work.  Okay, sure, a bit tedious and mechanical at times, but still lots and lots of difficult troubleshooting.  My hat's off to you, my friend.

nimmolo:
Thanks. It was like troubleshooting boot camp! 

I probably meant "Too many classes defined per file" but can't remember where i might have written that. Only the first class in a file gets auto-initialized, unless that first class calls other classes in the same file.

pellaea:
Got it.  It was in the description of the PR.  I'll ignore it!

test/models/localization_files_test.rb
nimmolo: Thanks. Removed FIXME comment

nimmolo:
@JoeCohen and @pellaea ... What should be our style here, for multiple `include` and `exclude`?
I don't mind either way, but codeclimate grouses about the one-liners.

JoeCohen:
Re `include`, `exclude`. If we need to chose, I'd use the default. But maybe we should turn this cop off.

pellaea:
My preference is to leave it up to the programmer -- that is, turn off the cop.  But maybe there's a long term cost to one over the other... I don't know, like maybe one style will eventually be deprecated?  Beats me.  I will personally probably try to stay consistent, whatever we choose.  If we choose to turn the cop off, I will probably keep doing one-liners, because I think that's what we've always done in the past.

nimmolo:
Reconsidering my style. 
I think having separate lines would help commenting out includes for debugging. 

But We can make the cop prefer one liners, too. But I'm also in favor of  turning it off

pellaea:
Needless to say, I don't mind the cop enforcing my preferred style. :)

PR #983, opened 2022-06-03
api/error — switch from string concatenations to interpolations

nimmolo:
Extremely minor rubocop fix aids in diffing these files

### DIFF
--- a/app/classes/api/errors.rb
+++ b/app/classes/api/errors.rb
@@ -113,7 +113,7 @@ class API
   class RenderFailed < Error
     def initialize(error)
       super()
-      msg = error.to_s + "\n" + error.backtrace.join("\n")
+      msg = "#{error}\n#{error.backtrace.join("\n")}"
       args.merge!(error: msg)
     end
   end
@@ -220,8 +220,8 @@ class API
 
     def help_message
       if keys_for_patch.any?
-        "query params: " + render_keys(keys_for_get) +
-          "; update params: " + render_keys(keys_for_patch)
+        "query params: #{render_keys(keys_for_get)}; " \
+          "update params: #{render_keys(keys_for_patch)}"
       else
         render_keys(all_keys)
       end


JoeCohen (approved):
Looks fine to me.
@pellaea: How untouchable is `app/classes/api`?
If we keep this version of the API, changes would be needed if new bugs are discovered.
Are minor changes like that proposed in this PR okay?
What about other Codeclimate/RuboCop changes?
Should we exclude Codeclimate and or Rubocop?
Maintaining this version of the API seems like extra work. Should we just remove it?

pellaea:
Excluding it from rubocop sounds like a great idea if it's causing you a
lot of headaches.

Let's step back for a second.  I'm not sure why we are having trouble
reaching a consensus on this.  The idea is to keep this around for a year
or two (maybe it's already been a year?) so that people have time to
migrate to the new API2.  If there are bugs in it, then there have *been*
bugs in it, and people either don't care or have worked around them.  If
they really need bug-free code, then they should migrate to API2.  So, no,
I haven't been fixing bugs in API2 recently.  I see no reason to fix
rubocop offenses because rubocop is all about increasing maintainability
and apparently smooths future upgrades, but we do not maintain API nor are
we going to be upgrading it.  If the upgrade to rails 6 completely breaks
it and it can't be kept on without nontrivial tweaks, then I say we get rid
of it.  But if, so far, we're just running into deprecations, well hell,
leave 'em.  If we upgrade to rails 7 or whatever in a year and those
deprecations finally break, then that sounds like a good time to officially
retire API.

I feel like you guys are working on the assumption that API needs to be
maintained.  But I consider it something that should *not* be maintained.
If it continues to work, great, fine, let's keep it, no harm.  But if it
doesn't work any more, then let's get rid of it.

I think the only area really up for debate is how much work is anyone
willing to do to keep it working associated with, e.g., the upgrade of ruby
or rails.  Outside of that we should not be even touching the code, in my
opinion.  Maybe with the exception of security risks or something like
that.  "Trivial maintenance only"... the question is what is "trivial"?

My understanding is that API is working fine in Rails 6.0 and 6.1.  What is
its status in rails 7?  Are we actually seriously going to essentially skip
rails 6 and jump straight to rails 7 in the coming weeks?  I had understood
from the last meeting that Nimmo was going to open a rails 7 PR so we could
see what it would take to upgrade.  Then we kinda know what our target is
and we can be moving things in that direction gradually and proactively.  I
have not personally had time to look at the new rails 7 PR yet.  How close
is it to going live?  I would personally have suspected that it's a bit too
bleeding edge still, and that it would be more stable if we waited a year
to upgrade to it.  I love the idea of "keeping an eye on it" in this draft
PR so that we can be making the migration proactively in the meantime so
that it will be really easys when we finally officially upgrade.  But I
don't recall being part of a conversation where we officially decided to
upgrade to the very latest version of rails immediately.  Was I just zoning
out during the last meeting and missed it??  Are there features in rails 7
that we really need?

Bottom line is -- *by definition* -- keeping API around should require
little to no work.  If it starts to do so, then we need to have a
conversation and decide whether it's time to drop it.  Having to work to
maintain legacy code is the sign that we're waiting for to tell us we need
to let it go.

PS. It's worth noting that in practice there is hardly any difference
between API and API2.  It should be trivial to migrate any API app to an
API2 app.  It was just such a sweeping refactor under the hood that I
couldn't guarantee identical behavior, so I figured it was an excellent
opportunity to test out how creating a new version in parallel with an old
version would work.  I think that's an important point to keep in mind: to
some extent **API is just a test**.  To that end, I think this conversation
about how to maintain the old version and when we should choose to retire
it is super important.  This is sort of the point of the test.  We're
making this up as we go.  I want you guys to understand and weigh in with
your opinion on how it should happen in the future when we upgrade the API
again.

nimmolo:
I think where this is coming from is that Joe and I are both working with VS Code. The loathsome Redmond critter has done a lot to facilitate my flurry of recent work, and made it way waaay easier to propose changes with some confidence. 

When we open a file, offenses like this are sometimes highlighted. Speaking for myself, it's **more** mental noise to keep ignoring them, than it is to fix them. With little things like this I'd always rather commit a "correction". 

I'm also in favor of deleting API. (And maybe we can talk about getting a Mailchimp account for MO? if that's what it takes to send out a bulk all-users email in cases like this, without getting spam-blocked.)

pellaea:
Thank you, Nimmo.  Let's talk about this at the next dev meeting?  Both
logistics of how to retire the old API, and share observations and
commentary on how this experiment with migrating from API to API2 worked
(and didn't work), with an aim for smoothing it out the next time around,
when and if that ever happens.

nimmolo:
Sounds good.

nimmo

JoeCohen:
Thanks Jason.
I will do that. It's a help for everyone.

PR #985, opened 2022-06-04
Dynamic find_by

JoeCohen:
This PR does some cleanup of our use of dynamic `find_by` methods:
- Deletes an unused method (`find_by_sql_with_limit`)
- Whitelists our dynamic `find_by` methods for the Rails/DynamicFindBy cop
- Enables autocorrection of Rails/DynamicFindBy offenses
- Corrects the remaining offenses

The net result is that future DynamicFindBy offenses will be safely auto-correctable, and about a 4% reduction in total RuboCop offenses


app/models/abstract_model.rb
```diff
@@ -131,17 +131,6 @@ def self.find_object(type, id)
...
-  #   sql = "SELECT id FROM names WHERE user_id = 123"
-  #   names = Name.find_by_sql_with_limit(sql, 20, 10)
-  #
-  def self.find_by_sql_with_limit(sql, offset, limit)
-    sql = sanitize_sql(sql)
-    add_limit!(sql, limit: limit, offset: offset)
-    find_by_sql(sql)
-  end
-
   # Wrap a normal SQL query in a <tt>COUNT(*)</tt> query, then pass it to
```
pellaea, line 134: Good catch, Joe.  These utilities get dug in like ticks.  Somehow we never realize when we've removed the last usage  and it's time to take them out.  I bet the last person to remove a use of this assumed it was a rails method!  Therefore it would never even occur to check to see if the method is now unused and can be removed.

pellaea: [Hey!  Wouldn't *that* be a useful cop -- catching code that is no longer used anywhere!]

pellaea:
Joe, I am a little confused by the opening comment -- where you talk about future offenses being auto-correctable.  And about white-listing some.  Are you saying there are some find_by_xxx that we are going to keep for some reason?  Are there in fact any find_by_xxx left at all in the code (after this PR)?

JoeCohen:
@pellaea Thanks for looking at this PR, the changes, and for the question. Sorry to be unclear.
What I meant was:
If someone were to add a DynamicFindBy offense (ex: `Name.find_by_text_name("foo")` then we could use the RuboCop autocorrect command line flag to fix it automatically (`Name.find_by(text_name: "foo"`).
I had previously disabled autocorrect for this cop because we have our own `find_by_xxx methods, and RuboCop would try to "correct" them.  For instance, it would 
"correct"
```ruby
Location.find_by_name_or_reverse_name(name)
```
to 
```ruby
Location.find_by(name_or_reverse_name: name)
```
In answer to your question about remaining `find_by_xxx` methods:
There are methods left in the code
>`find_by_sql, find_by_code_with_wildcards, find_by_name_with_wildcards, find_by_name_or_reverse_name, find_by_scientific_name_with_wildcards, find_by_title_with_wildcards, find_by_title_with_wildcards`

Tinkering with them is beyond the scope of this PR. 
(I might do a very short PR to change find_by_name_or_reverse_name so that it does only one query.)

pellaea:
Ah!!  I understand now, thank you for the clarification.  Wow, I didn't
realize we'd amassed so many look-alike railsy find_by_xxx custom methods.

nimmolo:
Good to know, i wasn't aware of this. 
After search, I'm reading that `find_by_{column_name}(value)` is **not** deprecated, but the more complicated ones like `find_all_by_{column_name}(value)` are deprecated? 
But... don't use the methods, I get it.

PR #986, opened 2022-06-04
Refactor Location.find_by_name_or_reverse_name

JoeCohen:
This is a one-line change that guarantees 1 db query for this method, instead of 1-2.

Delivers https://www.pivotaltracker.com/story/show/182381087


nimmolo:
Looks good!

PR #988, opened 2022-06-05
accessor grouping

JoeCohen:
This PR fixes all Rubocop Style/AccessorGrouping offenses by:
- Grouping accessors
- Exempting files where accessors should not be grouped because they are individually commented.

Result: The cop is now enabled, and ~137 offenses are removed from the todo list.


.rubocop.yml
```diff
@@ -220,9 +220,19 @@ Rails/HasAndBelongsToMany:
 Rails/WhereExists:
   Enabled: false
 
-# (0.87) pending cop
+# Enable with default EnforcedStyle: grouped; saves lines in some classes
 Style/AccessorGrouping:
```
JoeCohen, line 224: @pellaea @nimmolo @mo-nathan:
In enabling the cop, I decided to stick with the default configuration:
```ruby
EnforcedStyle: grouped (default)
# bad

class Foo
  attr_reader :bar
  attr_reader :baz
end

# good
class Foo
  attr_reader :bar, :baz
end
```
This makes the classes (marginally) shorter, and because it's presumably more widely accepted.
I don't have strong feelings about this.  If you'd rather use the [opposite configuration](https://docs.rubocop.org/rubocop/1.29/cops_style.html#enforcedstyle-separated) please let me know.
Another alternative is disabling the cop entirely.

pellaea:
Thanks, Joe, for giving the rationale behind choosing one over the other.
I couldn't figure out why this cop wasn't pure OCD!  I think I would
personally lean toward disabling it, because as usual I believe there are
good cases for both styles, and I find the comments to disable and
re-enable rubocops a little cluttering and annoying.  But I'm fine with you
submitting this PR as-is.
-Jason

JoeCohen:
@pellaea: Thanks for the comment and for permission to merge. 
I won't do that right away because I want to think it over.
I agree with you about the disable/enable comments. Here, I instead excepted entire files in `.rubocop.yml`. That's a less precise solution that works reasonably well for this cop. (And the disable/enable comments are causing another offense.)

JoeCohen:
"Coverage decreased": I will ignore this; it results from having fewer covered lines, which in turn results from combining multiple covered lines into a single line by grouping the accessors.

JoeCohen:
I will disable this cop (as sugested by @pellaea) and merge, but keep the other changes.
We'll have the option of grouping accessors in order to save a few lines in a particular class(es).

PR #990, opened 2022-06-08
Exclude old API from Rubocop

JoeCohen:
- Excludes the old API, including tests, etc., from Rubocop code analysis
- Regenerates the Rubocop todo list

The old API is frozen, so it shouldn't be changed.
See [Discussion in PR #983](https://github.com/MushroomObserver/mushroom-observer/pull/983#issuecomment-1148762125)

Note that this PR does not exclude Rubocop from [Brakeman](https://brakemanscanner.org/) analysis.
(An old verison of Brakeman is currently invoked by Codeclimate during CI.)
I think that's right, as we probably do want to fix security issues in the old API.


JoeCohen:
I will merge this, as it will make development a tiny bit easier.

PR #991, opened 2022-06-08
Exclude ApiTest

JoeCohen:
This PR fixes a typo in PR #990.

The result is that ApiTest really is excluded (and a few more offenses are removed form the todo list).


JoeCohen:
I will merge this as soon as all CI checks pass.

PR #994, opened 2022-06-08
Use https://mushroomobserver.org in tests and mailers

nimmolo:
Refactors mail tests using `mail` gem's `decoded` method (instead of examining the `encoded` email, which necessarily has the newlines). 

Newlines removed from mailer fixtures, tests should be less brittle!

Note on the other hand, that the `decode` method faithfully reproduces all indents in the mail builder templates. So all HTML indents in builder.erb files are removed by this PR, rather than trying to duplicate the expected indents in the fixtures.


JoeCohen:
I'd forgotten how brittle the tests were. 
It might be worth thinking about how to write less brittle tests.

pellaea:
Agreed.  In all our spare time! :)

pellaea:
Looks good to me.  Again, this must have been a big chunk of work, so thank you, Nimmo!

test/mailers/account_mailer_test.rb
```diff
@@ -28,47 +28,41 @@ def run_mail_test(name, user = nil)
...
+    last = ActionMailer::Base.deliveries.last
+    # Text emails expect a newline before the body, html does not
+    newline = space ? "\n" : ""
+    <<~"EMAIL"
+      From: #{last.from.first}
+      Reply-To: #{last.reply_to.first}
+      To: #{last.to.first}
+      Subject: #{last.subject}
+      #{newline + last.decoded}
+    EMAIL
```
nimmolo, line 65: This is the key new method.

test/mailers/account_mailer_test.rb
```diff
@@ -28,47 +28,41 @@ def run_mail_test(name, user = nil)
...
     end
 
     return unless html_files.any?
 
     user.email_html = true if user
     yield
-    email = ActionMailer::Base.deliveries.last.encoded
-    email = remove_extraneous_email_headers(email)
-    fix_mac_vs_pc!(email)
+    email = String.new(whole_email(space: false))
```
nimmolo, line 39: `String.new(string)` is necessary in Ruby 2.7 to create a mutable string. Takes some getting used to... you can't just `string + string` here.

JoeCohen (approved):
@nimmolo: Thanks so much for doing this and the other PR's while I was out in the woods.
I think you should merge it. It's a big improvement.
I did some quick manual testing, and it looks OK. (If there are hidden problems we'll hear about them and can fix them.)

nimmolo:
I’m on the case!

nimmo

PR #995, opened 2022-06-09
Add translation `api_password_incorrect` to config/locales/en.txt

nimmolo:
This was picked up by working on tests on the Zeitwerk branch — this translation for an API2 error is in fact missing. All the other errors have translations.

Would like to merge soon if no objections.


pellaea:
Thanks, Nimmo.

PR #996, opened 2022-06-09
Use https://mushroomobserver.org outside tests and mailers

nimmolo:
This is to split out the non-test-breaking part of the https:// changes


pellaea:
Looks innocent enough.  Thanks, Nimmo.

PR #998, opened 2022-06-10
Api2Controller fix `require_dependency "api2"` (not `"api"`)

nimmolo:
This fixes a longstanding (and until now, inconsequential) oversight, which would become consequential if we deleted the original API. Anyway, "api2" is obviously the intended require_dependency.

I'd like to merge today if it's OK.


nimmolo:
Actually I'm going to go ahead and merge this without bothering you guys. 
Jason and I discussed it in another context, and there's no question this was the intent of this line of code.

PR #1000, opened 2022-06-10
Update `api` references in Image, scripts and tests, where `api2` seems intended

nimmolo:
There are a few places in the code that only access one API, and i feel like they should probably be using API2 at this point.

Please check me on this!


script/make_eol_xml
```diff
@@ -84,8 +84,8 @@ end
 ################################################################################
 
 if testing
-  LOCAL_URL = "https://localhost:3000/api/observations?id=1"
-  REMOTE_URL = "https://mushroomobserver.org/api/observations?id=1"
+  LOCAL_URL = "https://localhost:3000/api2/observations?id=1"
```
JoeCohen, line 87: Should this (and line 90) be `http:`?

JoeCohen: 1. We can't go to `https://localhost...`. So things would blow up if someone tried that. 
2. I don't know if we are really using the eol code currently. @mo-nathan is the best person to answer that.

pellaea: Agreed, the local server on the dev VM isn't secure.

script/old/refresh_controllers
```diff
@@ -186,7 +186,7 @@ end
 
 norobots = [
   "/ajax/",
-  "/api/"
+  "/api2/"
```
JoeCohen, line 189: @pellaea Are the scripts in `script/old` used?
If yes, what's intended here (`api` vs `api2`)?

script/old/upload_observation.rb
```diff
@@ -515,7 +515,7 @@ def post_data
 
   def post_observation
     http = Net::HTTP.new(HOST, PORT)
-    path = make_path("/api/observations",
+    path = make_path("/api2/observations",
```
JoeCohen, line 518: @pellaea Are the scripts in `script/old` used?
If yes, what's intended here (`api` vs `api2`)?

pellaea:
It's probably time to remove the old scripts.  I'd initially left them in there in case they were still used, but after a few years and I haven't used a single one in all that time, I'd say it's safe to remove them!  Same with test_api.rb.  Totally unnecessary now.  (And might not even work...)

test/models/localization_files_test.rb
```diff
@@ -144,7 +144,7 @@ def source_files(*paths, &block)
 
   def test_api_error_translations
     tags = []
-    file = "#{::Rails.root}/app/classes/api/errors.rb"
+    file = "#{::Rails.root}/app/classes/api2/errors.rb"
```
JoeCohen, line 147: @pellaea: Should we run this for both `.../api/errors` and `.../api2/errors? 
If not, this looks right.
(And I think it explains something that puzzled me: How did we end up with the missing translation that @nimmolo fixed in #995?)

test/models/localization_files_test.rb
```diff
@@ -154,7 +154,7 @@ def test_api_error_translations
         tags << "api_#{Regexp.last_match(1).underscore.tr("/", "_")}".to_sym
       end
     end
-    Dir.glob("#{::Rails.root}/app/classes/api/parsers/*.rb").each do |file|
+    Dir.glob("#{::Rails.root}/app/classes/api2/parsers/*.rb").each do |file|
```
JoeCohen, line 157: See above question.

test/models/localization_files_test.rb
nimmolo: Ahaa!

test/models/localization_files_test.rb
nimmolo: I'd in favor of testing api2 only, because i'm hoping we will see fit to delete API very soon and it's less busywork. But I can add it now, merge, and add the delete to the "Sunset the API" branch if you like.

nimmolo: Oh also - I believe they use the same localization files, so testing one should be enough.

pellaea:
We definitely should not be doing any development on API!  Even if we were
to keep it, we should not be adding tests or fixing bugs or anything on
it.  I hope we can finally clear this up at the meeting tomorrow.  You both
seem to think that keeping API around is some big burden.  It should not be.

nimmolo:
Keeping the API frozen is a blocker for zeitwerk, so yea it is kind of a burden.

pellaea:
Right, which is the reason we're talking about retiring it now.  But why in
the world would we be considering adding tests for it in the meantime?  Or
ever?

nimmolo:
Right! Well this PR effectively _subtracts_ a test from the API, I think is why Joe raised this issue. However, the same translations are still getting tested.

Incidentally, I updated localhost to use http, and deleted the old scripts and test_api.rb. I'm basically doing these small PRs, where I'm sure of relatively minor changes, to try and remove the noise and cruft from the bigger ones. Let me know if you feel like it's mergeable, both of yous! :)

pellaea:
Yes, I'm totally down with merging the deletion of old scripts and
test_api.rb.

script/make_eol_xml
nimmolo: Changed to `http://localhost...` in both places

JoeCohen:
I'm fine with merging it.
Because TestApi is now toast, it makes sense to also revise `test_api_error_translations`.
NOTE: If we choose to retain the old API, do we lose something by the above changes?
In particular, could a change other than in API cause it to break? And could that breakage be detected by TestApi?

pellaea:
I don't think we lose anything.

PR #1001, opened 2022-06-10
Rails 6.1/Ruby 2.7 - update not-yet-mandatory things

nimmolo:
Two lines of code change in this PR.

A new Rails config, to switch to the new ActiveRecord connection handler that _can_ handle multiple database connections, [does not change anything in our app](https://guides.rubyonrails.org/active_record_multiple_databases.html#migrate-to-the-new-connection-handling), but the config needs to be present for later updates. 

The method `Enumerable#sum` will in Ruby 3.0 require an initial value. The value is the default and should match the type of the sum (`array` / `string` / `hash`). This one seems to be an array.

### DIFF
--- a/app/classes/eol_data.rb
+++ b/app/classes/eol_data.rb
@@ -232,7 +232,7 @@ class EolData
   end
 
   def image_id_to_names
-    make_list_hash_from_pairs(name_images_list.sum)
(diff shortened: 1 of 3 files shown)


pellaea:
I guess this is fine.  I'm sure I'll understand this better when it actually makes a difference!

PR #1004, opened 2022-06-11
Convert `for...in` to `each` in views

nimmolo:
One good thing came out of running Brakeman locally - it refused to parse the `for` loops in the views. So here you go... `for` loops are converted to `each` loops.

### DIFF
--- a/app/views/species_list/list_species_lists.html.erb
+++ b/app/views/species_list/list_species_lists.html.erb
@@ -10,7 +10,7 @@
 <%= paginate_block(@pages) do %>
   <% if @objects.any? %>
     <div class="list-group">
-      <% for species_list in @objects.each %>
+      <% @objects.each do |species_list| %>
         <div class="list-group-item">
           <big>
             <%=
--- a/app/helpers/pagination_helper.rb
+++ b/app/helpers/pagination_helper.rb
@@ -7,7 +7,7 @@ module PaginationHelper
   # Wrap a block in pagination links.  Includes letters if appropriate.
   #
   #   <%= paginate_block(@pages) do %>
-  #     <% for object in @objects %>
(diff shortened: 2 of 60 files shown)


app/views/species_list/list_species_lists.html.erb
```diff
@@ -10,7 +10,7 @@
 <%= paginate_block(@pages) do %>
   <% if @objects.any? %>
     <div class="list-group">
-      <% for species_list in @objects.each %>
```
nimmolo, line 13: Not sure I did this one right. Are these equivalent?

nimmolo: After researching a bit, i'm pretty sure these are equivalent.
`for species_list in @objects.each`
`@objects.each do |species_list|`

pellaea:
The only difference, as Joe explained it a short while ago, is subtle: in
the case of the for loop, the accumulator (species_list in your example)
and any local variables used in the body of the for loop are visible
outside the loop.  In the case of the each construct, they are not visible
outside the each block.

But as I was blissfully unaware of this dubious "feature" of for loops, I
never relied on it in my code, so it should not be an issue.

app/views/species_list/list_species_lists.html.erb
JoeCohen: Looks right.

nimmolo:
Thanks for the reviews. Will merge now.

JoeCohen (approved):
Looks good to me. That's a lot of work, but might as well get it right and be consistent with `app`.
NOTE: As a matter of principle, Rubocop refuses to look at `.erb` files. https://github.com/rubocop/rubocop/issues/1113
I wonder if there are other potential problems lurking there?
Maybe down the line we should add a `.erb` linter?

PR #1006, opened 2022-06-12
Brakeman Action

JoeCohen:
This PR uses the Github Action `brakeman-linter-action`, instead of Codeclimate, to run Brakeman.

Codeclimate uses an obsolete version of Brakeman, which gives many false positives, and misses some issues.
`brakeman-linter-action` adds another (5th) CI check. When that check fails, you see a nicely formatted report on Github.


JoeCohen:
Creating a persistent report isn't as trivial as I thought. 
Creating a file and saving it to `tmp` won't work because there's no way to get at it.
On the other hand, I don't want to save it someplace that's tracked by Git.


I'm done for today and will hack on this tomorrow. Anyone who has an idea how to deal with this is welcome to change it.

pellaea:
Can you have it save to a persistent file and add that file to .gitignore?

JoeCohen:
@pellaea - Thanks for the suggestion. I will give it a try.

JoeCohen:
PS: I will try it later. I'm having trouble getting my head around how to do this correctly.
At the moment, I need to work on mindless tasks (e.g. Rubocop), preferably small, mindless tasks.

JoeCohen:
NOTE: The Brakeman check will fail in this PR because there are ~37 outstanding Brakeman issues.
I will merge it despite that failure, then work on fixing those issues.

pellaea (approved):
I know nothing about this, so I trust you, Joe!

nimmolo (approved):

nimmolo:
@JoeCohen  - It seems we're getting two Brakeman reports now. 
Is that just how the new GitHub action works, or do we still need to remove it from CodeClimate?

JoeCohen:
Interesting.
I'm only seeing one Brakeman report: the one from Github Actions.
Where are you seeing two?
I disabled it in Codeclimate.

> plugins:
> # check Rails apps for security vulnerabilities
> brakeman:
> # Do not use Codeclimate's version of Brakeman as of 2022-06-18
> # It is dated and throws many false positives
> # Instead run brakeman locally or in CI rely on Github action plugin
> # https://github.com/devmasx/brakeman-linter-action
> enabled: false
>
> I left it that way hoping (probably in vain) that Codecimate one day will
get the current Brakeman version.
But maybe I need to remove it completely from the Codeclimate config.

nimmolo:
Ah. Just now, im only getting the one. Sorry for false alarm.

JoeCohen:
Super. Glad to know it's performing as expected.

PR #1009, opened 2022-06-12
Force everyone using old API to authenticate, announce deprecation

pellaea:
Instead of messing with routes, I chose to use an even simpler technique: simply force everyone using the old API to authenticate.  This gives us a way to contact anyone using the old API.  I've also overridden the old "must authenticate" error message to explain that we are retiring version 1 by the end of the month.


app/classes/api/errors.rb
```diff
@@ -252,6 +252,16 @@ def keys_for_patch
 
   # Request requires valid ApiKey.
   class MustAuthenticate < Error
+    def to_s
+      "API version 1 is going to retire in July 2022.  Please migrate to
+      version 2.  Superficially, version 2 is very similar.  It mostly involves
+      moving data fields around in the responses.  You may need only change the
```
JoeCohen, line 258:
```suggestion
      moving data fields around in the responses.  You may only need to change the
```

JoeCohen (commented):
Thanks Jason. This sounds much better than messing with the routes.
(Though I'm not sure how it works. No need to explain.)
I'm good with deploying it.

nimmolo:
Woohoo!

JoeCohen:
I just noticed the big drop in coverage (for obvious reasons).
Options:
- Ignore it. This is the simplest. it will fix itself when API is actually ripped out.
- Figure out how to tell Coveralls to exclude `app/classes/api.rb` and `app/classes/api`. IMO not worth the bother.

Oops. Looks like this ship has already sailed. But I'll leave this Comment hear in case it's useful in the future.

pellaea:
I thought "ignore it" was the obvious answer! :)

JoeCohen:
You were right. Sorry to have belabored it.

PR #1011, opened 2022-06-12
Fix Style/StringConcatenation offenses in test

JoeCohen:
This PR fixes the Style/StringConcatenation offenses in `test` and subfolders.
It also includes some manual changes to:
- improve readability; 
- change some string concatenations not flagged by Rubocop.

The PR reduces total number of Rubocop offenses by about 1%.


JoeCohen:
NOTE:  As the first step in this PR, I used Rubocop autocorrect in its default `aggressive` mode.  (`$ rubocop --only Style/StringConcatenation -A test`)
Autocorrect is "unsafe" in this mode because "it cannot be guaranteed that the receiver is actually a string, which can result in a false positive." https://docs.rubocop.org/rubocop/1.29/cops_style.html#safety-44

I then reviewed the corrections to determine if the receivers were strings. (All of them were.)

nimmolo:
Looks good, Joe.

PR #1012, opened 2022-06-12
Rails native `enum` for `Name.rank`

nimmolo:
Note: This PR is based on the Zeitwerk branch, to avoid excessive re-refactoring.
**The big difference here is that Rails native enums return a string rather than a symbol.** 
In other words, with the current `simple_enum` gem,
```
name = Name.last
name.rank
=> :Subgenus
```
**with Rails enum**,
```
name.rank
=> "Subgenus"
```
Tests must therefore compare results to strings. 

So far, `LocationDescription` and `NameDescription` enums have been converted, and are passing tests.


nimmolo:
If either of you has the chance to check this out, I would love a lifeline here. 

I've got the converted enums passing almost all tests, but these 2 fails/2 errors for `Name.rank` are escaping me!

nimmolo:
** NameControllerTest#test_update_add_icn_id**
I think there may be a mistake in the fixtures... or an incompatibility. The issue to investigate is that `test/fixtures/names.yml#stereum_hirsutum` did not have a rank to begin with, nor was there a default value for the enum `rank`. So `NameControllerTest#test_update_add_icn_id` correctly does not find a value for `name.versions.first.rank`, after updating the name version.

pellaea:
Good catch, Nimmo.  I find missing data in the fixtures all the time.
Please fix them whenever you notice them.  Thank you!

PR #1015, opened 2022-06-13
Fixed small bug and added minimal test to expose bug in acts_as_versioned.

pellaea:


pellaea:
Don't mind me, I couldn't figure out how to push my change directly to nimmo's branch because my branch had a different name!  This works instead...

PR #1017, opened 2022-06-13
Fix "conservative" StringConcatenation offenses

JoeCohen:
This PR fixes Style/StringConcatenation offenses 
- discovered by using that cop in "conservative" mode
- autocorrecting them
- fixing new offenses created by the autocorrection

NOTE: Even when "safely" corrected, interpolation performs differently from concatenation when one of the values to be combined is `nil`:
```ruby
> x = nil
 => nil 
> "string" + x
... TypeError (no implicit conversion of nil into String)
> "string#{x}"
 => "string" 
```
---

This PR does NOT fix StringConcatenation offenses found in the default `aggressive` mode. 
Those will be fixed in a later PR(s).

The end game is to (eventually) fix all StringConcatenation offenses 
because interpolation performs better than concatenation. 
(In Ruby 2.x, it performs better when combining > 2 things. In 3.x it's better or equal all the time.)
https://mrod.space/2018/06/24/ruby-concatenation-vs-interpolation.html
https://mrod.space/2020/07/16/ruby-concatenation-vs-interpolation-performance.html


JoeCohen:
Another trap for the unwary (i.e., me).
If we need to modify a string in place, we cannot simply create it from interpolated strings. 
For example, with
```ruby
x = "b"
str = "a#{x}"
```
creates the frozen string `str`.
This contrasts with 
```ruby
x = "b"
str = "a" + x
```
which creates the mutable string `str`.

JoeCohen:
Yes. And I got whiplashed yesterday with one of the changes in this PR. 
Also note that concatenating `nil` throws an error, but interpolating it does not.

pellaea:
Looks good to me.

nimmolo:
Thabks for the heads up about frozen strings Joe. I believe this is part of the transition to Ruby 3 (which 2.7.6 is meant to be preparatory for). Stricter typing and freezes of static objects to save memory.  I've encountered a lot of similar changes. I think it will settle a bitonce we're on 3 and get used to it.

PR #1019, opened 2022-06-13
HABTM > HMT for `observation_images`

pellaea:
This is a test to see if there is any performance hit from converting has-and-belongs-to-many relationships to has-many-through.  The latter is The Way To Do It Now™.  And it allows us to, for example, order an observation's images, which we've never been able to do before.  And it gives us more options for cleaner ActiveRecord queries.

It's a simple migration that mostly involves changing the name of the images_observations table to "image_observations" (so that the glue model is somewhat-sensibly named "ImageObservation" instead of "ImagesObservation").  And it adds an id and rank column to the glue table.

Presently the rank column is unused.


app/models/observation_image.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Glue table between observations and images.
+class ObservationImage < ApplicationRecord
```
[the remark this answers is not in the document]
JoeCohen: False positive generated by Codeclimate's old version of Brakeman. (It does no appear in Brakeman 5.2.3 run locally.) I

app/controllers/name_controller.rb
```diff
@@ -1002,11 +1002,11 @@ def eol_data(review_status_list)
     image_data = Name.connection.select_all(%(
       SELECT name_id, image_id, observation_id, images.user_id,
              images.license_id, images.created_at
-      FROM observations, images_observations, images
+      FROM observations, observation_images, images
       WHERE observations.name_id IN (#{name_ids})
```
JoeCohen, line 1006: Brakeman (current version) complains about this:
>

| Confidence | Class |  Method | Warning Type | Message |
| --- | --- |  --- | --- | --- |
| Medium | NameController | eol_data | SQL Injection | Possible SQL injection near line 1006: Name.connection.select_all("\n SELECT name_id, image_id, observation_id, images.user_id,\n images.license_id, images.created_at\n FROM observations, observation_images, images\n WHERE observations.name_id IN (#{{}.keys.map(&:to_s).join(",")})\n AND observations.vote_cache >= 2.4\n AND observations.id = observation_images.observation_id\n AND observation_images.image_id = images.id\n AND images.vote_cache >= 2\n AND images.ok_for_export\n ORDER BY observations.vote_cache\n ") |

It looks like a false positive to me, but I'm unsure.
It might be fixable by using Arel -- even if a false positive. The Arel is over my head.
```sql
Name.joins(observations: :images).
     where(observations: { name: name_ids }).
     where("observations.vote_cache >= 2.4").
     where("images.vote_cache >= 2").
     where(images: { ok_for_export: true})
```
will sanitize `name_ids`.
But I have no idea how to deal with the `SELECT` part of the query.

nimmolo:
@JoeCohen - thanks for reviewing. I will rewrite that controller method. (in about two hours when i'm back)

app/controllers/name_controller.rb
nimmolo: @JoeCohen shouldnt this method really be in the model?

JoeCohen:
Another thing I just noticed:
```ruby
vagrant@ubuntu-focal /vagrant/mushroom-observer (has_many_through-everything-but-descriptions)
$ bundle install
Your Ruby version is 3.0.4, but your Gemfile specified 2.7.6
```

nimmolo:
Right- good catch!. I was picturing these merging before ruby 3. Will update.

nimmolo:
@JoeCohen Let's tell codeclimate to ignore this eol_data method in NameController, after all.

Apart from the EOL Data-maybe-broken question, I'm realizing there are a lot of query methods that I'd like to move to the models, as i'm reading through the name controller and other controllers.

This is not just being OCD about separation of concerns. GraphQL does not have access to controller methods, but it does have access to model methods, and these queries may be useful in the new interface. I think that will be my next series of PRs.

As for the vagrant message, I think for the immediate future, you'll just have to tell Vagrant which version to use every time you switch branches if they have different Ruby versions. This HMT branch has Ruby 3.0.4 set in .ruby-version, so I believe we're good there.

JoeCohen:
Thanks @nimmolo.
I told Codeclimate to shut up.
As for the Vagrant message, I think I will locally 
1. Merge has_many_through-images_observations into has_many_through-everything-but-descriptions
2. Merge has_many_through-everything-but-descriptions into has_many_through-descriptions
That should take care of the ruby version issue and some other things in order to let me run the branches to do some extra testing.

nimmolo:
Ok. In the last 20 minutes or so i’ve merged the first HMT observation_images into master,
and merged master into the remaining two HMT branches and resolved conflicts.
Pulling from origin may clear this up for you more quickly at this point. Sorry i didn’t get your email before I did that.

nimmo

JoeCohen:
Thanks. I'll do that.

PR #1020, opened 2022-06-14
HABTM > HMT for others: `glossary_term_image`, `herbarium_curator` etc, except `*_description_*`

nimmolo:
This is just a way to split them up. Next PR will do the descriptions.

**This PR should probably merge after #1019, but before #1021**.


nimmolo:
**Scrutiny requested**:
Please check the `FIXME` at the top of `GlossaryTerm`.

pellaea:
Having lived with rails for way too long, I now have a healthy(?) dislike
of automagic, so I'd just as soon have clients of the images association
explicitly request the order they want anyway.  But wouldn't the correct
syntax be:

has_many(:images, through: :glossary_term_images, order: -> {
order(vote_cache: :desc })

Or something like that.  I can't get apidock to load.

nimmolo:
@pellaea apidock is a very slow website, way slower than GitHub for me. More to the point, https://api.rubyonrails.org is always faster to load for me.

Here's their v.6.1 description of what they call a `has_many` [scope](https://api.rubyonrails.org/v6.1.5.1/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many). It does not seem to use the key syntax you're thinking of, either in their examples or elsewhere in our code.

nimmolo:
But - more importantly, I agree about having the methods call their scope order explicitly, and just deleting this automatic scope. There must only be a couple of places calling GlossaryTermImage.

app/models/species_list.rb
```diff
@@ -121,24 +125,7 @@ def clear
...
-    # puts(delete_manager.to_sql)
-    SpeciesList.connection.delete(delete_manager.to_sql)
-  end
-
-  def arel_delete_observations_species_lists(id)
-    osl = Arel::Table.new(:observations_species_lists)
-    Arel::DeleteManager.new.
-      from(osl).
-      where(osl[:species_list_id].eq(id))
+    SpeciesListObservation.where(species_list_id: id).delete_all
```
pellaea, line 128: Nice!!!!

pellaea (approved):

nimmolo:
Ok, just re-introduced the GlossaryTerm#images order at the controller level.

app/controllers/species_list_controller.rb
```diff
@@ -485,7 +485,7 @@ def do_add_observations(species_list, query)
     # large species_lists, such as "Neotropical Fungi".
     # species_list.observation_ids += ids
     Observation.connection.insert(%(
```
JoeCohen, line 487:
```suggestion
    SpeciesListObservation.insert_all(
      ids.map { |id| { observation_id: id, species_list_id: species_list.id } }
    )
```
This fixes the Codeclimate/Brakeman SQL injection issue:
>Possible SQL injection NEW
        #{ids.map { |id| "(#{id},#{species_list.id})" }.join(",")}
Severity: MinorFound in [app/controllers/species_list_controller.rb](https://codeclimate.com/github/MushroomObserver/mushroom-observer/app/controllers/species_list_controller.rb?from_sha=6f1e5ccb&to_sha=ce0aecf9) by [brakeman](https://docs.codeclimate.com/docs/brakeman)

app/controllers/species_list_controller.rb
```diff
@@ -485,7 +485,7 @@ def do_add_observations(species_list, query)
     # large species_lists, such as "Neotropical Fungi".
```
JoeCohen, line 485: @pellaea Does the comment about inefficiency refer to `species_list.observation_ids += ids`? 
If yes, we can get rid of it.

JoeCohen (changes requested):
Very nice.
Thanks for all the work on this.

JoeCohen:
I'll see if there's a way to deal with the codeclimate "Mass assignment is not restricted using attr_accessible" issues via codeclimate configuration. 
If not, I'll tell Codeclimate (via the UI) that they are Invalid.

JoeCohen:
Pure curiosity question:
Why did you reverse the name of some of the join tables? Was that simply a matter of which models you dealt with first?

app/controllers/species_list_controller.rb
nimmolo: Yes! Nice one.

nimmolo:
@JoeCohen I tried to put the "mother" model (conceptually speaking) first. But let me know if i didn't get it right in some cases. 

For example, in the case of images, these are definitely ancillary to observations and glossary terms, to me.
Likewise, observations are a constituent of species lists, so: species_list_observations.
And to my thinking projects own species lists, more than the other way around. so: project_species_lists 
But now that i think about it, maybe that's debatable!

JoeCohen (approved):

app/controllers/species_list_controller.rb
pellaea: Yes, exactly.  I wanted to leave that in there but commented out so we would know not to try it again.  Or at least be sure to benchmark it before adding it back in.  The alternative is ugly, but was way faster.

pellaea:
Oh, okay, I see.  You just need to make it the second argument, then:

has_many(:images, -> { blah blah blah }, option1: val1, option2: val2, ...)

Does that work?

JoeCohen:
Thanks for the explanation.

PR #1021, opened 2022-06-14
HABTM > HMT for `descriptions`-`users`-`user_groups` join tables

nimmolo:
**This PR should probably merge after #1019 and #1020**


nimmolo:
Scrutiny requested: 
I redid the arel to AR in `Description#group_user_ids` and `#group_users`, but the `group_user_ids` has a double `join` with `pluck`. I think it should have an `includes` with `select` and an `each` block instead, for a more efficient query, but i wasn't able to get the syntax right... there are too many levels of association for me to get my head around!

app/models/description.rb
```diff
@@ -401,34 +401,24 @@ def group_users(table)
...
   # Do minimal query to enumerate a list of groups.  Return as an Array of ids.
   # Caches result.  (Equivalent to using <tt>association.ids</tt>, I think.)
   def group_ids(table)
     @group_ids ||= {}
     @group_ids[table] ||=
-      self.class.connection.select_values(select_group_ids(table).to_sql)
+      table.to_s.classify.constantize.
+        where("#{type_tag}_id" => id).
+        order(user_group_id: :asc).distinct.
+        pluck(:user_group_id)
```
pellaea, line 419: This is so much nicer!  Big win for the conversion to HMT!

app/models/user_group.rb
```diff
@@ -42,10 +42,34 @@
...
+  has_many :names_admined, through: :name_description_admins,
+                           source: :name_description
+
+  has_many :name_description_writers, dependent: :destroy
+  has_many :names_written, through: :name_description_writers,
+                           source: :name_description
+
+  has_many :name_description_readers, dependent: :destroy
+  has_many :names_read, through: :name_description_readers,
+                        source: :name_description
```
pellaea, line 69: Haha, this gets a bit excessive, doesn't it??  Works for me, though.

pellaea:
Looks good.  (How do you officially approve a PR?  I saw you do this on one of Joe's, Nimmo.)

nimmolo:
Thanks, @pellaea ! It's a button on the "files_changed" tab on the GitHub website itself. 
I'm using their gui all the time, idk if it will be available via CL. I understand github.com is slow to load for you!

pellaea:
Thanks, I didn't look for it there.

By the way, should we close my test PR for images_observations?  Is that
part of one of the other two PRs?

nimmolo:
No - that one should merge first. I did some tweaking on it, it’s good to go (after review).

nimmo

pellaea:
I'm fine with all three.  Merge at your pleasure.

nimmolo:
Thanks @pellaea. I'll wait for @JoeCohen's read...

nimmolo:
Does it make sense, in another PR, to do migrations to give the versioned tables a `primary_key`, and create models for them?

That would allow us to always use table.singularize.classify.constantize with AR methods, and eliminate those fancy Arel methods in User#delete.

pellaea:
It's not obvious this is a good idea, actually, as acts_as_versioned
creates these models itself.  Why not just use its models?  Name::Version,
etc., I think.

nimmolo:
Ah... I didn't realize those were the model names! Ok.
It'll be a bit more complicated logic to get the model name from a list of table names in that case, but doable.

pellaea:
Is this for SiteData?

nimmolo:
No it's for User#delete

pellaea:
Thanks.  Hmmm.  Looking at that code, it doesn't look like anything needs
to be done at all.  Am I missing something?

Note that versions are not HABTM relationships, just plain old has_many.

app/models/location_description_admin.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Glue table between location_descriptions and user_groups.
+class LocationDescriptionAdmin < ApplicationRecord
```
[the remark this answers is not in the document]
JoeCohen: Another codeclimate false positive due to old Brakeman version.
One of us can fix this through the codeclimate UI after the conflicts are resolved and we can see the Codeclimate check.

JoeCohen (commented):
It looks good. I don't have any suggestions.

nimmolo:
@JoeCohen, @pellaea  - I just pushed a commit to remove the Arel methods from `User.erase_user`. 
Now that everything's HMT we can call those tables directly with AR 🥇

JoeCohen:
The PR currently says
>[nimmolo](https://github.com/nimmolo) wants to merge 23 commits into [has_many_through-everything-but-descriptions](https://github.com/MushroomObserver/mushroom-observer/tree/has_many_through-everything-but-descriptions) from [has_many_through-descriptions](https://github.com/MushroomObserver/mushroom-observer/tree/has_many_through-descriptions)

Is that still corect?

nimmolo:
@JoeCohen - no those are merged. Try pull origin again...

JoeCohen:
So should the PR should be edited to make the base branch `master`. I.e., merge this branch into `master`.

nimmolo:
Whooops! Yes!!

JoeCohen:
I restarted the CI test job hoping (illogically) that Github Actions would run all 4 checks. 
But we're still missing two checks:  test (pull_request), and codeclimate.
I just pushed a trivial change to see if that gets all 4 checks going.
If not, we should probably close this PR and open a new one.

JoeCohen:
Oops. Pushed to the wrong (closed) HABTM branch. 
I'm trying again.

JoeCohen:
That worked.

JoeCohen:
I told Codeclimate that the mass assignment issues were Invalid.
The current version of Brakeman (run locally) does flag the SQL Injection issues, but with weak confidence:

| Confidence | Class | Method | Warning Type | Message |
| --- | --- | --- | --- | --- |
| Weak | Description | group_user_ids | [SQL Injection](https://brakemanscanner.org/docs/warning_types/sql_injection/) | Possible SQL injection near line 406: table.to_s.classify.constantize.joins(:user_group => :user_group_users).where("#{type_tag}_id" => id) |
| Weak | Description | group_ids | [SQL Injection](https://brakemanscanner.org/docs/warning_types/sql_injection/) | Possible SQL injection near line 417: table.to_s.classify.constantize.where("#{type_tag}_id" => id) |

JoeCohen:
I'm gone for an hour. Then can work on this for an hour. Then really gone all day Thu. Around Fri.

nimmolo:
@JoeCohen I am pretty sure they're ok, but maybe these two should be private methods?

JoeCohen:
I think the two SQL Injection issues are false positives. I don't see how either `type_tag` or `id` can come from a user.
If you agree, then let's completely ignore those issues in this PR (don't let them block a merge), and I'll add them to the `brakeman.ignore` file in another PR.

pellaea:
Agreed, those are both definitely safe.

PR #1022, opened 2022-06-14
Fix "aggressive" StringConcatenation offenses in classes

JoeCohen:
This PR fixes Style/StringConcatenation offenses in `app/classes` in "aggressive" mode.
It follows up on  PR #1017 (which fixed only the offenses that could be found in "conservative" mode).

I'm breaking the fix of remaining offenses into (artificially) small PRs because to facilitate review and because (per Rubocop) the cop is less safe in "aggressive" mode.


nimmolo:
That Query::LocationPatternSearch#initialize_flavor `CONCAT` `COALESCE` is too mind-boggling, but the rest of 'em seem a-OK!

nimmolo (approved):

JoeCohen:
Thanks @nimmolo!
The change I made to that query does seem wrong. I'll have a closer look.

nimmolo:
I don't believe it's necessarily wrong, I just don't get it!

JoeCohen:
I have some extra newlines in the SQL, but they shouldn't make a difference.
However, to be on the safe side, I'll get rid of them.

JoeCohen:
It was right, but still mind-boggling.
I added some line breaks and spacing, which hopefully make it more readable (but the boggle factor is unchanged).

app/classes/auto_complete.rb
```diff
@@ -100,7 +100,7 @@ def refine_matches
         return used + part if matches.length <= limit
       end
       if n < words.length
-        used += word + " "
+        used += "#{word} "
```
pellaea, line 103: I'm really surprised rubocop doesn't require: used = "#{used}#{word} ".  Likewise above used + part... I guess it just doesn't know that used is a String?  Not that I'm complaining, mind you!  I think the string interpolation version of concatenation is ugly...

app/classes/geocoder.rb
```diff
@@ -67,7 +67,7 @@ def set_extents(north, south, east, west)
   end
 
   def ajax_response
-    [north, south, east, west].join("\n") + "\n"
+    "#{[north, south, east, west].join("\n")}\n"
```
pellaea, line 70: (case in point! yikes! how important is that tiny bit of performance?)

app/classes/query/location_pattern_search.rb
```diff
@@ -14,10 +14,11 @@ def initialize_flavor
...
-      LocationDescription.all_note_fields.map do |x|
-        "COALESCE(location_descriptions.#{x},'')"
-      end.join(",") +
-      ")"
+    "CONCAT(" \
+      "locations.name," \
+      "#{LocationDescription.all_note_fields.map do |x|
+           "COALESCE(location_descriptions.#{x},'')"
+         end.join(",")}" \
+    ")"
```
pellaea, line 22: sorry, my mistake, *this* is case in point! :)

pellaea (approved):

pellaea:
This all looks correct and safe and proper.  I only wish it were possible to make correct code more readable!  But I think you've done as good a job as possible.

JoeCohen:
@pellaea: Thanks very much for taking the time to review. 
I agree that the interpolation format is sometimes ugly. But I'd like to leave he cop on, and would rather have an ugly format than the awkward enabling/disabling of the cop.

pellaea:
Agreed that lots of disabling/enabling cops is a bad idea!  I can live with
the cop.

PR #1030, opened 2022-06-17
ApplicationHelper: Allow for an array of refactored controllers in link_next/prev

nimmolo:
This PR doesn't change much, it just allows for adding other controllers to an array of `REFACTORED_CONTROLLERS` that will accept a `flow` `next`/`prev` `param`. The goal here is to start allowing for other controllers to emulate  @JoeCohen 's refactor of `HerbariaController`. That moves routes to default `resources` routes like `show` and `index`, and provides redirects for the old routes. 

Routes in a potential GraphQL app would ideally follow a similar regular pattern, so I'm thinking refactoring the controllers and updating paths ahead of time would "clear the paths" for minimal eventual URL disruption in the future. Hopefully it'll get easier following Joe's example.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -93,8 +93,8 @@ module ApplicationHelper
 
   # link to next object in query results
   def link_next(object)
-    path = if REFACTORED_CONTROLLERS.include? object.type_tag
-             send("#{object.type_tag}_path", { object.id, flow: "next" })
+    path = if REFACTORED_CONTROLLERS.include?(object.type_tag)
+             send("#{object.type_tag}_path", object.id, flow: "next")
            else
              { controller: object.show_controller,
                action: object.next_action, id: object.id }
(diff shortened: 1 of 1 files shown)


pellaea (approved):

nimmolo:
@JoeCohen - Does this look ok to you? It just allows me to extend your `flow` param to other controllers and get a default `next`/`prev`. I used it in the UsersController.

JoeCohen:
I'll try to look at it tonight in about an hour, or tomorrow.
(I'm having trouble keeping up with all your work.)

JoeCohen (approved):
I think I commented elsewhere about finding a more specific name for `REFACTORED_CONTROLLERS`.
Otherwise, this looks great.

nimmolo:
Good call Joe. Renamed that constant to REFACTORED_CONTROLLERS_WITH_FLOW

PR #1032, opened 2022-06-17
1 `UsersController``ChecklistsController`

nimmolo:
This converts UsersController to `resource` style routes `show` and `index`. Since `edit` and `update` are not being used, I just routed the `change_user_bonuses` to use these two default actions. It's much simpler to deal with. (Users edit their own profiles via the AccountController.) There's an alternate index for users too, `by_contribution`, accessed by a non "resource" route.

Checklists were moved to their own controller and route. 

Links for Users and Checklists were converted to use the path helpers, rather than `controller:`/`action:` keywords. Unfortunately, i got over-enthusiastic and converted some others; most are reverted. (Notably **not** reverted in the nav and searchbar partials, and in the Users views, where they seem to be OK.)


nimmolo:
@JoeCohen @pellaea Should `checklist` have its own controller, not under Users? 
It seems like it's not only a User thing, it could be for Project, SpeciesList, or the site.
Or is it always tied to a user?

pellaea:
Sure, whatever makes the most sense to you.

nimmolo:
Ugh. Close!

- With default resources, it seems Rails won't route a request for a single user without the ID. So the guard clause in #show (redirecting to #index), as well as the `test_user_no_id`, seem superfluous.
- Trying to figure out what `/users/#{id}/bonuses` should be doing, i believe just editing the bonuses string. I haven't quite figured out how to configure a controller like that, yet. I'm trying to understand how Joe built Herbarium's `request_to_be_curator`
- I don't get the error `ObserverControllerSupplementalTest#test_destroy_observation_from_search`. It's expecting the short `mushroomobserver.org/#{id}` URLs, but some toolbar or other isn't providing it
- Likewise `ObserverControllerTest#test_lookup_user` - it's expecting a different URL or controller action for Einstein.

pellaea:
re: first point about show user without an ID is now superfluous -- That
sounds plausible and just fine.  Whatever error the router gives is almost
certainly acceptable.  It should never occur in practice.  The point of the
guard clause was to prevent a "crash" if it ever arose., resulting in a
cron email report.  For a while I religiously whacked every single one of
those moles.  I'm less concerned nowadays.  It was never going to actually
cause any problems -- e.g. security hole or corrupt data or anything
actually substantive.

nimmolo:
Thanks... all suggestions welcome. 
This has been waaaay more trouble than I expected, nearly 3 days of work! I'm going out for a bike ride. 
But i'm very close, compared to the number of errors at the beginning.

app/controllers/location_controller.rb
```diff
@@ -599,8 +599,8 @@ def create_location
           if (user = User.safe_find(@set_user))
             user.location = @location
             user.save
-            redirect_to(controller: :observer,
-                        action: :show_user,
+            redirect_to(controller: :users,
+                        action: :show,
```
pellaea, line 603: Why isn't this user_path(@set_user) as elsewhere above?

app/controllers/users/bonuses.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+        # Add new bonuses
+        bonuses.each do |(points, _reason)|
+          contrib += points
+        end
+        # Update database.
+        @user2.bonuses      = bonuses
+        @user2.contribution = contrib
+        @user2.save
+        redirect_to(action: "show", id: @user2.id)
```
pellaea, line 65: Another case where I would have expected to use user_path() to be more consistent.  Can you help me understand why we use user_path() in some places but not others?  I haven't figured out the pattern yet.

app/controllers/users_controller.rb
```diff
@@ -1,40 +1,90 @@
...
-module ObserverController::UserController
+class UsersController < ApplicationController
+  # These need to be moved into the files where they are actually used.
+  require "find"
+  require "set"
+
+  before_action :login_required
+  before_action :disable_link_prefetching, except: [
+    :show,
+    :by_contribution
```
pellaea, line 11: Sorry to be picky, but show user is an expensive page.  I might prevent prefetching it, too.  Not sure about by_contribution.  Probably better to disable it, too.  Certainly, I don't expect it to be a common work flow!!

app/controllers/users/bonuses.rb
nimmolo: Highest probability: I missed it. I changed hundreds of paths, back and forth many times. This is the case here.

Small possibility: using the path without the specific builder keys sometimes removed the query param from the path, causing tests to fail. I was too bleary to notice where and when. Not in this case, though.

app/controllers/users_controller.rb
```diff
@@ -1,40 +1,90 @@
...
       show_selected_users(query)
     else
       flash_error(:permission_denied.t)
-      redirect_to(action: "list_rss_logs")
+      redirect_to("/")
     end
   end
 
+  # by_contribution.rhtml
+  def by_contribution
```
pellaea, line 44: Oh yes, if it's using SiteData, it's got to be expensive, therefore we should not allow it to be prefetched.

app/controllers/users_controller.rb
nimmolo: Please be picky!

app/views/publish_name_email/build.html.erb
```diff
@@ -11,7 +11,7 @@
   links.push([:email_links_show_object.t(type: :name),
     "#{MO.http_domain}/name/show_name/#{@name.id}"])
   links.push([:email_links_email_publisher.t,
-    "#{MO.http_domain}/observer/show_user/#{@publisher.id}"])
+    "#{MO.http_domain}/users/#{@publisher.id}"])
```
pellaea, line 14: I'm confused.  With name and image controllers the path is just "/name" and "/image", respectively.  Why is it plural for the user controller?  Is this something to do with user controller routing differently than the other controllers?

app/views/users/show.html.erb
```diff
@@ -5,53 +5,52 @@
...
+      :location_descriptions_by_editor ],
+    [ :locations, :location, :locations_by_user ],
+    [ :locations_versions, :location, :locations_by_editor ],
+    [ :name_description_authors, :name, :name_descriptions_by_author ],
+    [ :name_description_editors, :name, :name_descriptions_by_editor ],
+    [ :names, :name, :names_by_user ],
+    [ :names_versions, :name, :names_by_editor ],
+    [ :observations, :observer, :observations_by_user ],
+    [ :species_lists, :species_list, :species_lists_by_user ],
+    [ :life_list, :checklists, :show ]
```
pellaea, line 23: You don't like lined up table columns?  Or is this rubocop's preference?

app/views/publish_name_email/build.html.erb
nimmolo: It's like Joe's "/herbaria/" -- it's because they're the Rails "resource" CRUD routes. They seem to default to the plural.

app/views/users/show.html.erb
nimmolo: Lines were too long, so it was a train wreck in my editor. (It forces line-wraps)

app/views/publish_name_email/build.html.erb
pellaea: Ah!  Understood.  Thank you!

app/views/users/show.html.erb
pellaea: I thought it might be something like that.  No worries.

pellaea:
By the way, Nimmo.  I'm totally fine with inconsistency in using the three forms:

1. user_path(...)
2. action: "users", ...
3. "/users/..."

I just want to make sure I didn't misunderstand something more subtle that required one version over another.  And the query_params() mechanism is potentially one of those subtle things.  Point taken.  I just want to make sure I understand these issues, that's all.  ...To the extent that my poor little brain can keep up with all these fancy shmancy changes!

pellaea:
By the way, Nimmo.  I'm totally fine with inconsistency in using the three forms:

1. user_path(...)
2. action: "users", ...
3. "/users/..."

I just want to make sure I didn't misunderstand something more subtle that required one version over another.  And the query_params() mechanism is potentially one of those subtle things.  Point taken.  I just want to make sure I understand these issues, that's all.  ...To the extent that my poor little brain can keep up with all these fancy shmancy changes!

For the record, I think overall this is a great improvement in readability.  And presumably maintainability and consistency with standards and all that good stuff.  I'm sorry it has been so much work.  But I'm not surprised.  These are pretty fundamental changes.

nimmolo:
many of my problems were "own goals", in retrospect. Refactoring other things on the fly beyond the scope, then having to undo them. I'm pretty sure `user_path` works in all cases, for this refactored controller.

Einstein (in the controller test) remains a mystery. I don't get why he wants to go to the nonexistent route `users/index_rss_logs`

pellaea:
I can take a look if you like.

nimmolo:
That would be great.  The last 3 controller test issues in that list are really stumping me, in some cases because i don't understand how they're getting what they're getting.

pellaea:
Okay, I'm stuck installing and stripping a new database snapshot locally
for the time being.  But soon as that's done I'll take a look.

pellaea:
re: Einstein -- The problem is that lookup_xxx makes a special exception for user because for whatever reason we didn't want users to be able to request an index of all the users on MO.  So if the user for lookup_user is not found, it sends the requester to the home page -- /observer/list_rss_logs.  That code is no longer working because it's using the "show_controller" which is /users, so it is trying to send the user to /users/list_rss_logs.  See app/controllers/observer_controller/markup_controller.rb line 98.  It just needs a more complete separate clause for User, I think.  I'll push a quick fix.

pellaea:
I'm running the tests now to see where the remaining two test failures are...

pellaea:
I don't know how to fix the edit_user_bonus_path error.  It doesn't seem to understand that the bonuses controller is under the users controller.  I've never dealt with this mechanism, so I'm no help.

app/controllers/location_controller.rb
nimmolo: Changed to user_path(@set_user.id)

app/controllers/users/bonuses.rb
nimmolo: Yep i missed it. Changed

app/controllers/users_controller.rb
nimmolo: Removed

nimmolo:
Ok. Here's where i'm stuck now. 
I've got the methods `#edit_bonuses` and `#update_bonuses` back in the `UsersController` proper.

I've done the routes like this:
```
vagrant@ubuntu-focal:/vagrant/mushroom-observer$ rails routes -c=UserController
               Prefix Verb   URI Pattern                      Controller#Action
users_by_contribution GET    /users/by_contribution(.:format) users#by_contribution
        users_by_name GET    /users/by_name(.:format)         users#by_name
         bonuses_user GET    /users/:id/bonuses(.:format)     users#edit_bonuses
                      POST   /users/:id/bonuses(.:format)     users#update_bonuses
                      PATCH  /users/:id/bonuses(.:format)     users#update_bonuses
                users GET    /users(.:format)                 users#index
                      POST   /users(.:format)                 users#create
             new_user GET    /users/new(.:format)             users#new
            edit_user GET    /users/:id/edit(.:format)        users#edit
                 user GET    /users/:id(.:format)             users#show
                      PATCH  /users/:id(.:format)             users#update
                      PUT    /users/:id(.:format)             users#update
                      DELETE /users/:id(.:format)             users#destroy
```

but when i call `get(:edit_bonuses)` in the controller test, I get 
```
RuntimeError: Wrapped undumpable exception for: ActionController::UrlGenerationError: No route matches {:action=>"edit_bonuses", :controller=>"users"}
            test/functional_test_case.rb:22:in `get'
            test/controllers/users_controller_test.rb:167:in `test_change_bonuses'
```
and in #test_change_bonuses, 
```
    # Prove that non-admin cannot change bonuses and attempt to do so
    # redirects to target user's page
    login("rolf")
    get(:edit_bonuses)
    assert_redirected_to(user_path(user.id))
```
If I add the `id` param to `get(:edit_bonuses, { id: user.id })` i get `too many arguments`.

nimmolo:
It occurs to me that maybe (aside from the user bonuses) I actually don't need the put, post, patch, delete routes for `UsersController`. Ah! What if the bonus actions WERE the edit routes for UsersController?!

nimmolo:
Nope... Even that gets
```
Minitest::UnexpectedError:         
RuntimeError: Wrapped undumpable exception for: ActionController::UrlGenerationError: 
No route matches {:action=>"edit", :controller=>"users"}
            test/functional_test_case.rb:22:in `get'
            test/controllers/users_controller_test.rb:167:in `test_change_bonuses'
```
This is a fresh crotte de cheval, franchement. The route is definitely there. Calling it a night.

nimmolo:
It just had to be a params hash with the key `params: {}`!

nimmolo:
Regarding the broken paths in the nav partials - 

That's the kind of on-the-fly refactoring i was doing and had to revert — and I also meant to revert those. Basically, I'm chastened from converting anything else to use the path helpers until I'm refactoring _the relevant_ controller and I know how those specific path helpers are working.

I'll be happy to revert those too, if they're not working.
Can test them tomorrow. I believe the integration tests spot check some of them.

pellaea:
Understood.  Integration only spot checks a couple, and that's how I
noticed that any were broken.  But hardly reliable for all.

JoeCohen:
Should master be merged into this branch, then migrated
in order to include the new join tables in the schema?

test/controllers/checklists_controller_test.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+# Controller tests for checklists
+class ChecklistsControllerTest
```
JoeCohen, line 6:
```suggestion
class ChecklistsControllerTest < FunctionalTestCase
```

nimmolo:
@pellaea - re: paths
I'd like to review how they work, and note it in the discussion thread — do I have this right? 
What i understood is the old paths need the `id` (or whatever param) as a named parameter, while the new paths take it as an argument. so it has to be:
`observer_do_something_path(id: obs.id)`
but
`user_path(user.id)` or even `user_path(@user)`?
That's why i believe my nav links broke - I didn't name the param.

And then the last thing i discovered is that under `functional_test_case`, when you `get(:action)` or `post(:action)`, any params must be in a hash named `params`. Not just a hash! So:
`get(:edit, params: { id: user.id })`

@JoeCohen  - This is a pull off the `next` branch, which has zeitwerk and enums. So I believe the more efficient, and less git-confusing, way to go would be to keep `next` up to date with `master` (i.e. merge `master` into `next`), rather than this PR branch.

nimmolo:
Informal poll... do you all prefer the route `users/by_contribution` or the simpler `contributors`?
Since it's a separate controller now, it's not any more difficult to do.

nimmolo:
Update: I tested all the left-nav and searchbar links in the front end, everything is working on this branch.

pellaea:
> @pellaea - re: paths I'd like to review how they work, and note it in the discussion thread — do I have this right? What i understood is the old paths need the `id` (or whatever param) as a named parameter, while the new paths take it as an argument. so it has to be: `observer_do_something_path(id: obs.id)` but `user_path(user.id)` or even `user_path(@user)`? That's why i believe my nav links broke - I didn't name the param.

Yes, I was also getting confused about when to name the parameter explicitly and when not to.  Thanks for catching the additional need for params: {} in get/post in the controller tests!  So, can you help me understand the path helpers even more?  Are there two general forms:

1. controller_action_path(param: hash) -- e.g., observer_how_to_help_path()
2. resource_method_path(object_or_id) -- e.g., user_path(252) or ... user_post_path(form_data?)???

Thanks!  I'm sure I could look it up but I'm lazy and I can barely keep up with my email even without adding homework to the task! :)

pellaea:
> Informal poll... do you all prefer the route `users/by_contribution` or the simpler `contributors`? Since it's a separate controller now, it's not any more difficult to do.

I do like your "contributors" but... I have to confess that "by_contribution" is more MO-y and more clear that it's just an index with a custom sort order.  But it's a very weak preference.  This is a very minor point to me.

JoeCohen:
Would it be possible to get `gem("byebug")` restored under 
```ruby
group :test, :development do
```
It's still in master, but at some point got removed or moved on the road to this branch.
I use it a lot. (I'm kind of hopeless without it. Of course, some might say I'm hopeless with it.)

pellaea:
Definitely.  I thought we had agreed to do that.  Why not make a PR against
"next" to make sure that happens?

JoeCohen:
by_contribution

nimmolo:
Joe, I had deleted "byebug" in the "next" branch because in your email to me, where you were helping me with debugging, you mentioned:

> Big Problem: I just realized that byebug is incompatible with Zeitwerk. https://github.com/rails/rails/pull/43187. Rails 7 is switching to Ruby native debug. Looks like we need to do the same.

So, the `next` branch currently incorporates Zeitwerk. But I don't want to make any unnecessary changes to the Gemfile! I just restored it on `next`. If it doesn't work on that branch, i'll leave it to you to delete.

nimmolo:
@pellaea re: path helper formats — I'll respond in the discussion!

pellaea:
Sounds good... but where is the discussion??  Is that supposed to be one of the tabs on the PR page on github?

JoeCohen:
I completely forgot about that.
I'll find a substitute. Almost certainly the "debug" gem, which is (or was
supposed to be) bundled with Ruby 3.1.

nimmolo:
@pellaea Agreed. I apologize — That was not at all the right way for me to do a gem change. It was something I did locally in a debugging delirium, trying out the gem `pry` as a replacement, and I did not mean to have become permanently part of this branch. 

Everyone obviously would need to discuss implications, potential replacements, workarounds, etc.

JoeCohen:
@pellaea: No worries.
1. It reminded me about the impending change, and makes me think about how to move forward.
2. I can always temporarily install it on a branch I'm working on. (That's what I did with this branch.)

@nimmolo: Have you by any chance tried the VSCode rdbg Ruby Debugger extension?

nimmolo:
@JoeCohen No i haven't! I also still don't have rubocop working on vscode locally. 
This week, i'll ask you about related things on email if that's ok.

JoeCohen:
> 



> @JoeCohen No i haven't! I also still don't have rubocop working on vscode locally. This week, i'll ask you about related things on email if that's ok.

@nimmolo That would be great. I'll be in town but may take a while to get back to you because of visiting family.

app/controllers/users_controller.rb
```diff
@@ -1,40 +1,140 @@
 # frozen_string_literal: true
```
JoeCohen, line 1: FYI
It might be appropriate to deal with some of the following in this PR.
(I'm too tired to dig into them tonight.)
```sh
$ rubocop --ignore-parent-exclusion app/controllers/users_controller.rb -A
Inspecting 1 file
C

Offenses:

app/controllers/users_controller.rb:48:3: C: Metrics/AbcSize: Assignment Branch Condition size for show is too high. [<9, 17, 7> 20.47/20]
  def show ...
  ^^^^^^^^
app/controllers/users_controller.rb:84:5: C: Style/GuardClause: Use a guard clause (return unless @user2.bonuses) instead of wrapping the code inside a conditional expression.
    if @user2.bonuses
    ^^
app/controllers/users_controller.rb:86:17: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over unannotated tokens (like %s).
        format("%-6d %s", points, reason.gsub(/\s+/, " "))
                ^^^^
app/controllers/users_controller.rb:86:22: C: Style/FormatStringToken: Prefer annotated tokens (like %<foo>s) over unannotated tokens (like %s).
        format("%-6d %s", points, reason.gsub(/\s+/, " "))
                     ^^
app/controllers/users_controller.rb:92:3: C: Metrics/AbcSize: Assignment Branch Condition size for update is too high. [<16, 30, 9> 35.17/20]
  def update ...
  ^^^^^^^^^^
app/controllers/users_controller.rb:112:5: C: Style/GuardClause: Use a guard clause (return if errors) instead of wrapping the code inside a conditional expression.
    unless errors
    ^^^^^^
app/controllers/users_controller.rb:144:3: C: Metrics/AbcSize: Assignment Branch Condition size for show_selected_users is too high. [<4, 23, 8> 24.68/20]
  def show_selected_users(query, args = {}) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/controllers/users_controller.rb:144:3: C: Metrics/MethodLength: Method has too many lines. [32/30]
  def show_selected_users(query, args = {}) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

JoeCohen (changes requested):

app/views/checklist/show.html.erb
```diff
@@ -4,8 +4,7 @@
   if @show_user
     @title = :checklist_for_user_title.t(user: @show_user.legal_name)
```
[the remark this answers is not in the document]
nimmolo: Yes! fixed now

test/controllers/checklists_controller_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  # Prove that Life List goes to correct page which has correct content
+  def test_checklist_for_user
+    login
+    user = users(:rolf)
+    expect = Name.joins(observations: :user).
+             where(Observation[:user_id] == user.id).
+             where(Name[:rank] == Name.ranks[:Species]).distinct
+
+    get(checklist_path(user_id: user.id))
```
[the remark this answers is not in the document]
nimmolo: @JoeCohen I think i was working on this about the same time you were writing... I reverted back to this.
The path helpers don't work in controller tests because we are overriding `get` and `post` with our own methods. (Or they may not work in `FunctionalTestCase` at all... but it's limited to controller tests, thankfully.)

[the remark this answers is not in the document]
JoeCohen: That makes perfect sense. (The helpers do work in contexts other than the overridden methods.)

test/controllers/checklists_controller_test.rb
```diff
@@ -0,0 +1,77 @@
...
+    prove_checklist_content(expect)
+  end
+
+  # Prove that Project checklist goes to correct page with correct content
+  def test_checklist_for_project
+    login
+    project = projects(:one_genus_two_species_project)
+    expect = Name.joins(observations: :project_observations).
+             where(ProjectObservation[:project_id] == project.id).
+             with_rank("Species").distinct
```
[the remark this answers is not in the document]
nimmolo: @JoeCohen - I'm happy with the AR, and I want to commit it to have an example. I didn't even know how to write that in AR!

The Arel is a bit cleaner, I believe it's closer to plain Ruby (and I believe AR compiles to Arel before being sent to the db). It's basically just comparing tables as objects and accessing the columns as attributes.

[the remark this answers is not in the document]
nimmolo: Converted to AR in all three cases. Thanks for the correct syntax @JoeCohen

app/controllers/checklists_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      if (@project = find_or_goto_index(Project, proj_id))
+        @data = Checklist::ForProject.new(@project)
+      end
+    elsif list_id.present?
+      if (@species_list = find_or_goto_index(SpeciesList, list_id))
+        @data = Checklist::ForSpeciesList.new(@species_list)
+      end
+    else
+      @data = Checklist::ForSite.new
+    end
```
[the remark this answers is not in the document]
nimmolo: @JoeCohen Thanks - very nice. There's a typo in there, "cbecklist". I'll commit with that corrected.

app/controllers/observer_controller/email_controller.rb
```diff
@@ -12,7 +12,7 @@ def email_features
                                             params[:feature_email][:content])
         end
         flash_notice(:send_feature_email_success.t)
-        redirect_to(action: "users_by_name")
+        redirect_to(controller: "users", action: "by_name")
```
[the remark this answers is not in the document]
nimmolo: Committed

config/routes.rb
```diff
@@ -664,6 +653,8 @@ def route_actions_hash
   resources :articles, id: /\d+/
   redirect_legacy_actions(old_controller: "article")
 
+  get 'checklist', to: 'checklists#show'
```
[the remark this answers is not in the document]
nimmolo: Done

app/controllers/observer_controller/email_controller.rb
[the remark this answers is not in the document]
nimmolo: Done

JoeCohen:
Good idea. I'll do that after dinner. (About  an hour.)
I must have missed that discussion.
I didn't even realize it was missing/moved until this afternoon.

PR #1033, opened 2022-06-19
Update db/clean.sql with new glue table names.

pellaea:


JoeCohen:
Thanks for finding and fixing this.
Looks good.

PR #1036, opened 2022-06-20
2 `RssLogsController`

nimmolo:
Questions: 
- Currently three indexes ...?
- Why two different filtering params for index, show_type vs type
- Should "action denied" redirects specifically reroute you to rss_logs... or to site root?


pellaea:
re: why three indexes? -- The first (/observer/index) is just a routing thing.  Nowadays it would be done in config/routes.rb instead.  The remaining pair is repeated in all the controllers.  One is the main entry point for doing an index of whatever object: it creates the query.  The second is the re-entry point if you've already started a query and are resorting it or changing the filter parameters or whatever.  The query itself is passed back in via the "q" parameter.  Any index can be further modified on the fly with a bunch of other parameters -- e.g., "by" to change the order, "page" and "letter" for pagination, and I think in the unique case of rss_logs "type" or "types"... whatever parameter is used to allow users to specify which type or types they wish to see in the activity log.  Furthermore, there are additional entry points in some cases, especially for observations.  You can get to a page of observation results by several paths: all observations, observations by a given user, search results, advanced search results, observations for a species list, observations of a name, observations at a location, etc.  All of these use show_selected_observations(query) to display the actual results.  But then the sorting and pagination and prev/index/next controls all send the user back to list_observations.  It's the general purpose index that just takes whatever query is stored in "q", modifies it according to the sorting, pagination, type filter and whatever other modifier parameters are available no matter which entry point the user arrived at, and displays the result.  It probably would be okay to merge index_observations into list_observations, and just make sure that list_observations provides appropriate defaults -- just make sure "q" and "by" and all the rest are not passed in if you want it to create a new query!  I thought it was simpler and made more sense to keep the actions which *created* queries (index_observations, etc.) separate from the general purpose *re-entry* action (list_observation).  At very least I probably could have named list_observations better!  As you are the one doing all the tremendous amount of work involved in refactoring these behemoths, I say you get to choose going forward whatever makes most sense to you.  The above was my reasoning.  But my day is done.  This is a bright new day!  Enjoy!

pellaea:
re: show_type vs type -- I don't see "show_type" in that file.  My guess is that you're seeing something that has to do with the difference between the user clicking on something that says they want to see that type and nothing else, versus toggling the showing of a given type on and off.

re: action_denied → main page versus explicitly list_rss_logs -- (Hah!  I might have gotten index_xxx and list_xxx backwards! :)  You may have noticed that in the past few years I occasionally take the time to switch from the explicit list_rss_logs to simply "/" as I come across them.  I think that makes much more sense.  That lets us change what the main page is in a single place (config/routes.rb) instead of having to find and change list_rss_logs throughout the site.

Also, along the same line of reasoning you were using when you considered changing by_contribution to contributors... Maybe consider changing list_rss_logs to activity_log?  I for one struggle to remember that list_rss_logs is the Activity Log page.  It really isn't obvious from the name, is it??  The name "list_rss_logs" was just following the convention I talked about above, that's all.

JoeCohen:
I like the idea of changing `list_rss_logs` to `activity_log` for the reasons given by @pellaea,
and it also decreases the likelihood that a person or robot thinks that we have an rss feed.

pellaea:
Actually, we *do* have an RSS feed, right?  Just as an incidental note.

JoeCohen:
Doh! I just assumed we didn't because I could never get it to work with my
RSS reader.

pellaea:
Well, I didn't say that it *worked*... ;)  When's the last time anyone we
know actually successfully used it?  I've never tried it myself.

nimmolo:
> re: action_denied → main page versus explicitly list_rss_logs -- (Hah! I might have gotten index_xxx and list_xxx backwards! :) You may have noticed that in the past few years I occasionally take the time to switch from the explicit list_rss_logs to simply "/" as I come across them. I think that makes much more sense. That lets us change what the main page is in a single place (config/routes.rb) instead of having to find and change list_rss_logs throughout the site.

@pellaea - I want to ask about some specific routes, because in those cases the user might really want to be back on the activity log index. In the case of admins, i assume they will always prefer to be dropped off at the activity log:
AccountController
- turn_admin_on
- turn_admin_off
- switch_users
- add_user_to_group_admin_mode
VoteController
- refresh_vote_cache

pellaea:
As an admin I would far prefer to be dropped off at a blank page!  (I go to
the how_to_help page all the time for this very reason -- it's just about
the cheapest page we have.)  In general most admin actions return to
whatever page the admin was on before the action.  And that feels
appropriate to me, personally.

re: refresh_vote_cache should return to the observation page, I would have
thought.  Isn't that where it is accessed from anyway?  It may not be
accessible any more, assuming that anyone who wants to force a refresh need
only tweak one of their votes, admin or otherwise.

nimmolo:
@pellaea got it. These are `redirect_back_or_default` so I guess they're going back, generally. Should i put "/" as the default?

pellaea:
I would say, yes, make "/" be the default.

nimmolo:
Agreed about `activity_log`. Before this merges, i'd like to try a separate PR to rename `RssLog` `ActivityLog` everywhere, including the db columns named `rss_log_id`, but **not** in the translations. 

Does anyone perceive any gotchas with that? ("Yes" is the only valid answer!)

pellaea:
As a fellow sufferer of OCD, I feel like I have a duty to push back gently
-- how important is it really to rename "rss" to "activity" throughout the
code?  Do we actually gain anything of value that offsets the trouble and
inevitable unseen bugs it will introduce?

(I prefer "activity" to "rss", mind you, that's not it -- it's just the
general principle of avoiding fixing what isn't really broken.  I'd prefer
to see your prodigious energy better focused! :)

nimmolo:
Good call. There are only 682 uses of `rss_` in the code... what could go wrong? Anyway, maybe later.

New question: what partial is this test, `ObserverControllerTest#test_page_loads`, expecting? 
There is no partial, in the folder `views/observer` or elsewhere, called "_rss_log".
```
    get_with_dump(:list_observations)
    assert_template(:list_observations, partial: :_rss_log)
```
`shared/_matrix_box` maybe?

nimmolo:
Ok, further research on the ~rss_logs~ activity_logs tabset: 
I don't believe, in the present form of MO, that requests ever hit any of the RssLog indexes with the inclusive `show_#{type}` param. We may be able to rip out that param check.

Here's how it seems to work: The checkboxes are named `show_#{type}`, but when the sum of checkbox values is submitted (the bar is a form, after all) the submission of `types` is saved as a query, and the params are subsumed into the mysterious `q`. (Always mysterious, that one!) 

Also note: despite the efforts of the javascript `rss_feed_select_helper`, one _cannot_ in fact click on the labels to select the checkbox. Clicking on the labels effectively fires a link to the index with the _exclusive_ `type` param, so you narrow the index to rss_logs of "Projects", for example. To get an index of multiple types, you have to click directly on the checkboxes and `submit`. 

Please try it and lmk what you get. Does this seem correct to everyone?

pellaea:
> New question: what partial is this test, `ObserverControllerTest#test_page_loads`, expecting? There is no partial, in the folder `views/observer` or elsewhere, called "_rss_log".

Very good question.  _rss_log partial (and all the others -- _observation, _name, _location, etc. -- went away years ago when Ray build the matrix box presenter mechanism.  I have no idea how assert_template is working.

nimmolo:
On the `assert_template` mystery: This passes tests:
```
    get_with_dump(:list_observations)
    assert_template(:list_observations, partial: :_whatever)
```
so I believe we need to change how we call assert_template in a few places.
It seems to want the string of a directory path:
```
    assert_template("shared/_matrix_box")
```
This works, and does check that template.

Note that despite the method's current oblivion to the `partial:` named param, the following does not work:
```
    assert_template(: _whatever, partial: :_whatever)
```
That seems to mean it's only reading the first param, if they're named. We just need to use strings.

pellaea:
Thank you so much for noticing and getting to the bottom of that, Nimmo.
Yikes, I count around 64 uses of assert_template(blah, partial: blah).  :(
:( :(

nimmolo:
> I'm a little confused by the comment "requests never hit RssLog index with the show_#{type} params"...  As you go on to point out, the rss_log_tabset partial form submits a request with these parameters.  Is that not what you mean?  Yes, it immediately stores those in a Query instance.  But it definitely posts a request using those parameters.

Got it. See, i'm fuzzy even on the basics - really appreciate these clarifications. I feel like I may be able to get the checkboxes to modify the same param `type`, with joined values "project,observation,blah" - at least i'm going to try it. If not, well, it's back to working as expected now.

nimmolo:
Yikes. This is thorny, in controller_extensions:
```
  def assert_action(action, partials)
    if partials
      if partials.is_a?(Array)
        assert_action_partials(action, partials)
      else
        assert_redirected_to(action: action, partial: partials)
      end
    else
      assert_redirected_to(action: action)
    end
  end

  def assert_action_partials(action, partials)
    partials.each do |p|
      assert_template(action, partial: p)
    end
  end
```
Can i do (in the second method):
```
      assert_template("#{action}/#{p}")
```

pellaea:
Consider removing both assert_action and assert_response (just above it in
controller_extensions).  I think we may never use the customized part of
assert_response.  And I think we use assert_action only once.  I may be
missing something, though.  That was just a quick survey by grepping
"assert_response" and "assert_action".

I vaguely remember writing assert_response but apparently never actually
made any use of its powerful features.  And probably just as well!  I don't
have any memory of assert_action.

nimmolo:
I restored (and corrected) the partial checks all through the controller test.
However, I suspect anything calling `assert_action_partials` is blissfully ignoring those partials.
For removal of `assert_action` I'd ask for your assistance. (ditto for `assert_response`... it's used a lot).

I found that `assert_action` and `assert_action_partials` are not used many times, but they're used to build other methods in the controllers Location, Naming, Observer and SpeciesList that are used more than a few times.

pellaea:
re: assert_action and assert_action_partials are used to build other helper
methods -- Damn, I was wondering if that was the case.

re: assert_response is used a lot -- Yes, but only as
assert_response(:success), etc.  Note that those fall straight through to
super.  So those don't count.  I think there wasn't a single use that
actually uses non-standard capabilities, or maybe one.  No worries.  If
it's not actually broken, it can stay.  It's just a *lot* of complicated
code... and if it's not used...  well, you understand! :)  I thought
assert_action and assert_action_partials might be subroutines of
assert_response because there's no comment explaining what assert_action
does.  (Another clue that I am not the culprit for assert_action! :)  But I
guess not.  Phooey.

nimmolo:
Ok. well... what should we do about it? 
From all this, i believe it's safe to assume that `assert_action_partials` is not checking anything.

nimmolo:
Done. Wasn't actually that much work.

pellaea:
Good, that's what I would have done, too.  Thanks!

nimmolo:
Only one problem. Test show observation checks for a species list, but the template doesn't show one if there isn't one. The first assertion fails because the fixture `unknown_with_no_naming` has no list, i believe. Should I remove that partial from what it's testing, or use a different set of assertions for that first fixture?

nimmolo:
That's `ObserverControllerTest#test_show_observation` line 825

pellaea:
Really?  Only this one problem??  I'm surprised there aren't more!  I don't
know how I'd fix this.  It sounds like you're going to see a different set
of partials for each observation.

nimmolo:
Yeah, only this one. I removed `assert_template("observer/_show_lists")` from the to-do list of the method `assert_show_observation` and everything passes now. Better than not checking any partials, I guess!

I forgot, I'd FIXME'd one other unresolved issue. It's trying to test passing an ID to a partial, but i don't know if that can happen with the current `assert_template` syntax.

In RssLogsControllerTest,
```ruby
  def test_altering_types_shown_by_rss_log_index
    login
    # Show none.
    post(:index)
    assert_template(:index)

    # Show one.
    post(:index,
         params: { show_observations: observations(:minimal_unknown_obs).to_s })
    assert_template(:index)

    # Show all.
    params = {}
    RssLog.all_types.each { |type| params["show_#{type}"] = "1" }
    post(:index, params: params)
    # FIXME: assert_template now ignores `partial` param (always).
    # What's this intended to check?
    # assert_template(:index, partial: rss_logs(:observation_rss_log).id)
    assert_template(:index)
  end
```

pellaea:
Haha, I have no idea what that FIXME thing was supposed to do!

Removing _show_lists completely works for me.  Good work!

nimmolo:
Thanks. Just modified the checkboxes to send an array of types on submit, under the same `type` param as the single-type indexes.

pellaea:
Sweet.  I had no idea that was possible.  Can't wait to see how you managed
that.

nimmolo:
The only difference is in the HTML form with checkboxes, you can `name` the checkboxes with the name plus square brackets, implying it should send an array of checked values as the named param.

`<input type="checkbox" name="type[]" value="#{type}">`

nimmolo:
These controller refactors seem like they could be a bit much to add to the Zeitwerk/enums `next` branch. Should I open a new `edge` branch or something, for these controller refactors that each of them can be merged into, that builds on `next`? That way we can let the dust settle on Zeitwerk before we merge the controller/view changes.

pellaea:
Brilliant!

pellaea:
I don't suppose that would be a problem.  I'm also fine with pushing it all
out at once soon as API goes away at the end of the month.  I don't mind.

PR #1037, opened 2022-06-21
Prevent commas at start and end of some Name fields

JoeCohen:
This PR prevents (going forward) Names from having: (a) Authors ending in a comma, and (b) Citations starting with a comma. These are mistakes that we sometimes see and can be easily prevented.
The PR operates by stripping combinations of space(s), comma(s), and som other punctuation from the start and end of the entered Author and Citation fields before saving them.

Delivers https://www.pivotaltracker.com/story/show/182508428

####Manual Test
Try to create a Name with an Author ending in a comma, and a citation beginning with a comma. E.g.:
- Scientific Name:`Boletus test`
- Author: `TryToEndInComma,` 
- Citation: `, TryToStartWithComma`
Expected result: The Name is created without those commas.

### DIFF
--- a/app/controllers/name_controller/create_and_edit_name.rb
+++ b/app/controllers/name_controller/create_and_edit_name.rb
@@ -95,7 +95,8 @@ class NameController
   def set_unparsed_attrs
     set_locked_if_admin
     set_icn_id_if_unlocked_or_admin
-    @name.citation = params[:name][:citation].to_s.strip_squeeze
+    @name.citation = params[:name][:citation].to_s.
+                     strip_squeeze.strip_bad_chars_at_ends
     @name.notes    = params[:name][:notes].to_s.strip
   end
 
--- a/app/extensions/string_extensions.rb
+++ b/app/extensions/string_extensions.rb
@@ -520,8 +520,8 @@ class String
 
   # Strip leading and trailing whitespace and some punctuation chars
   def strip_bad_chars_at_ends
-    gsub(/\A[\s\,;-]+/, "").
-      gsub(/[\s\,;-]+\Z/, "")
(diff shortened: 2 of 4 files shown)


pellaea (approved):
I love it!  Only question is... which bad characters exactly does it mean?  Is there any easy way to characterize better this selection of commas, semicolons, dashes?  Maybe "strip_punctuation_from_ends"??  Not important, just a thought.

JoeCohen:
Good point. I will rename the method, plus try to generalize it a bit and test it directly.

pellaea:
FWIW, I think you can get away with what you have -- commas, semicolons,
dashes -- It works for the present purpose.  If we need to expand it later
for other purposes, we'll have better use-cases presumably at that time,
and will know better how to generalize it.  We don't, for example, want to
include dots, I assume, even though that's arguably a punctuation.  It's
tricky, for sure.  If it really does need to be highly specialized for
copying and pasting out of Index Fungorum (!!) then it might be worth
considering naming the method to make that clear.  And such a method might
not belong in the String extensions.  Note that if the method were right
there next to the single caller, then the name of the method is very
unimportant.  The name is only important if you need it to describe what
the method does.  If the method is right there, you can easily just glance
down to see what it does, so leaving it vaguely named is just fine.  Blah
blah blah.  Just random unordered thoughts on the subject.

JoeCohen:
You're right. I'm generalizing prematurely.
Furthermore, you catalyzed me to think that telling the user about the error is a better approach than automatic correction. 
So I'm closing this PR, and will redo it as validator(s) on Name.

pellaea:
Interesting, that wasn't my intention at all, but if this new approach
works better for you, then I support it!

PR #1041, opened 2022-06-21
Refactor ObserverController

nimmolo:
This combines all the branches that are splitting out the ObserverController components.

Please make new improvements or alterations as commits to this branch.


README_CODE
```diff
@@ -176,7 +176,7 @@ Note that the default controller is "observer" and the "id" parameter is
 shortened to a bare integer:
 
   http://mo.org/42                      Show observation #42.
-  http://mo.org/observer/show_user/252  Show summary for user #3.
```
pellaea, line 179: Would you mind correcting it from 252 to 3 (or vice versa)?  Trivial, but unnecessarily confusing.

app/controllers/lookups_controller.rb
```diff
@@ -1,7 +1,14 @@
...
-module ObserverController::MarkupController
+class LookupsController < ApplicationController
+  # These need to be moved into the files where they are actually used.
+  require "find"
+  require "set"
+
+  before_action :login_required, except: [
+    :lookup_observation,
+  ]
+
```
pellaea, line 11: Very minor point, but I bet you'll agree -- the "lookup" in the action name is now redundant, right?

app/controllers/vote_controller.rb
```diff
@@ -57,8 +57,8 @@ def refresh_vote_cache
     # Naming.refresh_vote_cache
     Observation.refresh_vote_cache
     flash_notice(:refresh_vote_cache.t)
-    redirect_with_query(controller: :observer,
-                        action: :list_rss_logs,
+    redirect_with_query(controller: :rss_logs,
+                        action: :index,
                         id: observation.id)
   end
```
pellaea, line 63: Oh my stars.  Have you looked at what this action actually does?  Or tries to do -- it would probably fail to do it in the age of the universe!  Fortunately, there is no link to this action.  Would you mind terribly ripping it out?  If we ever decide that we absolutely must recalculate the consensus for every single observation on the site, we need to rewrite Observation.refresh_vote_cache to take advantage of eager loading at very least.  I would recommend a mass-update, as well.  I think even if the calculation were trivial, simply running updates for every single observation in the database one at a time would be prohibitively slow.  I think if this has a prayer it will have to be rewritten to calculate a massive hash of all of the new consensus names and vote caches, and then change them all in one or a few gigantic mass updates.  (Then update the cached classification and text_name columns in observations table, too, but that part's already written.)

pellaea: By the way, the variable "observation" isn't even initialized the way this action is currently written, right?

app/helpers/footer_helper.rb
```diff
@@ -28,13 +28,13 @@ def show_authors_and_editors(obj)
...
         authors += link_with_query("(#{:review_authors_review_authors.t})",
-                                   controller: :observer,
+                                   controller: :authors,
                                    action: :review_authors,
                                    id: obj.id, type: type)
       elsif !is_author
         authors += safe_nbsp
         authors += link_with_query("(#{:show_name_author_request.t})",
-                                   controller: :observer,
+                                   controller: :authors,
```
pellaea, line 37: Is it worth removing the redundant "review" from these two actions?

Long, long ago Nathan and I decided the retain the redundancy -- thus name/show_name, image/show_image, etc.  But that clearly is not the CRUD Way™.  So since we're refactoring anyway, we might as well consistently remove the redundancy throughout, right?  Or does this add too much work?  (Such as changing rss_log to activity_log would have done.)

nimmolo: That's right, let's keep it on-brand! The CRUD Way™ hahaha

But i guess you meant the redundant "author"... oder?

app/controllers/vote_controller.rb
nimmolo: Let's rip it out. 💨

app/controllers/lookups_controller.rb
nimmolo: Agree. Seemed a bit... bare... without it, but i'll just go for it. These actions are namespaced under the class, so no one's going to get confused. Hopefully.

script/refresh_sitemap
```diff
@@ -75,17 +75,17 @@ OBJECT_TYPES = [
...
+  ["how to help",  "info/how_to_help",    "weekly"]
 ].freeze
 
 # These are aliases for pages allowed above.
 OTHER_ALLOWED_PATHS = [
   # "/observer/index", (would allow too many indexes we don't want to allow)
-  "/observer/lookup_observation",
-  "/observer/lookup_name",
+  "/lookups/lookup_observation",
+  "/lookups/lookup_name",
```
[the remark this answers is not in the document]
nimmolo: I'd guess that a sitemap should be an index of active routes, and the 301's should not be in it. Worth researching.

app/presenters/matrix_box_presenter.rb
```diff
@@ -111,7 +111,8 @@ def user_to_presenter(user, view)
                    #{:list_users_contribution.t}: #{user.contribution}<br/>
                    #{:Observations.t}: #{user.observations.count}".html_safe
     # rubocop:enable Rails/OutputSafety
-    self.what  = view.link_with_query(name, action: :show_user, id: user.id)
+    self.what  = view.link_with_query(name, controller: user.show_controller,
+                                      action: user.show_action, id: user.id)
```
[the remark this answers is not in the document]
nimmolo: Found it. I added a concern called "routeable" suggested by someone encountering similar difficulty.

But hey - is there some way to inline this presenter code right in the matrix box template? 

I guess it's the kind of logic that would be provided by a controller, but this is inter-controller so it has to be independent of any controller. It's called by several controllers.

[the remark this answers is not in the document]
pellaea: Sorry, I don't understand presenters or the theory behind them.  I do know that overuse of partials is a huge suck on performance, though.  So in-lining appeals to me, personally.  But I think in-lining is precisely what the presenter is supposed to be fixing.  Not much help, right?

script/refresh_sitemap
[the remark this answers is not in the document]
nimmolo: Hm. well, discoverlife.org is offline. Who else

I have to say i disagree about permanent URLs — this amounts to an obligation to other websites never to refactor our site, which means ever-increasing technical debt and garble for us. 301's should be sufficient — they get legacy links where they need to go, and in my opinion, even those should have an expiration date. We have to adjust for their changing organizations, why not the other way around?

Certainly URLs are not considered permanent in other organizations. If a site gives a 301, it's considered a clients' responsibility to update their links, in my experience.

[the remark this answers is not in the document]
nimmolo: For sitemaps, this is the kind of info i've been working with:

> These (XML) sitemaps give search engines a website’s URLs and offer data a complete map of all pages on a site. This helps search engines prioritize pages that they will crawl.
> 
> There is information within the sitemap that shows page change frequency on one URL versus others on that website, but it is unlikely that this has any effect on rankings.
> 
> An XML sitemap is very useful for large websites that might otherwise take a long time for a spider to crawl through the site and can help search engines just simply know what pages exist. Without this sitemap, they would not be able to build the full menu of available pages until they completed a crawl of the whole site.

I don't think update frequency has much effect on bot behavior; they're trying to crawl the whole thing. On the other hand, seems like we could keep things off the sitemap that we don't want crawled.

app/helpers/footer_helper.rb
nimmolo: Update: `request` seems to be a reserved word. So i made it `authors/email_request`

config/routes.rb
```diff
@@ -747,10 +826,26 @@ def route_actions_hash
...
+    lookup_species_list
+    lookup_user
+  ]
   # Accept non-numeric ids for the /observer/lookup_xxx/id actions.
-  lookup_xxx_id_actions.each do |action|
-    get("observer/#{action}/:id",
-        controller: "observer", action: action, id: /.*/)
+  LOOKUP_ACTIONS.each do |action|
+    match("lookups/#{action}(/:id)", to: "lookups##{action}", id: /\S.*/,
+          via: [:get, :post])
```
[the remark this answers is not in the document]
nimmolo: Post removed

app/controllers/vote_controller.rb
nimmolo: Commented out, with additional comment about expensive query.

app/controllers/lookups_controller.rb
nimmolo: Not feeling like changing these... too confusing without the prefix. 
But did you want to remove `lookup_observation`?

README_CODE
nimmolo: Updated

script/refresh_sitemap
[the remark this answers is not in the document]
pellaea: I couldn't disagree more strongly with you, Nimmo!  One, it doesn't prevent us from refactoring in the slightest.  Two, sites that change their canonical URLs are doing a huge disservice to the larger web community.  All it means is separating the public view from the internals of the site.  And that's just plain good programming practice.  It's convenient at first for the URLs to match internal labels, sure.  But to *rely* on that, as you say, forces us to either never refactor or to change the public interface willy nilly with every whim of the internal coding structure.

Perhaps we're misunderstanding each other.  I am only advocating for a public route /observer/lookup_name to stick around.  You can route it to whatever internal action you wish.  If you don't like the parameter format, we can make a wrapper that converts it.  You are free to refactor the controller and name it whatever you wish.  I don't see how keeping /observer/lookup_name puts any constraints on your refactor whatsoever!

pellaea:
re: sitemap -- Okay, whatever, then.  Do whatever you like with it.  If crawlers are going to disregard it anyway, I guess I don't care so much!  BTW, I wasn't saying anything about frequency, I was telling crawlers when pages were last changed.  Maybe it's just me, but if I'm a crawler prioritizing, why would I crawl a page which the site tells me hasn't changed since I last crawled it???  And we already tell it very exclusively which pages we want to be indexed.  I'm not arguing about that.  I want to discourage crawlers from getting the same unchanged page over and over and over again.  A real frustration when that wasted traffic is like 50% of the load on your server!  Of course, it could be taken as an opportunity to cache better.  If the page hasn't changed, then why can't we serve a cached static version to the crawler?  I've been advocating for that for years, longer than you've been with us, Nimmo!  But it's a major project, and no one has had the time to do it.

re: lookup_<type> actions staying the same -- Fair enough.  It's actually sort of the opposite case from changing name/show_name → names/show -- There we are removing the redundant "name".  Now we are removing the equivalent of "show"!  It's "lookup" not "name" that is redundant.  So again, fair enough, not the same case, so not surprising our preferences would differ.  And definitely not required for consistency.

script/refresh_sitemap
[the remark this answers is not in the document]
nimmolo: OK - Currently the `observer/lookup_name` routes are still there, but they are 301'd. 
You're saying: keep them there and **active**, not 301?

Either way you're right, this is not a big deal to maintain.

For background, my sense — coming from the bot/API/EOL/discoverlife.org situations — is that our code maintains a lot of things out of consideration for phantom "participants" who may be... hoodlum bots not even interested in mycology, or the original participants may not even be there at all, or that our code intended to provide for them may be broken and we're not aware. Maybe we can start implementing stronger customs of maintaining contact info for all end users...? There can't be _that_ many academic, institutional or even solo end users of MO... and my suspicion is that they're all capable of writing and receiving an email. It's not like we're providing translations of banned Chinese political texts, and the users need the anonymity... or do they?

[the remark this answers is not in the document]
nimmolo: and by solo end users i mean, end users of custom-tailored routes and functions, not of the site generally.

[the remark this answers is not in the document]
nimmolo: I guess what i'm saying is...\ i have a hard time mustering mental space for anyone or any institution who expects to mooch data off MO without so much as a courtesy email introducing themselves, and providing contact information. Heck, I introduced myself to you 11 years ago and I didn't even want anything! I'm persuadable though, if you can think of a situation where this kind of lurker mooching is valuable and unavoidable.

nimmolo:
New topic: I'm thinking of how to deal with certain actions in the `ObserverController/Other` that don't have an obvious home.

`turn_javascript_on` etc -> move to `AccountsController` (seems to make sense, even though these browser settings aren't saved in the account)

(currently) `info/w3c_tests`, `info/change_banner`, `observer/test_flash_redirection` - move to a new **AdminController**

`observer/set_export_status` -> move to `NameController` ?

script/refresh_sitemap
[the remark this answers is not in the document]
pellaea: Good conversation!  And I think your attitude is probably the healthier one in general.  I do think of it a little differently, though.  Not in terms of what people expect of us, but of what I've (we've) promised other people.  There aren't many things, so I don't think this is as big as you are worried.

1. image URLs -- https://mushroomobserver.org/images/$size/$id.jpg -- I've been trying to get people to use this URL for ages.  From the beginning, really, but that's literally where they were at the beginning, back when we actually stored the images locally on the webserver(!!)

2. observations at https://mushroomobserver.org/$id

3. content we've "published" -- By this, I mean basic observation data, including notes and comments.  This includes markup, in which you can find these lookup_xxx URLs.  We decided early on not to strip HTML / markup out of the notes.  That means these have been available for many years.  Many, many people have contacted us asking about downloading data from MO.  Foolishly perhaps, I have never bothered to keep track of them.  My position was, okay, fine, the easiest way to manage access is simply to give it away without any restrictions.  That meant structuring the markup so that it would be guaranteed to be stable.  So I created the _blah_ mechanism for creating internal links, made them point to this set of silly lookup_blah actions with the express (to myself at least) expectation that they be guaranteed to remain stable.  But nothing else.  We do allow users to enter their own links and HTML and stuff, but that's all implicitly at their own risk.  I make no guarantees, obviously, for that sort of thing.  I designed the lookup mechanism the way I did so that it did not rely on name ids, because the meaning of a name id could potentially change.  Instead I made it go by the scientific name.  I figured that was as stable and simple and transparent as possible.

That's about it.  Maybe there's something else somewhere, but I can't think of anything else right now.  I have never encouraged third parties to link directly to name pages, or project pages or anything else.  Just these three things.  The API is there for anything else.  And of course, that's a whole other can of worms.

So if we can just ensure these three canonical URLS -- images, observations, lookups -- I will be happy.  Redirects are okay, I guess, if we must, but rails makes it so easy to route them directly, so why not?  /1234 doesn't redirect to /observations/1234, it routes straight there.  I wish I could do that with the images, too, but I've never figured out how.  The problem with the redirect is people get the 301 or whatever the code is, and think "darn MO, they've changed it, I thought that was supposed to be stable, oh well, I guess I'll change it".  But that's wrong!  I do *not* guarantee the new route, only the old route.  So the redirect fools people into switching to an unstable route.  Ugh.  That's caused so many people problems with the images.  Including people whom I've told very explicitly in no uncertain terms not to do precisely that!  Arg.  So, yes, I would prefer rerouting, not redirecting, for that reason! :)

pellaea:
re: admin controller -- Funny, I was thinking exactly the same thing!

re: js stuff → account controller -- Sure, why not.  Good as anywhere!

pellaea:
re: admin controller -- Funny, I was thinking exactly the same thing!

re: js stuff → account controller -- Sure, why not.  Good as anywhere!

script/refresh_sitemap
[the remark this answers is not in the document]
nimmolo: Got it. Lookup routes shall endure!

For observations and images, I'll explore if there's a way to redirect and send a different HTTP code. `300` means "multiple choice", i.e. we're rerouting you, but both urls are valid. Maybe that's the one we should try — but TBD if Rails allows customizing the response code. (Happy to leave that alone, also... Is this an interesting thing to try?)

nimmolo:
I decided on a `JavascriptController`. What the hell... at least it is what it says it is!

nimmolo:
Hm. I broke `test_change_banner`.
Could you take a quick look at that test? The first assertion expects a non-logged-in user to get bounced to the login path:
```ruby
      get(:change_banner)
      assert_redirected_to(controller: :account, action: :login)
```
But the action actually seems to bounce them to the rsslogs index:
```ruby
  # Update banner across all translations.
  def change_banner
    if !in_admin_mode?
      flash_error(:permission_denied.t)
      redirect_to("/")
    ...
```

script/refresh_sitemap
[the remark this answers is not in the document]
nimmolo: Regarding redirecting images and observations, yes, it is possible. Idk if this is what you intend, but it does redirect and the status says "both routes valid".
```ruby
get ':id', to: redirect('/observations/%{id'}', status: 300)
```

nimmolo:
Nevermind! `before_action :login_required`

script/refresh_sitemap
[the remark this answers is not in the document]
pellaea: Interesting.  I think I'd like /$id to reroute not redirect, anyway... yes, the way it works right now.  And the image URLs are handled by nginx so they never make it to the rails router.  But maybe there's a way to tell nginx to use 300 instead of 301 when "rewriting" a request.  Thanks!

pellaea:
> I decided on a `JavascriptController`. What the hell... at least it is what it says it is!

There's really not much of a penalty against lots of controllers, is there?

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,205 @@
...
+    end
+  end
+
+  alias index_user index
+  # People guess this page name frequently for whatever reason, and
+  # since there is a view with this name, it crashes each time.
+  alias list_users index
+
+  # User index, restricted to admins.
+  def by_name
```
JoeCohen, line 31: Would this be better as a flavor or type of `index`, getting rid of one path? 
(Sorry for the belated comment. Maybe the above should be a separate PR, so as not to delay this PR.)

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,205 @@
...
+      query = create_query(:User, :all, by: :name)
+      show_selected_users(query)
+    else
+      flash_error(:permission_denied.t)
+      redirect_to("/")
+    end
+  end
+
+  # by_contribution.rhtml
+  def by_contribution
```
JoeCohen, line 42: See above comment about `by_name`.
Also (now paying attention to what this does) I agree with @nimmolo that -- provided that it's a separate route -- `contributors` is better than `by_contribution`.

app/controllers/users_controller.rb
pellaea: Yes, it is possible that GET /users?by=name is equivalent.  Likewise, below, GET /users?by=contribution could replace /users/contribution or /users/contributors.

nimmolo: Car's in the shop! It was a failure of the cooling system, recently worked on.

My arbitrary philosophy with "flavors of index" so far has been, if it's different arrangements of data using the same layout, make the flavor a param of the index path. But if it's a completely different layout, using different partials for example, use a different path. 

Going by that, i feel like contributors, which seems like a different presentation of user data from the index, should be a separate path. Actually, the most hassle-free way to refactor it would be to make it a separate "Contributors" controller with one route: index. This would be my preferred 'path'.  i.e., mo.org/contributors

What do you guys think?

nimmolo: Another way of saying that tentative "flavor-of-index-philosophy" is 

"if you can (or would) switch between these views of the data on the same view", like filtering the index.

pellaea:
That makes sense.  Also, isn't contributors publicly accessible?  But the
general-purpose user index is not?  Moving it to a new controller feels
like overkill to me, but if it sits better with you, Nimmo, I can
definitely live with it.

So sorry about the bust of your plans.

nimmolo:
Ok, commits coming. 
By the way, what's users_by_name? 
It's an admin-only view, but the user index can be sorted by name — is that the same thing? I assume not...

pellaea:
Honestly, I never use any of these user indexes, so I'm the wrong person to
ask!  I do all user administration on the back end by logging into the
server an running mysql by hand.  @jdcohenesq & @nathan -- Do you use any
of these admin-only user indexes?

nimmolo:
New `ContributorsController` is in. I deleted `by_name` since the "by" param is already coded into the `:index` action. All links to `users_by_name_path` are now rewritten as `users_path(by: "name")`

JoeCohen:
@nimmolo: I also like moving it out of UserController. An alternative way to do that would be to namespace it.
@pellaea: I have occasionally looked at the back end "Users By ..." pages. But I don't find them that helpful. I wouldn't mind if the methods and routes disappeared. (I'm pretty much SQL illiterate, but get by with Rails AR in the back end Rails console.)

nimmolo:
@pellaea somehow controller proliferation doesn't feel like overkill when writing it, it feels like less code...?
- less logic in each controller, less keeping track of routes and names. You just have to think of show, index (with whatever flavors) and CRUD; if it gets trickier than that, it probably could be another simple controller
- simpler routes
- easier to find what does what in the directories: controllers, views, tests

nimmolo:
@JoeCohen It could be namespaced, I agree. I wouldn't mind it that way at all.
"Contributors" just seems like a top level path to me, since it's in the main nav... and people are vain!

pellaea:
Yeah, the proliferation of tiny files is my problem.  But it's not a big
deal.

JoeCohen:
Grr. Looks like I may have caused conflicts instead of fixing them:
>This branch has conflicts that must be resolved
 to resolve conflicts before continuing.
Conflicting files
app/controllers/observer_controller/other.rb
app/controllers/observer_controller/search_controller.rb

I'll back out the last two commits. Maybe someone smarter than me can deal with this.

JoeCohen:
That didn't help. Looks like the conflicts were there before I "Re-fix[ed]" things. 
I'll still let someone smarter deal with this (and hopefully include the RCE fixes).

nimmolo:
@JoeCohen Handling it now. When conflicts are resolved I can merge this into `next` to simplify things!

JoeCohen:
Much thanks. Too many branches for me to figure out the right one to use as the base branch.
But that should improve by this weekend, as I think @pellaea will get rid of the old API on July 1.

nimmolo:
Considerations about merging this into `next`? 

I had earlier asked @pellaea this question and he felt it was ok to merge, but i then had second thoughts out of a concern for possible bug triage. But I feel like at this point it's pretty solid.

nimmolo:
My idea, if you guys agree, is to keep the `next` branch active for major PRs. 
I figure we're going to probably wait a day or two after deleting API and moving to zeitwerk, so i'll merge this branch into `next` now, and it'll be the new "next".

JoeCohen:
Maybe we should talk about this in today's tech meeting.
I can see the advantage of doing this.
But -- maybe because of my inexperience, poorer coding abilities, and/or
force of habit -- it makes me uncomfortable to always develop against next
rather than master; there's a logjam of changes; makes it harder to isolate
issues that appear when everything is deployed at once.
It also looked to me like Codeclimate doesn't report on pushes to PRs that
are not based on master.

pellaea:
I'm still good with merging everything into "next" and doing a massive push
to production right after API goes away.  I prefer one big push to
trickling distractions every day.

PR #1046, opened 2022-06-22
Prevent bad punctuation in Author and Citation

JoeCohen:
Prevents most punctuation at the end of Author and the start of Citation.
There are Names whose Authors ends in a comma or whose Citation starts with a comma.[1]
This PR gives an appropriate flash warning when a user tries to Save/Update a Name with one of those problems.

Follows-up on now-closed PR #1037 
Re-delivers https://www.pivotaltracker.com/story/show/182508428

#### Manual Test
Try to create a Name with an Author ending in a comma, and a citation beginning with a comma. E.g.:

Scientific Name:Boletus test
Author: TryToEndInComma,
Citation: , TryToStartWithComma
Expected result: User is returned to form with a flash message describing the mistake.

-----
1. I think this typically results from copying and pasting too many characters from Index Fungorum.

### DIFF
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -462,7 +462,7 @@ class Name < AbstractModel
   end
 
   def citation_start
-    # Should not end with punctuation other than:
+    # Should not start with punctuation other than:
     # quotes, period, close paren, close bracket
     # question mark (used for Textile italics)
     # underscore (previously used for Textile italics)
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -377,6 +377,7 @@ class Name < AbstractModel
   validate :icn_id_unique
 
   validate :author_ending
+  validate :citation_start
 
   # Notify webmaster that a new name was created.
   after_create do |name|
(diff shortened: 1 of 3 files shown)


JoeCohen:
My thought was that instead of silently correcting these mistakes, we should tell Users so that they know what's happening and can avoid the mistakes in the future.

app/models/name.rb
```diff
@@ -447,4 +450,27 @@ def icn_id_unique
...
+    # other than quotes, period, close paren, close bracket
+    return unless (
+      punct = %r{[\s!#%&(*+,\-/:;<=>?@\[^_{|}~]+\Z}.match(author)
+    )
+
+    errors.add(:base, :name_error_field_end.t(field: :AUTHOR.t, end: punct))
+  end
+
+  def citation_start
+    # Should not end with punctuation other than:
```
pellaea, line 465: very minor: I think you mean "start" not "end"

pellaea (approved):
Looks good.

PR #1047, opened 2022-06-22
Observer/Indexes - fix one Brakeman SQL alert

nimmolo:
The alert may be bogus, but this silences it. 
Is this use of `pluck` in Observer/Indexes optimal, though? If not, please advise.

~Note: this bit of NameController SQL is also converted in the upcoming Refactor ObserverController PR, but this PR backports that change because it seems independent of that PR and will keep our spiffy new Brakeman a bit quieter.~ 
Nope: the change in NameController is too complicated because of Zeitwerk changes. It can wait.

This PR does not conflict with the refactor, though, and can merge whenever.

### DIFF
--- a/app/controllers/observer_controller/indexes.rb
+++ b/app/controllers/observer_controller/indexes.rb
@@ -218,11 +218,11 @@ class ObserverController
 
     unless locations.empty?
       # Eager-load corresponding locations.
-      @locations = Location.connection.select_rows(%(
-        SELECT id, name, north, south, east, west FROM locations
-        WHERE id IN (#{locations.keys.sort.map(&:to_s).join(",")})
-      )).map do |id, *the_rest|
-        locations[id.to_i] = MinimalMapLocation.new(id, *the_rest)
+      @locations = Location.
+                     where(id: locations.keys.sort).
+                     pluck(:id, :name, :north, :south, :east, :west).
+                     map do |id, *the_rest|
+        locations[id] = MinimalMapLocation.new(id, *the_rest)
       end
       @observations.each do |obs|
         obs.location = locations[obs.location_id] if obs.location_id
--- a/app/controllers/observer_controller/indexes.rb
+++ b/app/controllers/observer_controller/indexes.rb
@@ -219,9 +219,9 @@ class ObserverController
     unless locations.empty?
       # Eager-load corresponding locations.
       @locations = Location.
-                     where(id: locations.keys.sort).
-                     pluck(:id, :name, :north, :south, :east, :west).
-                     map do |id, *the_rest|
+                   where(id: locations.keys.sort).
+                   pluck(:id, :name, :north, :south, :east, :west).
+                   map do |id, *the_rest|
         locations[id] = MinimalMapLocation.new(id, *the_rest)
       end
       @observations.each do |obs|
--- a/app/controllers/name_controller.rb
+++ b/app/controllers/name_controller.rb
@@ -162,28 +162,10 @@ class NameController < ApplicationController
   end
 
   # Display list of the most popular 100 names that don't have descriptions.
+  # NOTE!! -- all this extra info and help will be lost if user re-sorts.
   def needed_descriptions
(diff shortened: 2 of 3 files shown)


app/controllers/observer_controller/indexes.rb
```diff
@@ -218,11 +218,11 @@ def map_observations
...
-        SELECT id, name, north, south, east, west FROM locations
-        WHERE id IN (#{locations.keys.sort.map(&:to_s).join(",")})
-      )).map do |id, *the_rest|
-        locations[id.to_i] = MinimalMapLocation.new(id, *the_rest)
+      @locations = Location.
+                   where(id: locations.keys.sort).
+                   pluck(:id, :name, :north, :south, :east, :west).
+                   map do |id, *the_rest|
+        locations[id] = MinimalMapLocation.new(id, *the_rest)
       end
```
pellaea, line 226: Naturally I don't have the capacity to benchmark this, but from what Nimmo and I were seeing not too long ago in a not-altogether-different case, it is actually a good deal faster these days to instantiate all of the locations, and then grab the columns you need from the instances.  Admittedly, though, this could be *huge*.  I mean, for example, it is entirely possible that folks will try to map every single location in our database(!)  So it might be a good idea to benchmark this one properly.  And pay attention to memory use, as well as system/user time, this time around.

For clarity, I'm recommending comparing what you've written with:

`Location.where(id: locations.keys).map do |loc|
    locations[id] = MinimalMapLocation.new(id, loc.name, loc.north, loc.south, loc.east, loc.west)
end`

(Note that there is no reason to sort locations.keys... I don't think.  I guess it depends what mysql does on the database end.  Maybe if dbserver time is more precious than webserver time it really is better to sort it however mysql wants it sorted on the webserver end.  But how would we know what will make mysql's life easiest?  In the absence of any information, it stands to reason we should do nothing -- that is, leave locations.keys unsorted until we have some benchmark or documentation proving that it is better to sort it somehow.)

pellaea (approved):
I say merge this PR, and to the extent we want to do anything with my obnoxious suggestion, toss it in a pivotal story for another day!

app/controllers/observer_controller/indexes.rb
nimmolo: How about limiting the number of locations this method will map to a sane value?

nimmolo: (I'm waiting on a tow truck, wont make it to retreat tonight!)

pellaea:
Oh no!  What a drag!  (Pun not intended.)

I hate to seem like I'm always disagreeing with you, Nimmo, but...  I
actually want to support folks doing that.  Mapping all of our locations,
or all of our observations.  I think that would be a really neat map!  And
I tried to design the mapping code to be able to handle it.  It actually
simplifies the map by collapsing nearby points together, and keeps bumping
up the coarseness a factor two at a time until the number of objects it
draws on the map falls below a sane value.  It even -- get this! -- lets
you click on the aggregate objects to see what observations/locations were
collapsed into it.

Now, if it blows up the server and there is no way to make it work, fine.
But the way it was written, it did actually work.  Or did many years ago
when I wrote it!  I should probably test it... one minute... it took ca. 30
seconds to map all of the observations.  Yikes the map is ugly, though.  Oh
well.  iNat's "heat map" really would be better for large maps.  Add that
to the list of things to do...

Meanwhile... I say let it go.  If it blows up, it blows up.  This is not a
frequent request.  Not like image searches and show_user and site_stats,
which are also really expensive pages, and they show up much more often.
(Though thankfully not even remotely as frequent as list_rss_logs,
list_observations and show_observation.)

My two cents.

JoeCohen (commented):
I like the new code, but don't understand the performance issues. 
I also don't understand what @pellaea means by "let it go": 
Leave the code as is vs merge it.
I'm fine either way.

pellaea:
Sorry, @JoeCohen, I had to think hard to remember what I was responding to!  I was responding to @nimmolo suggesting limiting the number of locations.  I was suggesting "letting it go", meaning letting if match as many locations as it wanted to.

PR #1051, opened 2022-06-23
Fix ImageController SQL Injection warning.

JoeCohen:
This PR fixes a Brakeman SQL Injection warning (See below.)
The PR:
- Uses AR insert_all in place of raw SQL
- Removes now-unneeded lines designed to fix SQL injection
- Bonus: Fixes RuboCop method-length offense

NOTE: The warning may have been a false positive. But in any event, it's now gone. And IMO the code is much simpler.

-----
Screen Shot 2022-06-23 at 2 19 27 PM


app/controllers/image_controller.rb
```diff
@@ -780,27 +780,21 @@ def row_changed?(row)
...
   end
 
-  # Add license change records with raw SQL in order to use a single INSERT.
+  # Add license change records with a single insert to the db.
   # Otherwise updating would take too long for many (e.g. thousands) of images
   def update_licenses_history(images_to_update, old_holder, old_license_id)
     data = images_to_update.pluck(:id, :when)
+    values = data.each_with_object([]) do |datum, vals|
+      vals << { user_id: @user.id.to_s,
+                updated_at: Time.current,
```
JoeCohen, line 789: Is there a subtle difference here?
I think that `updated_at` be set to the time when Rails runs this method, rather than the time that the db inserts the record.
The difference is so small that I don't think it matters.
Also note that in Rails 7, we can get rid of this line.

JoeCohen:
@pellaea & @nimmolo 
I'm now wondering:
1. Do I need to explicitly set `created_at`? The raw SQL wasn't doing that, so I assumed that MySQL would take care of that field.
2. On the other hand, do I need the `updated_at`? Will our version of MySQL do the right thing if I don't supply this value?

nimmolo:
@JoeCohen — I don't know. I'd suggest calling it on a record in `rails console --sandbox` and inspect the record afterwards.

Jason may also have a good suggestion about `map`ping the records rather than `pluck`ing the values, i still haven't got that syntax under my belt but I believe it's generally preferable.

nimmolo:
@JoeCohen as you're doing these, just a heads up that the brakeman warning from [app/controllers/observer_controller/indexes.rb](https://github.com/MushroomObserver/mushroom-observer/pull/1051#diff-93dbb68a3dfa5c684d4abb77afd6ebbfd9df1d27ed9624123a8e1705b6b8024c) is already fixed in the `Refactor ObserverController` branch. I didn't backport it to `master` because the commits are moved to different files, and it would create a confusing merge conflict.

app/controllers/image_controller.rb
pellaea: Could this be simplified and made easier to understand (I always struggle with inject-style constructs):

```suggestion
CopyrightChange.insert_all(images_to_update.map do |image|
  { user: @user,
    updated_at: Time.current,
    target: iamge,
    year: image.when.year,
    name: old_holder,
    license: old_license }
end)
```

pellaea: Oops, I meant for that suggestion to replace the entire body of the method.

Also, I agree that the tiny difference in updated_at is negligible.  This is all about protecting ourselves legally -- having an audit trail of what the copyright used to be, in case someone did something with the image back when the copyright was different.  A few milliseconds won't make a difference! :)

pellaea:
I don't understand the question about created_at.  The copyright_changes table doesn't have a created_at column...  Am I missing something?

I don't think mysql "knows" anything about created_at or updated_at.  I think that's a rails convention.  It is possible that insert_all will automatically fill in updated_at, and that we can omit it.

JoeCohen:
@pellaea Thanks for the refactor. 
And thanks for pointing out that that the `copyright_change` table lacks a `created_at` field. I had assumed otherwise.
In Rails 6, `insert_all` does not automatically fill in timestamps.

(That will change in Rails 7, where "automatic setting of timestamp columns is controlled by the model's `record_timestamps` config". See, e.g., https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all
So CopyrightChange (and any other model tables that lack `created_at` or `updated_at`) will create an issue in Rails 7, 
We can deal that when we upgrade.)

nimmolo (approved):
Looks good.

PR #1052, opened 2022-06-23
`ExportController`

nimmolo:
This moves the last action from `ObserverController`/`Other` to its own controller, and readies `ObserverController` itself to be refactored. (Maybe in a separate PR.) 

Requests to `set_export_status` can come from `Image`, `Name`, `Location`, and the two `Description`s.

As a standalone controller, it seems analogous to the JavascriptController or the AdminController... but the test can't seem to figure out what controller to initialize. I've tried setting it explicitly here, but it throws "NameError: uninitialized constant ExportController".

I don't understand why that's a problem here, but not in the JavascriptController or any other of the new controllers. Suggestions welcome.


nimmolo:
> The Rails server instantiates the controller when it receives a request that's mapped to one of the handler methods. Rails is built so that you don't have to handle any instantiation yourself. Basically, Rails checks the controllers directory to satisfy the defined routes during runtime and instantiates whatever controller classes it finds. This programming pattern is referred to as Inversion of Control.

So, I believe the ExportControllerTest _is_ sending a request mapped to one of the handler methods. 
```ruby
  # Test setting export status of names and descriptions.
  def test_set_export_status
    name = names(:petigera)
    params = {
      id: name.id,
      type: "name",
      value: "1"
    }

    # Require login.
    get(:set_export_status, params: params)
    assert_redirected_to(controller: :account, action: :login)
...
```
Please correct me!

test/controllers/export_controller_test.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+# Controller tests for info pages
+class ExportControllerTest < FunctionalTestCase
+  def setup
+    @controller = ::ExportController.new
```
pellaea, line 8: Weird!  So this is what you need to add to this test for it to get the right controller??

nimmolo: That was a suggestion on SO. But the constant `ExportController` comes up uninitialized... and anyway the setup block shouldn't be necessary. 

Car was repaired, i'm headed back out. If you figure out what is going on here, please commit it and merge to the observer controller refactor!

pellaea:
Hooray!  Happy travels!

I'll look at this later maybe.

I'm helping write the description of a new species endemic to spruce-fir
forests of the southern Appalachians tonight, though.  I love doing tedious
measurements, whee!!  At least it has super-cool enormous muriform spores.
:)

JoeCohen:
@nimmolo The problem with the test is with the controller filename: it's missing `.rb`

JoeCohen:
@mo

nimmolo:
@JoeCohen re: `ok_for_export?` - if there is any relation between that method and `ok_for_export` (no `?`) i believe it is used a lot.

pellaea:
Yes, I think ok_for_export or variations are used in a lot of places... but I think all are associated with EOL, and I don't recall the status of that export.  It might be wise to keep the mechanism around, though, as we might find additional reasons to curate which content we approve for export in the future.

PR #1053, opened 2022-06-24
Replace byebug gem with debug gem

JoeCohen:
Replaces the `byebug` gem with the `debug` gem.
Byebug is incompatible with Zeitwerk. 
`debug` is the new default Ruby debugger, included in Ruby 3.1 and Rails 7.
(So we can remove the gem once we upgrade to Ruby 3.1.)

Delivers https://www.pivotaltracker.com/story/show/182558784

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -252,8 +252,9 @@ gem("graphql-batch")
 #
 
 group :test, :development do
-  # Use byebug as debugging gem
-  gem("byebug")
+  # https://github.com/ruby/debug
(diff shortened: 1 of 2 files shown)


nimmolo:
Any objections to merging this into `next`? Does anyone prefer to leave both gems activated on that branch for awhile?

JoeCohen:
I'm good with merging it (or leaving it activated until we have to remove
it).
I don't know if anyone else is using Byebug.

PR #1054, opened 2022-06-24
Fix export controller test

JoeCohen:
Makes `ExportControllerTest` find its controller
- Renames controller, adding `.rb` extension
- Removes test setup (unneeded)
- Corrects a comment


nimmolo:
Joe, THANK YOU. I found myself running over the code in my mind, even in the country, to try to discover the problem~!

PR #1055, opened 2022-06-25
Updates needed to get the system working natively on Apple M1 hardware

mo-nathan:


app/models/rss_log.rb
```diff
@@ -437,7 +437,7 @@ def decode_args(args)
   end
 
   def decode_time(str)
-    Time.parse(str).in_time_zone
+    Time.parse(str + "+0000").in_time_zone
```
pellaea, line 440: FWIW, Rubocop will almost certainly prefer "#{str}+0000".

pellaea (approved):
I *think* we can rip out jpegresize, too -- it's only referred to in comments in /app/models/image.rb and in README_DEVELOPMENT_INSTALL.  But it hardly seems to be causing any problems, and to rip it out will require a PR to startup, too.

This PR looks good to me just the way it is.

app/models/rss_log.rb
mo-nathan: I would have agreed with you, but I just checked and it doesn't seem to care.  Neither way generates a violation and it's exactly the same number of characters.  For some reason I don't fully understand I slight prefer the way I wrote it, but like Rubocop, I don't have strong opinion. :-)

pellaea:
Really??  I definitely prefer your way.  Sometimes I just don't understand
rubocop!  I am so sorry to have belabored the (non)issue.

nimmolo:
Maybe Rubocop doesn't care about the `str + var` concatenation here because `Time.parse` already casts the `var` as a string, or sanitizes it some other way...?

nimmolo:
I'm looking forward to getting this merged to make my 2013 iMac happy on 10.14 Mojave... 
just kidding! 
But glad you figured it out, Nathan.

PR #1056, opened 2022-06-27
Fix Brakeman RCEs

JoeCohen:
This PR fixes the Remote Code Execution issues flagged by Brakeman.
Instead of constantizing user params, it compares the param to a list of known good models.
See [The Safest Way to Constantize…](http://gavinmiller.io/2016/the-safesty-way-to-constantize/) (2016)
[Ruby Unsafe Reflection Vulnerabilities - Praetorian](https://www.praetorian.com/blog/ruby-unsafe-reflection-vulnerabilities/)
[Insecure Use of Dangerous Function | GuardRails](https://docs.guardrails.io/docs/vulnerabilities/ruby/insecure_use_of_dangerous_function#option-b-using-constantize-safely) (2022)

Delivers https://www.pivotaltracker.com/story/show/182567331

### DIFF
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -100,7 +100,8 @@ class CommentController < ApplicationController
   # "and more..." thingy at the bottom of truncated embedded comment lists.)
   def show_comments_for_target
     model = begin
-              params[:type].to_s.constantize
+              Comment.all_types.
+                      select {|model| model.name == params[:type]}.first
             rescue StandardError
               nil
             end
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -100,8 +100,7 @@ class CommentController < ApplicationController
   # "and more..." thingy at the bottom of truncated embedded comment lists.)
   def show_comments_for_target
     model = begin
-              Comment.all_types.
-                      select {|model| model.name == params[:type]}.first
+              Comment.all_types.find { |m| m.name == params[:type] }
             rescue StandardError
               nil
             end
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -99,7 +99,7 @@ class CommentController < ApplicationController
   # Shows comments for a given object, most recent first. (Linked from the
   # "and more..." thingy at the bottom of truncated embedded comment lists.)
   def show_comments_for_target
-    model = safe_model_from_param_type
+    model = Comment.safe_model_from_name(params[:type])
     if !model
       flash_error(:runtime_invalid.t(type: '"type"',
                                      value: params[:type].to_s))
@@ -316,9 +316,5 @@ class CommentController < ApplicationController
     redirect_with_query(controller: target.show_controller,
                         action: target.show_action, id: target.id)
     false
-  end
-
-  def safe_model_from_param_type
-    Comment.all_types.find { |m| m.name == params[:type] }
   end
 end
--- a/app/controllers/comment_controller.rb
+++ b/app/controllers/comment_controller.rb
@@ -104,7 +104,7 @@ class CommentController < ApplicationController
             rescue StandardError
               nil
             end
-    if !model || !model.acts_like?(:model)
+    if !model
       flash_error(:runtime_invalid.t(type: '"type"',
                                      value: params[:type].to_s))
       redirect_back_or_default(action: :list_comments)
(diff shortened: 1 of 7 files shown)


app/controllers/comment_controller.rb
```diff
@@ -99,12 +99,8 @@ def show_comments_for_user
   # Shows comments for a given object, most recent first. (Linked from the
   # "and more..." thingy at the bottom of truncated embedded comment lists.)
   def show_comments_for_target
-    model = begin
-              params[:type].to_s.constantize
-            rescue StandardError
-              nil
-            end
-    if !model || !model.acts_like?(:model)
+    model = safe_model_from_param_type
```
JoeCohen, line 102: NOTE: The new method performs slightly differently, and I think better than, the old one. In particular,
With params of a real, non-commentable model and real id:
- The old method would throw a runtime error, e.g., `Minitest::UnexpectedError:         RuntimeError: The model "APIKey" does not support comments!`
- The new method supplies a flash_error and returns gracefully.

app/controllers/observer_controller/other.rb
```diff
@@ -92,13 +95,12 @@ def set_export_status
     id    = params[:id].to_s
     type  = params[:type].to_s
     value = params[:value].to_s
-    model_class = type.camelize.safe_constantize
+    model_class = EXPORTABLE_MODELS.find { |m| m.name.downcase == type }
```
JoeCohen, line 98: Question: Should I get rid of the explanatory constant (and just plug the array in here?

app/controllers/observer_controller/search_controller.rb
```diff
@@ -65,7 +67,8 @@ def advanced_search_form
     @filter_defaults = users_content_filters || {}
     return unless request.method == "POST"
 
-    model = params[:search][:model].to_s.camelize.constantize
+    model = ADVANCED_SEARCHABLE_MODELS.
```
JoeCohen, line 70: QUESTION: Should I get rid of the explanatory constant?

app/controllers/observer_controller/other.rb
JoeCohen: @pellaea: Is there way to get the db to supply the names of tables that have a `okay_to_export` column?
I took a first stab with the following (reformatted fro readability).
```
ActiveRecord::Base.connection.execute(%(
  SELECT 
  sys.columns.name AS ColumnName, 
  tables.name AS TableName 
  FROM 
  sys.columns 
  JOIN 
  sys.tables ON sys.columns.object_id = tables.object_id 
  WHERE 
  sys.columns.name = 'ok_for_export'
))
```
And got
>client.rb:148:in `_query': Mysql2::Error: SELECT command denied to user 'mo'@'localhost' for table 'columns' (ActiveRecord::StatementInvalid)
 /home/vagrant/.rvm/gems/ruby-3.0.4/gems/mysql2-0.5.4/lib/mysql2/client.rb:148:in `_query': SELECT command denied to user 'mo'@'localhost' for table 'columns' (Mysql2::Error)

app/controllers/comment_controller.rb
```diff
@@ -321,4 +317,8 @@ def check_permission_or_redirect!(comment, target)
                         action: target.show_action, id: target.id)
     false
   end
+
+  def safe_model_from_param_type
+    Comment.all_types.find { |m| m.name == params[:type] }
+  end
```
pellaea, line 323: Just an idea -- why not move this into Comment, then Query will have access to it, too.

app/controllers/observer_controller/other.rb
pellaea: I don't think Nimmo will like any solution that involves explicit SQL.  He;'s nearly eradicated all SQL from the code by now.  Let's keep it that way.

It seems to me, though, you can probably work something out equivalent.  If there were a way to get a list of all ActiveRecord subclasses, for example, then you can filter those for ones that respond to okay_to_export, or have that attribute.  Since all of the models you care about here are subclasses of AbstractModel, maybe we could somehow use that to supply the list of all model classes.  Maybe provide a class method like AbstractModel.models_with_attribute(:okay_to_export).

Just some thoughts.

pellaea (approved):
I think the explanatory constants are fine.  And I think this PR is fine the way it is.

app/controllers/comment_controller.rb
JoeCohen: Good idea. Maybe I should rename it `safe_constantize`. :laughing:

PR #1059, opened 2022-06-30
Jdc sql injection site data field count

JoeCohen:
This PR fixes a single SQL injection warning in `SiteData#field_count`

It does this by refactoring `field_count` to replace
(a) SQL queries composed from string fragments with
(b) complete (uncomposed) AR queries.

(The refactor also fixes an ABC offense.)

### DIFF
--- a/app/classes/site_data.rb
+++ b/app/classes/site_data.rb
@@ -144,11 +144,11 @@ class SiteData
   FIELD_QUERIES = {
     contributing_users:
       User.
-        where(User[:contribution].gt(0)),
+        where(contribution: 1..),
     observations_with_voucher:
       Observation.
         where(specimen: true).
-        where("LENGTH(observations.notes) >= 10").
+        where(Observation[:notes].length >= 10).
         where.not(thumb_image_id: nil),
     observations_without_voucher:
       Observation.
--- a/app/classes/site_data.rb
+++ b/app/classes/site_data.rb
@@ -133,7 +133,7 @@ class SiteData
   FIELD_QUERIES = {
     contributing_users:
       User.
-        where("contribution > 0"),
+        where(User[:contribution].gt(0)),
     observations_with_voucher:
       Observation.
         where(specimen: true).
--- a/app/classes/site_data.rb
+++ b/app/classes/site_data.rb
@@ -50,11 +50,11 @@ class SiteData
   #  FIELD_TABLES::     Table to query.
   #  FIELD_CONDITIONS:: Additional conditions.
   #
-  #  The basic query for stats for the entire site is:
+  #  The default query for stats for the entire site is:
   #
   #    SELECT COUNT(*) FROM table
   #
-  #  The basic query for stats for a single user is:
+  #  The default query for stats for a single user is:
   #
   #    SELECT COUNT(*) FROM table WHERE user_id = 123
   #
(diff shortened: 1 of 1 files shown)


app/classes/site_data.rb
```diff
@@ -142,6 +138,34 @@ class SiteData
     contributing_users: "contribution > 0"
   }.freeze
 
+  # Non-default unified queries for stats for the entire site
+  # rubocop:disable Layout/MultilineMethodCallIndentation
+  # Rubocop 1.30.0 wants to allgn "where" with the open brace on the next line.
+  FIELD_QUERIES = {
+    contributing_users:
+      User.
+        where(User[:contribution].gt(0)),
```
JoeCohen, line 147: My First Arel! 
Once I figured out the syntax, this was trivial thanks to the examples and gems provided by @nimmolo.

app/classes/site_data.rb
```diff
@@ -142,6 +138,34 @@ class SiteData
...
+    observations_with_voucher:
+      Observation.
+        where(specimen: true).
+        where("LENGTH(observations.notes) >= 10").
```
JoeCohen, line 151: This could also be Arel-ized. I left it as-is because `
- LENGTH` seems to be a basic (universal?) SQL function
- the replace Arel looked complicated to me. I.e., something like:
```ruby
Arel::Nodes::NamedFunction.new('LENGTH', [Observation.arel_table[:notes]]).gteq(10)
```
But I wouldn't mind Arel if there's a shorter way to it, or even if others prefer the longer Arel to SQL.

JoeCohen:
@pellaea: Am I messing up performance by doing things this way (uncomposed, single AR query vs composed SQL string).

app/classes/site_data.rb
pellaea: Can you help me understand these named functions?  Are they really database-agnostic?  Couldn't we name any random mysql-specific function here?   The point, after all, is not to turn things into Arel, but to make it database-independent.  But there must be an Arel string length method along the lines of "gt", etc., right?  As you say, this must be a universal.  But I wouldn't be completely surprised if it's not spelled "LENGTH" in every database.

nimmolo: @JoeCohen — Congrats on the Arel! Here's a [reference for arel-extensions](https://github.com/Faveod/arel-extensions). 

I believe @pellaea is right about named functions, they do not create database-agnostic SQL but rather allow us to write db-specific SQL, which is not what we're going for. It took me awhile to figure that out.  

But arel-extensions is here to help. What I think you'd want is:
```ruby
Observation.
  where(specimen: true).
  where(Observation[:notes].length >= 10)
```
Note that it allows you to also keep the `>=` operators. It's pretty convenient!

app/classes/site_data.rb
nimmolo: I think a better way to go here is pure AR, thanks to our recent updates we can use endless ranges:
```ruby
User.
  where(contribution: 0..)
```

JoeCohen: @nimmolo Thanks for pointing me the right way on endless ranges.
After some experimentation, I realized that `0..` is `>= 0`. https://ruby-doc.org/core-3.1.1/Range.html#class-Range-label-Endless+Ranges
Fortunately, user contribution scores are integral, so we can do
```ruby
User.
  where(contribution: 1..)
```

nimmolo (approved):

pellaea:
Thank you for reminding me.  Yes, the site data stuff is very expensive.
It is used by both show_site_stats and show_user.  And since performance
now seems to be completely non-intuitive to me, I would strongly recommend
comparing performance before and after this PR.  On a real DB snapshot, of
course.

PR #1063, opened 2022-07-02
Fix confusion about whether Image#glossary_terms includes the thumbnail.

pellaea:
Answer is: it should.  That's how observation thumbnails work, so it really should work the same for glossary terms.  I'll make sure the database reflects this.  If not, we might need to do some more work on this PR.

Therefore I removed the now-redundant Image#all_glossary_terms, and I fixed the fixtures (only one glossary term was affected, I think).

Also, I changed Image#get_subjects to all_subjects because I think rubocop doesn't like methods with the prefix "get_".  I wish there were a better name for this to disambiguate it with respect to Image#subjects which refers exclusively to users who use this image for their profile.  Maybe we can change Image#subjects to profile_users?  Then use just plain old generic "subjects" for what used to be "get_subjects" (and is now at least temporarily called "all_subjects").

### DIFF
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -400,20 +400,6 @@ class Image < AbstractModel
     size == :original ? original_extension : "jpg"
   end
 
-  def has_size?(size)
-    max = width.to_i > height.to_i ? width.to_i : height.to_i
-    case size.to_s
-    when "thumbnail" then true
-    when "small" then max > 160
-    when "medium" then max > 320
-    when "large" then max > 640
-    when "huge" then max > 960
-    when "full_size" then max > 1280
-    when "original" then true
-    else; false
-    end
-  end
-
   # Calculate the approximate dimensions of the image of the given size.
   def size(size)
     w = width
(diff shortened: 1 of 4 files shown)


pellaea:
There were several cases of glossary_terms.thumb_image_id not being in glossary_term_images, but they were all old, so I will conclude that the code does now ensure that all images attached to a glossary_term appear in glossary_term_images, even the thumbnail.  Meanwhile, I fixed all the old cases where that was not true.

JoeCohen:
Thanks for doing this; it looks great.
Given that the PR touches `app/models/image.rb`, could we rip out `Image.has_size?`; I can't find any reference to it.

pellaea:
Why not?  Sure.  Would you like to do it?

JoeCohen:
Sure. Sounds like fun!

pellaea:
Yes!  Ripping code out is always fun! :)

JoeCohen:
Bonus: Coveralls check passed, coverage increased.

PR #1070, opened 2022-07-06
Pending performance cops

JoeCohen:
This PR sets up pending Performance cops. It
- Configures the cops by adding appropriate lines to `.rubocop.yml`;
- Fixes current offenses

It's part of an effort to deal with the backlog of unconfigured Pending cops. 
The main reason for fixing the offenses is to avoid "new" Codeclimate offenses in other developers' branches.
(I assume -- from the cops' names -- that each fix also improves performance very slightly. 
The cop documentation is at https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancesum.)


test/models/name_test.rb
```diff
@@ -290,7 +290,7 @@ def test_format_string
   end
 
   def test_upper_word_pats
-    pat = /^#{::Name::Parse::UPPER_WORD}$/
+    pat = /^#{::Name::Parse::UPPER_WORD}$/o
```
pellaea, line 293: I like having the regular expression right there where it's used, but I feel like the creation of the regex doesn't have to happen each time we run test_upper_word_pats... could it be turned into a (frozen?) global just outside of the method?  Or does the "o" flag basically make the regex a persistent constant that's equivalent to a global?

pellaea (approved):
Looks like this was a lot of work.  Thanks, Joe.  I made a few trivial comments to make it look like I was contributing something, but they can be blissfully ignored. ;)

JoeCohen:
@pellaea re frozen globals. You're probably right. I was too lazy to do that in every case.

JoeCohen:
@pellaea Re regex `/o` option. It memoizes the interpolation. 
https://ruby-doc.org/core-3.1.2/Regexp.html#class-Regexp-label-Options
It's a help, but not as good as creating a frozen regexp.

pellaea:
I was curious, so I dug some more.  I think it is essentially as good as a
frozen global.  See this test:

def test(var)
  pat = /#{var}/o
  puts pat.inspect
end
test("one")
test("two")

This produces:

/one/
/one/

So it *is* remembering from one call to the next.  Exactly as you said.
"memoize" is the perfect way to think of it.  That is perfect for our case
-- no need to create that regex until it is actually needed (lazy
creation).  But once it's done so, no need to ever do it again.  And that's
exactly what the //o flag does.  Sweet.  Thanks!

JoeCohen:
@pellaea Thanks for digging into the `/o` option. I didn't know about it until working on this branch. 
Rubocop suggested it as one way to fix the offense.

pellaea:
Its origins can be traced all the way back to Perl in the 90s I think, but
I am pleasantly surprised by ruby's particular implementation of it.  That
is very useful, indeed.  Good to know!

PR #1074, opened 2022-07-08
Remove legacy redirects of form routes from routes and `RedirectTests`

nimmolo:
Per discussion with @pellaea and @JoeCohen, form routes wouldn't be bookmarked anywhere and don't need redirects from the old URLs. People needing the forms should be able to find them from the redirected :show, :index pages and their generated (updated) links.

This also removes the RedirectTests for those routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -614,7 +614,9 @@ MushroomObserver::Application.routes.draw do # rubocop:todo Metrics/BlockLength
 
   # ----- Articles: standard actions --------------------------------------
   resources :articles, id: /\d+/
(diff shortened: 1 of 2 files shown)


nimmolo:
Note: the Article#destroy redirect tests were causing trouble on the ObservationsController PR, somehow. Since we don't need them, this eliminates that problem.

This seemed like it should a separate PR however, so i'd like to merge it first if it passes review.

pellaea (approved):
I only dimly understand routes, so I trust you.  Looks like you removed quite a number of apparently superfluous tests.  Hopefully this speeds up the test suite nontrivially!

Issue #1076, opened 2022-07-08
Links to textile broken


JoeCohen:
They are:
`https://mushroomobserver.org/info#textile_sandbox`
should be
`https://mushroomobserver.org/info/textile_sandbox`

JoeCohen:
merged & deployed fix.

pellaea:
Thanks, Joe!  I noticed this with how_to_help, too.

PR #1077, opened 2022-07-08
Fix textile_sandbox route

nimmolo:
It was an ordering question in the routes, `textile` must come after `textile_sandbox` perhaps


JoeCohen:
I should have clarified Issue #1076 by saying that it was broken on the website. Sorry.
It's not broken (or at least not broken the same way) in the master branch as of [2ecb07e](https://github.com/MushroomObserver/mushroom-observer/commit/2ecb07e15782122021f6618df0aa6ceff143d766).
To replicate:
- goto any Observation or Name
- click `Add Comment`. 
- click `Textile markup system`.

JoeCohen:
Because this works in master, I will deploy master now, then look at this PR.

JoeCohen (commented):
The Textile link does work locally in this branch. 
Should there be a test to prevent reversion? 
If you don't think so, please just merge.

nimmolo:
Ok. 
I'm not 100% sure `textile` actually has to come after `textile_sandbox`; I believe the routes are matched between `/`, but this couldn't hurt (other than providing a possibly-misleading note.) Additionally, the aliased action "textile" is not used in the code anywhere anymore, and the entire route and alias could be deleted, I believe.

JoeCohen:
One of us should merge this after all the checks (other than Brakeman) pass.

nimmolo:
Aha Joe - the obsolete route "/observer/textile" is hardcoded into the translation string in en.txt and presumably in all translations!

Let's fix it

nimmolo:
Waait a minute. Will my changing "en.txt" effectively void a bunch of translations? Should I revert that?

There has to be a way to change these routes embedded in the translations master file.

nimmolo:
@pellaea please check my question above about changing hardcoded old routes in en.txt, that end up directly output as links in the front end. Will changing the english translation void all other translations? Or would that only happen when changing **keys**?

pellaea:
Here's a copy of my overly-energetic response from earlier today. :)

"What's wrong with translations "hard-coding" old routes?  This is one of
the reasons we keep aliases for the old routes!  Changing the content of an
existing tag in en.txt will not void anything.  And there's no harm in
updating any old-style links you may find in there.  I just wouldn't stress
over it.  Way more important fish to fry!"

nimmolo:
Ok, well - they're updated now! It was only 5 minutes work. Glad it doesn't mess anything up.

nimmolo:
Also, the reroute for "observer/textile" is still active, so there shouldn't be a problem with existing translations either. Will merge.

PR #1078, opened 2022-07-08
Unlock jQuery version in `application.js`

nimmolo:
Up to now we've been bundling a specific and very old version of `jquery.js` in the `javascripts` folder via `application.js`.
This PR removes that version and allows Rails to use the current version.


pellaea:
I'm mightily confused.  It looks for all the world like you simply deleted jquery.js without replacing it with anything!  How is the one rails wants loaded?  Where is that included in the page layout?  or application.js?

pellaea:
(Sorry, pressed the wrong button!)

nimmolo:
Somehow it comes bundled with the `jquery-rails` gem. You don't have to save the file in /javascripts
If you spin up the local app on this branch, you'll see that the version has bumped from 1.11 to 1.16.

However, thanks for questioning! because I just noticed "unlocking" it only gets to the latest version 1. To get on jQuery 3, we have to `//= require jquery3` in applications.js... 

I just did that locally, but have picked up a js error, will fix then commit.

nimmolo:
The error related to a single change in syntax for jQuery:

`$(window).load(function () {`
is now
`$(window).on('load', function () {`

The PR now specifies jQuery 3, changes all instances where we called window.load, and introduces a polyfill to avoid future gotchas.

pellaea:
Oooh, I use the old style all the time.  Yikes.  I hope that doesn't cause
a lot of grief on MO.  Does this just apply to load or to all callback
types?

nimmolo:
I think it's for events. jQuery's (window).load was always a shorthand for (window).on("load") but it 
... uhhh... 
maybe conflicts with other uses of load? Anyway they deprecated it. I can't remember why. I replaced it everywhere in our JS anyway.

nimmolo:
Could be, but this was the only one causing a javascript error. I think we're in the clear here.

nimmolo:
https://bugs.jquery.com/ticket/11733

pellaea:
It looks like it just applies to load because -- as you say -- ambiguous
signature.  It could apply to javascript or window or image.  Not sure of
the details (what does "signature" mean in this context??), but that was
good enough for me!  Thanks!

nimmolo:
`Signature` refers to getting your autograph on this PR!

nimmolo:
....code delirium

pellaea:
Oh man, I guess I do need to do that...  Fair warning: this is not going to
be a terribly thorough review!  I have half a mind to say: just throw the
dice already! :)

pellaea:
Oh!  *This* PR...  uhhhh...  oh hell, sure, why not.  I'm *terrible* at testing code on the dev VM.  (It's just so blasted slow...)  Go ahead and merge it.  Will it go out tonight with the cookie change, too?

nimmolo:
What's a jQuery-heavy page I could test?

nimmolo:
Aha... I found instances of window.unload. Need to fix every command in that document.

nimmolo:
Will fix after i disable my autoformatter for JS, too many changes to review.

pellaea:
Probably the only really important pages, now that you have me thinking
about it, are the observation form, especially create observation, and the
lightbox thingy for showing images.  Pay special attention to any popups
like autocomplete.  And the image stuff.  That image stuff was all Ray
Suelzer's code, so it's probably solid.  I might not trust my autocomplete
code as much.  And the show vote popup on the observation page is another
one I bet is used a lot.  The pulldowns for language and user prefs are
worth a glance, too.

Got all that???  Show and create observation, basically.

JoeCohen (commented):
I'm js ignorant.

nimmolo:
Methods updated, interface seems copacetic.

PR #1084, opened 2022-07-12
Fix LookupObjects SQL injection warning

JoeCohen:
Use AR instead of SQL

### DIFF
--- a/app/classes/query/modules/lookup_objects.rb
+++ b/app/classes/query/modules/lookup_objects.rb
@@ -41,10 +41,9 @@ module Query
         project_ids = lookup_projects_by_name(vals)
         return [] if project_ids.empty?
 
-        SpeciesList.connection.select_values(%(
-          SELECT DISTINCT species_list_id FROM project_species_lists
-          WHERE project_id IN (#{project_ids.join(",")})
-        ))
+        SpeciesList.joins(:project_species_lists).
+          where(project_species_lists: { project_id: project_ids }).distinct.
+          pluck(:id)
       end
 
       def lookup_species_lists_by_name(vals)
--- a/app/classes/query/modules/lookup_objects.rb
+++ b/app/classes/query/modules/lookup_objects.rb
@@ -43,7 +43,7 @@ module Query
 
         SpeciesList.joins(:project_species_lists).
           where(project_species_lists: { project_id: project_ids }).distinct.
-          pluck(:id)
+          map(&:id)
       end
 
       def lookup_species_lists_by_name(vals)


app/classes/query/modules/lookup_objects.rb
```diff
@@ -41,10 +41,9 @@ def lookup_lists_for_projects_by_name(vals)
         project_ids = lookup_projects_by_name(vals)
         return [] if project_ids.empty?
 
-        SpeciesList.connection.select_values(%(
-          SELECT DISTINCT species_list_id FROM project_species_lists
-          WHERE project_id IN (#{project_ids.join(",")})
-        ))
+        SpeciesList.joins(:project_species_lists).
+          where(project_species_lists: { project_id: project_ids }).distinct.
+          pluck(:id)
```
nimmolo, line 46: This looks good. But I'd suggest generally switching from `pluck(:id)` to `.map(&:id)` following @pellaea's benchmarking. 

I'm amazed you're finding things to convert in the Query class, i was scared to touch it!

nimmolo (approved):

app/classes/query/modules/lookup_objects.rb
JoeCohen: Thanks for the reminder about pluck vs map.
I will make that change.

JoeCohen: I actually benchmarked this with a single run of each alternative:
```ruby
[#<Benchmark::Tms:0x00007f92d86b6f30
  @cstime=0.0,
  @cutime=0.0,
  @label="pluck:",
  @real=0.0013135079989297083,
  @stime=3.399999999997849e-05,
  @total=0.0008969999999989264,
  @utime=0.0008629999999989479>,
 #<Benchmark::Tms:0x000055c1ef274ca8
  @cstime=0.0,
  @cutime=0.0,
  @label="map:",
  @real=0.0012400790001265705,
  @stime=2.999999999997449e-05,
  @total=0.0007929999999991555,
  @utime=0.000762999999999181>] 
3.0.4 :028 > project_ids
 => 50..100 
```

nimmolo:
@JoeCohen Can you show me how you do your benchmarks?

JoeCohen:
I did a crude Benchmark in the console, arbitrarily setting local variables.
See How to Benchmark my code in Ruby or Rails
<https://www.lynnbright.com/learn-how-to-use-ruby-benchmark-module/>

nimmolo:
Thanks for the article. Did you write out the test as a file, and execute it in the console as she suggests? Or just type the code directly into console... 

sorry for my ignorance! I don't even know how to execute a file from the console, in the context of rails. Do you use `rails runner`?

JoeCohen:
I typed/inserted the text in the console.
I'm equally ignorant, so that seemed like the easiest way to do it. I was
also unclear about how to do it as a file in the context of MO.

nimmolo:
Got it, thanks. Do you use a debug console, or just `rails console`? 

I find that rails console acts unpredictably if I paste in multi-line code, at least in VSC. I had better luck with `pry`, when I had that gem installed, and also with debug.

JoeCohen:
I use rails console. I hadn't thought of other ways to do it. And VS Code
does display unpredictably when the lines are too long.

nimmolo:
Roger. If i find a better way, I’ll share it!

nimmo

pellaea:
Your way works great and there really is no clear reason to ever use the
command line, but since that is the way I do it, I thought I would briefly
share how I write command line scripts that run the MO env.

Basically, just follow the scripts.  In particular, I usually use
scripts.mushroom_mapper.rb as the template:

#!/usr/bin/env ruby
# frozen_string_literal: true
require(File.expand_path("config/boot.rb", __dir__))
require(File.expand_path("config/environment.rb", __dir__))
<do stuff here>
exit(0)

Save that as, say, /vagrant/mushroom-observer/test.rb, then

vagrant/mushroom-observer> chmod a+x test.rb
vagrant/mushroom-observer> ./test.rb

That's it!

(Note that I removed the ".." from the path of boot and environment.rb
because the script lives in the main MO dir, not scripts.)

nimmolo:
Ahaa. This is helpful!

Definitely preferable to have the script in a fiie. It can get annoying running a multi line bit of code more than once in the console, because I have to keep pasting it in.

I may add this to the developer docs.

nimmo

PR #1085, opened 2022-07-13
Add legacy observation action redirects, reorganize `routes.rb` a bit

nimmolo:
I forgot to add the legacy redirects for old Observation routes to the ObservationsController PR.
This adds redirects for those old routes. 

It also reorganizes `routes.rb` so the "normalized" routes are in one block, the original routes are in another, and the legacy redirects are in a third block. Goal is to improve legibility, both of the `routes.rb` file and of the generated routes.


nimmolo:
This PR should probably merge in some form before the ObservationsController PR is deployed, because it is intended to keep bookmarked search links working (per @JoeCohen, thank you!)

JoeCohen (approved):
Thanks for re-organizing this. It looks good to me.

PR #1086, opened 2022-07-13
Bump rails constituents to 6.1.6.1

JoeCohen:
Responds to [[CVE-2022-32224] Possible RCE escalation bug with Serialized Columns in Active Record](https://groups.google.com/g/rubyonrails-security/c/MmFO3LYQE8U?pli=1).
This is an alternative to Dependabot PR #1083, which would bump us to 7.0.3.1.


nimmolo (approved):
Hmm... the line `group :rails do`is pretty brilliant! Where'd you find that?

Trying `bundle install` on this branch locally I find that line doesn't install all of rails.

JoeCohen:
Bundler lets you group gems to make it easier to manage the group.
It's just a group of gems that I decided to name `rails`. I could have chosen anything.  It might be better as `selected_rails_constituents`.

PR #1089, opened 2022-07-14
`NamesController` and `NameDescriptionsController` — normalized CRUD+

nimmolo:


nimmolo:
@mo-nathan — In this case I feel like "Description" is indeed a use case for a concern. It's not hiding complexity and is a class of shared code.

In any case it is already abstracted out in current master as a separate file that is included in both controllers, but "concerns" have an established Rails usage, even in core Rails, in the sense of a "shared concern" of two classes, and they do have a handy syntax. Calling it a `concern` highlights it as a shared thing and makes it easier to locate, to my way of thinking.

nimmolo:
Question for the refactor - How do people feel about doing the routing of descriptions like this: 
from
`/name_descriptions(:id)` 
to
`/names/:name_id/descriptions(:description_id)`

The list of descriptions for a name would be at `names/:name_id/descriptions/` and the edit form for a description would be at `names/:name_id/descriptions/:description_id/edit`

To me this makes more sense. A NameDescription is not a free-floating resource, but always a resource of a Name.

pellaea:
Yes!  I like those routes.

nimmolo:
Note: I'm aiming to finish this up with the minimum refactoring, so it's easier to review. It's already so muuch renaming and debugging.

In other words, I'm not aiming to fix ABC or method length violations this PR, and can either disable the cop everywhere it calls out existing code, or just leave it to merge with the violations. If @JoeCohen wants to tackle them, though, I'd be happy about it.

JoeCohen:
@nimmolo I'll have a look once you think it's in shape to merge. (But I'm out of pocket tomorrow.)

nimmolo:
@JoeCohen Thanks. 
If we're refactoring for ABC in separate PR's though, maybe ObervationsController is a better one to start on, since it's already merged and has plenty of jumbo methods.

JoeCohen:
@nimmolo Another option for you -- if you get this to passing (and maybe fix the conflicts) I could pick up where you left off: deal with the remaining Codeclimate non-metric issues (and the conflicts if you don't have time to deal with them).

JoeCohen:
Oh! I see. Not much of an offer, as all the Codeclimate issues are metric issues or TODOs.

nimmolo:
Want to confirm that i am indeed retiring `bulk_name_edit`: the route, the action and its tests with this PR.

pellaea:
I still think that's a good idea.  People have had ample opportunity to
voice disagreement.

nimmolo:
New thing: This PR introduces nested routes for descriptions (in this case only Name). 
`/names/:name_id/descriptions/:id`
`/names/:name_id/descriptions/:id/edit` etc
This means that all the route builders in controllers and views need an additional named parameter to route a NameDescription resource, the `name_id`. Many route builders are generic for both types of descriptions, though, so I needed a method that would return this extra param, built properly for whichever resource. 

I settled on a new method in the Description model, `#parent_param`, that returns a merge-able hash containing the named param and the description's parent_id. You can call it on a description instance, like `desc.parent_param`, or in a route builder,
```ruby
redirect_to( { controller: desc.show_controller, action: desc.show_action, id: desc.id }.merge(desc.parent_param))
```
For a `Model.controller_normalized? == false`, i.e. `LocationDescription`, it returns an empty hash, so it's safe to use in generic description route builders. If `LocationDescription` gets normalized, it just works with no further modifications. Eventually we can take out the guard clause about `controller_normalized?`.

app/controllers/names_controller/index.rb
```diff
@@ -0,0 +1,200 @@
...
+    # Add "show descriptions" link if this query can be coerced into a
+    # description query.
+    # FIXME: This link makes no sense. Needs a name_id?
+    if query.coercable?(:NameDescription)
+      @links << [:show_objects.t(type: :description),
+                 add_query_param({ controller: :name_descriptions,
+                                   action: :index },
+                                 query)]
+    end
+
```
nimmolo, line 179: @JoeCohen @pellaea  — 

In this block generating the right-tab `@links` for a Names index, I don't understand what the above and below blocks are intended do here: the ones generating links that should be added to the instance variable array if `query.coercable?(:NameDescription)` or if `query.flavor == :with_descriptions`.

First of all, I can't find where these links are output in any Name index i'm able to get by interacting with the front end. But i'm probably missing some kind of unfamiliar interaction. 

Also relevant, with this refactor, is the fact that any `NameDescriptionsController#index` requires a `name_id` - there's no such thing as an index of NameDescriptions per se, if there ever was. Whatever the above link is supposed to generate, it can't do it under this refactor without a `name_id`.

pellaea:
re: descriptions link -- Fair enough.  I can't convince the existing site
to produce that link, either.  Probably just means it's not working!  But
no loss, as no one has noticed its lack up till now.

re: Show Observations link on name index -- Easy to get this on the
existing site.  Do any observation index, click on "Show Names", and voila,
the resulting name index will include a "Show Observations" link to allow
you to return to the original observation index.

"Coerce" in this context means convert a query for observations into a
query for corresponding names (or locations or images).  And vice versa.

FWIW, at the time I didn't know anything about subqueries, so I implemented
it incorrectly.  I literally cobbled together a new query, duplicating all
sorts of code in the process.  So a "by user" observation query would
become a "with observations by user" name query.  Every observation query
flavor has a corresponding "with observations xxx" flavor for name,
location and image.  Instead, I should have done it as a generic "with
observations" name/location/image query where it just wrapped the original
observations query as a subquery "select * from names where id in (select
name_id from observations where ...)".  Live and learn, I guess.  But it's
worth knowing this to understand what Query#coerce is doing.

nimmolo:
@pellaea — Also, thank you for explaining coerced queries. 

If i take on the project to convert the Query class to use AR chained queries, do you imagine that I'd want to rewrite the coerced-queries as subqueries, then? 

I won't take that on in this PR, but it's interesting to start thinking about.

pellaea:
Okay, sorry, that code concerning with_descriptions is adding extra columns
to the name index for with_descriptions name queries.  I'm still struggling
to figure out how to create a with_descriptions name query on the site.
Maybe if we could figure that out we can also figure out how to get the
other link I failed to find.

I never use name descriptions.  Despite having written that code very early
on, I've never used it, don't know if anyone else uses it any more, and
don't remember how most of it works.  It was all written in support of some
classes Anne Pringle was teaching years ago.

Sorry!

pellaea:
Dude!  Here it is!  There's a hidden page:

https://mushroomobserver.org/name/authored_names

Check it out.  Wacky.  No idea what it's for.  Some admin thing, I would
guess. :)

PS. concerning how to rewrite Query / phase Query out -- One thing I would
personally like to see is to remove the concept of Query flavors.  Instead
have all queries be "all" queries with various parameters.  That would have
some significant bearing on the implementation of coercion.  It's also
possible that we should move coercion outside of the query altogether.  I
thought coercion would be more broadly and abstractly applicable to a
number of situations.  Instead it is restricted to just three cases:
displaying an alternate index of names / locations / images corresponding
to observations.  Well, hell, why not just rip out coercion altogether and
write those three alternate indexes?  Just thoughts to consider.

nimmolo:
Nice! 

I love that the text styling tips you off that it’s an Easter Egg page, and things may not be as they seem. 
It’s like the bonus hidden castle in Atari Adventure...
image

Need to eat dinner, and I will think on these things.

pellaea:
Haha!  Yes, totally an easter egg. :)

nimmolo:
re **Query** — I'll start a dedicated thread for this query-redo roadmap, in "Discussions". Psyched to plan work on it.

re **Names**:
That's very funny to think that it was purpose built code for a class 10 years ago, and may not be much used currently. But, it seems useful that one can access a list of academic descriptions of a taxon from the **show name** view. 

I suppose it's useful to see this list for the purposes of assigning description writing assignments to students. That's not exactly what this index is, though, is it? That index would be more like "Names needing descriptions". Also, this index of "Names with descriptions" doesn't seem to show multiple descriptions of the same name in the same row. or.... Does it?!

- It seems to me you're saying, and i notice from comparing "Names-index" to "Names-with-descriptions-index", that these are basically both versions of the name index (layout-wise, a table), but one has more columns. Are there other differences? (There are different rows!) Maybe the version of the table with the extra columns should be the default? Or a simplified version of it? But it will be slower due to the extra queries. 🤔 Hmmm.

- In current `name/authored_names`, what does " - - -    n o t   t h e   d e f a u l t   - - - " mean?

- I'm trying to figure out if the new `NameDescriptionsController` needs an `index` method. The most obvious reason, i guess, would be to output a **list of descriptions** for a given name (a partial, in the `NamesController#show` view). But is `NameDescriptionsController#index` really the place for that? 
- 
- The other possible "indexes" i can imagine for name descriptions seem to be not _lists of name descriptions_, but **lists of names having descriptions authored (or edited) by a specified user**. I'm going to play around and try to find these in the front end (dev version) by logging in as Nathan and poking around. Surely he has authored and edited descriptions.

pellaea:
I'm afraid if I ever knew, I've forgotten all about this code.  (names with
descriptions index)  Your guess is as good as mine!

Also, minor point, but I'm not so sure we're "redoing" Query as getting rid
of it.  The parts of it that we still need -- e.g., persistence (that is,
parsing of a hash of parameters and turning them into a chain of scopes) --
should maybe be separated into their own classes?

nimmolo:
Noted. Thanks for letting me know, even if it's that you don't remember something. 

For sure the most stressful and time-consuming part of these PRs, by far, is trying to decipher code that i don't know the purpose of, and handle the functionality properly when I have no idea what that purpose is. 

I will probably need to ask a bunch more questions like this as I go through the controllers, and all answers are helpful — even to know that you don't remember what something does. I will try to ask them more at the outset, because the time sink happens when i'm already knee-deep in the controller and I realize i don't actually understand a whole batch of routes.

PR #1090, opened 2022-07-14
Pending lint cops

JoeCohen:
This PR sets up Lint department cops that are pending as of Rubocop 1.31.1:
- Configures the cops (explicitly enables/disables each of them in `.rubocop.yml`
- Fixes offenses or disables them inline in the code

This PR delivers part of https://www.pivotaltracker.com/story/show/182027741


.rubocop.yml
```diff
@@ -66,8 +66,59 @@ Layout/LineContinuationLeadingSpace: #`new in 1.31.1
...
+  Enabled: true
+Lint/IncompatibleIoSelectWithFiberScheduler: # new in 1.21
+  Enabled: true
+Lint/LambdaWithoutLiteralBlock: # new in 1.8
+  Enabled: true
+Lint/NoReturnInBeginEndBlocks: # new in 1.2
+  Enabled: true
+Lint/NonAtomicFileOperation: # new in 1.31
+  # Disable until I understand the cop's implications - jdc 2022-07-14
+  # https://docs.rubocop.org/rubocop/1.31/cops_lint.html#lintnonatomicfileoperation
```
JoeCohen, line 99: Here are the offenses:
>app/models/language_tracking.rb:97:28: W: [Correctable] Lint/NonAtomicFileOperation: Remove unnecessary existence checks File.exist?.
    FileUtils.mkpath(path) unless File.exist?(path)
                           ^^^^^^^^^^^^^^^^^^^^^^^^
test/general_extensions.rb:478:40: W: [Correctable] Lint/NonAtomicFileOperation: Remove unnecessary existence checks File.exist?.
        File.delete("#{filename}.old") if File.exist?("#{filename}.old")
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/general_extensions.rb:488:33: W: [Correctable] Lint/NonAtomicFileOperation: Remove unnecessary existence checks File.exist?.
      File.delete(new_filename) if File.exist?(new_filename)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/models/ip_stats_test.rb:115:35: W: [Correctable] Lint/NonAtomicFileOperation: Remove unnecessary existence checks File.exist?.
    File.delete(MO.ip_stats_file) if File.exist?(MO.ip_stats_file)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/models/script_test.rb:115:31: W: [Correctable] Lint/NonAtomicFileOperation: Remove unnecessary existence checks File.exist?.
    FileUtils.rm(output_file) if File.exist?(output_file)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^

app/classes/query/modules/google_search.rb
```diff
@@ -70,13 +70,13 @@ def google_conditions(search, field)
...
-    if str.sub!(/^-"([^""]+)"( |$)/, "") ||
+    if str.sub!(/^-"([^"]+)"( |$)/, "") ||
        str.sub!(/^-(\S+)( |$)/, "")
       bads << Regexp.last_match(1)
-    elsif str.sub!(/^(("[^""]+"|\S+)( OR ("[^""]+"|\S+))*)( |$)/, "")
+    elsif str.sub!(/^(("[^"]+"|\S+)( OR ("[^"]+"|\S+))*)( |$)/, "")
       str2 = Regexp.last_match(1)
       or_strs = []
-      while str2.sub!(/^"([^""]+)"( OR |$)/, "") ||
+      while str2.sub!(/^"([^"]+)"( OR |$)/, "") ||
```
JoeCohen, line 79: Autocorrected by Lint/DuplicateRegexpCharacterClassElement cop
https://docs.rubocop.org/rubocop/1.31/cops_lint.html#lintduplicateregexpcharacterclasselement

app/classes/query/modules/validation.rb
```diff
@@ -195,9 +195,7 @@ def validate_name(arg, val)
           @params_cache ||= {}
           @params_cache[arg] = val
           val.id
-        elsif val.is_a?(String)
-          val
-        elsif val.is_a?(Integer)
+        elsif val.is_a?(String) || val.is_a?(Integer)
```
JoeCohen, line 198: This is a typical correction of a DuplicateBranch offense.

app/models/name/taxonomy.rb
```diff
@@ -158,18 +158,18 @@ def all_children
...
+        if accepted.text_name.include?(" ")
+          genus_name = accepted.text_name.split(" ", 2).first
+          genera     = Name.with_correct_spelling.where(text_name: genus_name)
+          accepted   = genera.reject(&:deprecated)
+          genera     = accepted if accepted.any?
+          nonsensu   = genera.reject { |n| n.author.start_with?("sensu ") }
+          genera     = nonsensu if nonsensu.any?
+          genera.first
+        end
+      end
```
JoeCohen, line 172: Fixes Lint/NoReturnInBeginEndBlocks
https://docs.rubocop.org/rubocop/1.31/cops_lint.html#lintnoreturninbeginendblocks

app/classes/query/modules/google_search.rb
pellaea: The reason I (occasionally remember to) do the redundant double quotes is that it sometimes works better with syntax highlighters, especially when there is a typo somewhere which causes the parser to go off the rails.

A better example of the same thing (which might not offend rubocop) is to avoid unbalanced parentheses, brackets or braces, adding the closing paren even if unnecessary, or if that would break the expression, adding the closing paren in a comment.  In vim there is a command to jump between the opening and closing parens surrounding the cursor.  That breaks if you have a string literal with just one paren "(".

I take it the syntax highlighters you and Nimmo use will not be fooled by the unbalanced double quotes in some of these lines?  If so, then I have no qualms about this particular example, personally.  (I don't even like or use syntax highlighting!)

pellaea (approved):
Looks good to me!

app/classes/query/modules/google_search.rb
JoeCohen: @pellaea: Thanks for reviewing yet another PR so carefully.
Good point about syntax highlighting; I hadn't thought of that.
Fortunately, VSCode (which both @nimmolo and I use) isn't fooled by the changes. Or at least the changes don't make any difference in the display.

PR #1092, opened 2022-07-16
Add assertions to test_observations_of_name

JoeCohen:
Adds some assertions to ObservationsControllerTest#test_observations_of_name

### DIFF
--- a/test/controllers/observations_controller_test.rb
+++ b/test/controllers/observations_controller_test.rb
@@ -527,13 +527,21 @@ class ObservationsControllerTest < FunctionalTestCase
   end
 
   def test_observations_of_name
-    login
-    params = { species_list_id: species_lists(:unknown_species_list).id,
-               name: observations(:minimal_unknown_obs).name }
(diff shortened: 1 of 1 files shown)


JoeCohen:
This was the first step in something I've now put on hold. I thought it was worth merging independently.
(What I put on hold was: Add a `Review` link to the Observation index template. Clicking that link would create a new query and display its results -- Observations in set, where the members of the set are the result of the old query. I thought this would be a quick and dirty way to address Alan's need for a better way to ID unknowns. But it seems that Alan would like something more ambitious, and @nimmolo is in the mood to tackle it.)

pellaea (approved):
Useful test.  Your idea of turning a query into an "in_set" query to fix the order is clever!

nimmolo (approved):
I also like that `in_set` idea...

JoeCohen:
Thank you @pellaea & @nimmolo for the feedback. Maybe I'll take that Idea up next week.

PR #1093, opened 2022-07-16
Set up pending Naming cops

JoeCohen:
Configures Naming department cops which are pending in Rubocop 1.31.1

(There's just one cop and there are no current offenses, so it's just a matter of explicitly enabling it.)


JoeCohen:
I'm merging this. It's a trivial change only in Rubocop configuration.

PR #1094, opened 2022-07-17
Stop images_observations.csv from expanding every time the script is run

mo-nathan:
I noticed that the file /var/web/mo/public/images_observations.csv had become unreasonably large (41 million lines) and had a huge amount of duplication.  Traced it down to some recent changes in script/update_snapshots.  I believe this fix will now correctly create both observation_images.csv and images_observations.csv.  I'm assuming we want to maintain the original name (images_observations.csv) since folks might be downloading from that URL while also enabling the now more correctly named observation_images.csv.  This should also clean up the current mess once this fix gets deployed.

### DIFF
--- a/script/update_snapshots
+++ b/script/update_snapshots
@@ -6,10 +6,8 @@ source $(dirname $0)/bash_include
 config_file=$app_root/config/mysql-$RAILS_ENV.cnf
 public=$app_root/public
 
-echo "image_id	observation_id" > $public/observation_images.csv
-mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/observation_images.csv
-# Maintain a copy of the old file for backwards compatibility.
-cp $public/observation_images.csv $public/images_observations.csv
+echo "image_id	observation_id" > $public/images_observations.csv
+mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/images_observations.csv
 
 echo "id	name	north	south	east	west	high	low" > $public/locations.csv
 mysql --defaults-extra-file=$config_file -q -s -e 'select id, name, north, south, east, west, high, low from locations' >> $public/locations.csv
--- a/script/update_snapshots
+++ b/script/update_snapshots
@@ -7,7 +7,8 @@ config_file=$app_root/config/mysql-$RAILS_ENV.cnf
 public=$app_root/public
 
 echo "image_id	observation_id" > $public/observation_images.csv
-mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/images_observations.csv
+mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/observation_images.csv
+cp $public/observation_images.csv $public/images_observations.csv
 
 echo "id	name	north	south	east	west	high	low" > $public/locations.csv
 mysql --defaults-extra-file=$config_file -q -s -e 'select id, name, north, south, east, west, high, low from locations' >> $public/locations.csv
--- a/script/update_snapshots
+++ b/script/update_snapshots
@@ -8,6 +8,7 @@ public=$app_root/public
 
 echo "image_id	observation_id" > $public/observation_images.csv
 mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/observation_images.csv
+# Maintain a copy of the old file for backwards compatibility.
 cp $public/observation_images.csv $public/images_observations.csv
 
 echo "id	name	north	south	east	west	high	low" > $public/locations.csv


script/update_snapshots
```diff
@@ -7,7 +7,8 @@ config_file=$app_root/config/mysql-$RAILS_ENV.cnf
 public=$app_root/public
 
 echo "image_id	observation_id" > $public/observation_images.csv
-mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/images_observations.csv
+mysql --defaults-extra-file=$config_file -q -s -e 'select image_id, observation_id from observation_images' >> $public/observation_images.csv
+cp $public/observation_images.csv $public/images_observations.csv
```
pellaea, line 11: Huh???  Does "echo > file" not clobber the file???  Oh!!!  Nimmo must have changed one of the "images_observations" to "observation_images" but not both.  Can we simply change that back instead of doing the copy?  (Shouldn't it have been a mv -f instead of a cp, anyway?  Won't this code just keep observation_images.csv around and grow it ad infinitum?

I would like to see this file remain images_observations.csv.  The fact that we changed the table name internally need not affect the various users that use this file.

Summary: Please revert line 10, remove the copy, and fix line 9 to output the header to the correct file, images_observations.csv.

mo-nathan:
I updated images_observations.csv manually on the production server to save some space, but it will continue to grow until this change gets deployed.  I have a local copy of the big version if anyone cares to see it.  Initially I just gzipped the old one, but then figured I'd probably just forget about cleaning it up .

script/update_snapshots
pellaea: Okay, I've finally gotten my head fully around this.  (I think!)

The original bug was introduced when Nimmo changed one of the filenames from images_observations.csv to observation_images.csv, but not the other.  Nice and simple.  It's the particular fix that had me so confused.

Maybe you are deliberately "keeping the bug" because it creates a new file (observation_images.csv) separately that you can then move into place atomically?  If that is the reason wouldn't it be better to move than copy?  Does keeping the copy around serve some purpose I've not thought of?  And shouldn't we also do it for all the other files, not just images_observations.csv?  If that one file can benefit from being changed atomically, wouldn't they all?  In fact, maybe we should create the whole *suite* of files separately first, then move them all into place in one go.  They kind of need to be all the same date.  For that matter, if we're really being strict, wouldn't we want to lock the database to prevent any new writes while we dump all the tables?  How careful do we need to be?  And lastly, wouldn't the intent be clearer if we called the new file "images_observations.csv.new" instead of "observation_images.csv"?  That's such a random and/or potentially confusing name in this context!

mo-nathan: So my thinking was let's first create observation_images.csv correctly (since it appears to be the shiny new name), but then maintain backwards compatibility by immediately copying it to the old name.  This would also clean up both files that are currently getting created incorrectly.  If we're at all worried about the disk space or the time it takes to do the copy, then we could use a hard (or a symbolic) link instead of a making a copy.  However, at the end of the day I don't care as long as it doesn't just keep growing.

I was definitely not thinking about atomicity and I don't think we need to.  Looks like the worst possible case would be an observation getting created after the join table gets dumped which doesn't seem like a bit deal since any associated images would just appear in the next dump.

nimmolo: I approve of the superior analysis going on here. Sorry to introduce the bug, i don't think I would have intentionally changed a filename output like that, but i should have caught it.

I do like the idea of having it available under the new name of the table, and the old one for compatibility. Who uses these csv's, by the way?

pellaea:
Ah!!!  Thank you for taking the time to explain the reasoning, Nathan.

I have a mild preference for trying to separate the external view from the
internal structure.  I don't like the idea that we essentially need to make
a new copy every time we change the internal structure.  Of course, in
practice, such changes happen exceedingly rarely, so who cares!

(As an aside, I'm curious about your general thoughts on this basic
pattern.  My instinct is to make the external view identical to the
internal structure in the first pass -- it just makes everything easy the
first time around.  But the general pattern seems to be that superficial
similarity leads subsequent workers to conclude erroneously that the
external view *must* be identical to the internal structure -- in effect a
violation of separation of concerns.  The question is: would it be better
to make the first version deliberately and conspicuously (and yes,
artificially) different so that it is 100% clear that it is separate from
the internal structure?  I think if I had done so, there would have been no
cause for this conversation to come up.  No one would ever have considered
changing the filename, or as it seems in this case, it would not have
accidentally gotten changed in an app-wide global search and replace.
However careful Nimmo might have been, such mistakes are bound to happen.
And they're damned hard to catch -- I approved the PR, I'm equally "to
blame"!  So I could argue that the superficial similarity of public to
private is actually dangerous from this point of view.  Thoughts?)

I'm totally fine with maintaining two copies, by the way, it almost goes
without saying.  Although, it's confusing enough (to me at least) that it
may be worth a quick comment, which I will happily add myself in a minute.

re: who uses these? -- There's no way to tell.  But I've pointed countless
users at them, so they may be used by quite a large number of people.
These csv files are *by far* the easiest way for researchers to mine our
database.  *Way* easier and less burden on our server than using the API or
scraping the web interface.  And easier than me providing them a custom
dump.  And it stays up to date nightly to boot.  It is missing a few
things, admittedly.  And every so often I will add an extra column or two
to one of the files to handle new requests.  But it's been a long time
since I've had to do that.

mo-nathan:
Regarding the "external view" vs. the "internal view" discussion, I don't think there is a simple answer.  It's essentially a form of the notoriously hard naming problem.  Different languages have taken different approaches (Old School Java being at one extreme where everything has to be declared separately even if they have the same name).  However, the general conclusion (even in current day Java now) is to keep it simple and let the obvious thing be the easiest to implement.  If this is a strictly internal change (which was clear to me after a casual glance), then I'd be in favor of just maintaining the old name.  I just made it work correctly with minimal effective change in case someone did care.

mo-nathan:
I got rid of the new file name.  I don't really see any reason to support that name as well given that the old name has been shared a bunch and who knows who is using it at this point.

pellaea:
Thanks, Nathan.  That is my preference for whatever it's worth.  Thanks
also for your insight re: internal vs. external naming conventions, even if
there is no simple answer.

PR #1101, opened 2022-07-18
Fix some routes

pellaea:
I noticed a profusion of "no routes exist" errors in the log file after deploying Nimmo's observer controller refactor.  I tried to fix them all.

Also there was no route for POSTing /observations/download ... probably should write a test for that one!

Lastly, I accidentally uncovered an old bug in query pagination where it was literally loading and instantiating *every* observation in certain cases.  (Click on an observation, then click on "Index" between "« Prev" and "Next »".)  I simplified and cleaned up that code to ensure nothing like that happens again.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -378,9 +378,10 @@ Style/DateTime:
   Enabled: false
 
 # This doesn't apply in routes.rb because Rails routes don't accept annotated
-# tokens (at least yet).
+# tokens (at least yet).  Also, enabling this produce a ton of new errors.
+# Let's talk about this before we enable it, I guess.
 Style/FormatStringToken:
-  Enabled: true
+  Enabled: false
   Exclude:
     - "config/routes.rb"
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -377,6 +377,13 @@ Style/AsciiComments:
 Style/DateTime:
   Enabled: false
 
+# This doesn't apply in routes.rb because Rails routes don't accept annotated
+# tokens (at least yet).
+Style/FormatStringToken:
+  Enabled: true
+  Exclude:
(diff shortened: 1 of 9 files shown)


nimmolo:
@pellaea What does the `:id` param do in an index? If i knew i've forgotten, it never made sense to me.

pellaea:
The workflow -- if anyone actually uses it! -- is:

1. do a search or index
2. click on a result
3. go to next or prev as much as you like
4. return to the index at any point by clicking in "Index" (between « Prev
and Next »)

This takes you to the page of results in the index that has that last
result in it.

Someone must be using it because there were quite a number of requests
(possibly all from robots, though).

.rubocop.yml
```diff
@@ -377,6 +377,14 @@ Style/AsciiComments:
 Style/DateTime:
   Enabled: false
 
+# This doesn't apply in routes.rb because Rails routes don't accept annotated
+# tokens (at least yet).  Also, enabling this produce a ton of new errors.
+# Let's talk about this before we enable it, I guess.
+Style/FormatStringToken:
+  Enabled: false
+  Exclude:
+    - "config/routes.rb"
```
JoeCohen, line 386: @pellaea Re configuration of `Style/FormatStringToken`
It shouldn't be producing **new** errors. I'll try to look at this tomorrow.

pellaea:
Yeah, that was weird.  Thanks for taking a look at it.  In particular it
was going crazy in the scripts directory.

JoeCohen:
I looked at this, and cannot figure out what's happening. I think it's a
Rubocop bug.
I therefore left the cop disabled.
An alternative would be fixing the 38 offenses and enabling the cop.
For more details, see
https://www.pivotaltracker.com/story/show/182759112/comments/232331567

PR #1100, opened 2022-07-18
pending style cops 182027741

JoeCohen:
This PR sets up Style department cops that are pending as of Rubocop 1.31.1:
- Configures the cops (explicitly enables/disables each of them in `.rubocop.yml`
- Fixes offenses or disables them inline in the code

This PR delivers part of https://www.pivotaltracker.com/story/show/182027741

PR #1102, opened 2022-07-19
Bisect cop failure

JoeCohen:
DO NOT MERGE
The purpose of this PR is to isolate the source of a CI test failure in the pull_request check of #1100.
My plan is 
- Bisect the commits of that PR to find the offending commit, then
- Bisect, divide and/or analyzing the offending commit to further isolate the problem.

I need an actual to replicate the failure because the test failure occurs only the in pull_request check (and not in the push check or in local tests).


JoeCohen:
Closing this, deleting the branches as it appears I can work on #1100 locally.

PR #1104, opened 2022-07-19
Improve test_propagate_<xxx> assertions

JoeCohen:
This PR changes assertion in these tests to assert that bulk updates leave `updated_at` unchanged (instead of asserting that `updated_at` is earlier than a minute before now).
This fixes a latent bug that can cause intermittent test failures if one adds a fixture that lacks an explicit `updated_at`.

Delivers https://www.pivotaltracker.com/story/show/182759067.

### DIFF
--- a/test/models/cache_test.rb
+++ b/test/models/cache_test.rb
@@ -78,15 +78,24 @@ class CacheTest < UnitTestCase
   # changed in all of a genus's subtaxa (and that updated_at not changed).
   def test_propagate_classification
     name = names(:agaricus)
+    saved_obs_updated_ats =
+      Observation.where(Observation[:text_name].matches("Agaricus%")).
+      pluck(:updated_at)
     new_classification = names(:peltigera).classification
+
     name.update(classification: new_classification)
     name.propagate_classification
+
     Observation.where(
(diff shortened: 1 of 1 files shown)


test/models/cache_test.rb
```diff
@@ -78,47 +78,73 @@ def test_changing_observation_name
...
       Observation[:text_name].matches("Agaricus%")
     ).each do |obs|
       assert_equal(new_classification, obs.classification)
-      assert_operator(obs.updated_at, :<, 1.minute.ago)
     end
+    assert_equal(
+      saved_obs_updated_ats,
+      Observation.where(Observation[:text_name].matches("Agaricus%")).
+        map(&:updated_at)
+    )
```
pellaea, line 98: Oh wow, that's actually really clean.  I was scared to try this approach because I couldn't figure out how to do so easily.  This works nicely.

pellaea (approved):
Strongly approve.  Thanks, Joe!

JoeCohen:
@pellaea: Thanks for the review and the kind words. 
@nimmolo's improvements (Arel and the Name scopes) were key in keeping this short and readable. 
I'll merge it.

PR #1105, opened 2022-07-20
Pending layout cops

JoeCohen:
Sets up the pending cops in the Layout department:
- Explicitly enables/disables them;
- Fixes any offenses

Delivers the final part of https://www.pivotaltracker.com/story/show/182027741


JoeCohen:
This is small and simple enough so that I will just merge it.

PR #1106, opened 2022-07-20
Turn Style/FormatStringToken cop again

pellaea:
I changed the configuration to ignore the redirect method instead of skipping the entire config/routes.rb file.

I cannot reproduce the offenses that it was finding in the scripts directory earlier.

We have now officially gone 360° and returned almost precisely to where we were before I turned this silly cop off.  I'm so confused.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -377,13 +377,11 @@ Style/AsciiComments:
 Style/DateTime:
   Enabled: false
 
-# This doesn't apply in routes.rb because Rails routes don't accept annotated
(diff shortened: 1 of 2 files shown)


JoeCohen:
I think Codeclimate will tell you only about "new" offenses.

pellaea:
But I thought the point was that it was complaining about the scripts even
though they were not new?  Well, whatever, it is turned back on and it is
not complaining about any old offenses and I think that makes everyone
happy, so I'm going to merge this silly thing.

JoeCohen:
Sounds good.

PR #1108, opened 2022-07-21
Remove merge_source_id code

JoeCohen:
Removes all code related to unused attributes `LocationDescription.merge_source_id` and  `NameDescription.merge_source_id`
(Does not  remove the columns themselves. That will be handled by a later migration.)

Delivers the code portion of https://www.pivotaltracker.com/story/show/182778592

### DIFF
--- a/app/controllers/location_controller.rb
+++ b/app/controllers/location_controller.rb
@@ -441,23 +441,7 @@ class LocationController < ApplicationController
     return unless @description
 
     @location = @description.location
-    if params[:merge_source_id].blank?
-      @description.revert_to(params[:version].to_i)
-    else
-      @merge_source_id = params[:merge_source_id]
-      version = LocationDescription::Version.find(@merge_source_id)
-      @old_parent_id = version.location_description_id
-      subversion = params[:version]
-      if subversion.present? &&
-         (version.version != subversion.to_i)
-        version = LocationDescription::Version.
-                  find_by(
-                    version: params[:version],
-                    location_description_id: @old_parent_id
-                  )
-      end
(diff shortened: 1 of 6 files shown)


JoeCohen:
Bonuses:
- Fixes a Brakeman SQL Injection warning. (That's what got me started down this path.);
- Fixes 3 ABC offenses; 
- Improves coverage, as the removed code was largely untested.

app/controllers/location_controller.rb
```diff
@@ -800,7 +784,6 @@ def edit_location_description
         if (params[:delete_after] == "true") &&
            (old_desc = LocationDescription.safe_find(params[:old_desc_id]))
           v = @description.versions.latest
-          v.merge_source_id = old_desc.versions.latest.id
           v.save
```
pellaea, line 787: It doesn't look like "v" is used anywhere in this method any more.  I think you can get rid of "v = @description.versions.latest" and "v.save", too.

app/controllers/name_controller.rb
```diff
@@ -583,7 +568,6 @@ def edit_name_description
         if (params[:delete_after] == "true") &&
            (old_desc = NameDescription.safe_find(params[:old_desc_id]))
           v = @description.versions.latest
-          v.merge_source_id = old_desc.versions.latest.id
           v.save
```
pellaea, line 571: Probably ditto.

pellaea (approved):
Looks good!

JoeCohen:
Thank you! 
I'll get rid of the lines you pointed out and merge this.

PR #1111, opened 2022-07-22
Return symbols, not strings, for `AbstractModel#show_controller`, `show_action`, `index_action`, `destroy_action` etc.

nimmolo:
To make the `HerbariaController` refactor work, it seems @JoeCohen discovered that [Rails automatically prepends the controller namespace to the controller name](http://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing), if the caller of a route builder is in a namespaced controller like `Herbaria::MergesController`. However, this can unfortunately lead Rails to build inaccurate controller names like `herbaria/herbaria`, if the caller of the route builder is trying to build a route to an **action in the main controller** (i.e., no namespace needed). 

However, [if you send the controller name as a **string** prepended by a leading forward slash](https://stackoverflow.com/questions/20057910/rails-url-for-behaving-differently-when-using-namespace-based-on-current-cont), though, Rails will **not** prepend the namespace, and you'll get `controller: "/herbaria"` which works as desired.

This has some odd and unpredictable effects elsewhere, though, as the `Model.show_controller` is used for other purposes in the code. The forward slash has to be accounted for, even when this is not as expected. (I encountered these problems on the `NamesController`/`NameDescriptionsController` refactor and decided to try to regularize it upstream.)

Generally I've been switching to symbols everywhere in the app for url builders with a controller/action hash `{ controller: :ctrlr, action: :action }`, and while Rails will accept anything, including an IOU, I'd rather be able to count on getting a controller symbol when I call for the show_controller, and an action symbol for the index_action. (If we need to call a namespaced controller, we can always explicitly call it.)

So here I've i worked out a way to return symbols, eg `:location`, `:observations` for the show_controller, `:show` (or `:show_location`) for the show action, etc, while still fixing the issue of Rails erroneously prepending a controller namespace, closer to the source, in Application Controller. There only seems to be one place it's needed: the source of the issue seems to be Rails `redirect_to`, and this PR prepends the forward slash to any controller name just prior to calling that method. Meanwhile, `Model.show_controller` now always returns the symbol for the controller.


pellaea (approved):
I think this might be good.  My memory from earlier -- when we had to add the leading slashes because of a rails upgrade or somesuch -- was that this is not terribly well-tested and it should therefore be at least somewhat tested by hand in the dev VM before pushing it out.

JoeCohen (changes requested):
@nimmolo I finally found the time, and slept enough, to focus on this PR.
I have no substantive suggestions. The comments should be revised to reflect the changes you made to the code.
I can do that if you like; just let me know.

app/models/abstract_model.rb
```diff
@@ -413,9 +415,9 @@ def show_action
   # Note that show_controller has a leading forward slash
```
JoeCohen, line 415: Revise documentation (`show_controller` is now a symbol without a leading forward slash)

JoeCohen (approved):

JoeCohen:
I:
- pushed my changes to the comments
- pushed the deletion of Observation's now-unnecessary override of `show_controller`.
- merged master into this branch and resolved conflicts.

As @pellaea already approved this branch, I intend to merge it.

nimmolo:
Sounds good. Thanks, Joe.

JoeCohen:
It's now running on the server

nimmolo:
Hi Joe, 
Thanks for checking it out. I agree with moving the comments, i'd be glad for you to handle it if you can.

PR #1114, opened 2022-07-24
Enable HABTM cop

JoeCohen:
This PR:
- Enables Rubocop's Rails/HasAndBelongsToMany cop.
(It does that by removing the explicit `Enabled: false`, instead relying on the Rails default of `Enabled: true`.
The cop had been because we had many HABTM associations. 
But they've now all become HMT, thanks to @nimmolo .)
- Delivers https://www.pivotaltracker.com/story/show/182783737


JoeCohen:
I will merge this. It's a trivial PR and I'd like to get it out of the way.

PR #1115, opened 2022-07-24
Fix update snapshots for ml

mo-nathan:
Doh!


pellaea (commented):
Mo' better.

JoeCohen (approved):
I can see that this is linux and/or sql, and I can sort of see what it's doing, and the changes make sense. Sorry to be unable to offer any useful comments.

PR #1117, opened 2022-08-02
Rails/InverseOf

JoeCohen:
This PR fixes all current [Rubocop Rails/InverseOf](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsinverseof) offenses.
In many cases, the cause of the offense was an explicit, redundant `foreign_key`. 
(Rails does not automatically compute an inverse of associations that specify a `foreign_key`. 
See https://guides.rubyonrails.org/v6.1/association_basics.html#bi-directional-associations.)
In those cases, I fixed the InverseOf offense by removing the redundant `foreign key`.
Thus this PR incidentally fixes all [Rails/RedundantForeignKey](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsredundantforeignkey) offenses.


pellaea (approved):
Looks fine to me.  Thanks, Joe.

JoeCohen:
Thanks Jason for taking the time to look and comment.

PR #1118, opened 2022-08-03
Rubocop safe autocorrect

JoeCohen:
This PR fixes a few offenses by:
- running `rubocop -a --ignore-parent-exclusion`
- inspecting the results, and 
- making necessary tweaks.


JoeCohen:
Right now I only have energy for small/simple PRs. But at least it's something.

PR #1124, opened 2022-08-16
Normalize Sequences Controller

JoeCohen:
This PR changes `SequenceController`, and associated files, routes, actions, and views to the Rails defaults, including
- pluralizes "Sequence";
- drops "sequence" prefix from file and method names;
- `index` (rather than `list`) as the only index action,
combining legacy indexing actions into "flavors" of normalized `index`;
- passes Observation id to `new` and `create` as a query param, rather than a route param, 
i.e. ~~`/sequence/create_sequence/123`~~ => `sequences/new?obs_id=123`,
changing routes accordingly;
- uses `form_with` (not `form_for`), extracting form builder calls to the form file;
- changes controller test to exclusively "request" tests, i.e., one request per method;
- `destroy` UI is a button, rather than a link;
- a variety of other improvements and fixes.

### Suggested Manual Test
Manual test
- Goto any Observation
- Try to create a Sequence with a Locus (any text will do) but without other attributes.
Expected result: Flash error, returned to new Sequence form
- Add the bases (any combination of ACGT, spaces, returns).
Expected result: Sequence created, returned to Observation
- Edit the Sequence (whatever you like).
Expected result: Update successful, returned to Sequence
- Destroy Sequence (in Dev Mode this must be confirmed 2x).
Expected result: Sequence destroyed, returned to Observation
- http://localhost:3000/sequences
Expected result: complete Sequence index


JoeCohen:
@pellaea: Should the sitemap include Sequences?

test/integration/sequencer_test.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require("test_helper")
+# require("capybara_helper")
+
+# Test adding, editing, and deleting a Sequence
+# Uses Capybara; MO SessionFormExtensions don't work on form_with forms
+# Use an integration test because click_mo_link needs an id, and
+# there's no  id (at least by default) for a form_with resource-oriented
+# form field.
```
JoeCohen, line 10: @pellaea:
### Should I try to modify SessionFormExtensions so that `click_mo_link` etc.  work on `form_with` forms?
(That probably belongs in another PR.)

app/controllers/sequences_controller.rb
```diff
@@ -0,0 +1,230 @@
...
+# destroy_sequence (delete)         destroy (delete)
+# edit_sequence (get)               edit (get)
+# *edit_sequence (post)             update (patch)
+# index_sequence (get)              index (get) -- lists query results
+# list_sequences (get)              index (get, flavor: all) -- all Sequences
+# *next_sequence (get)              show { flow: :next } (get))
+# *prev_sequence (get)              show { flow: :prev } (get)
+# observation_index (get)           index (get, flavor: observation)
+#                                   -- list Sequences for one Observation
+# sequence_search (get)             index (get, pattern: present)
```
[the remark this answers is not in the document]
pellaea: Yeah, I kinda agree with you that MO really isn't well positioned to replace BLAST or other features of GenBank.  We don't even provide a way to download a FASTA or anything.  Not sure what to think about this.  I guess we have a recent trend toward simplifying the code and removing old unused features.  Maybe that applies here?

app/views/sequences/new.html.erb
```diff
@@ -10,6 +10,9 @@
...
 
-<%= form_for(:sequence, url: add_query_param(action: :create_sequence)) do |form| %>
-  <%= render(partial: "form_sequence", locals: { form: form }) %>
-<% end %>
+<%= render(
+  partial: "form",
+  locals: {
+    url: add_query_param(sequences_path(params: { obs_id: @observation.id }))
+  }
+) %>
```
pellaea, line 18: I agree.  This makes more sense than having the form outside the partial.  That was a bit weird.

config/routes.rb
```diff
@@ -817,6 +808,40 @@ def route_actions_hash
   get("/observer/show_rss_log/:id", to: redirect("/activity_logs/%{id}"))
   get("/observer/rss", to: redirect("/activity_logs/rss"))
 
+  # ----- Sequences: legacy action redirects
+  redirect_legacy_actions(
+    old_controller: "sequence", new_controller: "sequences",
+    # actions: [:show, :list, :index]
+    actions: [:show, :index]
+  )
```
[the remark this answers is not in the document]
pellaea: Nah.  You'd be surprised.  They will actually continue to show up for a while.  Maybe from cached pages?  Not sure.  I saw it with observer controller.  But I don't think it matters.

config/routes.rb
```diff
@@ -817,6 +808,40 @@ def route_actions_hash
...
+  # ----- Sequences: nonstandard legacy action redirects
+  # get("/sequence/index", to: redirect("/sequences"))
+  # get("/sequence/index_sequence", to: redirect("/sequences"))
```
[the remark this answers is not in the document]
pellaea: Same as the comment at the top about whether to keep unused features or not.  Probably not.

config/routes.rb
```diff
@@ -817,6 +808,40 @@ def route_actions_hash
...
+      # https://api.rubyonrails.org/classes/ActionDispatch/Routing/Redirection.html#method-i-redirect
+      to: redirect { |_params, request| # rubocop:disable Style/BlockDelimiters
+            # Does original request have params?
+            URI.parse(request.original_url).query ? # rubocop:disable Style/MultilineTernaryOperator
+              # preserve original pattern param (and any others)
+              "/sequences?#{URI.parse(request.original_url).query}"
+            :
+              "/sequences"
+          })
+
```
pellaea, line 844: Yikes!  Is there any better way to do this?  Except for the specialized /id parameter, don't all the ?params come along for the ride automatically with redirects?  I do remember having to explicitly route /observer/blah/id and /observer/blah separately, but I don't remember seeing all this complex parsing in the legacy observer controller actions.  Sorry, I should look into this more myself, but maybe you've already been there and done that?

test/integration/sequencer_test.rb
pellaea: Shouldn't we be migrating away from our custom-rolled integration helpers to Capybara and (now) normal integration tests anyway?  I think I updated click_mo_link a few weeks/months ago because it was the quickest way to get some big PR of Nimmo's working.  Not because I want to be maintaining out-dated test helpers!

nimmolo:
Joe - 
Capybara has its own click helpers - maybe check docs for those. Or - how about just giving the form an ID? Speaking from the ignorance of a phone of course.

app/controllers/sequences_controller.rb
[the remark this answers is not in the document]
JoeCohen: @pellaea: I greatly appreciate your taking time to look through the PR and comment.
As to this particular action/method, I'll delete it, and the associated route, redirect route, and tests.
As you point out it simplifies the code. And it prunes the routes. And it's easy enough to reinstate if there's a need.

config/routes.rb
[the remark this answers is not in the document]
JoeCohen: Interesting. I'll add redirects for them.

test/integration/sequencer_test.rb
JoeCohen: I agree.
(Even though I think there are downsides:
- The home-rolled helpers feel faster than Capybara.
- The home-rolled helpers sometimes do some extra testing.)

config/routes.rb
JoeCohen: Bottom lines: 
1. It doesn't matter here as this redirect route will be trashed. See discussion above https://github.com/MushroomObserver/mushroom-observer/pull/1124#discussion_r947067090
2. I think you're right for all practical purposes. And I was confused
For instance, the legacy URL
```http
https://mushroomobserver.org/observer/pattern_search?pattern=date%3A2015-11-20+user%3A%22kundabungkid%22+region%3AAustralia
```
lands on the right page (with a weird flash message `Search is invalid: ‘nil’`), with the address bar showing
```http
https://mushroomobserver.org/observations?pattern=date%3A2015-11-20+user%3A%22kundabungkid%22+region%3AAustralia
```
I got confused by web posts like
this: https://relativkreativ.at/articles/how-to-preserve-the-query-string-when-redirecting-routes-in-a-rails-application
and this: https://www.viget.com/articles/merging-query-strings-when-redirecting-in-rails/
and this: https://stackoverflow.com/questions/19549203/rails-routes-redirect-for-query-string

JoeCohen:
@nimmolo: Thanks for the suggestions. To clarify, 
I am using -- as you suggest -- the Capybara `click` and `fill_in` methods which work with id's, names, labels, etc.
I did think about adding id's. It's a good idea, but I eventually decided to keep the form as standardized and simple as possible.

JoeCohen:
I think this is now ready to merge. 
I'll let it sit for a day or so in case there are more comments.

nimmolo:
@JoeCohen — I wanted to circle back to this. 

As i'm dealing more with integration tests on the Name/NameDescription PR, trying to select links and forms with these CRUD routes, i'm belatedly noticing, thanks to @pellaea pointing out the obvious, that a link URL pattern is often not enough to select the right link. 

    # names/:name_id/descriptions/:description_id = show
    # names/:name_id/descriptions/:description_id = also, destroy!
    # names/:name_id/descriptions/new = create
    # names/:name_id/descriptions/:description_id/edit = edit

I forgot which form we were talking about in your sequences integration test here, but going through the form templates, i'm realizing that a rule of thumb I follow elsewhere in my life as a developer is probably applicable here: it's best if test-able elements are identifiable by HTML attribute. HTML attributes are cheap, and you never want to be without a handle for a door. (Or is it a key for a lock? or a handle for a window-sash? Or a long blade of grass for an ant-hill?)

For example, a form element should probably always get an ID, for the sake of identifying submit events. 

Likewise, on the NamesController PR, I think I'm going to start giving links for "create", "edit", "destroy" etc each an identifying class, like "destroy-name-description-link". On some pages, there's more than one "create" "destroy" or "edit" link, so identifying a link by label is not enough, and the class needs to be specific enough to grab the right link.

pellaea:
FWIW, I agree with Nimmo on this.  I think it's a good idea to give easy id
"handles" for elements we might want to test, including, definitely,
forms.  Why not links and buttons, too?  It costs nothing.

PR #1125, opened 2022-08-18
Cover ImageController

JoeCohen:
This PR increases coverage of ImageController.
The goal is to increase it to 100%, preferably all via ImageControllerTest
(The purpose is to prepare for refactoring of ImageController, and make any such refactoring more reliable.)


app/controllers/image_controller.rb
```diff
@@ -121,7 +126,7 @@ def advanced_search
     show_selected_images(query)
   rescue StandardError => e
     flash_error(e.to_s) if e.present?
-    redirect_to(controller: :observations, action: :advanced_search)
+    redirect_to(search_advanced_path)
```
JoeCohen, line 129: This is a bug fix.
(I think it originally was
```ruby
controller: :observation, action: :advanced_search
```
Then in the ObservationController refactor, got changed via search-and-replace to:
```ruby
controller: :observations, action: :advanced_search
```
and then was overlooked because there was no relevant test assertion.

app/controllers/image_controller.rb
```diff
@@ -272,7 +277,7 @@ def cast_vote
     if params[:next]
       redirect_to_next_object(:next, Image, params[:id].to_s)
     else
-      redirect_with_query(action: "show_image", id: id)
+      redirect_with_query(action: "show_image", id: params[:id])
```
JoeCohen, line 280: Bug fix. The local variable `id` is undefined.

app/controllers/image_controller.rb
```diff
@@ -714,7 +742,7 @@ def transform_image
       end
     end
     if params[:size].blank? ||
-       params[:size].to_sym == (@user ? @user.image_size : :medium)
+       params[:size].to_sym == (@user ? @user.image_size.to_sym : :medium)
```
JoeCohen, line 745: Bug fix. Rails native enums return strings, not symbols, so
we have to compare `params[:size].to_sym` to a symbol.
(This was probably OK when we used the simple_enum gem.)

test/controllers/image_controller_test.rb
```diff
@@ -485,6 +640,82 @@ def test_update_image
...
+        "when(2i)" => "5",
+        "when(3i)" => "12",
+        "copyright_holder" => "Rolf Singer",
+        "notes" => "",
+        "original_name" => "new name"
+      }
+    }
+
+    login(image.user.login)
+    Image.any_instance.stubs(:save).returns(false)
```
JoeCohen, line 711: @pellaea: Is there an image (or pseudo-image) file which will make `@image.save` fail?
(I'd prefer to avoid stubbing Image#save if that can be done easily.)

test/controllers/image_controller_test.rb
```diff
@@ -693,6 +997,30 @@ def test_add_images_strip_gps
     assert_true(img.gps_stripped)
   end
 
+  def test_add_images_process_image_fail
+    login("rolf")
+    obs = observations(:coprinus_comatus_obs)
+    setup_image_dirs
+    fixture = "#{MO.root}/test/images/geotagged.jpg"
+    fixture = Rack::Test::UploadedFile.new(fixture, "image/jpeg")
+    Image.any_instance.stubs(:process_image).returns(false)
```
JoeCohen, line 1006: @pellaea: Is there an image (or pseudo-image) file that will make `Image#process_image` fail?

test/fixtures/users.yml
```diff
@@ -25,6 +25,7 @@ rolf:
   name: Rolf Singer
   license: ccwiki30
   image: rolf_profile_image
+  image_size: <%= User.image_sizes[:medium] %>
```
JoeCohen, line 28: Repeats the default in order to make it less likely that a coder will change the value.

test/controllers/image_controller_test.rb
pellaea: Sorry, no, I think the only way to get Image.save to fail is to stub something.  I suppose you could also stub validate_upload?  Or maybe stub upload_length to be huge or something like that?  You might also try uploading a plain old text file or something like that?

JoeCohen:
Jason:
Thanks for the feedback.
I actually did try a text file. I forget why it didn't work; probably there
was an error before `image.save` got called.
I'll try again.
Joe

JoeCohen:
@nimmolo: TLDR is `ImageCpontroller#remove_image` (singular) dead code? If not, where is use?

Everything in ImageController is now covered from ImageControllerTest
**except `remove_image`** which is quoted below.
I can't find any call to `ImageController#remove_image` (which is not an action), 
even though the comment preceding the method says that it's linked from `observations\edit`.
(The comment took its present form in PR #1060.)

Of course, my old eyes and brain could easily be missing something subtle (or obvious).
_____

```ruby
  # Callback to remove a single image from an observation.
  # Linked from: observations/edit
  # Inputs: params[:image_id], params[:observation_id]
  # Redirects to observations/show.
  def remove_image
    pass_query_params
    @image = find_or_goto_index(Image, params[:image_id])
    return unless @image

    @observation = find_or_goto_index(Observation, params[:observation_id])
    return unless @observation

    if !check_permission!(@observation)
      flash_error(:runtime_image_remove_denied.t(id: @image.id))
    elsif @observation.images.exclude?(@image)
      flash_error(:runtime_image_remove_missing.t(id: @image.id))
    else
      @observation.remove_image(@image)
      @image.log_remove_from(@observation)
      flash_notice(:runtime_image_remove_success.t(id: @image.id))
    end
    redirect_with_query(controller: :observations,
                        action: :show, id: @observation.id)
  end
```

JoeCohen:
@pellaea: Any thoughts on the following?

> @nimmolo: TLDR is `ImageCpontroller#remove_image` (singular) dead code? If not, where is use?
> 
> Everything in ImageController is now covered from ImageControllerTest **except `remove_image`** which is quoted below. I can't find any call to `ImageController#remove_image` (which is not an action), even though the comment preceding the method says that it's linked from `observations\edit`. (The comment took its present form in PR #1060.)
> 
> Of course, my old eyes and brain could easily be missing something subtle (or obvious).
> 
> ```ruby
>   # Callback to remove a single image from an observation.
>   # Linked from: observations/edit
>   # Inputs: params[:image_id], params[:observation_id]
>   # Redirects to observations/show.
>   def remove_image
>     pass_query_params
>     @image = find_or_goto_index(Image, params[:image_id])
>     return unless @image
> 
>     @observation = find_or_goto_index(Observation, params[:observation_id])
>     return unless @observation
> 
>     if !check_permission!(@observation)
>       flash_error(:runtime_image_remove_denied.t(id: @image.id))
>     elsif @observation.images.exclude?(@image)
>       flash_error(:runtime_image_remove_missing.t(id: @image.id))
>     else
>       @observation.remove_image(@image)
>       @image.log_remove_from(@observation)
>       flash_notice(:runtime_image_remove_success.t(id: @image.id))
>     end
>     redirect_with_query(controller: :observations,
>                         action: :show, id: @observation.id)
>   end
> ```

pellaea:
Agreed.  This looks dead to me, as well.  It should be handled through
remove_images instead.  I can confirm that the edit_observation form does
not, in fact, provide a remove_image link.  Maybe the old one did?

nimmolo:
@JoeCohen — Sorry I missed this review request! I just tucked my head down and started ploughing my row, didn't I?

I don't know much about ImageController. Maybe I wrote this action in the runup to Bootstrap 4? 
It does seem like it would be useful to be able to remove an image from an observation, but if this action ain't being used, i'd say it's good to delete. Anyway, if we need it in the future, here it is. Thanks for dealing with it.

PR #1128, opened 2022-08-23
Convert `get_with_dump` to `get`, use named params hash everywhere

nimmolo:
When we updated to Rails 6/Ruby 2.7, and then on to Ruby 3.0, if I recall correctly, there was a new requirement that when a hash is the last argument in any method call, it must be **named**. No more automatic conversion of keyword arguments. So in the case of `FunctionalTestCase` methods `get`, `post`,  `put`, and `patch` which modify the `super` methods with the same names, it became necessary to change them to send a named `params: params` hash. 

In other words, no more `get(:action, params)`; it seems it has to be `get(:action, params: params)`.

> _(Note: I _believe_ this change was necessary, but I am not 100% sure. There is a chance I misunderstood the requirement, but I seem to remember trying various less intrusive changes to the methods in functional test case, and I couldn't get any of them to work. With the changes I did implement, I also had to change quite a few test method calls to send that named `params` hash.)_
> Check [this discussion](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) of the change to Ruby.
> _If anyone feels that my changes to FunctionalTestCase made the test code unnecessarily verbose, or is aware that I missed something obvious about Ruby, please speak up now! Even if a few months ago might have been more convenient. :)_

So. This PR extends the same expectation/behavior to MO's other methods descended from `get` et al: the `ControllerExtensions` methods `get_with_dump`, `post_with_dump`... It brings their syntax in line with the format used in functional test case's `get` et al. All methods now expect a hash named `params`, and the method splats the hash. 

If this doesn't make sense, just check the code in the PR, it should be obvious at a glance. **New positive outcome**: test calls to `get_with_dump` now have the same syntax as calls to `get`, as you will notice from the changes in the PR. There may be other test methods that will need a syntax change, like `requires_login` maybe, but I believe not. 

Reason for this PR: `get_with_dump` was throwing a lot of errors when it was called with the extra "parent_id"  parameter that is necessary for route nesting on the `NamesController` / `NameDescriptionsController` branch. 

`names/:name_id/descriptions/:desc_id`

This issue with `get_with_dump` seemed to have a separate scope from the refactoring in that controller, so I wanted to tackle it separately.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -21,12 +21,6 @@ class AdminController < ApplicationController
     end
   end
 
-  # Simple list of all the files in public/html that are linked to the W3C
-  # validator to make testing easy.
-  def w3c_tests
-    render(layout: false)
-  end
-
   # Update banner across all translations.
   def change_banner
     if !in_admin_mode?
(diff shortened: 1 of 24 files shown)


JoeCohen (changes requested):
1. I agree that changes to `<action>_with_dump` should be in their own PR, 
and that this is the way to do that.
2. But I think you should instead get rid of the various `<action>_with_dump` as well as `Admin#w3c_tests`. See https://groups.google.com/forum/#!topic/mo-developers/Qi8N2TVSp6c. 
(I had created a [Pivotal story](https://www.pivotaltracker.com/story/show/173424826) to do that, but never followed through.  Sorry.
If you like, you can forget about this PR, and get rid of `<action>_with_dump` in another PR today.)
3. If retaining `<action>_with_dump` as the easiest short-term way to move forward with the NamesController PR, I suggest doing a global search for `/\w+_with_dump/`. There are some calls to `post_with_dump` plus one to `send(:get_with_dump` that might need fixing.

nimmolo:
@JoeCohen — 🎖 that's some amazing recall! 

Okay - should I 
1. Delete all these assertions that `get_with_dump` `post_with_dump` etc?
2. Or just switch them to use `get`, `post`, etc?

JoeCohen:
I'm not sure what you mean by "Delete all these assertions that get_with_dump post_with_dump etc?"
I would:
-  change all calls to `<action>_with_dump` to call plain old `<action>`;
- Then delete the various defs of `<action>_with_dump`;
- Then delete `w3c_tests` all the calls to it .

nimmolo:
Thanks Joe. I didn't really read through the methods again before I responded, and i thought I had remembered some duplication. But it wasn't true duplication; it was `get` after login, for example. 

Converted all `#{method}_with_dump` to `#{method}`, and removed the `w3c_tests` stuff. The bulk of the PR work seems to have been worthwhile, since mostly it was adding the `params: ` name to all those params hashes. A residual benefit to keeping method syntax parallel.

JoeCohen (approved):

PR #1130, opened 2022-08-24
Views/SCSS - replace invalid HTML, update some class names

nimmolo:
This PR replaces the now-invalid HTML tags `<center>` and `<big>`, used in many places in MO, with the Bootstrap equivalent `<div class="text-center">` and `<span class="text-larger">`. I'd been replacing these haphazardly but it's easy enough to grep and replace all.

It also updates and reorganizes certain class names in `mushroom-observer.scss` to use Bootstrap 4 class name conventions. This should ease any potential future updates. The PR also converts some inline styles in views, in favor of these standard classes.

Lastly, it removes a few undefined classes that were used in views, but were undefined anywhere in the CSS.

To test, navigate to any of the affected views and check that the layout is as intended. In a few cases there will be 1px adjustments to padding or margin where units have changed from px to em, but in general, layouts should not have perceptible changes.


nimmolo:
If it's ok with everyone i'm just going to go ahead and merge this. 

It's only CSS. I've proofread my consolidations pretty carefully, but the truth is i am not only unfamiliar with many views; I don't have any idea how to get to some of these views through the front end, other than typing out the route. 

But this PR is pretty low stakes; if i mess up the padding in a layout i can easily fix. 

The idea here is basically to make views easier to style and alter. Both by standardizing styling classes, and by bringing some existing class names in line with Bootstrap 4 and 5.  

A further move will be to split up the mushroom_observer.scss file to make the theme variable assignment logic in it a bit more intelligible, and to make it easier to find classes by view.

PR #1131, opened 2022-08-27
Make name change email more robust with respect to missing descriptions.

pellaea:
Super trivial.


JoeCohen:
Looks good, and everything is now passing.
Thanks for dealing with this.

PR #1132, opened 2022-08-27
Add Visual Group class

mo-nathan:


app/controllers/visual_groups_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  before_action :login_required
+
+  # GET /visual_groups or /visual_groups.json
+  def index
+    @visual_groups = VisualGroup.joins(
+      "LEFT JOIN names ON visual_groups.group_name_id = names.id"
+    ).order(
+      "names.text_name"
+    )
+  end
```
nimmolo, line 13: @mo-nathan 
Joe, Jason and I have done a lot of work to rephrase SQL literals with the equivalent AR — our goal for test coverage and database portability is no SQL strings. I believe this will be just as efficient with
```ruby
  def index
    @visual_groups = VisualGroup.joins(:names).order(:text_name)
  end
```

mo-nathan: Actually that doesn't work.  I tried something like that initially but the problem is that generates:

`SELECT visual_groups.* FROM visual_groups INNER JOIN names ON names.visual_group_id = visual_groups.id ORDER BY text_name ASC`

rather than the intended

`SELECT visual_groups.* FROM visual_groups LEFT JOIN names ON visual_groups.group_name_id = names.id ORDER BY text_name ASC`

The first query returns a bunch of duplicate VisualGroups.  However, I did adopt the `.order(:text_name)`.  I wasn't sure that was going to work, but it does.

nimmolo: Agh sorry Nathan... It's `VisualGroup.left_outer_joins(:names).order(:text_name)`

https://guides.rubyonrails.org/v6.1.6/active_record_querying.html#left-outer-joins

mo-nathan: Actually that's even worse.  That generates:
`SELECT visual_groups.* FROM visual_groups LEFT OUTER JOIN names ON names.visual_group_id = visual_groups.id ORDER BY text_name ASC`
which generates even more rows due to the outer join.  The point is that we want to be joining on `visual_groups.group_name_id` not on `names.visual_group_id`.  VisualGroups have only one `group_name_id`, but the relationship from Name to VisualGroup is a many to one relationship.  My code will generate exactly one row for each VisualGroup.  You code generates one row for each Name that has a VisualGroup (and the outer join case throws in a couple more because of the NULL case).

nimmolo:
Aha, ok, yes. This is why the SQL conversions have been so time consuming… we tested all of them as you are doing now. I’ve easily spent 200 hours just on these conversions.

There’s definitely a way to do it in AR, i’ll research it more carefully when I have a minute.

Nimmo

mo-nathan:
I'm actually debating changing the way I have the data structured completely.  Circular references like that are bad and dangerous.  Thanks for encouraging me to rethink it.

mo-nathan:
I actually think it would be better to just give the VisualGroup a separate name field that isn't a Name.  Then it could be more free form and I wouldn't have to deal with the weird "clade" and "complex" groups and such.

nimmolo:
Ok. Well in any case, sorry for the roundabout. 

Now I did my homework, and for future reference: If you change VisualGroup to inherit from AbstractModel, like our other models, it will get the ArelHelpers. Then, for example, you can write this, which may narrow down to the results you need, but it generates an ugly double `where` condition.:
```ruby
VisualGroup.left_outer_joins(:names).where(VisualGroup[:group_name_id] == Name[:id]).order(:text_name)
```
Result (I just tested in rails console):
```
"SELECT `visual_groups`.* FROM `visual_groups` LEFT OUTER JOIN `names` ON `names`.`visual_group_id` = `visual_groups`.`id` WHERE `visual_groups`.`group_name_id` = `names`.`id` ORDER BY `text_name` ASC"
```

With straight Arel, however, you can be as precise as you like. The following is exactly equivalent to your SQL, and I believe preferable:
```ruby
VisualGroup.joins(VisualGroup.arel_table.join(Name.arel_table, Arel::Nodes::OuterJoin).on(VisualGroup[:group_name_id] == Name[:id]).join_sources).order(:text_name)
```
and generates the very succinct:
```
"SELECT `visual_groups`.* FROM `visual_groups` LEFT OUTER JOIN `names` ON `visual_groups`.`group_name_id` = `names`.`id` ORDER BY `text_name` ASC"
```

pellaea:
An hour late and a dollar short, but... for the initial problem I would
suggest trying eager loading instead:

@visual_groups = VisualGroup.includes(:group_name).sort_by { |v|
v.group_name.text_name }

It turns it into two straight-up non-joined queries.  Just a thought.

Also, did you consider sorting based on Name#sort_name instead of
Name#text_name?  May not make a difference if all the names are guaranteed
to be the same rank (i.e. "group").

mo-nathan:
@nimmolo Thanks for digging into finding the correct way to do this with Arel.  It's good to be aware of even if I decide to not go that way in the end.

@pellaea Interesting question.  My initial goal was to try something that roughly grouped taxa together that I expect to be visually similar.  My guess was that author differences more often than not would refer to very similar taxa so for a first pass I decided to lump rather than split.  However, I definitely expect that these groups will need to be reviewed by a human to decide if they make sense.  Just now I looked for cases where using sort_name would result in something different from text name using this query:

`select n1.id, n1.sort_name, n2.id, n2.sort_name from names n1, names n2 where n1.id < n2.id and n1.visual_group_id = n2.visual_group_id and n1.text_name = n2.text_name and n1.sort_name != n2.sort_name and n1.visual_group_id is not null;`

It's pretty slow query (~18 seconds on my machine) and it requires that my migrations already be run, but it finds 83 cases.  The examples I've looked at generally confirm my guess.  Here are a few cases in case you're interested:

```|     7 | Xerocomus subtomentosus  (L.) Quél.                           |  64455 | Xerocomus subtomentosus  sensu auct. PNW                                               |
|    93 | Ramaria botrytis  (Pers.) Ricken                              | 109919 | Ramaria botrytis  (Pers.) Bourdot                                                      |
|   154 | Morchella elata  Fr.                                          |  43442 | Morchella elata  Fr. sensu Clowez                                                      |
|   209 | Inocybe  sensu lato                                           | 109255 | Inocybe  (Fr.) Fr.                                                                     |
|   223 | Hygrophorus chrysodon  (Batsch) Fr.                           |  60273 | Hygrophorus chrysodon  sensu CA                                                        |
|   227 | Hygrocybe miniata  (Fr.) P. Kumm.                             | 108565 | Hygrocybe miniata  auct. amer.                                                         |
```

One case that I noticed that's more interesting is: _Amanita franchetii_


```|   378 | Amanita franchetii  sensu Thiers                              |  18337 | Amanita franchetii  sensu auct. europ.                                                 |
|   378 | Amanita franchetii  sensu Thiers                              |  18416 | Amanita franchetii  (Boud.) Fayod                                                      |
|   378 | Amanita franchetii  sensu Thiers                              |  20269 | Amanita franchetii  sensu Dav. T. Jenkins                                              |
| 18337 | Amanita franchetii  sensu auct. europ.                        |  18416 | Amanita franchetii  (Boud.) Fayod                                                      |
| 18337 | Amanita franchetii  sensu auct. europ.                        |  20269 | Amanita franchetii  sensu Dav. T. Jenkins                                              |
| 18416 | Amanita franchetii  (Boud.) Fayod                             |  20269 | Amanita franchetii  sensu Dav. T. Jenkins                                              |
```

This is an interesting case since "Amanita franchetii  sensu Thiers" now has the name "Amanita augusta" and I believe is endemic to the western US.  The diligent Amanita lovers on MO apparently know this since "Amanita franchetii  sensu Thiers" is longer the consensus name for any observations.  True "Amanita franchetii" is from Europe and might be visually distinct enough for a ML model to notice, but I'd have to test that.  At first I figured "Amanita franchetii  sensu Dav. T. Jenkins" was just a synonym for "Amanita franchetii  sensu Thiers", but it led me to this interesting observation https://mushroomobserver.org/21306 which probably should be excluded from the ML training set.

Ultimately this again leads me to thinking that VisualGroups should have their own names rather than trying to use an existing Name.  I still think this is a reasonable way to create some initial VisualGroups, but we'll have to see what lumping and splitting makes sense beyond that.

So many interesting rabbit holes!

pellaea:
Nice!  I've played with queries like that where you link a table to itself,
e.g., through synonym_id.  A lot of fun.  But strangely expensive.  I don't
know what to say.  I am very glad that you are pursuing this approach --
visual groups -- I suggested it to the last 10 groups working on ML using
our MO's data, and no one ever paid any attention.  So it's exciting that
you're taking it on!!  But as I probably should have expected, the deeper
you go into the problem the more strange cases you find.  Isn't that always
the way it goes when fungal taxonomy is involved??? ;)

app/controllers/visual_groups_controller.rb
mo-nathan: Decided to not use Names directly with VisualGroups making this kind of moot.

mo-nathan:
Did you mean to merge this?  It’s no where near ready.

pellaea:
No, my mistake.  I've already reverted the PR and made a new one branched
from master instead of visual-group-scaffold.  Sorry about that.

PR #1133, opened 2022-08-28
View templates: fix `form.select` syntax (with class)

nimmolo:
To reproduce: go to your account preferences, select the Amanita theme, and reload the page. Preference selects are not really very legible; insufficient contrast.

Diagnosis:
In the view templates, form selects (in user prefs and elsewhere) correctly specify the Bootstrap `class: "form-control"`, but the class attribute is not created in the HTML, due to incorrect syntax in the erb tag. Result: the select options are not very legible; the text color is incorrectly inherited from the parent element.

Fix: The `form.select` helper requires two options hashes. The `class` has to be in the second hash, rather unintuitively. (Someone seems to have discovered this already, because `views/name/form_name.html.erb` already had the correct syntax.)

### DIFF
--- a/app/assets/stylesheets/Agaricus.scss
+++ b/app/assets/stylesheets/Agaricus.scss
@@ -11,20 +11,22 @@ $xanthodermus_stain:   #D4A833;
 $polyozellus_light:    #a5bffa;
 $polyozellus_dark:     #5167b2;
 
+$BUTTON_FG_COLOR:             $cupreobrunneus_gills;
+
 $BODY_FG_COLOR:               $cupreobrunneus_gills;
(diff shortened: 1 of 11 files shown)


nimmolo:
This is preliminary to the other SCSS consolidation PR. Will merge today, if no objections
Let me know if your `select` elements appear wacky, wonky, or otherwise. They shouldn't be.

Note: if you don't feel comfortable reviewing SCSS, the easy way to review these PRs is to simply navigate to the page affected and note any front-end changes vs master. The change here just fixes a long-term legibility bug in the Amanita theme, which apparently nobody noticed until now because i'm the only one using Amanita. (Blame "Amanita Rita" aka MO user Debbie Viess!)

PR #1139, opened 2022-09-02
Make it so non-logged in users can see glossary.

pellaea:
Next installment in my series of one-line PRs!

The glossary is so rarely used I believe this cannot possibly cause a drain on performance.  Since I think at least some visitors encountering pages that require login are finding MO to be unwelcoming, it doesn't seem worth doing so for the glossary.  Site stats?  User summaries?  Yes, those are expensive pages, I'm fine with requiring login.  But the glossary is different.

Feel free to disagree and veto this PR.  It's not like I worked hard on it. :)

### DIFF
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -2,7 +2,7 @@
 
 # create and edit Glossary terms
 class GlossaryTermsController < ApplicationController
-  before_action :login_required # except: [:index, :show, :show_past]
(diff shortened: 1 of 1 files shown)


pellaea:
Oh wow, good point, Joe.  Then let's add paging to the index before accepting this PR.  (I'll do it when I have time.  It will probably be several weeks.  This is especially busy time for me.)

nimmolo:
Closing this in favor of a new one that does the same thing, this seemed daunting to bring up to date.

PR #1143, opened 2022-09-05
Upgrade labels

pellaea:
Upgrade the Print Labels feature to produce a rich text version.  This was requested by Joanne Schwartz of FunDiS.  If anyone would like to alter the format or content, we might need to consider providing multiple options, as she was quite comprehensive with her needs.  But I can try to negotiate with her.


pellaea:
Okay, this now has literally everything associated with the observation except the name proposals, comments and images!  It seems really excessive to me, but this is what Joanne wants.  And probably she's not alone.

I feel like this is okay as a first draft, just so we can get this out so she and others associated with FunDiS can use it.  Eventually it may be handy to have a set of checkboxes allowing users to deselect things like notes, projects, species lists, sequences, etc.  For example, if we wanted to use this feature to produce labels for most lichen herbaria I work with, we would need to radically simplify the content of the labels in order to fit on standard-sized lichen packets.  But I think it's fine to relegate that kind of work for a future PR when and if someone requests it.  (And is as persistent as Joanne... ;)

Does this work for everyone else?

mo-nathan (approved):
Not sure what happened to my previous review.  In any case this looks fine to me.  My thoughts were that I'd prefer the lat/long to be on a separate line from the location name and having the field labels be all lower case looks odd to me, but if this is what the Fundis folks want it's fine with me.

pellaea:
We can definitely change the style. When i can I'll try another and see if
you like it better. Joanna as only really concerned with content, not order
or style.

mo-nathan:
Ok, if lat/long could be on another line I think it would tighten them up in cases where there isn't a huge amount of other stuff.  I'd make the field names have a leading capital.  I'd also consider making either make the field names or the contents bold just so they are visually distinct.  Making the field names bold is probably simpler in case any of the values are formatted and might contain stuff in bold.

pellaea:
Sounds good, I'll try to get to that soon.  Should be really easy changes.  Thanks for looking at it, Nathan!

pellaea:
I still can't run my dev VM server locally, so I don't know what these changes I just made look like.  But I think they are pretty straightforward, so hopefully I didn't mess anything up.

pellaea:
Aha!  Got my dev VM working and cleaned up the labels.  Should be good to go now.

I will be out for a bit.  If this is now acceptable would someone merge and deploy it and tell Joanne?  explorerdj at mac.com

Thanks!

JoeCohen:
Some of the fields can take Textile. That probably should be rendered, as opposed to exporting raw Textile.
I don't have time to deal with that until next week, so I will merge and deploy today as-is.
FunDis can let us know if they want changes.

pellaea:
Yes, I couldn't think of any easy way to handle that. Maybe there's an HTML
to RTF translator gem?

JoeCohen:
I couldn't find a gem.
But I did find this.
https://gemfury.com/svassallo/ruby:rtf/-/content/lib/rtf/converters/html.rb
Not sure how to use it.

pellaea:
Wow, that's dead simple.  I don't want *exactly* that but I might steal the
code (with appropriate attribution of course).

JoeCohen:
Jason:
I'll look at this tomorrow a.m. I'm too tired to do anything requiring
thought right now.

PR #1144, opened 2022-09-06
Fix image float-right in comments

nimmolo:
In my previous SCSS refactor moving inline styles to Bootstrap-4-syntax utility classes, I forgot to define the classes for `float-right` and `float-left`, although I used `float-right` in the comment markup. 

This defines the classes and fixes the layout.


pellaea:
Thank you, Nimmo!

PR #1145, opened 2022-09-18
add newrelic configs

gete76:
This adds the New Relic gem and agent config file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,6 +126,9 @@ gem("mail")
 # for detecting file type of uploaded images
 gem("mimemagic")
 
+
+#for monitoring
+gem('newrelic_rpm')
 # for creating zip files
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -126,6 +126,9 @@ gem("mail")
 # for detecting file type of uploaded images
 gem("mimemagic")
 
+# for monitoring
+gem("newrelic_rpm")
+
```
JoeCohen, line 131: I suggest moving to the end of the file. But more important, put it in a production group. Otherwise it will tryh to talk to the web in the test environment.
```suggestion
group :production do
  # Application performance monitoring
  # https://one.newrelic.com/marketplace
  gem('newrelic_rpm')
end
```

PR #1146, opened 2022-09-21
Separate IntegrationTestCase <-> CapybaraIntegrationTestCase

nimmolo:
MO currently has two discrete styles of integration test: 
- those that depend on `rails-dom-testing` methods and MO's extensions to those methods in `IntegrationExtensions` and `SessionExtensions`, and 
- those that depend on `capybara` methods, and none of MO's extensions (currently).

Currently also, the `Capybara::DSL` class is included in all integration tests, but not `Capybara::Minitest::Assertions`. The latter class of assertions would help us write tests with better error messages, but if it is included in `ActionDispatch::IntegrationTest`, it causes current non-Capybara integration tests to fail due to method name conflicts.

This PR separates our `integration_test_case.rb` into two test case classes inheriting from `ActionDispatch::IntegrationTest`, with `capybara_integration_test_case.rb` being the new one. The former would continue to include MO integration/session extensions, the latter would now be the only Integration test case that requires Capybara, includes Capybara classes, and gets a couple new `login` session extension methods especially for capybara, to DRY up the tests. 

This PR also 
- sorts all the tests into one of two folders within `/test/integration`, to avoid confusion.
- folds the two `IntegrationExtensions` methods into `SessionExtensions`. (They are only ever used together, by "regular" IntegrationTests. `IntegrationExtensions` contained the more session-oriented methods for `login`.)


Gemfile
```diff
@@ -268,7 +268,9 @@ end
 
 group :test do
   # Use capybara to simulate user-browser interaction
-  gem("capybara")
+  gem("capybara", "~> 3.37", ">= 3.37.1")
```
nimmolo, line 271: I don't know why this doesn't already pull the latest Capybara (even with gem update). I upped the version at the suggestion of the Capybara author.

test/capybara_integration_test_case.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+# Allow simuluation of user-browser interaction with capybara
+require("capybara/rails")
+require("capybara/minitest")
+
```
nimmolo, line 6: Moved these requires to the test case, from `test_helper.rb`

Gemfile
```diff
@@ -268,7 +268,9 @@ end
...
+
+  # gem("selenium-webdriver")
```
nimmolo, line 273: Maybe later, in case we want to test js...

test/capybara_session_extensions.rb
```diff
@@ -3,24 +3,24 @@
 #
 #  = Integration High-Level Test Helpers
 #
-#  Methods in this class are available to all the integration tests.
+#  Methods in this class are available to Capybara integration tests.
 #
 #  login::   Create a session with a given user logged in.
 #  login!::  Same thing,but raise an error if it is unsuccessful.
 #
```
nimmolo, line 10: Note that the `IntegrationExtensions` methods `login` and `login!` were moved to `SessionExtensions` for rails-dom-testing. They're always included together.

test/capybara_session_extensions.rb
```diff
@@ -3,24 +3,24 @@
...
-      form.submit("Login")
-    end
+    visit("/account/login")
+
+    fill_in("user_login", with: login)
+    fill_in("user_password", with: password)
+    check("user_remember_me") if remember_me == true
+
+    click_button("Login")
   end
```
nimmolo, line 24: Rewrite of `login` for Capybara

test/integration/capybara/filter_test.rb
```diff
@@ -15,10 +15,7 @@ def test_user_content_filter
                  where.not(thumb_image_id: nil)
 
     reset_session!
-    visit("/account/login")
-    fill_in("User name or Email address:", with: user.login)
-    fill_in("Password:", with: "testpassword")
-    click_button("Login")
+    login(user)
```
nimmolo, line 19: DRY: Use the new `CapybaraSessionExtensions` method

test/integration/capybara/observations_controller_supplemental_test.rb
```diff
@@ -7,16 +7,7 @@
...
-    visit("/account/login")
-    fill_in("User name or Email address:", with: user.login)
-    fill_in("Password:", with: "testpassword")
-    click_button("Login")
-  end
-
-  # ------------------------------------------------------------
-
+class ObservationsControllerSupplementalTest < CapybaraIntegrationTestCase
   # Prove that when a user "Tests" the text entered in the Textile Sandbox,
```
nimmolo, line 11: Method can now inherit from `CapybaraSessionExtensions#login`

test/test_helper.rb
```diff
@@ -102,15 +100,15 @@ class TestCase
...
 
     # Instantiated fixtures are slow, but give you @david where
     # otherwise you would need people(:david).  If you don't want to
     # migrate your existing test cases which use the @david style and
     # don't mind the speed hit (each instantiated fixtures translates
     # to a database query per test method), then set this back to
-    # true.
+    # true. Defaults to false.
     self.use_instantiated_fixtures = false
     ##########################################################################
```
nimmolo, line 113: `test_case.rb` is quite a long file and seems more complicated than it really is.

Do we still need these? They seem to date from the original setup of MO; the linked article is offline.
Both methods reiterate the defaults and don't seem to have changed in eons... but if we want to keep, i'm fine with it.

test/test_helper.rb
```diff
@@ -171,10 +169,3 @@ def clear_logs
     end
   end
 end
-
-# Make the Capybara DSL available in all integration tests
-module ActionDispatch
-  class IntegrationTest
-    include Capybara::DSL
-  end
-end
```
nimmolo, line 180: Moved to `CapybaraIntegrationTestCase`

test/capybara_integration_test_case.rb
```diff
@@ -0,0 +1,86 @@
...
+#        using_session(mary) { visit('/edit_rolfs_stuff') }
+#        mary.assert_redirect
+#
+#        # You can also create anonymous sessions.
+#        session = Capybara::Session.new(:rack_test, Rails.application)
+#        session.visit('/show_index')
+#        session.assert_selector('span', :text => 'Rolfs Thing', :count => 0)
+#      end
+#    end
+#
```
nimmolo, line 47: Revised comments for Capybara.

PR #1147, opened 2022-09-23
`div.results` - give it an id, not a class

nimmolo:
First of several PRs adding html `id` attributes to elements in the site, very useful for integration tests.

`div.results` is one of the main page sections lacking a unique identifier. There's almost always only one `.results` div in an index, and this PR switches it to `div#results`. References in tests to `div.results` are changed to `#results`. 

> **EDIT**: Because the locations index has two paginated results divs, I just added a param `html_id` to `paginate_block`. This allows a second paginated results div to have a custom `id`. For undefined locations, it's `locations_undefined`. I'm going to try to follow that naming convention for the many dozens of ids I'm adding to the next PR. 

That's all there is to this PR. Will merge very soon if no objections.

### DIFF
--- a/app/views/observations/index.html.erb
+++ b/app/views/observations/index.html.erb
@@ -36,7 +36,7 @@
 <% end %>
 
 <%= paginate_block(@pages) do %>
-  <div class="row results mt-3">
+  <div class="row mt-3">
     <%= render(partial: "shared/matrix_box", layout: "shared/matrix_table",
                collection: @objects, as: :object) %>
     <div style="clear: left"></div>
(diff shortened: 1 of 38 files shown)


app/views/observations/index.html.erb
```diff
@@ -36,7 +36,7 @@
 <% end %>
 
 <%= paginate_block(@pages) do %>
-  <div class="row results mt-3">
+  <div class="row mt-3">
     <%= render(partial: "shared/matrix_box", layout: "shared/matrix_table",
```
nimmolo, line 40: This was always a superfluous class here. It's rendered in the front end as basically:
```
<div class="results">
  <br>
  <ul class="pagination pagination-sm">
  <div class="row results mt-3">
```

PR #1149, opened 2022-09-24
Title and logo — HTML classes and ids

nimmolo:
Part of a series of PRs to generate more predictable page elements for test refactoring.

- Clean up logo classes and ids, currently somewhat confusing. One `id` was not unique on the page
- Eliminate unused CSS `#navigation .logo` style block (logo moved out of #navigation when logo changed, apparently)
- Refactor page title bar and title "caption" to use an `h1` element, and rename the ids and classes used there.
- Change "#title" to "#title_bar" and "#title_caption" to "#title"
- Change ".owner-id" to "#owner_id" and "#flash-notices" to "#flash_notices"
- Adjust tests to find the new ids, removes element tag names from selectors ("div#title" to "#title_bar"

Curiously, MO does not currently generate a machine-readable page heading (generally, that would be an `h1` element). It doesn't have to be styled like an h1... Bootstrap allows for restyling headings like so, keeping the same appearance we have currently: 

`<h1 class="h3" id="title"><%= @title %></h1>`

### DIFF
--- a/app/assets/javascripts/vote_popup.js.erb
+++ b/app/assets/javascripts/vote_popup.js.erb
@@ -77,7 +77,7 @@ function VotePopupModule(translations) {
               var title = html.children().eq(0);
               var div   = html.children().eq(1);
               document.title = title.attr("title");
-              $("#title-caption").html(title.html());
+              $("#title").html(title.html());
               $("#naming_partial").html(div.html());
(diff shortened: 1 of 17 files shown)


nimmolo:
Small PR. 
Will merge today if no objections.

PR #1151, opened 2022-09-25
Api delete user

pellaea:
Enable user to request to have their account deleted from mobile app.

This creates a new User#disable_account_and_delete_private_objects method that deletes all of a user's content that isn't integral to other users' content.  If there is nothing left after doing so, it completely removes the account from the database, otherwise it just disables the account (sets the password to "", marks the account as blocked).  Always succeeds.


app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+    observations.delete_all
+  end
+
+  # Delete user's descriptions that don't have any other authors or # editors.
+  def delete_private_name_descriptions
+    (name_descriptions -
+      name_descriptions.joins(:name_description_authors).
+        where.not(name_description_authors: { user_id: id }) -
+      name_descriptions.joins(:name_description_editors).
+        where.not(name_description_editors: { user_id: id }))
```
JoeCohen, line 1012:
```suggestion
        where.not(name_description_editors: { user_id: id })).
```

app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+      delete_all
+  end
+
+  # Delete user's descriptions that don't have any other authors or # editors.
+  def delete_private_location_descriptions
+    (location_descriptions -
+      location_descriptions.joins(:location_description_authors).
+        where.not(location_description_authors: { user_id: id }) -
+      location_descriptions.joins(:location_description_editors).
+        where.not(location_description_editors: { user_id: id }))
```
JoeCohen, line 1022:
```suggestion
        where.not(location_description_editors: { user_id: id })).
```

JoeCohen:
Should test/models/api2_test.rb line 4185 be changed as follows?
```diff
-     do_help_test(:delete, :user, fail: true)
+     do_help_test(:delete, :user)
```

test/models/user_test.rb
```diff
@@ -335,4 +335,194 @@ def test_notes_template_validation
...
+  end
+
+  def test_delete_interests
+    assert_not_zero(Interest.where(user: junk).count)
+    junk.delete_interests
+    assert_zero(Interest.where(user: junk).count)
+  end
+
+  def test_delete_notifications
+    assert_not_zero(Notification.where(user: rolf).count)
```
JoeCohen, line 362: I have no idea why `assert_not_zero` blows up here, and didn't want to take the time to figure it out. 
So I suggest
```suggestion
    assert_operator( 0, "<", Notification.where(user: rolf).count)
```

test/models/user_test.rb
```diff
@@ -335,4 +335,194 @@ def test_notes_template_validation
...
+    rolf.delete_notifications
```
JoeCohen, line 363: Once I get past the prior line, this blows up with:
```ruby
ERROR UserTest#test_delete_notifications (334.58s)
Minitest::UnexpectedError:         ActiveRecord::NotNullViolation: Mysql2::Error: Column 'user_id' cannot be null
            app/models/user.rb:991:in `delete_notifications'
            test/models/user_test.rb:364:in `test_delete_notifications'
```
I'm way out of my depth here.

JoeCohen:
Sorry, can't make any more progress on the test failures.

test/models/user_test.rb
nimmolo: I think you need `assert_not_nil` here. That seems to work for me (cloned the branch, trying it locally)

app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+  # I think it's okay to delete observations even if they are attached to a
+  # project.  But species_lists are potentially a much more collaborative
+  # effort, so I don't think it's okay to delete lists that are attached to
+  # public projects just because the user happened to originally create them.
+  # -JPH 20220916
+  def delete_private_species_lists
+    (species_lists -
+      species_lists.joins(:project_species_lists)).
+      delete_all
+  end
```
nimmolo, line 1048: Tried this locally. `delete_all` needs an AR relation. I think it needs to be something like 
```ruby
    ids = (species_lists -
            species_lists.joins(:project_species_lists)).map(&:id)
    SpeciesList.where(id: ids).delete_all
```
This works, but reveals another issue: test fixtures shows 8 SLs for mary, but only 3 where SL.joins(PSL). 

That leaves 3 not deleted by `UserTest#test_delete_private_species_lists`, and the test is only expecting one.

app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+
+  def delete_unattached_collection_numbers
+    (collection_numbers -
+      collection_numbers.joins(:observation_collection_numbers)).
+      delete_all
+  end
```
nimmolo, line 1054: I tried this, like the above, but it says `undefined local variable or method` for `collection_numbers`

```ruby
    ids = (collection_numbers -
      collection_numbers.joins(:observation_collection_numbers)).map(&:id)
    CollectionNumber.where(id: ids).delete_all
```

app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+
+  def delete_unattached_images
+    (images -
+      images.joins(:glossary_term_images) -
+      images.joins(:observation_images) -
+      images.joins(:project_images) -
+      images.joins(:subjects)).
+      delete_all
+  end
```
nimmolo, line 1069: This worked, or at least the UserTest passes:
```ruby
    ids = (images -
      images.joins(:glossary_term_images) -
      images.joins(:observation_images) -
      images.joins(:project_images) -
      images.joins(:subjects)).map(&:id)
    Image.where(id: ids).delete_all
```

app/models/user.rb
```diff
@@ -940,6 +940,184 @@ def remove_image(image)
...
+    (collection_numbers -
+      collection_numbers.joins(:observation_collection_numbers)).
+      delete_all
+  end
+
+  def delete_unattached_herbarium_records
+    (herbarium_records -
+      herbarium_records.joins(:observation_herbarium_records)).
+      delete_all
+  end
```
nimmolo, line 1060: I tried this:
```ruby
    ids = (herbarium_records -
      herbarium_records.joins(:observation_herbarium_records)).map(&:id)
    HerbariumRecord.where(id: ids).delete_all
```
and got `ActiveRecord::RecordNotFound expected but nothing was raised`.

That seems to be because the test expects the `rec` fixture `herbarium_records(:interesting_unknown)` to be deleted,
but the one getting deleted in the last part of the test is `herbarium_records(:fundis_record)` according to debugger.

test/models/user_test.rb
JoeCohen: 1. While `assert_not_nil` avoids the Error, it can't be right.
The assertion (`assert_not_zero`) ensures that the fixtures are OK before proceeding with the rest of the test.
In particular, the test makes sense only if Rolf starts with some notifications.
`assert_not_nil` will give the wrong result (`true`) if Rolf has 0 Notifications.
2. `assert_not_zero` just blows up in this particular test method. 
With a breakpoint at line 362, we manually confirm that the `count` is correct:
```ruby
 Notification.where(user: rolf).count => 1
```
But `assert_not_zero` throws a NoMethodError:
```ruby
assert_not_zero(1) =>
ERROR UserTest#test_delete_notifications (230.16s)
Minitest::UnexpectedError:         NoMethodError: undefined method `assert_not_zero' for #<UserTest:0x0000555f48ca6578 @NAME="test_delete_notifications", @failures=[#<Minitest::UnexpectedError: Unexpected exception>], @assertions=1, @fixture_cache={"users"=>{"rolf"=>#<User 241228755: "Rolf Singer (rolf)">}}, @fixture_connections=[], @connection_subscriber=#<ActiveSupport::Notifications::Fanout::Subscribers::Timed:0x0000555f48ca4980 
...
Did you mean?  assert_not_same
               assert_not_nil
  (rdbg)//vagrant/mushroom-observer/test/models/user_test.rb:1:in `test_delete_notifications'

```
How did `assert_not_zero` get undefined just for this particular test? It works fine elsewhere in this file. See following screen shot.
Screen Shot 2022-09-26 at 7 20 11 AM

JoeCohen: I was wrong. `assert_zero` and `assert_not_zero` are not working in other places.
I think there's no such thing as `asert_zero` or `assert_not_zero`.
So these should probably be globally changed to
`assert_equal(0, ...` or `assert_operator( 0, "=", ...)` 
`assert_operator( 0, "<",  ...)`

pellaea:
Thanks, Joe.  I didn't have access to internet when I wrote this.  I didn't
expect assert_zero / assert_not_zero to exist, but wanted to look up what
*does* exist when I finally could do so.

JoeCohen:
1. `assert_zero` read so nicely that I thought it must exist.
2. There's a typo in my prior email. It should have a double equals
operator:
> assert_equal(0, ... or assert_operator( 0, "==", ...)

mo-nathan:
I looked at the tests a bit and I don't fully understand the intent of the code.  For example, one of the issues is that on line 973 of app/models/user.rb it says `self.blocked = true`.  This suggests adding a boolean to the User class which might well make sense, but I don't see any migration or other indication that this was the real intention of the code.  The same test then fails a bit later if I comment out that line with `Mysql2::Error: Column 'user_id' cannot be null`, which I suspect has to do with an issue in the fixtures, but I haven't put the time into hunting down exactly what the issue is there.

The other test I looked at is in api2_test.rb in the `test_help` method.  This test is failing on the very last line which I think is expecting the user delete to fail, but it appears to be succeeding.  Removing the `fail: true` from that line allow the test to pass.  However, I'm again not 100% sure that is the intent here.

JoeCohen:
Thanks Nathan! That's helpful.
I'm slooowly working my way through this.
Started with 10 Errors, 5 Failures.
Now at 5 Errors, 6 Failures.

app/models/user.rb
```diff
@@ -940,6 +940,189 @@ def remove_image(image)
...
+          map(&:id)
+    LocationDescription.where(id: ids).delete_all
+  end
+
+  # Delete all the user's projects that don't have any other users on them.
+  def delete_private_projects
+    Project.joins(:admin_group, :user_group).
+      where(user: self,
+            admin_group: { name: "user #{id}" },
+            user_group: { name: "user #{id}" }).
```
JoeCohen, line 1033: This is short and it works.
But perhaps is a bit hacky; it depends on the convention for naming a user's personal UserGroup.
See `UserGroup#self.create_user`.
IMO we should get quickly to something that works so that the iOS mobile app can be published.
We can refactor later.

test/models/user_test.rb
```diff
@@ -335,4 +335,204 @@ def test_notes_template_validation
...
+                            count
+    assert(deleteable_proj_count >= 1,
+           "user should have >= 1 project that's deleted when s/he's deleted")
+    old_proj_count = Project.where(user: dick).count
+
+    dick.delete_private_projects
+
+    assert(Project.where(user: dick).count ==
+             old_proj_count - deleteable_proj_count)
+    assert_not_empty(Project.where(id: bolete.id))
```
JoeCohen, line 445:
```ruby
assert_not_empty(Project.where(id: bolete.id))`
``` 
instead of 
```ruby 
assert_not_nil(Project.find(bolete.id))
```
because the latter throws `ActiveRecord::RecordNotFound` if the Project doesn't exist.

JoeCohen:
@pellaea: What, if anything, do I immediately need to do with `blocked` other than setting it?

> I looked at the tests a bit and I don't fully understand the intent of the code. For example, one of the issues is that on line 973 of app/models/user.rb it says `self.blocked = true`. This suggests adding a boolean to the User class which might well make sense, but I don't see any migration or other indication that this was the real intention of the code. The same test then fails a bit later if I comment out that line with `Mysql2::Error: Column 'user_id' cannot be null`, which I suspect has to do with an issue in the fixtures, but I haven't put the time into hunting down exactly what the issue is there.

JoeCohen:
Converting to a Draft PR until tests pass.

JoeCohen:
Thanks for stealing some time in order to respond.
I'll see how much progress I can make on the current PR, then recalibrate.

JoeCohen:
PS: I need to review `User.erase_user(user.id) unless Comment.where(user:
user).any?` as a shortcut.

JoeCohen:
I am putting this PR aside for now: I'm out of time and there's too much for someone of my skill level to deal with quickly.
I agree with @pellaea that the immediate way forward is the temporary fix he proposed [here](https://github.com/MushroomObserver/mushroom-observer/pull/1151#issuecomment-1259530597).

pellaea:
CI is failing because of a pending migration.  Why is it not doing the migration?  The tests are all passing for me.  I don't want to check in my db/schema.rb because it has diverged from the one on master for many months now.  That might be the problem.  But if I can't check it in without causing a mess on the production server, how do I solve it?  (My schema has all the old glue tables still in it, it's a real mess.)

app/models/user.rb
pellaea: I didn't like this because there may be other user groups with just that one user in it.  In fact the test fixtures bolete_admins and bolete_users both contain only Dick!  I think I found a better way to do it which is more in line with all the other related methods.

Also, on a related note, I noticed that the test had been rewritten to basically use the exact same logic as the method it was testing (it looked like a copy and paste from the method), so of course if the logic is wrong in the code, then the test won't catch it.  I realize the motivation is to make the test more robust with respect to changing fixtures, but it also makes the test less useful, much harder to understand, and well, prone to being outright wrong without realizing it!  (Such as in this case.)  I rewrote it hard-coded the way you presumably hate, but compromised by documenting better what all the requirements are that are being tested and making sure they are all met before I even begin the actual testing itself.  At least this way when the test breaks because of someone monkeying around with the fixtures, it should be easier to understand how to modify the test to accommodate the changes.

(I personally think the better solution to this problem, instead of writing more complex and potentially buggy tests, is to leave the fixtures alone!  Instead, if you need some very particular setup for one or two tests, create it on the fly in those tests, and not in the application-wide fixtures.  That way the details of the new fixtures are clearly "documented" where they are needed (and you don't have to go searching ad nauseum through the fixtures to understand them), and they won't break any other tests.  Everyone wins.  Except maybe it's a bit slower.  I am a very strong fan of keeping tests as logic-free as possible.  Tests do not need to be DRY.  But that's just my two cents.)

JoeCohen: Thanks for the re-write. I won't have time to look at it.
I'm good with more hard-coding in the tests (and also testing the fixtures to make sure they're not broken.)
I agree with the factory approach -- creating fixtures on the fly, especially when the relationships are as complicated as they are here.

JoeCohen:
IWhat if I run the migration, then push the changed schema?

pellaea:
Yes, that might work.

I also wonder if I just need to recreate my schema.rb.  I'm using a
snapshot of the live database, so those old tables must not actually be
there, right?  That would therefore mean that my schema.rb is out of sync
with the actual database on my dev VM.  That doesn't seem smart or safe...
something that should be fixed regardless of this PR.  I'll try grabbing a
fresh snapshot and doing that.

PR #1153, opened 2022-09-27
Simplify matrix views

nimmolo:
This greatly simplifies the logic and html on `matrix_box` pages. 

I'd also like to get rid of the `matrix_table` layout (now called `matrix_grid`) because it is now superfluous, but i seem to need the logic of it for some reason on `reuse_image` and `remove_image` templates. 

If i try to bypass the call to `render(layout: "shared/matrix_grid", locals: {objects: @object.images}) do |image|` and directly render the block of html by iterating over each image, like this: 
```ruby
  @object.images.each do
    ...an html block with the image and a checkbox...
  end
```
... the loop outputs the images correctly, but then adds a "proxy" of the first **image record** at the end of the block, which is not intended, and rightly flagged as `unsafe_html`.

Suggestions welcome, it would be nice to solve this now, but not really necessary. This is already a big improvement on the current HTML for matrix box layouts, eliminating many useless divs.


app/views/image/_reuse_image.html.erb
```diff
@@ -12,18 +12,25 @@
...
+      <%= @objects.each_with_index do |image, idx| %>
+        <li class="matrix-box col-xs-12 col-sm-6 col-md-4">
+          <div class="panel panel-default">
+            <div class="panel-body">
+              <%= thumbnail(image, link: url_args.merge(img_id: image.id), original: true) %>
+            </div>
+          </div>
+        </li>
+      <% end %>
+    </ul>
```
nimmolo, line 34: Why does this generate HTML with an `unsafe_html` error:
```ruby
      <%= @objects.each_with_index do |image, idx| %>
        <li class="matrix-box col-xs-12 col-sm-6 col-md-4">
          <div class="panel panel-default">
            <div class="panel-body">
              <%= thumbnail(image, link: url_args.merge(img_id: image.id), original: true) %>
            </div>
          </div>
        </li>
      <% end %>
```
But this does not?
```ruby
      <%= render(layout: "shared/matrix_table",
                 locals: {objects: @objects}) do |image| %>
        <div class="mt-3 mb-3">
          <%= thumbnail(image, link: url_args.merge(img_id: image.id), original: true) %>
        </div>
      <% end %>
```

nimmolo:
@JoeCohen This is ready to go, too.

JoeCohen:
Deployed.

app/views/image/_reuse_image.html.erb
pellaea: That's an easy one!  I can answer that!  You're doing <%= each %> when you should only be doing <% each %>. :)

pellaea:
Oops, looks like I'm late to the party.  I saw the review request and started, not realizing it had already been deployed.  Sorry!

JoeCohen:
The comment was great. Now Nimmo can fix it.

app/views/image/_reuse_image.html.erb
nimmolo: Agh! ❤️ You're the best! 
Had me completely flummoxed, researching the inner secrets of `yield`...

Will do a small PR to get this fixed

JoeCohen:
Are you intending to leave the TODO
in  app/assets/javascripts/box_resizer.j?

nimmolo:
Yes... Hoping to move to bootstrap 4 within the month. The resizer script is unnecessary with display: flex.

PR #1155, opened 2022-09-28
Revert change to `.help-block` position, add `.position-relative` where tooltips are needed

nimmolo:
To make the positioning of "tooltip-arrows" more predictable, I had added `position: relative` to the class `.help-block`. This had the unintended consequence of blocking clicks on **links in nested, floated divs** such as the one for "Propose a name". 

This PR reverts that change and adds  the class `.position-relative` only to help-blocks with tooltips, which (i checked) do not have nested button links.


nimmolo:
@JoeCohen Ready to go

JoeCohen:
@nimmolo I'm no use here; I don't have a mental picture how the scss and classes work with these views.
I only looked at the trivial parts (tidying up the forms.)
I'll merge and deploy it.

PR #1158, opened 2022-09-29
Revise matrix-box : lazy-load images, vote on hover, refactor some JS

nimmolo:
Changes the layout of the matrix box, to show more image and less clutter:
- Extra large screens have four columns of matrix boxes instead of three
- Thumbnail fills the width of the box, up to a certain proportion of height/width
- Vote meter bar appears only on hovering over the image.


nimmolo:
Actually let me test and about this a bit more in the morning before deploying. Might be the wrong approach, i have another in mind.

nimmolo:
Closing in favor of #1379

PR #1159, opened 2022-09-30
CSS: Refactor div container classes

nimmolo:
In terms of layout, all of these are equivalent in Bootstrap 3. 
```
<div class="row">
  <div class="col-xs-12">
    ...block of stuff
  </div>
</div>
```

```
<div>
  <div>
    ...block of stuff
  </div>
</div>
```

```
...block of stuff
```

In other words, this particular nesting of divs with these particular classes (provided there are no other classes on either `div`, and the content is itself a block, and not inline content like a `<span>`) is for layout only, and it doesn't do anything (`row` adds negative margin, and then `col-xs-12` adds the same amount of padding, at all viewport sizes, resulting in two extra wrapping divs with no effect on layout). 

This PR tries to remove these divs everywhere relevant, to simplify layouts and ease the transition to Bootstrap 4.


nimmolo:
Yikes - blech! So many files changed. But I'd like to draw a line under this PR and I believe this is ready to send out. 

All the views I've checked are in a workable state, and if anything has snuck by, i can fix.

nimmolo:
@pellaea, @JoeCohen  This is a views-only PR. 

It's a vast swath of changes and i've been pretty careful with it, but I can't quite imagine visiting every view to check the layouts. I'm inclined to let'er rip and let users do the QC, with me standing by to fix, since a lot of it is going to keep changing in the next week anyway. But I will defer to your opinion.

I added a new method in `application_helper` (for generating CSS `container_class` in a layout) and a change in `vote_popup.js` to reflect a changed html `id`.

Other than that, it's all about CSS class names, ids and div nesting.

PR #1161, opened 2022-09-30
Fix remove glossary term image

pellaea:
Background: Elsewhere in the code (e.g. in observations) the thumbnail is included in the images, but for whatever reason great pains were taken to exclude the thumbnail from GlossaryTerm#images... only some cases were missed and when I fixed them I didn't realize thumbnails were being deliberately excluded, so I made a mess of things.  I've been trying to clean this up and make GlossaryTerm thumbnails and images behave the same way they do in Observation ever since.

Recently I noticed yet again that there were three glossary terms with thumbnails but no images.  I cleaned these up by hand and went looking at the code yet again.  I found more code that was still behaving as if the thumbnail was not included in the images, so I fixed them.

I still don't understand how this pathological case arises in the wild -- that is how images are removed without the thumbnail being set to null.  But these changes definitely fix bugs, in any case, and would have to be done eventually one way or another.


nimmolo:
(I emailed you about this, but just posting it here so it's easier to find.)

I was checking out the `Image` methods `#reuse_image` and `#remove_images_from_object`, and they seem to be able to work on any model, theoretically. `#reuse_image` uses a `@mode` ivar, and `#remove_images_from_object` uses a `target_class` param.

However, they may be behaving unexpectedly in Glossary Terms, because there is a whole different action for `#reuse_image_for_glossary_term` that does not call `#reuse_image`. And there is _also_ a separate action `#remove_images_for_glossary_term`, but it calls `#remove_images_from_object`. In any case it seems like these separate actions could be DRY'd up to use the same routes with the ivar or param.

In any case it seems that either these "special" attach or remove methods for glossary term images are possibly related to the mystery of how thumbnail assignment/deletion gets confused?

pellaea:
@nimmolo -- I looked into it briefly.  It *looks* like images are added or removed from glossary terms exclusively via GlossaryTerm#add_image and #remove_image, and those two methods are well-tested for all possible cases, seemingly.  While I agree the current state of image controller actions is clearly not ideal, and that they should be DRYed, it doesn't appear to be the source of this problem.  I added one more test to be sure anyway.  But I'm going to have to give up on this and merge it before it gets behind again.

PR #1162, opened 2022-10-01
Restore `matrix_table` until Bootstrap 4 switch

nimmolo:
Sorry for the hassle. It shouldn't be too hard to restore this.  

Simple PR restores all previous uses of `matrix_table` as a layout.


nimmolo:
I believe this is good to go, and will resolve HuaFang's issue

PR #1163, opened 2022-10-02
Allow xyz

pellaea:
A user with an email address at dropkick-ego.xyz tried to sign up and got the response "We grow weary of this. Please go away."  I personally thought it was hilarious, but needless to say, some people may have a different sense of humor than I.  Let's hope the onslaught of bogus accounts from .xyz addresses has stopped.  This removes that block on signups from .xyz accounts.


pellaea:
Woah!  This was branched off of Nathan's visual models branch apparently.  Let me back it out...

PR #1160, opened 2022-09-30
Mega CRUDify (split, to be retired)

nimmolo:
This PR is now obsolete, keeping here for reference and will close soon.

### Controllers CRUDified ###
**Non-CRUD actions mostly namespaced and CRUDified**
- Observations
  _Namings, Namings/Votes, Namings/Suggestions_

### Integration tests added ###
(In progress.) In addition to updating/namespacing all the controller tests, this PR adds new integration tests for changed forms' UI.

PR #1166, opened 2022-10-02
VisualModels and VisualGroups for Machine Learning

mo-nathan:


app/classes/visual_group_data.rb
```diff
@@ -0,0 +1,59 @@
...
+    query.project(attribute(:observation_images, :image_id),
+                  attribute(:visual_group_images, :included),
+                  attribute(:observations, :vote_cache))
+  end
+
+  def add_conditions(name, vote_limit)
+    query.where(tables[:names][:text_name].matches("%#{name}%"))
+    query.where(tables[:observations][:vote_cache].gteq(vote_limit))
+  end
+end
```
nimmolo, line 59: The only potential for using model scopes here that i can spot, is maybe instead of joining to `Name.arel_table` with the conditions, you define `tables[:names]` as `Name.text_name_includes(name)`... passing the `name` var down to `def tables`. You could do the same thing for Observations, but you'd have to add a new scope like `Observation.votes_above(limit)`.

I don't know if that's usable, or what that gets you, but it's nice that scopes are defined in one place — so if the definition of a scope needs to change, it changes everywhere.

mo-nathan: Hmm, tried it, but couldn't get it to fly.  It complained about "no implicit conversion of Symbol into Integer" when it did the `join_table` with :names after I make these changes:
```% git diff
diff --git a/app/classes/visual_group_data.rb b/app/classes/visual_group_data.rb
index 8fe0e2635..bdc9ee5a5 100644
--- a/app/classes/visual_group_data.rb
+++ b/app/classes/visual_group_data.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 class VisualGroupData
+  attr_accessor :name
   attr_accessor :query
 
   def initialize(name, vote_limit)
+    self.name = name
     self.query = tables[:observation_images]
     add_joins
     add_project
@@ -22,7 +24,7 @@ class VisualGroupData
       observation_images: ObservationImage.arel_table,
       visual_group_images: VisualGroupImage.arel_table,
       observations: Observation.arel_table,
-      names: Name.arel_table
+      names: Name.text_name_includes(name) # Name.arel_table
     }
   end
 
@@ -53,7 +55,7 @@ class VisualGroupData
   end
 
   def add_conditions(name, vote_limit)
-    query.where(tables[:names][:text_name].matches("%#{name}%"))
+    # query.where(tables[:names][:text_name].matches("%#{name}%"))
     query.where(tables[:observations][:vote_cache].gteq(vote_limit))
   end
 end
```

nimmolo: Aha. Sorry for not testing that first.
🐥 This guy is telling me the place for the scope might be rather in the conditions, like

```ruby
   def add_conditions(name, vote_limit)
     query.where(Name.text_name_includes(name))
     query.where(tables[:observations][:vote_cache].gteq(vote_limit))
   end
```

I'm standing by to help Jason with the email update. Can test later, or you can try locally.

app/assets/javascripts/application.js
```diff
@@ -9,6 +9,7 @@
 //= require external_link
 //= require image_vote
 //= require image_export
+//= require visual_group_status
```
nimmolo, line 12: Rather than requiring `visual_group_status` on all pages, consider using `javascript_include` to include it only in the templates where it's needed. 

Check the observations/show template for an example (`suggestions` and `vote_popup`) or pivotal/index.

app/views/visual_groups/_form.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= form_with(model: [@visual_model, @visual_group], local: true) do |form| %>
```
nimmolo, line 1: I suggest giving the form an html `id` to help with javascript, and also with integration tests. I'm starting to do that everywhere in my refactoring:

```
form_with(model: [@visual_model, @visual_group], local: true,
          id: "visual_group_form")
```

app/views/visual_groups/edit.html.erb
```diff
@@ -0,0 +1,52 @@
...
+  <% @visual_group.distinct_names.each do |name| %>
+  <%= link_to(name[0],
+              edit_visual_group_path(@visual_group, filter: name[0]))%><br>
+  <% end %>
+</p>
+
+<hr>
+
+<strong><%= :edit_visual_group_filter_options.t %></strong>
+<%= form_with(url: edit_visual_group_path(@visual_group), method: :get) do |form| %>
```
nimmolo, line 28: Here again i'd add an html `id` to help with the JS selector (and Capybara selector)
```suggestion
<%= form_with(url: edit_visual_group_path(@visual_group), method: :get,
              id: "visual_group_filters_form") do |form| %>
```

app/views/visual_groups/_form.html.erb
nimmolo, line 1: I suggest giving the form an html `id` to help with javascript, and also with integration tests. I'm starting to do that everywhere in my refactoring:
```suggestion
<%= form_with(model: [@visual_model, @visual_group], local: true,
              id: "visual_group_form") do |form| %>
```

app/views/visual_groups/edit.html.erb
```diff
@@ -0,0 +1,52 @@
...
+      <%= form.radio_button(:status, a_status, checked: @status == a_status, id: a_status) %>
+      <%= form.label(:"status_#{a_status}", :"visual_group_#{a_status}".t, value: a_status, for: a_status) %>
+    <% end %>
+  </p>
+
+  <p>
+    <div class="actions">
+      <%= form.submit(:edit_visual_group_update_filter.t) %>
+    </div>
+  </p>
```
nimmolo, line 41: I think you could use a simple inline JS, something like this (i'm not sure I have the actual HTML id's right):

```suggestion
  </p>
  <script>
    jQuery("#included, #excluded").on("click", function(e){
    // or jQuery('input[name="status[included]"], input[name="status[excluded]"]').on("change click", function(e){
      jQuery("#filter").val("");
    });
  </script>
```

mo-nathan: Thanks!  That helped.  Needed a bit more tweaking to get it to work, but it seems to be working now.  I also added an anchor to avoid bouncing back to the top on a small screen.  I like @pellaea's idea of disabling the text box as well when "Included" or "Excluded" are selected, but I'll need to figure out how to do that.

app/assets/javascripts/application.js
mo-nathan: Done.

app/views/visual_groups/edit.html.erb
mo-nathan: Now that I've played with it a bit, it seems like it should go even further and do the reload when you click "Included" or "Excluded".  Then the button could be renamed "Review" and be next to the text box.  I'm also thinking this really should be it's own thing and the stuff for editing the name and description should be on it's own page.

mo-nathan: Hit a bug.  If you use the links to mark something as "Included" or "Excluded" under the images and then click "Update Filter" (with idea being that it should remove the images you've provided info for), the "Update Filter" button becomes disabled.  This isn't anything I did consciously so it's a bit of a head scratcher for me.  Any suggestions welcome!

pellaea:
@nimmo, I think there may be two schools of thought here, re: include all
javascripts on all pages, or include only the bits you need on the pages
that need them.  Thanks to automatic compression of assets, it might be
more efficient to include all scripts on all pages.  Then there i only a
single (compressed, concatenated) script that is cached, so browsers can
take advantage of the fact that it is the same on every page.  If you
include a kaleidoscope of varying fragments from each page, there will be
(potentially) multiple requests for each page, and they will all be
different, so the browser cannot take advantage of any caching or whatever.

Your suggestion is a worthwhile optimization to consider, but it should
definitely be tested carefully, because it may (or may not) do more harm
than good.

app/views/visual_groups/edit.html.erb
nimmolo: @mo-nathan  - rails-ujs probably disables the buttons on submit. The `data-disable-with` attribute is automatically added to the submit button. (...and you can change the text of that button while disabled. I also made a `safe_spinner` helper method that you can use to add a spinner after whatever text you use.)

There's a rails-ujs way to re-enable a disabled button with your jQuery snippet, on page load, or on whatever event.
```
setTimeout(function(){
  document.querySelectorAll('[data-disable-with]').forEach(function(element){
    $.rails.enableElement(element);
  });
});
```

nimmolo:
Agree with @pellaea, i'd forgotten about that. 

We have a bunch of page-loaded snippets of JS sprinkled around the existing app, including on templates involved in my upcoming PR, so i may start requiring those into application.js. We also have a bunch of inline js, which doesn't involve a resource load, and is convenient to consult because it's right there, but it kinda clutters the templates, which i'm trying to simplify to make the transition to Bootstrap 4 easier (in terms of scanning the templates for HTML class structure, which must change). Not sure what to do about inline js and may leave them where they are, but if they're moved to assets and required in application.js, as long as they're consistently named and easy to find it should be ok.

app/views/visual_groups/edit.html.erb
mo-nathan: Sorry, I need more help on this.  JavaScript just doesn't make any sense to me.  I don't know how to get it to run that code so it makes any difference.  I just seeing the exact same behavior.

pellaea: Actually, in this case, maybe it would be better to remove the data-disable-with attribute from these buttons?  I wonder if there is an option you can pass into form.submit() that does this?  If there were some way like that to avoid having it disable the button in the first place, that would seem like a far cleaner and simpler solution.

However, in answer, perhaps, to your question about how to run the javascript, you would need to add an event handler to the submit button that runs the setTimeout(...) snippet @nimmolo supplied.  Something like this:

<script>jQuery("#<submit_button_id>").on<Event>(function() { setTimeout(...) })</script>

I forget which event you would want to attach it to, and of course fill in the correct button id.  Also, I'm not sure if you want to run this immediately, or place it all in a document.onLoad() callback to have it run after the page has loaded!  Maybe @nimmolo knows these details off of the top of his head?  Also, don't you have to add a 1 second delay or something to the setTimeout() call?

mo-nathan: Decided to mostly bail on that approach and just go with links rather than radio buttons and provide a reload button.  The "Update Filter" button can still get stuck if you click the links under the images and then click "Update Filter" without change the name, but if you hit reload it's recovers.  Not sure it's worth spending more time on at this point.

app/views/visual_groups/_form.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      <%= form.text_field(:name, size: 40) %>
+      <%= :VISUAL_GROUP.t %>
+    </div>
+
+    <div class="field">
+      <%= form.label :description %><br>
+      <%= form.text_area(:description, cols: 60, rows: 10) %>
+    </div>
+
+    <div class="field">
```
nimmolo, line 30: Suggest adding Bootstrap class "form-group" on these label/field combos (where there is currently "field").

https://getbootstrap.com/docs/3.4/css/#forms

app/views/visual_models/_form.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <div id="error_explanation">
+      <h2><%= pluralize(visual_model.errors.count, "error") %> prohibited this visual_model from being saved:</h2>
+
+      <ul>
+        <% visual_model.errors.each do |error| %>
+          <li><%= error.full_message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
```
nimmolo, line 12: Don't we usually raise flash errors if a form doesn't save? That seems to be how it works on Prefs and New obs.

app/views/visual_models/_form.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+  <div class="field">
+    <%= form.label :name %>
+    <%= form.text_field :name %>
+    <%= :VISUAL_MODEL.t %>
+  </div>
```
nimmolo, line 18: Ditto here with "form-group"

app/views/visual_groups/_form.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      <%= form.label :approved %>
+      <%= form.check_box :approved %>
+    </div>
+  </div>
+
+  <div class="actions">
+    <%= form.submit %>
+  </div>
```
nimmolo, line 38: I forgot if you still want this, but if you want Rails to disable the `submit` button on submit, you do have to pass a `data-disable-with` attribute to the button helper (you can use something like `data-disable-with: "Filtering... #{safe_spinner}"` to get alternate text with a "spinner".) 
`rails-ujs` will disable the button until the form is re-rendered. 

Invalid fields will keep the submit button disabled, though... so, to re-enable the submit button when an input changes, i have added a little snippet on #1160 after line 51 in `assets/javascripts/main.js`. Not sure if you need this here, but i imagine so. 

```javascript
 // Not a great solution, but ok for now.
 jQuery('form :input').on('change', function() {
   var disabled_buttons = $('[data-disable-with]');
   $(disabled_buttons).each(function() {
     $.rails.enableElement(this);
   })
 });
```

app/views/visual_groups/edit.html.erb
nimmolo: @mo-nathan @pellaea  I had forgotten how i'd already dealt with this on #1160.
Made a suggestion in the code review (small addition to `main.js`) that will re-enable the button if any inputs change.

You could also disable `data-disable-with` but I'd much prefer to keep it on and get submit-disabling working properly on the forms - then we don't need that alert telling people not to submit twice, we simply disable it.

pellaea: I realize this is a dead horse, because Nathan's taking a different approach now, but I just want to understand the issue better...

@nimmolo -- Are you saying that we have to explicitly pass an option into the submit button to make it disable when you press it?   Are there cases where we don't want that behavior?  Maybe this is one of those cases?  Or if, say, it is calling an AJAX method, it might make sense to have the AJAX callback(s) re-enable the button, so that even if the user doesn't change an input field they can still press the button again after the first press finishes?  Are we on more or less the same page?  Thanks!

nimmolo: Yep i did say that and sorry — that was sloppy and incorrect advice.

I eventually went back to my PR and review what i learned (more than a month ago)! 
Editing that comment now.

nimmolo: Also - yes, AJAX callbacks could re-enable submit buttons using the same bit of JS.

app/views/visual_groups/_form.html.erb
mo-nathan: That submit button is not the issue I was concerned about.  I've been struggling with the extra form in edit.html.erb for updating the set of image that the user is viewing.  It's mostly working now, but it's still possible to get into a state where the button magically becomes disable for no good reason, but I've changed the way the UI work enough that I think it will be rare.

app/views/visual_models/_form.html.erb
mo-nathan: Did it for new/create.  Haven't dug into edit/update yet.

mo-nathan: Did it for new/create.  Haven't dug into edit/update yet.

mo-nathan: Done.

app/views/visual_groups/_form.html.erb
mo-nathan: Tried this in the edit.html.erb, but still no joy.  I added it verbatim at the beginning of the `inject_javascript_at_end` block.  The bad behavior is, create a visual group (start at http://localhost:3000/visual_models and do the CRUD thing) with the same name as some mushroom in your local database (e.g., "Agaricus campestris").  Goto to the edit Visual Group page.  Click "Include" or "Exclude" under some image.  Then click "Update Filter".  It may work correct the first time, but if you do it again the "Update Filter" button it won't refresh and it gets disabled and remains disabled until you reload the page either in the usual way or by clicking the "Reload" button.  Once it has failed once it will continue to fail.

pellaea: I'll grab the latest code and take a look at this shortly.  I bet it's something really small.  Have you checked if there are any javascript errors in the browser error console?

mo-nathan: I suspect you're right.  I did look for JS errors and didn't see anything obvious.  My JS debugging skills have never been very good and they are very rusty right now.

pellaea: @mo-nathan -- Urk, I don't see nimmo's suggested code snippet anywhere.  Is it possible I don't have your latest commits?

pellaea: @mo-nathan -- Well, I played with it anyway.  I'm at commit 848d888f5c81818d5f36da7387c82926624cb58d, by the way.  And I confirm your behavior.  The first time I ask it to update the filter it works, but subsequent clicks do nothing.  I don't think this has anything to do with javascript.  I think it's the anchor.  Firefox (at least for me) doesn't seem to see fit to actually reload the page when you click on same URL if there's a "#anchor" in it.  I bet if you removed that it would work fine...  Yes, that's correct, I just tried it.

pellaea: Here's what I did, for the record:

--- a/app/views/visual_groups/edit.html.erb
+++ b/app/views/visual_groups/edit.html.erb
@@ -28,7 +28,7 @@
   <hr>
 
 <strong><%= :edit_visual_group_filter_options.t %></strong>
-<%= form_with(url: edit_visual_group_path(@visual_group, anchor: "filter_options"), method: :get,
+<%= form_with(url: edit_visual_group_path(@visual_group), method: :get,
               id: "visual_group_filters_form") do |form| %>
   <%= form.text_field(:filter, value: @filter, size: 60, id: "filter") %> <%= form.submit(:edit_visual_group_update_filter.t) %>

mo-nathan: Nice find!  I played around a bit more and ended up going with your suggestion.  Dumped the JS I had in there since it wasn't actually being used on the page any more.

I didn't commit Nimmo's code.  It was earlier in this thread if you want to see it.

nimmolo: That sounds right. Sorry for the red herrings.

PR #1167, opened 2022-10-02
Remove MO's extra js confirm. `data-confirm` is handled by rails-ujs

nimmolo:
This should be good to deploy after review.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,4 @@
-// = require jquery3
+//= require jquery3
 //= require jquery_extensions
 //= require rails-ujs
 //= require bootstrap
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,11 +1,10 @@
-//= require jquery3
+// = require jquery3
 //= require jquery_extensions
-//= require jquery_ujs
+//= require rails-ujs
 //= require bootstrap
 //= require jstz
 //= require autocomplete
 //= require box_resizer
-//= require confirm
 //= require csrf_token
 //= require external_link
 //= require image_vote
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -686,6 +686,13 @@
     t.boolean "blocked", default: false, null: false
   end
 
+  create_table "visual_groups", charset: "utf8mb3", force: :cascade do |t|
```
mo-nathan, line 689: I don't think this should be part of this PR.

mo-nathan (changes requested):

app/assets/javascripts/application.js
```diff
@@ -1,11 +1,10 @@
-//= require jquery3
+// = require jquery3
```
mo-nathan, line 1: Is this change intentional?

mo-nathan:
I was able to confirm that these changes have the desired effect on my branch.  Pushed the changes to the visual-models branch.

db/schema.rb
nimmolo: Oh... that must have come in via merging master. I must have misunderstood what you were trying to say in the email thread... probably trying to save me the extra step. 

Removing from schema now.

app/assets/javascripts/application.js
nimmolo: Nope! Thank you

db/schema.rb
nimmolo: @mo-nathan When I run `bin/rails db:migrate" this keeps recreating the table in the schema, even though there is no migration in this branch that creates this table. 

I've forgotten how to resolve this situation! do you know offhand?

mo-nathan: It means the table exists in your database, but there's no corresponding migration.  The simplest thing to do is just copy db/schema.rb from master onto your branch and commit it.  To get `rails db:migrate` to work correctly, you can either reload the database or you can just go in to the database console and drop the table.  Then you should be able to run the migrations and schema.rb should not change.

nimmolo: Got it. Doing that now. Thank you!

PR #1174, opened 2022-10-15
Switch ActionMailer to send via `smtp` to GMail, rather than directly from server.

nimmolo:
Joe mentioned the verification emails getting stuck in people’s spam folders despite our burnished DMARC credentials. 

I believe a good solution would be to send the emails out through GMail SMTP rather than directly from our own server. Although that sets up a technically unnecessary extra layer of routing, it will certainly certify the emails as valid, since Google owns the Internet.

I found [this article](https://dev.to/morinoko/sending-emails-in-rails-with-action-mailer-and-gmail-35g4) about how to set up the config. The password, however, should be an “app password” which we would need to create in the GMail interface for `webmaster@mushroomobserver.org`.

[More about Gmail "app passwords".](https://support.google.com/mail/answer/185833?hl=en#zippy=%2Cwhy-you-may-need-an-app-password%2Capp-passwords-revoked-after-password-change)

Maybe this is a good place to try our first use of Rails 6 [credentials in production](https://blog.saeloun.com/2019/10/10/rails-6-adds-support-for-multi-environment-credentials.html#global-credentials).


.gitignore
```diff
@@ -33,3 +33,5 @@ coverage
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
```
pellaea, line 35: Actually, maybe we want master.key to be checked in, now?  I guess it comes down to how "secret" are any development or test passwords?  I would guess there aren't any I care about people getting hold of.  But maybe you can think of something I'm not?

nimmolo: I think there may be dev secrets that we don't want checked in, especially if i resume work on the GraphQL api, because in dev mode we will conceivably need to test giving viable credentials to some API or another. 

I'm also wary of automated code scanners, especially after that automated alert this morning, that i was "exposing" smtp credentials. I don't think we want bots scouring the repo and finding credentially-looking things or master keys.

`master.key` is automatically added to .gitignore by Rails when you first set up credentials. My motive for starting to use them is just to keep moving our setup closer to a Rails default, so potential devs don't need to spend time deciphering the basics of it.

nimmolo:
What i did at Gmail/Google Workspace was 
- configure an SMTP relay at the workspace level, on [admin.google.com](http://admin.google.com/) . Private Gmail addresses don’t need to set up a relay, but workspace emails (like ours @mushroomobserver.org) do. Bonus: this allows us to potentially send to larger lists, although we should consult Gmail docs before doing, so we don’t get flagged by Google.)
- activate 2FA for the webmaster@ account. 2FA is mandatory for an app password, and considering that i couldn’t get into the account anyway without being authorized by Nathan, it didn’t seem like much of a change. NOTE: you can add your own phone number as a verification method! 
- make an app password for webmaster@ (and news@ and no-reply@, just in case)
Notes:
- The production ActionMailer config uses Rails 6 production-only credentials, with their own `production.key` (analagous to the `master.key`, that serves as a default for other environments). `production.key` currently exists only on the production server and on my local env - feel free to copy it locally. I chucked in some other credentials from `consts-site.rb` and `gmaps_api` into the encrypted `production.yml.enc` file, but am not using them anywhere, we can test them separately.
- DKIM was correctly set up (by you!) for Gmail. The mystery we didn’t know was, in order to be usable by Ruby mail, we would have needed to add a DKIM gem and configure it, and possibly add a different, non-Google DKIM DNS record. (That’s because DKIM is a header added to all emails, that allows the recipient to check a key in the header against a TXT record on the domain DNS. Gmail's DKIM record seems to have a custom syntax.)
- The Google tech said that even after this is merged and we’re sending via the SMTP relay, we probably still need to request removal from Hotmail’s blocklist.

nimmolo:
Further notes:

`AccountMailer` now renamed `ApplicationMailer`,  re: Rubocop convention.
All mailer classes now renamed from `*Email` > `*Mailer` 

To enable sending from two different smtp accounts, `ApplicationMailer` now loads two different SMTP configs from the production credentials, and these credentials are applied in each Mailer class via the `after_action` hook.

Production credentials are now to be found on the production server, under `config/credentials`. The `production.key` lives there too.

An entire hash of `action_mailer.smtp_settings` for each email account we send from, `webmaster@` and `news@`, is now stored in the production credentials.

.gitignore
mo-nathan: Also we should probably be running `git secret` in general and it would get grumpy about that.  Doesn't seem like a big deal to share around the master.key outside of git as needed.

pellaea:
Fair enough.  Disregard my suggestion!

PR #1182, opened 2022-10-22
Added a bunch of additional self-governing overseas territories

pellaea:
to the list of countries, even though their status as independent countries is somewhat ambiguous.

Any objections?


mo-nathan:
Looks reasonable to me.

pellaea:
Thanks, Nathan.

PR #1194, opened 2022-11-08
Added test for valid email address

pellaea:
And have it silently not send mail if the "to" address is invalid.  This became a problem when we switched to using google as an SMTP proxy server.


nimmolo:
J'approuve. 

As a bonus, it seems to introduce a callback point where we could conceivably insert a "whoa, reverify this user" action.

pellaea:
Yeah, that's what I was thinking, too.  There's probably a gem that does a
better job, but I was too lazy!

PR #1195, opened 2022-11-08
Oops, didn't actually stop attempted delivery of undeliverable emails!

pellaea:


nimmolo:
Huh. You getting more invalid emails? Or just the test...?

PR #1198, opened 2022-11-16
Add two new `User` validations: email format and content_filter[:region]

nimmolo:
This adds two new validators for the `User` model:
- `email` format `address@service.tld`
checks the `email` format against the same regex used in `ApplicationMailer`. This doesn't ensure that someone has entered a valid working email address, but at least it catches typos and deliberate non-addresses. (The PR also updates Jason's recently-added regex to use the Ruby native library `URI::MailTo` regex.)

- `region` format `maybe place, province, country`
checks if the content_filter text string `region` is a valid MO region, i.e. if MO has any locations in that region. It uses the `Location.in_region` scope to check if the string matches the end of any existing Location strings. If not, it won't validate.

### Dev note ###
**I discovered a major Rails gotcha with adding new validations to existing records**: if you add a normal Rails model attribute validation _without conditions_, and if our existing records have _newly-invalid_ data (ours certainly do), users and admins will not be able to update (correct) the invalid fields. Not even in the console! 

**Solution**: You **must** add an `if` `new_record?` or `email_changed?` condition to the validator, so it works as expected and doesn't block updates.

#### Further solutions for users who deliberately enter invalid email addresses after signup ####
Jason and I are working out other ideas to help satisfy people who don't want any emails, and to alert them to provide a valid and verifiable email address.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -15,13 +15,9 @@ ActionMailer::Base.raise_delivery_errors = false
 
 #  Base class for mailers for each type of email
 class ApplicationMailer < ActionMailer::Base
-  # This more or less follows RFC-5321 rules, minus the ridiculous quoting.
-  DOT_ATOM = %r{[0-9A-Za-z_!$&*\-=\\^`|~#%‘+/?{}]+}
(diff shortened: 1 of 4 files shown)


mo-nathan (approved):
Look good!

PR #1203, opened 2022-11-25
Add yet another email preference to opt out of all emails.

pellaea:
The purpose of this is to let admins essentially disable emails to users who have invalid email addresses.  One of the requirements for "responsible" emailers is that they not repeatedly try to send mail to nonexistent accounts.  But out users will occasionally deliberately change their address to a bogus address in order to prevent us from sending them emails.  (Even though they could simply uncheck all the email boxes on the preference form.)  And in some cases email accounts simply get closed out and no longer work.  In both cases I would like, as an admin, to be able to sudo over to that user and check the "opt out of all emails" preference to stop MO from trying to send mail to bogus addresses.

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -305,6 +305,7 @@ class AccountController < ApplicationController
       [:locale, :string],
       [:location_format, :enum],
       [:login, :string],
+      [:no_emails, :boolean],
       [:notes_template, :string],
       [:theme, :string],
       [:thumbnail_maps, :boolean],
--- a/app/controllers/emails_controller.rb
+++ b/app/controllers/emails_controller.rb
@@ -8,7 +8,8 @@ class EmailsController < ApplicationController
 
   def features
     if in_admin_mode?
-      @users = User.where("email_general_feature=1 && verified is not null")
+      @users = User.where(email_general_feature: true, no_emails: false).
+               where.not(verified: nil)
       if request.method == "POST"
         @users.each do |user|
           QueuedEmail::Feature.create_email(user,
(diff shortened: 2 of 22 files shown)


pellaea:
The only failure here is a pre-existing rubocop issue with one of the notify methods being too long, and which I do no intend to fix in this PR.

I believe this is ready to go.  Does anyone have time or sufficient interest to review it?

nimmolo:
💐 Hooray for this. It will mean a bit of adaptation for my CRUD PR, but I'm very in favor.

nimmolo:
I'm sure you noticed, but CI is reporting a bunch of language tag test failures

```
= FAIL AccountControllerTest#test_edit_prefs (40.49s)
        Language tag(s) are missing..
        --- expected
        +++ actual
        @@ -1 +1 @@
        -[]
        +[:prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails, :prefs_no_emails]
        test/test_helper.rb:140:in `teardown'
```

nimmolo:
It could be smart to add a line in the `ApplicationMailer` that if a user's "email" doesn't even pass the formatting test, the app adds the "no emails" attribute to that user, right there. That would prevent sending emails a bit higher up in the chain.

pellaea:
You know, @nimmolo, there might be an even better solution.  Since it's now impossible to enter a *new* value into User#email that's invalid (thanks to you), all we need to do is add it to the *migration*.  A one time process.  Hell, I don't even have to do that.  I might just do it by hand after deploying.  And I could cull the millions of bogus unverified users while I'm at it.  It's been a long time, I think, since anyone's done that.

Thanks for looking at it!  As usual for me, hardly award-winning work, but it gets the job done.  Without making anything worse, I think.  And fixing one bug into the bargain!  (Existing code checks if the *sender* allows personal emails, not if the *receiver* allows them in EmailController#ask_user_question.  Even the test tests the wrong thing!  Oops!)

PR #1206, opened 2022-11-26
Have account controller strip leading and trailing spaces from all inputs

pellaea:


pellaea:
As before, the failure is a pre-existing rubocop metric problem that is outside of the scope of this PR to fix.

app/controllers/account_controller.rb
```diff
@@ -213,8 +205,8 @@ def login_get
 
   def login_post
     user_params = params[:user] || {}
-    @login = user_params[:login].to_s
-    @password = user_params[:password].to_s
+    @login = user_params[:login].to_s.strip
+    @password = user_params[:password].to_s.strip
     @remember = user_params[:remember_me] == "1"
     user = User.authenticate(login: @login, password: @password)
     user ||= User.authenticate(login: @login, password: @password.strip)
```
nimmolo, line 212: @pellaea - Does your new `strip` on the password param make this second `authenticate` request redundant?

Asking because i'm adapting it to the CRUD branch.
```ruby
@login = user_params[:login].to_s.strip
@password = user_params[:password].to_s.strip
@remember = user_params[:remember_me] == "1"
user = User.authenticate(login: @login, password: @password)
user ||= User.authenticate(login: @login, password: @password.strip)
```

pellaea: Oh, would you look at that!  Yes, good catch.

PR #1211, opened 2022-12-07
Restore `AbstractModel` method `show_controller` the way it was

nimmolo:
This reverts a reversion. Took me a year to learn again the value of what Joe committed two years ago: 

You can't pass a symbol as a controller name, assuming that Rails will use that controller. 

`controller: :controller_name` <= SOMETIMES DOES NOT WORK!
`controller: "/controller_name"` <= ALWAYS WORKS 

Now that we have namespaced controllers, with many more on the way, generalized methods like `show_controller` MUST string-interpolate the controller name after a leading forward slash, in order to explicitly specify a "top level" controller. Without the forward slash, requests from a nested controller will assume the same class nesting as the originating call, e.g. nonsense like 

`controller: "/herbaria/name", action: :show_name` 

It's very confusing to debug, and almost never what you want. 

This is counter-intuitive and a real Rails quirk. **Symbols** seem like a Ruby way of disambiguating a parameter, and you might assume that the use of a symbol would always specify a top-level controller. 

Not so.


nimmolo:
Also fixes an obsolete Herbarium path in a mailer.

pellaea:
Looks good to me!

PR #1213, opened 2022-12-07
`AuthorsController` CRUDify 

nimmolo:
I'm splitting up the giant CRUD controllers PR into chunks, starting with the smallest.

They'll still be big, just not as big. Hopefully easier to review.

This one's dainty, though. Truly manageable!


nimmolo:
This is ready to go, will merge today if no objections. 
Next PR is larger, requires this to merge first, and needs more scrutiny.

app/helpers/application_helper.rb
```diff
@@ -133,52 +137,72 @@ def link_prev(object)
...
   #             path: herbaria_merges_path(that: @merge.id,this: herbarium.id),
-  #             confirm: :are_you_sure.t)
-  def post_button(name:, path:, confirm: nil)
-    data = confirm ? { confirm: confirm } : nil
-    button_to(name, path, method: :post, class: "text-info", data: data)
+  #             data: { confirm: :are_you_sure.t })
+  def post_button(name:, path:, **args)
+    html_options = {
+      method: :post,
+      class: "text-info"
```
pellaea, line 169: This is such a minor thing, and totally not important, but I wonder if maybe you have seen this problem solved in your perusal of forums and such...  In an ideal world, this would not _replace_ the CSS class, but add to it, right?  I bet this is a common thing.  Basically where you want a merge but where it concatenates the new value if it's not already in there.  I could almost see it being worthwhile to write little Hash extensions which do this:

html_args = args.overwrite(method: :post).concat(class: "text-info")

So totally not important here because of course none of your callers have provided the "class" argument anyway! :)  But just thought I'd throw that out there in case it resonated as something worth doing some day...

pellaea:
I'm sorry, Nimmo, but I find reviewing PRs where files move around nearly impossible.  I cannot follow all this CRUD stuff, moving actions into sub-controllers and all.  It makes my brain explode.  I'm so glad you've (apparently?) got it down.  I didn't notice anything questionable, FWIW.  If it passes the tests...

app/helpers/application_helper.rb
nimmolo: YES! I was puzzling over this. Thank you for raising it.
Sometimes, though, I do want to replace the class, not concatenate. So I left it like this for now. "text-info" in Bootstrap enforces a text color, and I figured i might want to change the color in some places. Or make it a button, or whatever.

pellaea: Exactly, my suggestion would let you choose how to treat each argument, cleanly, transparently, and conveniently.  (I always thought the args = { override: this }.merge(input_args) mechanism was clunky.)

nimmolo:
Thanks for checking it out. Yep, it's hard to follow. 

Joe mentioned this too, and I was examining a method yesterday for maintaining git line blames while splitting up files, but it doesn't help when methods move from one file to another.

Still, i hope it helps breaking it up this way. The next PR is ready, and much bigger, but at least it's not the whole giant one. 

Can't think of what else to do. They all kind of build on each other, and breaking these up is consequently very labor-intensive.

pellaea:
I appreciate the effort!  It would be awesome to connect those files.  I
assume you are doing "git mv" when you rename files, right?  (As opposed to
"git rm" the old one, then "git add" the new one.)  I see to remember
trying that one time as a test and it did nothing for the output of "git
diff".

Still, in all, it is probably "better" to have the smaller PRs, from the
perspective of making it easier to track down which PR is to blame in case
something goes wrong.

Someday, eventually, you will be able to work on "normal" PRs that don't
completely rewrite or move large portions of code!  Nice little things
where you just add a little here or there, make a few small changes, nice
and easy and fun. :)

PR #1214, opened 2022-12-07
Send author requests via `QueuedEmail`

nimmolo:
Just what it says.

### DIFF
--- a/app/models/queued_email.rb
+++ b/app/models/queued_email.rb
@@ -129,18 +129,6 @@ class QueuedEmail < AbstractModel
   self.inheritance_column = "flavor"
   self.store_full_sti_class = true
 
-  # Ensure that all the subclasses get loaded.  Problem is some subclasses have
-  # the same name as toplevel classes, e.g., QueuedEmail::Comment.  Thus the
-  # constant QueuedEmail::Comment will already be "defined" if Comment is
-  # loaded, so it won't know to try to load the one in QueuedEmail.  This way,
-  # soon as QueuedEmail is defined, we know that all subclasses are also
-  # properly defined, and we no longer have to rely on autoloading.
-  #
-  Dir["#{::Rails.root}/app/models/queued_email/*.rb"].each do |file|
-    match = /(\w+)\.rb$/.match(file)
-    require("queued_email/#{match[1]}") if match
-  end
-
   # ----------------------------
   # :section: General methods.
   # ----------------------------
--- a/app/models/queued_email/author_request.rb
+++ b/app/models/queued_email/author_request.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Request to be the author of an object
+# Request to be the author of an object, usually a Description
 class QueuedEmail::AuthorRequest < QueuedEmail
   def obj_id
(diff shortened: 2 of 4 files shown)


app/models/queued_email.rb
```diff
@@ -129,18 +129,6 @@ class QueuedEmail < AbstractModel
...
-  # loaded, so it won't know to try to load the one in QueuedEmail.  This way,
-  # soon as QueuedEmail is defined, we know that all subclasses are also
-  # properly defined, and we no longer have to rely on autoloading.
-  #
-  Dir["#{::Rails.root}/app/models/queued_email/*.rb"].each do |file|
-    match = /(\w+)\.rb$/.match(file)
-    require("queued_email/#{match[1]}") if match
-  end
-
   # ----------------------------
```
nimmolo, line 132: This is not necessary because of zeitwerk.

nimmolo:
Note that this is only a PR off the `AuthorsController` PR, merging it will only combine this with that PR.

app/models/queued_email.rb
pellaea: Nice!

app/models/queued_email/author_request.rb
```diff
@@ -0,0 +1,41 @@
...
+  def self.create_email(sender, recipient, object, subject, message)
+    result = create(sender, recipient)
+    raise("Missing object!") unless object
+
+    result.add_integer(:obj_id, object.id)
+    result.add_string(:obj_type, object.type_tag.to_s)
+    result.add_string(:subject, subject)
+    result.set_note(message)
+    result.finish
+    result
```
pellaea, line 30: Oooh!  I just thought of something.  What if QueuedEmail#add_xxx, set_note, finish all returned self?  Then these could all be chained:

create().add_xxx(xxx).add_xxx(xxx).set_note(xxx).finish

That would be so slick! :)

pellaea:
Wow, that was easy.  Maybe I'll do that with some of the other non-QueuedEmail mail, too.  Once all of our email uses this mechanism, then we can contemplate alternative delivery methods, something that's been near the top of my personal priority list for ages.

Looks good to me.  Thanks for doing this and demonstrating how easy it is to migrate direct emails to QueuedEmails.

app/models/queued_email/author_request.rb
nimmolo: The `return unless object` currently interrupts that chain, should it come before?

nimmolo: Trying it.  Like so?
```
    raise("Missing object!") unless object

    create(sender, recipient).
      add_integer(:obj_id, object.id).
      add_string(:obj_type, object.type_tag.to_s).
      add_string(:subject, subject).
      set_note(message).finish
```

pellaea:
Yes, that's _exactly_ what I mean.  It's silly syntactical sugar, but what
can I say, I'm just in that mood right now. :)  Please feel free to ignore!

nimmolo:
`add_integer` actually throws an error chained immediately after `create` - I think it has to wait for the minimal record to be instantiated. I'll split it into create/chain

PR #1217, opened 2022-12-08
`AccountController` & `AdminController` CRUDify

nimmolo:
### Controllers CRUDified ###
- Account
  _APIKeys, Login, Preferences, Profile, Verifications_
- Admin
  _Session, Emails/Feature, AddUserToGroup, Banner, BlockedIps, Donations, Users_

This PR moves relevant methods from `Account`, `Email`, and `Users` under the `Admin` namespace.

### Integration tests added ###


app/assets/stylesheets/mo/_utilities.scss
```diff
@@ -114,41 +114,41 @@
 }
 
 .mx-1 {
-  margin-left: 0.25em !important;
-  margin-right: 0.25em !important;
+  margin-left: 0.25rem !important;
+  margin-right: 0.25rem !important;
```
pellaea, line 118: Neat.  Are "rem" units new?  Looks like most posts about it are from 2022.

app/controllers/account/verifications_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+      :create,
+      :reverify,
+      :resend_email
+    ]
+    before_action :disable_link_prefetching, except: [
+      :new,
+      :create
+    ]
+
+    # Regular signup verifications hit this page only (via GET).
```
pellaea, line 16: Thank you for the comments!  I would have no clue what the difference between account/login/new and account/login/create would be otherwise!

app/controllers/admin/add_user_to_group_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+module Admin
+  class AddUserToGroupController < ApplicationController
+    include Admin::RestrictAccessToAdminMode
+
+    before_action :login_required
+
+    def new; end
+
+    def create
```
pellaea, line 11: Okay, I think I'm starting to get it.  If I ignore "new" and "create" and read them instead as "GET" and "POST" and then interpret "controller" as "action", then this all maps pretty well from the old system.  It gets a _little_ confusing when there are non-crud actions mixed in (like /account/verifications/resend).  I'm not recommending this, but would another approach for things like that be to do yet another layer of namespacing?  e.g., /account/verifications/resend_controller?  (Just for the purpose of understanding and getting used to this way of doing routing...)

app/controllers/admin/restrict_access_to_admin_mode.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Defines admin access level used across all the admin controllers.
+# It extends the "login_required" filter defined in classes/login_system
+module Admin
+  module RestrictAccessToAdminMode
+    def authorize?(_user)
```
pellaea, line 7: Clever!

app/controllers/admin/switch_users_controller.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Admin
+  class SwitchUsersController < ApplicationController
+    before_action :login_required
```
pellaea, line 5: Just thinking out loud, not making a recommendation...  I think I understand that you can get away with the RestrictAccessToAdminMode mechanism because your new controllers are all so small that each controller is more or less guaranteed to be entirely admin-required or not.  Okay, so then what about login_required?  How often are the new controllers divided where some methods are open access and others require login?  Or is it just more complicated when it comes to login, e.g. for login and verification controllers?

app/controllers/admin/turn_off_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Admin
+  class TurnOffController < ApplicationController
+    before_action :login_required
+
+    def show
```
pellaea, line 7: I gotta ask...  Why two controllers for this?  Why not a single controller that takes a parameter?  I mean, sure, they're tiny, so who cares!  I was just wondering what your thought process was?

app/controllers/admin/users_controller.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+module Admin
+  class UsersController < ApplicationController
+    before_action :login_required
+
```
pellaea, line 6: Why not just use require RestrictAccessToAdminMode?  Just because the redirects are different if denied?

app/controllers/admin_controller.rb
```diff
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 
+#  ==== Admin utilities
+#  test_flash_redirection::      <tt>(R . .)</tt>
+
 class AdminController < ApplicationController
   before_action :login_required
```
pellaea, line 7: Why not just in-line the RestrictAccesstoAdminMode module here so all admin (sub)controllers inherit it by default?

app/extensions/string_extensions.rb
```diff
@@ -517,6 +517,45 @@ def gsub_html_special_chars
       html_safe # rubocop:disable Rails/OutputSafety
   end
 
+  # For integration test comparisons:
+  # Render special encoded characters as they appear in HTML
+  def render_html
```
pellaea, line 522: Would it be too annoying for me to request a more descriptive name? Without reading this method I would have no clue what "blah".render_html does.  Hmmm.  Okay, the really obvious choice ("unescape_html", because, well, that's _exactly_ what the method does) would also be inscrutable to me, most likely.  String and character encoding is always a pain in the butt no matter how you dress it up... :(

app/controllers/admin/add_user_to_group_controller.rb
nimmolo: yes, exactly — in fact i did stuff like that on the next PR for Observations/Namings/Votes. 

With the routes for that kind of nesting, you can configure the nested `resource` with `shallow: true` so that the route only gives as many levels as necessary to specify the right record — also, so you're not trying to grab and pass around an ID at each level of nesting. That was a big conceptual hurdle on my attempt at namespacing NameDescriptions, for example, i thought i had to make all these methods pass a `parent_id`. Now that i've understood this, those should be easier to deal with.

`verifications/resend` was such a simple action it didn't seem worth it but that's the idea.

app/extensions/string_extensions.rb
```diff
@@ -517,6 +517,45 @@ def gsub_html_special_chars
...
+    CGI.unescapeHTML(self)
+  end
+
+  # Insert a line break between the scientific name and the author
+  # (for styling taxonomic names legibly)
+  def break_name
```
pellaea, line 528: Did you intend for these changes to go out with this PR?  I would, in fact, like to see what this looks like, but it seems to be entirely unrelated to CRUDifying the account and admin controllers.

Totally unimportant.  Let's just say the purpose of this comment is to bring these changes to the attention of other potential revewiers! :)

app/controllers/admin/turn_off_controller.rb
nimmolo: I thiiiink i was just trying to preserve a route name that made sense, and i couldn't think of a general word for "turn" that would work for both "on" and "off"~! But if you can...

app/views/account/login/login2021.html.erb
```diff
@@ -0,0 +1,34 @@
+<%
```
pellaea, line 1: Did you intend to delete this when you were done with the PR?

app/extensions/string_extensions.rb
nimmolo: Not too annoying. Yes — i considered `unescape_html` but i can never keep straight what escaping and unescaping mean, these are truly inscrutable words that seem to sometimes mean exactly the opposite of what i think. I chose `render_html` thinking more like, "render this html code `&8931;` in the page like it will be shown in the browser." 

That method was a pain in the butt to figure out, btw.

app/extensions/string_extensions.rb
nimmolo: Right. These are for the Observations show view, namings table, where the namings are kind of hard to read on a phone. Breaking the name and putting the author in slightly smaller text made them a lot easier to scan. 

I also would like to use these two methods in the handsome "matrix box" update i've been sitting on for nearly 3 years, but that is still in traffic waiting behind controller refactors and Bootstrap 4.

app/views/account/login/login2021.html.erb
nimmolo: Ah, yes. Let's delete. I didn't know the origin or need for this file.

config/initializers/inflections.rb
```diff
@@ -30,4 +30,5 @@
   inflect.acronym("TSV")
   inflect.irregular("bonus", "bonuses")
   inflect.irregular("info", "info")
+  inflect.irregular("account", "account")
```
pellaea, line 33: I presume this is to avoid having to change a thousand account controller references to "accounts"... which sounds wrong anyway, admittedly?  There's just something fishy about summarily deciding that the plural of "account" is no longer "accounts"!!  If we ever introduced the concept have a single user being linked with multiple accounts this could get _really_ confusing! :)

app/controllers/admin_controller.rb
nimmolo: That was my original idea. Tried it - I think something in the `DonationsController` does not work with `RestrictAccesstoAdminMode`. I tried it in each controller independently.

app/controllers/admin/add_user_to_group_controller.rb
pellaea: Understood, thanks!

config/initializers/inflections.rb
nimmolo: Good point. Lemme research what else i could do to keep the controller name singular, that isn't too verbose.

app/controllers/admin/turn_off_controller.rb
pellaea: Yeah, "turn" is a bit inscrutable isn't it! :)  But maybe it could just be a GET to /admin?  No worries.

Out of curiosity, if we need to tell Rails to pluralize "account" as "account", why don't we need to do the same with "admin"?

app/controllers/admin/switch_users_controller.rb
nimmolo: `RestrictAccessToAdminMode` basically just changes two parameters of `login_required`. 
`login_required` still has to be a `before_action` on the controller, for the mechanism to work. 

But that might not be the nature of your question!

app/controllers/admin/users_controller.rb
pellaea: Totally doesn't matter.  In fact, I think the user index is not paginated(?) and very expensive (and utterly useless!) so I've been meaning to change it to redirect to / anyway!

If this redirect is the only thing standing in the way of significant simplification, then toss it!

As I see it, and maybe I'm missing something still, we could rip out all of the RestrictAccessToAdminMode stuff and put it in admin_controller.rb to be shared automatically with all admin subcontrollers.  One fewer file, and more standardized admin behavior, and less chance of forgetting to include "require RestrictAccessToAdminMode" in any new admin controllers.

app/controllers/admin_controller.rb
nimmolo: Oh- I remember. `DonationsController` has different redirects, if you're not an admin, than `RestrictAccesstoAdminMode`.

pellaea: Looks like /admin/donations has the same redirect problem, and again, it would be totally fine to have it redirect to / instead of /support/donate.

app/extensions/string_extensions.rb
pellaea: I believe you.  Understood.

app/assets/stylesheets/mo/_utilities.scss
nimmolo: Not new. They were the [hot new thing in 2012-ish](https://caniuse.com/?search=rem), but there are other new units turning heads these days. 

These style classes are all cribbed directly from Bootstrap-4, with the idea there would be minimal adjustment once we're on that branch. They're also just so freakin handy.

pellaea:
I don't have any substantive criticism, so as far as I'm concerned, this is ready to go!

app/controllers/admin/switch_users_controller.rb
pellaea: Oh, good point.  There is probably some magic you can do with modules to have them "run a class method" in the requirer each time the module is required.  That is, have it add a login_required before filter in the requiring controller.  If it is important to anyone (hah!) it can be done just fine in a future PR. :)

app/controllers/admin/turn_off_controller.rb
nimmolo: re: `Admins` — No idea!
re: turn_on/turn_off = get to `AdminController` with params, hmmm. Yes.... Lemme try that.

app/controllers/admin_controller.rb
nimmolo: Ohhh, yeah. The blocker for having `RestrictAccesstoAdminMode` everywhere in Admin is... `TurnAdminOn`. You can't restrict access to that! 

That's why I ended up leaving `Admin::UsersController` and `Admin::DonationsController` with those polite custom tailored redirects. 

One solution. I could do a separate top level controller that doesn't have restricted access, called `AdminAccessController`. Then you'd hit the show action of that controller at `/admin-access/` with your `turn_on`, `turn_off`, or `drop_out` param. Sound good?

Let me know and I will reconfigure this. I like the potential simplification of the `AdminController`.

nimmolo: Am i correct in thinking that the only "turn_admin_on" button is in the top bar [A] button ?

And that the sidebar only ever shows a "turn_admin_off" link (to admins)?

nimmolo: Oh hey, i've got a great name for this controller. `AdminModeController`

nimmolo: Okay, one final item. The `AdminController` method `test_flash_redirection` cannot require admin access. Where should I put that...?

pellaea: re: turn_admin_on require nonadmin access -- of course!  thanks.

re: admin_mode_controller -- excellent!  Good job sticking with it until the solution appeared!

re: test_flash_redirection -- test controller??  Oooh!  There would be a test_controller_test.rb :)

nimmolo: Hahaha `test_controller_test.rb` ok

One more action requires special handling... `switch_users`. 
It's got an extra OR condition governing access to the form, `session[:real_user_id].blank?`

Idea: I can put the `switch_user` form in the `AdminModeController` (as `edit`?) to keep the conditions of `AdminController` the same.

config/initializers/inflections.rb
nimmolo: Inflection removed, turns out it only mattered in routes. 

The workaround is
```ruby
resource :account, controller: "account" do
  stuff
end
```

app/controllers/admin_controller.rb
pellaea: If that works, sounds good to me.  creative.

config/initializers/inflections.rb
pellaea: Ah!  That _is_ much better.  Hope that feels worthwhile to you, too.

app/controllers/admin_controller.rb
nimmolo: It turns out that controllers namespaced under Admin like `Admin::DonationsController` do not automatically inherit anything from `AdminController` unless they explicitly inherit from that controller class, instead of `ApplicationController`. (of course!) 

So I can move the admin mode controller back under `Admin::ModeController` and just have it inherit from `ApplicationController` instead of `AdminController`. 'doh

pellaea: Haha!  I didn't think of that either!

nimmolo: ...But i'm not going to move it. It's fine as `AdminModeController`

pellaea: That's great.  Thanks!

nimmolo:
All clear. Going to merge now if no objection.

PR #1218, opened 2022-12-09
`CollectionNumbersController` `HerbariumRecordsController` CRUDify

nimmolo:
### Controllers CRUDified ###
**Non-CRUD actions namespaced and CRUDified**
- CollectionNumbers
  _RemoveObservations_
- HerbariumRecords
   _RemoveObservations_

### Integration test coverage ###
`HerbariumCuratorTest`


nimmolo:
I believe this is ready to merge.

app/controllers/application_controller.rb
```diff
@@ -282,6 +282,15 @@ def redirect_to(*args)
     end
   end
 
+  # Objects that belong to a single observation:
+  def redirect_to_observation_or_object(obj)
```
pellaea, line 286: I know we talked about this earlier, so I want to focus on only the most important part -- is there a way we can better expose that this method relies on @back_object?

I see it is used 7 times ... 7? why an odd number? strange! ... in herbarium_records_controller and collection_number_controller.  That's not that many.  How about explicitly passing @back_object in?

app/controllers/collection_numbers_controller.rb
```diff
@@ -200,7 +198,27 @@ def post_edit_collection_number
       @collection_number.destroy
       @collection_number = @other_number
     end
-    redirect_to_observation_or_collection_number
+
+    redirect_to_observation_or_object(@collection_number)
+  end
+
+  def check_for_form_errors?
```
pellaea, line 205: Super minor thing, but I thought the recommended convention for method names ending in "?" and "!" was:

? = just checks a condition and returns boolean, no side-effects

! = does whatever, like a normal method without the exclamation, but raises an error if something's wrong (think of the difference between ActiveRecord.create vs. create!, one silently fails, the other raises an error if it fails)

There are probably other additional cases, but these are the important ones here, because while, yes, this returns a boolean, it actually _does_ something if there are form errors, and sort of along the lines of raising an error, it redirects and that means all further processing should stop in this action.

So I would have expected this to be called "make_sure_there_are_no_form_errors!".  Or even "check_for_form_errors!" to keep the change minimal.

Another alternative would be to split this method up and call it like this:

reload_form && return if any_form_errors?

Wait, but again, it's not just checking for errors it's rendering an (appropriate) error message if there is one.  That should be made clear in the name.

reload_form && return if check_form_form_errors!

or

reload_form && return if flash_error_if_form_has_errors

or

flash_error_and_reload_if_form_has_errors

Yikes!  But at least it's now totally transparent what it does.  This is why the gods invented copy and paste. :)

app/controllers/herbarium_records_controller.rb
```diff
@@ -220,7 +229,25 @@ def post_edit_herbarium_record
       return
     end
 
-    redirect_to_observation_or_herbarium_record
+    redirect_to_observation_or_object(@herbarium_record)
+  end
+
+  def check_for_form_errors?
```
pellaea, line 235: Same quibbling little comment about the "?"

app/controllers/herbarium_records_controller.rb
```diff
@@ -256,7 +283,7 @@ def normalize_parameters
     @herbarium_record.notes = @herbarium_record.notes.to_s.strip
   end
 
-  def validate_herbarium_name!
+  def validate_herbarium_name! # rubocop:disable Metrics/AbcSize
```
pellaea, line 286: Good on you!  Fight the Man! :)

app/models/observation.rb
```diff
@@ -314,6 +314,10 @@ class Observation < AbstractModel
         ->(name) { where(name: Name.text_name_includes(name)) }
   scope :by_user,
         ->(user) { where(user: user) }
+  scope :mappable,
+        -> { where.not(location: nil).or(where.not(lat: nil)) }
+  scope :unmappable,
+        -> { where(location: nil).and(where(lat: nil)) }
```
pellaea, line 320: Is this sort of like pork in a congressional bill?  Throwing a bone to a special interests group? :)

app/controllers/application_controller.rb
nimmolo: Yeah it's kind of a switch, isn't it? It's a weird method.

Should it be called this and have a safety back to "/"
```ruby
redirect_to_back_object_or_object(back_obj, obj)
  if back_obj
    redirect_with_query(back_obj.show_link_args)
  elsif obj
    redirect_with_query(obj.index_link_args)
  else
    redirect_with_query("/")
  end
end
```

app/controllers/collection_numbers_controller.rb
nimmolo: I never knew what those method punctuations really implied. I'll change these.

app/controllers/herbarium_records_controller.rb
nimmolo: I've got those in a salt shaker right here on my desk. 
But i am breaking down a lot of methods, too.

app/controllers/application_controller.rb
nimmolo: Seven times because there's an extra check for Herbarium permission in the HR controller

app/controllers/collection_numbers_controller.rb
nimmolo: Revision up. Let me know if that's not right

app/controllers/application_controller.rb
pellaea: Yes, that works so much better for me, thank you!

app/controllers/collection_numbers_controller.rb
pellaea: Yes, so much better for me, thank you!

app/models/observation.rb
pellaea: :)

pellaea:
I did my level best.  I tried to at least put my eyes on every change, but I'm sure I didn't really see most of this long PR.

I appreciate your assertions on the fixtures making it clear what the test relies on.  I guess in some cases at least we used to do that in comments, but I doubt I was doing it everywhere.  That should help with understanding those tests better in the future.

I haven't seen any error reports (automated or otherwise) from the last PR.  When do you think it will be safe to push out this one?  Is one day long enough between monster PRs??

nimmolo:
Hopefully i didnt actually change anything, just move code around. Can you give an example of my "assertions on the fixtures making it clear what the test relies on"? If i'm doing something helpful i want to keep doing it. I feel like I'm still pretty random. With the Account controller, i guess we should hear about it within a couple days if anything's off. Plus, it's covered by integration tests. Maybe not tonight but by tomorrow i feel like it should be safe.  ?Today i finished Comments, in draft PRs because I think it depends on Obs, but more ready for review than Obs. And started hacking on an ImagesController, what a beastie! SpeciesList is also a freakin scary chunk of code. On Dec 9, 2022, at 9:31 PM, Jason Hollinger ***@***.***> wrote:﻿
I did my level best.  I tried to at least put my eyes on every change, but I'm sure I didn't really see most of this long PR.
I appreciate your assertions on the fixtures making it clear what the test relies on.  I guess in some cases at least we used to do that in comments, but I doubt I was doing it everywhere.  That should help with understanding those tests better in the future.
I haven't seen any error reports (automated or otherwise) from the last PR.  When do you think it will be safe to push out this one?  Is one day long enough between monster PRs??

pellaea:
Arg, I can't find any examples right now, for some reason, sorry.  You were
doing something like:

def test_blah
   obj = <fixture>
   assert(<condition>, "need fixture with <condition>")
   <do test now that we know the fixtures have the required features>
end

nimmolo:
Thanks @pellaea — I think it's been enough time from the Account PR to merge this. 
Going once...

PR #1222, opened 2022-12-10
Added simple script to keep rolling backups of blocked ips list.

pellaea:


JoeCohen:
Question: Will `cp` automatically overwrite the existing file? 
(If I type it on the command line, it asks me for confirmation before overwriting).

pellaea:
Oops!  Right, those should be "cp -f" to force overwrite.  Thank you!  I
only ran it once to test it.

PR #1223, opened 2022-12-10
Fix path for `Account::LoginController#login_unverified` and add integration test

nimmolo:
... should be `account/verifications/reverify`

### DIFF
--- a/app/controllers/account/login_controller.rb
+++ b/app/controllers/account/login_controller.rb
@@ -94,7 +94,7 @@ module Account
 
     def login_unverified(user)
       @unverified_user = user
-      render("/account/reverify")
+      render("/account/verifications/reverify")
     end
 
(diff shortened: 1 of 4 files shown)


nimmolo:
More goodness:
- Wrote new capybara session extensions to check delivered mail contents (last, or _n_th from last) and to grab the first link out of that email's `href` attribute
- Renamed `string_extensions#render_html` to `unescape_html` and made a new method `translation_string.t.as_displayed` for integration test comparisons. Turns out you also need to `strip_squeeze` the translation string if there are double spaces.
- Tests that the reverify email goes out
- Tests that the link in the verify email is the same as the link in the reverify email
- Verifies with the link from the delivered email
- Tests attempt to reuse verification link
- Logs out and logs back in

PR #1224, opened 2022-12-11
Add integration test for unverified user trying to login

nimmolo:
This was the actual issue encountered in the wild. 

Logging in an unverified user tried to render a route, but it should have called `render` on the path to the template. This tests that scenario, that they actually get to a `reverify` link.

### DIFF
--- a/app/controllers/account/verifications_controller.rb
+++ b/app/controllers/account/verifications_controller.rb
@@ -68,7 +68,7 @@ module Account
       raise("This action should never occur!")
     end
 
-    # This is used by the "reverify" page to re-send the verification email.
+    # Route linked from the "reverify" template. Re-sends verification email.
     def resend_email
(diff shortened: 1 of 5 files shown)


nimmolo:
More goodness:
- Wrote new capybara session extensions to check delivered mail contents (last, or _nth_ from last) and to grab the first link out of that email's `href` attribute
- Renamed `string_extensions#render_html` to `unescape_html` and made a new method `translation_string.t.as_displayed` for integration test comparisons. Turns out you also need to `strip_squeeze` the translation string when there are double spaces, if you are comparing against browser output.
- Assert that the reverify email goes out
- Assert that the link in the verify email is the same as the link in the reverify email
- Verifies with the link from the delivered email
- Tests attempt to reuse verification link
- Logs out and logs back in

PR #1227, opened 2022-12-12
Rename `String#unescape_html` just in case

nimmolo:
Plus, call a different Ruby class that is [more appropriate for what this method does](https://stackoverflow.com/questions/32065012/ruby-unescape-html-string), and handles stuff like `&mdash;`.


pellaea:
Definitely not the problem.  I'm fixing the real problem in a separate PR.

PR #1229, opened 2022-12-12
`ImageController` break up methods to prepare for CRUD

nimmolo:
Another smallish PR, one of a series to reduce file changes in the upcoming `ImagesController` CRUD to facilitate review.

This mostly breaks up long methods for comprehensibility and ABC.

@JoeCohen I'd really appreciate your read-through here, if you have time.  Everything's still in the same file. 

The CRUD refactor PR will split up the controller, but the method changes there should be pretty minor.


app/controllers/image_controller.rb
```diff
@@ -268,7 +230,74 @@ def prev_image
     redirect_to_next_object(:prev, Image, params[:id].to_s)
   end
 
+  private
+
+  def image_exists
+    @image = find_or_goto_index(Image, params[:id].to_s)
```
pellaea, line 236: I would prefer to somehow expose the fact that @image is set as a side-effect.  I might add an exclamation, too, to emphasize that it's doing error handling and that the caller should abort if it fails.

app/controllers/image_controller.rb
```diff
@@ -268,7 +230,74 @@ def prev_image
...
+  end
+
+  def goto_next_image
+    query = find_or_create_query(Image)
+    query.current = @image
+    @image = query.current if query.index(@image) && (query = query.next)
+  end
+
+  def grab_list_of_votes
+    @votes = @image.image_votes.sort_by do |v|
```
pellaea, line 290: Again, it would be nice to expose the fact that @votes is set as a side-effect.

pellaea: Can you just set @votes in the caller instead?

app/controllers/image_controller.rb
```diff
@@ -300,37 +329,50 @@ def cast_vote
...
+    check_observation_permission
+
+    return new_image if request.method != "POST"
+
+    create_image
+  end
+
+  private
+
+  def new_image
```
pellaea, line 343: Ditto, @image is set.

app/controllers/image_controller.rb
```diff
@@ -300,37 +329,50 @@ def cast_vote
...
-                          action: :show, id: @observation.id)
     end
+    redirect_with_query(permanent_observation_path(id: @observation.id))
+  end
+
+  def observation_exists
+    @observation = find_or_goto_index(Observation, params[:id].to_s)
+  end
+
+  def check_observation_permission
```
pellaea, line 373: Exclamation might be appropriate here, too.

app/controllers/image_controller.rb
```diff
@@ -300,37 +329,50 @@ def cast_vote
...
         process_image(args, params[:upload]["image#{i}"])
         i += 1
       end
-      redirect_with_query(controller: :observations,
-                          action: :show, id: @observation.id)
     end
+    redirect_with_query(permanent_observation_path(id: @observation.id))
+  end
+
+  def observation_exists
```
pellaea, line 369: Ditto.

app/controllers/image_controller.rb
```diff
@@ -349,6 +391,10 @@ def process_image(args, upload)
     @image.original_name = "" if @user.keep_filenames == "toss"
     return flash_object_errors(@image) unless @image.save
 
+    add_image_to_observation
```
pellaea, line 394: (I love this!  Rubocop really wants to have short methods, but the result of its whining is this!  This is no different from a single long method!  I do the same thing.  It just makes me laugh.  "Oh, you really got me there, rubocop, good catch! " :)

app/controllers/image_controller.rb
```diff
@@ -364,62 +410,96 @@ def process_image(args, upload)
...
+
+    done = figure_out_if_done(xargs)
+    if done
+      redirect_with_query(action: "show_image", id: @image.id)
+    else
+      init_project_vars_for_reload(@image)
+    end
+  end
+
+  def set_licenses_ivar
```
pellaea, line 456: This is clear... but just throwing it out there, would it be even better to have the caller set @licenses and then name the method something that actually reveals what it's setting @licenses to, such as "current_names_and_ids" or something like that?  Even _more_ transparent, and just as short / concise.

app/controllers/image_controller.rb
```diff
@@ -364,62 +410,96 @@ def process_image(args, upload)
...
+    @licenses = License.current_names_and_ids(@image.license)
+  end
+
+  def check_image_permission
```
pellaea, line 460: Ditto exclamation.

app/controllers/image_controller.rb
pellaea, line 456: This is clear... but just throwing it out there, would it be even better to have the caller set @licenses and then name the method something that actually reveals what it's setting @licenses to, such as "current_names_and_ids" or something like that?  Even _more_ transparent, and just as short / concise.

app/controllers/image_controller.rb
pellaea, line 460: Ditto exclamation.

app/controllers/image_controller.rb
pellaea, line 373: Exclamation might be appropriate here, too.

app/controllers/image_controller.rb
pellaea, line 369: Ditto.

app/controllers/image_controller.rb
pellaea, line 343: Ditto, @image is set.

app/controllers/image_controller.rb
nimmolo: Are you suggesting that the caller should assign like `@image = image_if_exists!` and then 
```ruby
def image_if_exists!
  find_or_goto_index(Image, params[:id].to_s)
end
```

app/controllers/image_controller.rb
```diff
@@ -364,62 +410,96 @@ def process_image(args, upload)
...
+    redirect_with_query(action: "show_image", id: @image) unless
+      check_permission!(@image)
+  end
+
+  def set_xargs
```
pellaea, line 465: Oh wow, this is super embarrassing.  xargs is never used.  I think this had to do with a project where we were recording every mutation as a "transaction" (xargs = arguments to make that transaction), in order to enable multiple independent instances of MO to run and stay more or less in sync.  I ripped that code out ages ago, but it looks like I missed a piece! :)

app/controllers/image_controller.rb
nimmolo: I like. By the way this way of refactoring is way better than the mega-chunk, huh? Solicit comments on the methods, and then reorganize the controllers in a separate PR using improved methods.

app/controllers/image_controller.rb
```diff
@@ -364,62 +410,96 @@ def process_image(args, upload)
...
+      xargs[:set_copyright_holder] = @image.copyright_holder
+    end
+    if @image.original_name_changed?
+      xargs[:set_original_name] = @image.original_name
+    end
+    xargs[:set_license] = @image.license if @image.license_id_changed?
+    xargs
+  end
+
+  def figure_out_if_done(xargs = {})
```
pellaea, line 479: This could be done much more simply without xargs by just checking if any important columns changed: when_changed?, notes_changed?, copyright_holder_changed?, original_name_changed?, license_id_changed?

app/controllers/image_controller.rb
```diff
@@ -364,62 +410,96 @@ def process_image(args, upload)
...
+    else
+      xargs[:id] = @image
+      @image.log_update
+      flash_notice(:runtime_image_edit_success.t(id: @image.id))
+      update_projects(@image, params[:project])
+      true
+    end
+  end
+
+  def maybe_update_projects
```
pellaea, line 495: Haha!  Great name!  ...If your goal is to make readers laugh.  Less great if you're interested in them actually understanding what this method does. :)

pellaea: Right, this always attempts to update projects, and renders an error if it fails, so "update_projects!" would be a whole lot more clear.

app/controllers/image_controller.rb
```diff
@@ -805,37 +939,51 @@ def update_licenses_history(images_to_update, old_holder, old_license_id)
...
                        pluck(ImageVote[:id].count.as("total"))&.first
       @num_public = ImageVote.where(user_id: @user.id).
                     where(anonymous: false).
                     pluck(ImageVote[:id].count.as("total"))&.first
+    end
+  end
 
+  private
+
+  def create_anonymity_change
```
pellaea, line 966: Just a question: I saw this pattern above, too.  "create_some_action".  What does the "create" mean?  Is that code-speak referring to the fact that this is a POST method?

app/views/image/_image_thumbnail.html.erb
```diff
@@ -9,9 +9,14 @@
...
-   #If the image is part of an image set.
+  # If the image is part of an image set.
   is_set = is_set || false
-
+  # NOTE: The local var `link` might be to #show_image as you'd expect,
+  # or it may be a GET with params[:img_id] to the actions for #reuse_image
+  # or #remove_image ...or any other link. Firing a POST to those actions
+  # might require printing a Rails post_button and putting something like
+  # Bootstrap's .stretched-link class on the generated form input.
+  # However, the whole reuse_image page is currently a form - refactor?
```
JoeCohen, line 19: Consider moving lines 2-17 to a helper

app/controllers/image_controller.rb
nimmolo: Yes. More of a note to self. This is prep for making this one the POST action and the previous action the GET. 

It would probably be `Images::AnonymityController#new` (or maybe `#edit`).

app/views/image/_image_thumbnail.html.erb
nimmolo: Gooooooood idea.

app/controllers/image_controller.rb
nimmolo: `my_goal_is_always_to_make_readers_laugh!`

app/controllers/image_controller.rb
pellaea: Yes pretty much.  I might personally have preferred "return unless (@image = find_image!)".

This is all just personal aesthetics, you realize that right?  I'm not trying to prove in any absolute, cosmic sense that your solution here is "wrong".  I'm just giving feedback.  These are things I would prefer, and why.

In these sorts of refactors what I like to do for what very little it is worth -- you may approach it differently, so please feel free to ignore! -- is I like to rewrite the main method as a sort of executive bullet-point kind of thing.  It does A, B, C, and D, then returns.  And I like to manage as many globals (instance variables in this case) as possible in the top level.  It feels really good if I can somehow structure it so that each bullet-point is very simple and stand-alone, e.g., relying on 0-2 parameters only.  If it requires more parameters then maybe there's a way to split it?  Sometimes it's not possible, of course.

But then, when you read the top-level method you know exactly what it does, what all of the side-effects are, ideally even _how_ the subroutines are doing the work.  Then I can go and write all the little helper methods (the "bullet-points").

Obviously it was all a great deal easier when we could just throw it all in one big method, no need to worry about passing parameters around or anything, super free-form and forgiving.  But given that we're now being encouraged to write smaller and more modular and all that, I find that this top-down approach can really yield some nice code.  The very rare cases where I've actually done it, that is. :)

Another approach which we can call "bottom-up" is take the big long old method and just slice it up, changing the minimum amount, ideally cutting it into as meaningful chunks as possible.  Certainly _way_ easier and quicker and _way_ less likely to break code.  Like what I did when I "refactored" the observations_controller -- I just literally cut it up into several files but changed nothing fundamentally.  It was meant only to be a baby step in the right direction, and as such it was an improvement and just fine.

I'm not recommending a complete overhaul here.  I just want to make sure when you cut up the existing code the pieces get good names so that we know more or less what the pieces do without having to look up the code.  What's the point of cutting it up into pieces if you still have to read it all to make sense of any of it?  It's just a perverse game that helps nothing except fooling rubocop into thinking it passes all its silly metrics! :)  I understand this is just meant to be a step in the right direction.  And this is.

It's all good.  I'd just like to improve the names a bit, and I think it's good to go.

app/controllers/image_controller.rb
pellaea: Totally!  If it works for you, it sure is easier to track the changes at this point.  Once they're moved into separate files, I won't stand a chance.  Of course, then, you can get away with murder and I'll never know better. ;)

app/controllers/image_controller.rb
pellaea: Ah, thank!  Totally makes sense.

app/controllers/image_controller.rb
```diff
@@ -268,7 +230,74 @@ def prev_image
...
+    return unless obs.present? && obs.to_s.match(/^\d+$/) && !browser.bot?
+
+    obs_query = find_or_create_query(:Observation)
+    obs_query.current = obs
+    img_query = create_query(:Image, :inside_observation,
+                             observation: obs, outer: obs_query)
+    query_params_set(img_query)
+  end
+
+  def maybe_cast_user_vote
```
JoeCohen, line 267:
```suggestion
  def cast_vote
```
See @pellaea's comment re `maybe_update_projects`

app/controllers/image_controller.rb
pellaea: "maybe_update_projects_maybe_not_whats_it_to_you_punk?"

app/controllers/image_controller.rb
```diff
@@ -268,7 +230,74 @@ def prev_image
...
+
+  def set_default_size
+    # Decide which size to display.
+    @default_size = @user ? @user.image_size : :medium
+    @size = params[:size].blank? ? @default_size : params[:size].to_sym
+
+    # Maybe make this size the default image size for this user.
+    unless @user && (@default_size != @size) && (params[:make_default] == "1")
+      return
+    end
```
JoeCohen, line 247:
```suggestion
    return unless @user && 
                  (@default_size != @size) && 
                  (params[:make_default] == "1")
```
I like the guard clause (because it shows the non-normal path) instead of the old nested `if ... end`.
I'm merely suggesting avoiding making the guard clause itself nested. I think that also could be done with "`\`" line continuations.

app/controllers/image_controller.rb
JoeCohen: This method is only called once, so consider leaving things as they were?
```ruby
  @observation = ...
  return unless @observation
```
It's transparent, and the class is shorter.

And maybe for the other methods that @pellaea is commenting on.

app/controllers/image_controller.rb
```diff
@@ -349,6 +391,10 @@ def process_image(args, upload)
     @image.original_name = "" if @user.keep_filenames == "toss"
     return flash_object_errors(@image) unless @image.save
 
+    add_image_to_observation
+  end
+
+  def add_image_to_observation
     if @image.process_image(strip: @observation.gps_hidden)
```
JoeCohen, line 398: Use a guard clause?

JoeCohen:
@nimmolo: 
1. When this gets divided up, consider a `reuse` controller instead of a `reuse` action.
2. I'm out of steam for tonite. Will continue review tomorrow.

app/controllers/image_controller.rb
nimmolo: I broke it out because checks like this are repeated between new/create and edit/update actions, in CRUD. But i agree, if i'm assigning with the caller, it's better how it was

But here's a question: do they need to be repeated ? Does the POST action need to check/set the ivar again? I've been doing that, but never sure if necessary

pellaea: I assume POST set it just in case there was an error and it needed to re-serve the form?

app/controllers/image_controller.rb
nimmolo: I've re-read this a couple times. You're a good writer and the coding style you describe is exactly what i've been going for.  This clarifies a lot.

The hitch is my bottom-up understanding of the code. I often wade in with the vaguest possible idea what any of the methods do, including the obvious ones. I sometimes bump into blocks of code and discover entire scopes of action on MO that i didn't suspect existed. So i'm usually slogging through, just chopping code into manageable lengths until i have an inkling what they do.

I would write bullet point, directive sub-methods if i had a clear idea of where i was going. I'm grasping for that top-down understanding... and almost never arriving until i've gotten through fixing all the tests, at which point i don't feel like messing with the controller methods again. Hah!

pellaea: Excellent, sounds like we are totally on the same page.  I totally support what we're calling the bottom-up approach here.  There are many times and places where it makes sense.  And this is one of them.

Chop away and worry about making it "make sense" later when it's had time to percolate and you've gotten your head around the whole problem.  I withdraw my (already weak) criticisms.

nimmolo: Almost all these suggestions are now committed locally. But i'm getting an error on the thicket of update methods without `xargs`. Question coming...

app/controllers/image_controller.rb
nimmolo: Nevermind! Got it. It was IF image data was NOT changed, ...

app/controllers/image_controller.rb
nimmolo: Done

app/controllers/image_controller.rb
nimmolo: Done

app/controllers/image_controller.rb
nimmolo: Done

PR #1230, opened 2022-12-12
Bump gem versions for three recent Dependabot security updates

nimmolo:
For some reason the dependabot PRs don't pass tests, they lack the Rails credentials. 
This PR does the same updates "manually": #1216 #1234 & #1235.

Gems updated: `nokogiri 1.13.10`, `loofah 2.19.1`, `rails-html-sanitizer 1.4.4`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,12 +35,15 @@ group :rails do
   gem("sprockets-rails")
 end
 
-# security fix for CVE-2021-41817 regex denial of service vulnerability
+# Security fix updates via Dependabot
+# CVE-2021-41817 regex denial of service vulnerability
 gem("date", ">= 3.2.1")
-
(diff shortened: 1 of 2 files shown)


nimmolo:
The difference between this PR and the dependabot version is that I spec'd a minimum nokogiri version in the Gemfile.

Does anyone know a reason not to put nokogiri in the Gemfile?

JoeCohen:
Makes sense.(The only thing I can think of is the future theoretical
possibility that Nokogiri get abandoned by all the gems that we use, so we
no longer need it, so we shouldn't install it. That's pretty far-fetched.)

nimmolo:
Thanks @JoeCohen . Merging now

nimmolo:
Updated to add two new Dependabot suggestions.

PR #1232, opened 2022-12-13
Make temp file name more unique in exifautotran

mo-nathan:
Fix for blank and possibly corrupt images during image resizing.

### DIFF
--- a/script/exifautotran
+++ b/script/exifautotran
@@ -26,11 +26,11 @@ do
   esac
 
   if test -n "$transform"; then
-    convert $transform $file tempfile.jpg
+    convert $transform $file temp-$file.jpg
     if test $? -ne 0; then
       echo Error while transforming $file - skipped.
     else
-      mv -f tempfile.jpg "$file"
+      mv -f temp-$file.jpg "$file"
     fi
   fi
 done
--- a/script/exifautotran
+++ b/script/exifautotran
@@ -26,11 +26,11 @@ do
   esac
 
   if test -n "$transform"; then
-    convert $transform $file temp-$file.jpg
+    convert $transform "$file" /tmp/exifautotran-$$.jpg
     if test $? -ne 0; then
       echo Error while transforming $file - skipped.
     else
-      mv -f temp-$file.jpg "$file"
+      mv -f /tmp/exifautotran-$$.jpg "$file"
     fi
   fi
 done


script/exifautotran
```diff
@@ -26,11 +26,11 @@ do
   esac
 
   if test -n "$transform"; then
-    convert $transform $file tempfile.jpg
+    convert $transform $file temp-$file.jpg
```
pellaea, line 29: Just one consideration -- are you certain $file doesn't contain any directories?  I'll check....  Shoot, you can't tell from this script, depends how it's called.  How about using the pid instead?  temp-$$.jpg  Does that work in bourne shell?  Yes, it does.  That way it won't matter how it's called.

I'll check the process_image script to see if there are any other temp files with similar issues.

pellaea: PS. exifautotran is definitely be called with a relative path not just a file name.  image_root/orig/id.jpg.  Looks like this is the only tempfile used in the entire process_image script.  Although, the initial file from nginx may be a temp file, but I'm sure rails handles that intelligently.  This exifautotran script was quick and dirty, not surprised it was missing something so critical.

mo-nathan:
Good catch!  Anything reasonably unique should work.  PID is fine.
$file-temp.jpg would probably work as well.  Up to you.

pellaea:
> Good catch! Anything reasonably unique should work. PID is fine. $file-temp.jpg would probably work as well. Up to you.

Good idea.  Both should work.  The only thing I can think of that might make me lean toward temp-$$.jpg is that the temp files would be more "exposed", in case some processes got killed leaving temp files stranded.  If they're in images_root/orig directory, they might build up unseen for a long time.  But if they're in /var/web/mo, I'll see them periodically and clean them out.

Yet another option is to stuff 'em in /var/web/mo/tmp/exifautotran-$$.jpg or /tmp/exifautotran-$$.jpg.  Those are probably the standard conventions.  The latter is automatically cleaned periodically, right?  I don't know about the former.

mo-nathan:
I like the idea of dropping the file in /tmp and using the PID, so I made those updates.  I also noticed a spot where $file should probably be in double quotes as well.

pellaea:
Excellent, good catch re: adding quotes, too.  I say we let this thing loose and see if our troubles disappear!

Thank you _so_ much for figuring that out, Nathan.

mo-nathan:
Sounds good!  Can you make it happen?  I'm up to my eyeballs with work
today and I have an evening thing.

PR #1233, opened 2022-12-13
`CommentsController` CRUD

nimmolo:
### CommentsController CRUDified ###
- Comments

### Integration test coverage ###
Covered by `AmateurTest` and `ExpertTest`.


nimmolo:
If i could request review of one file in this PR, it's the `CommentsController` itself. Changes are fewer than in the `ImageController` PR and reflect suggestions made there. 

Most other changes are just routes.

app/controllers/comments_controller.rb
```diff
@@ -156,35 +155,62 @@ def show_selected_comments(query, args = {})
...
     @target = @comment.target
     allowed_to_see!(@target)
   end
 
-  # Go to next comment: redirects to show_comment.
-  def next_comment
-    redirect_to_next_object(:next, Comment, params[:id].to_s)
+  private
+
+  def find_comment!
```
pellaea, line 187: Thank you!

app/controllers/comments_controller.rb
```diff
@@ -197,25 +223,43 @@ def prev_comment
...
+    @comment = Comment.new(target: @target)
     @comment.attributes = whitelisted_comment_params if params[:comment]
-    if @comment.save
-      @comment.log_create
-      flash_notice(:runtime_form_comments_create_success.t(id: @comment.id))
-      redirect_with_query(controller: @target.show_controller,
-                          action: @target.show_action, id: @target.id)
-    else
+
+    save_comment_or_flash_errors_and_redirect!
```
pellaea, line 240: nice!  I love it!  I know exactly what this is going to do.

app/models/comment.rb
```diff
@@ -98,9 +113,22 @@ class Comment < AbstractModel
   after_create :notify_users
   after_create :oil_and_water
 
+  scope :by_user,
+        ->(user) { where(user: user) }
+  scope :for_user,
```
pellaea, line 118: I like that you've done this without a join.  But do we need to send the list of target ids back to the webserver at all?  Can we do this as a subquery instead?  Or is this somehow being done as a subquery under the hood?  Actually I really wonder how this _is_ being done under the hood as it is!

I think this would be easier to do if we did the polymorphism the way rss_logs does it: select where observation_id in (select id from observations where user_id = 1234) or name_id in (...) or ...

But something somewhat equivalent should be possible with comments: select where target_type = "Observation" and target_id in (select ...) or target_type = "Name" and target_id in (select ...) or etc.

Can that be built in a for loop somehow?  Same number of queries on the db, ultimately, but all of the subqueries' results stay local on the db.  Is that any better?

nimmolo: I think that's exactly what it does, right? Ohhh, i get it, you mean bake the loop into the sql so the db does it, not in ruby.

`where(target: proc)` or something?  

I don't know how to do that. I could chain the `where().or(where())` scopes maybe by hardcoding each model. The current scope is nice because the target types are pulled off the class definition further down, so if the target classes change... they only have to change in one place.

Let's set that as an aspiration and i'll put a note in the comments. This scope isn't called anywhere yet, it's just another "note to self", for when we try to summit the Query class with scopes.

Come to think of it, it should probably `pluck` the id, right?

pellaea: Aspiration, yes!  The pluck is a good idea to do immediately.

Just messin' around, can you do this?

`
Comment.all_types do |type|
  model = get_model_from_type_tag(type)
  scope = scope.or.where(target_type: type, target_id: model.where(user: user))
end
`

nimmolo: I'm intrigued, because i don't get how chaining works in scopes yet.

Consider this attempt, class method defs included for clarity. This doesn't fly (nor with substituting `self` for `scope`). The first problem is we have to start with a `where`, not an `or`, or else the initial scope will be `Comment.all` and then we're just `or`-ing onto everything. I'm doing this in the rails console and inspecting results.

```ruby
  scope :for_user,
        lambda { |user|
          all_types.each do |model|
            scope = scope.or(where(target_type: model.name.underscore.to_sym,
                                   target_id: model.where(user: user)))
          end
        }

  # Returns Array of all models (Classes) which take comments.
  def self.all_types
    all_type_tags.map { |t| t.to_s.camelize.constantize }
  end

  # Returns Array of all valid +target_type+ values (Symbol's).
  def self.all_type_tags
    [:location, :name, :observation, :project, :species_list]
  end
```

nimmolo: Ok. Consider this, which works as of Rails 6. 
```ruby
# in the Person model:
scope :john, -> { where(name: "John") }
scope :smith, -> { where(lastname: "Smith") }
scope :john_or_smith, -> { john.or(self.smith) }
```
Can we iterate to make a bunch of scopes, and put them together?

nimmolo: We can also do this. It's confounding me how to combine them with the above:
```ruby
  [:small, :medium, :large].each do |s|
    scope s, where(size: s) 
  end
``

Comment (pellaea) [reply app/models/comment.rb:118]: What about something like this?

```
scope :for_user,
  lambda do |user|
    all_types.inject(nil) do |scope, model|
      scope2 = where(target_type: model.name.underscore.to_sym,
                     target_id: model.where(user: user))
      scope ? scope.or(scope2) : scope2
    end
  end
```

Comment (pellaea) [reply app/models/comment.rb:118]: Less certain this will work, but it would be shorter...

```
scope :for_user,
  lambda do |user|
    all_types.inject do |scope, model|
      where(target_type: model.name.underscore.to_sym,
            target_id: model.where(user: user)).or(scope)
    end
  end
```

Comment (nimmolo) [reply app/models/comment.rb:118]: The first one works. What the heck? No comprende.

I'm guessing `nil` pushes nil as the first el of the array. 
How are you iterating two args? What is `do |scope, model|`? in other words

Comment (pellaea) [reply app/models/comment.rb:118]: Look up how inject works.  I hate that method, it always hurts my brain.  It is equivalent to

```
lambda do |user|
  scope = nil
  all_types.each do |model|
    scope2 =where(blah)
    scope = scope ? scope.or(scope2) : scope2
  end
  scope
end
```

Comment (nimmolo) [reply app/models/comment.rb:118]: Ugh. Gross. But impressive. Kind of the Cordyceps of iterators.

Comment (nimmolo) [reply app/models/comment.rb:118]: Got to respect it.

Comment (nimmolo) [reply app/models/comment.rb:118]: I'm so impressed.
```ruby
3.1.2 :002 > Comment.for_user(252).to_sql
 => "SELECT `comments`.* FROM `comments` WHERE (((`comments`.`target_type` = 'location') AND (`comments`.`target_id` IN (SELECT `locations`.`id` FROM `locations` WHERE `locations`.`user_id` = 252))) OR ((`comments`.`target_type` = 'name') AND (`comments`.`target_id` IN (SELECT `names`.`id` FROM `names` WHERE `names`.`user_id` = 252))) OR ((`comments`.`target_type` = 'observation') AND (`comments`.`target_id` IN (SELECT `observations`.`id` FROM `observations` WHERE `observations`.`user_id` = 252))) OR ((`comments`.`target_type` = 'project') AND (`comments`.`target_id` IN (SELECT `projects`.`id` FROM `projects` WHERE `projects`.`user_id` = 252))) OR ((`comments`.`target_type` = 'species_list') AND (`comments`.`target_id` IN (SELECT `species_lists`.`id` FROM `species_lists` WHERE `species_lists`.`user_id` = 252))))"                                                                          
```

pellaea: Woah!!  Check it out!  It works!  What a crazy query!  I love it!  Those sorts of puzzles are fun... when we can actually work out the solution(!)

nimmolo: Wait a minute. Are we doing an unnecessary check for `target_type`? It's not necessary to specify the `comment` record.

Couldn't we just have it scanning the `target_id` column only? Or is this faster.

app/models/comment.rb
```diff
@@ -98,9 +113,22 @@ class Comment < AbstractModel
...
+          where(target: targets)
+        }
+  scope :for_target,
+        ->(target) { where(target: target) }
+
   # Returns Array of all models (Classes) which take comments.
   def self.all_types
-    [Location, Name, Observation, Project, SpeciesList]
+    all_type_tags.map { |t| t.to_s.camelize.constantize }
   end
```
nimmolo, line 152: Is this unnecessary DRY? Does it force a class lookup on all those type tags? Yes, it does...

The inverse, deriving the tags from a list of classes,  would also reference the classes.
Maybe best to leave it like it was, a double list.

pellaea:
I agree, let 'er rip!

PR #1236, opened 2022-12-14
`InterestController` ABC

nimmolo:
Simple PR preparing for CRUD.

Also adds scope `:for_user` to `Interest` and `Notification` models.

### DIFF
--- a/app/controllers/interest_controller.rb
+++ b/app/controllers/interest_controller.rb
@@ -64,37 +64,37 @@ class InterestController < ApplicationController
   # Outputs: none
   def set_interest # rubocop:disable Metrics/AbcSize
     pass_query_params
-    type   = params[:type].to_s
-    oid    = params[:id].to_i
-    state  = params[:state].to_i
-    uid    = params[:user]
-    target = Comment.find_object(type, oid)
+    target_type = params[:type].to_s
+    target_id   = params[:id].to_i
+    state       = params[:state].to_i
+    user_id     = params[:user]
+    target      = Comment.find_object(target_type, target_id)
     return unless @user
 
     interest = Interest.find_by(
-      target_type: type, target_id: oid, user_id: @user.id
+      target_type: target_type, target_id: target_id, user_id: @user.id
     )
-    if uid && @user.id != uid.to_i
+    if user_id && @user.id != user_id.to_i
       flash_error(:set_interest_user_mismatch.l)
     elsif !target && state != 0
-      flash_error(:set_interest_bad_object.l(type: type, id: oid))
+      flash_error(:set_interest_bad_object.l(type: target_type, id: target_id))
     else
       set_interest_state_for_target(interest, state, target)
     end
-    redirect_to_target_or_interests(target)
+    redirect_to_target_or_interests(target, target_id)
   end
 
   private
 
-  def redirect_to_target_or_interests(target)
+  def redirect_to_target_or_interests(target, target_id)
     unless target
-      redirect_back_or_default(controller: "interest",
-                               action: "list_interests")
+      return redirect_back_or_default(controller: "/interest",
+                                      action: "list_interests")
     end
 
     redirect_back_or_default(
       add_query_param(controller: target.show_controller,
-                      action: target.show_action, id: oid)
+                      action: target.show_action, id: target_id)
     )
   end
 
--- a/app/controllers/interest_controller.rb
+++ b/app/controllers/interest_controller.rb
@@ -17,6 +17,11 @@
 class InterestController < ApplicationController
   before_action :login_required
   before_action :disable_link_prefetching
+  before_action :pass_query_params,
+                except: [
+                  :list_interests,
+                  :destroy_notification
+                ]
 
   # Show list of objects user has expressed interest in.
   # Linked from: left-hand panel
@@ -62,29 +67,29 @@ class InterestController < ApplicationController
   # Redirects back (falls back on show_<object>)
   # Inputs: params[:type], params[:id], params[:state], params[:user]
   # Outputs: none
-  def set_interest # rubocop:disable Metrics/AbcSize
-    pass_query_params
+  def set_interest
     target_type = params[:type].to_s
     target_id   = params[:id].to_i
     @state      = params[:state].to_i
-    user_id     = params[:user]
-    @target     = Interest.find_object(target_type, target_id)
-    return unless @user
-
-    @interest = Interest.find_by(
+    @target     = AbstractModel.find_object(target_type, target_id)
+    @interest   = Interest.find_by(
       target_type: target_type, target_id: target_id, user_id: @user.id
     )
-    if user_id && @user.id != user_id.to_i
+    set_interest_or_flash_errors(target_type, target_id)
+    redirect_to_target_or_interests
+  end
+
+  private
+
+  def set_interest_or_flash_errors(target_type, target_id)
+    if (user_id = params[:user]) && @user.id != user_id.to_i
       flash_error(:set_interest_user_mismatch.l)
     elsif !@target && @state != 0
       flash_error(:set_interest_bad_object.l(type: target_type, id: target_id))
     else
       set_interest_state_for_target
     end
-    redirect_to_target_or_interests
   end
-
-  private
 
   def set_interest_state_for_target
     create_interest_if_not_exists
--- a/app/controllers/interest_controller.rb
+++ b/app/controllers/interest_controller.rb
@@ -26,13 +26,13 @@ class InterestController < ApplicationController
     store_location
     @title = :list_interests_title.t
     # notifications = Notification.find_all_by_user_id(@user.id).sort do |a,b|
-    notifications = Notification.where(user_id: @user.id).sort do |a, b|
+    notifications = Notification.for_user(@user).sort do |a, b|
       result = a.flavor.to_s <=> b.flavor.to_s
       result = a.summary.to_s <=> b.summary.to_s if result.zero?
       result
     end
     # interests = Interest.find_all_by_user_id(@user.id).sort do |a,b|
-    interests = Interest.where(user_id: @user.id).sort do |a, b|
+    interests = Interest.for_user(@user).sort do |a, b|
       result = a.target_type <=> b.target_type
       if result.zero?
         result = (a.target ? a.target.text_name : "") <=>
--- a/app/controllers/interest_controller.rb
+++ b/app/controllers/interest_controller.rb
@@ -17,11 +17,7 @@
 class InterestController < ApplicationController
   before_action :login_required
   before_action :disable_link_prefetching
-  before_action :pass_query_params,
-                except: [
-                  :list_interests,
-                  :destroy_notification
-                ]
+  before_action :pass_query_params, except: [:list_interests]
 
   # Show list of objects user has expressed interest in.
   # Linked from: left-hand panel
@@ -72,28 +68,41 @@ class InterestController < ApplicationController
     target_id   = params[:id].to_i
     @state      = params[:state].to_i
     @target     = AbstractModel.find_object(target_type, target_id)
-    @interest   = Interest.find_by(
-      target_type: target_type, target_id: target_id, user_id: @user.id
-    )
-    set_interest_or_flash_errors(target_type, target_id)
-    redirect_to_target_or_interests
+
+    if check_params_or_flash_errors!(target_type, target_id)
+      @interest = find_or_create_interest
+      set_interest_state_for_target
+    end
+
+    redirect_to_target_or_list_interests
   end
 
   private
 
-  def set_interest_or_flash_errors(target_type, target_id)
+  def check_params_or_flash_errors!(target_type, target_id)
     if (user_id = params[:user]) && @user.id != user_id.to_i
       flash_error(:set_interest_user_mismatch.l)
+      return false
     elsif !@target && @state != 0
       flash_error(:set_interest_bad_object.l(type: target_type, id: target_id))
-    else
-      set_interest_state_for_target
+      return false
     end
+    true
+  end
+
+  def find_or_create_interest
+    interest = Interest.find_by(
+      target_type: @target.type_tag, target_id: @target.id, user_id: @user.id
+    )
+    return interest unless !interest && @state != 0
+
+    interest = Interest.new
+    interest.target = @target
+    interest.user = @user
+    interest
   end
 
   def set_interest_state_for_target
-    create_interest_if_not_exists
-
     if @state.zero?
       remove_interest_from_target_and_flash_notice
     elsif @interest.state == true && @state.positive?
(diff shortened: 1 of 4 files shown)


app/controllers/interest_controller.rb
```diff
@@ -25,100 +25,133 @@ class InterestController < ApplicationController
...
 
+  public
+
   # Callback to change interest state in an object.
   # Linked from: show_<object> and emails
   # Redirects back (falls back on show_<object>)
   # Inputs: params[:type], params[:id], params[:state], params[:user]
   # Outputs: none
-  def set_interest
+  def set_interest # rubocop:disable Metrics/AbcSize
```
pellaea, line 65: While I personally have absolutely no problem with this method as-is... in this case it _could_ be debigulated in a few ways.

1. pass_query_params and requires_login could be outsourced to before_filters along the lines of other controllers.
2. you could move the find_object and error checking to another method, then the only params[:xxx] you need in this top level method is state, I think
3. even state could be moved out if you moved set_interest_state out into a fairly trivial submethod, too.

I can't believe rubocop has brainwashed me into arguing its case for it...  ¡Fight the system!

pellaea: Oops, I see you already do #3, ignore that one.

app/controllers/interest_controller.rb
```diff
@@ -25,100 +25,133 @@ class InterestController < ApplicationController
...
+
+    redirect_back_or_default(
+      add_query_param(controller: target.show_controller,
+                      action: target.show_action, id: target_id)
+    )
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def set_interest_state_for_target(interest, state, target)
+    if !interest && state != 0
```
pellaea, line 103: I wonder if this would pass ABC if this initialization of new interests were outsourced?  Once a method is reduced to a simple switch statement, and it still fails ABC, then fair enough.  But maybe this one can actually be fixed without silly contortions.

app/controllers/interest_controller.rb
```diff
@@ -25,100 +25,133 @@ class InterestController < ApplicationController
...
+      interest = Interest.new
+      interest.target = target
+      interest.user = @user
+    end
+    if state.zero?
+      remove_interest_from_target_and_flash_notice(interest, target)
+    elsif interest.state == true && state.positive?
+      flash_notice(
```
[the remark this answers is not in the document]
nimmolo: Got it. Committing a change now.

Question: is there any reason not to make ivars for `@interest`, `@state`, and `@target` here?
Instead of passing them around everywhere?

app/controllers/interest_controller.rb
```diff
@@ -25,100 +25,133 @@ class InterestController < ApplicationController
...
-    uid    = params[:user]
-    target = Comment.find_object(type, oid)
-    if @user
-      interest = Interest.find_by(
-        target_type: type, target_id: oid, user_id: @user.id
+    target_type = params[:type].to_s
+    target_id   = params[:id].to_i
+    state       = params[:state].to_i
+    user_id     = params[:user]
+    target      = Comment.find_object(target_type, target_id)
```
nimmolo, line 71: Is there a more direct way to get the `target` object than using a `Comment` class method?

I get it, it works, but `target_type.to_s.camelize.constantize.find(target_id)` is simpler.

nimmolo: Oh, find_object is from AbstractModel and does some error handling. Changing it back to
`Interest.find_object(target_type, target_id)`

app/controllers/interest_controller.rb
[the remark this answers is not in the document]
pellaea: Since you ask... (!)

One purpose of ivars is as a screwy mechanism for "passing" parameters to the view templates.  But that's not needed here.

The other is to store state information in a class.  I mean, yes, you could spin your suggested use as "state information", but it's stretching it a bit.  It smells more like laziness to me.  (No offense intended!  I've done it myself, no judgement!)

I personally think that a better solution in this case is to re-imagine the design so that parameters don't need to be passed around so much.  Whenever lots of parameters are flying all over, it means concerns aren't be separated properly, or we need a class to hold that information (e.g. a "bounding box" class that can hold north, south, east, west extents in a single object instead of forcing you to pass around four parameters in mapping routines).

In this case, note that target already provides access to target.type and target.id, so there go two parameters, nicely!  And login_required assures that @user.id is the same as params[:user_id], so there it goes.  I think you end up with: target, interest and state, right?  And doesn't interest give you access to the target?

Among the other things I like to see in "good" code --in addition to things like well-named methods, well-separated concerns, few parameters, globals all explicitly set (as opposed to being set by side-effect in submethods) -- is to have top-level parameters clearly exposed.  In the case of controllers, your parameters are all in params[:xxx], of course.  And there's no real reason to pass "params" around, as it's already freely available to everyone.  My own code has been all over the board with respect to this.  Basically, if it's convenient and I'm not feeling lazy, I'll pass params[:state] explicitly to the submethods that need it, if it thereby moves all accesses to params[:xxx] to the top level action method.  It's just nice to be able to see exactly what a method's inputs and outputs are.

I'm also fine with simply documenting the list of parameters in a comment.  But note that if you're trying to track down a bug associated with a parameter, or if you're trying to see how an action actually uses a parameter, it can be a pain in the butt drilling down multiple levels to find where the parameter is actually used.  So, I think it is still "best" to move them all into the top-level and pass them around as parameters.  But not critical.

One classic counter-case is whitelisted_parameters.  But we all know to look for parameters there.  That's a standard mechanism for bundling and processing large numbers of parameters all associated with a single model.

app/controllers/interest_controller.rb
pellaea: "Interest" makes more sense than "Comment".  I admit I did a double-take on that.  I'm sure there is a perfectly sensible historical reason for that ... maybe the method was moved from Comment into AbstractModel after writing this?  In any case, good to see it moved to something more reasonable.  Wait, hell, why not "AbstractModel.find_object"??  That would really emphasize that this has nothing to do with Interest per se, and that we're just looking up an object by type and id!

nimmolo:
You had me at “lazy” - i’m definitely feeling lazy. I get what you’re saying, though.

Is there a cost to setting ivars? 
Don’t they get cleared as soon as we leave the realm of the InterestController?

pellaea:
Haha!  No cost to ivars that I know of.  Larry Wall considered laziness one of the Three Great Virtues. :)

Just because there's no harm in repeating it -- these comments are just feedback, suggestions, ideas.  Feel free to push back or outright ignore any or all of it.

nimmolo:
I really appreciate ALL the feedback, and especially the explanation of programming styles and virtues. It has been very clarifying!!! Plus I appreciate that it takes time to write. It takes me forever to write.

I guess my main goal is to improve legibility and ease for maintainers, so ivars seem to fill the bill here.

nimmolo:
The other thing i wanted to ask is, what is the exact relation between Interests, Notifications and QueuedEmails? I notice Interests and Notifications don't belong to each other as records.

If you set an interest, you get notifications, which fire queued emails?

Are Notifications what we would use for an in-app "account messages" view?

I wonder if for CRUD clarity i might move the `destroy_notification` method to either `Interests::NotificationsController` or a separate one-action `NotificationsController`. Not in this PR though.

pellaea:
Crap, I was hoping you wouldn't notice.  Interests and Notifications are the same thing.  It's entirely my fault.  Nathan wrote the Notifications template, but I didn't understand his intentions because it seemed to be focused on allowing users to send notifications to _other_ users, e.g., researchers sending specimen requests to observers.  So I created Interests for a way to request notifications yourself when various objects had activity.  But I could totally have done that within the existing Notification system.  Sorry!

Notification/Interest are a way of specifying what kind of emails you want, QueuedEmail is the email itself.  Does that make sense?

nimmolo:
Whoops, wrong button. LOL! Why weren't/aren't the tables just combined?

pellaea:
(I wonder if the name "Notification" was part of why I missed what Nathan was doing so long ago?  You're right, "Notification" sounds like the email itself, not what you are interested in being notified of.  This is why names are important! :)  Not to excuse my mistake or anything, but it is an interesting lesson...)

pellaea:
Oh I know!  I've done that so many times!!

nimmolo:
So... can we write a script to combine the tables under "Interests"? Seems easy?!

pellaea:
Well, sorta.  But the mechanisms are not identical, so I think if we chose
to merge Notification into Interest we'd have to refactor all the
Notification code.

Of course, I totally support doing so, if you have the time, interest,
energy, patience, etc. to do the work.  Up to you!

nimmolo:
Notifications are only used for Names? (According to Notification.all_flavors)
I also notice that Notification stores the email template and body, that's a difference. 

Just thinking... are those attributes Interest needs, or should they be handled by a queued email, selected depending on the type of interest (if Interest.target_type == :name)?

nimmolo:
I've got the ABC all down but i feel like i'm not handling the `@user == params[:user]` check at the top level like i should, here. I also don't like that i try to set an `@interest` at the top level, and then buried is this other `create_interest_if_not_exists`.

It's all twisted. I've tried three times to un-twist it, but i don't find where i'm messing it up. I feel like there should be a simple method like `@interest = find_or_create_interest`, but whenever I try to write that, it throws errors that something isn't set, or an object is nil.

nimmolo:
Ah! I got it untwisted! 
Still interested in the question about Notification attributes, if you have bandwidth.

pellaea:
> Notifications are only used for Names? (According to Notification.all_flavors) I also notice that Notification stores the email template and body, that's a difference.
> 
> Just thinking... are those attributes Interest needs, or should they be handled by a queued email, selected depending on the type of interest (if Interest.target_type == :name)?

No, QueuedEmail can't handle it.  Here:  The user goes to a name page, and on that page they select "Email Tracking" or some such.  This takes them to a form inquiring details about how they would like to be notified and if they want an automated (spam!) notification sent to the observer who posts an observation of that name.  They save that form and it creates a Notification with all that detail.  There are no QueuedEmails yet.  So there would be no way to store that information in a QueuedEmail.  Then another user posts an observation.  Someone proposes a name you are tracking for that observation.  This triggers the Notification.  The Notification, in turn, creates two QueuedEmails -- one to the researcher, one to the observer.  Yes, the QueuedEmail to the observer will have the custom note template attached, but that has to come from somewhere.  It's not in the users table.  It has to be stored somewhere.  It's stored in the notifications table now.  Interests do not currently have a way to store that information.

So do you see how it's non-trivial to merge these two?  There's a reason we haven't done it yet!

Maybe you can allow interests.target_type to take on a new value which is not a model class name, say "NameTracking" or something like that?  Would that break the rails polymorphic relationship too badly?  Then add interests.template and interests.require_specimen columns.  Then you can fold notifications table into interests table.  Do a global search and replace of "Notification" => "Interest".  Hopefully that will only create a handful of inevitable twiddly bugs, and you're good to go??  Undoubtedly extremely naive and overly optimistic, but it's the direction I would go, personally.

nimmolo:
How about, if `interests.target_type == :name` it just behaves differently: it looks for something in the `interests.template` and `interests.require_specimen` columns, which can otherwise remain blissfully blank.

By the way, I love this.  huh?
```
class Notification < AbstractModel
  belongs_to :user

  scope :for_user,
        ->(user) { where(user: user) }

  # Do not change the integer associated with a value
  enum flavor:
       {
         name: 1,
         observation: 2,
         user: 3,
         all_comments: 4
       },
       _suffix: :flavor

  # List of all available flavors (strings).
  def self.all_flavors
    ["name"]
  end
```

pellaea:
Okay, so just throwing this out there, maybe stupid...  What about adding "Notification" as a new Interest target_type?  Actually, I would rename the class then from Notification to "NameTracking" (or whatever).  Then Interest becomes the front point for _all_ such requests for notifications, but the polymorphic relationship remains strict (there is still an actual object pointed to for the name-tracking notification request).  Interest itself doesn't need to hold any information that isn't required by all notification requests.  We still have a place to store details like spam template and whether specimens are required.  It allows additional "flavors" to be created in the future, each corresponding to a new table in the database and ActiveRecord model class.  Blah blah blah.  I think we could do away with notifications.flavor column, then, as it's useless -- it always contains the string "Naming" I think?  Totally superfluous at this point.  But otherwise the database doesn't change!  We just add a record to interests corresponding to each record in Notifications.  (And change the name "Notification" to something less misleading!!)

nimmolo:
That sounds solid!

pellaea:
(responding to your  "love this huh?" email -- yes!  see?  that clearly proves that Nathan had something in mind concerning expanding Notifications that would eventually include all the various email_xxx columns now in users... but which weren't there at the time... too bad we hadn't been communicating better so I actually knew what he had in mind! :)

nimmolo:
By the way, point of clarity... do Notifications track  _namings_ or names? 

And should I remove the unused options in that Notifications enum?

pellaea:
I don't remember the precise details.  I think it waits until an
observation's consensus name changes.  And if the new name is one that a
user is tracking, then it triggers.  That includes when the observation is
created with an initial name.  You'd have to look at the details in
app/models/observation.rb.  Sorry!

But maybe the question is what notifications.obj_id refers to?  Definitely
names.id.

nimmolo:
Thanks. OK, Just checked, there are no enum values in the db other than for name. I will remove the unused enum values and set `Notification.all_flavors` to `Notification.flavors.keys`, that just grabs the keys of the Rails native enum.

Sometimes I forget how much work i've done. Native enums! Woohoo!

nimmolo:
Merging this now.

I like that idea and i'm definitely into working on it. 

Do all notification requests currently, already, generate an `Interest` as well? (I don't think so.) 
Would we need to recode so they do create that record? `Interest.target_type = :name_tracker`

I guess the first thing is a migration renaming the `Notifications` table to `NameTrackers`

pellaea:
Yep, I think that's a fine place to start.  And yes, the next step would be
to have notifications stuff (is there a controller?) start creating an
interest in addition to the NameTracker.

nimmolo:
Thanks. There’s no notifications controller, i think it all happens in the Names controller, yet to be crudified. 
I will probably make a Names::NameTrackersController for this though.

Next question: should Interest.target_type be converted to an enum? 
That actually seems like the first actionable item.

nimmolo:
Or maybe it’s more like this:
Copy the values from Interest.target_type column into a new Interest.flavor column, converting values to enums.
Check the result
Drop the Interest.target_type column
Make Interest.target_type an instance method that converts the flavor (enums are stored as integers but return a string!) to a symbol.

Presto, don’t change any methods. Everything can still call .target_type

pellaea:
Interest#target_type should be the same as whatever Comment#target_type is,
and whatever ActiveRecord polymorphic association needs, right?  If that's
an enum, then that's what it should be!

nimmolo:
Oh wow. I didn’t realize type_tag was auto generated by activerecord when you call an assoc polymorphic 

Huh.
Got to read up on that.

nimmolo:
Right. It's not an enum. 

The way to limit allowed classes for a polymorphic association is apparently through a [validation](https://stackoverflow.com/questions/40681599/rails-polymorphic-association-restrict-allowed-classes) - this also conveniently could make a constant available listing the flavors of Interest.

PR #1238, opened 2022-12-15
Remove `flavor` from `Notification`

nimmolo:
This is in preparation for renaming `Notification` > `NameTracker` and making it one of the possible targets of an `Interest`.
Discussion here: #1237

**Note**: This drops the column `notifications.flavor` from the DB. Please review!

Background: Notifications are only used to track names (namings). Possible other projected "flavors" of notifications, which have not yet been built, could now be achieved by making other models like this, trackers that target objects, and hold mail templates and other data. 

Interests could then target any of these tracker classes.

The goal is to clear up the functional overlap between Interests and Notifications.

### DIFF
--- a/app/controllers/name_controller.rb
+++ b/app/controllers/name_controller.rb
@@ -1062,9 +1062,8 @@ class NameController < ApplicationController
     @name = find_or_goto_index(Name, name_id)
     return unless @name
 
-    flavor = Notification.flavors[:name]
     @notification = Notification.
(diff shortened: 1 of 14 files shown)


pellaea:
Looks good.  Love how much it simplifies things in the short term! :)

nimmolo:
Shall I merge? This PR seems simple enough.

pellaea:
Yup, looks good to me.

PR #1249, opened 2022-12-25
Record observation source

pellaea:
This is a fairly minimal first attempt.  It just adds an enum "source" column to the observations table with four options: (null is allowed for pre-existing observations)

mo_website
mo_iphone_app
mo_android_app
mo_api

Any observation created on the website from now on will be recorded as "mo_website".

When creating observations on the API you can now pass in a source parameter which can take any of those four values.  Right now I'm just stupidly trusting the API user to correctly supply this value.  In the future we should change this to authenticate the app.  Let's talk about this later.  The default is "mo_api" if no value is supplied.  Therefore all observations created via the API will now be marked as "mo_api" unless the app explicitly tells it to record it as something else, e.g., "mo_iphone_app" or "mo_android_app".

It displays this source information reasonably prominently right below the comments.  Of course, if there are a lot of comments, it will get pushed way off the screen, but I think it's reasonably visible for new observations without comments or name proposals yet.  It displays nothing for observations created before this PR or which were created on the website (that made it too hard to notice which were created by the app or API). 

Any recommendations or suggestions?


app/models/observation.rb
```diff
@@ -1218,6 +1218,32 @@ def collector_and_number
...
+          mo_website: 1,
+          mo_android_app: 2,
+          mo_iphone_app: 3,
+          mo_api: 4
+        }
+
+  # Which agent created this observation?
+  def self.all_sources
+    [:mo_website, :mo_android_app, :mo_iphone_app, :mo_api]
+  end
```
JoeCohen, line 1238: Should this return strings instead of symbols?
If symbols, maybe the following to DRY things
```suggestion
  def self.all_sources
    self.sources.keys.map(&:to_sym)
  end
```
or maybe just define a frozen constant.
(And analogous suggestions if strings are needed.)

JoeCohen (approved):

pellaea:
Thanks, Joe.  I've just removed Observation.all_sources altogether.  I
don't see any need for it if that trick works!

app/models/observation.rb
pellaea: Thank you, good suggestion.  Taken care of, I think.

app/controllers/application_controller.rb
```diff
@@ -354,6 +354,7 @@ def autologin
     make_logged_in_user_available_to_everyone
     track_last_time_user_made_a_request
     block_suspended_users
+    clear_admin_mode_for_nonadmins
```
mo-nathan, line 357: I don't see a problem with this, but I don't understand why it's needed or why it's part of this PR.

pellaea:
I'd be happy to do it in a separate PR.  It's such a trivial, unimportant
thing.  It basically protects us against the extremely unlikely possibility
that an admin turns on admin mode on a public computer and forgets to turn
it off before leaving.  Then -- somehow, and I'm no longer certain of how
this happened -- another user can come along and log in and end up in admin
mode.  The autologin cookie must have expired before the session?  Anyway,
it seemed cheap and prudent to make absolutely certain admin mode is turned
off for non-admins.  It happened to me on my dev VM somehow while testing
this PR.  That's why I did it here.

mo-nathan:
Not a big deal either way.  Was just curious why it came up.  I'm fine with
leaving it as part of this PR.

mo-nathan:
This diff let it show up on the main page:
`diff --git a/app/presenters/matrix_box_presenter.rb b/app/presenters/matrix_box_presenter.rb
index e60ee0509..0ab56431e 100644
--- a/app/presenters/matrix_box_presenter.rb
+++ b/app/presenters/matrix_box_presenter.rb
@@ -54,6 +54,7 @@ class MatrixBoxPresenter
       end
     return unless (temp = rss_log.detail)
 
+    temp = target.source_credit.tpl if target.respond_to?(:source_credit) && target.source.present? && target.source != "mo_website"
     # To avoid calling rss_log.detail twice
     self.detail = temp
   end
`
The conditional is pretty much the same as the one in app/views/observations/show.html.erb.  I think if that were refactored into a method in the Observation class that would make sense.

mo-nathan:
I'm going to look at what it might take to update the mobile app to do the right thing here.

pellaea:
Oh great!  I was wondering if anyone knew how to do that!  Probably
needless to say, it should be a trivial change: just add a source parameter
to the POST to /api2/observations.

pellaea:
> This diff let it show up on the main page:
> 
> ```
> diff --git a/app/presenters/matrix_box_presenter.rb b/app/presenters/matrix_box_presenter.rb
> index e60ee0509..0ab56431e 100644
> --- a/app/presenters/matrix_box_presenter.rb
> +++ b/app/presenters/matrix_box_presenter.rb
> @@ -54,6 +54,7 @@ class MatrixBoxPresenter
>        end
>      return unless (temp = rss_log.detail)
>  
> +    temp = target.source_credit.tpl if target.respond_to?(:source_credit) && target.source.present? && target.source != "mo_website"
>      # To avoid calling rss_log.detail twice
>      self.detail = temp
>    end
> ```
> 
> The conditional is pretty much the same as the one in app/views/observations/show.html.erb. I think if that were refactored into a method in the Observation class that would make sense.

Ah!! I understand now.  I think.  Yes, I'll do that.

JoeCohen:
Does anyone object to my merging this? (PRs are piling up again.)

pellaea:
No, please do.

config/locales/en.txt
```diff
@@ -1219,6 +1219,11 @@
   log_object_updated: "[:log_object_updated_by_user]"
   log_object_updated_at: "[:log_object_updated_by_user]"
 
+  source_credit_mo_website: Created using the "Mushroom Observer website":https://mushroomobserver.org.
+  source_credit_mo_android_app: Created using the "Mushroom Observer iPhone app":https://blah.de.blah.
+  source_credit_mo_iphone_app: Created using the "Mushroom Observer android app":https://blah.de.blah.
```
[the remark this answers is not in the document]
mo-nathan: I went ahead and fixed these.  I also fixed the mapping between the strings and the platform.

mo-nathan (approved):

mo-nathan:
I forgot to actually submit the review I did, so I just made the changes.  Still not used to that GitHub workflow.

pellaea:
Oh good!  I was just starting to fix them myself.

My thought concerning blahdeblah.com was to direct users straight to the
apple store or play store page for Olivia's app.  But I'm reconsidering.
You're probably right, the article is better.  Let's assume I approve of
your change. :)

mo-nathan:
Yeah, given that most folks are going to be looking at things from a web
browser that isn't on a mobile app, I think the article is the better
approach.

PR #238, opened 2016-08-29
Fix Pivotal #129185765 (blank tracking template messed up).

JoeCohen:
Deliver bug fix of [Pivotal #129185765](https://www.pivotaltracker.com/story/show/129185765).

The bug is described in more detail at [MO Developers › Fwd: [MO] Research Request](https://groups.google.com/forum/#!topic/mo-developers/tKE1zLQ_L40) and
[MO Developers ›Re: [MO] Mailing Address for Collections](https://groups.google.com/forum/#!topic/mo-developers/lkYesX_Jf8w).

Manual test script:
- Log in
- if your Profile (/account/profile) includes a Mailing Address for Collections, temporarily delete it.
- Go to any Name which you are not already tracking.
- Click on Email Tracking

Result:
- You should see the entire default Template.  (in master you see only 2 lines.)
- The template should include `**insert mailing address for specimens**`.  (in master, it instead includes `:mailing_address`.)

(When done, restore your Mailing Address for Collections.)

PR #1254, opened 2022-12-28
`ImagesController` CRUDify

nimmolo:
### Image Controllers CRUDified ###
- Images
  _Filenames, Licenses, Transformations, Votes, Votes/Anonymity_
- Account
  _Profile/Images_
- GlossaryTerms
  _Images_
- Observations
  _Images_

Note that `Image#add_image` and `Image#edit_image` are moved under `Observations::ImagesController` (because `new`, at least, specifically relates to images belonging to Observations and requires an obs id to get started.) `Image#reuse_image` and `Image#remove_image` are implemented differently on Observations, GlossaryTerms and Profile, but share the partials for reuse and removal.

### Integration tests added ###

**The most helpful thing other than a review would be to test drive the new/reuse image forms, as well as bulk filename purge, bulk license update, transformations,votes and bulk vote anonymity updater.**


app/models/image.rb
```diff
@@ -984,6 +984,83 @@ def self.update_copyright_holder(old_name, new_name, user)
...
+  #   {"license_count"=>1, "copyright_holder"=>"Tim Wheeler", "license_id"=>2}]
+  #   map(&:attributes) gives you a hash of your selects with their keys
+
+  def self.licenses_for_user_by_type(user)
+    Image.includes(:license).where(user_id: user.id).
+      select(Arel.star.count.as("license_count"),
+             :copyright_holder, :license_id).
+      group(:copyright_holder, :license_id).
+      map(&:attributes).map do |datum|
+      next unless (license = License.safe_find(datum["license_id"].to_i))
```
pellaea, line 1017: Oh, this cannot be good!  At very least do a join or ... wait, you're eager-loading, so why isn't it accessible ... because of the `map(&:attributes)`, that's throwing away the eager-loaded associations.  Oh geez, and look at what `License.current_names_and_ids` does.

How about taking the `includes(:license)` out and skip including the `license_name` and `current_names_and_ids` completely, and instead build a reference hash for all licenses -- there are only 3 or 4!! -- and get the names and stuff from that reference hash instead of trying to include it in this structure?

I guess this is grouping all records by `copyright_holder` and `license_id`, and 99.9% of users post only their own images... but what if some user decides to post images from thousands of copyright holders?  Yikes!  This query would be hideous as it's written.

What about this:

```
display_names = {}
current_names_and_ids = {}
License.each do |license|
  display_names[license.id] = license.display_name
  current_names_and_ids[license_id] = License.current_names_and_ids(license)
end
Image.where(user: user).
      select(Arel.star.count.as("license_count"),
             :copyright_holder, :license_id).
      group(:copyright_holder, :license_id).
      map(&:attributes).map do |datum|
        license_id = datum["license_id"].to_i
        datum["license_name"] = display_names[license_id]
        datum["licenses"] = current_names_and_ids[license_id]
        datum.except!("id")
      end
```

pellaea: (The rest looks good.)

pellaea (commented):
Still working on this review, too.  I'll see if I can get to these two again later today.

app/models/image.rb
nimmolo: Nice. Thanks for the edit! What i had there was just genersting the identical SQL to the original, and i'm glad you understand it well enough to rethink it.

nimmolo: And should i move that chunk of query to the Image (or License) model?

pellaea:
Excellent, thanks for the well-deserved little kick in the shin of my
pride. :)  I guess I should be happy, because this proves that I'm a better
programmer now, if I missed it back then.  Maybe getting old isn't all bad?

re: punt this to the Image model? -- Since you asked... I'd vote "nay".  As
I see it, this is particularly specialized for a single consumer.  I can't
imagine reusing this anywhere else.  To me that means it belongs in the
controller not the model.

nimmolo:
:) Getting better!My idea was the second consumer for everything like this will be GraphQL controller. The license updater does seem useful...On Dec 28, 2022, at 8:06 PM, Jason Hollinger ***@***.***> wrote:﻿
Excellent, thanks for the well-deserved little kick in the shin of my
pride. :)  I guess I should be happy, because this proves that I'm a better
programmer now, if I missed it back then.  Maybe getting old isn't all bad?

re: punt this to the Image model? -- Since you asked... I'd vote "nay".  As
I see it, this is particularly specialized for a single consumer.  I can't
imagine reusing this anywhere else.  To me that means it belongs in the
controller not the model.

pellaea:
Oh!  Move the whole license updater to the model (or module of the
model?)...  actually, sure, yes, that does make sense.  Sort of like
Name#merge and User#delete and other single-use things that are purely
model logic.  Okay, yes, that does make sense, especially in the context of
*everything* needing two entry points with graphql.  (Good reminder.)

Thanks for pushing back.  I agree with you now.

pellaea:
Test driving now...  Problems with the profile page:

```
account/users:
  check out the number of queries when you load my summary!!!
account/profile/edit:
  can't see the account_profile_remove_image "put" button
account/profile/images:
  image vote controls are hard to see
  clicking on images does nothing
  there's no way to "go back" to the edit profile page (no tabs at all; usually there's a "Cancel (Show Observation)" or equivalent tab in the upper right)
```

User summary -- this may be unrelated to your PR, would you like me to look into that?

nimmolo:
> Test driving now... Problems with the profile page:
> 
> ```
> account/users:
>   check out the number of queries when you load my summary!!!
> account/profile/edit:
>   can't see the account_profile_remove_image "put" button
> account/profile/images:
>   image vote controls are hard to see
>   clicking on images does nothing
>   there's no way to "go back" to the edit profile page (no tabs at all; usually there's a "Cancel (Show Observation)" or equivalent tab in the upper right)
> ```
> 
> User summary -- this may be unrelated to your PR, would you like me to look into that?

Yes please! At a family dinner rn.

pellaea:
I'm seeing lots of eager-loading issues, sigh.  Probably all predate this PR.  See, this is why I try never to actually use or test the site myself!  I find way too many bugs every time I do. :)

Glossary term stuff:

```
glossary_terms:
  image vote controls hard to see
glossary_terms/nnn/images/reuse:
  image vote controls hard to see (I won't mention this any more!)
  clicking on images does nothing
glossary_terms/nnn/images/remove:
  I'm seeing a weird outline around the "Okay" vote for the first image if there are two images, might happen in other circumstances, too?
  the bottom "Remove" button does nothing
```

pellaea:
Next installment:

```
/account/preferences/edit
  Change Image Vote Anonymity is hard to see
  as are some help(?) messages in the privacy section
/images/nnn
  controls at the top are also hard to see, just like the image vote controls
/images/transformations
  the action could be DRYed quite a bit:
    ----- 
    def transform_image_and_flash_notices(image)
      case params[:op]
      when "rotate_left", "rotate_right", "mirror"
        image.transform(params[:op].to_sym)
        flash_notice(:image_show_transform_note.t)
      else
        flash_error("Invalid operation #{params[:op].inspect}")
      end
    end
    -----
/images/nnn/edit
  Edit Translations links should be at the bottom, but are in the image column, probably meaning there is an unclosed div somewhere
  "Projects: Check any projects" is being repeated for each project user is in!
  notes are invisible or missing
/images/licenses/edit
  needs "back to preferences" tab or something like that
  If I try to change the name or license of any of my images it throws an error:
    def self.process_license_changes_for_user(user, updates)
      updates.each_value do |row|  <----------- updates is nil!
        next unless row_changed?(row)
      ...
```

nimmolo:
Fixed vote link color, clickable images, and weird outline thing. 
Simplified license query as suggested, and transformation action.

pellaea:
Super!  I don't know if I will be able to get back to this tomorrow...
shoot, actually for some time, maybe weeks.  I'll be in the field and on
the road for a week soon, and still have a few things I really need to wrap
up here first.  You might consider moving forward without me.

nimmolo:
Thanks. That sounds good, but to be honest i'm still stumped by the two remaining test errors on this branch. If you get a minute, thats where i'd like assistance!On Dec 29, 2022, at 1:59 AM, Jason Hollinger ***@***.***> wrote:﻿
Super!  I don't know if I will be able to get back to this tomorrow...
shoot, actually for some time, maybe weeks.  I'll be in the field and on
the road for a week soon, and still have a few things I really need to wrap
up here first.  You might consider moving forward without me.

pellaea:
Okay, I'll look those before I dive into untangling ugly herbarium database
conflicts today...

pellaea:
I figured out (more or less) the stupid StandardError failure at `test/controllers/images_controller_test.rb:109`.  Notice that the test is doing something screwy with stubs:

`ImagesController.any_instance.expects(:index).raises(StandardError)`

I have no idea what this does.  But it is obviously no longer working as intended.  Presumably the overall point of the test is to make sure that if advanced search raises a StandardError it gets redirected back to the advanced search form(?)  Presumably advanced search must as a matter of course raise StandardErrors for various reasons and we want those handled correctly.  Displayed in a flash?  Something?  That's where to start.

pellaea:
The remaining two errors are CSS-related (multiple submit buttons, unable to find a CSS element).  Those sound like your department.

nimmolo:
THANK YOU!On Dec 29, 2022, at 12:36 PM, Jason Hollinger ***@***.***> wrote:﻿
The remaining two errors are CSS-related (multiple submit buttons, unable to find a CSS element).  Those sound like your department.

JoeCohen:
GitHub blames me for the test that's doing something screwy with stubs.
It says:
>[Cover request for too many images](https://github.com/MushroomObserver/mushroom-observer/commit/6fe5ce1458fa39513a8aa743ea3cb311b0b7639a)
+22 −0 
[Cover ImageController #1125](https://github.com/MushroomObserver/mushroom-observer/pull/1125)
[JoeCohen](https://github.com/MushroomObserver/mushroom-observer/commits?author=JoeCohen) committed on Aug 17

I was preparing to refactor ImageController, and wanted to start by getting coverage to 100%.  (I kept the increased coverage, but eventually put the refactor aside.)

The "screwy" line is associated this the following commit message, but it doesn't jog my memory.
>Write test to cover error in Image#advanced_search
Test errors because of bug in `Image#advanced_search`
```ruby
ERROR ImageControllerTest#test_advanced_search_error (4.01s)
Minitest::UnexpectedError:         ActionController::UrlGenerationError: No route matches {:action=>"advanced_search", :controller=>"observations"}
            app/controllers/application_controller.rb:281:in `redirect_to'
            app/controllers/image_controller.rb:130:in `rescue in advanced_search'
            app/controllers/image_controller.rb:122:in `advanced_search'
            app/controllers/application_controller.rb:236:in `catch_errors_and_log_request_stats'
            test/functional_test_case.rb:22:in `get'
            test/controllers/image_controller_test.rb:312:in `test_advanced_search_error'
```
> I figured out (more or less) the stupid StandardError failure at `test/controllers/images_controller_test.rb:109`. Notice that the test is doing something screwy with stubs:
> 
> `ImagesController.any_instance.expects(:index).raises(StandardError)`
> 
> I have no idea what this does. But it is obviously no longer working as intended. Presumably the overall point of the test is to make sure that if advanced search raises a StandardError it gets redirected back to the advanced search form(?) Presumably advanced search must as a matter of course raise StandardErrors for various reasons and we want those handled correctly. Displayed in a flash? Something? That's where to start.

nimmolo:
You guys seem to have a tool that tells you coverage line by line. How do you do that?

JoeCohen:
I probably was covering the rescue in `app/controllers/images_controller.rb#advanced_search`:
```ruby
  def advanced_search
    return if handle_advanced_search_invalid_q_param?

    query = find_query(:Image)
    show_selected_images(query)
  rescue StandardError => e
    flash_error(e.to_s) if e.present?
    redirect_to(search_advanced_path)
  end
```

mo-nathan:
There should be a directory called `coverage` at the top level of your source tree that contains an index.html file.  If you open that in a browser the rest should be clear.  It should get refreshed each time you run `rails test`.  If it doesn't you might try `SIMPLECOV=1 rails test`, but that doesn't seem to be necessary any more.

JoeCohen:
> There should be a directory called `coverage` at the top level of your source tree that contains an index.html file. If you open that in a browser the rest should be clear. It should get refreshed each time you run `rails test`. If it doesn't you might try `SIMPLECOV=1 rails test`, but that doesn't seem to be necessary any more.

And if the tests pass in CI (unfortunately not yet true for this PR), then Github also has line-by-line coverage.
Click the "Details" link to the right of "coverage/coveralls"
Screenshot 2022-12-29 at 8 16 01 PM

pellaea:
Let me pile on!  Everything so far sounds correct.  The "screwy" line --
sorry! no judgment intended! I just don't understand the syntax of those
stubs-like mechanisms, they always look "screwy" to me :) -- is clearly
intended to cause the failure which will result in those rescue lines being
executed.  I just don't know *how* it's doing so.  And I don't know where
to look up the docs.  I've seen "stubs" before, but not this "expect"
thing.  Joe, you must have learned that trick somewhere, at least
briefly... do you remember where???  Where is all the stubbing-like stuff
documented, anyway?  I don't know how to google it.

JoeCohen:
Re: `ImagesController.any_instance.expects(:index).raises(StandardError)`

It's my lazy way to stub instance methods. I agree that the syntax is screwy. And I have to look it up each time.

Stubbing class methods in minitest is straighforward. `Class.stub(:method, response)`. https://github.com/minitest/minitest#stubs-
For instance methods, you can stub a particular object instance: `instance.stub(:method, response)`. https://stackoverflow.com/questions/7211086/how-do-i-stub-things-in-minitest
But I'm lazy/easily confused. So I use the `mocha` gem to stub all instances of a class. https://github.com/freerange/mocha#quick-start

I'll try to have a look at this failure before 7 a.m. or after 2 p.m. Anyone else is welcome to dig into it.

JoeCohen:
The currently failing test used to be:
```ruby
  def test_advanced_search_error
    query = Query.lookup_and_save(:Image, :advanced_search,
                                  name: "Don't know",
                                  user: "myself",
                                  content: "Long pink stem and small pink cap",
                                  location: "Eastern Oklahoma")
    ImageController.any_instance.expects(:show_selected_images).
      raises(StandardError)
    login

    get(:advanced_search, params: @controller.query_params(query))
    assert_redirected_to(search_advanced_path)
  end
```
The change from stubbing `show_selected_images` to stubbing `index` looks right.
But what about the change from ` get(:advanced_search ...` to `get(:index ...`?

JoeCohen:
The  change from  `get(:advanced_search ...` to `get(:index ...` does look correct. 
Need to dig deeper, but might not have time this morning.

pellaea:
Thanks, Joe!  Those pointers to the mocha gem etc. docs should be enough
for me to get unstuck.  I'll take it from here!

pellaea:
Yes, okay, this totally makes sense now.  It's raising the error too
early.  We need to ensure a StandardError is raised within the block that
is rescued inside ImagesController#advanced_search.  Why not stub
"show_selected_images" instead, for example?  ... Yep, that works.  Easy
peasy!

(Yeah, this mocha gem was what confused me.  I see now that "expects" is
just a synonym essentially for "stubs"... but somehow isn't intuitive for
me.  But that's okay, now that I know it's just using a different but
equivalent mechanism, I'm okay with it.  Thanks again!)

.rubocop.yml
```diff
@@ -266,6 +266,15 @@ Style/SymbolArray:
...
+    - index
+    - show
+    - new
+    - create
+    - edit
+    - update
+    - destroy
 Rails/SkipsModelValidations:
   # We skip validations all the time.  I'm tired of disabling this cop.
   Enabled: false
```
JoeCohen, line 280: Suggestions:
1. Move these settings up to where the file has the other settings Rails department cops. (The out-of place setting might be my fault.)
2. Note that we have two configurations for `Rails/ActionOrder`. They should be combined, at least in this file. Otherwise we get this warning when running RuboCop:
>.rubocop.yml:159: `Rails/ActionOrder` is concealed by line 269

3. I prefer the settings above, which comment out `edit` because that's more flexible, resulting in no offenses.
If we demand that `create` appear before `edit`, we end up with 6 offenses, which should be fixed.  (With the default order there are 8 offenses.)
>920 files inspected, 6 offenses detected:
[app/controllers/articles_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/articles_controller.rb)
[app/controllers/glossary_terms_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/glossary_terms_controller.rb)
[app/controllers/herbaria_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/herbaria_controller.rb)
[app/controllers/publications_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/publications_controller.rb)
[app/controllers/visual_groups_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/visual_groups_controller.rb)
[app/controllers/visual_models_controller.rb - 1 offense](https://github.com/MushroomObserver/mushroom-observer/pull/1254/files#offense_app/controllers/visual_models_controller.rb)

Options for fixing include:
- change the action order in the offending files. Issues:
  -  I deliberately use the default order
```ruby
 # ---------- Actions to Display forms -- (new, edit, etc.) -------------------

  def new
...
  def edit
...
  # ---------- Actions to Modify data: (create, update, destroy, etc.) ---------
  def create
...
  def update
...
```
  - Based on the `visual` stuff, Nathan (and other devs) will probably naturally choose the default order.
- disable the cop inline in the offending files.
- add additional .rubocop.yml files in the appropriate directories.

app/assets/stylesheets/mo/_map_theme_vars_to_bootstrap_vars.scss
```diff
@@ -156,4 +156,4 @@ $panel-default-heading-bg: $LIST_HEADER_BG_COLOR;
 
 $hr-border: $LIST_BORDER_COLOR;
 
-$well-bg: $WELL_BG_COLOR;
```
JoeCohen, line 159: - Add newline at EOF. 
- Unless it messes you up on other projects, I suggest enabling the VSCode `files.insertFinalNewline` setting

app/assets/stylesheets/mo/_utilities.scss
```diff
@@ -380,5 +421,4 @@
   .mt-sm-3 {
     margin-top: 1em !important;
   }
-}
-
+}
```
JoeCohen, line 424:
```suggestion
}

```
Insert newline at EOF

JoeCohen:
Half-baked idea for the future.
What about 
- One controller that creates (uploads), edits, deletes images
- Another controller that creates, edits, and deletes uses of images
- An image_uses table with one entry per image use, kind of like the `comments` table. This would replace both `glossary_term_images` and `observation_images`, and give us the option of attaching images to other objects (e.g. articles). This would also handle profile image.

mo-nathan:
I can see having an ImagesController and an ImageUsesController (or whatever you want to call it).  I'm not a big fan of an image_uses model though.  It seems like it is essentially doing exactly what the standard many-to-many foreign keys do, but in a less standard way that would probably require more expensive SQL (each query of image_uses requires some sort of "type" field, eg, comment.target_type).  At work one of the core objects uses a mechanism like you're describing and it's one of the biggest and hardest problems in our code base.  Conceptually you're starting down the path of polymorphic objects which leads to key/value tables which has in turn created the whole concept of "no SQL" databases.  Not a path to be taken lightly and I don't see a compelling reason for it our current data model.

nimmolo:
I know what you mean.  I was trying to solidify an `ImagesController` CRUD independent of the parent object, here, but ultimately bailed. (I moved those actions back and forth between the `ImagesController` and `Observations::ImagesController` three separate times!) Still like the idea, but let's do it in a future PR. Seems like a perfect operation to encapsulate with Trailblazer. 

To describe the current situation more precisely, as you probably noticed: Actually, only new/create requires an observation, and this adds observation-specific fields to the image form. The edit/update is independent of the image's parent and does not require a parent object ID.

There is no new/create action in the `GlossaryTerms::ImagesController` or the `Account::Profile::ImagesController` for uploading images directly to those objects, because in those cases it's handled by the `AjaxController`.

The last target of my CRUD is going to be to break up the AjaxController and move the actions to the respective controller CRUD actions, so that 
- similar CRUD logic is in the same place, easier to reason about and easier to track down in a the logic service object refactor
- the templates returned by the AJAX call are all flattened into template erb files for the Bootstrap 4 refactor, no more HTML hiding in JS

app/controllers/emails_controller.rb
```diff
@@ -15,7 +15,7 @@ def ask_webmaster_question
     @email = @user.email if @user
   end
 
-  def ask_user_question
+  def ask_user_question # rubocop:disable Metrics/AbcSize
```
JoeCohen, line 18: I think this is something for another PR. But just in case:
We could get rid of the ABC offense by refactoring UserMail.build to take kwargs and eliminating unnecessary local variable assignments. 
See https://www.pivotaltracker.com/story/show/184120794 and https://www.pivotaltracker.com/story/show/184124094

.rubocop.yml
nimmolo: I think i did that provisionally, when you first added the cop. Will remove

I have to say that with these fat controllers I'm working on, I can't understand why anyone would prefer the default order. It's great that the new Visual controllers are skinny, but for all the other CRUD controllers i've had to hack through, if you read through them, I think you will find it is way too much scrolling to try to follow logic from `new` to `create`, via `edit`.  Moving the private methods to the end doesn't seem helpful either — I tried that. (I don't know, even though I have a scroll mouse i have lost a lot of time and frustration just through scrolling, and I didn't pick this organization without experimentation.)

When we move to service objects it won't matter though, because they will be so slim that skimming the controller methods won't even require scrolling.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,328 @@
...
+    return unless project
+
+    query = create_query(:Image, :for_project, project: project)
+    show_selected_images(query, always_index: 1)
+  end
+
+  # Display matrix of images whose notes, names, etc. match a string pattern.
+  def image_search
+    pattern = params[:pattern].to_s
+    if pattern.match(/^\d+$/) &&
```
JoeCohen, line 119:
```suggestion
    if pattern.match?(/^\d+$/) &&
```
See https://docs.rubocop.org/rubocop-performance/cops_performance.html#performanceregexpmatch
(I don't know why rubocop doesn't flag this.)

nimmolo: Oh wow. Great catch. 

These are all probably over the CRUD controllers, in every index. 
I will commit this now, and make another PR to get them elsewhere.

app/assets/stylesheets/mo/_map_theme_vars_to_bootstrap_vars.scss
nimmolo: Thanks. I think these sneaked in while i was trying to set up Rubocop on the other machine.

nimmolo:
Note: I've added a `destroy image` button to the tabset on the Edit Image page, in order to pass my own integration test. But at this point I have no idea why my integration test expected to find a destroy button there in the first place. 

I'm happy to remove both the button and the expectation in the test, or have it navigate back to the show image template and click the destroy button from there. Two considerations:

- It kind of makes sense that there would be a delete button on an edit page, but we don't have those elsewhere, and we didn't have one here before either.
- I'm not trying to change the UI in these PRs, as tempting as that is. 

Inclined to remove the button and alter the test, but maybe this is as good a time as any to ask for feedback about destroy links and where they should be!

app/controllers/images_controller.rb
JoeCohen: Our `.rubocop.yml` may explain why rubocop doesn't catch this:
`something.match?(regexp)` throws an error if `something` is `nil`.
That won't be true here.

app/controllers/emails_controller.rb
nimmolo: I like this idea, and I hope you will tackle it! 

I'm also wanting to convert `assert_request` and the `either_requires_either` feeder methods into kwargs.

app/controllers/images_controller.rb
nimmolo: Can we be sure the pattern won't be `nil`?

nimmolo:
Whew, this is passing tests!  I'm going to commit one more TODO just for codeclimate's enjoyment :)

app/controllers/images_controller.rb
pellaea: Yes, the `.to_s` above ensures it is not nil.

nimmolo: Thanks @pellaea , that's good to know.

nimmolo:
@JoeCohen I'll merge as soon as i hear from you.

app/controllers/glossary_terms/images_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+
+# Clicking on an image currently fires a GET to these actions... because it
+# comes from a link made by thumbnail_helper#thumbnail(link: url_args)
+# with CRUD refactor, change thumbnail helper to fire a POST somehow?
+
+module GlossaryTerms
+  class ImagesController < ApplicationController
+    before_action :login_required
+
+    before_action :login_required
```
JoeCohen, line 11: Suggestion: Remove duplicate lines
```suggestion
    before_action :login_required
```

app/controllers/glossary_terms/images_controller.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+# Clicking on an image currently fires a GET to these actions... because it
+# comes from a link made by thumbnail_helper#thumbnail(link: url_args)
+# with CRUD refactor, change thumbnail helper to fire a POST somehow?
+
+module GlossaryTerms
+  class ImagesController < ApplicationController
```
JoeCohen, line 8: Could this be a little more CRUDified by making it AttachmentsController, 
with actions
~~reuse~~ new
~~attach~~ create
~~remove~~ edit
~~detach~~ update
(This is one of the places I stumbled when I tried refactoring this.)

app/controllers/glossary_terms/images_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+    def remove
+      @object = find_or_goto_index(GlossaryTerm, params[:id].to_s)
+      return unless @object
+
+      return if check_permission!(@object)
+
+      redirect_with_query(glossary_term_path(@object.id))
+    end
+
+    def detach
```
JoeCohen, line 78: If keeping non-standard action names, perhaps add a comment here like
```suggestion
    # The remove form submits to this action
    def detach
```

app/controllers/images/votes/anonymity_controller.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
```
JoeCohen, line 1: For future reference:
We should get rid of anonymous voting (both for images and proposed names).
See https://www.pivotaltracker.com/story/show/153902213, https://www.pivotaltracker.com/story/show/155106697
For image votes @pellaea wanted to cut the cord, I wanted to phase it out. 
If cutting the cord, this might be a good opportunity?
But likely beyond the scope of this PR.

app/controllers/glossary_terms/images_controller.rb
nimmolo: Thanks for really considering this. (I initially did try it with just those CRUD _action_ names, but without your more clarifying _controller_ name `AttachmentsController`... I could just picture Jason looking over my shoulder and asking, "Are those names really clearer?" and thought, no indeed... `edit` is a misleading name in an `ImagesController` for a `detach` action. But it makes sense for an `AttachmentsController`.)

If you would like to commit those changes to all the attachment controllers in this branch, please do. Or, maybe better to do that in another PR. 

Don't consider anything in any of these controllers PRs set in stone, I'm thinking of the controller directory as one giant TODO list. I hope they will be in a constant state of digestion until the logic is in service objects and can be used for GraphQL directly.

app/controllers/glossary_terms/images_controller.rb
nimmolo: Hah! Good suggestion :)

app/controllers/observations/images_controller.rb
```diff
@@ -0,0 +1,369 @@
...
+    def create_image
+      if params[:upload].blank?
+        flash_warning(:runtime_no_changes.t)
+      else
+        args = params[:image]
+        i = 1
+        while i < 5 || params[:upload]["image#{i}"].present?
+          process_image(args, params[:upload]["image#{i}"])
+          i += 1
+        end
```
JoeCohen, line 81: First time I noticed this. What is this doing? Does it need a comment?

app/controllers/images/votes/anonymity_controller.rb
nimmolo: I like it. I'm trying to keep removing/altering functionality outside these PRs, but i hope they make alterations easier to achieve. 

To me that's a nice thing about the controller (and eventual service object) split up — it's hopefully more evident, if we're taking out functionality, what the likely boundaries of its effects are.

app/controllers/observations/images_controller.rb
nimmolo: I believe: There are five upload spots on the form. This says, for as many inputs as have an image, process the image. 

Adding comment now

JoeCohen:
Maybe use an explaining constant, which is referenced here and in the form.

app/controllers/observations/images_controller.rb
```diff
@@ -0,0 +1,369 @@
...
+      # Don't know how to correctly test this.
+      @image.when = @observation.when
+      @image
+    end
+
+    def create_image
+      if params[:upload].blank?
+        flash_warning(:runtime_no_changes.t)
+      else
+        args = params[:image]
```
nimmolo, line 76:
```suggestion
        # There are five upload spots on the form. 
        # For as many as have an image, process the image. 
        args = params[:image]
```

app/controllers/images/filenames_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Images::FilenamesController
+module Images
+  class FilenamesController < ApplicationController
+    before_action :login_required
+    # NO VIEW TEMPLATE
+    # Linked from account/preferences/_privacy
+    # Move to new controller Images::FilenamesController#update
+    # Move test from images_controller_test
```
nimmolo, line 10:
```suggestion
```

app/controllers/images/licenses_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# Images::LicensesController
+module Images
+  class LicensesController < ApplicationController
+    # def access_denied
+    #   flash_error(:permission_denied.t)
+    #   redirect_to(edit_account_preferences_path)
+    # end
+
```
nimmolo, line 10:
```suggestion
```
Remove outdated comment. I was playing with where it would send you if you weren't authorized.

app/controllers/observations/images_controller.rb
```diff
@@ -0,0 +1,369 @@
...
+          flash_notice(:attached_to_project.t(object: :image,
+                                              project: project.title))
+        else
+          project.remove_image(img)
+          flash_notice(:removed_from_project.t(object: :image,
+                                               project: project.title))
+        end
+        any_changes = true
+      end
+      any_changes
```
JoeCohen, line 172: I think we could get rid of the `any_changes` flag with the following suggestion. On the other hand, the flag does explain why it's returning `true` or `false`. Maybe ditch the flag and add a comment before the definition?
```suggestion
      projects.each do |project|
        before = img.projects.include?(project)
        after = checks["id_#{project.id}"] == "1"
        next if before == after

        if after
          project.add_image(img)
          flash_notice(:attached_to_project.t(object: :image,
                                              project: project.title))
        else
          project.remove_image(img)
          flash_notice(:removed_from_project.t(object: :image,
                                               project: project.title))
        end
        true
      end
      false
```

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,328 @@
...
+#  edit_image::            Edit notes, etc. for image.
+#  transform_image         Rotate or flip image
+#  destroy_image::         Callback: destroy image.
+#  remove_image::          Callback: remove image from observation.
+#  reuse_image::           Choose images to add to observation.
+#  remove_images::         Choose images to remove from observation.
+#  remove_images_for_glossary_term Choose images to remove from GlossaryTerm.
+#  license_updater::       Change copyright of many images.
+#  bulk_vote_anonymity_updater:: Change anonymity of image votes in bulk.
+#  bulk_filename_purge::   Purge all original image filenames from the database.
```
nimmolo, line 37:
```suggestion
#  new::                   Upload images for observation.
#  edit::                  Edit notes, etc. for image.
#  create::                New form commits here
#  update::                Edit form commits here
#  destroy::               Callback: destroy image.
```

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,328 @@
...
+#  advanced_search::       matching Advanced Search results
+#  index_image::           current query
+#  show_selected_images::  search results
+#
+#  ==== Show Images
+#  show::
+#  show_original::         show full_size image (for backwards compatibility)
+#  next_image::
+#  prev_image::
+#  cast_vote::             change user's vote & go to next image
```
nimmolo, line 24: Update method outline for `:show`
```suggestion
```

nimmolo, line 24: Update method outline for `:show`
```suggestion
#  show::                  can use params[:flow] :next, :prev
```

app/controllers/observations/images_controller.rb
```diff
@@ -0,0 +1,369 @@
...
+      img.observations.each do |obs|
+        obs.projects.each do |project|
+          projects << project unless projects.include?(project)
+        end
+      end
+
+      attach_images_to_projects_and_flash_notices(img, projects, checks)
+    end
+
+    def attach_images_to_projects_and_flash_notices(img, projects, checks)
```
nimmolo, line 154: Add comment about what this method returns
```suggestion
    # Returns true if any changes made, false if none
    def attach_images_to_projects_and_flash_notices(img, projects, checks)
```

app/controllers/observations/images_controller.rb
nimmolo: I'm not sure this suggestion will work. After the `projects.each` iterator, it still will hit the `false` line, no?

JoeCohen:
Sorry to leave this comment here, but GitHub won't let me put it in the right place.
The Brakeman warning (quoted below) for app/models/image.rb line 746 looks like a false positive to me:
- `operator` is restricted to one of three valid strings
- `id` is used by the script only as the filename.

If you agree, then get Brakemen to ignore it. See https://brakemanscanner.org/docs/ignoring_false_positives/
>Check warning on line 746 in app/models/image.rb
GitHub Actions
/ Brakeman
Medium - Execute
Possible command injection

JoeCohen:
@nimmolo:
I'm done looking at this PR. (Warning: I mostly ignored the `.erb` files.)
It's a huge improvement. It's easy on the eyes and brain. Thanks so much for all this work.
My only remaining suggestion is to add coverage to
- `/app/controllers/glossary_terms/images_controller.rb#detach`  (and any other uncovered lines that you want to take the time to deal with). See https://coveralls.io/builds/55549711/source?filename=app%2Fcontrollers%2Fglossary_terms%2Fimages_controller.rb#L75
 `/app/controllers/account/profile/images_controller.rb#detach` (and any other uncovered lines that you want to take the time to deal with). See https://coveralls.io/builds/55549711/source?filename=app%2Fcontrollers%2Faccount%2Fprofile%2Fimages_controller.rb#L25

JoeCohen:
> Sorry to leave this comment here, but GitHub won't let me put it in the right place. The Brakeman warning (quoted below) for app/models/image.rb line 746 looks like a false positive to me:
> 
> * `operator` is restricted to one of three valid strings
> * `id` is used by the script only as the filename.
> 
> If you agree, then get Brakemen to ignore it. See https://brakemanscanner.org/docs/ignoring_false_positives/
> 
> > Check warning on line 746 in app/models/image.rb
> > GitHub Actions
> > / Brakeman
> > Medium - Execute
> > Possible command injection

I take it back. The script also has this line:
```bash
if [[ $1 == "-h" || $1 == "--help" ]]; then cat <<END; exit 1; fi
```
and I don't know if it could be used for command injection.

JoeCohen:
I started to fix some of the Codeclimate issues, and noticed a few more things.
I'll leave comments in the files

JoeCohen:
app/models/image.rb 
I think these are all old, easily fixable offenses which could be easily corrected in this PR without expanding its scope.
Rubocop says (in part):
Line #238 – Convention: Rails/HasManyOrHasOneDependent: Specify a :dependent option.
  has_many :thumb_clients, class_name: "Observation",
Line #245 – Convention: Rails/HasManyOrHasOneDependent: Specify a :dependent option.
  has_many :subjects, class_name: "User"
Line #246 – Convention: Rails/HasManyOrHasOneDependent: Specify a :dependent option.
  has_many :best_glossary_terms, class_name: "GlossaryTerm",
Line #449 – Warning: Lint/DuplicateMethods: Method Image#image_dir is defined at both app/models/image.rb:447 and app/models/image.rb:449.
  def image_dir

nimmolo:
@JoeCohen Thanks for the commits. I added a test for removing an image from GlossaryTerm. 

Will merge if this passes.

PR #1256, opened 2022-12-31
`SessionExtensions` — pass args to `open_form` from methods that use it

nimmolo:
## For integration tests ##

`SessionExtensions` test method `open_form` accepts args to specify the form, but defaults to open the **first form on the page**. 

`SessionExtensions` helper methods `assert_form_has_correct_values` and `submit_form_with_changes` have so far been able to assume the defaults, but with CRUD controllers they're not specific enough. (There are other forms on the page, like delete links.)

This just adds an arg to those methods, and passes a default.

### DIFF
--- a/test/session_extensions.rb
+++ b/test/session_extensions.rb
@@ -211,8 +211,8 @@ module SessionExtensions
   #
   ##############################################################################
 
-  def assert_form_has_correct_values(expected_values)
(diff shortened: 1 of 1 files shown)


nimmolo:
Merging this because it passes tests and doesn't change anything materially. 

Isolates a change from #1254

PR #1257, opened 2023-01-01
Change `pattern.match(str)` to `pattern.match?(str)` where applicable

nimmolo:
@JoeCohen caught this on #1254, but applicable elsewhere.

Reviewers - please check if this change is also applicable in `name/parse` and `name_test` as i've done it here. 
I may have misunderstood something.

### DIFF
--- a/app/models/name/parse.rb
+++ b/app/models/name/parse.rb
@@ -323,7 +323,7 @@ module Name::Parse
 
   def parse_below_genus(str, deprecated, rank, pattern)
     results = nil
-    if (match = pattern.match(str))
+    if (match = pattern.to_s.match?(str))
       name = match[1]
       author = match[2].to_s
       name = standardize_sp_nov_variants(name) if rank == "Species"
--- a/app/models/name/parse.rb
+++ b/app/models/name/parse.rb
@@ -323,7 +323,7 @@ module Name::Parse
 
   def parse_below_genus(str, deprecated, rank, pattern)
     results = nil
-    if (match = pattern.to_s.match?(str))
+    if (match = pattern.match(str))
       name = match[1]
       author = match[2].to_s
       name = standardize_sp_nov_variants(name) if rank == "Species"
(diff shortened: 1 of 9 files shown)


app/models/name/parse.rb
```diff
@@ -323,7 +323,7 @@ def parse_genus_or_up(str, deprecated = false, rank = "Genus")
 
   def parse_below_genus(str, deprecated, rank, pattern)
     results = nil
-    if (match = pattern.match(str))
+    if (match = pattern.to_s.match?(str))
```
JoeCohen, line 326:
```suggestion
    if (match = pattern.match(str))
```
We need the `match` method here because the subsequent lines expect the `match` local variable to be a MatchData object, rather than true/false.
See https://ruby-doc.org/3.1.2/Regexp.html#method-i-match

test/models/name_test.rb
```diff
@@ -55,7 +55,7 @@ def do_name_parse_test(str, expects)
   end
 
   def assert_name_match_author_required(pattern, string, first_match = string)
-    assert_not(pattern.match(string),
+    assert_not(pattern.to_s.match?(string),
```
JoeCohen, line 58: I don't have time to look into this further tonight, but:

1. If this method is being called with a regexp as the `pattern` argument, I suspect that the assertion won't work correctly if `pattern` is converted to a string.
2. On the extreme, if the method is always called with a regex at the `pattern` argument, then you don't need `to_s` because a regex cannot be `nil` -- `nil` is in its own Ruby class.

test/models/name_test.rb
```diff
@@ -96,7 +96,7 @@ def assert_name_match_various_authors(pattern, string, first_match)
   end
 
   def assert_name_match(pattern, string, first, second = "")
-    match = pattern.match(string)
+    match = pattern.to_s.match?(string)
```
JoeCohen, line 99:
```suggestion
    match = pattern.match(string)
```
The subsequent lines require the `match` local variable to be a MatchData object.

JoeCohen (changes requested):

test/models/name_test.rb
nimmolo: sorry i was on autopilot. thanks for catching this.

PR #1259, opened 2023-01-02
`SpeciesListsController` CRUDify

nimmolo:
### SpeciesList Controllers CRUDified ###
- SpeciesLists
  _Downloads, NameLists, Observations, Projects, Uploads_
- Observations
  _SpeciesLists_

Note that `manage_species_lists` is moved under `Observations::SpeciesListsController` (it's linked from an obs view, and really about removing an Obs from a Species list.) `add_remove_observations` is moved to `SpeciesLists::ObservationsController#new`, and handles the addition or removal of a _query-full_ of observations from a species list.

One non-CRUD action has been kept, `SpeciesListsController#clear`. I guess it could be `#update` with a `params[:clear]`, couldn't it?

**The most helpful thing other than a review would be to test drive the forms.**


nimmolo:
@pellaea — The only failure now is a Bullet regarding name synonyms. 
I tried fixing, but i couldn't find the spot to `includes([:synonym])`

pellaea:
Feel free to disable Bullet for that test!  Those can be damned hard to fix.

app/controllers/location_controller.rb
```diff
@@ -548,7 +548,7 @@ def create_location
                       action: :show,
                       id: @set_observation)
         elsif @set_species_list
-          redirect_to(controller: :species_list, action: :show_species_list,
+          redirect_to(controller: :species_lists, action: :show,
                       id: @set_species_list)
```
JoeCohen, line 552: I should have picked up on this earlier (in ProjectsController)
If trying to use helper paths when possible, then
```suggestion 
          redirect_to(species_list_path(@set_species_list)
```

app/controllers/observations_controller/index.rb
```diff
@@ -230,7 +230,8 @@ def define_index_links(query)
     @links << [
       :list_observations_add_to_list.t,
       add_query_param(
-        { controller: :species_list, action: :add_remove_observations },
+        { controller: "/species_lists/observations",
+          action: :edit },
         query
       )
```
JoeCohen, line 236:
```suggestion
      add_query_param(edit_species_list_observations_path, query)
```

app/controllers/location_controller.rb
nimmolo: Thanks for catching these. I think i intentionally skipped them to avoid having extra conflicts to resolve on the upcoming Names/Locations. Double-checking now... yes they are corrected there.

Coming soon to a console near you! (I hope!)

app/controllers/species_lists/projects_controller.rb
```diff
@@ -0,0 +1,191 @@
...
+      if check_permission!(@list)
+        @projects = projects_to_manage
+        @object_states = manage_object_states
+        @project_states = manage_project_states
+        commit_and_redirect
+      else
+        redirect_to(species_list_path(@list.id))
+      end
+    end
+
```
JoeCohen, line 37:
```suggestion

    ############################################################################

    private

```

app/controllers/species_lists/projects_controller.rb
```diff
@@ -0,0 +1,191 @@
...
+    def projects_to_manage
+      projects = @user.projects_member
+      if @list.user == @user
+        projects += @list.projects
+        projects.uniq!
+      end
+      projects
```
JoeCohen, line 44:
```suggestion
      if @list.user == @user
        @user.projects_member.union(@list.projects)
      else
        @user.projects_member
      end
```

app/controllers/species_lists/projects_controller.rb
```diff
@@ -0,0 +1,191 @@
...
+        img: params[:objects_img].present?
+      }
+    end
+
+    def manage_project_states
+      states = {}
+      @projects.each do |proj|
+        states[proj.id] = params["projects_#{proj.id}"].present?
+      end
+      states
```
JoeCohen, line 63: Your choice: (a) implement the following which I like because it's shorter and may be faster, (b) leave as-is.
WARNING: @pellaea dislikes `each_with_object`
```suggestion
      @projects.each_with_object({}) do |proj, states|
        states[proj.id] = params["projects_#{proj.id}"].present?
      end
```

app/controllers/species_lists/projects_controller.rb
```diff
@@ -0,0 +1,191 @@
...
+
+        flash_warning(:runtime_no_changes.t)
+
+      else
+        flash_error("Invalid submit button: #{params[:commit].inspect}")
+      end
+      render(:edit)
+    end
+
+    def attach_objects_to_projects
```
JoeCohen, line 87: It would be nice to get rid the `@any_changes` flag. 
I haven't come up with anything yet. Maybe something for the future.

app/controllers/species_lists/projects_controller.rb
nimmolo: Question about `each_with_object`: Does `states` get set to `{}` even if there are no `@projects`?

app/controllers/species_lists/projects_controller.rb
JoeCohen: Maybe have each of the sub-methods return a truthy value instead of setting a global flag.
And in `attach_objects_to_projects` have a local, memoized flag? 
There's still a flag, but at least it's local.
Still might be something for the future.

app/controllers/species_lists/projects_controller.rb
JoeCohen: Great question. I hadn't thought about that.
If `@projects` is any Enumerable, including an empty Array, it works fine.
```ruby
irb(main):002:0> @projects = []
=> []
irb(main):003:0> @projects.each_with_object({}) {|proj, states| states[proj.id] = params["projects_#{proj.id}"].present? }
=> {}
```
If @projects is `nil` (or any non-Enumerable) it throws an error
```ruby
irb(main):001:0> @projects.each_with_object({}) {|proj, states| states[proj.id] = params["projects_#{proj.id}"].present? }
(irb):1:in `<main>': undefined method `each_with_object' for nil:NilClass (NoMethodError)
                                                                                                                                               
@projects.each_with_object({}) {|proj, states| states[proj.id] = params["projects_#{proj.id}"].present? }                                      
         ^^^^^^^^^^^^^^^^^                                                                                                                     
```

app/controllers/species_lists/projects_controller.rb
nimmolo: If this logic moves to trailblazer, which i hope we can start doing soon, ivars like will be all turned into context args, to be passed between service objects, so i say let's leave it for now.

JoeCohen: Agree. Even if we don't go to Trailblazer soon, the refactor would (at least for me) require enough thought so that it best be done elsewhere.

PR #1268, opened 2023-01-07
Really fix Comment email test error

nimmolo:
The new test for name tracker emails turns `queue_emails(true)` on, but forgets to set it back to `false`.

It seems that this only causes an issue if the `CommentTest` runs after `NameControllerTest`.


JoeCohen (approved):
Great work! These order-dependent failures are a pain to track down and fix.
I think you should merge it ASAP in order to prevent failures in other branches.

JoeCohen:
Hang on a second. There may be a an issue with my commit. Double-checking now.

JoeCohen:
False alarm. Everything good. (But I just pushed a commit that deletes an old comment. It's fine to merge when the tests pass. And ignore the epsilon purported drop in coverage.)

PR #1269, opened 2023-01-07
`NamesController` `LocationsController` CRUDify

nimmolo:
### Name and Location Controllers CRUDified ###
**Non-CRUD actions namespaced and CRUDified**
- Locations
  _Countries, Descriptions, Help, Maps, Merges, Observations, ReverseNameOrder, Versions_
  - Descriptions
  _Defaults, Merges, Moves, Versions_

- Names
  _Classification, Descriptions, EolData, Lifeforms, Maps, Synonyms, Trackers, Versions_
  - Classification
  _Inherit, Propagate, Refresh_
  - Descriptions
  _Defaults, Merges, Moves, Permissions, Publish, ReviewStatus, Versions_
  - EolData
  _ExpandedReview, Preview_
  - Lifeforms
  _Propagate_
  - Synonyms
  _Approve, Deprecate_
  - Trackers
  _Approve_


nimmolo:
Yahoo! This PR is green. I'll be looking where coverage went down, to write some new tests. Also, working on some codeclimate stuff.

Meanwhile, please do test drive if you are so inclined.

nimmolo:
A couple notes:

- Yes, this is a massive PR. After working for months on a previous attempt at `NamesController`, I realized it's not worth it to break it up. If i try to do Names but not Locations, the exceptions I have to write for descriptions are confusing and kind of absurd, since they're more complex and they're temporary (until a Locations PR).
- This uses concerns. I considered alternatives and I prefer concerns because they're core Rails, so they get autoloaded from the controllers directory with Zeitwerk, and they're a well-documented way for controllers to share functionality. 
- If we move to Trailblazer or another framework, the question of sharing logic via concerns will become obsolete.

nimmolo:
New test coverage for description merges, moves, permission adjustments, publishing drafts, reverse location name.

Note: This PR now proposes removing controllers for `LocationDescription` permissions and drafts. They may be useful for NameDescriptions, but this level of control does not seem likely to be useful for Location descriptions.

@JoeCohen Do you think we need these?

nimmolo:
Ready to rumble!

mo-nathan:
Formatting issue when I got to http://localhost:3000/names/402/versions?q=1oAhh&version=2.
Screenshot 2023-01-20 at 7 56 25 PM

nimmolo:
Formatting issue fixed, plus ABC suggestions in `versions_helper.rb`

nimmolo:
Thank you @mo-nathan for clicking around.

I hope it's clear that i'm not expecting code review here!

I'm asking for everyone to spend 5 minutes clicking around Names and Locations on this branch, to be sure everything is as expected. Very little code has changed, and all tests are passing, so anything that isn't right is likely to be in the front end, as in previous CRUD PRs.

JoeCohen:
Clicking around:  In the GlossaryTerm index, thumbnails are too small, but that's not new in this PR.

JoeCohen:
On show NameDescription page, the Destroy link/button should be next to Edit.
Screenshot 2023-01-21 at 5 44 35 PM

JoeCohen:
PS: I wouldn't mind if the Other Available Locations column went away. 
It's approximately useless because it includes almost every defined Location.
And because it's not paged, it's awfully slow. (And if it were paged, it would be even hard to use.)
 `locations/merges/new` lists every location,

nimmolo:
@JoeCohen Thank you for that detailed walk-through. 

I fixed the Edit | Destroy layout, and also improved the irregular "colored_notes_box" panels to use bootstrap default classes. I removed the "even/odd" arg that isn't used in the layout anymore, in any case, and renamed the method "notes_panel", because it is only used for description notes.

Re: Merges
I agree, the Other Locations column is next to useless. It takes 10 days to load all the locations — and this list is not exactly useful when it does arrive, by covered wagon. 

However, I notice now that the action is available by other tab links, and it seems the "Locations::MergesController#new" action is the real problem here. 

I think it should not be returning all locations as a fallback — this renders the page unresponsive. If the location doesn't match any existing location string, I think the form should offer simply an autocomplete box. 

I'd also like to move the string matcher functions to a series of `Location` model scopes, since they are db queries.

What do you think?

nimmolo:
I've got a new possible solution for the location merges page. #1302

PR #1271, opened 2023-01-08
Make print_labels a POST rather than a GET

mo-nathan:
Seems to fix the issue reported on Facebook about print labels from a species list.


nimmolo:
This seems good, it has to be a POST because it's the result of a form commit. 

Thanks Nathan.

PR #1276, opened 2023-01-10
Fix edit location form commit, edit obs image upload 

nimmolo:
**In the edit observation form:**
The issue was the image voting interface, which is now made of `post_button`s that each create a mini `form`. (Rails does not want you to write to the db via GET, and voting originally did that. We do that still in a few other places.) Its presence induces a browser to automatically close out the enclosing form tag, and open a new one around the button.

So, any form embedding thumbnails, for now, will have to do without image votes in the form context.  I feel that is actually preferable, because it’s not a context where people should be trying to vote on images.  (Maybe someone is editing their own obs, and wants to "vote" on their own image, but those people will have to wait for another template, like the very next one, `show`.)

**In the edit location form:**
Despite locations not yet being CRUDified on master, Rails somehow figured out this was an edit form and added its automatic `method: :put` special sauce to the form. This explicitly sets `method: :post`.

### DIFF
--- a/app/views/location/edit_location.html.erb
+++ b/app/views/location/edit_location.html.erb
@@ -18,7 +18,7 @@
       action: :edit_location,
       id: @location.id,
       approved_where: @display_name
-    ), id: "location_form") do |form| %>
+    ), method: :post, id: "location_form") do |form| %>
   <%= render(partial: "form_location",
             locals: { form: form, button: :UPDATE }) %>
 <% end %>
--- a/app/views/observations/_form.html.erb
+++ b/app/views/observations/_form.html.erb
@@ -34,7 +34,10 @@
 
   <% if include_images %>
 
-    <%= render(partial: "observations/form/images", locals: { f: f }) %>
+    <% if @good_images.any? %>
+      <%= render(partial: "observations/form/images", locals: { f: f }) %>
+    <% end # if @good_images %>
+
     <%= render(partial: "observations/form/images_upload", locals: { f: f }) %>
 
   <% end # if include_images %>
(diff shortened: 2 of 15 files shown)


nimmolo:
The form refactor was a part of the 2-hour debugging process, wherein i forgot everything i wrote in the thumbnail helper. Keeping the refactor though! Trying to scan nested templates is awful.

pellaea:
I can't see the "real" changes amid all of the (very helpful!) closing if comments.  And the extra partials seem to be a good idea.  If it works, it works, it's fine with me!

nimmolo:
sorry for the wild goose chase... 😅 that's cause there are no real changes! 

After I eliminated every possible other option, turned out to just be the thumbnail votes.

JoeCohen:
I agree with @pellaea. And the changes work for me, so I'm going to deploy this.
Re erb files. Like Jason, I'm beginning to hate them in general. They are long, full of cryptic formatting stuff and complicated logic. There must be a better way. Does Trailblazer address this?

nimmolo:
Alan reported a new issue in create obs, with undefined location. It doesn't create the obs. I'm heading out for a bit and can debug later, but can either of you check the issue? Should write a test to cover this too. Re: TB, it does seem to have a separate layer to put view logic, yes. Believe all layers are separate gems, so it's opt-in. On Jan 10, 2023, at 4:26 PM, Joseph D. Cohen ***@***.***> wrote:﻿
I agree with @pellaea. And the changes work for me, so I'm going to deploy this.
Re erb files. Like Jason, I'm beginning to hate them in general. They are long, full of cryptic formatting stuff and complicated logic. There must be a better way. Does Trailblazer address this?

JoeCohen:
Nimmo:
Thanks for putting out the image and location "freez" fires.
I'll look into the issue that Alan reported

PR #1277, opened 2023-01-11
Set default val for `goodImageVals` in multi_image_upload.js

nimmolo:
Fixes a js error discovered by @pellaea today.

Unlikely to cause trouble. It just sets a backup default value for a var that might not get assigned.

```
Uncaught TypeError: goodImageVals is undefined
onreadystatechange https://mushroomobserver.org/assets/application-a3015424f037a08d1bce306ea77c078eef407c0180ed92424e216cc89981a2e3.js:1
```

**Background**: `goodImageVals` is jQuery's assessment of the value of a hidden form field in the Obs form, whose name and id are both `good_images`. It's printed in the form erb by Ruby mapping the ids of the `@good_images` ivar.  If there are no already uploaded images, this value will be nil.

### DIFF
--- a/app/assets/javascripts/multi_image_upload.js
+++ b/app/assets/javascripts/multi_image_upload.js
@@ -537,7 +537,7 @@ function MultiImageUploader(localized_text) {
       if (xhrReq.readyState == 4) {
         if (xhrReq.status == 200) {
           var image = JSON.parse(xhrReq.response);
-          goodImageVals = $goodImages.val();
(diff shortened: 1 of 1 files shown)


pellaea:
It sounds like, while this does avoid the js error, it's not fixing it
correctly.  I suspect it was supposed to be grabbing the image ids from the
AJAX return string and adding them to this array so that it could pass that
in with the form data.  But it's not, and that's why the images are not
being attached to the new observation.

PR #1280, opened 2023-01-12
Fix Remove Collection Number

JoeCohen:
Removing a Collection Number from an Observation currently throws an error. (And one user has complained.)
```ruby
Started GET "/collection_numbers/11072/remove_observation?observation_id=512701&q=1oAhh" for 127.0.0.1 at 2023-01-11 23:55:58 +0000
ActionController::RoutingError (No route matches [GET] "/collection_numbers/11072/remove_observation")
```
This PR:
- Adds a test that provokes the bug;
- Fixes the bug by replacing the Collection Numbers `Remove` link with a PATCH button.
- Delivers [Pivotal #184210741](https://www.pivotaltracker.com/story/show/184210741)

### Manual Test

- Create an Observation, checking the Specimen Available box, and adding a Collector's Number
- Remove the Collection Number

### DIFF
--- a/app/views/observations/show/_collection_numbers.html.erb
+++ b/app/views/observations/show/_collection_numbers.html.erb
@@ -27,12 +27,14 @@
                                  id: number.id,
                                  params: { back: observation.id }
                                )) %> |
-           <%= link_with_query(:REMOVE.t,
-                               collection_number_remove_observation_path(
-                                 collection_number_id: number.id,
-                                 params: { observation_id: observation.id }
-                               ),
-                               { data: { confirm: :are_you_sure.t } }) %>]
+           <%= patch_button(name: :REMOVE.t,
+                            path: collection_number_remove_observation_path(
+                              collection_number_id: number.id,
+                              observation_id: observation.id,
+                              q: get_query_param,
+                            ),
+                            data: { confirm: :are_you_sure.t }) %>]
+
         </li>
       <% end %>
     </ul>
--- a/test/integration/capybara/observations_controller_supplemental_test.rb
+++ b/test/integration/capybara/observations_controller_supplemental_test.rb
@@ -92,13 +92,15 @@ class ObservationsControllerSupplementalTest < CapybaraIntegrationTestCase
 
   def test_observation_remove_collection_number
     obs = observations(:minimal_unknown_obs)
+    assert_not_empty(obs.collection_numbers,
+                     "Test needs a fixture with a collection number(s)")
     user = obs.user
 
     login(user)
     visit(observation_path(obs.id))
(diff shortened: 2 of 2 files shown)


test/integration/capybara/observations_controller_supplemental_test.rb
```diff
@@ -90,6 +90,17 @@ def test_observation_edit_uncheck_species_list
     assert_not_includes(species_list.observations, observation)
   end
 
+  def test_observation_remove_collection_number
```
JoeCohen, line 93: In which file does this test belong?

app/views/observations/show/_collection_numbers.html.erb
```diff
@@ -27,12 +27,14 @@
...
-                                 params: { observation_id: observation.id }
-                               ),
-                               { data: { confirm: :are_you_sure.t } }) %>]
+           <%= patch_button(name: :REMOVE.t,
+                            path: collection_number_remove_observation_path(
+                              collection_number_id: number.id,
+                              observation_id: observation.id,
+                              q: get_query_param,
+                            ),
+                            data: { confirm: :are_you_sure.t }) %>]
```
JoeCohen, line 36: Is this the best way to solve the problem?
Should the `q` param be included?

nimmolo: Yes! To the first one. 
re: q, i confess i don't know when to keep it. But there doesn't seem to be any harm in passing it.

Thanks for catching this, it's a bewildering number of new CRUD routes I'm making.

test/integration/capybara/observations_controller_supplemental_test.rb
nimmolo: This one is fine. 
For MO's integration tests, they seem to have started out with one of two organizing principles: either by use case or by template. This fits both.

nimmolo (approved):

JoeCohen:
@nimmolo Thanks for the review. 
I am re-running the tests because the Check display is weird:
>codeclimate/diff-coverage Pending — Waiting for coverage results to be reported
codeclimate/total-coverage Pending — Waiting for coverage results to be reported
codeclimate — All good!
I will merge and deploy this as soon as that's fixed.

nimmolo:
@JoeCohen - Forgot to say yesterday, I'm out of town visiting my aunt this weekend... i'll have a phone but no computer until Monday.

The diff-coverage and total-coverage are new, right? I feel like they were still "pending" when i reviewed last night, but not sure.

JoeCohen:
Thanks for the response.
I think you're right that the pending lines were there when you reviewed it.
Don't worry about the issue; I'll take care of it. Worst case is that I
just ignore it.
Have a great visit with your aunt.

PR #1281, opened 2023-01-13
Redirect lookup_name

JoeCohen:
- Redirects the legacy lookup_name action. 
We ok'd external websites to use the old url, so we need to redirect it.
- Includes a test to prevent reversion.

#### Manual test

- Paste the old url (with a valid id) into the browser address bar.
Example: `http://localhost:3000/observer/lookup_name/5077`
Result: It should display the Name with that id. 
(For the example above, it should display the Name page for _Peltigera_.)


nimmolo (approved):

pellaea:
Yes, looks good, thanks, Joe!

JoeCohen:
Thanks. I'll get deployed.

PR #1282, opened 2023-01-15
Observation pattern search speedup

mo-nathan:
This should substantially speed up the search for observations by removing an unneeded join against the locations table as well as removing the search of the notes fields.


JoeCohen:
@mo-nathan Thanks for changing the tests.
I did some quick manual testing. It looks good. 
Although this changes the functionality, it's a win because:
- It's highly likely that users searching for Observations of _x_ do not want Observations where _x_ is merely in the Notes. Cf. [Pivotal #109575312](https://www.pivotaltracker.com/story/show/109575312)
-  Users who want to search Notes specifically can use the pattern search `notes` variable, e.g.
`notes:Piximètre user:"Joseph D. Cohen"`

pellaea:
Looks great!  So glad you did this, Nathan.

PR #1283, opened 2023-01-15
Fix destroy thumbnail

pellaea:
This was a subtle bug.  As it turns out some associations have already been updated before the before_destroy callback is called, therefore image.observations, for example, is already empty at that point.  This is why unit testing is so important!

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -139,8 +139,8 @@ class ImagesController < ApplicationController
     args = {
       action: "index",
       matrix: true,
-      include: [:user, { observations: :name }, :subjects, :best_glossary_terms,
-                :glossary_terms, :image_votes]
+      include: [:user, { observations: :name }, :subjects,
+                :glossary_term_thumbnails, :glossary_terms, :image_votes]
     }.merge(args)
 
     # Add some alternate sorting criteria.
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -140,7 +140,7 @@ class ImagesController < ApplicationController
       action: "index",
       matrix: true,
       include: [:user, { observations: :name }, :subjects,
-                :glossary_term_thumbnails, :glossary_terms, :image_votes]
+                :thumbnail_glossary_terms, :glossary_terms, :image_votes]
     }.merge(args)
 
     # Add some alternate sorting criteria.
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -139,8 +139,8 @@ class ImagesController < ApplicationController
     args = {
       action: "index",
       matrix: true,
-      include: [:user, { observations: :name }, :subjects,
-                :thumbnail_glossary_terms, :glossary_terms, :image_votes]
+      include: [:user, { observations: :name }, :profile_users,
+                :thumb_glossary_terms, :glossary_terms, :image_votes]
     }.merge(args)
 
     # Add some alternate sorting criteria.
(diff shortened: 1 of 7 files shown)


nimmolo:
Hi @pellaea  - sorry I'm getting to this late, I couldn't read the code on my phone. 

The attribute names seem a bit better, but still pretty confusing in my opinion. 

It makes sense to me for an image to be a "glossary_term_thumbnail", "observation_thumbnail" or a "user_thumbnail", but "thumbnail_glossary_term" reads like a type of glossary term, and "subject" is none too clear, especially when users may pick mushrooms as their thumbnail image. Making these consistently "x_thumbnail" would read better in both directions, IMO.

It also goes along with the table names, "observation_images", "project_images" etc.

pellaea:
Well, I made the names going from image to glossary term, so that's right!
But maybe we can reverse it the other direction?  Wait, it is reversed the
other way:

image.thumbnail_glossary_terms --> yields glossary terms
glossary_term.thumb_image --> yields an image

I didn't like the randomness, best_glossary_terms and thumb_clients.
Neither made sense to me!

I guess in long-hand it would be
"glossary_terms_whose_thumbnail_is_this_image".  So what's a better
shortened version of that?  I didn't like my choice at first either, but it
grew on me and I like it now.  But I could probably learn to like something
else, too.  Let's just make sure it's the same convention for both
glossary_terms and observations!!

-Jason

nimmolo:
Oh! If that's what it is, then i like it. Sorry for misreading it. 
Definitely agree with the impulse to make it not random. Let's make it similar for obs, glossary and user all three.

So `@image.thumbnail_glossary_term` is the glossary term that has this image as a thumbnail...

Instead of `subject` for the user whose thumbnail is this image, how about `thumbnail_user`.

pellaea:
Don't forget there could be multiple of each going from images to
users/terms/obs (not sure why, but it's allowed!)

observations.thumb_image <-> image.thumbnail_observations
glossary_term.thumb_image <-> image.thumbnail_glossary_terms
user.image <-> image.(what?)_users

Is it okay that it is "thumb" on one side, and "thumbnail" on the other?  I
mean, we could normalize that, too.  I would recommend against changing
"thumb_image" to "thumbnail_image" because I think that's all over the
place (and would require a change to the database!), but the reflected
association is barely used at all, so we could easily change it to
image.thumb_observations and image.thumb_glossary_terms.

I have a little resistance to image.thumbnail_users.  Ideally we would also
change user.image to something else.  Technically it's the "profile image",
so that would be my preference:

user.profile_image <-> image.profile_users

Ugh, no, that doesn't work... or does it?  I don't know, maybe thumb would
be better...

user.thumb_image <-> image.thumb_users

Hmmm.  There is such a thing as making it too symmetric.

And it is entirely possible we are obsessing over this too much. :)

nimmolo:
It's ok, discussion's good. Naming is half the battle in documentation. You don't want some future numbskull like Nimmo asking you what the heck a "subject" is for an image.

I vote for symmetrical and simple. Howsabout `thumb_profile_users`, `thumb_observations`, `thumb_glossary_terms`.

It's definitely preferable to the current situation with `best_glossary_terms` and `subjects`.

pellaea:
Okay, since I'm doing the actual PR, I broke the ties -- which is to say,
in effect I completely ignored your votes! :)

thumb_glossary_terms
thumb_observations
profile_users

But note that User#image remains because, alas, renaming it to
profile_image would require changing the name of the database column.  Not
worth it!

nimmolo:
I approve!

PR #1286, opened 2023-01-16
Catch `q` and `id` params for all refactored `index_{object}` methods

nimmolo:
In my controller refactors my pattern has been to route all `index_{object}` methods through `index` with params, and change all links (e.g. `{ controller: "/images", action: :index, by_user: user.id }` or `images_path(by_user: user.id)`. 

However, I did not always encounter conditionals to handle the query param, and i forgot to handle it in my `index` switch. The result of this  that a link sending a `q` param to certain controllers will not get an index of the relevant query.

No one has reported a problem so far. But it's not the right pattern: any request for an index with a param should hit one of the sub-methods built to handle params, rather than a plain `index`.

This adds a catch for the `q` and `id` params to the `index` switch in these controllers. I'm not sure the `id` is always necessary, but if someone's requesting an index with an `id` this is likely what they want.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -54,7 +54,7 @@ class CommentsController < ApplicationController
       show_comments_by_user
     elsif params[:for_user].present?
       show_comments_for_user
-    elsif params[:by].present?
+    elsif params[:by].present? || params[:q].present? || params[:id].present?
       index_comment
     else
       list_comments
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -47,7 +47,7 @@ class ImagesController < ApplicationController
       images_by_user
     elsif params[:for_project].present?
       images_for_project
-    elsif params[:by].present?
+    elsif params[:by].present? || params[:q].present? || params[:id].present?
       index_image
     else
       list_images
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -35,7 +35,7 @@ class ProjectsController < ApplicationController
   def index
     if params[:pattern].present?
       project_search
-    elsif params[:by].present?
+    elsif params[:by].present? || params[:q].present? || params[:id].present?
       index_project
     else
       list_projects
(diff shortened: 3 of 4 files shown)


nimmolo:
My question for reviewers, before I try to write tests, is: is handling the `q` param desirable for an index in these 4 controllers?
- Comments
- Images
- Projects
- SpeciesLists

Comments, I believe, are reachable by pattern search. For example if your query is `Candolleomyces candolleanus`, you can get an index of comments for `Candolleomyces candolleanus` as a pattern search. I don't know if you'd also want a link to give you comments for `Candolleomyces candolleanus in North Carolina after June 1, 2016`... or if there even is such a link, i haven't paid close enough attention to `link_with_query`, i've just been trying to strictly maintain it wherever encountered.

Ditto for the others. 

(The reason this came up is that it certainly is an issue for Locations, Names and their Descriptions!)

pellaea:
So, maybe it's time to have a refresher on what the purpose of the q
parameter is?

The idea is that a user does some search or index, and then starts going
through the results.  In effect, the q parameter remembers that outer
search.  So then a user could click on a result, do some work on it, the
return to the index -- always going forward in the browser, not using the
back button -- and it would resume right where they left off.

So what are the various things a user might do in the normal workflow of
examining the results of a search?  The obvious use case is the activity
log.  Go to the front page and start going through the latest observations,
proposing names, adding comments, voting, going to the next or previous
observation in the activity log, returning to the index, and so on.  Note
that when it returns to the index from an observation -- there are three of
these "sequence" links in the upper left: « Prev | Index | Next » -- if you
click on Index, it needs to take you not just to an index of the original
query, but ideally to the page that contains the observation that you were
just on.  (It would be annoying if it always returned you to the top and
make you page down to get back to where you left off.)

Therefore add_comment, cast_vote, propose_name all pass the q parameter
through, because these actions will all return you to the observation in
the end, from which the user may want to continue browsing results of the
original query.

But "how to use" or a random user's profile page or "about xxx..." will not
pass the q parameter through.  Because these are interpreted as starts of
new lines of inquiry, not normal processing in the workflow of the old line
of inquiry.

Does that help you get a better feel for why some actions pass the q param
and others don't?  I admit that I probably was a bit too generous in
preserving the q param, but I figured it was better to err on the side of
preserving it than on throwing it away.  And it was so easy to do with the
"pass_query_param" and "link_with_query" helpers.

As is frequently the case, I don't have a strong opinion on this feature.
I would not be heart-broken if we got rid of it altogether and removed the
prev/index/next buttons.  It might be worth soliciting a little feedback
before doing so, however.  And maybe looking at usage stats.

PR #1284, opened 2023-01-15
Automatically Populate New Visual Groups

mo-nathan:
Made it so creating a VisualGroup automatically includes a bunch of images (confidence > 1.5).  This works reasonably for adding single names.

I also added a way to add larger lists of names.  Ended up implementing it as a rake task.  It was able to create Alan Celestino's list of 1246 names in just under 12 minutes.

PR #1287, opened 2023-01-17
Fix Visuals `validates` messages

JoeCohen:
The old messages caused this problem:
` :cannot_include_tabs.t` was called when starting
 any test (even if unrelated to VisualGroup or VisualModel).
This, in turn, threw errors when testing certain changes in `Textile`.
 To verify the problem
 - In the master/main branch, put a breakpoint in `StringExtensions#t`
 - run `rails t test/models/donation_test.rb -n test_truth`
 - when it halts at the breakpoint, `display self`.

### Manual Test
There is none; I couldn't figure out how to insert a tab into the `name` field on the forms.

### DIFF
--- a/test/controllers/visual_groups_controller_test.rb
+++ b/test/controllers/visual_groups_controller_test.rb
@@ -109,7 +109,7 @@ class VisualGroupsControllerTest < FunctionalTestCase
     patch(:update, params: {
             id: @visual_group.id,
             visual_group:
-              { name: "Tab %{\t} here",
+              { name: "Tab\there",
                 approved: @visual_group.approved }
           })
     assert_flash_text(/#{:cannot_include_tabs.t}/)
--- a/test/controllers/visual_groups_controller_test.rb
+++ b/test/controllers/visual_groups_controller_test.rb
@@ -104,18 +104,6 @@ class VisualGroupsControllerTest < FunctionalTestCase
     assert_redirected_to edit_visual_group_url(@visual_group)
   end
 
-  test "should prevent names with tabs" do
-    login
-    patch(:update, params: {
-            id: @visual_group.id,
-            visual_group:
-              { name: "Tab\there",
-                approved: @visual_group.approved }
-          })
-    assert_flash_text(/#{:cannot_include_tabs.t}/)
-    assert_redirected_to edit_visual_group_url(@visual_group)
-  end
-
   test "should destroy visual_group" do
     login
     assert_difference("VisualGroup.count", -1) do
--- a/test/controllers/visual_groups_controller_test.rb
+++ b/test/controllers/visual_groups_controller_test.rb
@@ -98,9 +98,21 @@ class VisualGroupsControllerTest < FunctionalTestCase
     patch(:update, params: {
             id: @visual_group.id,
             visual_group:
-              { name: "",
+            { name: "",
+              approved: @visual_group.approved }
+          })
+    assert_redirected_to edit_visual_group_url(@visual_group)
+  end
+
+  test "should prevent names with tabs" do
+    login
+    patch(:update, params: {
+            id: @visual_group.id,
+            visual_group:
+              { name: "Tab %{\t} here",
                 approved: @visual_group.approved }
           })
+    assert_flash_text(/#{:cannot_include_tabs.t}/)
     assert_redirected_to edit_visual_group_url(@visual_group)
   end
 
--- a/app/models/visual_group.rb
+++ b/app/models/visual_group.rb
@@ -8,7 +8,9 @@ class VisualGroup < AbstractModel
   validates :name, presence: {
     message: proc { :cannot_be_blank.t }
   }
-  validates :name, format: { without: /\t/, message: :cannot_include_tabs.t }
+  validates :name, format: {
+    without: /\t/, message: proc { :cannot_include_tabs.t }
+  }
 
   def image_count(status = true)
     return visual_group_images.count if status.nil? || status == "needs_review"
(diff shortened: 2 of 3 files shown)


mo-nathan:
Interesting.  I also ran into this issue when I created the rake task in https://github.com/MushroomObserver/mushroom-observer/pull/1284 where I solved it by adding `require "extensions"`.  I'll merge your change with the branch from that PR and see which solution works better.

mo-nathan:
I can trigger the check by going to a text editor (like Apple Notes) and creating a line that has tab in it, then copying and pasting it into the form.

mo-nathan:
I'm not able to recreate the issue you're seeing on current master:
```
(base) nathan@Nathans-MBP mushroom-observer % rails t test/models/donation_test.rb -n test_truth
Started with run options -n test_truth --seed 8808

[375, 384] in ~/src/mushroom-observer/app/extensions/string_extensions.rb--=---] 0% Time: 00:00:00,  ETA: ??:??:??
   375|   HTML_TAG_PATTERN = %r{</*[A-Za-z][^>]*>}
   376| 
   377|   ### Textile-related methods ###
   378| 
   379|   def t(sanitize = true)
=> 380|     debugger
   381|     Textile.textilize_without_paragraph_safe(self, false, sanitize)
   382|   end
   383| 
   384|   def tl(sanitize = true)
=>#0	String#t(sanitize=false) at ~/src/mushroom-observer/app/extensions/string_extensions.rb:380
  #1	Symbol#t(args=[]) at ~/src/mushroom-observer/app/extensions/symbol_extensions.rb:289
  # and 44 frames (use `bt' command for all frames)
(rdbg) display self    # command
0: self = cannot include tabs
(rdbg) 
```

mo-nathan:
To reproduce the issue I saw, create the `lib/tasks/test.rake` containing
```
namespace :test do
  task(count_visual_groups: :environment) do
    puts VisualGroup.count
  end
end
```
Then run: `rake test:count_visual_groups`.  This should give you something like:
```
rake aborted!
NoMethodError: private method `t' called for :cannot_include_tabs:Symbol

  validates :name, format: { without: /\t/, message: :cannot_include_tabs.t }
                                                                         ^^
/Users/nathan/src/mushroom-observer/app/models/visual_group.rb:11:in `<class:VisualGroup>'
/Users/nathan/src/mushroom-observer/app/models/visual_group.rb:3:in `<top (required)>'
/Users/nathan/.gem/ruby/3.1.2/gems/zeitwerk-2.6.6/lib/zeitwerk/kernel.rb:30:in `require'
/Users/nathan/.gem/ruby/3.1.2/gems/zeitwerk-2.6.6/lib/zeitwerk/kernel.rb:30:in `require'
/Users/nathan/src/mushroom-observer/lib/tasks/test.rake:3:in `block (2 levels) in <top (required)>'
Tasks: TOP => test:count_visual_groups
(See full trace by running task with --trace)
(base) nathan@Nathans-MBP mushroom-observer % cat lib/tasks/test.rake
namespace :test do
  task(count_visual_groups: :environment) do
    puts VisualGroup.count
  end
end
```

mo-nathan:
It looks like your fix works for my case as well.  I'll update my branch to use that approach as well since I now see that it's being used earlier in the files.

JoeCohen:
Thanks so much!
The testing thing is strange. It must be an issue with whence I'm running
the tests (a remote terminal in VSCode). I'll try it directly in Terminal.
In the VSCode remote terminal I see:

> $ git status
> On branch master
> Your branch is up to date with 'origin/master'.
> nothing to commit, working tree clean
> ***@***.*** /vagrant/mushroom-observer  (master)
> $ rails t test/models/donation_test.rb -n test_truth
> Started with run options -n test_truth --seed 4893
> [375, 384] in
> /vagrant/mushroom-observer/app/extensions/string_extensions.rb---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=--]
> 0% Time: 00:00:00,  ETA: ??:??:??
> ....
> => 380|     debugger
>    381|     Textile.textilize_without_paragraph_safe(self, false, sanitize)
>    382|   end
>    383|
>    384|   def tl(sanitize = true)
> =>#0    String#t(sanitize=false) at
> /vagrant/mushroom-observer/app/extensions/string_extensions.rb:380
>   #1    Symbol#t(args=[]) at
> /vagrant/mushroom-observer/app/extensions/symbol_extensions.rb:289
>   # and 44 frames (use `bt' command for all frames)
> (rdbg) display self    # command
> 0: *self = cannot include tabs*

test/controllers/visual_groups_controller_test.rb
```diff
@@ -98,8 +98,8 @@ class VisualGroupsControllerTest < FunctionalTestCase
     patch(:update, params: {
             id: @visual_group.id,
             visual_group:
-              { name: "",
-                approved: @visual_group.approved }
+            { name: "",
+              approved: @visual_group.approved }
```
mo-nathan, line 102: This doesn't seem like the best way to do this and there is another case in the same file.  In my PR I just updated these cases to look like:
```
            visual_group: {
              name: @visual_group.name,
              approved: @visual_group.approved
            }
```
which Rubocop is also fine with and is more consistent with another spot in the file and seems cleaner to me.

mo-nathan:
I think if you're good with the changes I made in my PR, then we can just merge that and close this one.

JoeCohen:
They look good to me.

JoeCohen:
For future reference (I'm not going to try digging into it further)
I get the same result running directly in Terminal.
Could the difference (between what you and I are seeing) be that I'm
running on ***@***.*** and you're not.

PR #1288, opened 2023-01-18
Fix bracketed years

JoeCohen:
- Stops RedCloth from turning bracketed years (Ex: `[1878]`) into footnote references.
- Explanation: 
- RedCloth turns bracketed integers (ex: `[1]`) into footnote references.
But some Name Citations contain bracketed years. Ex:
`Hyménomycètes (Alençon): 103 (1874) [1878]`
We want them to render as such, not footnote references.
The PR therefore stops RedCloth from munging four-digit bracketed integers  
in **any** Textilized text, letting RedCloth munge other bracketed integers.
- Delivers [Pivotal #184232904](https://www.pivotaltracker.com/story/show/184232904)

### Manual Test
Compare the citations for the same Name 
- main/master https://mushroomobserver.org/name/show_name/36 (bracketed year munged) to
- this branch http://localhost:3000/name/show_name/36  (bracketed year not munged)


app/classes/textile.rb
```diff
@@ -114,6 +115,14 @@ def textilize(do_object_links = false, sanitize = true)
...
     # Let Textile munge the thing up now.
     red = RedCloth.new(self)
+
+    # Quote bracketed years
+    # in order to stop RedCloth from turning them into footnote calls
+    # Some Name Citations contain bracketed years. Ex:
+    #   Hyménomycètes (Alençon): 103 (1874) [1878]
+    # We want them to render as such, not footnote references.
+    red.gsub!(BRACKETED_YEAR, '==[==\1]') if red.match?(BRACKETED_YEAR)
+
```
pellaea, line 125: I could have sworn this had to come before the RedCloth.new(self) line...  Yet, you have tested it and it apparently works.  I am so confused!

JoeCohen: Why didn't I think of that? It will be a little simpler that way.
I'll make the change.

PR #1290, opened 2023-01-18
Added script to automatically backup the database.

pellaea:
I changed it to store the snapshots on the image server, so there is no manual test now, sorry.

It keeps 7 daily backups, 8 weekly backups, 12 monthly backups, all yearly backups, and it removes anything else from the backup directory.  This will soon take up ca. 10 GB of disk space, but that's a drop in the bucket on the image server, so I don't anticipate any problems.

### DIFF
--- a/script/backup_database.sh
+++ b/script/backup_database.sh
@@ -6,9 +6,9 @@
 #
 #  DESCRIPTION::
 #
-#  Makes a new snapshot of the database in:
+#  Makes a new snapshot of the database on the image server:
 #
-#    db/backups/
+#    /data/images/mo/backup/database-YYYYMMDD.gz
 #
 ################################################################################
 set -e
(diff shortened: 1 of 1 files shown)


JoeCohen:
My instinct is to store them on the image server.

pellaea:
Changed script to store the snapshots on the image server.  It's a lot harder to test and potentially more dangerous.  The delete command at the end really wigs me out.  For example, if something goes wrong and it's not able to create the backup directory remotely, it can return a listing of everything in mo's home directory on the image server, and then attempt to delete it all.  I'll add sanity checks in a few places to prevent that from happening.

JoeCohen:
Would the GoogleCloud be any better?

pellaea:
Google Cloud is a good idea.  However it's harder ... if we want to keep the snapshots down to a reasonable number.  If we just stuff snapshots into the cloud every day... oh wait, we would need to encrypt them, then, too.  There is sensitive information in these snapshots.

I'm pretty happy with this script now.  I think the new assertions prevent anything crazy or unexpected from happening.

JoeCohen:
Oops. I forgot about encryption.

pellaea:
Yeah, I think these will be more convenient to use when and if we need them
later.

mo-nathan:
Look great to me!

pellaea:
Cool.  I'll let 'er loose tonight.

PR #1292, opened 2023-01-19
Bundle update

JoeCohen:
This PR 
- Solves the security issues addressed by failing #1291.
(That's a Dependabot PR which fails because of credentials issues.)
- Updates all gems with a `bundle update`.

I updated all gems because an attempt to update only `activerecord` failed.
```sh
$ bundle update activerecord
...
Bundler attempted to update activerecord but its version stayed the same
Bundle updated!
```
Furthermore, we cannot `bundle update rails` because only parts of rails are installed
```sh
bundle update rails
Could not find gem 'rails'.
Did you mean mail?
```
So `bundle update` was the path of least resistance.


mo-nathan (approved):
Looks good!

nimmolo (approved):

PR #1296, opened 2023-01-21
Use new page_dispatch for indexes

nimmolo:


JoeCohen (commented):
I like 
- Getting rid of the `case` chains in the controllers
- Keeping the controllers DRY by having `page_dispatch` in `ApplicationController`.

But for me, this implementation is too meta/abstract (due to both structure, and names). 
If I didn't know the intent in advance and/or have the comments, I would have puzzled over this for a long time. Maybe that's because I'm an amateur and the rest of you are pros.
So I like my solution, https://groups.google.com/g/mo-developers/c/asJ1AbwHQXQ/m/FkC8UFCcAAAJ
or something closer to it.  
(If you and others would prefer my idea, I'd be glad to code and push it.)

Otherwise:
- Do you need the `params` parameter in `page_dispatch` and `key_match`?
- Could you simplify `page_dispatch` (as follows) by always making the first parameter of the dispatch table an array?  e.g. `[ [:pattern], :name_search]`
```ruby
  def key_match(key)
    key.each do |subkey|
      return true if params[subkey].present?
    end
    false
  end
```
- Indeed, could you then get rid of `key_match` entirely by inlining it in `page_dispatch`:
```ruby
    dispatch.each do |method|
      params.each do |param|
        return method.call if params[param].present?
      end
    end
```

pellaea:
I'm trying to stay silent in this, as I think the fact that I don't really
code anymore makes my opinion less important for making general policy
decisions like this going forward, but...

I don't understand why the original method wasn't perfectly readable,
understandable and maintainable.  We seem to be tying ourselves in knots to
fix something that isn't broken just because rubocop says we should.

I think Nathan raises an interesting point that this is a symptom of a
deeper design decision.  And maybe making a radical restructure of how
indices and such are done we could avoid the problem.  Fair enough, but
that seems way beyond the scope of the original problem.

Also, I'm increasingly doubting that strict CRUD really is the ideal.

I feel like all three of these points -- strict adherence to ABC < 15 or
20, strict CRUD, and trying to ensure really deep design decisions are
correct in order to avoid something superficial much later on that is
really hard to foresee -- are examples of taking good ideas too far.

I think these sorts of discussions are really important.  I, for one, have
learned a number of things.  (For example, I really appreciate the way
Nathan framed keeping everything at the same level of abstraction in a
given method, something I've instinctively tried to do, but it's good to
clarify exactly what that instinct is trying to do!)  And I'm very
interested to see where it all goes in the meeting -- that is today...
right now(!), right?  Oh... ;)

-Jason

JoeCohen:
All good things to discuss in next week's meeting.

pellaea:
Oops!  Thank you, Joe!  It's been a long week... :)

JoeCohen:
Another improvement? on the proposed inlining
```ruby
    dispatch.each do |method|
      return method.call if params.any? { |param| param.present? }
    end
```

JoeCohen:
For your consideration, I modified my solution to use a hash (thanks to @mo-nathan) looks like the following.
The extra one-line methods are gone. But there's a lot of ink on the page, and it's hard on the eye.
```ruby
  LOCATIONS_INDICES = [
    { condition: ->(params) { params[:advanced_search].present? },
      method: :advanced_search },
    { condition: ->(params) { params[:pattern].present? },
      method: :location_search },
    { condition: ->(params) { params[:country].present? },
      method: :list_by_country },
    { condition: ->(params) { params[:by_user].present? },
      method: :locations_by_user },
    { condition: ->(params) { params[:by_editor].present? },
      method: :locations_by_editor },
    { condition: lambda { |params|
                   [:by, :q, :id].any? { |param| params[param].present? }
                 },
      method: :index_location },
    { condition: ->(_params) { true },
      method: :list_locations }
  ].freeze

  def index
    LOCATIONS_INDICES.each do |type|
      return send(type[:method]) if type[:condition].call(params)
    end
  end
```

test/controllers/locations_controller_test.rb
```diff
@@ -179,6 +179,12 @@ def test_list_locations
     assert_template("index")
   end
 
+  def test_bogus_param
+    login
+    get(:index, params: { bogus: "234" })
+    assert_template("index")
+  end
+
```
JoeCohen, line 187: Would it make sense to move this test to #1269?

JoeCohen, line 187: Would it make sense to move this test to #1269?

JoeCohen:
I'm going to refactor everything based on @mo-nathan 's suggestion. 
(May a branch off this one, refactor, then merge back into this one.)

nimmolo:
@JoeCohen - That sounds good!

JoeCohen:
On 2nd thought, I'll do this off master so that it's independent of #1269
and can be applied to other controllers immediately.

JoeCohen:
Replaced by #1307

PR #1297, opened 2023-01-22
Make a bug report button in the main nav

nimmolo:
Let me know what you think.


pellaea:
Why not!

JoeCohen (changes requested):
Thanks @nimmolo for doing this.. A great idea to get us earlier bug reports.
Suggestion:
Rather than adding a category to the left nav-bar, put the button under More and style it differently. (E.g. in a the black-and-white them, a red or yellow background.)

nimmolo:
@JoeCohen - Took your suggestion and moved it under More. 

In most themes, it's a contrasting color, the menu subhead color. But not for B&W, and it's too hard to find a color that works with all the themes and isn't a hardcoded color. I don't want to write in a non-theme exception color when i'm about to redo all the theme CSS. 

This is less prominent but i'm ok with it.

pellaea:
@nimmolo -- B&W needs to have a contrasting color, too. I thought we
already had one for "Delete" buttons and such, no?  Or does that only work
for normal text, not the sidebar?

pellaea:
You've convinced me.  Sorry!  Yes, definitely not worth the effort.  We
already have a big-report button, after all -- it's called "send the
webmaster an email"!  And empirically, people manage to find it all too
often. :)  So carry on and ignore me!

nimmolo:
I'm carrying on... and ignoring your request to ignore you ;)

`theme_class = @user&.theme ? " theme-#{@user&.theme.underscore.dasherize}" : ""`

I feel like this looks up @user twice. Isn't there a ternary assignment in Ruby, like `foo = bar ?: bip` ?
Or is it okay?

pellaea:
@user should already be looked up when it is assigned.  ...  Yikes, it's
about 50 levels down from ApplicationController#autologin, but here it is:
ApplicationController#session_user: User.safe_find(session[:user_id])...
and that is *eventually* assigned to @user and User.current.  So, accessing
@user causes no additional lookups.

... That is what you were concerned about, right?

nimmolo:
Yes. Thanks!.... But also the phrasing of ternary assignments!

pellaea:
Mmm, yeah, I don't know how to solve the ternary problem here without a
temp variable: temp = @user&.theme; theme_class = temp ? "
theme-#{temp.underscore.dasherize}" : ""

Or maybe theme_class = @user&.theme&.underscore&.dasherize; theme_class &&=
" theme-#{theme_class}"...  but I think that long chain of &. will tick off
rubocop.

nimmolo:
Hm. I do want to solve this sooner or later, so: 
```ruby
theme_name = @user&.theme.present? ? @user&.theme.underscore.dasherize : default_theme
```
If i do this for darv, I get `theme-null`. His theme is saved as "NULL" in the db.

The deeper issue is, `NULL` results in the random (sample) theme function.
So it's going to be trickier to be sure what theme is displayed, at least calling from the application layout.

Further up, the `css` var is set in the `<head>`, with a default value of `MO.default_theme`.
If i make `css` an ivar, like `@user_theme`, i can access later in the partial that creates this body class. 

What do you reckon?

nimmolo:
Ok, i've got the new class in `<body>` now, plus an exception for `theme-black-on-white`. I believe this helper function may be useful in the near future for bootstrap-4.

@JoeCohen - The `css_theme` helper function is new here, logic was moved from the `layouts/app/_head.html.erb` template to the `application_helper.rb`. 

Moving it to the helper has exposed the existing ABC violation (since templates don't get scanned), and in keeping with the new M.O. i'm not disabling the cop, but keeping the existing violation (20.71) (The language for this is so "violent"!)

If you would like to try to think of a way to reduce ABC on this, please do.

pellaea:
> Hm. I do want to solve this sooner or later, so:
> 
> ```ruby
> theme_name = @user&.theme.present? ? @user&.theme.underscore.dasherize : default_theme
> ```
> 
> If i do this for darv, I get `theme-null`. His theme is saved as "NULL" in the db.
> 
> The deeper issue is, `NULL` results in the random (sample) theme function. So it's going to be trickier to be sure what theme is displayed, at least calling from the application layout.
> 
> Further up, the `css` var is set in the `<head>`, with a default value of `MO.default_theme`. If i make `css` an ivar, like `@css_theme`, i can access later in the partial that creates this body class.
> 
> What do you reckon?

Why in the world do we have users.theme == "NULL" instead of nil??  Good grief!  Sounds like a bug got introduced during some migration???

If this is actually intended, then we should change "NULL" to "random" or something actually clear, right??

nimmolo:
I'd also be in favor of changing `"NULL"` to `"BlackOnWhite"`. 
But maybe Darvin actually prefers the kaleidoscope effect... It is kinda cool, like fungal variety itself.

nimmolo:
In any case, the helper method works. This PR is good to go!

pellaea:
> I'd also be in favor of changing `"NULL"` to `"BlackOnWhite"`. But maybe Darvin actually prefers the kaleidoscope effect... It is kinda cool, like fungal variety itself.

Originally theme was optional, and nil was the default state which was the random "kaleidoscope" setting, and you could set it to a definite value if you had a preference.  (This was actually the very first thing I did for MO!)  Given that there are almost 6000 users with this setting (possibly all inactive, admittedly, and I'm too lazy to check), I think it is worth keeping.  It should be really easy to implement, after all.

If the PR works as it is, then I say let 'er loose.

If the random mode is gone, then I'll make a quick PR to fix it (including a migration to fix the theme = "NULL" bug).

nimmolo:
Random mode works here too, darv's the one i've been testing.  A value like "NULL" still works because the code just tests if the saved theme is one of "MO.themes". Anything else reads as "random". 

I agree, it would be preferable to have the saved theme be something definite, like "Random".

pellaea:
Okay, cool, low priority, then! :)

PR #1298, opened 2023-01-22
Refactor checklist HTML

nimmolo:
This is a multi-column list, not a table, so this uses CSS columns and a `<ul>`.

Note: This controller could use a little attention: the site list is absurdly expensive to calculate.

Not only that, but `@data` seems to return a list of name strings only, and then to form the name links, the helper must look up the names again one by one. It's an n+12,800+1 problem currently.

But that's outside the scope of this tiny PR.

### DIFF
--- a/app/helpers/observations_helper.rb
+++ b/app/helpers/observations_helper.rb
@@ -140,10 +140,10 @@ module ObservationsHelper
       proposer_links = ""
     end
 
-    [name_link(naming), proposer_links].safe_join
+    [naming_name_link(naming), proposer_links].safe_join
   end
 
-  def name_link(naming)
+  def naming_name_link(naming)
     link_with_query(
       naming.display_name_brief_authors.t.break_name.small_author,
       show_name_path(id: naming.name)
--- a/app/views/checklists/show.html.erb
+++ b/app/views/checklists/show.html.erb
@@ -1,6 +1,8 @@
 <%
   # NOTE: action was `observations_of_name`. Whether before or after refactor,
   # this action needed a `name` param but there's none here. Reverting to index
+  # LAYOUT NOTE: This is not a table, it's a multi-column list.
+  # Convert to CSS columns
   name_link = { controller: :observations, action: :index }
 
   if @show_user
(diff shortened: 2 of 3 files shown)


app/views/checklists/show.html.erb
```diff
@@ -49,35 +51,22 @@
...
-        <% end %>
-      </tr>
-    <% end %>
-  </table>
-</div><!--.text-center-->
+<div class="panel panel-default">
+  <div class="panel-body checklist">
+    <ul class="list-unstyled">
+      <% names.each do |name| %>
+        <li><i><%= link_to(name, name_link.merge(name: name)) %></i></li>
```
pellaea, line 68: So, so throwing this out there, but if this is the only reason it has to look up every single name in the list... why not use the lookup_name link instead?  That defers the lookup until after you click on the name, right?

pellaea: (I mean, obviously, a deeper refactor could potentially solve the problem much better, and all!  But this could be a super-simple way of solving the huge N+1 problem in the short term!)

mo-nathan:
I would like to understand how we could have avoided having two fixes for this issue.  See https://github.com/MushroomObserver/mushroom-observer/pull/1299.  I attempted to claim the issue in Slack, but that was apparently missed.  I think we need a better way to communicate about this sort of thing.  I will work to merged the two PRs into a single solution.  I've also gotten more actual end user feedback from Darvin via Facebook, so I'm going to make a few more functional changes.  I'll also review the queries that are getting generated.  When I looked at it, it didn't seem quite N+1 (I thought it only did additional queries in some obscure cases involving synonyms), but I'm sure it could be better.

app/helpers/observations_helper.rb
```diff
@@ -140,10 +140,10 @@ def name_html(naming)
       proposer_links = ""
     end
 
-    [name_link(naming), proposer_links].safe_join
+    [naming_name_link(naming), proposer_links].safe_join
```
mo-nathan, line 143: While I don't disagree with this renaming, what is it's relevance to this PR?

mo-nathan:
Ok, updated my PR to include your much better way of doing the ERB stuff.

Regarding the "N+1" thing, it is as I suspected.  For my own page 568 species and the extra query only gets called 57 times.  So it seems to be something like N/10+1 which still isn't great, but it's as bad as it seemed at first.  It is adding some additional data, but I think it can be done in a single query that might grab more data, but would be more efficient.  I'll continue this on the other PR.

nimmolo:
@mo-nathan thanks for taking this on. Sorry for the duplicate work... i also made myself late to something i wanted to do, fussing with it. 😅

I'll make a point to check Slack before I start on unclaimed issues. I don't like having Slack or anything else open when i'm working, for me it's a distraction engine.

I believe it would be good if someone with a Facebook account posts something there to try to steer people to the webmaster email, so we have one focused bug report venue.

app/helpers/observations_helper.rb
nimmolo: My checklist erb was getting this `name_link` method when it was trying to call `name_link` from `object_link_helpers`. 

I suppose that helpers maybe can't have private methods?! 

So this was effectively a name collision, although in a different class. I'll do it in a separate PR

PR #1300, opened 2023-01-22
Rename `naming_name_link` in `observations_helper.rb`

nimmolo:
Helpers [cannot have truly private methods](https://stackoverflow.com/questions/12835367/why-can-private-helper-methods-still-be-accessed-in-views), I just found out!

So `ObservationsHelper#name_link` was colliding with the method of the same name in `object_link_helper.rb`, and in a recent PR was getting called first.

This PR also renames a couple other methods in `ObservationsHelper` to be more specific.


app/helpers/observations_helper.rb
```diff
@@ -125,9 +121,7 @@ def observation_naming_buttons(observation, do_suggestions)
     buttons.safe_join(tag.br)
   end
 
-  private
```
mo-nathan, line 128: Why did you entirely get rid of private from this?  I'm not seeing any external references.

app/helpers/observations_helper.rb
```diff
@@ -33,8 +33,6 @@ def owner_naming_line(obs)
     end
   end
 
-  private
-
   # name portion of Observation title
   def obs_title_consensus_id(name)
```
mo-nathan, line 37: Seems like this should stay private.  I noticed that there are some direct calls from tests, but it's not clear to me why.  It seems like the tests should know exactly what they are expecting.

app/helpers/observations_helper.rb
nimmolo: @mo-nathan - See the commit comment. Helpers cannot have private methods.

mo-nathan: Oh, I see the link now.  Reading it over, it doesn't seem to say that they can't be private, just that they get mixed in so it doesn't have as big an effect as you might expect.  However, it would still be private for things that try to access whatever is loading the helper.  In my view the biggest reason to make things private is as a declaration of what the author of the code expects other folks to use and what what they shouldn't.  There are ways of calling private methods if someone is really determined to do.  The key difference to me is if they do, then if you remove or change the interface it's on them.

nimmolo: OK, thank you for reading through that discussion. I didn't get how any other context could expose a private method, but i guess if it's mixed into the calling class, then... that sort of makes sense. Still — I had no idea I needed to be aware of namespace collisions in separate helpers!

I have been declaring tons of methods private throughout these CRUD refactors, I definitely prefer the convention. Will restore the `private` here.

nimmolo:
I had disabled the "no ivars in helpers" cop here in the helper awhile back, and am now refactoring the method to take kwargs and re-enabling the cop on the method. (the ivar is already defined by the caller when necessary, so can be easily passed as an arg)

app/helpers/observations_helper.rb
nimmolo: Done. I had disabled the "no ivars in helpers" cop here in `obs_title_consensus_id` awhile back, and am now refactoring the method to take kwargs and re-enabling the cop on the method. (the ivar is already defined by the caller when necessary, so can be easily passed as an arg)

app/helpers/observations_helper.rb
nimmolo: Restored private methods

PR #1302, opened 2023-01-23
Change the offered `@matches` for Location merges. 

nimmolo:
Removes "all locations" as an option.

New way of calculating possible matches - it takes the `where` string in segments separated by commas, and tries the first one. If there are different words, it tries to match the first word, too. 

Example in the current stripped db:
"Carara National Park, Costa Rica"
tries
"Carara National Park" then "Costa Rica", then finally "Carara" (bingo)

Hypothetical example: 
"Van Cortland Park, Bronx, New York, USA"
tries 
"Van Cortland Park", then "Bronx", then "New York", then "USA". 
It lists all of them in that order. Eventually in that list will be:
"Van Cortlandt Park, Bronx, New York, USA"

Potential issue: "USA" is a helluva location. But it's still less than "all locations".


JoeCohen (dismissed):
Thanks for doing this. 
The code looks good to me. 
I have not tested it manually because I forget (if I ever knew) how a user would create an "undefined" Location via the UI.
Do you have any other ideas for manually testing.

Oh wait! I just noticed that this is failing. 
Maybe dose this PR and deal with this later the issues after #1269 is merged into master/main.

nimmolo:
Just noticed that too!
That sounds like a plan. #1269 i believe is ready, all the other issues raised by click-through are cleared up.

nimmolo:
@pellaea 
Joe and I are wondering — What are the ways a location can end up in the db as "undefined"?

Do we still allow observations to be created without a defined location? I guess we send people to the "Create Location" form but they don't have to follow through...

pellaea:
Yes, exactly.  That's how.  Possibly also via species lists in a similar
manner(?)

JoeCohen:
It's great that the Other Available Locations column has gone away.
But the Near Misses can be anything but, and waaay too long. 
Example "Austin, Texas, USA" has 16,988 "Near Misses" with the best one (Austin, Travis Co., Texas, USA) buried somewhere in the core of the earth.
IMO, unless you're in love with this problem, you should (a) continue to ditch Other Available Locations, and (b) go back to the old algorithm for Near Misses. You have so many better things to do and the Locations#merge UI is rarely used.

nimmolo:
LOL. Not really in love with the problem... will go back to old logic. But, maybe we can keep something like my new phrasing — i find the current kind of multiple assignment syntax pretty obscure:

```
variable1, variable2 = method_that_returns_an_array(substrate, thing(arg))
```

nimmolo:
@JoeCohen - I believe this is resolved and back to the original `@matches + @others`. 

It's passing, and "Carara National Park, Costa Rica" on local now shows "Parque Nacional Carara, Puntarenas, Costa Rica" as a merge option, as it should. I didn't even find this location scanning the list myself, before!

nimmolo:
Merging because it passes the "Austin, Texas, USA" test, and it's so much better than the current list of "All Locations".  Still open to improvements... but i believe it's Good Enuff.

PR #1304, opened 2023-01-24
Dispatch index action to subactions

JoeCohen:
- Dispatches certain `index` actions to subactions through a dispatch table instead of switches
(`case` statements of `if elsif else`).
- Improves ABC metric of those actions, fixing some Rubocop ABC offenses.
- See [@mo-nathan's email](https://groups.google.com/g/mo-developers/c/asJ1AbwHQXQ/m/wdN3tL3eAAAJ) and related discussion.

There is no manual test.

### DIFF
--- a/app/controllers/collection_numbers_controller.rb
+++ b/app/controllers/collection_numbers_controller.rb
@@ -23,7 +23,7 @@ class CollectionNumbersController < ApplicationController
   }.freeze
 
   def index
-    dispatch_index_to_subaction
+    dispatch_to_index_subaction
   end
 
   def show
--- a/app/controllers/collection_numbers_controller.rb
+++ b/app/controllers/collection_numbers_controller.rb
@@ -14,16 +14,16 @@ class CollectionNumbersController < ApplicationController
   before_action :pass_query_params, except: :index
   before_action :store_location, except: [:index, :destroy]
 
-  def index # rubocop:disable Metrics/AbcSize
-    if params[:pattern].present? # rubocop:disable Style/GuardClause
-      collection_number_search and return
-    elsif params[:observation_id].present?
-      observation_index and return
-    elsif params[:by].present? || params[:q].present? || params[:id].present?
-      index_collection_number and return
-    else
-      list_collection_numbers and return
-    end
+  @dispatch_table_for_index_subactions = {
+    pattern: :collection_number_search,
+    observation_id: :observation_index,
+    by: :index_collection_number,
+    q: :index_collection_number,
+    id: :index_collection_number
+  }.freeze
+
+  def index
+    dispatch_index_to_subaction
   end
 
   def show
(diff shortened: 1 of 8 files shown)


app/controllers/collection_numbers_controller.rb
```diff
@@ -14,16 +14,16 @@ class CollectionNumbersController < ApplicationController
...
-    if params[:pattern].present? # rubocop:disable Style/GuardClause
-      collection_number_search and return
-    elsif params[:observation_id].present?
-      observation_index and return
-    elsif params[:by].present? || params[:q].present? || params[:id].present?
-      index_collection_number and return
-    else
-      list_collection_numbers and return
-    end
+  @dispatch_table_for_index_subactions = {
```
JoeCohen, line 17: Is there a better way to make this table available to the superclass?

app/controllers/collection_numbers_controller.rb
```diff
@@ -14,16 +14,16 @@ class CollectionNumbersController < ApplicationController
...
+    pattern: :collection_number_search,
+    observation_id: :observation_index,
+    by: :index_collection_number,
+    q: :index_collection_number,
+    id: :index_collection_number
+  }.freeze
+
+  def index
```
JoeCohen, line 25: Question: Should I try to completely push up `index` to the superclass (leaving this file without an explicit `index` method). 
Advantage: It would be DRYer.
I left it this way because you don't have to look at another file to understand how `index` works here.

app/controllers/observations_controller.rb
```diff
@@ -29,4 +29,21 @@ class ObservationsController < ApplicationController
...
+    related_taxa: :observations_of_related_taxa,
+    name: :observations_of_name,
+    user: :observations_by_user,
+    location: :observations_at_location,
+    where: :observations_at_where,
+    project: :observations_for_project,
+    by: :index_observation,
+    q: :index_observation,
+    id: :index_observation
+  }.freeze
```
JoeCohen, line 48: Question: How best to make this variable available to the superclass?
I'd like to put it in `app/controllers/observations_controller/index.rb`, but it blows up there.

JoeCohen:
Closed for the reasons given by @mo-nathan:
>Keys in a dictionary are not guaranteed to retain their order so if you have a case where more than one of the subaction params could be there, then you wouldn't be guaranteed to get the right subaction.  For another, the goal for new controllers is for KEY_TO_SUBACTION to be empty.  The idea is that it only captures exceptions to the direct mapping between key and value.  And finally, if you did just use a dictionary, then you'd have to [have] keys that are the same as their subaction.  If you address all of those concerns, you'd essentially end up with the same implementation as I first proposed where you have an ordered list of tuples (or maybe dictionaries or whatever).  You could simplify that solution a tiny bit by getting rid of the default case as we've discussed and you could deal with the "or" case just by giving each key its own row, but it's not very different.

PR #1305, opened 2023-01-24
Fix problem with deprecate name.

pellaea:


pellaea:
If you click "Deprecate" from a name page it would complain "The name '' is not recognized." before you even have a chance to type a name in.  This PR disabled that feedback if the "what" field is empty.

I am too lazy to revert the code to find out how it worked originally!  I think this will have the desired effect.

PR #1306, opened 2023-01-24
Tweaked Name#brief_author to leave qualifiers at the end intact

pellaea:
I think at least some qualifiers are critically important, such as "nom. prov." and "ined."  This PR currently tells Name#brief_author to preserve all such qualifiers -- defined as any number of lowercase words and dots following the final comma.

I'm working on getting a list of all currently used qualifiers so we can have an informed discussion about whether this is the right thing to do...

### DIFF
--- a/app/models/name/format.rb
+++ b/app/models/name/format.rb
@@ -108,13 +108,21 @@ module Name::Format
 
   private
 
+  PROV                       = /[a-z]+\.? prov\.?|ined\.?|ad ?int\.?/
+  INVAL                      = /[a-z]+\.? (inval|illeg(it)?)\.?/
+  ANY_ENDING_AFTER_COMMA     = /^(.*)(, [a-z. ]+)$/
+  SOME_ENDINGS_WITHOUT_COMMA = /^(.*)( (#{PROV}|#{INVAL}))$/
+  ENDINGS_WORTH_KEEPING      = / (#{PROV}|#{INVAL})$/
+
   # author(s) string shortened per ICN Recommendation 46C.2
   # Relies on name.author having a comma only if there are > 2 authors
   def brief_author
     str = author
     # pull of any qualifiers at the end, like "ined.", "nom. prov.", etc.
-    if (match = author.match(/^(.*)(, [a-z. ]+)$/))
+    if (match = author.match(ANY_ENDING_AFTER_COMMA) ||
+                author.match(SOME_ENDINGS_WITHOUT_COMMA))
       str, ending = match[1, 2]
+      ending = "" unless ending.match(ENDINGS_WORTH_KEEPING)
     end
     str.sub(/,.*\)/, " et al.)"). # shorten > 2 authors in parens
       sub(/,.*/, " et al.") +     # then shorten any remaining > 2 authors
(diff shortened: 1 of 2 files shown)


pellaea:
There are only 21 names and 14 cases in my snapshot:

ad int.
comb. nov.
comb. prov.
et al.
gen. nov.
ined.
nom. cons.
nom. illeg.
nom. inval.
nom. prov.
sp. nov.

1. The "xxx. nov." should all be fixed.  There is no reason to include that in the author except in the protologue itself (where it is required!)
2. "et al." is clearly not what I intended.  It should probably be fixed.  I think our system should include the full list of authors, and let Name#brief_name take care of eliding long lists. 
3. Clearly "nom. prov.", "comb. prov." and "ined." are the cases I had in mind for this PR.
4.  I don't recognize "ad int." ... is that another form of "ined."?
5. "nom. cons". seems extraneous and should probably be removed, agreed?
6. What about "nom. illeg.", "nom. illegit.", "nom. inval."?  Maybe it's fine to omit these?

Regardless of whether we keep or omit these qualifiers, they need to be dealt with, because the way it used to work, it does not handle these right, as "Hollinger & Lendemer et al." proved!

I don't have the time to handle this "correctly" -- which would be to change the parser itself and ensure that these qualifiers are dealt with correctly from the start: Auto-remove the superfluous ones, auto-correct the spelling of nom. illeg., remove the comma from ", et al.", and so on.

I propose to simply omit them all but "ad int."(?), "ined." and "xxx. prov.".  Does that work for everyone?  Other proposals?

pellaea:
Oh my stars.  There are a *lot* of cases where the comma is omitted.  There are 69 "et al.", 267 more "xxx. prov.", several more creative "ined", and so on.  What a mess!

JoeCohen:
> 1. The "xxx. nov." should all be fixed.  There is no reason to include that in the author except in the protologue itself (where it is required!)

Agree. Users (including those who should know better) do it anyway.

> 2. "et al." is clearly not what I intended.  It should probably be fixed.  I think our system should include the full list of authors, and let Name#brief_name take care of eliding long lists.

Agree in principle. But what if the authors' names are too long for the current field limit (or would make search_name, display_name, or sort_name too long)? We had this problem a while back, and bumped the size of those fields.

> 3. Clearly "nom. prov.", "comb. prov." and "ined." are the cases I had in mind for this PR.

Agree, but see 6 below. 

> 4. I don't recognize "ad int." ... is that another form of "ined."?

Must be. "ad int.—ad interim, in the meantime". https://www.govinfo.gov/content/pkg/GPO-STYLEMANUAL-2008/pdf/GPO-STYLEMANUAL-2008-11.pdf#page=31

> 5. "nom. cons". seems extraneous and should probably be removed, agreed?

Agree. 

> 6. What about "nom. illeg.", "nom. illegit.", "nom. inval."?  Maybe it's fine to omit these?

I think those are useful. 

> 
> Regardless of whether we keep or omit these qualifiers, they need to be dealt with, because the way it used to work, it does not handle these right, as "Hollinger & Lendemer et al." proved!
> 
> I don't have the time to handle this "correctly" -- which would be to change the parser itself and ensure that these qualifiers are dealt with correctly from the start: Auto-remove the superfluous ones, auto-correct the spelling of nom. illeg., remove the comma from ", et al.", and so on.
> 
> I propose to simply omit them all but "ad int."(?), "ined." and "xxx. prov.". Does that work for everyone? Other proposals?

Please consider keeping "nom. illeg.", "nom. illegit.", "nom. inval."

pellaea:
Okay, I'll keep the illeg/inval, too, throw away the rest.  Thanks for the
feedback, Joe and Nimmo.

PR #1307, opened 2023-01-24
Dispatch index to subactions

JoeCohen:
- Dispatches certain `index` actions to subactions using a dispatch table instead of switches.
- Improves ABC metric of those actions, fixing some Rubocop ABC offenses.
- See [@mo-nathan's email](https://groups.google.com/g/mo-developers/c/asJ1AbwHQXQ/m/wdN3tL3eAAAJ) and related discussion.
(This is v2 of now-closed #1304)

There is no manual test.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1541,6 +1541,22 @@ class ApplicationController < ActionController::Base
 
   private ##########
 
+  class << self
+    attr :index_subaction_dispatch_table, :index_subaction_param_keys
+  end
+
+  def dispatch_to_index_subaction
+    self.class.index_subaction_param_keys.each do |subaction|
+      if params[subaction].present?
+        return send(
+          self.class.index_subaction_dispatch_table[subaction] ||
+          subaction
+        )
+      end
+    end
+    default_index_action
+  end
+
   def apply_content_filters(query)
     filters = users_content_filters || {}
     @any_content_filters_applied = false
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1366,7 +1366,24 @@ class ApplicationController < ActionController::Base
   #
   ##############################################################################
 
-  # Render an index or set of search results as a list or matrix. Arguments:
+  class << self
+    attr :index_subaction_dispatch_table, :index_subaction_param_keys
+  end
+
+  # Dispatch to a subaction
+  def index
+    self.class.index_subaction_param_keys.each do |subaction|
+      if params[subaction].present?
+        return send(
+          self.class.index_subaction_dispatch_table[subaction] ||
+          subaction
+        )
+      end
+    end
+    default_index_subaction
+  end
+
+# Render an index or set of search results as a list or matrix. Arguments:
   # query::     Query instance describing search/index.
   # args::      Hash of options.
   #
@@ -1540,22 +1557,6 @@ class ApplicationController < ActionController::Base
   end
 
   private ##########
-
-  class << self
-    attr :index_subaction_dispatch_table, :index_subaction_param_keys
-  end
-
-  def dispatch_to_index_subaction
-    self.class.index_subaction_param_keys.each do |subaction|
-      if params[subaction].present?
-        return send(
-          self.class.index_subaction_dispatch_table[subaction] ||
-          subaction
-        )
-      end
-    end
-    default_index_subaction
-  end
 
   def apply_content_filters(query)
     filters = users_content_filters || {}
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1554,7 +1554,7 @@ class ApplicationController < ActionController::Base
         )
       end
     end
-    default_index_action
+    default_index_subaction
   end
 
   def apply_content_filters(query)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1383,7 +1383,7 @@ class ApplicationController < ActionController::Base
     default_index_subaction
   end
 
-# Render an index or set of search results as a list or matrix. Arguments:
+  # Render an index or set of search results as a list or matrix. Arguments:
   # query::     Query instance describing search/index.
   # args::      Hash of options.
   #
--- a/app/controllers/collection_numbers_controller.rb
+++ b/app/controllers/collection_numbers_controller.rb
@@ -14,7 +14,7 @@ class CollectionNumbersController < ApplicationController
   before_action :pass_query_params, except: :index
   before_action :store_location, except: [:index, :destroy]
 
-  INDEX_SUBACTION_PARAM_KEYS = [
+  @index_subaction_param_keys = [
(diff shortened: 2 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -1541,6 +1541,22 @@ def show_index_of_objects(query, args = {})
...
+    self.class.index_subaction_param_keys.each do |subaction|
+      if params[subaction].present?
+        return send(
+          self.class.index_subaction_dispatch_table[subaction] ||
+          subaction
+        )
+      end
+    end
+    default_index_subaction
+  end
```
JoeCohen, line 1558: @mo-nathan:  Does this look OK?
As you suggested, things break if the key list and the dictionary are constants in the subclasses. 
This is my attempt at instead using class instance variables.
I could use constants if you don't mind including them as arguments. That would also deal with another issue.  See the comment on app/controllers/observations_controller/index.rb

app/controllers/observations_controller/index.rb
```diff
@@ -3,36 +3,10 @@
...
-    elsif params[:where].present?
-      observations_at_where and return
-    elsif params[:project].present?
-      observations_for_project and return
-    elsif params[:by].present? || params[:q].present? || params[:id].present?
-      index_observation and return
-    else
-      list_observations and return
-    end
+  # NOTE: dipatch tables are in ObservationController
```
JoeCohen, line 6: The dispatch table must be a class instance variable of a subclass of ApplicationController.
But I defined it here, it's not a ivar of ObservationController.
@mo-nathan, @nimmolo: Is there a way to define those ivars here as ivars of ObservationsController?

app/controllers/collection_numbers_controller.rb
```diff
@@ -14,16 +14,24 @@ class CollectionNumbersController < ApplicationController
...
+    pattern: :collection_number_search,
+    observation_id: :observation_index,
+    by: :index_collection_number,
+    q: :index_collection_number,
+    id: :index_collection_number
+  }.freeze
+
+  def index
+    dispatch_to_index_subaction
   end
```
JoeCohen, line 35: @mo-nathan: Should I add `index` as a method of ApplicationController and get rid of `dispatch_to_index_subaction`?
The obvious advantages are shorter, DRYer code.
I choose 
```ruby
def index
    dispatch_to_index_subaction
```
because it lets readers of this file know what's going on without having to look at another file.

app/controllers/application_controller.rb
mo-nathan: What you have here looks right to me.

app/controllers/collection_numbers_controller.rb
mo-nathan: I'd get rid of dispatch_to_index_subaction either by renaming it index in ApplicationController or by introducing a class between ApplicationController and these controllers, say IndexedController, that just defines index.  It depends on whether we expect all ApplicationControllers to have some index action or not.

app/controllers/observations_controller.rb
```diff
@@ -29,4 +29,35 @@ class ObservationsController < ApplicationController
...
+    :where,
+    :project,
+    :by,
+    :q,
+    :id
+  ].freeze
+
+  @index_subaction_dispatch_table = {
+    pattern: :observation_search,
+    look_alikes: :observations_of_look_alikes,
```
mo-nathan, line 52: You might consider renaming some of the subaction methods to be the same as the key.  That could be a next step, but it seems a bit weird to have a bunch of actions called "observations_of_"...

app/controllers/observations_controller/index.rb
JoeCohen: @mo-nathan Thanks for the review!
I will keep them in ObservationController.

app/controllers/collection_numbers_controller.rb
JoeCohen: 1. I will rename `ApplicationController#dispatch_to_index_subaction` to `index`, 
and, in the relevant subclasses delete `dispatch_to_index_subaction` and `index` (which is now redundant).
2. There are many controllers that lack any kind of `index` action.
3. I'd rather leave an intermediate controller to a future PR; I want to keep this PR relatively small and uncomplicated.

app/controllers/observations_controller.rb
nimmolo: No problem. Sounds good - I had imagined at some point to go back through and try making the index params more consistent. thanks!

app/controllers/species_lists_controller.rb
```diff
@@ -42,22 +42,25 @@ class SpeciesListsController < ApplicationController
...
+    :by_user,
+    :for_project,
+    :by
+  ].freeze
+
+  @index_subaction_dispatch_table = {
+    pattern: :species_list_search,
+    by_user: :species_lists_by_user,
+    for_project: :species_lists_for_project,
+    by: :by_title_or_selected_by_query
```
JoeCohen, line 56: Dispatches to subaction which merely conditionally switches to different subactions.

app/controllers/species_lists_controller.rb
```diff
@@ -42,22 +42,25 @@ class SpeciesListsController < ApplicationController
...
-      species_lists_for_project
-    elsif params[:by] == "title"
-      species_lists_by_title
-    elsif params[:by].present?
-      index_species_list
-    else
-      list_species_lists
-    end
+  @index_subaction_param_keys = [
+    :pattern,
```
JoeCohen, line 46: @nimmolo & @pellaea: I deleted `advanced_search` because I don't think there's a Species List advanced search.
Is that okay?

pellaea:
Yes, thank you.

JoeCohen:
I think this is ready for review and merger.
The changes to the controllers are routine except:
- for Observations, the dispatch table is in different file from the `index` action and subactions
- one of the SpeciesLists subactions merely switches to other subactions.

I'd like to deal with name changes and a possible intermediate ApplicationIndexController in separate PRs.
@nimmolo: Would you be okay if I (in another PR) I changed subaction names like this:
- If possible make the subaction name the same as the param key.
Ex: `pattern` instead of `<model>_search`.
This lets shorten the dispatch table / dictionary.
- Else if possible, use the same name in different controllers.
Ex: something like `results_of_query` instead of `index_<model>`
- Else if possible, something that makes redundant any accompanying comment. 
This allows the comments to be deleted.

app/controllers/species_lists_controller.rb
pellaea: (I thought I replied to this...)

Yes, that's fine, thank you.

pellaea:
I approve.

JoeCohen:
Thanks @pellaea!
I wait to see if @mo-nathan weighs in again.

nimmolo:
I like the sub-method renaming proposal above.

app/controllers/species_lists_controller.rb
nimmolo: 👍

PR #1316, opened 2023-01-28
"Observations Needing Identification" UI

nimmolo:
## New Structural Features ##
- New Index "Observations Needing Identification"
- New "Observations" left menu item: "Help Identify"
- New Controller: `ObservationsIdentifyController` to generate the index
- New Query: `:observations` `needs_id` to generate the index query
- New Observation Model scopes: `needs_identification`, `with_vote_by_user`, `without_vote_by_user`, `reviewed_by_user`, `not_reviewed_by_user`, `without_confident_name` for future use
- New db column: `observation_views.reviewed` to register that user does not want to further review this observation
- New Controller: `ObservationViewsController` to set `reviewed` state
- New Pattern Search `params[:needs_id]` added to enable routing and rendering of search results directly on this index

#### New matrix_box elements #### 
- Naming vote select directly below the name for Observations with unconfident namings, so that you can boost/downgrade an already proposed name. This is useful in case the name you'd propose has already been proposed, but with a weak vote. (Otherwise you could go to the show_obs page, but it's so slow)
-  "Propose a name" button directly below the name for Observations with no namings (showing `Fungi` as default)
- "Mark as reviewed" checkbox at the bottom of the matrix_box, to clear the obs from subsequent loads of the "Obs Needing ID" index for the current user

#### New lightbox caption features #### 
- Observation info (from previous PR)
- "Propose a Name" button (launches the new modal naming form)
- "Mark as reviewed" checkbox in the caption to clear the obs from subsequent loads of the "Obs Needing ID" index for the current user
- Lightbox keyboard shortcuts `<-` and `->` to navigate to the previous or next unidentified observation. 
- New Lightbox keyboard shortcuts use keys near the cursor keys, to keep it accessible. 
  - `return/enter` to launch the "Propose a name" modal 
  - `/` to "Mark as reviewed"

#### New Tests ####
- `Observations::IdentifyControllerTest`, `ObservationViewsControllerTest`, and in `LurkerTest`
We don't have JS testing yet.


nimmolo:
This is finally all tested and ready to go.

But there's one thing I don't like. 
If you type a pattern search into the search bar **from this index** and commit, it doesn't update **this index**. 

It redirects to the obs index, and does save your query. You can then come back to obs_needing_ids, and your obs are in fact filtered.  That's good.

But i'd like to be able to send a param with "pattern search" that we're on a different query page, and this index is the one to come back to.

nimmolo:
Maybe "Observations needing identification" should be one of the `pattern_search` options in the dropdown?
It will really throw off the layout, though.

Otherwise it could be a hidden field...`params[:needs_id] = true` that gets added to the form `if @controller == IdentifyController` ?

nimmolo:
Whew. Looked into this. 

It's complicated by the fact that a pattern search has to hit the obs index so the query is stored by `PatternSearch::Observation.new(pattern)`. Unless I introduce that functionality here, hmm.

I guess i could pass another param and introduce a conditional in `ObservationsController#index`:
```ruby
  def render_observation_search_results(pattern, needs_id = false)
    search = PatternSearch::Observation.new(pattern)
    if search.errors.any?
      search.errors.each do |error|
        flash_error(error.to_s)
      end
      render("index", location: observations_path)
    else
      if needs_id
        redirect_to(controller: "/observations/identify", action: :index) and return
      end
      @suggest_alternate_spellings = search.query.params[:pattern]
      show_selected_observations(
        search.query, no_hits_title: :title_for_observation_search.t
      )
    end
  end
```
Maybe that's too much for this PR though.

pellaea:
Good point, Nimmo.  You're right, that would feel odd.  Can you check
HTTP_REFERER (or whatever it's called) in the pattern_search action?  Too
cheesy?

nimmolo:
Thanks. true… That seems brittle, though, with the hardcoded referer.

It’s not too hard to print that hidden field in the search form, and pass the additional param through. 
No? Search would send from this page (only) with both params[:search] and params[:needs_id].

Handling it adds a bit of overhead to the search controller, and the obs index, passing the extra param around.

Or can :needs_id be nested under params[:search]? 
The problem with that, when i just tried it, was that the controller then thinks we’re looking for “needs_id"

nimmolo:
Think I've got it now. 
Made a PR off this branch to enable direct filtering of this index.

nimmolo:
- Search bar pattern searches originating from this index now render results on this index.
- New custom keyboard bindings for Lightbox (Propose = `enter`, Reviewed = `/`)

pellaea:
Cool, I'd like to try this out later!

nimmolo:
Wheeee-ew!  Bullet had something to say about it. 
I added a `LurkerTest` to filter the index, and it had a long list of who's naughty and nice.

Now adding those includes to the index args.

app/controllers/observations/identify_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      old_query = find_or_create_query(:Observation)
+      new_query = Query.lookup_and_save(old_query.model, old_query.flavor,
+                                        old_query.params.merge(needs_id: true))
+      show_index_of_objects(new_query,
+                            { matrix: true,
+                              include: [:name, :location, :user, :namings,
+                                        :rss_log, { name: :synonym },
+                                        { namings: :name },
+                                        { thumb_image: :image_votes }] })
+    end
```
nimmolo, line 21: Q: Are there any duplicates here? 
Example: I'm not sure if `{ name: :synonym }` implies you're including `name` already.

app/controllers/search_controller.rb
```diff
@@ -24,7 +24,13 @@ def pattern
...
     session[:search_type] = type
 
-    special_params = type == :herbarium ? { flavor: :all } : {}
+    special_params = if type == :herbarium
+                       { flavor: :all }
+                     elsif params[:needs_id] == "true"
+                       { needs_id: true }
+                     else
+                       {}
+                     end
```
nimmolo, line 34: This keeps us from having to add a fourth positional arg.

app/controllers/search_controller.rb
```diff
@@ -71,7 +78,8 @@ def forward_pattern_search(type, pattern, special_params)
          :observation, :project, :species_list, :user
       redirect_to_search_or_index(
         pattern: pattern,
-        search_path: send("#{type.to_s.pluralize}_path", pattern: pattern),
+        search_path: send("#{type.to_s.pluralize}_path",
+                          params: { pattern: pattern }.merge(special_params)),
         index_path: send("#{type.to_s.pluralize}_path", special_params)
```
nimmolo, line 83: As of now, `special_params` for the `search_path` will only add a param if it's coming from obs_needing_ids.

pellaea:
Yes, `include: { name: :synonym }` includes `:name`.  But `{ namings: :name
}` does not!

config/locales/en.txt
```diff
@@ -1379,6 +1380,12 @@
   create_naming_title: "Propose Name for Observation #[id]"
   create_new_naming_warn: Sorry, someone else has given this a positive vote, so we had to create a new Naming to accomodate your changes."
 
+  # observations/identify
+  obs_needing_id: Observations Needing Identification
+  obs_needing_id_intro: "Here you can quickly scan unidentified observations, propose a name, or vote on a current name. Use the search bar above to filter location, taxon or user.\n\nIf you feel you can't improve on an identification, \"[:mark_as_reviewed]\" and it won't come up again. (It will still appear for others.)\n\n*\"Lightbox\" shortcuts:*\n[&larr;] [&rarr;] move between observations, [return] propose a name, [/] mark as reviewed."
+  mark_as_reviewed: Mark as reviewed
```
nimmolo, line 1386: Please review the intro I wrote here, it could be better. Please edit directly on this PR!

Also - there's something wrong with my Textile, because it gets cut off, but i can't figure out what.

test/controllers/observation_views_controller_test.rb
```diff
@@ -0,0 +1,25 @@
...
+    obs = Observation.needs_identification(users(:mary))
+    obs_count = obs.count
+
+    # Have to create the o_v, none existing
+    obs.take(5).pluck(:id).each do |id|
+      ObservationView.create({ observation_id: id,
+                               user_id: users(:mary).id,
+                               reviewed: true })
+    end
+
```
nimmolo, line 20: TODO: This test needs to hit the controller.

Needs to send a bunch of `put(:update, params)` instead of directly changing OV, here

nimmolo: Improved controller and fixed test.

nimmolo:
I think this thing is good to go. 
It certainly meets Apple standards for a 1.0 release. 
(that is a joke)

JoeCohen:
Suggestion: 
Mark it as reviewed if the user suggests a Name and the result is that the Obs has a high enough confidence level (whatever number you picked in the first place as the threshold for including Observations in Needs identification).

JoeCohen:
Suggestion:
Clicking on the image should have the same effect as clicking on the link to the Obs.
(I think identifiers will often want to view more than the first image. My suggestion would make that a little bit easier.)

nimmolo:
re: mark as reviewed.... yes, that's a somewhat confusing aspect of the UI, isn't it.

When you propose a name, it is taken out of the "Help Identify" query on the next load.
The same thing happens if you "Mark as reviewed". 

The difference, though, is in an outside case. These are two different situations in the db, and behave a bit differently. 
- If you mark it as reviewed, it sets the column in "ObservationViews.reviewed". You'll never get that obs in your "Needing ID's" feed again. It's effectively an "Ignore This" button.
- If a relative newbie "proposes a name" on an unidentified obs, say, with middling confidence, and somebody more senior comes through and quashes the name with a downvote, it will again have a low confidence, and can reappear in the newbie's "Needing ID's" query. That could be useful feedback for them — I figured that automatically changing the "ObsViews" column would prevent that feedback.

nimmolo:
re: clicking on image

As of the day before yesterday, that should be the case. The image link is the same. 

Maybe check to be sure you have the updated branch...

app/controllers/observations_controller/index.rb
```diff
@@ -114,12 +114,22 @@ def render_observation_search_results(pattern)
       search.errors.each do |error|
         flash_error(error.to_s)
       end
-      render("index", location: observations_path)
+      if params[:needs_id]
+        redirect_to({ controller: "/observations/identify", action: :index,
```
JoeCohen, line 118: Probably outside the scope of this PR:
It would be nice to supply a path here instead of hash.
That would required that our `ApplicationController#redirect_to` work with a string, just like [ActionController#redirect_to](nimmo-observations-needing-identifications).
I dont think it would need many changes.

test/integration/capybara/lurker_test.rb
```diff
@@ -205,6 +205,21 @@ def test_search
...
 
+  def test_search_from_obs_needing_ids
+    login
+
+    visit("/observations/identify")
+    # Search for a location.
+    place = "Massachusetts"
+    fill_in("search_pattern", with: place)
+    click_button("Search")
+    assert_match(/#{:obs_needing_id.t}/, page.title, "Wrong page")
```
JoeCohen, line 216: If I replicate this test manually:
- Go to http://localhost:3000/observations/identify
- Fill in Search box with "Massachusetts"
- Click Search
I end up at `Observations Matching ‘Massachusetts’` rather than  `Observations Needing Identification`

(Is this what's supposed to happen?:
Starting on /observations/identify, I can perform an Observation pattern search, 
which gets me Observations needing ID's limited by that search.
That would be incredibly useful!)

test/controllers/observations/identify_controller_test.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+module Observations
+  class IdentifyControllerTest < FunctionalTestCase
+    def test_identify_observations_index
```
JoeCohen, line 7: I personally like (and I think @mo-nathan likes) one request per controller test method.
This reads more like an integration test.

JoeCohen (commented):
I did a little manual testing + a cursory look at the PR changes.
This is very cool. Thank. you for all the work.

test/integration/capybara/lurker_test.rb
nimmolo: Joe - I believe your local server is almost certainly running the wrong version. This works for me locally.

Maybe something's amiss with your version control client? Or, you may just need to `Vagrant reload` and reboot the server. Are you still using VSCode for GitHub? 
Just asking because this seems to have happened a lot recently.

JoeCohen:
Thanks Nimmo.
In about 30-40 minutes I'll check the version. (I have some other stuff I
have to do right now.)
I'm still using VSCode. I don't have an extension named "VSCode for GitHub"

nimmolo:
Sounds good, thanks.
I don't use VSCode for version control — it seems buggy enough at some of its core functions, although I still use it for editing.  On the other hand, there may be a manual step you're unintentionally omitting, like pulling the latest version of the branch from source.

I haven't had any trouble with GitHub Desktop in a long time, that's what I'd recommend.

test/controllers/observations/identify_controller_test.rb
nimmolo: True. I wanted to write an integration test, but the JS blocks that for now. 

This seemed DRYer, although i know there's a way to refactor so that the repetitive steps are performed each time. That has some advantages, but it also makes the test slower. I could be convinced, though.

JoeCohen:
I was behind.
I'm simply not noticing when that happens. VSCode used to warn me about
that.
I am using GitLens.

pellaea:
I tested things a bit.  There are still the same issues about input focus, e.g. after proposing a name you have to click somewhere before it will let you use the arrow keys again, but as you said before, these are details that can be worked out later.

One thing isn't working at all for me, though.  If I change a vote (in regular not lightbox view) it just spins away.  It does actually record the vote (reload the page and it changes).

I'm mildly concerned about losing precious real estate with the help message at the top.  Not critical for this PR, but maybe we can stuff it in the "how to use" page and just put a link at the top of the page, just one line lost.

pellaea:
PS. No javascript errors reported when AJAX returns and the result is ignored.

nimmolo:
@pellaea good catch, i fixed the whirly selector so it does disappear after voting via the regular view.

Also added a js focus change after proposing a name in Lightbox, to get you back to the keyboard shortcuts.

test/controllers/observations/identify_controller_test.rb
nimmolo: (The issue, by the way, is that I have to make two requests for each test... because what i'm testing is if underlying alterations to the db alter the query results. So I need a before/after. Maybe i should move most of this to an integration test? But it's not strictly an integration test since it does the db alterations directly.)

nimmolo:
Please feel free to merge and deploy, if that fix works. I think it's ready for the world.
We can continue to improve it.

PR #1317, opened 2023-01-28
Update Account Verifications

JoeCohen:
- Cease emailing webmaster every time a verification email is send to a hotmail or live.com address.
(Those emails to webmaster were used to debug hotmail's rejecting us as spam, which has since been fixed.)
- Refactors terminating new user signups when user is malicious.

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -74,10 +74,7 @@ class AccountController < ApplicationController
     UserGroup.create_user(@new_user)
     flash_notice(:runtime_signup_success.tp + :email_spam_notice.tp)
     VerifyMailer.build(@new_user).deliver_now
-    if verification_emails_blocked_as_spam?(user) ||
-       bogus_login?(user)
-      notify_root_of_blocked_verification_email(@new_user)
-    end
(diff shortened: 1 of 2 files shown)


JoeCohen:
Temporarily close. There may be one improvement possible.

app/controllers/account_controller.rb
```diff
@@ -111,17 +103,38 @@ def strip_new_user_param(arg)
...
   # it's a known spammer.
-  def block_evil_signups!
-    return false unless evil_signup_credentials?
-
-    true
+  def abort_signup_with_client_error
+    # Too Many Requests == 429. Any 4xx status (Client Error) would also work.
+    render(status: :too_many_requests,
+           content_type: "text/plain",
+           plain: "We grow weary of this. Please go away.")
```
pellaea, line 110: I seem to remember one innocent user actually receiving this message accidentally.  Who knows how many were so offended by it that we never heard from them!  Maybe it's time to soften it?  "We've received a lot of malicious sign-up attempts that match your chosen login name or email address.  Please contact us if you are not a robot."  Would that work okay?

JoeCohen: Good point. I'll change the message.

PR #1318, opened 2023-01-29
Improve coverage of ApplicationController#show_index_of_objects

mo-nathan:
Working towards getting 100% coverage before refactoring.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1436,57 +1436,7 @@ class ApplicationController < ActionController::Base
     query_params_set(query)
 
     # Supply default error message to display if no results found.
-    @error ||= case query.flavor
-               when :all
-                 :runtime_no_objects.t(type: type)
-               when :at_location
-                 loc = query.find_cached_parameter_instance(Location, :location)
-                 :runtime_index_no_at_location.t(type: type,
-                                                 location: loc.display_name)
-               when :at_where
-                 :runtime_index_no_at_location.t(
-                   type: type, location: query.params[:location]
-                 )
-               when :by_author
-                 user = query.find_cached_parameter_instance(User, :user)
-                 :runtime_user_hasnt_authored.t(type: type,
-                                                user: user.legal_name)
-               when :by_editor
-                 user = query.find_cached_parameter_instance(User, :user)
-                 :runtime_user_hasnt_edited.t(type: type, user: user.legal_name)
-               when :by_rss_log
-                 :runtime_index_no_by_rss_log.t(type: type)
-               when :by_user
-                 user = query.find_cached_parameter_instance(User, :user)
-                 :runtime_user_hasnt_created.t(type: type,
-                                               user: user.legal_name)
-               when :for_target
-                 :runtime_index_no_for_object.t(type: type)
-               when :for_user
-                 user = query.find_cached_parameter_instance(User, :user)
-                 :runtime_index_no_for_user.t(type: type, user: user.legal_name)
-               when :in_species_list
-                 spl = query.find_cached_parameter_instance(SpeciesList,
-                                                            :species_list)
-                 :runtime_index_no_in_species_list.t(type: type,
-                                                     name: spl.title)
-               when :inside_observation
-                 id = query.params[:observation]
-                 :runtime_index_no_inside_observation.t(type: type, id: id)
-               when :pattern_search
-                 :runtime_no_matches_pattern.t(
-                   type: type, value: query.params[:pattern].to_s
-                 )
-               when :regexp_search
-                 :runtime_no_matches_regexp.t(type: type,
-                                              value: query.params[:regexp].to_s)
-               when :with_descriptions
-                 :runtime_index_no_with.t(type: type, attachment: :description)
-               when :with_observations
-                 :runtime_index_no_with.t(type: type, attachment: :observation)
-               else
-                 :runtime_no_matches.t(type: type)
-               end
+    @error ||= :runtime_no_matches.t(type: type)
 
     # Get user prefs for displaying results as a matrix.
     if args[:matrix]
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -161,24 +161,21 @@ class ApplicationController < ActionController::Base
 
   # Kick out agents responsible for excessive traffic.
   def kick_out_excessive_traffic
-    return true unless IpStats.blocked?(request.remote_ip)
-    return true if params[:controller] == "account" &&
-                   params[:action] == "login"
-    return true if session[:user_id].present?
+    return true if is_cool?
 
     logger.warn("BLOCKED #{request.remote_ip}")
-    msg = "We have noticed a lot of server-intensive traffic from this IP" \
-          "address (#{request.remote_ip}). There may be better ways of" \
-          "doing what you are trying to do. Please contact the webmaster" \
-          "(#{MO.webmaster_email_address}) so that we can talk about it." \
-          "So that we can best help you, please: \n" \
-          "- include a copy of this message; \n" \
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -1436,53 +1436,57 @@ def show_index_of_objects(query, args = {})
     query_params_set(query)
 
     # Supply default error message to display if no results found.
-    if (query.params.keys - query.required_parameters - [:by]).empty?
```
mo-nathan, line 1439: Currently this PR essentially disables this condition.  I don't see a reason for it.  It only appears to disable the code from providing a more detailed error report is the key in question if there are no non-required parameters.  I tried turning it off in this PR and had to fix two tests and got a whole lot more coverage.

mo-nathan:
An example of the changed behavior on my local system is: http://localhost:3000/names?by_user=10000&id=1

The current code reports: "No matching names found."

The new code reports: "Faun hasn’t created any names."

app/controllers/application_controller.rb
pellaea: I'm trying to puzzle this out, too.  The condition is essentially asking whether the user has customized the search in some way.  That is, added additional filter parameters which are not required by the flavor.  There are a whole slew of these filters one can apply to all observation queries, for example.  You can restrict the query to a given user, certain dates, months of the year, names or clades (Agaricales or lichens, say), confident observations, ones with sequences, etc. etc. etc.

If someone did this, for example, to a "by user" query -- let's say, give me all of Alan's observations restricted to Antarctica.  This is probably empty (although who knows, given it's Alan! :).  This code will produce the error message "There are no observations by Alan".  That seems misleading.  The default error message "there are no matching observations" seems more appropriate.

Of course, I don't know how someone would actually effect such a query.  You can easily add filters like "region:Antarctica" to pattern searches.  But the error message for pattern searches is "no observations matching "blah blah blah"", which seems entirely appropriate.  You can add site-wide content filters via the user prefs.  This is currently restricted to: 1) only show obs in a given clade, 2) only show obs with images, 3) only show obs with specimens, 4) only show obs from a given region.

So here's the question, I guess: What should the error message be when someone asks to see, e.g., all of a give user's observations, but has a content filter which yields an empty result?  That is, say, if you click on "Your Observations" -- "There are no observations by Nathan"?  "There are no matching observations?"  Or... ?

I don't know what the ideal answer is.  But I think this is a fair summary of the context, so that you can make an informed decision about whether to change the behavior or not while refactoring this.

Enjoy! :)

JoeCohen:
@mo-nathan: Thanks for tackling this gnarly method.
@pellaea: Thanks for the very helpful explanation!
I think I prefer the current message:
- The new message seems better (at least at first glance), in [Nathan's example](https://github.com/MushroomObserver/mushroom-observer/pull/1318#issuecomment-1407682705),  
  but the old message is acceptable.
- The new message is misleading in [the case cited by Jason](https://github.com/MushroomObserver/mushroom-observer/pull/1318#discussion_r1090014742). 
For example (via the UI), if:
  -  I change my Preferences to "Oregon, USA"; 
  - do a User-type PatternSearch for "RenegadeMycologist"; 
  - click Observations 

 The new result is "Dušan hasn’t created any observations" for a user who actually created 956 Observations, but none in Oregon. 
I supposed that could be tweaked to something like "Dušan hasn’t created any observations matching your filter". But implementing that that seems overly complicated to, and outside the scope of this PR.

JoeCohen:
How many people actually use `content_filter`?
```ruby
users_with_non_redundant_filters = User.where.not(content_filter: {}).where.not(content_filter: {:has_images=>nil, :has_specimen=>nil, :region=>"", :lichen=>"", :clade=>""}).count
 => 1936 
```

pellaea:
Wow.  I had no idea it would be that popular.  Amazing.

JoeCohen:
> Wow. I had no idea it would be that popular. Amazing.

@pellea: My number is **way** wrong. My AR query doesn't do what I expect. 
```ruby
> users_with_non_redundant_filter = User.where.not(content_filter: {}).where.not(content_filter: {:has_images=>nil, :has_specimen=>nil, :region=>"", :lichen=>"", :clade=>""})
 => 
> users_with_non_redundant_filter.first.content_filter
 => {:has_images=>nil, :has_specimen=>nil, :lichen=>"", :region=>"", :clade=>""} 
```
Note the different order of the keys.  
Maybe the query is converting converting `content_filter` to a string instead of using `Hash#==`? Maybe that's because we serialize the filter?

JoeCohen:
435 is the actual number of users with a non-redundant content_filter. (It's still higher than I expected.)
```ruby
> foo = []
> users_with_non_redundant_filter.each {|u| foo << u unless u.content_filter == {:has_images=>nil, :has_specimen=>nil, :region=>"", :lichen=>"", :clade=>""} }
...
> > foo.length
 => 435 
```
(Couldn't figure out how to do this as an AR query.)

pellaea:
Thanks, Joe!  That's tricky!
-Jason

JoeCohen:
That's still too high. Actual 254. (I had wondered if the number was small enough so that we didn't have to care about the correct error message.)
```ruby
> foo = []
> User.where.not(content_filter: {}).each { |u| foo << u if u.content_filter[:has_images].present? || u.content_filter[:has_specimen].present? || u.content_filter[:lichen].present? || u.content_filter[:region].presen
t? || u.content_filter[:clade].present? }
...
 > foo.length
 => 254
```

app/controllers/application_controller.rb
mo-nathan: Thanks!  I got buried in a semi-urgent disk recovery situation that has been a lot more painful than expected.

Anyway, the challenge I have with the current code is there doesn't appear to be an obvious way to get to the more specific error messages as the code is currently written, but if you turn off that condition you can.  Personally, I don't see a big problem with "There are no observations by Alan" if the person has somehow bent over backwards trying to get observations about Antarctica by Alan, then they should have the context.  However, I'm also fine with the more generic error.  That's certainly the simplest thing to do and it eliminates code which is great for helping with the ABC metric!  Given your descriptions, I'm leaning towards just making it simple and having a single error message.

pellaea:
I can live with any of the three options, including especially deleting the
whole thing and replacing it with a single generic "no matches!" message. :)

app/controllers/application_controller.rb
```diff
@@ -1517,29 +1464,11 @@ def show_index_of_objects(query, args = {})
...
-      @objects = query.paginate(@pages, include: include)
-      @timer_end = Time.current
-      logger.warn("QUERY finished: model=#{query.model}, " \
-                  "flavor=#{query.flavor}, params=#{query.params.inspect}, " \
-                  "time=#{(@timer_end - @timer_start).to_f}")
+      calc_pages(args, query)
+      find_objects(args, query)
 
       # Give the caller the opportunity to add extra columns.
+      #### Refactor problem?????
```
mo-nathan, line 1471: Do we really need to be able to pass a block to this function?  Makes refactoring a lot harder and I was only able to find one seemingly obscure usage of this functionality.

pellaea: The only place it's used is the old /name/authored_names (≡ /names?with_descriptions=1).

My personal opinion is basically that descriptions are so rarely used, contribute so little to our core mission, and contribute so much to our tech debt, that I encourage cutting back on that functionality wherever possible.

JoeCohen (approved):
IMO we should merge this.
- We gain a big benefit from increased coverage of a key method.
- This change in the flash (when an object isn't found) is an improvement for most users. 
- The arguable detriment (caused by the message change) to the other users is:
  - Outweighed by the above benefits, and
  - If anyone complains, can be improved in the future.

pellaea:
Agreed, go for it!

JoeCohen:
Thanks! Will do.

PR #1324, opened 2023-02-01
Improve index coverage

JoeCohen:
- Gets 100% coverage of `index` actions that dispatch on `@index_subaction_dispatch_table`
(This is preliminary to pruning and revising that table.)
- Delivers [Pivotal #184362765](https://www.pivotaltracker.com/story/show/184362765).
- Mostly adds new tests.
- In a few cases modifies misbehaving controller code.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1655,7 +1655,23 @@ class ApplicationController < ActionController::Base
     nil
   end
 
+  # Like find_or_goto_index, but allows redirect to a different index
+  def find_obj_or_goto_index(model:, obj_id:, index_path:)
+    model.safe_find(obj_id) ||
+      flash_obj_not_found_and_goto_index(
+        model: model, obj_id: obj_id, index_path: index_path
+      )
+  end
+
   private ##########
+
+  def flash_obj_not_found_and_goto_index(model:, obj_id:, index_path:)
+    flash_error(
+      :runtime_object_not_found.t(id: obj_id, type: model.type_tag)
+    )
+    redirect_with_query(index_path)
+    nil
+  end
 
   # Redirects to an appropriate fallback index in case of unrecoverable error.
   # Most such errors are dealt with on a case-by-case basis in the controllers,
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -93,7 +93,10 @@ class ImagesController < ApplicationController
 
   # Display matrix of images by a given user.
   def images_by_user
-    user = find_or_goto_index(User, params[:by_user].to_s)
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_user].to_s,
+      index_path: images_path
+    )
     return unless user
 
     query = create_query(:Image, :by_user, user: user)
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -93,11 +93,7 @@ class ImagesController < ApplicationController
 
   # Display matrix of images by a given user.
   def images_by_user
-    user = if params[:by_user]
-             find_or_goto_index(User, params[:by_user].to_s)
-           else
-             @user
-           end
+    user = find_or_goto_index(User, params[:by_user].to_s)
     return unless user
 
     query = create_query(:Image, :by_user, user: user)
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -105,7 +105,7 @@ class LocationsController < ApplicationController
   # Display list of locations that a given user is editor on.
   def locations_by_editor
     user = find_obj_or_goto_index(
-      model: User, obj_id: params[:by_user].to_s,
+      model: User, obj_id: params[:by_editor].to_s,
       index_path: locations_path
     )
     return unless user
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -92,11 +92,7 @@ class LocationsController < ApplicationController
 
   # Display list of locations that a given user is author on.
   def locations_by_user
-    user = if params[:id]
-             find_or_goto_index(User, params[:by_user].to_s)
-           else
-             @user
-           end
+    user = find_or_goto_index(User, params[:by_user].to_s )
     return unless user
 
     query = create_query(:Location, :by_user, user: user)
@@ -105,11 +101,7 @@ class LocationsController < ApplicationController
 
   # Display list of locations that a given user is editor on.
   def locations_by_editor
-    user = if params[:id]
-             find_or_goto_index(User, params[:by_editor].to_s)
-           else
-             @user
-           end
+    user = find_or_goto_index(User, params[:by_editor].to_s)
     return unless user
 
     query = create_query(:Location, :by_editor, user: user)
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -92,16 +92,30 @@ class LocationsController < ApplicationController
 
   # Display list of locations that a given user created.
   def locations_by_user
-    user = User.safe_find(params[:by_user].to_s)
+    user = other_model_obj_or_goto_index(
+      other_model: User,
+      obj_id: params[:by_user].to_s,
+      index_path: locations_path
+    )
+    return unless user
 
-    if user
-      query = create_query(:Location, :by_user, user: user)
-      show_selected_locations(query, link_all_sorts: true)
-    else
-      flash_error(:runtime_object_not_found.t(id: params[:by_user].to_s,
-                                              type: User.type_tag))
-      redirect_to(locations_path)
-    end
+    query = create_query(:Location, :by_user, user: user)
+    show_selected_locations(query, link_all_sorts: true)
+  end
+
+  def other_model_obj_or_goto_index(other_model:, obj_id:, index_path:)
+    other_model.safe_find(obj_id) ||
+      flash_other_model_error_and_goto_index(
+        other_model: other_model, obj_id: obj_id, index_path: index_path
+      )
+  end
+
+  def flash_other_model_error_and_goto_index(other_model:, obj_id:, index_path:)
+    flash_error(
+      :runtime_object_not_found.t(id: obj_id, type: other_model.type_tag)
+    )
+    redirect_with_query(index_path)
+    nil
   end
 
   # Display list of locations that a given user is editor on.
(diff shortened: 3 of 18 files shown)


JoeCohen:
🚧 Temporarily closing while I make some repairs.

app/controllers/locations_controller.rb
```diff
@@ -90,13 +90,12 @@ def list_locations
...
   def locations_by_user
-    user = if params[:id]
-             find_or_goto_index(User, params[:by_user].to_s)
-           else
-             @user
-           end
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_user].to_s,
+      index_path: locations_path
+    )
```
JoeCohen, line 98: Changes functionality when `by_user` is not an existing user:
- in main: display locations of current user
- new: display all locations with a flash error

app/controllers/images_controller.rb
```diff
@@ -93,11 +93,10 @@ def list_images
...
   def images_by_user
-    user = if params[:by_user]
-             find_or_goto_index(User, params[:by_user].to_s)
-           else
-             @user
-           end
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_user].to_s,
+      index_path: images_path
+    )
```
JoeCohen, line 99: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/locations_controller.rb
```diff
@@ -105,11 +104,10 @@ def locations_by_user
...
   def locations_by_editor
-    user = if params[:id]
-             find_or_goto_index(User, params[:by_editor].to_s)
-           else
-             @user
-           end
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_editor].to_s,
+      index_path: locations_path
+    )
```
JoeCohen, line 110: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/names/descriptions_controller.rb
```diff
@@ -79,11 +79,10 @@ def list_name_descriptions
...
     def name_descriptions_by_author
-      user = if params[:by_author]
-               find_or_goto_index(User, params[:by_author].to_s)
-             else
-               @user
-             end
+      user = find_obj_or_goto_index(
+        model: User, obj_id: params[:by_author].to_s,
+        index_path: name_descriptions_path
+      )
```
JoeCohen, line 85: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/names_controller.rb
```diff
@@ -121,7 +121,10 @@ def names_with_descriptions
 
   # Display list of names that a given user is author on.
   def names_by_user
-    user = params[:id] ? find_or_goto_index(User, params[:by_user].to_s) : @user
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_user].to_s,
+      index_path: names_path
+    )
```
JoeCohen, line 127: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/names_controller.rb
```diff
@@ -133,11 +136,10 @@ def names_by_user
...
   def names_by_editor
-    user = if params[:id]
-             find_or_goto_index(User, params[:by_editor].to_s)
-           else
-             @user
-           end
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:by_editor].to_s,
+      index_path: names_path
+    )
```
JoeCohen, line 142: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/names/descriptions_controller.rb
```diff
@@ -92,11 +91,10 @@ def name_descriptions_by_author
...
     def name_descriptions_by_editor
-      user = if params[:by_editor]
-               find_or_goto_index(User, params[:by_editor].to_s)
-             else
-               @user
-             end
+      user = find_obj_or_goto_index(
+        model: User, obj_id: params[:by_editor].to_s,
+        index_path: name_descriptions_path
+      )
```
JoeCohen, line 97: Changes functionality when by_user is not an existing user.
See [Comment about LocationsController](https://github.com/MushroomObserver/mushroom-observer/pull/1324/files#r1095161355).

app/controllers/observations_controller/index.rb
```diff
@@ -28,12 +28,6 @@ def list_observations
     show_selected_observations(query)
   end
 
-  # Displays matrix of all Observations, alphabetically.
-  def observations_by_name
-    query = create_query(:Observation, :all, by: :name)
-    show_selected_observations(query)
-  end
```
JoeCohen, line 35: Removes unreferenced code.
@pellaea, @nimmolo: Am I overlooking some use of this code e.g., in meta-programming?

app/extensions/regexp_extensions.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Regexp
+  # Spaces are not Regexp meta-characters, so
+  # stop Ruby from escaping them
+  # Cf. https://stackoverflow.com/questions/73933394/regexp-escape-adds-weird-escapes-to-a-plain-space
+  def self.escape_except_spaces(str)
+    escape(str).gsub("\\ ", " ")
+  end
+end
```
JoeCohen, line 10: Should I use a different approach?
Back story:
In a couple of places, I wanted to check for the correct flash message with `assert_flash(<matcher>)`,
where <matcher> was something like `:runtime_no_matches.t(type: :observation)`.
See ObservationControllerTest#test_index_at_location_without_observations.

I wanted to use a regexp as the matcher. But the string `:runtime_no_matches.t(type: :observation)` didn't work because the flash included other stuff (in particular <p><\p>). 
I tried turning that string into a regexp by running it through `Regexp.escape` and using the result as the matcher. 
That didn't work because -- to my surprise -- `Regexp.escape` escapes spaces (in addition to meta characters).
So I created this simple `Regexp` extension.

nimmolo: Joe - I wrote a string_extensions method you might be able use here. 
Check out `string_extensions#as_displayed` — it strips out html tags and extra spaces. You'd use it like `assert_flash(<matcher>.as_displayed)`.

But i'm not sure that's what you need. Is it your matcher that has the `<p>` tag, or the flash it's matching against?

JoeCohen:
Thanks Nimmo. I'll have a look at that method.
The tags are in the flash.

JoeCohen:
I could manually add the tags to the matcher.
I didn't like that because it requires additional knowledge about the flash
(and is potentially brittle).

JoeCohen:
@pellaea:
**Can a `Locations#index` html request ever result in a query that's coerceable into a `LocationDescription` query?**

I've 100% covered the relevant controller `index` actions except for one line in `LocationsController`:

Screenshot 2023-02-02 at 3 23 56 PM

app/controllers/application_controller.rb
```diff
@@ -1655,8 +1655,24 @@ def flash_error_and_goto_index(model, id)
     nil
   end
 
+  # Like find_or_goto_index, but allows redirect to a different index
+  def find_obj_or_goto_index(model:, obj_id:, index_path:)
+    model.safe_find(obj_id) ||
+      flash_obj_not_found_and_goto_index(
+        model: model, obj_id: obj_id, index_path: index_path
+      )
+  end
```
JoeCohen, line 1664: I need something like this in order to be able to both:
- flash that an object of class A was not found, and
- redirect to the controller for class B.

app/controllers/observations_controller/index.rb
pellaea: Naw, I never used this even when there was still a way to get to this!

app/extensions/regexp_extensions.rb
pellaea: Yeah, I think (if I understand the situation) I agree with Nimmo.  I think it would be better to remove html markup from the strings before comparing them instead of messing with matchers.  Maybe.  I'm open to being convinced otherwise, though.

pellaea:
Nice work, Joe.

app/extensions/regexp_extensions.rb
JoeCohen: Jason/Nimmo:
I don't get to remove the html markup before `flash_alert(<string or regexp>)` makes the comparison.
So the choices that jump out at me are:
- Use `flash_alert` with a regexp;
```ruby
assert_flash(flash_matcher)
```
- Use `flash_alert`, manually wrap the expected string with tags, 
```ruby
assert_flash(tag.p(:runtime_no_matches.t(type: :observation))
```
- Don't use `flash_alert`. Instead use:
```ruby
assert_equal(:runtime_no_matches.t(type: :observation),
             get_last_flash[1..-1]. # strip the error level code
             strip_html) # strip the p tags
```
What do you like best?

JoeCohen: P.S.: I forgot to mention one of the things that made it hard to create the matcher:
In this case, the flash message contains regexp meta characters. 
So if using a regexp as a matcher, I have to quote those characters, but not spaces.

pellaea:
> @pellaea: **Can a `Locations#index` html request ever result in a query that's coerceable into a `LocationDescription` query?**

Yes, nicely buried!  Sorry I missed that.  No, I don't think it is possible with the existing UI, so go right ahead and remove these lines instead of covering them.  (The same should occur with name_descriptions, right?)

test/controllers/observations_controller_test.rb
```diff
@@ -508,6 +516,58 @@ def test_observation_search_with_spelling_correction
...
+    get(:index, params: params)
+
+    assert_select("#title", text: "Observations from #{location.name}")
+  end
+
+  def test_index_at_location_without_observations
+    location = locations(:unused_location)
+    params = { location: location }
+    flash_matcher = Regexp.new(
+      Regexp.escape_except_spaces(
```
pellaea, line 533: Okay, I see where you're using it.  I've got no problem with this.

JoeCohen:
Thanks so much! I'll remove it.
You're right, the same should with name_descriptions, but ... it's not
there.

JoeCohen:
Thanks for the feedback. I appreciate your opinion.

PR #1325, opened 2023-02-02
Added two new files to nightly csv dumps.

pellaea:
This is in response to a request from some users that want a little more information.

(We might consider zipping these files some day.  images.csv is like 100-200 Mb at this point.  But if we do so it will break any scripts out there that are using the unzipped files.  I don't know any way to warn users before making the change.  The same problem we had with upgrading the API.)


nimmolo:
I guess somebody does use those descriptions...

pellaea:
No, they aren't going to.  They *think* they want them.  It was easier to
just show them the data than to convince them they don't want it! :)

PR #1326, opened 2023-02-02
Slight refactor of Observation form methods

nimmolo:
Moves a couple methods into the `observations_controller/form_helpers.rb` and refactors `create_new_observation` for ABC.

~(was: Fix "Unpermitted parameters" message when reloading create obs form)~

> **Old PR message, for the record only:**
> 
> This makes an "Unpermitted parameters" message go away, that i discovered in the log while testing an unrelated aspect of the create_obs form. (Log output below.) 
> 
> The "issue" doesn't seem to cause any problems and may not be worth fixing; it has been existing for awhile, probably. It arises from a switch in Rails 5, when form params became no longer a hash, but an instance of `ActionController::Parameters`. 
> 
> ### TLDR: if we are permitting nested params, the nested params must come last in the permitted params hash. ###
> 
> More reading on parameter permitting [here](https://flaviabastos.ca/2020/09/30/how-to-permit-nested-parameters-in-rails/) and [here](https://smartlogic.io/blog/permitting-nested-arrays-using-strong-params-in-rails/).
> 
> I'm not sure where else we may have nested params incorrectly (...? incompletely?) permitted, but one place is the create name form. However, the unpermitted params there should not be permitted anyway.
> 
> This Rails switch to a params service object has cropped up and required a change in MO code once before, in [descriptions](https://github.com/MushroomObserver/mushroom-observer/blob/main/app/controllers/concerns/descriptions.rb#L216), but only in checking for the existence of the params.
> 
> Here was the log error, on committing a blank create obs form. I cut at the last line to accentuate the error, but the log continues as the page is reloaded.
> 
> ```shell
>   Parameters: {"utf8"=>"✓", "authenticity_token"=>"suZPk9j30d8PCV2ZA8UF9NT4KPmddRu1kyaD4emFL3MhhDgpZGXhtizNtJaemKEhvyB5viTGh16Q0DPFdfNMcQ==", "observation"=>{"when(3i)"=>"2", "when(2i)"=>"2", "when(1i)"=>"2023", "place_name"=>"", "is_collection_location"=>"1", "lat"=>"", "long"=>"", "alt"=>"", "gps_hidden"=>"0", "specimen"=>"0", "notes"=>{"Other"=>""}, "thumb_image_id"=>""}, "naming"=>{"name"=>"", "vote"=>{"value"=>""}, "reasons"=>{"1"=>{"check"=>"0", "notes"=>""}, "2"=>{"check"=>"0", "notes"=>""}, "3"=>{"check"=>"0", "notes"=>""}, "4"=>{"check"=>"0", "notes"=>""}}}, "was_js_on"=>"yes", "collection_number"=>{"name"=>"Darvin DeShazer", "number"=>""}, "herbarium_record"=>{"herbarium_name"=>"Darvin DeShazer (darv): Personal Herbarium", "herbarium_id"=>"", "notes"=>""}, "good_images"=>"", "project"=>{"id_3"=>"0", "id_162"=>"0", "id_295"=>"0", "id_222"=>"0"}, "list"=>{"id_114"=>"0", "id_31"=>"0", "id_185"=>"0", "id_75"=>"0", "id_102"=>"0", "id_1201"=>"0", "id_99"=>"0", "id_58"=>"0", "id_23"=>"0", "id_6"=>"0", "id_24"=>"0", "id_8"=>"0", "id_11"=>"0", "id_338"=>"0", "id_186"=>"0", "id_1277"=>"0"}, "approved_name"=>""}
>   User Load (0.4ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 2 LIMIT 1
>   User Load (0.5ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 2 LIMIT 1
> user=2 robot=N ip=10.0.2.2
>   TRANSACTION (0.2ms)  BEGIN
>   User Load (0.6ms)  SELECT `users`.* FROM `users` WHERE `users`.`login` = 'darv' LIMIT 1
>   User Update (1.2ms)  UPDATE `users` SET `users`.`updated_at` = '2023-02-02 19:26:25', `users`.`last_activity` = '2023-02-02 19:26:25' WHERE `users`.`id` = 2
>   TRANSACTION (2.7ms)  COMMIT
> [I18n] loading locale: en from @user
>   Language Load (0.4ms)  SELECT `languages`.* FROM `languages` WHERE `languages`.`locale` = 'en' LIMIT 1
> [I18n] Locale set to en
>   CACHE Language Load (0.0ms)  SELECT `languages`.* FROM `languages` WHERE `languages`.`locale` = 'en' LIMIT 1
>   UserGroup Load (42.1ms)  SELECT `user_groups`.* FROM `user_groups` INNER JOIN `user_group_users` ON `user_groups`.`id` = `user_group_users`.`user_group_id` WHERE `user_group_users`.`user_id` = 2
>   User Load (0.6ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 2 LIMIT 1
> Unpermitted parameter: :notes   << THIS
> ```

### DIFF
--- a/app/controllers/observations_controller/form_helpers.rb
+++ b/app/controllers/observations_controller/form_helpers.rb
@@ -16,10 +16,38 @@
 module ObservationsController::FormHelpers
   private
 
-  def whitelisted_observation_args
+  # Note this array syntax. VSCode won't allow nested square brackets (?).
+  # The nested attributes MUST come last.
+  def permitted_observation_args
     [:place_name, :where, :lat, :long, :alt, :when, "when(1i)", "when(2i)",
-     "when(3i)", :notes, :specimen, :thumb_image_id, :is_collection_location,
-     :gps_hidden]
+     "when(3i)", :specimen, :thumb_image_id, :is_collection_location,
+     :gps_hidden, { notes: [:Other] }]
+  end
+
+  def update_permitted_observation_attributes
+    @observation.attributes = permitted_observation_params || {}
+  end
+
+  # NOTE: You MUST call `to_h` on the permitted params, because param nesting.
+  # As of rails 5, params are an ActionController::Parameters object,
+  # not a hash.
+  def permitted_observation_params
+    return unless params[:observation]
+
+    params[:observation].permit(permitted_observation_args).to_h
+  end
+
+  # Symbolize keys; delete key/value pair if value blank
+  # Also avoids whitelisting issues
+  def notes_to_sym_and_compact
+    return Observation.no_notes unless notes_param_present?
+
+    symbolized = params[:observation][:notes].to_unsafe_h.symbolize_keys
+    symbolized.compact_blank!
+  end
+
+  def notes_param_present?
+    params[:observation] && params[:observation][:notes].present?
   end
 
   def init_license_var
@@ -120,7 +148,7 @@ module ObservationsController::FormHelpers
             name = upload.original_filename.force_encoding("utf-8")
           end
           # image = Image.new(args2) # Rails 3.2
-          image = Image.new(args2.permit(whitelisted_image_args))
+          image = Image.new(args2.permit(permitted_image_args))
           # image = Image.new(args2.permit(:all))
           image.created_at = Time.zone.now
           image.updated_at = image.created_at
@@ -175,7 +203,7 @@ module ObservationsController::FormHelpers
       args = param_lookup([:good_image, image.id.to_s])
       next unless args
 
-      image.attributes = args.permit(whitelisted_image_args)
+      image.attributes = args.permit(permitted_image_args)
       next unless image.when_changed? ||
                   image.notes_changed? ||
                   image.copyright_holder_changed? ||
--- a/app/controllers/observations_controller/form_helpers.rb
+++ b/app/controllers/observations_controller/form_helpers.rb
@@ -16,8 +16,8 @@
 module ObservationsController::FormHelpers
   private
 
-  # Note this array syntax. VSCode won't allow nested square brackets (?).
-  # The nested attributes MUST come last.
+  # Note this array syntax: The nested attributes MUST come last.
+  # VSCode won't allow nested square brackets (?).
   def permitted_observation_args
     [:place_name, :where, :lat, :long, :alt, :when, "when(1i)", "when(2i)",
      "when(3i)", :specimen, :thumb_image_id, :is_collection_location,
--- a/app/controllers/observations_controller/form_helpers.rb
+++ b/app/controllers/observations_controller/form_helpers.rb
@@ -16,19 +16,18 @@
 module ObservationsController::FormHelpers
   private
 
-  # Note this array syntax: The nested attributes MUST come last.
-  # VSCode won't allow nested square brackets (?).
+  # NOTE: potential gotcha... Any nested attributes must come last.
   def permitted_observation_args
     [:place_name, :where, :lat, :long, :alt, :when, "when(1i)", "when(2i)",
-     "when(3i)", :specimen, :thumb_image_id, :is_collection_location,
-     :gps_hidden, { notes: [:Other] }]
+     "when(3i)", :notes, :specimen, :thumb_image_id, :is_collection_location,
+     :gps_hidden]
   end
 
   def update_permitted_observation_attributes
     @observation.attributes = permitted_observation_params || {}
   end
 
-  # NOTE: You MUST call `to_h` on the permitted params, because param nesting.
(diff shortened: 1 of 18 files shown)


mo-nathan:
Seems like it shouldn't be hard to create test for this.

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -16,20 +16,38 @@
...
+  # Also avoids whitelisting issues
+  def notes_to_sym_and_compact
+    return Observation.no_notes unless notes_param_present?
+
+    symbolized = params[:observation][:notes].to_unsafe_h.symbolize_keys
+    symbolized.compact_blank!
+  end
+
+  def notes_param_present?
+    params[:observation] && params[:observation][:notes].present?
```
pellaea, line 50: May I ask a favor?  If it's really important to you to move methods around like this, please ignore the request, but...  Where possible, it is better from the perspective of reviewers if you don't move methods around.  When you do there is no way to (easily) see what (if anything) has been changed.  It just makes reviewing and understanding your PR more difficult.  I'm sure there will be excellent cases for moving things, but just in case you aren't aware of that side-effect (messing up the diffs), I thought I'd throw that out there.

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -16,20 +16,38 @@
...
   private
 
+  # Note this array syntax: The nested attributes MUST come last.
+  # VSCode won't allow nested square brackets (?).
   def permitted_observation_args
     [:place_name, :where, :lat, :long, :alt, :when, "when(1i)", "when(2i)",
-     "when(3i)", :notes, :specimen, :thumb_image_id, :is_collection_location,
-     :gps_hidden]
+     "when(3i)", :specimen, :thumb_image_id, :is_collection_location,
+     :gps_hidden, { notes: [:Other] }]
```
pellaea, line 24: So, what I'm still confused about -- sorry! -- is how this ever passed tests.  The log message claimed "notes" was unpermitted.  Surely we have a test which involves creating an observation with notes, no?

Also, why single out :Other here, and none of the other notes fields?  I mean, sure, that's user-defined, so how do you apriori know which notes fields to permit?

Yea, I guess I'm just confused.  Like you say, this must already work... somehow...  It just bugs me when I can't imagine how.  Or at least understand why it's displaying an error message in the log that apparently doesn't mean anything(!)

Again, feel free to ignore and move on to greener pastures.  I have no problem with "don't fix it if it ain't broke".  Just hoping you have some amazing new insight that will make all sorts of heavenly lights go on! :)

mo-nathan:
This seems to be related to some complex parsing going on in app/models/observation.rb that @JoeCohen last touched back in 2017.  It doesn't look familiar to me and it seems like that form is getting rendered wrong since I see 

`<textarea class="form-control" rows="10" id="observation_notes_Other" name="observation[notes][Other]"></textarea>`

in the HTML which doesn't make sense to me.  Why isn't it just "observation[notes]"?

pellaea:
Ooh!  I can answer that!  Because users can provide a "notes template", basically a list of fields.  "Other" is the catch-all at the end that we tack onto their list of requested fields.  You probably are logged in as a user who doesn't have any notes template fields specified, so it looks strange.

app/controllers/observations_controller/form_helpers.rb
JoeCohen: I'm equally confused. (And just not focussing well this a.m.)

app/controllers/observations_controller/form_helpers.rb
JoeCohen, line 50:
```suggestion
    params.dig(:observation, :notes).present?
```

nimmolo: @JoeCohen thanks — I'd forgotten about `dig`, good suggestion. 

MO has a helper method that does something similar, `param_lookup` in application_controller. Maybe a good idea to do a PR for a global replace for this method — although it is not applied as often as would be useful, `dig` fills the bill.

I may be retiring this PR, because I misunderstood the original issue. Nathan and Jason have sorted me out.

nimmolo: Sure thing. I'll move in a separate PR. 

Moved these either because they're used by both `edit_and_update` and `new_and_create` or were closely interrelated with methods that were.

nimmolo:
Retiring this PR, unless there's anything here about the params hash worth keeping. 

Thanks for the diagnosis

nimmolo:
Actually there's some ABC reduction here, that i might as well keep. 
Re-opening and renaming the PR.

PR #1328, opened 2023-02-03
Adjust Rubocop config

JoeCohen:
- Changes Max Abc size from 20 to the default (currently 17). 
- Removes (almost) all in-code disables of the Abc cop.
- Disables the cop that requires top-level documentation of classes and modules ([Style/Documentation](https://docs.rubocop.org/rubocop/1.42/cops_style.html#styledocumentation)).
- Regenerates the Rubocop todo file.

### Notes

- I don't understand the interaction among Codeclimate, Rubocop, and the Rubocop todo file.[^1] 
- I might need to update the PR to deal with that. 
- If Codeclimate (or Rubocop) spits out Abc violations that you don't want to fix in the PR, feel free to 
  - Use Codeclimate's wontfix feature;
  - Ignore them; 
  - If you think think that fixing them would make the code worse, disable them with a comment.

---

[^1]: The Rubocop todo file disables Metric violations in a unique way. 
Instead of disabling the cop completely or for a list of files, it sets the Max Abc size.


JoeCohen:
Temporarily closing this PR until I have a better understanding of Codeclimate/Rubocop/`.rubocop_todo.yml` interaction.

JoeCohen:
Please ignore this PR until it changes from Draft to Ready for Review.
(I need to experiment with Codeclimate/Rubocop/.rubocop_todo.yml but can do that only if this is a PR, rather than a mere branch.)

JoeCohen:
I will close this branch and create a fresh one with a clean Git history.
After all my experiments, I still don't know how Codeclimate interacts with Rubocop with respect to Abc and the Rubocop todo list. So I'll just do the simplest thing for now, and make adjustments later.

PR #1330, opened 2023-02-05
Rubocop config adjustment

JoeCohen:
- Changes [Metrics/AbcSize](https://docs.rubocop.org/rubocop/1.42/cops_metrics.html#metricsabcsize) Max from 20 to the default (currently 17). 
- Removes all in-code disables of [Metrics/AbcSize](https://docs.rubocop.org/rubocop/1.42/cops_metrics.html#metricsabcsize).
- Disables the cop that requires top-level documentation of classes and modules ([Style/Documentation](https://docs.rubocop.org/rubocop/1.42/cops_style.html#styledocumentation)).
- Regenerates the Rubocop todo file.

### Notes

- I don't understand the interaction among Codeclimate, Rubocop, and the Rubocop todo file.[^1] I will have to see how things play out in practice.
- If Codeclimate (or Rubocop) spits out Abc violations that give you don't want to fix in the PR, feel free to 
  - Use Codeclimate's wontfix feature;
  - Ignore them; 
  - If you think think that fixing them would make the code worse, disable them with a comment.

---

[^1]: The Rubocop todo file disables Metric violations in a unique way. 
Instead of disabling the cop completely or for a list of files, it sets the Max Abc size.


JoeCohen:
Closing because this branched from the wrong place. I will redo it.

PR #1336, opened 2023-02-08
Make Query default_order available 

JoeCohen:
- Uses Query as the "source of truth" for default sort sort order.
- Makes the default sort order available without instantiating a query as:
```ruby
::Query::<model>Base.default_order
```
Example:
```ruby
::Query::SpeciesListBase.default_order
```


pellaea:
Looks right to me.

PR #1338, opened 2023-02-08
Slight refactor of naming vote AJAX

nimmolo:
This fixes a mistake I made in the JS that was intended to update the "community vote" percentage when you vote on a naming. (The % was updated in the db, and showed up on page refresh, but should update it "live" on the page. The prob was an errant CSS selector.)

It also moves the ajax HTML response into a partial. 
(Was `render(inline: string_of_html)` now: `render(partial: "observations/namings/votes/ajax_response")`

### DIFF
--- a/app/assets/javascripts/naming_vote_ajax.js
+++ b/app/assets/javascripts/naming_vote_ajax.js
@@ -33,26 +33,26 @@ function VoteByAjaxModule(translations) {
         }
 
         // bootstrap modal printed in layout already, just activate it
-        $('#naming_ajax_progress_caption').append(
+        $('#naming_ajax_progress_caption').empty().append(
           $("<span>").text(translations.show_namings_saving + "... "),
           $("<span class='spinner-right mx-2'></span>")
         );
         $("#naming_ajax_progress").modal('show');
 
         $.ajax("/ajax/vote/naming/" + naming_id, {
-          data: {value: value, authenticity_token: csrf_token()},
+          data: { value: value, authenticity_token: csrf_token() },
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/naming_vote_ajax.js
```diff
@@ -65,8 +65,8 @@ function VoteByAjaxModule(translations) {
               save_vote_buttons().hide();
             } else {
               change_vote_selects().each(function () {
-               _this.val(_this.data("old_value"))
-                       .attr("disabled", null);
+                _this.val(_this.data("old_value"))
+                  .attr("disabled", null);
               });
```
nimmolo, line 70: Sorry for the reformat here, this is just VSCode busybodying.

app/helpers/observations_helper.rb
```diff
@@ -180,6 +180,7 @@ def pct_html(naming)
       content_tag(:button, h(percent),
                   class: "vote-percent btn btn-link",
                   data: { toggle: "modal",
+                          id: naming.id.to_s,
                           target: "#show_votes_#{naming.id}" })
```
nimmolo, line 184: The missing selector

app/views/observations/namings/votes/_table.html.erb
```diff
@@ -36,7 +36,7 @@
               else
                 "..."
               end
-            end.safe_join(",") + ")"
+            end.safe_join(", ") + ")"
           end
```
nimmolo, line 40: Puts a space between user_links of voters

PR #1342, opened 2023-02-09
Change find_query boolean positional arg to kwarg

JoeCohen:
A two-line change to call `find_query` like this:
```ruby
find_query(:LocationDescription, update: false)
```
instead of the puzzling
```ruby
find_query(:LocationDescription, false)
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1098,7 +1098,7 @@ class ApplicationController < ActionController::Base
   end
 
   # Lookup the given kind of Query, returning nil if it no longer exists.
-  def find_query(model = nil, update = !browser.bot?)
+  def find_query(model = nil, update: !browser.bot?)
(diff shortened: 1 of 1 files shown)


JoeCohen:
I might merge this without waiting for approval so that I can use it in [jdc-default_order-184416852](https://github.com/MushroomObserver/mushroom-observer/tree/jdc-default_order-184416852).

JoeCohen:
No rush on this -- I won't be able to use it for my desired purpose.
But it's still a reasonable, if trivial, change. See https://docs.rubocop.org/rubocop/1.44/cops_style.html#styleoptionalbooleanparameter

pellaea:
That's my kind of PR!  Two characters changed?  Looks just fine by me.  Better, wish I did that from the start.  (Maybe kwargs didn't exist yet in the version of ruby we were using at the time?)

PR #1345, opened 2023-02-09
Move naming votes by AJAX into the `Observations::Namings::VotesController`

nimmolo:
This relocates the handling of the naming votes from the `AjaxController` to the `Observations::Namings::VotesController`. This is currently where non-ajax votes are handled (although rarely, if ever, used), so this brings the logic together in one place.

## Please test this manually ##

**Improved functionality:** 
- The entire naming table now updates when you vote, to better reflect the current state of consensus, plus the fact that only one naming can have your `I'd Call It That` vote.

**Coding notes:**
- The refactor uses `respond_to do |format|` to route the request for different formats, and follows the same pattern as the upcoming "Propose Name via AJAX" PR. That means the same controller action can respond to HTML and AJAX requests. (Nathan I believe was the first to use this on MO, in `VisualModelsController#show`)
- JS requests to an action look for a JS template file named after the action, which is `Observations::Namings::VotesController#update`, so the live update (to the show_obs template) is now handled by a new template file, `update.js.erb`.

**Manual tests:** 
- On an observation, try voting on namings, or changing your vote. The entire table should update to reflect your vote and the percentage of other votes, after each vote, without reloading the whole page. You should be able to change a vote more than once this way, or cast multiple votes on other namings. If you're logged in as a major contributor, you should be able to change the consensus vote each time, as if you were Sam Rayburn; the "eyes" should move.
- Try casting a vote with JS turned off in the browser. The whole page will reload of course, but the votes should be updated as usual. (It doesn't break existing functionality).


nimmolo:
Sorry for the noise. I forgot to move the test out of `AjaxControllerTest`. 

This should not interfere with manual testing, though — which is the most important thing i'm requesting on this PR.

I won't merge until I have a new test written and passing.

nimmolo:
Test now moved to `Observations::Namings::VotesControllerTest`.

**Coding note:**
We can test an AJAX request to a regular controller action in a controller test using the usual syntax, with one extra arg:
`get(:update, xhr: true, params)`

PR #1350, opened 2023-02-12
Add observation data to lightbox caption

nimmolo:
Adds basic observation data to the lightbox caption, including an unimplemented capability to produce a link to the "Propose a name" form.

This is part of the process of breaking #1316 into smaller components that MO developers can test separately(!)

## Manual Test ##
Try the lightbox links on the activity log index or the observation index. You should get something like this under the image:

Screen Shot 2023-02-14 at 7 59 17 PM

## Question ##
Is getting this data piling any new includes onto the index query? 
I don't think so — it's the basic, where, when, who — but i'm not sure.


app/helpers/thumbnail_helper.rb
```diff
@@ -59,6 +61,62 @@ def image_link_html(link = "", link_method = :get)
...
+                   locals: { observation: obs_data[:obs] })
+    # render(partial: "observations/namings/form",
+    #        locals: { action: :create, url: url, show_reasons: true })
+  end
+
+  # Disabled because interpolation produces unsafe html
+  # rubocop:disable Style/StringConcatenation
+  def caption_image_links(image_id)
+    orig_url = Image.url(:original, image_id)
+    original_image_link(orig_url) + " | " + image_exif_link(image_id)
```
pellaea, line 88: I think the preference in such cases is safe_concat or something like that?  @JoeCohen, do you know the right syntax?  There's a way to do this that doesn't require disabling a cop.

nimmolo: Fixed it with a safe_join

app/controllers/rss_logs_controller.rb
```diff
@@ -73,7 +73,7 @@ def show_selected_rss_logs(query, args = {})
       glossary_term: :user,
       location: :user,
       name: :user,
-      observation: [:location, :name, :user,
+      observation: [:location, :name, :namings, :user,
                     @user ? { thumb_image: :image_votes } : :thumb_image],
```
nimmolo, line 77: Note the new `includes(:namings)` requested by Bullet here.

app/controllers/observations_controller/index.rb
```diff
@@ -183,7 +183,7 @@ def define_index_args(query, args)
     args = { controller: "/observations",
              action: :index,
              matrix: true,
-             include: [:name, :location, :user, :rss_log,
+             include: [:name, :location, :user, :namings, :rss_log,
                        { thumb_image: :image_votes }] }.merge(args)
```
nimmolo, line 188: Note the new `includes(:namings)` requested by Bullet here.

app/controllers/rss_logs_controller.rb
nimmolo: Ah, ok. This extra `includes(:namings)` significantly slows down the page load; `namings` table lookups are an order of magnitude slower than everything else.
 
I'd like to figure out where i'm loading namings for the lightbox caption info, and hopefully get rid of it, before merging this.

pellaea:
Good catch!  It could be something to do with observer's preferred naming
or something like that(?)

Will all of that info be needed when you do the "needs id" interface?

Another option, by the way, is to add an index to namings.observation_id.
That's on our to-do list.  Sounds like you already have a sorta base-line,
so it could be a good time to try it.

nimmolo:
Found it. It's in the `observations/show/_observation` partial itself. Is there another way to do this, or can we delete this `Textile.register_name(*observation.namings.map(&:name))`? 

It seems to be for presentation of Textile markup that may be in the notes, regarding a naming that is not the consensus name.
```
<div class="obs-notes" id="observation_notes_<%= obs_id %>">
  <%= if observation.notes?
    Textile.clear_textile_cache
    Textile.register_name(*observation.namings.map(&:name))
    Textile.register_name(observation.name)
    content_tag(
      :div, observation.notes_show_formatted.sub(/^\A/, "#{:NOTES.t}:\n").tpl
    )
  end %>
</div><!--.obs-notes-->
```
```
ERROR StudentTest#test_creating_drafts (8.51s)
Minitest::UnexpectedError:         Bullet::Notification::UnoptimizedQueryError: user: vagrant
        GET /
        USE eager loading detected
          Observation => [:namings]
          Add to your query: .includes([:namings])
        Call stack
          /vagrant/mushroom-observer/app/views/observations/show/_observation.html.erb:76:in `map'
          /vagrant/mushroom-observer/app/views/observations/show/_observation.html.erb:76:in `_app_views_observations_show__observation_html_erb__1215123354350024259_54240'
          /vagrant/mushroom-observer/app/helpers/thumbnail_helper.rb:78:in `image_observation_data'
          /vagrant/mushroom-observer/app/helpers/thumbnail_helper.rb:67:in `image_caption_html'
          /vagrant/mushroom-observer/app/views/shared/_image_thumbnail.html.erb:17:in `_app_views_shared__image_thumbnail_html_erb___2159241190805785047_54160'
          /vagrant/mushroom-observer/app/helpers/thumbnail_helper.rb:30:in `thumbnail'
          /vagrant/mushroom-observer/app/presenters/matrix_box_presenter.rb:51:in `rss_log_to_presenter'
          /vagrant/mushroom-observer/app/presenters/matrix_box_presenter.rb:21:in `initialize'
          /vagrant/mushroom-observer/app/views/shared/_matrix_box.html.erb:4:in `new'
          /vagrant/mushroom-observer/app/views/shared/_matrix_box.html.erb:4:in `_app_views_shared__matrix_box_html_erb__1283571071300029790_54060'
          /vagrant/mushroom-observer/app/views/rss_logs/index.html.erb:17:in `block in _app_views_rss_logs_index_html_erb__706207341511921174_54000'
          /vagrant/mushroom-observer/app/helpers/pagination_helper.rb:19:in `paginate_block'
          /vagrant/mushroom-observer/app/views/rss_logs/index.html.erb:15:in `_app_views_rss_logs_index_html_erb__706207341511921174_54000'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:1571:in `show_index_of_objects'
          /vagrant/mushroom-observer/app/controllers/rss_logs_controller.rb:102:in `show_selected_rss_logs'
          /vagrant/mushroom-observer/app/controllers/rss_logs_controller.rb:38:in `index'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:244:in `catch_errors_and_log_request_stats'
          /vagrant/mushroom-observer/test/session_extensions.rb:123:in `get'
          /vagrant/mushroom-observer/test/session_extensions.rb:102:in `process_with_error_checking'
          /vagrant/mushroom-observer/test/session_extensions.rb:125:in `get'
          /vagrant/mushroom-observer/test/integration/rails-dom-testing/student_test.rb:74:in `create_draft'
          /vagrant/mushroom-observer/test/integration/rails-dom-testing/student_test.rb:31:in `test_creating_drafts'
```

pellaea:
Good catch!  Yes, remove that for the purpose of the index at very least.
Would you mind slipping it up to the show_obs controller so that it's not
happening for indexes?

nimmolo:
Will do. I didn't know you could `Textile.register` in the controller. I guess i can also repeat the conditional `if observation.notes?`

Is the situation that you have to register specific strings to get Textile to parse `__**Stuff likethisia**__`, otherwise the punctuation goes straight through unescaped? If so that seems awfully unwieldy, compared to Markdown, doesn't it?

pellaea:
It's only needed to be able to handle abbreviations _S. likethisia_.  It
needs to know what the "S" might stand for.

nimmolo:
Re: add an index to `namings.observation_id` -  I'm guessing that is a migration... is it this:
```ruby
add_index :namings, :observation_id, unique: true
```
and should I do that in a separate PR?

I don't understand what adding an index on a column really does, actually. I notice there's one on `observation_images` - 
```ruby
    t.index ["observation_id"], name: "index_observation_images_on_observation_id"
```
I'm guessing it has obs_id's in one column, and the second column has an array that points to all rows that have that obs_id. The rails_db gem used to allow me to inspect the db, but it's not working for me currently.

nimmolo:
I propose moving discussion of this to the new separate draft PR.

app/controllers/observations_controller/index.rb
nimmolo: Removed in [4880aa6](https://github.com/MushroomObserver/mushroom-observer/pull/1350/commits/4880aa6330fd93872e6df1afcd6fdae168e98772)

app/controllers/observations_controller/show.rb
```diff
@@ -117,4 +118,12 @@ def external_sites_user_can_add_links_to(obs)
...
   end
+
+  # Incurs a costly namings lookup if called in the partial outside show_obs
+  # e.g., in the lightbox caption
+  def register_namings_for_textile_in_notes
+    return unless @observation.notes?
+
+    Textile.register_name(*@observation.namings.map(&:name))
+  end
 end
```
nimmolo, line 129: This avoids the costly extra `naming` includes on the indexes where the captions will appear (in the lightbox).
It only registers the `Textile` naming on the show_obs page.

app/presenters/matrix_box_presenter.rb
```diff
@@ -50,7 +50,8 @@ def rss_log_to_presenter(rss_log, view)
       if target&.respond_to?(:thumb_image) && target&.thumb_image
         view.thumbnail(target.thumb_image,
                        link: { controller: target.show_controller,
-                               action: target.show_action, id: target.id })
+                               action: target.show_action, id: target.id },
+                       obs_data: obs_data_hash(target))
```
nimmolo, line 54: This code allows the matrix box photo itself to serve as an AJAX link, e.g. to the naming form. Defaults to current link type, to minimize refactoring. As yet unimplemented.

app/views/observations/show/_observation.html.erb
```diff
@@ -32,44 +36,43 @@
...
-  <div id="observation_sequences">
+  <div class="obs-sequence" id="observation_sequences_<%= obs_id %>">
     <%= render(partial: "observations/show/sequences",
                 locals: { observation: observation }) %>
   </div>
 <% end %>
 
-<div class="obs-notes" id="observation_notes">
+<div class="obs-notes" id="observation_notes_<%= obs_id %>">
   <%= if observation.notes?
```
nimmolo, line 74: More specific IDs necessary when this partial appears with other versions of the same partial on the page (in lightboxes). The IDs are not unique in that case.

app/views/observations/show/_observation.html.erb
```diff
@@ -32,44 +36,43 @@
...
     Textile.clear_textile_cache
-    Textile.register_name(*observation.namings.map(&:name))
```
nimmolo, line 72: Removed to avoid unnecessary `includes`. Further comment above,.

app/controllers/rss_logs_controller.rb
nimmolo: Resolved by moving a `Textile.register_name` line in the partial. (Note below)

app/helpers/thumbnail_helper.rb
```diff
@@ -71,7 +71,8 @@ def image_caption_html(image_id, obs_data, link_type)
   end
 
   def image_observation_data(html, obs_data, link_type)
-    if link_type == :naming || obs_data[:obs].vote_cache <= 0
+    if link_type == :naming ||
+       (obs_data[:obs].vote_cache.present? && obs_data[:obs].vote_cache <= 0)
       html << caption_propose_naming_link(obs_data)
```
nimmolo, line 76: It may be an artifact of the old improperly stripped db snapshot, but my local install crashed without this.
obs.vote_cache was nil.

pellaea:
I already fixed that same problem somewhere.  There are so many PRs
floating about I can't keep them straight!  Yes, vote-cache can be nil.

nimmolo:
Yeah, let's merge these if they pass. Caption redesign will come later.

PR #1353, opened 2023-02-13
Make community votes table an AJAX popup with live data

nimmolo:
Before, the community vote table was pre-rendered in the page with the data available on page load. Voting via Ajax, which does not reload the page, would not affect the table; it would always display the votes that were originally there on page load.

Now, the vote table is rendered by `VotesController#show`, as it would be for no-js users, and therefore should show live data at the time of the request. `VotesController#show` is refactored to handle both `html` and `js` requests.

**New feature**: The votes table now shows the naming that's being voted on. Confusingly this was missing before.


pellaea:
Sweet.  If it works, it looks great.

JoeCohen:
I'm starting to play with this and noticing 2 things that I'll put in separate comments.

JoeCohen:
When I first visit an Observation, it doesn't show my existing vote. 
Compare https://mushroomobserver.org/441942
Screenshot 2023-02-13 at 6 03 48 PM
with http://localhost:3000/441942
Screenshot 2023-02-13 at 6 04 32 PM
(Of course, you'll only see this on a Obs where you already have a vote.)

nimmolo:
@JoeCohen - Thank you! 
Is that the case also on the live site, for you? 

I noticed that too while i was working on #1354 and that PR should fix it, it should maybe merge first.

JoeCohen:
If I click on one of the `%` links in the Community Vote column, it shows all the voters as `admin`, and does not show the weights or the totals.
Screenshot 2023-02-13 at 7 38 21 PM

JoeCohen:
The top screenshot is from the live site, which does show my votes when I first go to the Obs.

> @JoeCohen - Thank you! Is that the case also on the live site, for you?
> 
> I noticed that too while i was working on #1354 and that PR should fix it, it should maybe merge first.

nimmolo:
@JoeCohen - Hm. I did a double-take because I have experienced that bug on another branch, but not the case for me here. I'm logged in as "nathan".

Untitled

Screen Shot 2023-02-13 at 8 04 16 PM

One hypothesis: I've noticed VSCode has been unreliable switching branches in the past, maybe happening again. (I'm using GitHub Desktop to switch branches here, and it's been good so far.)

nimmolo:
New feature: The votes table now shows the naming that's being voted on. 

Screen Shot 2023-02-13 at 8 45 32 PM


Confusingly this was missing before.

pellaea:
Nice!

nimmolo:
@JoeCohen did this ever clear up for you?

nimmolo:
Actually, now I am definitely getting @JoeCohen 's reported problem of the user votes reported as `admin` in the table.
Screen Shot 2023-02-16 at 3 54 10 PM

When i inspect the `Vote.user_id`, it is `0`, and that's reported as `admin`. I'm working off the fresh `checkpoint_zipped.gz` I just created off the live site, so i can easily find the same observation and compare the vote tables.
Screen Shot 2023-02-16 at 3 54 33 PM

@pellaea I feel like i might need your help debugging this. I can't tell why the vote record lookup would suddenly find a `0` user.

pellaea:
Okay, let me check it out and spin it up and see if I can also reproduce
the bug.  Remind me what the steps are?

pellaea:
It's working fine for me.  Help?

nimmolo:
Yeah it was working for me too, until updating my db today. Maybe try the new snapshot?

I'm on this obs: http://localhost:3000/478351

Checking the db directly, the `Vote` record most definitely shows this. Note the `0` user_id on the second vote:
```
3.1.2 :009 > Vote.where(naming_id: 661059)
  Vote Load (356.0ms)  SELECT `votes`.* FROM `votes` WHERE `votes`.`naming_id` = 661059
 =>                                                                                                                                 
[#<Vote:0x00007f722ebe2d08                                                                                                          
  id: 937316,                                                                                                                       
  created_at: Sun, 14 Nov 2021 00:16:04.000000000 EST -05:00,                                                                       
  updated_at: Sun, 14 Nov 2021 00:16:04.000000000 EST -05:00,                                                                       
  naming_id: 661059,                                                                                                                
  user_id: 667,                                                                                                                     
  observation_id: 478351,                                                                                                           
  favorite: true,                                                                                                                   
  value: 2.0>,                                                                                                                      
 #<Vote:0x00007f722ec5bb18                                  
  id: 997262,                                               
  created_at: Wed, 15 Feb 2023 23:18:39.000000000 EST -05:00,
  updated_at: Wed, 15 Feb 2023 23:18:39.000000000 EST -05:00,
  naming_id: 661059,
  user_id: 0,
  observation_id: 478351,
  favorite: false,
  value: -2.0>] 
3.1.2 :010 > 
```

nimmolo:
By the way the user `bobby_bo` (user `12090`) definitely exists on the new snapshot.

nimmolo:
Ahaa: 
```
User.find(12090).votes_anonymous?
true
```

pellaea:
Oh, hell, that's because you're using a sanitized database where it
anonymizes votes for users who want their votes anonymous!

update votes
inner join users on users.id = votes.user_id
set votes.user_id = 0
where users.votes_anonymous = 1;

pellaea:
Yup, mystery solved! :)

nimmolo:
Whee-ew! 
So there IS no anonymous voting on the live site?! Why does it show up?
It also weights their votes differently on local, because user `0` has different reputation.

pellaea:
No, there is anonymous voting on the site.  When we create a "stripped"
version of the database, in order to maintain that anonymity we need to
anonymize those anonymous votes.  We use the snippet I provided above (from
the end of db/clean.sql) to do so.  It felt better to preserve the vote and
instead to change all the anonymous votes to be owned by admin.  One way or
another, it's imperfect.  This way, since admin has no contribution to
speak of, it would weight the votes differently when owned by admin versus
the original user.  But it's better than deleting the vote altogether, in
my opinion.  But we can't leave the user_id on those votes because then
anyone could slurp in the public copy of the database and query all those
anonymous votes and find out who had cast them.

nimmolo:
Ok. That creates some interesting artifacts:

When I first load an obs that has two anonymous votes, it has a vote percentage, say `48% (2)`.
When I click on the voting table, it seems to "realize" they're anonymous and... 
on page reload, the percentage for the same naming is suddenly `0% (2)`.

nimmolo:
@pellaea I've implemented the whirly on click of the vote percentage. Once this merges, i'll add it to the other links.

nimmolo:
Also added whirly to the `Propose another name` link.

pellaea:
Yeah, I highly recommend actually testing on an unsanitized snapshot for
this reason.  There are actions which will automatically refresh the vote
caches and conensus and everything.  The thing which grabs the vote table
must do this. Makes sense.  It already has all the data it needs, right?

pellaea:
Thank you and thank you!

nimmolo:
Yes. But - note that on the live site, the "anonymous" votes ARE labeled with the actual user, in the vote table. 

ref: https://mushroomobserver.org/478351 — _**Ossicaulis**_ naming
e.g., it seems `bobby_bo` votes anonymous, but his name appears in the vote table on the live site.

pellaea:
Well, that's a bug, then, isn't it?  Alright, I'll look at it.  That has to
be resolved ASAP.  Some people take privacy on-line very seriously.
Imagine that.

pellaea:
Arg, it's more screwed up than that.  users.votes_anonymous == 1 means
*not* anonymous, 2 = all anonymous, 3 = grandfather in.  So the live site
is working.  The clean script is not.

nimmolo:
Well that's a relief. 
I vote to merge this if you're satisfied with it, it will reduce confusion on the next PRs.

I have some questions about passing translations to the new MOEvents, but i'll save them for later.

pellaea:
Fine with me!  I'm creating a new stripped snapshot now.

PR #1354, opened 2023-02-14
Make naming vote selects submit directly to `VotesController#update` without a custom AJAX handler

nimmolo:
This refactors the Naming Votes form to use the same `rails-ujs` js-request handlers that other naming events use, such as the "show community vote table" and the "Propose a name" form. 

There's nothing wrong with our previous `naming_vote_ajax.js` making an AJAX call, but this is less code, and hopefully easier to maintain since it follows the same pattern as the other remote-events things I'm building.

It allows for a much simpler handler for naming vote changes:
- The handler only puts up a modal that notifies the user we are saving the votes. (Vote change can take up to 10 seconds~!) The modal now cannot be dismissed by any keystroke, to avoid multiple concurrent ajax form submissions.
- It doesn't have to downgrade other selects if the user votes `I'd call it that`, because the whole naming table should get refreshed
- I don't believe we need to handle saving a backup of the select "old value" data on the event of unsuccessful AJAX. The selects just won't be changed by a naming table refresh.

## Please test manually ##
Since this is almost purely a js change, it's not covered by integration tests yet. However, it should not break non-js versions of the forms.


nimmolo:
Currently this saves the vote for any vote except "I'd call it that", which saves as "No opinion"

app/assets/javascripts/naming_vote_ajax.js
```diff
@@ -1,86 +1,36 @@
...
-        if (value == "3.0") {
-          change_vote_selects().each(function () {
-            var _this2 = $(this);
-            if (_this2.data("id") != naming_id && _this2.val() == "3.0") {
-              _this2.val("2.0");
-            }
-            _this2.attr("disabled", "disabled");
-          });
-        }
-
```
pellaea, line 34: Any particular reason to do away with this behavior?

nimmolo: Yes - sorry I modified the PR comment after requesting your review, it probably didn't go out in the email.

> - It doesn't have to downgrade other selects if the user votes "I'd call it that", because the whole naming table will get refreshed if the ajax request is successful
> - I don't believe we need to handle saving a backup of the select "old value" data on the event of unsuccessful AJAX. The selects just won't be changed by a naming table refresh.

pellaea: Ah, right.  Okay, then the only difference is that we are no longer disabling the controls while waiting... is that also covered by the new "fire" thingy?

nimmolo: Yes, but it's handled a different way. Now, the "Saving... whirly" modal disables all events on screen.

I was surprised to find that on the current live version, you can dismiss the "saving" modal while you're waiting, but in my book, there's no need to interact with the page at all.

pellaea:
That works for me.  Caveat ... something.  You get what you deserve if you
dismiss the whirly and try to change something that's about to change from
under you!

nimmolo:
"Caveat whirly" is the expression, I believe!

pellaea:
> "Caveat whirly" is the expression, I believe!

LOL!

PR #1355, opened 2023-02-14
Enable "destroy naming" via AJAX

nimmolo:
This makes the `destroy` link in the namings table (only available to the user who proposed the name) able to submit "remotely", i.e. via rails-ujs ajax. It adds a handler for the successful destroy action as a js template, `destroy.js.erb`.

## Please test manually ##


pellaea:
Groovy.  Looks like you got this down! :)

nimmolo:
The whirly has been added to the remote delete button.

nimmolo:
@pellaea Potential issue: For `destroy`, the whirly goes away if you confirm the destroy, after the db action has happend. But it stays on the screen if you cancel at the "Are you sure?" prompt. You have to click outside the whirly to dismiss the whirly.

The choices are: no whirly, or this. 
I don't know how to bind to the JS-generated `confirm` `dismiss` event, it would need to be written into the original handler, in this case extending rails-ujs probably. Not worth it i believe.

pellaea:
Not important, do away with the whirly?

nimmolo:
Ok. Yeah, the confirm dialog is enough feedback to start with, for this one.

pellaea:
(I appreciate you taking the time to try it, though.  I think it would be
the "correct" way to do it.  But since it's not possible, so be it.)

PR #1356, opened 2023-02-14
Enable "edit naming" via AJAX

nimmolo:
Generalizes the code for the "Propose a name" modal form, so it also works for editing a proposed name.

Adds a hidden form and input `ur_name_autocomplete` that initializes a name autocompleter. This can be transferred via jQuery to modal naming forms (that way, they don't have to query `Name.primer` again).

The "Edit" link under a naming (only available to the user who proposed the name) should now open the modal form if JS is on. If not, it loads the form page.

## Please test manually ##
- Propose a name for an observation
- Then click "Edit" under your proposed name, and change the name. Check autocomplete functionality!
- Try something deprecated
- Try a nonexistent name
- Try changing your vote on other namings


app/views/observations/namings/edit.js.erb
```diff
@@ -0,0 +1,38 @@
...
+  var naming_input = $("#naming_name");
+  var naming_form = $("#naming_form");
+  var namings_table = $("#namings_table");
+  var observation_namings = $("#observation_namings");
+  var this_modal = $(this);
+
+  new MOAutocompleter(jQuery.extend({
+    input_elem: naming_input,
+    ajax_url: "/ajax/auto_complete/name/@",
+    collapse: 1
```
pellaea, line 18: So there's a trick you can do with the MO autocompleter that allows you to establish a single name autocompleter that can be used with multiple name fields.  (I developed it for the species list bulk editor which I think is now defunct, but the code should still be sound.)  The concern is that each autocompleter instance loads in *a lot* of names.  Modern browsers use a lot of memory, sure, and can probably handle it... but why not have all the modals share the same autocompleter?

Hmmm.  There it is, not documented, alas.  Check the "reuse" method.  I think I get it.  I think you bind an MOAutocompleter to the first field like usual, then for all subsequent ones you do `$('#first_field').reuse($('#second_field'))` and so on.  This binds a bunch of events on #second_field to switch the input element over to the second one, etc.  It basically rebinds the autocompleter on the fly to whichever field you click into.  It worked well, as I recall.

nimmolo: Thanks! I did notice that, and wanted to use it. 

The trick is... the fields have the same ID. The first form is `remove()`d from the DOM, and then the second form is appended, and... there's nothing to reuse.

So I guess what I should do is this:
- Change the `id` attribute of the _old_ field, to `naming_name_old` or something.
- Hide the old form
- reuse the autocompleter `$('#naming_name').reuse($('#naming_name_old'))`
- _Then,_ remove the old form

Do you reckon that would work?

nimmolo: Oh. My comment related to `form_reload.js.erb` and you were commenting on `edit.js.erb`.

With `edit.js.erb` and `new.js.erb` there's no name autocompleter to reuse (we're on the show_obs page)... is there?

pellaea:
This is awesome.  If you wait a day or two, I can probably test drive it in the dev VM and see if I can break anything.

app/views/observations/namings/edit.js.erb
nimmolo: More background... 

The partial that renders the naming form into the modal already has a `turn_into_name_auto_completer`, and I think that should work, but it does not. (Maybe because the modal is still hidden at this point?):
```ruby
        <%= f_n.label(:name, :WHAT.t + ":") %>
        <%= f_n.text_field(:name, value: what, size: 40, class: "form-control",
                           autocomplete: "off", # turn off browser autocomplete
                           data: { autofocus: focus_on_name }) %>
        <% turn_into_name_auto_completer(:naming_name, primer: Name.primer) %>
```
In any case, i found that it didn't work; I had to manually bind the autocompleter with the response template (`new.js.erb` etc) after the modal is shown.

pellaea: I think there is, even on the show obs page.  Each time you propose a name or edit an existing one, it will popup this naming modal, right?  Unless it's the identical modal in each case.  (Sorry, I could check that myself if I were more patient.)  But certainly on the index this will become an issue.

How about this solution?  Create a hidden field, and bind a name autocompleter to it!  Then "reuse" it for all name fields, wherever they appear, whether in the initial page load or in the results of any ajax request.  (Unfortunately the way the autocompleter I adapted worked, it *had to* have a jquery element to work off of.  It's too much work to change that to work off of an "autocompleter" instance instead.)

pellaea: re: why turn_into_autocompleter doesn't work in ajax response -- I believe it's doing an "inject javascript at end" thingy.  That will never happen in an ajax response!

nimmolo: Right. Copying from a hidden one already in the page is a nice idea, i like it. 

Will add that field to observations index and show and adjust this. Is the correct syntax:
```js
AUTOCOMPLETERS[jQuery('##{first_id}').data('uuid')].reuse('#{new_id}')
```
?

pellaea: I don't think you need to mess with AUTOCOMPLETERS.  I think just `$('##{first_id}').reuse('#{new_id}')`.  But I'd have to test it, too! :)

PR #1359, opened 2023-02-15
Add index to `namings` table on `observation_id`

nimmolo:
Adds an index to the namings table to speed up observation namings queries, which are so slow.

Interestingly, this index key `observation_id` cannot be unique ~yet, although it seems like it should be, because we have duplicates in the db presently. (At least one, `140296`, was raised by my first migration attempt).~

I don't know if we have duplicates by design, ~but they seem like they could have been caused by namings proposed on an obs at the same time.~   EDIT: Apparently we do. 

~I found [an article about it](https://dev.to/nodefiend/rails-migration-adding-a-unique-index-and-deleting-duplicates-5cde), with suggestions for cleanup. If this is the situation, it seems like we should maybe merge, rather than delete, the duplicate data though.~

### DIFF
--- a/db/migrate/20230215221813_add_index_to_namings_observation_id.rb
+++ b/db/migrate/20230215221813_add_index_to_namings_observation_id.rb
@@ -1,5 +1,5 @@
 class AddIndexToNamingsObservationId < ActiveRecord::Migration[6.1]
   def change
-    add_index :namings, :observation_id, unique: true
+    add_index :namings, :observation_id
   end
 end
--- a/db/migrate/20230215221813_add_index_to_namings_observation_id.rb
+++ b/db/migrate/20230215221813_add_index_to_namings_observation_id.rb
@@ -1,0 +1,5 @@
+class AddIndexToNamingsObservationId < ActiveRecord::Migration[6.1]
+  def change
+    add_index :namings, :observation_id, unique: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_12_26_160800) do
+ActiveRecord::Schema.define(version: 2023_02_15_221813) do
 
   create_table "api_keys", id: :integer, charset: "utf8mb3", force: :cascade do |t|
     t.datetime "created_at"
@@ -431,6 +431,7 @@ ActiveRecord::Schema.define(version: 2022_12_26_160800) do
     t.integer "user_id"
     t.float "vote_cache", default: 0.0
     t.text "reasons"
+    t.index ["observation_id"], name: "index_namings_on_observation_id"
   end
 
   create_table "observation_collection_numbers", charset: "utf8mb3", force: :cascade do |t|


pellaea:
There should definitely be multiple namings per observation.  Does that
mean you can't put an index on it??  That doesn't make sense.  We have
indexes on observation_images.image_id and observation_id, also neither of
which is unique.

nimmolo:
Ok - that's just me not understanding what the index does. Deleted `unique: true`.

Separate question. Recently all my migrations wind up with dozens of our obsolete tables re-added to the schema.
- We renamed tables like `collection_numbers_observations` `observation_collection_numbers` a while ago

Do i need to delete and re-do all migrations in the db? I feel like I've done that, recently, for this exact same reason, and i'm back where i started... Is it maybe because the recent snapshot I installed is out of sync with the schema? Seems doubtful.

pellaea:
> Separate question. Recently all my migrations wind up with dozens of our obsolete tables re-added to the schema.

My understanding is this happens in two cases:

1. the snapshot itself has two copies
2. you rebuild the dev db from a snapshot without first dropping the existing db

The latter usually is fine, but if you have old unused tables in your db, then installing a new one over top of it will leave the old unused tables in there.  You have to drop the entire database first, then restore the new snapshot.

It's easy to check the first case: just open the snapshot and see what tables it's creating. (unzip first, of course)

nimmolo:
Got it. Yes, the checkpoint_stripped has the outdated tables. 
Can I generate a checkpoint? I'll chuck this one.

--
-- Table structure for table `collection_numbers_observations`
--

DROP TABLE IF EXISTS `collection_numbers_observations`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!50503 SET character_set_client = utf8mb4 */;
CREATE TABLE `collection_numbers_observations` (

pellaea:
Yes, actually, it would be wonderful if you would generate a new one.  You
might know that it is backing up the database nightly now.  Install that on
your dev VM and then run clean.sql:

```
/vagrant/mo> scp
***@***.***:/data/images/mo/backup/db-YYYYMMDD.gz
snapshot.gz
/vagrant/mo> mysql -u mo -pmo mo_development -e "drop database
mo_development"
/vagrant/mo> gunzip -c snapshot.gz | mysql -u mo -pmo mo_development
/vagrant/mo> mysql -u mo -pmo mo_development -e "source db/clean.sql"
/vagrant/mo> mysqldump -u mo -pmo mo_development | gzip -c - >
checkpoint_stripped.gz
/vagrant/mo> scp snapshot_stripped.gz ***@***.***:
/data/images/mo/
```

Does all that make sense?  (I just typed it out off the top of my head, so
there could be typos as always.)

nimmolo:
Ok. Thanks for sending.
 
But it seems like the new snapshot will also have the outdated tables.
I suspect the production db has them!

pellaea:
Which tables are extra?  The "drop database mo_development" line should fix
the problem.  If the production db has extra tables, I'll fix it now and
generate a new schema.db.

pellaea:
```

Tables_in_mo_production
api_keys
ar_internal_metadata
articles
collection_numbers
comments
copyright_changes
donations
external_links
external_sites
glossary_term_images
glossary_terms
glossary_terms_versions
herbaria
herbarium_curators
herbarium_records
image_votes
images
interests
languages
licenses
location_description_admins
location_description_authors
location_description_editors
location_description_readers
location_description_writers
location_descriptions
location_descriptions_versions
locations
locations_versions
name_description_admins
name_description_authors
name_description_editors
name_description_readers
name_description_writers
name_descriptions
name_descriptions_versions
name_trackers
names
names_versions
naming_reasons
namings
observation_collection_numbers
observation_herbarium_records
observation_images
observation_views
observations
project_images
project_observations
project_species_lists
projects
publications
query_records
queued_email_integers
queued_email_notes
queued_email_strings
queued_emails
rss_logs
schema_migrations
sequences
species_list_observations
species_lists
synonyms
translation_strings
translation_strings_versions
triples
user_group_users
user_groups
users
visual_group_images
visual_groups
visual_models
votes
```

nimmolo:
None of those are extra. But the last snapshot had at least 11 of them, starting with `collection_numbers_observations`.

Is there anything to fear from a possible typo? 
It seems right to me. I will wait till after dark to run it.

nimmolo:
> You might know that it is backing up the database nightly now. Install that on your dev VM and then run clean.sql

I don't understand what are you referring to installing there, above...

pellaea:
The backup is at images.mushroomobserver.org:
/data/images/mo/backup/db-YYYYMMDD.gz.

"Install" that by downloading it, unzipping it and loading it into your
database.  That's what the instructions I just wrote up do.  You do not, of
course, even need to strip (clean.sql) it for use in your dev VM.  It's not
like your dev VM can send email to people or anything!  But it would be
good to keep that stripped, "sanitized" snapshot up to date, too.

The nightly backups should be identical to the production database --
that's sorta the point! :)  So they should not have any extra tables,
either.  I don't have a copy of the stripped version anymore.  Let's assume
I forgot to drop my database before loading the new one when I last updated
the stripped snapshot.

nimmolo:
I get it now, I think. So the checkpoint_stripped.gz is always generated locally.

It’s not done on the production server — I always thought it was! 

You’ve been generating it locally from a backup that you downloaded and installed into your db.

pellaea:
Oops!  I just moved the backups one directory up to
...:/data/images/backup  They were publically visible where they were!

pellaea:
Right, because the sanitizer is extremely destructive.  It has to be done
somewhere safe.

nimmolo:
Should this merge now too? Seems very simple...

pellaea:
Yes, we can use Alan's stats to monitor what effect it has.  I like that
this PR just adds one index.  There are several that NewRelic recommends,
but this was the top one.  Let's wait a week at least before trying the
next one(s).  I'll merge and deploy now.

What about the other recent UI branches?  Haven't they all been tested now?

nimmolo:
They've been tested by me, and I think you. I'd like to merge them!

nimmolo:
Except for the draft big one "Observations Needing Identificaton" UI

pellaea:
Right, is that ready for testing?  I am eager to see it in action!

nimmolo:
It is, yea!

pellaea:
Excellent, soon as I'm done with this thing I'm doing now...

nimmolo:
Is it okay if I add snapshot.gz and checkpoint_stripped.gz to gitignore, so i dont' have to worry about them in the mo folder?

pellaea:
Yes, that's fine.  Actually, checkpoint.gz is already in db/.gitignore.
How about putting them both in db, then?

PR #1362, opened 2023-02-17
Implement User pref `image_size` for lightbox

nimmolo:
Uses the old user `image_size` pref to choose which resolution of image pops up in the lightbox. (Changes the English translation string to make the "lightbox" part clear.)  Also does a migration to set everybody's pref at `huge` which is the size image served by the lightbox the last few years. 

Choices are:
```
Medium
Large
Huge
Full Size
```

If certain lichenologists 👀 select `full_size` for example, they'll get all the pixels and won't have to wait for the large image to load, then click through to the `Show original image` link.

Adds tests for the new pref: controller and integration.

### DIFF
--- a/app/views/account/preferences/_appearance.html.erb
+++ b/app/views/account/preferences/_appearance.html.erb
@@ -60,7 +60,9 @@
   <%= f.select(:image_size,
                options_for_select(
                  User.image_sizes.filter_map { |key, value|
-                   [key.titleize, User.image_sizes.key(value)] if (value >= 3)
+                   if value > User.image_sizes[:small]
+                     ["image_show_#{key}".to_sym.l, key]
+                   end
                  }, @user.image_size
                ),
                {}, { class: "form-control" }) %>
--- a/app/views/account/preferences/_appearance.html.erb
+++ b/app/views/account/preferences/_appearance.html.erb
@@ -55,4 +55,15 @@
   <%= f.number_field(:layout_count, class: "form-control", min: 1) %>
 </div>
 
+<div class="form-group mt-3 form-inline">
+  <%= f.label(:image_size, :prefs_image_size.t) %>
+  <%= f.select(:image_size,
+               options_for_select(
+                 User.image_sizes.filter_map { |key, value|
+                   [key.titleize, User.image_sizes.key(value)] if (value >= 3)
+                 }, @user.image_size
+               ),
+               {}, { class: "form-control" }) %>
+</div>
+
 <%= f.submit(:SAVE_EDITS.l, class: "btn btn-default center-block mt-3") %>
--- a/app/views/shared/_image_thumbnail.html.erb
+++ b/app/views/shared/_image_thumbnail.html.erb
@@ -14,6 +14,7 @@
   # Lightbox stuff. obs_data comes from the matrix_box presenter,
   # link_type from view (for naming links)
   lightbox_id = is_set ? "observation-set" : SecureRandom.uuid
+  lightbox_url = (User.image_size == "full_size") ? full_url : huge_url
   caption = image_caption_html(image_id, obs_data, link_type)
 
   # Show original name?
@@ -33,7 +34,7 @@
                       data: { toggle: "tooltip", placement: "bottom" }) %>
         <%= image_link %>
       </div>
-      <%= link_to("", huge_url,
+      <%= link_to("", lightbox_url,
                   class: "glyphicon glyphicon-fullscreen theater-btn",
                   data: { lightbox: lightbox_id, title: caption }) %>
       <div class="mt-3">
--- a/app/views/shared/_image_thumbnail.html.erb
+++ b/app/views/shared/_image_thumbnail.html.erb
@@ -14,7 +14,7 @@
   # Lightbox stuff. obs_data comes from the matrix_box presenter,
   # link_type from view (for naming links)
   lightbox_id = is_set ? "observation-set" : SecureRandom.uuid
-  lightbox_url = (User.image_size == "full_size") ? full_url : huge_url
+  lightbox_url = (User.current&.image_size == "full_size") ? full_url : huge_url
   caption = image_caption_html(image_id, obs_data, link_type)
 
   # Show original name?
--- a/app/views/shared/_image_thumbnail.html.erb
+++ b/app/views/shared/_image_thumbnail.html.erb
@@ -14,7 +14,13 @@
   # Lightbox stuff. obs_data comes from the matrix_box presenter,
   # link_type from view (for naming links)
   lightbox_id = is_set ? "observation-set" : SecureRandom.uuid
-  lightbox_url = (User.current&.image_size == "full_size") ? full_url : huge_url
+  p_map = {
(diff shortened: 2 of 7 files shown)


app/views/shared/_image_thumbnail.html.erb
```diff
@@ -14,6 +14,7 @@
   # Lightbox stuff. obs_data comes from the matrix_box presenter,
   # link_type from view (for naming links)
   lightbox_id = is_set ? "observation-set" : SecureRandom.uuid
+  lightbox_url = (User.current&.image_size == "full_size") ? full_url : huge_url
   caption = image_caption_html(image_id, obs_data, link_type)
```
nimmolo, line 18: This just falls back to "huge", where it's been. But i'd like to take anything from "medium" up. 
The problem is many users have a `null` size.
I could use a frozen array mapping values to urls, but is that the elegant move?
```ruby
{ "medium" => medium_url, "large" => large_url, "huge" => huge_url, "full_size" => full_url  }
```

app/views/account/preferences/_appearance.html.erb
```diff
@@ -55,4 +55,15 @@
   <%= f.number_field(:layout_count, class: "form-control", min: 1) %>
 </div>
 
+<div class="form-group mt-3 form-inline">
+  <%= f.label(:image_size, :prefs_image_size.t) %>
+  <%= f.select(:image_size,
+               options_for_select(
+                 User.image_sizes.filter_map { |key, value|
+                   [key.titleize, User.image_sizes.key(value)] if (value >= 3)
```
pellaea, line 63: Maybe I'm misunderstanding, but it looks like this is not translating anything?  It looks like it is printing the internal key representation in the dropdown.  That really needs to be protected, in case for any reason we have to change the data representation.  Also can we do something other than "3"? Lastly, isn't User.image_sizes.key(value) just == key?

```
["image_show_#{key}".to_sym.l, key] if value > User.image_sizes[:small]
```

nimmolo: oh yeah. Forgot to translate this...

app/views/shared/_image_thumbnail.html.erb
pellaea: I mentioned this earlier, we really need to do a migration with this PR that resets everyone's preference to "huge".  That's the setting everyone's been stuck on for the past few years, regardless of what their (unused) preference claims.  We should make it so that it works the same after we deploy this PR, until they choose a different size preference.

app/views/account/preferences/_appearance.html.erb
nimmolo: Yeah that last one was absurd. I've got a cold

app/views/shared/_image_thumbnail.html.erb
nimmolo: The migration sounds good to me. 
I'd like to make medium and large a possibility though. Is that hash/array the way?

nimmolo: Is it just 
```ruby
  def change
    User.update_all(image_size: "huge")
  end
```

app/views/shared/_image_thumbnail.html.erb
```diff
@@ -14,6 +14,7 @@
   # Lightbox stuff. obs_data comes from the matrix_box presenter,
   # link_type from view (for naming links)
   lightbox_id = is_set ? "observation-set" : SecureRandom.uuid
+  lightbox_url = (User.current&.image_size == "full_size") ? full_url : huge_url
```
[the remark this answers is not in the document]
nimmolo: Had a comment somewhere about this. Updating to reflect the pref.

app/views/shared/_image_thumbnail.html.erb
nimmolo: Note that in the fixtures, `rolf` has his pref set to `medium`, which is also the default.
I could update the fixture, but it's also fine as is... I do test changing it to `full_size`.

In case js integration tests ever load real images, `medium` seems definitely preferable for testing.

app/views/account/preferences/_appearance.html.erb
nimmolo: Thanks for that rewrite.

app/views/shared/_image_thumbnail.html.erb
pellaea: > The migration sounds good to me. I'd like to make medium and large a possibility though. Is that hash/array the way?

I answered this earlier, but it must've been missed.  Why not `Image.url(@user.image_size, image_id)`.

> `User.update_all(image_size: "huge")`

Yeah, that sounds right.

nimmolo: You sound like you know the code.

LOL, I was really tired yesterday, huh? That solutions was in the same file, four lines away, wasn't it. 

So - If we're going to call @user.image_size, I think that db column needs a default in the migration, too. My server crashed when not finding an image_size.

nimmolo: Actually: Will your suggestion incur an extra lookup of a value we already have? Or would it be cached?
I guess the advantage of my verbose map is that it uses the urls we've already pulled, above.

nimmolo: Ah, ok. `Image.url` is just a path builder, but it does check `source_exists?`, which must have some cost.

pellaea:
Sorry, yeah, I could tell you were a bit off yesterday.  I just got over a
cold, too, I sympathize.  I was also really frustrated with other people I
was working with last night, so I had more of an edge than I would have
liked.  Sorry about that, not your falt.

That whole image thumbnail helper thing is weird.  Did you notice that
nothing actually uses most of those pre-calculated urls?  But you've
already done enough for this PR, none of this is a big deal, most likely.
The thing I would be most concerned with is more database hits.
User.current is probably better than @user, I was just being lazy, but I
also suppose it doesn't make any difference(?)  You definitely want to
check if image_size pref is set and provide a default, again, I was just
throwing out shorthand out of laziness, assuming you would deal with the
details like supplying defaults, etc.

nimmolo:
No problem. I was also on a separate work deadline last night, so i shouldn't have been doing this - at all!

Oh yeahhhh. The other URLs are me starting to build `srcset` attributes for the images... it saves a lot on load time, because it background-loads only as high res of an image as you have pixels to show, at your current browser zoom size. The tricky part is that the `srcset` attribute has to tell the browser exactly which `src` at what screen dimension, which seems silly... the browser should calculate it on the image's own container, which only it knows! And not expect the programmer to supply it as an attribute.

I think that spec is evolving in that direction, but i'll have to read up again. And i will comment them out here.

Then let's merge this thing! 🙌

nimmolo:
Oh no comment-out necessary, it doesn't load the small one. It's just Med-Full. Merging!

nimmolo:
I wonder if any lichenologists will use this, or if they're all far enough off the grid that they only use MO via the API!

pellaea:
I'll let you know... ;)

PR #1364, opened 2023-02-18
Enable comments via AJAX

nimmolo:
Just what it says. Works!

### DIFF
--- a/app/assets/javascripts/mo_events.js
+++ b/app/assets/javascripts/mo_events.js
@@ -7,14 +7,14 @@ MOEvents.alert = function (element) {
 }
 
 MOEvents.whirly = function () {
-  $("#naming_ajax_progress").modal('show');
+  $("#mo_ajax_progress").modal('show');
 }
 
 MOEvents.savingWhirly = function () {
-  // $('#naming_ajax_progress_caption').empty().append(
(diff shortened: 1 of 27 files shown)


pellaea:
I test drove it and couldn't break anything (and I really tried hard).  The code changes all looks reasonable, too.  You seem to really have this down, Nimmo.  Well done!

And it's really slick.  This is something I've wanted for a long time.  I hope all of these modals will save lots of unnecessary show_observation page loads.  And the modal requests surely must be snappier even on the production server.  (They're super snappy on my nice new laptop!)

Awesome all around.  What's next??

nimmolo:
Thanks. 
What's next? I'm thinking of rewriting the Vagrantfile... in Commodore BASIC.

pellaea:
Then schawarma?

nimmolo:
Sounds ideal! 
Somebody said there's a legit Döner place around here somewhere, but i don't remember where.

nimmolo:
Thank you for testing it! 

Now i gotta write tests for the dang Obs Needing ID's so we can unleash Alan on it. 

If you haven't tried the "Help Identify" index in the last day or two, I added some bells and whistles to the matrix boxes, so you can mark a bunch of obs as skippable before even opening the lightbox. You can also vote on unconfident IDs.

PR #1369, opened 2023-02-20
Automatically add a tag whenever a deploy gets done

mo-nathan:


pellaea:
Cool.  How do I view tags and return to a tag?  I've never used tags before.  I presume this is a tool to help with reverting deployments?

JoeCohen (approved):

mo-nathan:
Tags are very much like branches (which are sometimes called "branch
tags"), only simpler.  A tag names a specific commit so you can find it
again.  Branch tags are just names that move when you push new commits to
that branch.  You can see what tags exist with git tag.  You can move a
local repo to a tag with git checkout <tag> (just like with branch tags).
However, it will give you a warning about being in "detached HEAD state"
which essentially means that adding commits will not update any branch
tag.  You can create a new branch after checking out a tag by using git
switch -c <new-branch-name> (as the warning tells you).  Note that switch is
really just another name for checkout so you can also use git checkout -b
<new-branch-name>.

Tags can be used to conceptually "revert" deployments.  You probably
already know this, but this is not at all what git revert is used for and
personally I never use git revert.  Technically what most people want when
they want to revert a deployment is to get the main branch back to having
the same code as it did before some deployment happened.  In my opinion the
simplest, safest way to do that is to "fail forward", meaning add new
commits that make the main branch have the same code as it had before.
Assuming no new files have been added or removed, this can be done with:

git checkout <tag>
tar cf tmp.tar `git diff --name-only main`
git checkout main
tar xf tmp.tar
git commit -a -m 'Revert back to the code for <tag>'
git push

If files have been deleted, then they should get added to tmp.tar while on
the tag.  If files have been added, then they will need to be explicitly
removed from main using git rm.  This will allow everyone else's local
repos to get the changes by just doing a normal git pull and the GitHub
repos will not need anything tweaking.

There are more complicated ways to do reverts, but I don't see any reason
why we would need to do that.

The main practical use for tags is to compare the current code with the
some previous release just using git diff <tag>.  Occasionally, I'll use a
tag to run a previous version of the code so I can figure out if a bug
existed before the latest deploy or not.  This can be difficult if there
has been a lot of change to the database schema or the Gemfile.  I also use
tags with git bisect to figure out when a specific change was introduced.

Hope that helps!

JoeCohen:
Nathan
Thanks for doing this.
Is there a way to do automatic (or semi-automatic) versioning?
- Joe

mo-nathan:
There are definitely ways to do semantic versioning if that's what you mean, but it requires more work both to set it up and to use it.  What I'd do is have the deploy script automatically update a version file and commit it before the tag gets applied. However, it's a fair amount of mechanism and if it fails you end up with a bit of a mess.  It also means humans have to decide for each deploy whether it's a major vs. minor vs. patch and have whoever is doing the deploy provide that info in someway.  This is cognitive overhead that I don't think we want.  That's why I went with simple timestamp based versioning since it's fully automatic, way better than what we have at the moment, and in my view fits with how we like to do things.  The timestamp also has obvious meaning.  Happy to discuss further if you think it's warranted.

JoeCohen:
No need to discuss further.
I just wanted to know if it could be done easily.

pellaea:
Thanks for the great overview, Nathan!  It all sounds perfectly fine to me.

I will have to look up `git bisect`, another command I've never heard of.

Also -- and I don't mean to argue anything here, just conversational --
I've done the equivalent of `git checkout $previous-tag` to get production
up and running again.  That buys time to figure out what happened and fix
it, then I just `git checkout main` and deploy as per usual.  That avoids
all the rigmarole around taring changes, etc.   But that's only acceptable,
presumably, if the bug is anticipated to be easy to find and fix so that
production stays in "detached head" mode for only a short amount of time.
And I would communicate that clearly to the team while it was in effect.
Importantly: these new tags will make this workflow *much* easier, too.  I
always struggled to identify which commit was the one I wanted in the
output of `git log`.

mo-nathan:
Yeah, that's a totally legit short term use of tags.  I was really talking about reverting code within the code base vs. getting production working if it's broken.  Tags are great for just getting back to where things were in production.

PR #1370, opened 2023-02-20
`herbaria_helper.rb` SQL -> AR/Arel

nimmolo:
Existing method:
```sql
SELECT u.name, u.login, COUNT(u.id)
FROM herbarium_records hr JOIN users u ON u.id = hr.user_id
WHERE hr.herbarium_id = #{herbarium.id}
GROUP BY u.id ORDER BY COUNT(u.id) DESC LIMIT 5
```

This rewrite produces the SQL:

```sql
SELECT `users`.`name`, `users`.`login`, COUNT(`users`.`id`) 
FROM `users` INNER JOIN `herbarium_records` ON `herbarium_records`.`user_id` = `users`.`id` 
WHERE `herbarium_records`.`herbarium_id` = #{herbarium.id} 
GROUP BY `users`.`id` ORDER BY COUNT(`users`.`id`) DESC LIMIT 5
```

Does this seem correct?

### DIFF
--- a/app/helpers/herbaria_helper.rb
+++ b/app/helpers/herbaria_helper.rb
@@ -4,8 +4,8 @@
 module HerbariaHelper
   def herbarium_top_users(herbarium_id)
     User.joins(:herbarium_records).
-      where(HerbariumRecord[:herbarium_id].eq(herbarium_id)).
-      select(User[:name], User[:login], User[:id].count).
-      group(User[:id]).order(User[:id].count.desc).take(5)
+      where(herbarium_records: { herbarium_id: herbarium_id }).
+      select(:name, :login, User[:id].count).
+      group(:id).order(User[:id].count.desc).take(5)
   end
 end
--- a/app/helpers/herbaria_helper.rb
+++ b/app/helpers/herbaria_helper.rb
@@ -6,6 +6,6 @@ module HerbariaHelper
     User.joins(:herbarium_records).
       where(HerbariumRecord[:id].eq(herbarium.id)).
       select(User[:name], User[:login], User[:id].count).
-      group(User[:id]).order("COUNT(`users`.`id`) DESC").take(5)
+      group(User[:id]).order(User[:id].count.desc).take(5)
   end
 end
--- a/app/helpers/herbaria_helper.rb
+++ b/app/helpers/herbaria_helper.rb
@@ -4,7 +4,6 @@
 module HerbariaHelper
   def herbarium_top_users(herbarium)
     User.joins(:herbarium_records).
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):

pellaea:
No, that's not quite the same.  The third line needs to be `herbarium_records.herbarium_id` not `herbarium_records.id` to be identical:

```
WHERE `herbarium_records`.`herbarium_id` = #{herbarium.id} 
```

app/helpers/herbaria_helper.rb
```diff
@@ -3,11 +3,9 @@
...
 module HerbariaHelper
   def herbarium_top_users(herbarium)
-    Herbarium.connection.select_rows(%(
-      SELECT u.name, u.login, COUNT(u.id)
-      FROM herbarium_records hr JOIN users u ON u.id = hr.user_id
-      WHERE hr.herbarium_id = #{herbarium.id}
-      GROUP BY u.id ORDER BY COUNT(u.id) DESC LIMIT 5
-    ))
+    User.joins(:herbarium_records).
+      where(HerbariumRecord[:id].eq(herbarium.id)).
```
pellaea, line 7: Yep, this line is off.  It might be worth checking to see why no tests catch this.  This should totally be unit-testable.

nimmolo: Good catch, thanks. Added tests

app/helpers/herbaria_helper.rb
```diff
@@ -2,12 +2,10 @@
...
-      SELECT u.name, u.login, COUNT(u.id)
-      FROM herbarium_records hr JOIN users u ON u.id = hr.user_id
-      WHERE hr.herbarium_id = #{herbarium.id}
-      GROUP BY u.id ORDER BY COUNT(u.id) DESC LIMIT 5
-    ))
+  def herbarium_top_users(herbarium_id)
+    User.joins(:herbarium_records).
+      where(HerbariumRecord[:herbarium_id].eq(herbarium_id)).
+      select(User[:name], User[:login], User[:id].count).
+      group(User[:id]).order(User[:id].count.desc).take(5)
```
JoeCohen, line 9: If you felt like it, you could eliminate some of the Arel.
But it may read better with all Arel.
```ruby
    User.joins(:herbarium_records).
      where(herbarium_records: { herbarium_id: herbarium_id }).
      select(:name, :login, User[:id].count).
      group(:id).order(User[:id].count.desc).take(5)
```

nimmolo: I like it. Committed it, thanks!

PR #1372, opened 2023-02-21
Enable direct filtering of obs needing ids index

nimmolo:
Adds a hidden field `params[:needs_id]` to the site search form if the form is rendered on the Obs Needing Ids index, only.

Passes `params[:needs_id]` thru to the `ObservationsController#index`, which will save the search as the new query param, and then render the results via `Observations::IdentifyController#index` if the param is present.

The alternative to this is to make a whole 'nother `type` of `PatternSearch`, and i don't believe that's necessary.

### DIFF
--- a/app/controllers/observations_controller/index.rb
+++ b/app/controllers/observations_controller/index.rb
@@ -114,12 +114,22 @@ module ObservationsController::Index
       search.errors.each do |error|
         flash_error(error.to_s)
       end
-      render("index", location: observations_path)
+      if params[:needs_id]
+        redirect_to({ controller: "/observations/identify", action: :index,
+                      q: get_query_param })
+      else
+        render("index", location: observations_path)
+      end
     else
       @suggest_alternate_spellings = search.query.params[:pattern]
-      show_selected_observations(
-        search.query, no_hits_title: :title_for_observation_search.t
-      )
+      if params[:needs_id]
+        redirect_to({ controller: "/observations/identify", action: :index,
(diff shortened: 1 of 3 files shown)


app/controllers/observations_controller/index.rb
```diff
@@ -114,12 +114,22 @@ def render_observation_search_results(pattern)
...
+      end
     else
       @suggest_alternate_spellings = search.query.params[:pattern]
-      show_selected_observations(
-        search.query, no_hits_title: :title_for_observation_search.t
-      )
+      if params[:needs_id]
+        redirect_to({ controller: "/observations/identify", action: :index,
+                      q: get_query_param })
+      else
```
nimmolo, line 128: @pellaea I'm having an issue here. 

Calling the identify index with `q: get_query_param` doesn't pull the query we just made, so it calls the identify index without the recent search query.

At the top of this method:
```ruby
  def render_observation_search_results(pattern)
    search = PatternSearch::Observation.new(pattern)
```
Does instantiating `PatternSearch::Observation.new(pattern)` save the query record? Or not yet?
I think i need to save the record before redirecting.

nimmolo: Ah. `show_selected_observations` does `store_query_in_session`. That's what i need to do here.

nimmolo: Actually i can do the same thing via `get_query_param(search.query)` - it also saves it.

PR #1375, opened 2023-02-24
Fix thumbnail links on obs index

nimmolo:
Sets local defaults at `matrix_box.html.erb` for link_type, link_method and identify, called in presenter and thumbnail. 

Please test before deploying if possible. Works for me locally.


pellaea:
Will do.  I can barely get this page to load right now.  I'll do it later tonight.

PR #1379, opened 2023-02-28
Add lazy-loading to thumbnails, vote on hover

nimmolo:
#### For js users ####
Adds a script, along with the markup to implement it, that 
- loads a placeholder instead of an ``
- calculates the image proportions for the placeholder, to eliminate resizing layout shift
- loads the `` when the image is within the browser viewport
- can handle `srcset` and `sizes` for future implementation
- can utilize browser native lazyloading when that becomes available

#### For non-js users ####
- loads the regular image within a `<noscript>` tag, right on top of the placeholder.

#### For both js and non-js #### 
 - Only on matrix_boxes, moves the vote interface over the image, on hover. 
 - JS is not required to access the vote UI — it's pure CSS

Screen Shot 2023-02-28 at 3 51 22 PM


nimmolo:
This slight change in the thumbnail UI is ready for test drive. I've updated the commit notes above.
I don't expect a code review, but it's welcome if anyone wants to. 
It's a bit of JS and CSS, and a bit of refactoring.

PR #1380, opened 2023-02-28
Refactor `thumbnail_helper` to use a new `thumbnail_presenter`, disentangle from `matrix_box_presenter`

nimmolo:
MO's existing presenter is not the easiest part of our code to understand, but it's kinda indispensable because of our multifarious `RssLog` template partials. `MatrixBoxPresenter` is essentially a service object that determines what data to pull for the matrix_box, (e.g. the title, and other attributes). It also figures out what image relates to the object at hand, and thanks to me, cooks up an increasingly complex set of nested elements for the thumbnail, depending on context. 

With more planned changes to `thumbnail`, it's gotten too confusing because the thumbnail args are passed around so many times. This PR attempts to simplify the situation by separating the concerns of **matrix box** data and **thumbnail** data.

### Details ###
In the current state of MO, there are a lot of local args being passed through the `matrix_box` template that are destined for the increasingly complex `image_thumbnail` template. Basic confusion arises from the routing of these passed arguments. 

Usually, these image presentation args are passed something like this:
```
rss_logs/index >  
    # An index or show template calls the matrix box partial, with locals. 
    # About half of the locals are for the matrix_box, and half are for the thumbnail.
shared/matrix_box.html.erb >  
    # instantiates the presenter with an object, and contextual args
MatrixBoxPresenter >  
    # object determines what image to show, and what kind of caption
ThumbnailHelper#thumbnail >  
    # creates some complex nested HTML, which it passes as locals to the 
    # thumbnail template
shared/image_thumbnail.html.erb  
    # draws the thumbnail, its links, lightbox caption, data...
```
It's too many layers, especially when we need to add new args to the image template. How to pass them; where to change them contextually?

This PR
- removes the handling of thumbnail args from `MatrixBoxPresenter`
- creates a new `ThumbnailPresenter`, to prepare the data for the image template separately. It removes the image args being passed to the `MatrixBoxPresenter`, and passes them directly on to the `ThumbnailPresenter` as a subset of the `local_assigns`. It moves all of the helper methods that don't need to be globally accessible from `ThumbnailHelper` into the `ThumbnailPresenter`.
- creates a new base class for presenters, to facilitate calling helper methods

I think everyone will find it's more comprehensible via reading through the code itself.

### DIFF
--- a/app/assets/stylesheets/mo/_links_buttons_alerts.scss
+++ b/app/assets/stylesheets/mo/_links_buttons_alerts.scss
@@ -2,7 +2,7 @@
 // link, button, alert themed styles
 // --------------------------------------------------
 
-a:not(.btn):not(.list-group-item):visited {
+a:not(.btn):not(.list-group-item):not(.theater-btn):visited {
   color: $LINK_VISITED_FG_COLOR;
 }
 
--- a/app/helpers/thumbnail_helper.rb
+++ b/app/helpers/thumbnail_helper.rb
@@ -11,24 +11,16 @@ module ThumbnailHelper
   #   html_options::     Additional HTML attributes to add to <img> tag.
   #   notes::            Show image notes??
   #
-  def thumbnail(image, args = {})
-    image_id = image.is_a?(Integer) ? image : image.id
-    locals = {
-      image: image,
-      link: image_path(image_id),
-      link_type: :target, # or :remote
-      link_method: :get,
-      size: :small,
-      votes: true,
-      original: false,
-      theater_on_click: false,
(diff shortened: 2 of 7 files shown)


nimmolo:
Question: I'd like to add tests for the presenter, but i need some pointers.

There are no tests for the existing presenter.

nimmolo:
Merging, because it's blocking further work. 
Feel free to suggest presenter tests to be added later.

nimmolo:
Do not be afraid: A lot of the files changed were the result of a single change of `local_assigns` variable, from `link` to `image_link`. 

This PR doesn't change much except variable routing.

PR #1385, opened 2023-03-03
Add `db/schema.rb` to `.gitignore`

nimmolo:
The deploy script runs migrations, thereby generating a fresh schema.rb every time.
So do developers, when they run `db:migrate` after setting up the VM or migrating.

According to @pellaea, nothing uses the `schema.rb` version committed to github, and it causes many deploy conflicts, so let's get it out of the repo!


mo-nathan:
This is not the right thing.

mo-nathan:
The schema.rb is used to create initial environments.  If it gets removed from the repo there will be no way to build a new system.  I very much doubt the current migrations will do the right thing.  There has to be a way to get a schema.rb that is consistent.

pellaea:
Nathan, do you mean dev-startup needs it?  Does the test suite use it?

I'm totally fine with it staying in there, but if so we need to find a way
to keep it more stable.  It is constantly in conflict with my local copy.

For example, every time I run a db migration it tweaks the options on the
"create table" lines uselessly, so I'm constantly having to clobber my
local copy.

Another case I've seen a number of times is that a dev accidentally gets
their local db in a weird state, resulting in old superfluous tables.  This
gets written into their copy of the schema and accidentally added to PRs.

I thought maybe we could skip the whole thing by simply removing it from
the repo (but allowing rails to create it locally).  But if we somehow need
to get a copy to dev-startup, that doesn't make any sense, I agree.

mo-nathan:
This indicates that there is an issue with one or more of our local databases.  That's what needs to get fixed.  Can you provide a copy of the conflicting schema.rb?  That would help trace down what is wrong.

pellaea:
No, I don't have time for that now.  I've got multiple projects due in the
next few days.  Sorry.  I guess when it comes up again I'll re-open this.
It won't be long, I'm sure! :)

mo-nathan:
Just in case you have a moment with your local environment running, all I need is for you to run: `bundle exec rake db:schema:dump`, send me db/schema.rb, and then do `git checkout db/schema.rb`.

pellaea:
Okay, I can do that...  Mine is currently identical to the one on main.
Maybe we can get Joe and Nimmo and you to do the same test?  At least track
down whose is different?

It is conceivable that when I recently tested Nimmo's new Vagrantfile --
and therefore reinstalled from scratch -- that fixed some long-standing
problem with configuration of my local db...   Ah, and if so, I bet the
production db has the same issue that I used to have...

mo-nathan:
Interesting!  Mine is also identical to the one on main as is the one currently in production.

pellaea:
Nah, I just tried `bundle exec rake db:schema:dump` on production, and it
too is identical to what's currently in main.

JoeCohen:
I agree with Nathan.

From schema.db:

db:schema:load`. When creating a new database, `bin/rails db:schema:load`
tends to
# be faster and is potentially less error prone than running all of your
# migrations from scratch. Old migrations may fail to apply correctly if
those
# migrations use external dependencies or application code.
#
# It's strongly recommended that you check this file into your version
control system.


I do sometimes see the first issue as Jason (conflicting options in the
schema). For me, it takes only one click to fix it locally.

I have seen old superfluous tables in my local db, and can have fixed that
locally. I haven't seen that in a while.
In response to Nathan's question, it was caused -- in my case -- by some
combination of switching between branches with different schemas and/or
using an older db snapshot and/or running migrations in the wrong branch at
the wrong time.
There's a short write-up at:
https://github.com/MushroomObserver/developer-startup/blob/main/DEVELOPER-WORKFLOW.md#loading-a-snapshot-of-the-live-database-optional

I don't remember those tables being added back to the repo, which to me
indicates that it's a rare event. It also can be fixed.
- Joe

pellaea:
Yeah, I guess we just wait until it happens again and see if we can figure
out why it just happened... :(

pellaea:
Yeah, it should be rare, but it seems to happen way too often.  For
example, it just happened to Nimmo out of the blue on production while
deploying the latest main.  WTF?  What possible reason did it have for
suddenly messing with the schema??

mo-nathan:
I'd have to see that example.  The way I look at it, any unexpected change to schema.rb indicates that something is wrong in someone's database and that needs to be hunted down.  You can actually look at that as one of the benefits of having the schema.rb checked into main.  There can be several sources of this problem from people not checking in the new schema.rb after running a migration (which is kind of harmless) to folks getting turned around after running migrations from different branch and doing the correct rollbacks.

nimmolo:
Closing this... it'll be here to preserve the conversation for reference.

PR #1386, opened 2023-03-03
Paginate `GlossaryTerms` index and add `pattern_search`

nimmolo:
Adds a very simple `Query` class for `GlossaryTerm`, which allows us to paginate the glossary terms index, and the `pattern_search` subclass that allows searching. 

Modeled directly on Joe's exceptionally clear code for `ArticlesController`... that is to say, copied verbatim!

@JoeCohen i believe this satisfies these Pivotal stories:
Pagination https://www.pivotaltracker.com/story/show/167657202
Searchable https://www.pivotaltracker.com/story/show/167809123

Also - cleans up `names_controller_test` in a couple spots. Doing it while I think about it.

### DIFF
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -8,12 +8,6 @@ class GlossaryTermsController < ApplicationController
   # ---------- Actions to Display data (index, show, etc.) ---------------------
 
   def index
-    # Index should be paged with alpha and number tabs
-    # See https://www.pivotaltracker.com/story/show/167657202
-    # Glossary should be query-able
-    # See https://www.pivotaltracker.com/story/show/167809123
-    # includes = @user ? { thumb_image: :image_votes } : :thumb_image
-    # @glossary_terms = GlossaryTerm.includes(includes).order(:name)
     filter_index? ? index_filtered : index_full
   end
 
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -12,8 +12,9 @@ class GlossaryTermsController < ApplicationController
     # See https://www.pivotaltracker.com/story/show/167657202
     # Glossary should be query-able
     # See https://www.pivotaltracker.com/story/show/167809123
-    includes = @user ? { thumb_image: :image_votes } : :thumb_image
-    @glossary_terms = GlossaryTerm.includes(includes).order(:name)
+    # includes = @user ? { thumb_image: :image_votes } : :thumb_image
+    # @glossary_terms = GlossaryTerm.includes(includes).order(:name)
+    filter_index? ? index_filtered : index_full
   end
 
   def show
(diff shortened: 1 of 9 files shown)


app/controllers/glossary_terms_controller.rb
```diff
@@ -80,6 +75,47 @@ def destroy
...
+
+  def index_filtered
+    query = find_or_create_query(:GlossaryTerm, by: params[:by])
+    show_selected_articles(
+      query,
+      id: params[:id].to_s,
+      always_index: true
+    )
+  end
+
```
nimmolo, line 93: @JoeCohen - Is this necessary? I didn't include any filtering tabs on the index and now that I think about it, it seems not. I also can't imagine that a `:q` param would be relevant to a glossary term index. We basically just want pagination and pattern search, I reckon.

nimmolo: Maybe this also needs to handle pattern searches explicitly. Looking to `articles_controller` for guidance... I don't find an index action for the `pattern` param in that controller, although there is an `Query::ArticlePatternSearch` class. 

`Articles` are also not available in the search bar dropdown menu. This may be an unimplemented feature for articles.

app/controllers/glossary_terms_controller.rb
```diff
@@ -80,6 +75,47 @@ def destroy
...
+      action: :index,
+      letters: "glossary_terms.name",
+      num_per_page: 50,
+      include: includes
+    }.merge(args)
+    @links ||= []
+
+    show_index_of_objects(query, args)
+  end
+
```
nimmolo, line 116: @pellaea now that i'm adding something with `show_index_of_objects` for the first time, gotta say it's been pretty smooth!

app/controllers/glossary_terms_controller.rb
```diff
@@ -8,12 +8,7 @@ class GlossaryTermsController < ApplicationController
...
 
   def index
-    # Index should be paged with alpha and number tabs
-    # See https://www.pivotaltracker.com/story/show/167657202
-    # Glossary should be query-able
-    # See https://www.pivotaltracker.com/story/show/167809123
-    includes = @user ? { thumb_image: :image_votes } : :thumb_image
-    @glossary_terms = GlossaryTerm.includes(includes).order(:name)
+    filter_index? ? index_filtered : index_full
   end
```
nimmolo, line 12: If we are going to make `pattern_search` available to the glossary, we may want to do a very short `index_dispatch` for this controller. @JoeCohen I will go with your preference, since you've been working on those dispatches. Please feel free to suggest a commit here and below.

app/controllers/glossary_terms_controller.rb
JoeCohen: What you're saying about filtering, :by, and :q makes sense. 
Perhaps in the future we could have filtering, by, and q. But in this PR  I think they should be omitted.

It's been so long since I wrote Articles that I don't remember anything about why I wrote Articles the way I did.
It was probably a combo of copying some example from the cloud and groping in the dark.

test/controllers/names_controller_test.rb
```diff
@@ -266,7 +260,7 @@ def test_pagination_page2
   def test_pagination_letter
```
JoeCohen, line 260: It's great to get rid of the raw sql, but do these changes belong in this PR?

JoeCohen (changes requested):

test/controllers/names_controller_test.rb
nimmolo: I can move to another PR, no prob

JoeCohen:
Just saw this.
I'd go without a dispatch for now because there are only a few cases. It's
simplest that way, and not hard to change.
We can always change it later. (There are multiple indices that use "case"
or if/then/else.)
The motive for the dispatching was to make the Abc score <= 20.
I don't think that will be a problem here.

app/controllers/glossary_terms_controller.rb
nimmolo: Ok, thanks. In a separate PR I may remove the pattern search query class from Articles since it is as yet unimplemented, what do you think? Or,  I could implement it. But it doesnt seem that useful. 

I'm glad it was there for me to copy here, though!

JoeCohen:
I'm in favor of removing unused code.

nimmolo:
@JoeCohen Thanks. I've added search with tests, and removed filtering.

PR #1393, opened 2023-03-04
Add git stash to deploy.sh

mo-nathan:
This allows there to be persistent local changes on the production server.


nimmolo:
Sounds good to me. How would I use?

nimmolo:
Oh, nifty. It just stashes and pops automatically. That's great!

nimmolo (approved):

PR #1403, opened 2023-03-06
Copy working nginx.conf from mo.org

mo-nathan:


nimmolo (approved):
Ohhhhh thank you!! got it! My second capture group didnt include the file ext.

PR #1406, opened 2023-03-08
Location description form fix - commit action and method

nimmolo:
This attempts to fix a form error spotted by @JoeCohen.

The sense I make of it is that `form_with` seems to require more explicit args here because the `scope` that the params are attached to, `description`, is not the same as the relevant model name, `location_description`.


app/views/locations/descriptions/_form.html.erb
```diff
@@ -1,30 +1,40 @@
-<%= form_with(scope: :description, url: action,
+<%
```
JoeCohen, line 1: Would it make sense to move this to a helper?

JoeCohen (approved):
Thanks for fixing this. I scratched my head for a long time, but couldn't figure it out.
I tried it locally and it works like a charm.
Just the one question about whether to move some stiff from the form to a helper. It's your call.

app/views/locations/descriptions/_form.html.erb
nimmolo: Joe, you're right. Actually I think this would be a proper use of a presenter, rather than a helper. The difference as far as I can tell is that presenters are designed to just perform logic and return a set of data, where helpers return HTML markup.

I don't know if you noticed my recent PR where I cleaned up the matrix box presenter and thumbnail presenter to bring them back to the intended purpose, and as a result they're less confusing. (that PR was only around briefly, I merged immediately).

I don't picture this getting more complex, so i'm ok with it like this, but let's try a new presenter. It will be good practice to show the pattern.

JoeCohen:
I know that presenters exist, but nothing else.
Thanks for explaining the difference.
I did notice the PR, but didn't have time to look at it.
Maybe I've been using helpers where I should have used presenters... and am
about to do it once more.
I'll have to read up on them.

app/views/locations/descriptions/_form.html.erb
nimmolo: Presenters are a good example of "service objects".

You initialize them with a few args, and they give you an object full of data. Simple as that. 
An advantage is that they do not pollute the global namespace for methods, like helpers do.

Check the three we have for examples. I think the old `RssLogsPresenter` scared everyone off them, because it was generating all kinds of HTML through very convoluted means. 

The key is to keep html generation accessible in helpers, and just use presenters to figure out logic that's not coming straight off the db. They can lighten the load for controllers and remove logic from views.

nimmolo:
Shoehorning a little stuff in here from gigantic upcoming form-cleanup PR, to eliminate conflicts.

PR #1411, opened 2023-03-09
Delete redundant id param on User page

JoeCohen:
- Deletes, on the user page, `id:` params which repeat other param's values.
Ex:
```ruby
https://mushroomobserver.org/locations/descriptions?by_editor=4468&id=4468
```
The `id=4468` is unnecessary and can create a query which later causes an index to display starting at the wrong index page.
- Delivers [Pivotal 184539899](https://www.pivotaltracker.com/story/show/184539899)

### Suggested Manual Test
- Go to any User page (/users/nnnn)
- Make sure the links work in the RH navbar and in the contributions area.

### DIFF
--- a/app/helpers/user_stats_helper.rb
+++ b/app/helpers/user_stats_helper.rb
@@ -3,12 +3,10 @@
 # Helpers for user view
 module UserStatsHelper
   def user_stats_links(user)
-    links = {}
-    user_stats_links_table(user).each do |key, controller, action, params|
-      links[key] = url_for(controller: controller, action: action,
-                           params: params)
+    user_stats_links_table(user).each_with_object({}) do |row, links|
+      links[row[0]] = url_for(controller: row[1], action: row[2],
+                              params: row[3])
     end
-    links
   end
 
   #########################################################
--- a/app/helpers/user_stats_helper.rb
+++ b/app/helpers/user_stats_helper.rb
@@ -5,11 +5,6 @@ module UserStatsHelper
   def user_stats_links(user)
     links = {}
     user_stats_links_table(user).each do |key, controller, action, params|
-      unless [key].intersect?([:location_description_authors,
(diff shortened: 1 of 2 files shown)


app/helpers/user_stats_helper.rb
```diff
@@ -3,17 +3,10 @@
 # Helpers for user view
 module UserStatsHelper
```
JoeCohen, line 4: @nimmolo Should this be a Presenter instead of a Helper?

test/controllers/users_controller_test.rb
```diff
@@ -90,6 +90,18 @@ def test_show
...
+      /\?.*=(\d+).*&id=\1/, # some param=n followed by id with same value
+      false,
+      "Links should not use the same value for id and another param"
+    )
+    assert_select(
+      "a:match('href', ?)",
+      /\?.*id=(\d+).*&\S+=\1/, # id=n followed by another param with same value
+      false,
+      "Links should not use the same value for id and another param"
+    )
```
JoeCohen, line 104: Is there a way to combine the regexp's (in order to use a single `assert_select`)/

test/controllers/users_controller_test.rb
```diff
@@ -90,6 +90,18 @@ def test_show
     assert_select(
```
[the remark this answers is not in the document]
nimmolo: I personally feel like the controller is the place to test this... but testing is the last thing anybody should listen to me about!

test/controllers/users_controller_test.rb
nimmolo: Only think i can imagine is if you just omit the `id` and search for two `=(\d+)`. 
Doesn't seem safe though, what if one of the params has very few digits?

nimmolo (approved):

PR #1419, opened 2023-03-11
Form submit button helper, more form consolidation

nimmolo:
DRY up the form submit buttons. There are more than 100 submit buttons in /views! 

Part of the process of writing helpers to make it easier to code consistent forms, and apply repetitive Bootstrap classes in one place (the helper rather than in every view). This PR continues to evolve other existing helpers to be versatile enough to use on almost all inputs. More helpers on the way.

Before: 
```ruby
f.submit(button.t, class: "btn btn-default center-block my-3")
```

After:
```ruby
submit_button(form: f, button: button, center: true)
```

Also refactors the few remaining form inputs that didn't use form helpers yet.

### DIFF
--- a/app/helpers/forms_helper.rb
+++ b/app/helpers/forms_helper.rb
@@ -38,11 +38,12 @@ module FormsHelper
     end
   end
 
-  def center_form_submit(**args)
+  # <%= centered_submit(form: f, button: button.t) %>
+  def centered_submit(**args)
     opts = args.except(:form, :button, :class)
     opts[:class] = "btn btn-default center-block my-3 #{args[:class]}"
 
-    form.submit(button.t, opts)
+    form.submit(button, opts)
   end
 
   # Bootstrap checkbox: form, field, (label) text, class,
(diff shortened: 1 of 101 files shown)


app/helpers/forms_helper.rb
```diff
@@ -38,57 +38,77 @@ def panel_block(**args, &block)
...
+    end
+
+    opts = args.except(:form, :button, :class, :center)
+    opts[:class] = "btn btn-default"
+    opts[:class] += " center-block my-3" if args[:center] == true
+    opts[:class] += " #{args[:class]}" if args[:class].present?
+
+    args[:form].submit(args[:button], opts)
+  end
+
```
nimmolo, line 54: @JoeCohen - Here's an example. This has been applied to every submit button on Earth. 

The main comment for this PR shows the difference in syntax. 

I'm currently writing helpers for all the other input types on Earth, and applying them in the forms. An eventual payoff from converting the forms to `form_with`, but whew it's a load of shoveling. Helpers are still evolving to accommodate the different situations/args i'm encountering.

JoeCohen:
Very nice.
It's great to be able to do that without all the css stuff.

PR #1420, opened 2023-03-12
Allow city + county

JoeCohen:
- Lets users include both city and county in Location name.
We used to tell users that we prefer to omit county in order to have consistent Location names.
But county is now included in so many Locations that this no longer makes sense.
See e.g.:
>Jason Hollinger  Sun, Jan 15, 2023 at 4:47 PM
To: Joseph Cohen 
Cc: "webmaster mushroomobserver.org" 
And while we're at it, can we finally remove the bit about "county may not be necessary"?  The community has clearly
voted to ignore that rule!

- Delivers [Pivotal 184449491](https://www.pivotaltracker.com/story/show/184449491)

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -531,7 +531,6 @@ class Location < AbstractModel
     unless check_db && location_exists(name)
       reasons += check_for_empty_name(name)
       reasons += check_for_dubious_commas(name)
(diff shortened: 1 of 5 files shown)


JoeCohen:
Marginal drop in coverage is due to the removal of covered lines; there are no newly uncovered lines.

PR #1422, opened 2023-03-12
Remove fancy_time display

JoeCohen:
- Displays timestamps as dd Mon yyyy hh:mm:ss instead of as fancy_time
The fancy_time display broke tests whenever we switched from daylight to standard or vice versa.
Normal display is more precise and accurate than fancy time. 
See [Slack conversation starting 2023-03-1-2 0619 PDT](https://mushroomobserver.slack.com/archives/C040TH9FV/p1678627166657409)
- Delivers [Pivotal 184679272](https://www.pivotaltracker.com/story/show/184679272)

### Manual Tests
- Go to Activity Feed
- Timestamps should look like `15 Feb 2023 22:10:30`
-


JoeCohen:
The marginal coverage decrease is due to the removal of previously tested lines.
There is no newly-uncovered line.
Coveralls [hard-to-interpret statistics](https://coveralls.io/builds/57767562) say that coverage of app/extensions fell,
But every extension has coverage that is unchanged, or -- for `TimeExtensions` -- increased!

PR #1424, opened 2023-03-13
`views/account` form-helper overhaul

nimmolo:
This PR is part of a sweep through the forms to complete the conversion of most form markup to use the new form helpers.

It will add more form helpers as needed.

The idea of the helpers is to standardize styling in the helpers, and make it easier to code forms.


nimmolo:
Seeming loss of coverage is due to the build-out of new, as-yet-unimplemented form helpers.

They will be tested when they get used.

PR #1426, opened 2023-03-13
Widen query for "Observations needing identification"

nimmolo:
It turns out that observations of "Fungi" aren't the only ones that need help with identification. 
This PR widens the basic "Needs ID" query to include observations whose:

- naming is "Fungi" (current), regardless of confidence
- naming is specific, but lacks community confidence (current)
- naming is general, i.e. of a taxon wider than genus, regardless of confidence

To be more useful to specialists of a taxon or a region, the "Needs ID" page seems like it should be able to filter unidentified observations to within a broad area of familiarity. The page UI introduces two filters that replace "Pattern search", on this page only. 

- search by "clade": narrows the search to "general namings, or specific namings without confidence", within a taxonomic clade
- search by "region": narrows the search to general or unconfident observations within a region
- ~search by "user": narrows the search to observations by a specific user~ (projected, not implemented)

The clade filter benefits from a new autocompleter, that completes only **names above genus** (rather than all names).

Note that you can only use one filter at a time, for now. Ideally you could apply multiple filters at once, but i'm going to wait to implement that. I think it will work better with autocomplete-selects that can submit the name/location/user ID as the filter value, rather than a text string requiring a table lookup.

Summary
These changes make the query on "Needs ID" more costly, but hopefully make it a lot more useful. We should monitor the page's performance for ways to reduce the cost of the query.

[Further discussion here](https://github.com/MushroomObserver/mushroom-observer/discussions/1404).

### DIFF
--- a/app/classes/query/observation_needs_id.rb
+++ b/app/classes/query/observation_needs_id.rb
@@ -6,7 +6,7 @@ module Query
     def parameter_declarations
       super.merge(
         in_clade?: :string,
-        in_region?: :string #,
+        in_region?: :string # ,
         # by_user?: :string
       )
     end
@@ -60,11 +60,14 @@ module Query
       end
     end
 
-    # The tricky thing here is, without the user.id as the value passed in
+    # The tricky thing here is, without the user.id being the value passed in
     # params[:filter][:term], we're hunting for a user from a string like
     # "Name <name>". Better to have the id as the value!
+    # Below uses the method in query/initializers/advanced_search to get a
+    # string but is expensive. Something like
+    # joins(:users).where((User[:login] + User[:name]).matches(str))
     # def by_user_condition
-    #   user = params[:by_user]
+    #   user = params[:by_user].to_s.gsub(/ *<[^<>]*>/, "")
     # end
   end
 end


nimmolo:
Interesting finding:

In `Query::ObservationBase#initialize_needs_id`, when I wrote it in pure SQL like this, the query was VERY slow:
```ruby
    def initialize_needs_id
      names_above_genus = Name.with_rank_above_genus.select(:name_id).to_sql
      where << "observations.name_id IN (#{names_above_genus}) OR " \
               "observations.vote_cache <= 0"
      where << "observations.id NOT IN (SELECT DISTINCT observation_id " \
               "FROM votes WHERE user_id = #{User.current_id})"
      where << "observations.id NOT IN (SELECT DISTINCT observation_id " \
               "FROM observation_views WHERE observation_views.user_id = " \
               "#{User.current_id} AND observation_views.reviewed = 1)"
    end
```
Note that the inner query above, `names_above_genus`, produced the SQL the way we would normally write it in the Query class,
```SQL
SELECT `name_id` FROM `names` WHERE (`names`.`rank` > 9) AND (`names`.`rank` != 16)
```

When I let AR map the ids, and passed the array of ids as a joined string, it seemed to take probably 1/10 the time:
```ruby
    def initialize_needs_id
      names_above_genus = Name.with_rank_above_genus.map(&:id).join(", ")
      where << "observations.name_id IN (#{names_above_genus}) OR " \
               "observations.vote_cache <= 0"
      where << "observations.id NOT IN (SELECT DISTINCT observation_id " \
               "FROM votes WHERE user_id = #{User.current_id})"
      where << "observations.id NOT IN (SELECT DISTINCT observation_id " \
               "FROM observation_views WHERE observation_views.user_id = " \
               "#{User.current_id} AND observation_views.reviewed = 1)"
    end
```

This suggests to me that some of our inner queries in the Query class may be much slower than the current AR equivalents, and that we may get quite a performance boost

JoeCohen:
One other thought
IMO the search should include only Names classified as _Fungi_ or the name Undetermined (and maybe _Protozoa_ if we want to encourage more specific id of slime molds).
IMO we should not encourage people to ID things in other Kingdoms. 
I lilke to hear what @mo-nathan thinks about that?

pellaea:
Yikes!  Let me look at this later when I get a chance.  These inner queries
look hideous, like selecting all observations a user has voted on?  Eek!  I
just don't have time now.

nimmolo:
Thanks guys. 

@JoeCohen 

What you described re: "only Names classified as Fungi" — that's the current query. It has a few problems, probably you agree:
- You cant search for an upper taxon in your area of expertise (or familiarity, or interest)
- It occurred to me, that it's preferable not encouraging the "narrowing down" approach, but I think that ship has sailed. There are a lot of observations that already have fairly useless upper taxa, like "Agaricales sensu lato", that still need ID. EXD can't be the only person doing this.

What you described above, with filtering, is my ideal. Requires more code, but maybe this is the time for it.
- I don't think a `select` filter for higher taxa, like family, is the way to go. There are too many of them. 
- What I think would work is a text field for filtering by children of the given taxon. Then we'd do the query with the scope `include_subtaxa_of(given_name)`
- To autocomplete the higher taxa, it would be nice to implement a new NameAboveGenus primer, no? That would make it quicker to work with.

@pellaea 
Thanks for considering the inner queries.  Note that the hideous votes query is no change from the current query, though! i'm open to reconsidering it of course.

pellaea:
re: votes inner query -- Yeah, joining to votes is probably prohibitive.
But that's probably just because we haven't added an index to votes.id
yet?  I would be interested in testing -- in the fullness of time! :) --
the difference between this current approach and doing a join, and then
doing a join after adding an index.  Maybe there is little difference at
our current database size?  It just feels crazy to me!  But that's what
computers are good at -- crazy repetition! :)

nimmolo:
Ok, i've got a filter that kind of works! There's a new "NameAboveGenus" autocomplete. Kinda nice! Probably not passing tests and no new tests have been written, so that's coming.

Check it out locally if you can, and please examine the query, it seems inefficient. The filter aims to get every obs of (any subtaxon under the given taxon above genus), plus any obs with (subtaxon of given taxon at or below genus AND unconfident votes).

mo-nathan:
Just catching up on this discussion.  I don't have a strong opinion about observations in other kingdoms.  I don't really see a strong reason to exclude them.  I do wonder if rather than using complex runtime queries if the identification status of an observation should be associated with the observation itself.  These complex queries could be use to initialize these values for existing observations, but going forward it could just be maintained on new observations.  Exactly what "confidence" means is an interesting question, but it's some combination of the precision of the id and the experience of the person (or persons) making the id.  I do think someone like Alan should be able to mark an observations as "Needs Review" or "Reviewed".  Ultimately it's a continuum from photos of a type specimen to "Fungi" and we have to make a decision where to draw the line.  The question is where that line should be drawn and how much control we want to give users over where a given observation falls.

app/models/observation.rb
```diff
@@ -258,9 +258,14 @@ class Observation < AbstractModel
         -> { where.not(name: Name.unknown) }
   scope :without_name,
         -> { where(name: Name.unknown) }
-  scope :without_confident_name, lambda {
-    without_name.or(where(vote_cache: ..0))
+  scope :with_name_above_genus,
+        -> { where(name_id: Name.with_rank_above_genus.map(&:id)) }
```
JoeCohen, line 262:
```suggestion
        -> { where(name_id: Name.with_rank_above_genus.ids) }
```
a wee bit shorter

app/models/observation.rb
```diff
@@ -279,10 +284,21 @@ class Observation < AbstractModel
...
-  scope :needs_identification, lambda { |user|
-    without_confident_name.without_vote_by_user(user).
-      not_reviewed_by_user(user).distinct
+  scope :needs_id_for_user, lambda { |user|
+    needs_id.without_vote_by_user(user).not_reviewed_by_user(user).distinct
+  }
+  scope :needs_id_by_taxon, lambda { |user, name|
+    name_plus_subtaxa = Name.include_subtaxa_of(name)
+    subtaxa_above_genus = name_plus_subtaxa.with_rank_above_genus.map(&:id)
+    lower_subtaxa = name_plus_subtaxa.with_rank_at_or_below_genus.map(&:id)
```
JoeCohen, line 293: `ids` instead of `map(&:id)`
See above.

app/models/observation.rb
nimmolo: Thanks @JoeCohen, i was completely unaware of that method.  I'm having trouble finding available documentation of it, and it's hard to google. 

APIDock implies it's an alias for `pluck`, in which case I don't think we would want to use it. In my long research detour on this subject a few months ago, I found that `pluck` was definitively slower than `map`. (`select` was way slower too.)

Please let me know if you're able to find out what it does.

nimmolo:
I'm putting this PR back in draft until I've got a better filter working, and some tests.

@mo-nathan - Let's continue on a GitHub "discussion" or on Slack if you prefer.

Sorry for the premature PR, and thanks everybody for checking it out.

JoeCohen:
You're right. `.ids` plucks the primary id.

# File activerecord/lib/active_record/relation/calculations.rb, line 220def ids
  pluck primary_keyend

app/classes/query/observation_base.rb
nimmolo, line 1: Moved the `needs_id` query entirely out of `observation_base`, and into its own subclass

app/classes/query/observation_needs_id.rb
nimmolo, line 1: This is the new subclass

app/views/layouts/_search_bar.html.erb
nimmolo, line 1: Note here the pattern search is replaced by a filter with a similar UI, only on this index.

PR #1430, opened 2023-03-16
Confirm Clear List

JoeCohen:
- Adds an "Are you sure?" confirmation to the Clear List buttons on the admin/blocked_ips page.
This makes it harder to accidentally toast  the OK ips or Blocked ips lists. 
(The lists are backed up, so that isn't absolutely fatal. But it it's a bit of trouble to restore them. 
I've been meaning to do this for a while.)


nimmolo (approved):

JoeCohen:
Thx!

PR #1448, opened 2023-04-04
Preliminary alterations to matrix boxes

nimmolo:
This PR does some alterations to matrix-box layouts adapted from the old Bootstrap 4 branch.
Basically it tries to rough out my intended "redesign" as far as possible in Bootstrap 3. 

With all the recent refactoring, the move to 4 seems close, but i'd like to try some of these changes and get feedback before that.

- Eliminates some of the "mat" around the images in the matrix box
- Tries to fit the images full-width, or if it can't, puts them in a darker full-width mat area. (There is a max vertical proportion on the images to keep a tall skinny image from throwing off the entire layout. It could be adjusted to fit taller images than currently, though.)
- Allows for 4 columns of boxes on larger screens
- Legibility:  Left aligns the text in the box
- Legibility:  Makes all text the primary (most legible) color, with links getting the link color and an underline on hover
- Legibility:  Breaks the taxon name (authority on new line) and makes authority smaller
- Legibility:  Floats the ID to the right, away from the observation name
- Legibility:  Gives matrix boxes a bigger title, when they don't have an image
- Legibility:  Puts the "log" details in a "panel-footer" so the log is easier to scan

matrix_box_bootstrap_3

- In Bootstrap 4, the "log-footer" is auto-fit within cards of the same height, and will not need the JS used here to equalize heights.


nimmolo:
Since the changes in this PR has been designed, proposed and discussed sporadically over the last 3 years, will merge tomorrow if no objections. It's been a long road to get here!

No need for code review, but comments on the screenshot above, or a test drive welcome.

PR #712, opened 2020-11-17
Search for name

pellaea:
Changes behavior of plain old observation search for a name.

Type a name like "Xanthoria elegans" into the search bar, then search for observations.
It used to return only observations of Xanthoria elegans, not its synonym "Rusavskia elegans".
I am arguing that the vast majority of users will expect it to include synonyms by default.

Also, I think that the majority of users, at least new users, will expect a search for "Amanita" to show all observations of Amanita, not just those too poor to identify to species.  Therefore this PR now makes the default behavior of all searches with the names parameter supplied to be include_subtaxa:yes.

This PR changes it so that the observation pattern search recognizes that the pattern the user entered is a name (checks for that name in the database under text_name and search_name), then moves the pattern parameter into the names parameter.  And if the names parameter is present, it supplies defaults for both include_synonyms and include_subtaxa.

PR #1449, opened 2023-04-05
Reboot of "Search for name" PR

nimmolo:
This is a manual reconstruction of @pellaea's PR #712. I couldn't resolve merge conflicts since the test methods have moved, and the tests in `ObservationsControllerTest` seem to have been refactored.

> Changes behavior of plain old observation search for a name.
Type a name like "Xanthoria elegans" into the search bar, then search for observations.
It used to return only observations of Xanthoria elegans, not its synonym "Rusavskia elegans".

> I am arguing that the vast majority of users will expect it to include synonyms by default.

> Also, I think that the majority of users, at least new users, will expect a search for "Amanita" to show all observations of Amanita, not just those too poor to identify to species. Therefore this PR now makes the default behavior of all searches with the names parameter supplied to be include_subtaxa:yes.

> This PR changes it so that the observation pattern search recognizes that the pattern the user entered is a name (checks for that name in the database under text_name and search_name), then moves the pattern parameter into the names parameter. And if the names parameter is present, it supplies defaults for both include_synonyms and include_subtaxa.


nimmolo:
The test failures in this reboot all have to do with generated titles. @pellaea seems to have solved it in #712, but things must have changed.

> Ugh, the failing tests all had to do with query titles. That is definitely a larger problem than I have time to solve completely. (How do we get observation queries to describe themselves in English when there are so many possible permutations of parameters??) But I did take a stab at it, trying at least to preserve the titles for the flavor=all versions of the old flavors like by_user, of_name, at_location, etc. Almost certainly not perfect, but at least all the tests pass again!

> This is definitely something that needs to happen eventually if we ever want to convert all those old "flavorful" queries into "unflavored" (i.e. flavor = :all) queries using filter parameters instead. (Which would be a useful way to make various queries and indexes more transparent and RESTful and persistent. "Observations By User" then becomes a generic observation query with "user=xxx" filter parameter, and we can automatically fill in the search bar with "user:xxx". Then it will be very easy for users to modify the query in nontrivial and useful ways by simply adding terms to the search bar. And so on and on. You get the point, I'm sure. Useful direction to move, I think, just going to be a long journey...

Currently failing: ObservationsControllerTest, ObservationsControllerSupplementalTest, FilterTest. 

The current issue seems to be that pattern search is generating titles like 
`Observations with Observations of #{name}` 
or 
`Map of Observations with Observations of #{name}`

nimmolo:
I believe the remaining failures in `ObservationsControllerTest` may not be failures, but rather that the test expectations need to be adjusted. 

@JoeCohen and @pellaea please clarify if the below are the new expected titles, or not:
`ObservationsControllerTest`
`#test_index_look_alikes` - old title: `Observations by Confidence Level`, new title: `Observations of Tremella`
`#test_index_name ` - old title: `Observations by Confidence Level`, new title: `Observations of Fungi`
`#test_index_related_taxa ` - old title: `Observations by Confidence Level`, new title: `Observations of Tremella`
`#test_index_pattern_multiple_hits ` - old title: `Observations Matching ‘Agaricus’`, new title: `Observations of Agaricus`

`NamesControllerTest`
`#test_index_with_saved_query ` - old title: `Names with Observations created by Tricky Dick`, new title: `Names with Matching Observations`

nimmolo:
I've redone the expectations as described in note above, and tests are passing.

The one exception i'd like to make is for "Names with Observations created by [USER]", but i'd try that in a separate PR.

nimmolo:
Figured out "Images with observations by user" title exception, merged into this PR. 

I'm going to merge.

PR #1450, opened 2023-04-06
Switch to `fastimage` gem to read image dimensions

nimmolo:
There's a persistent issue with certain devices' images getting uploaded. When they are stored in the db, the width and height are swapped, leading to layout problems that only became apparent when I added lazyloading and image dimensions.

So - This is what happens in our code during the action of uploading the image to an observation. The obs controller calls

- `ObservationsController#create_image_objects`…
- `Image.process_image`...
- `Image.set_image_size`. This is the action that actually saves the width and height to the database, using our script `jpegsize`.

```ruby
  # Get image size from JPEG header and set the corresponding record fields.
  # Saves the record.
  def set_image_size(file = local_file_name(:full_size))
    script = Rails.root.join("script/jpegsize").to_s
    output, _status = Open3.capture2(script, file)
    w, h = output.to_s.chomp.split
    return unless /^\d+$/.match?(w.to_s)

    self.width  = w.to_i
    self.height = h.to_i
    save_without_our_callbacks
  end
```

After experimenting with some user images that were getting the proportions swapped, I found that the `fastimage` gem reads the dimensions correctly, so without further ado, I propose changing the dependency from our own `jpegsize` to the gem.

Two other scripts depend on `jpegsize` though: `rotate_image` and `process_image`.

### DIFF
--- a/script/process_image
+++ b/script/process_image
@@ -75,8 +75,9 @@ fi
 # Make sure image is oriented correctly.
 log_cmd exifautotran $full_file
 
+# Changing from $app_root/script/jpegsize to identify (ImageMagick)
 if (( $set_size )); then
-  size=$( $app_root/script/jpegsize $full_file )
+  size=$( identify $full_file )
   w=$( echo $size | sed "s/ .*//" )
   h=$( echo $size | sed "s/.* //" )
   log_mysql "UPDATE images SET width=$w, height=$h WHERE id=$id" || true
@@ -121,7 +122,7 @@ if (( !$development && !$errors && $transferred_any )); then
 fi
 
 if (( !$development && $errors )); then
-  send_mail -s "[MO] process_image" $webmaster_email < $log_file 
+  send_mail -s "[MO] process_image" $webmaster_email < $log_file
 fi
 
 exit $errors
--- a/script/process_image
+++ b/script/process_image
@@ -75,9 +75,8 @@ fi
 # Make sure image is oriented correctly.
 log_cmd exifautotran $full_file
 
-# Changing from $app_root/script/jpegsize to identify (ImageMagick)
 if (( $set_size )); then
-  size=$( identify $full_file )
+  size=$( $app_root/script/jpegsize $full_file )
   w=$( echo $size | sed "s/ .*//" )
   h=$( echo $size | sed "s/.* //" )
   log_mysql "UPDATE images SET width=$w, height=$h WHERE id=$id" || true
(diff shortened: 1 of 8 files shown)


script/process_image
```diff
@@ -75,8 +75,9 @@ fi
 # Make sure image is oriented correctly.
 log_cmd exifautotran $full_file
 
+# Changing from $app_root/script/jpegsize to identify (ImageMagick)
 if (( $set_size )); then
-  size=$( $app_root/script/jpegsize $full_file )
+  size=$( identify $full_file )
   w=$( echo $size | sed "s/ .*//" )
```
nimmolo, line 81: @pellaea is this the correct syntax?

script/rotate_image
```diff
@@ -80,7 +80,8 @@ case $op in
 esac
 
 # Update size and clear "transferred" bit in database.
-size=$( $app_root/script/jpegsize $orig_file )
+# Changing from $app_root/script/jpegsize to identify (ImageMagick)
+size=$( identify $orig_file )
 w=$( echo $size | sed "s/ .*//" )
```
nimmolo, line 85: @pellaea ditto

test/models/image_script_test.rb
```diff
@@ -114,6 +114,7 @@ def mysql_config_file
       file.puts("#{local_root}/320//#{in_situ_id}.jpg")
       file.puts("#{local_root}/thumb//#{in_situ_id}.jpg")
     end
+    # Not sure how to convert this to ImageMagick `identify`
     output, _status = Open3.capture2(script_file("jpegsize"), "-f", tempfile)
```
nimmolo, line 118: @pellaea Wasn't sure what to do here.

test/models/script_test.rb
```diff
@@ -34,17 +34,17 @@ def script_file(cmd)
...
+  #   script = script_file("jpegsize")
+  #   [
+  #     ["Coprinus_comatus.jpg", 2288, 2168],
+  #     ["perf.jpg", 4288, 2848],
+  #     ["sticky.jpg", 407, 500]
+  #   ].each do |file, width, height|
+  #     result = `#{script} #{::Rails.root}/test/images/#{file}`.chomp
+  #     assert_equal("#{width} #{height}", result)
+  #   end
+  # end
```
nimmolo, line 47: Maybe we don't need this test?

script/process_image
nimmolo: Resolved by calling fastimage from the jpegresize script itself

PR #1451, opened 2023-04-06
Add required gem class to Image

nimmolo:
Fixes bug on uploading images to an Obs. Gem class needs requiring!

### DIFF
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -2,6 +2,7 @@
 
 require("open3")
 require("mimemagic")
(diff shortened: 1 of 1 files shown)


JoeCohen:
Thanks for jumping on this.

nimmolo:
Put this in the email thread, but to have it in the right place: 
If we still get an error on uploads, we may have to also add this at the top of image.rb:
```ruby
require("stringio")
```

PR #1454, opened 2023-04-08
Fix edit Classification

JoeCohen:
- Fixes a bug that throws an error when the Edit Classification form is submitted.
- Adds a test assertion to expose the bug:
```ruby
Routing Error
No route matches [POST] "/names/883/classification"
```
- Submits with PUT rather than POST. 
- Delivers [Pivotal #184899433](https://www.pivotaltracker.com/story/show/184899433)

### Manual Test
- Edit a Classification. Ex: http://localhost:3000/names/883/classification/edit
- erase or modify the Classification string
- Save
**Expected Result**: It should display the Name with the changed Classification.


JoeCohen:
Provided that CI checks pass, I'll probably merge and deploy this immediately so that I can use it live.

PR #1458, opened 2023-04-08
Fix versions table

### LINKED ISSUE #1456: Name version links are missing
The Show Name Versions page is no longer linking to other versions.
Ex: https://mushroomobserver.org/names/27419/versions?version=11
**Each** "Version nn" should be a link to the appropriate version. 
Screenshot 2023-04-08 at 8 27 35 AM

nimmolo:
Fixes #1456. The table helper was just returning the link *text* if it wasn't the current version of the object.

- In some places, we've been passing the argument name `link` for the `text` of the link. This is likely to cause confusion when methods are broken down for abc, because we may assume we're dealing with a link when we're only dealing with a string. Recently I've been trying to replace that arg name wherever I encounter it.
- Helper methods are all in the global namespace, even potentially "private" methods. So i'm renaming some of these methods to be more specific.

### DIFF
--- a/app/helpers/versions_helper.rb
+++ b/app/helpers/versions_helper.rb
@@ -99,9 +99,9 @@ module VersionsHelper
   end
 
   def initial_version_link_text(ver)
-    link = "#{:VERSION.t} #{ver.version}"
-    return link unless ver.respond_to?(:format_name)
+    text = "#{:VERSION.t} #{ver.version}"
+    return text unless ver.respond_to?(:format_name)
 
-    link + " #{ver.format_name.t}"
+    text + " #{ver.format_name.t}"
   end
 end
--- a/app/helpers/versions_helper.rb
+++ b/app/helpers/versions_helper.rb
@@ -43,7 +43,7 @@ module VersionsHelper
 
   def previous_version_link(previous_version, obj)
     str = :show_name_previous_version.t + " " + previous_version.version.to_i
-    initial_html(obj) +
+    initial_version_html(obj) +
       link_with_query(str,
                       controller: "#{obj.show_controller}/versions",
                       action: :show, id: obj.id,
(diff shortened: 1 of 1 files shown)


JoeCohen:
Thanks for doing this. 
I'll try to look it over before the Tech Meeting -- mostly so that I get caught up about the helpers.
But you're welcome to merge it if you like.

nimmolo:
@JoeCohen - I can leave it to you to merge when you've checked it out.

app/helpers/versions_helper.rb
```diff
@@ -43,80 +43,65 @@ def show_past_versions(obj, args = {})
...
                       controller: "#{obj.show_controller}/versions",
                       action: :show, id: obj.id,
                       version: ver.version)
     end
   end
 
-  def initial_link(ver)
+  def initial_version_link_text(ver)
     link = "#{:VERSION.t} #{ver.version}"
     return link unless ver.respond_to?(:format_name)
```
JoeCohen, line 103: @nimmolo, @pellaea 
Under what circumstances could the conditional (`ver.respond_to?(:format_name)`) be true?
(I'm asking because I'd like to cover the next line, which is currently not covered.)

JoeCohen (approved):
I've tested it manually, and it works fine, so I think you should merge it.

(I haven't reviewed the code in detail and am a bit confused by how parts of it work.)

app/helpers/versions_helper.rb
nimmolo: @JoeCohen Under no circumstance I can figure. 

Name versions respond to 
 `text_name`: "Cabalodontia",
 `search_name`: "Cabalodontia M. Piatek",
 `display_name`: "**__Cabalodontia__** M. Piatek",
 `sort_name`: "Cabalodontia  M. Piatek",

I guess the question is, would any of these make this table more useful?

nimmolo: Incidentally I wonder why Piątek lost the accent under the "a". That accent is more or less equivalent to a nasal "n" and is important in Polish...

JoeCohen: 1. It would be slightly more useful -- provided you thing there's space -- to show for each version:
version #, date, user, Display Name
That would show the key changes in the table, without having to click through each of the versions.
BUT, I don't know if it's worth your time.
2. Good catch on Piątek. I fixed it. My guess is that the accent got dropped because of user error/carelessness.

nimmolo: We can make space in the table. But i can't get the dang html tags in the display name to render.

Even after calling html_safe in about 5 different places. I despise dealing with that stuff! It takes me forever to figure out where to permit the tags.

Screen Shot 2023-04-08 at 4 19 23 PM

I can check it in and hand it over if you think you can figure it out.

nimmolo: Oh hey, i got it. Checking it in now.

PR #1460, opened 2023-04-08
Fix typo and mistake in form

### LINKED ISSUE #1455: Edit Classification help_block renders incorrectly
The class renders as text instead of as a tag.
Ex: https://mushroomobserver.org/names/58195/classification/edit
>`< class="help-block">List the higher taxa containing this family, one name per rank. Do not include genus or lower ranks. ...`

I also wonder if this should be translated (i.e. called with `.t` rather than `.l`.)

nimmolo:
Fixes #1455

The first arg for `help_block` should be a symbol: `:p` not `p`

As @JoeCohen  and @pellaea noted, the second arg should get a `.t` not a `.l`

### DIFF
--- a/app/views/names/classification/edit.html.erb
+++ b/app/views/names/classification/edit.html.erb
@@ -15,7 +15,7 @@
 
   <%= text_area_with_label(
         form: f, field: :classification, rows: 10, value: @name.classification,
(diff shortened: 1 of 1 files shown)


JoeCohen:
Aha!
Thanks for fixing it and for the explanation.

PR #1468, opened 2023-04-18
Change home page to Observations#index by RssLog

nimmolo:
Changes the default home page from RssLogs to the Observations index, ordered by rss_log activity.

>~and adds a new preference to restore the RssLogs as homepage for power-users who want that~.  (Actually, no. Adding a preference like that is a lot of work. If people want the activity log there is a link in the main nav for it.)

There seems to be consensus as of the last meeting that this is a good idea. It will make the home page query slightly less expensive, and quicken the many default redirects on the site. And in my opinion, getting the housekeeping out of the homepage is better UX, especially for beginners.

**Revised - see comment below**. The query used for this page is "Observations by: :rss_log", and it adds an "Activity" button to the sort tabs.

>So... for this Obs home index, I’d like to hear peoples' ideas of which query to use. We discussed Obs by RssLog. Obs by date is currently the default query for the obs index, but i prefer the idea of obs by updated_at, which is in the spirit of the the activity log but less expensive. 
>
>Question:
>- I notice that the query class `ObservationByRssLog` is different from `ObservationAll by: :rss_log`
>
>Both of these involve a join to RssLogs, so they are slower than `ObservationAll by: :updated_at`, and i believe the main difference is that the join to `:rss_logs` returns deleted observations. In my opinion, deleted observations don't need to be in the homepage feed. If somebody wants to check the log, they’ll be there.

### DIFF
--- a/app/controllers/images_controller.rb
+++ b/app/controllers/images_controller.rb
@@ -197,7 +197,7 @@ class ImagesController < ApplicationController
   # Outputs: @image
   def show
     store_location
-    return false unless (@image = find_image!)
+    return false unless find_image!
 
     case params[:flow]
     when "next"
@@ -221,7 +221,14 @@ class ImagesController < ApplicationController
   private
 
   def find_image!
-    find_or_goto_index(Image, params[:id].to_s)
+    @image = Image.includes(show_image_includes).find_by(id: params[:id]) ||
+             flash_error_and_goto_index(Image, params[:id])
+  end
+
+  def show_image_includes
+    [
+      { image_votes: :user }
+    ]
   end
 
   def set_default_size
--- a/app/controllers/observations_controller/index.rb
+++ b/app/controllers/observations_controller/index.rb
@@ -206,12 +206,13 @@ class ObservationsController
 
       # Add some alternate sorting criteria.
       links = [
-        [(query.flavor == :by_rss_log ? "rss_log" : "updated_at"),
-         :sort_by_updated_at.t],
+        ["rss_log", :sort_by_activity.t],
         ["date", :sort_by_date.t],
+        ["created_at", :sort_by_posted.t],
+        # kind of redundant to sort by rss_logs, though not strictly ===
+        # ["updated_at", :sort_by_updated_at.t],
         ["name", :sort_by_name.t],
         ["user", :sort_by_user.t],
-        ["created_at", :sort_by_posted.t],
         ["confidence", :sort_by_confidence.t],
         ["thumbnail_quality", :sort_by_thumbnail_quality.t],
         ["num_views", :sort_by_num_views.t]
--- a/app/controllers/observations_controller/index.rb
+++ b/app/controllers/observations_controller/index.rb
@@ -21,10 +21,6 @@ class ObservationsController
       show_selected_observations(query)
     end
 
-    def default_sort_order
-      ::Query::ObservationBase.default_order
-    end
-
     # Displays matrix of selected Observations (based on current Query).
     # NOTE: Why are all the :id params converted .to_s below?
     def index_query_results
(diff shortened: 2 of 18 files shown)


nimmolo:
Realization: the join to` :rss_logs` is necessary to catch observation comments, which do not trigger a new `:updated_at `date in the observation.

So observations by :rss_log it is.

There is currently a `(Observation, :by_rss_log)` query flavor, distinct from the sorting criterion `(Observation, :all, by: :rss_log)`. I believe they produce the same results.

I plan to use the sorting arg not the flavor, because I'd like all the sorts to use the same query class. Seems simpler. I believe the `:by_rss_log` query flavor is used only in coerced queries.

nimmolo:
A remaining question is what to call this index, and what to call the sort order.

Mindful that people are used to "Activity Log", i'm proposing "Observation Activity" as a temporary title of this "Observations by rss_log". "Observation Activity Log" is good too, as is "Observation Log". 

In the future, if people like the top navbar redesign idea I'm thinking of, it would just be 

"Observations" with a dropdown for the sort order, something like 

## Observations   sort by: `Recent Activity` ##

But that's for another day.

pellaea:
Any of those work for me, too.

app/controllers/images_controller.rb
```diff
@@ -221,7 +221,14 @@ def show
...
   def find_image!
-    find_or_goto_index(Image, params[:id].to_s)
+    @image = Image.includes(show_image_includes).find_by(id: params[:id]) ||
+             flash_error_and_goto_index(Image, params[:id])
+  end
+
+  def show_image_includes
+    [
+      { image_votes: :user }
+    ]
```
nimmolo, line 231: This change is because Bullet asked for the include

app/controllers/observations_controller/index.rb
```diff
@@ -206,12 +207,13 @@ def define_index_args(query, args)
 
       # Add some alternate sorting criteria.
       links = [
-        ["name", :sort_by_name.t],
+        ["rss_log", :sort_by_activity.t],
```
nimmolo, line 210: This adds the new "Activity" sort button to the sort links

app/controllers/observations_controller/index.rb
```diff
@@ -206,12 +207,13 @@ def define_index_args(query, args)
...
         ["date", :sort_by_date.t],
-        ["user", :sort_by_user.t],
         ["created_at", :sort_by_posted.t],
-        [(query.flavor == :by_rss_log ? "rss_log" : "updated_at"),
-         :sort_by_updated_at.t],
+        # kind of redundant to sort by rss_logs, though not strictly ===
+        # ["updated_at", :sort_by_updated_at.t],
+        ["name", :sort_by_name.t],
+        ["user", :sort_by_user.t],
```
nimmolo, line 216: A bit of reordering here. Comments welcome. Note that i removed "updated_at".

test/controllers/observations/images_controller_test.rb
```diff
@@ -138,7 +138,7 @@ def test_update_image
       image = images(:agaricus_campestris_image)
       obs = image.observations.first
       assert(obs)
-      assert(obs.rss_log.nil?)
+      assert_not_nil(obs.rss_log)
```
nimmolo, line 141: ... because `obs.rss_log` ain't nil anymore.
All obs fixtures now have (and must have) a corresponding rss_log fixture.

nimmolo:
re: coveralls, I don't understand why it says the species_lists/projects lines are "newly" uncovered here...
Don't believe I touched anything to do with those.

nimmolo:
Thanks @pellaea ... @mo-nathan @JoeCohen — Any preferences whether I merge this ahead of the weekend, so we can all test drive it before the meeting, or keep it open?

PR #1471, opened 2023-04-19
Cinch down the tall thumbnails a bit more

nimmolo:
This would further limit the tall images on the indexes to 4:3, from 3:2 currently. Responds to HuaFang's feedback.

The max proportion of our users' cameras is likely 16:9, so it's a significant crop from a new iPhone, but these proportions would affect matrix box thumbnails only. Lightboxes and the new Bootstrap 4 carousel on show-obs would show the uncropped image.

Here's the proportion difference. 
Yellow = newest phones
Mycena = good photographer's camera (go Alan!)
Pink = proposed proportion cap

proportions


pellaea:
Hooray!  I for one am certain that I will like this look better.  Thank
you, Nimmo. :)

nimmolo:
I'm just gonna merge this. It's a one-line change, if we don't like it we can go back.

PR #1473, opened 2023-04-21
Refactor application helper

nimmolo:
This pulls some changes from the bs4/5 branch into `main` to make the PR branch easier to comprehend for me

Helper file reorg
Some helper methods can now take a block


nimmolo:
Rubocop complaints about "booleans prefer kwargs" may be valid, but predate this PR. 
Not clear how to fix in this case. Leaving the cop on duty and ignoring.

PR #1474, opened 2023-04-22
Refactor Query Ordering

JoeCohen:
- Refactors `Query::Modules::Ordering#initialize_order_specs` to fix AbcSize offense.
(Uses GPTChat suggestion and model code snippets. See #a37a4534 for details of the suggestion.)
- Fixes remaining offenses in the file.

There is no manual test.


JoeCohen:
Turning this to a draft PR while I add some more coverage.

app/classes/query/modules/ordering.rb
```diff
@@ -1,121 +1,219 @@
...
+          )
+        end
+
+        send(sorting_method, model)
+      end
+
+      #######
+
+      private
+
```
JoeCohen, line 36: Question: Should these methods be in alphabetical order?
That would make it easier to find them in the future.
(The are currently in an order that facilitates comparison with the prior version.)

app/classes/query/modules/ordering.rb
```diff
@@ -1,121 +1,219 @@
...
-      "#{table}.#{by} DESC" if columns.include?(by)
+      def sort_by_num_views(model)
+        return unless model.column_names.include?("num_views")
+
+        "#{model.table_name}.num_views DESC"
+      end
+
+      def sort_by_date(model)
+        if model.column_names.include?("date")
+          "#{model.table_name}.date DESC"
```
JoeCohen, line 63: @pellaea, @nimmolo:
I think the first condition should be deleted because there is no table with a `date` column.
Am I missing something?

app/classes/query/modules/ordering.rb
```diff
@@ -1,121 +1,219 @@
...
-      elsif columns.include?("title")
-        "#{table}.title ASC"
       end
 
-    when "title", "login", "summary", "copyright_holder", "where",
-         "initial_det", "accession_number"
-      "#{table}.#{by} ASC" if columns.include?(by)
+      def sort_other_models_by_name(model)
+        if model.column_names.include?("sort_name")
+          "#{model.table_name}.sort_name ASC"
```
JoeCohen, line 118: @pellaea, @nimmolo 
This line is untested, and I think it's unreachable in practice because:
- There are only two tables with a `sort_name` column:
`names` and `names_version`.
- `names` is already handled by `sort_names_by_name`
-  We don't sort `names_version` by `sort_name`.

**Should I delete the first condidtion?**

app/classes/query/modules/ordering.rb
```diff
@@ -1,121 +1,219 @@
...
+
+      def sort_by_accession_number(model)
+        return unless model.column_names.include?("accession_number")
+
+        "#{model.table_name}.accession_number ASC"
+      end
+
+      def sort_by_user(model)
+        unless model.column_names.include?("user_id") || model == Herbarium
+          return
```
JoeCohen, line 164: @pellaea, @nimmolo 
This line is untested because:
The only relevant, sortable tables are `herbaria` or have `user_id` column
**Question Should I eliminate the condition? **

app/classes/query/modules/ordering.rb
nimmolo: Sounds right.
It seems like it would make sense then to also change the name of the sorting condition to `by: :when`, but it's not necessary.

app/classes/query/modules/ordering.rb
nimmolo: Don't know enough to have an opinion here.

app/classes/query/modules/ordering.rb
JoeCohen: That would work, but I think would necessitate more changes elsewhere, which I'd rather avoid in this PR.

pellaea: Yes, let's please resist making unnecessary changes to the public interface! :)

app/classes/query/modules/ordering.rb
pellaea: Sounds right, rip it out.

app/classes/query/modules/ordering.rb
pellaea: Yes.

Why is there a special exception for Herbarium any way?  It looks like it serves no purpose.  Oh I get it, this condition *should never happen*, that's exactly why it's there.  We just wanted it to give a useful error message if we ever tried to sort a table without user_id by user... or apparently we wanted to remind ourselves never to try to sort herbaria by user... for whatever reason.  Removing this condition will just make it either 1. crash if there is no such column (still an error message!) or 2. actually sort herbaria by uiser... which is bad exactly why?  I think herbaria.user is just the one who created it.  I guess that's pretty meaningless.  But harmless.

I suspect this conversation has already consumed more resources than the exception would ever have saved!! :)  Get rid of it!

pellaea:
Nice job using rails's methods as a dispatch table.  Very clean and slick.

JoeCohen:
@nimmolo & @pellaea The feedback is much appreciated. I will make the changes.

PR #1480, opened 2023-04-26
Prevent reversion of bug fixed in PR #1479

JoeCohen:
Adds an integration test to prevent reversion of the bug fixed by  #1479.
The test
-  Passes in this branch; and
-  Fails in # e39d5fa45 (just before #1479), with the same error as thrown in production (and development):
```ruby
ERROR ObservationsControllerSupplementalTest#test_edit_observation_with_query (6.37s)
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `[]=' for nil:NilClass
        
              params[:q] = query_param if query_param
                    ^^^^^^
            app/controllers/application_controller.rb:1034:in `add_query_param'
            app/controllers/application_controller.rb:1039:in `add_query_param'
            app/helpers/link_helper.rb:23:in `link_with_query'
            app/views/observations/edit.html.erb:5
            app/controllers/application_controller.rb:146:in `skip_bullet'
            app/controllers/application_controller.rb:241:in `catch_errors_and_log_request_stats'
            test/integration/capybara/observations_controller_supplemental_test.rb:65:in `test_edit_observation_with_query'
```

### DIFF
--- a/test/controller_extensions.rb
+++ b/test/controller_extensions.rb
@@ -204,7 +204,6 @@ module ControllerExtensions
   #
   def either_requires_either(method, page, altpage, params = {},
                              username = "rolf", password = "testpassword")
-    # binding.break
     assert_request(
       method: method,
       action: page,
--- a/test/controllers/observations/downloads_controller_test.rb
+++ b/test/controllers/observations/downloads_controller_test.rb
@@ -197,7 +197,6 @@ module Observations
           commit: "Print Labels"
         }
       )
-      # binding.break
       assert_flash_error
     end
   end
--- a/test/integration/capybara/observations_controller_supplemental_test.rb
+++ b/test/integration/capybara/observations_controller_supplemental_test.rb
@@ -51,6 +51,22 @@ class ObservationsControllerSupplementalTest < CapybaraIntegrationTestCase
                  "Wrong page")
   end
 
+  # Prevent reversion of the bug that was fixed in PR #1479
+  # https://github.com/MushroomObserver/mushroom-observer/pull/1479
+  def test_edit_observation_with_query
+    user = users(:dick)
+    obs = observations(:collected_at_obs)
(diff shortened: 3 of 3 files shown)


JoeCohen:
### Why didn't prior tests catch the bug, even though they visited Edit Observation?

They didn't visit that page under the conditions that provoked the bug.
```ruby
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `[]=' for nil:NilClass
        
              params[:q] = query_param if query_param
                    ^^^^^^
            app/controllers/application_controller.rb:1034:in `add_query_param'
```
In particular, the line above blows up if `query_param` is present but `params` is `nil`.

All of our pre-existing tests visited Edit Observation without a query param.

nimmolo:
Thanks for figuring this out Joe!

JoeCohen:
@nimmo:
1. I tried for a **very** long time to write a controller test that would provoke this exact bug. (I wanted a controller test because I still think our controller tests are faster than integration tests.) But I always failed. 
2. The following is probably unrelated: I noticed that `app/controllers/application_controller.rb`#`get_query_param`, and therefore `add_query_param` -- the method that blew up -- depend (conditionally) on a global variable `@query_params`.  That bothers me, but I can't articulate why.

Nimmo:
> @Nimmo:
> 
> 1. I tried for a **very** long time to write a controller test that would provoke this exact bug. (I wanted a controller test because I still think our controller tests are faster than integration tests.) But I always failed.
> 2. The following is probably unrelated: I noticed that `app/controllers/application_controller.rb`#`get_query_param`, and therefore `add_query_param` -- the method that blew up -- depend (conditionally) on a global variable `@query_params`.  That bothers me, but I can't articulate why.

You've tagged the wrong user here I think...

JoeCohen:
Thanks! so much.

PR #1483, opened 2023-04-26
`Query` class: eliminate second query for `num_results`

nimmolo:
This tiny PR may shave more than 40% off the load time of the home index.

When I paid close attention to the load of the new obs by rss_log home index in the dev environment, i found something in the log that seems off. Turns out this had been bugging @pellaea also.

There are two giant queries being executed on page load:
```
(4355.3ms)  
      SELECT COUNT(*)
      FROM `observations` JOIN `rss_logs` ON observations.rss_log_id = [rss_logs.id](http://rss_logs.id/)
    
QUERY starting: "\n      SELECT DISTINCT [observations.id](http://observations.id/)\n      FROM `observations` JOIN `rss_logs` ON observations.rss_log_id = [rss_logs.id](http://rss_logs.id/)\n      ORDER BY rss_logs.updated_at DESC, [observations.id](http://observations.id/) DESC\n"
   (5806.0ms)  
      SELECT DISTINCT [observations.id](http://observations.id/)
      FROM `observations` JOIN `rss_logs` ON observations.rss_log_id = [rss_logs.id](http://rss_logs.id/)
      ORDER BY rss_logs.updated_at DESC, [observations.id](http://observations.id/) DESC
```

The first query is just whatever the current query is, minus `group` and `order` args, and it's run by `HighLevelQueries#num_results`, to get an accurate count of the results, for pagination. It covers edge cases where a `group` arg in the query can throw off the number of results.

@pellaea suggested an alteration: just accepting the number of results returned by the main query. On a large query like the home page, any discrepancy is not going to matter, and it's hard to think of the edge cases where it will matter (and especially, matter more than the effect of basically doubling the page load time for indexes).

### DIFF
--- a/app/classes/query/modules/high_level_queries.rb
+++ b/app/classes/query/modules/high_level_queries.rb
@@ -32,23 +32,8 @@ module Query::Modules::HighLevelQueries
   INSTANTIATE_ARGS = [:include].freeze
 
   # Number of results the query returns.
-  def num_results(args = {})
-    @num_results ||=
-      if @result_ids
-        @result_ids.count
-      else
-        # Explicitly disable GROUP BY and ORDER clauses for the purposes of
-        # simply counting the number of results.  This is important because
-        # GROUP BY in particular will mess up the COUNT(*) spec.  Passing in
-        # empty strings for group and order will tell it to explicitly override
-        # anything in self.group and self.order.
-        rows = select_rows(args.merge(select: "COUNT(*)", group: "", order: ""))
-        begin
-          rows[0][0].to_i
-        rescue StandardError
-          0
-        end
-      end
+  def num_results(_args = {})
+    @num_results ||= result_ids&.count || 0
   end
 
   # Array of all results, just ids.


pellaea:
That looks right.  Fingers crossed...

nimmolo:
Searching through the code, i'm not sure we have a single index query that uses a `group` arg outside of scripts and migrations. `site_data` and `report` do, but i don't believe they're ever used in a paginated context.

pellaea:
I don't think it matters.  GROUP can't be used with COUNT.  But we aren't
using COUNT anymore, so GROUP is fine.

nimmolo:
Merrging!

nimmolo:
Deployed. Idk where we access those page load statistics, but i'll be curious.

JoeCohen:
Try here: https://mushroomobserver.org/stats/index.html
Then click on a graph for more history.

nimmolo:
Thanks Joe. I wouldn't say it moved an inch.

nimmolo:
Current time is far left of the graph? 4am GMT? 

Noffing! LOL! It did speed up my load time in dev, though. I'm baffled!

image

nimmolo:
Also, it did in fact deploy the change. I checked the log of the deploy script.
```
From github.com:MushroomObserver/mushroom-observer
   59e707f18..873667abd  main                    -> origin/main
 * [new branch]          jdc-prevent-1479-reversion -> origin/jdc-prevent-1479-reversion
   c62c93a75..5ad5dcd4a  nimmo-bootstrap-4-reboot -> origin/nimmo-bootstrap-4-reboot
 * [new branch]          nimmo-query-num-results -> origin/nimmo-query-num-results
 * [new branch]          nimmo-sorting-links-fix -> origin/nimmo-sorting-links-fix
Updating 59e707f18..873667abd
Fast-forward
 app/classes/query/modules/high_level_queries.rb       | 19 ++-----------------
 app/controllers/application_controller.rb             |  2 +-
 app/helpers/application_helper.rb                     |  6 ++++++
 app/views/layouts/content/_sorter.html.erb            | 16 ++++++++++++++++
 .../layouts/content/_title_and_tab_sets.html.erb      | 18 +-----------------
```

pellaea:
Yeah, I just double checked everything, too.  It is definitely not
executing the query twice.  But as you can see in the graph, which averages
real load times over each hour for the last 27 or so hours based on the
production logs, there has been no change whatsoever.  Are we actually
certain that it ever executed the query twice?

pellaea:
Sure looks like it should have been doing the query twice.  Beats me.  Oh
well.

nimmolo:
I'm not sure about production, but i'm sure about dev doing a double query.

Maybe tomorrow it will show a diff.  I believe it might be loading a bit quicker for me right now. ~6s vs ~10s

nimmolo:
The weekly graph shows a significant and stable drop, back on saturday when we switched homepages. Nothing since.



But the units are also kind of baffling. 1100ms? I don't think so! More like 11000ms.

PR #1488, opened 2023-04-30
Cache `rss_logs.updated_at` on `observations` table to eliminate join on home page

nimmolo:
Adds new column `observations.log_updated_at`, to cache the value of the obs's corresponding `rss_log.updated_at`. Migration populates it with correct value, or if `rss_log` is missing, it leaves the column `null`. 

(An earlier idea of populating missing values with the obs' own `updated_at`, and making the column non-nullable, tabled for now. Many obs with missing logs are draft obs from the name_lister; others are very early days of MO. We may want to treat these differently.)

Changes the new home page, obs by rss log index, to order by the cached column, so it doesn't have to join to `rss_logs` just to find out the correct order.

#### Code note: ####
Alters `AbstractModel#log`, to call an MO method `touch_when_logging` instead of AR `touch`, which can then be defined at the model (default is `touch`). Allows any model in the future to piggyback on `touch` in this way and cache a value, or do anything else, on a log event.

### DIFF
--- a/app/classes/query/modules/ordering.rb
+++ b/app/classes/query/modules/ordering.rb
@@ -193,8 +193,14 @@ module Query
       def sort_by_rss_log(model)
         return unless model.column_names.include?("rss_log_id")
 
-        add_join(:rss_logs)
-        "rss_logs.updated_at DESC"
+        # use cached column if exists, and don't join
+        # calling index method should include rss_logs
+        if model.column_names.include?("log_updated_at")
+          "#{model.table_name}.log_updated_at DESC"
+        else
+          add_join(:rss_logs)
+          "rss_logs.updated_at DESC"
+        end
       end
 
       def sort_by_summary(model)
--- a/app/models/abstract_model.rb
+++ b/app/models/abstract_model.rb
@@ -659,9 +659,16 @@ class AbstractModel < ApplicationRecord
   #
   def log(tag, args = {})
     init_rss_log unless rss_log
-    touch unless new_record? ||
-                 args[:touch] == false
+    touch_when_logging unless new_record? ||
+                              args[:touch] == false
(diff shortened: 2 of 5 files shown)


nimmolo:
The error is about the column not being nullable, but also not having a default value. 

Should I remove the `null: false` or add a default?

app/classes/query/modules/ordering.rb
```diff
@@ -193,8 +193,14 @@ def sort_by_records(_model)
...
-        "rss_logs.updated_at DESC"
+        # use cached column if exists, and don't join
+        # calling index method should include rss_logs
+        if model.column_names.include?("log_updated_at")
+          "#{model.table_name}.log_updated_at DESC"
+        else
+          add_join(:rss_logs)
+          "rss_logs.updated_at DESC"
+        end
       end
```
pellaea, line 204: I like this!

JoeCohen:
I think `observations.log_updated_at` is not updated via an AR callback per se,  
but rather via `app/controllers/observations_controller/new_and_create.rb#create` line 114:, see below
```ruby
  def create
    ...
    # Once observation is saved we can save everything else.
    if success
      @observation.log(:log_observation_created)
```

`observations.log_updated_at` does get set when I run the api tests (perhaps because it's a simulated, rather than real POST??).
(Example: `rails t test/controllers/api2_controller_test.rb -n test_post_minimal_observation`)

The user (gillyj, User 22647) says 
>All my new observations have been uploaded from the MO mobile app
https://mail.google.com/mail/u/0/#inbox/FMfcgzGsmqzscjjrTRDcxvzfklMdfvKq

This is probably unrelated, but just in case: gillyj's recent Observations list the source as "mo api" rather than "mo android app" or "mo iphone app". Example: https://mushroomobserver.org/520284

PR #1489, opened 2023-04-30
Should `SpeciesList#name_lister` log "draft" observations?

nimmolo:
`SpeciesList` has not been logging the observations it creates (via `name_lister`). This seems to account for many of the 25,000 observations in the db without a corresponding `rss_log`. (Others are from early days of MO.)

This PR should also probably
- create an `RssLog` for all the observations that didn't get logged so far, because the existence of an rss_log governs them showing up on the home page
- add a test that new obs from name_lister are getting a log. This seems tricky, because the name_lister controller test doesn't actually create any obs (yet).

### DIFF
--- a/app/models/species_list.rb
+++ b/app/models/species_list.rb
@@ -355,6 +355,8 @@ class SpeciesList < AbstractModel
       is_collection_location: args[:is_collection_location],
       specimen: args[:specimen]
     )
+    obs.log(:log_observation_created)
+
     args[:projects].each do |project|
       project.add_observation(obs)
     end


pellaea:
I think we might want to discuss this more before we "fix" this.  Can we work through what this will look like?  
What will happen if someone creates a species list for a foray and records, say, 50 names without any photos or other data?  It will create 50 "empty" observations.  Do we want those observations to completely swamp the activity log?  It's a fair question to ask if we even want those observations at all.  The answer has historically been "sure, why not?"  But clearly we also decided not to let those empty observations overwhelm the activity log.  That was before EXD came around and proved that an individual user is perfectly capable of flooding the activity log with entirely meaningless logs without resorting to any questionable tools like the Name Lister.  But the point still stands.  Why make it easy to do that?

These are the sorts of changes that should probably be discussed at a tech meeting.

pellaea:
By the way, what if that user that created the 50 empty observations then proceeds to add observations and notes to a few of those observations?  Hopefully that will create an rss_log for those select observations -- the observations that are no longer "empty" observations.  These detailed observations *so* belong on the activity log... and I believe they probably already would end up there with the code as it is.

nimmolo:
This is key. Lets discuss. It is nice that "draft" observations dont get a log, in a way. Let them stay in limbo until they get fleshed out better!On Apr 30, 2023, at 7:13 PM, Jason Hollinger ***@***.***> wrote:﻿
By the way, what if that user that created the 50 empty observations then proceeds to add observations and notes to a few of those observations?  Hopefully that will create an rss_log for those select observations -- the observations that are no longer "empty" observations.  These detailed observations so belong on the activity log... and I believe they probably already would end up there with the code as it is.

pellaea:
Agreed.

PR #1498, opened 2023-05-05
Refactor Lookups Controller

JoeCohen:
- Reduces AbcSize of `LookupsController#lookups_general` from 82.13 to <= 17 (the current default).
- Fixes other RuboCop offenses in the controller.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -635,7 +635,6 @@ Style/OptionalBooleanParameter:
     - 'app/classes/query/modules/sql.rb'
     - 'app/classes/textile.rb'
     - 'app/controllers/application_controller.rb'
-    - 'app/controllers/lookups_controller.rb'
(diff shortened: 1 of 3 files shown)


JoeCohen:
@nimmolo, @jason
I'd like to merge this in the next day or two unless there are objections.

nimmolo:
Thanks Joe. I'm not sure I entirely get what's going on in this controller, but the refactor seems good fwiw. 

I'm guessing it has to do with links to objects within Textile such as comments and descriptions, and not instantiating the referred objects.

JoeCohen:
Nimmo:
Thanks for the feedback. You are correct.
It also allows external links that are independent of the object id. (The
object id can change, at least for Names and Locations, at least in the
event of merger.)
- Joe

pellaea:
This looks good to me, too.  Thanks for cleaning it up, Joe.

PR #1503, opened 2023-05-06
Fix "previous version" links in `versions_helper.rb`

nimmolo:
I refactored `link_with_query` a couple weeks ago to take the same args as `link_to(text, path, **html_options, &block)`, and it now requires something like one of the following for the path arg:
 - `name_path(obj.id)`
 - `obj.show_link_args`
 - `{ controller: :obj.show_controller, action: obj.show_action, id: obj.id }`

In other words, you can't pass `controller` and `action` as separate `kwargs` anymore, just have to enclose them in a hash. These helpers seem to be the last that I missed converting, of the 201 uses of `link_with_query`.


JoeCohen:
Thanks for doing this!
I tried the functionality in question and it's fixed. I didn't poke around
more or look at the failing test.

PR #1505, opened 2023-05-07
Redirect legacy name_search

JoeCohen:
- Redirects legacy `/name/name_search` actions to `/names`
iNat is using the old  url in order to link to MO Name pages from INat name pages.
As with #1499, it costs epsilon to make the old url work. 
- Delivers https://www.pivotaltracker.com/story/show/185114318

=== Manual test
- http://localhost:3000/name/name_search?pattern=Tremella%20mesenterica
Result: http://localhost:3000/names?pattern=Tremella+mesenterica
Names Matching ‘Tremella mesenterica’  (in any field)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -922,6 +922,7 @@ MushroomObserver::Application.routes.draw do # rubocop:todo Metrics/BlockLength
 
   # ----- Names: legacy action redirects -----------------------------------
   get("name/eol", to: redirect("names/eol_data#show"))
+  get("name/name_search", to: redirect(path: "names"))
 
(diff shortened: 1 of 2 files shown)


JoeCohen:
If I can find time, I will also try to fix this (or at least suggest a fix) on the iNat end. 
- `htttps` for all links to MO
- update the `name_search` url, BUT
- they should probably be using `lookup` instead of `name_search`

JoeCohen:
I reported this to iNat and suggested that they use a lookup instead of a pattern search.
https://forum.inaturalist.org/t/taxa-broken-link-out-to-mushroom-observer/41537

PR #1511, opened 2023-05-20
Fix truncated Symbiota notes

JoeCohen:
- Deletes consecutive newlines before textilizing.
- Includes test.
- Delivers  [Pivotal #185214534](https://www.pivotaltracker.com/story/show/185214534).

### Manual Test
- https://mushroomobserver.org/observations?pattern=user%3A2873+notes%3AUBC+date%3A2007-11-10
- click Download Observations
- select Symbiota
- click Dowload
- save the resulting tsv file
- open that file
Result: the fieldNotes field for row 2 should be 
>Original Herbarium Label: Lepiota pseudofelina J.E. Lange as in Moser & Breitenbach et al. Herbarium Specimen: UBC F29500 

(In the main branch, the fieldNotes field (= MO Notes) of row 2 (= https://mushroomobserver.org/65468) is truncated.
See [Pivotal #185214534](https://www.pivotaltracker.com/story/show/185214534) for details.

### DIFF
--- a/app/classes/report/symbiota.rb
+++ b/app/classes/report/symbiota.rb
@@ -92,7 +92,10 @@ module Report
     end
 
     def clean_notes(str)
-      str.strip.t.html_to_ascii.gsub(/\s+/, " ")
+      str.strip.
+        # Compress conssecutive newlines because they confuse Textile
+        gsub(/(\r|\n)+/, "\n").
+        t.html_to_ascii.gsub(/\s+/, " ")
     end
 
     def image_urls(row)
--- a/test/models/report_test.rb
+++ b/test/models/report_test.rb
@@ -345,6 +345,52 @@ class ReportTest < UnitTestCase
     do_tsv_test(Report::Symbiota, obs, expect, &:id)
   end
 
+  def test_symbiota_compress_consecutive_newlines
+    obs = observations(:detailed_unknown_obs)
+    obs.notes = {
+      Substrate: "wood\tchips",
+      Habitat: "lawn",
+      Host: "_Agaricus_",
+      Other: "First line.\r\n\r\nSecond line."
+    }
+    obs.save!
+
+    img1 = images(:in_situ_image)
+    img2 = images(:turned_over_image)
+    expect = [
+      "Fungi",
+      "",
+      "Kingdom",
+      "Fungi",
+      "",
+      "",
+      "Mary Newbie",
+      "174",
+      "NY",
+      "2006-05-11",
+      "2006",
+      "5",
+      "11",
+      "USA",
+      "California",
+      "",
+      "Burbank",
+      "34.185",
+      "-118.33",
+      "148",
+      "294",
+      "#{obs.updated_at.api_time} UTC",
+      "wood chips",
+      "Agaricus",
+      "Habitat: lawn Other: First line. Second line.",
+      obs.id.to_s,
+      "https://mushroomobserver.org/#{obs.id}",
+      "https://mushroomobserver.org/images/orig/#{img1.id}.jpg " \
+        "https://mushroomobserver.org/images/orig/#{img2.id}.jpg"
+    ]
+    do_tsv_test(Report::Symbiota, obs, expect, &:id)
+  end
+
   def test_rounding_of_latitudes_etc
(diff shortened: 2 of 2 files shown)


JoeCohen:
This PR solves one piece of a cascade of issues that originate with user Adolf Ceska.
In some of his Observations, he used consecutive newlines in the Notes. 
Example: https://mushroomobserver.org/65468

On Thu 18 May, Dr. Mary Berbee of University of British Columbia herbarium asked us for the Ceska MO records that have UBC accession numbers (FXXXXXX) with their corresponding MO metadata and photographs in a format that would allow us uploading them to a ... Symbiota database. See [Berbee email chain.pdf](https://github.com/MushroomObserver/mushroom-observer/files/11522732/Berbee.email.chain.pdf).
It's easy to locate those records with an Observation Search. https://mushroomobserver.org/observations?pattern=user%3A2873+notes%3AUBC. It's also easy to export the results as a Symbiota file.

The problem is that the records with consecutive newlines in the Notes do not export correctly -- the fieldNotes section is  truncated at the double newline. And that's especially bad for UBC because it drops the accession number.

I also noticed that those notes _appear_ to be truncated in other forms of export (e.g., csv). In fact, the notes are not truncated for those other exports. But most (all?) csv parsers will drop the notes when viewing the exported file.

This PR fixes the problem just for Symbiota exports.

### Questions

- Should we instead somehow attempt to quote the consecutive newlines for Symbiota exports? IMO: No.
(I actually tried to do this but couldn't figure it out.)
- Should we alternatively/additionally edit the Ceska Observations that have consecutive newlines to turn them into a single `\r\n`.  (IMO probably yes, but I think it's non-trivial because of how Notes are stored in the db.)
- Should we delete consecutive newlines in other csv exports? (I think that should be in a different PR.) (IMO yes.)

pellaea:
Hi Joe, sorry I missed the meeting today.  Yes, this newlines inside fields in CSV files (or Symbiota files or whatever) is a recurring problem I keep running into.

In my opinion -- just my opinion -- the safest solution is to convert the newlines to something else with a search-and-replace in the SELECT.  That is, something like this:

```
SELECT REPLACE(notes, "\n", "\\n") FROM ...
```

I don't know what that would look like in AR.  Or maybe there's a mechanism in the observation download class that allows for some post-processing after the SELECT returns.  If not, we could add it.

Yet another option is to convert it to HTML.  I remember doing that somewhere in some export.  We can't really expect most third parties to understand Textile, and even less our peculiar extended version of it!  So it kinda actually makes sense to do so... only, the problem is that will probably kill the server.  Imagine downloading all observations in the database.  We can actually support that presently... albeit I'd prefer if users did not start doing it regularly!  But imagine now also running Textile on the notes for every single observation.  Yikes!  Oh, I know where I did the HTML conversion: I bet the API processes Textile before serving responses.  That's okay because the API has the ability to limit resource usage.  We don't currently limit the size of downloads.  (And personally I'd prefer not doing so, otherwise we'd have to add the ability to paginate the download, etc.  Much less user-friendly.  Or we could turn it into an asynchronous process -- not a bad idea, and maybe there are gems to help out with that sort of thing, but again a nontrivial development.)

Anyway, the point is, in my experience it is a coin flip whether or not any given spreadsheet reader will be able to understand newlines inside quoted values in CSV files.  So one way or another they need to be encoded to ensure the download is stable and universally accepted.

JoeCohen:
@pellaea: Sorry I didn't wait for your response before deploying this PR.
(I did the simplest thing I could think of to make the Symbiota export work for UBC.)
I'll review your response to see if I can find an improved solution.

JoeCohen:
The Symbiota export does custom post-processing on our Notes field. 
The key part is 
```ruby
def clean_notes(str)
  str.strip.
    t.html_to_ascii.
    gsub(/\s+/, " ")
end
```
There's a problem in that Textile can get confused by some combinations of whitespaces.
Examples: `\r\n\r\n` `\r\n \r\n`
My solution (fully implemented in #1512, is simply to do the gsub before Textilizing the string:
```ruby
def clean_notes(str)
  str.strip.
    gsub(/\s+/, " ").
    t.html_to_ascii
end
```

pellaea:
If it works!...

PR #1519, opened 2023-05-31
Observation log_updated_at

### LINKED ISSUE #1518: User gillyj's Observations missing from main page
>[MO] Question From gillyj@yahoo.com
webmaster mushroomobserver.org <webmaster@mushroomobserver.org>
To: gillyj@yahoo.com
Bcc: Jason Hollinger <pellaea@gmail.com>, Andrew Nimmo <andrnimm@fastmail.fm>
Tue, May 30, 2023 at 8:18 AM
Jeff:
Thanks for letting us know. It looks like a new bug, rather than something relating to your settings. (I also don't see your new Observations when I click the MO logo.)
I will look into it later today. (I'm not in a good spot right now,)
Regards,
-- Joe (member MO volunteer Development Team)
On Tue, May 30, 2023 at 7:45 AM <webmaster@mushroomobserver.org> wrote:
I've noticed my new observations do not show up under the "Activity" filter option on the web site, however, in the past they would. Typically after uploading an observation, I just click the MO logo in the upper left corner on the screen which defaults to "activity" to check if my observation uploaded correctly. I'm wondering if this is a bug, something I did with my default settings or was done intentionally for a reason?
I'm using Windows 10 Pro, and Firefox 113.0.1 (64bit).

This is user 22647, Jeff (gillyj).
I don't see his recent (or maybe any) of his Obs's on the main page.

JoeCohen:
- Insures that `observation.log_updated_at` is saved
- Fixes #1518, https://www.pivotaltracker.com/story/show/185297843

### Manual Test

- Create an Observation with a valid Location but no Scientific Name
- Goto the home page, http://localhost:3000/
Result: the Observation you just created should be the first one displayed in the Observation Activity feed

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -587,7 +587,7 @@ class Observation < AbstractModel
   # to cache a log_updated_at value
   def touch_when_logging
     self.log_updated_at = Time.zone.now
-    touch
+    save
   end
 
   ##############################################################################
--- a/test/controllers/api2_controller_test.rb
+++ b/test/controllers/api2_controller_test.rb
@@ -155,6 +155,8 @@ class API2ControllerTest < FunctionalTestCase
     assert_equal(false, obs.specimen)
     assert_equal(true, obs.is_collection_location)
     assert_equal(Observation.no_notes, obs.notes)
+    assert(obs.log_updated_at.is_a?(ActiveSupport::TimeWithZone),
+           "Observation should have log_updated_at time")
     assert_obj_arrays_equal([], obs.images)
     assert_nil(obs.thumb_image)
     assert_obj_arrays_equal([], obs.projects)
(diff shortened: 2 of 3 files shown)


pellaea:
This still confuses me a bit, but I've convinced myself your solution is correct.

The thing that bothered me is that `save` will not actually do anything if there are no changes to the record.  On the other hand `touch`, while it will definitely update the `updated_at` attribute, apparently doesn't actually save any other outstanding changes to the record (bizarre).  But the line `self.log_updated_at = Time.zone.now` *should* guarantee that there is a change in the record that needs to be saved, and that will trigger `save` to update `updated_at` automagically and save the changes.  The only other difference is that `touch` doesn't trigger some callbacks (like `before_save`) but `save` does.  Is this important??  I think this will cause `cache_content_filter_data` to be triggered unnecessarily.  But no harm done. ...  Well, it looks to me like the way I wrote `cache_content_filter_data` it will unnecessarily load the associated name and location even though they haven't changed.  Arg.  All I needed to do was check `name_id` and `location_id` instead of `name` and `location`, sigh.  Oh well.  Sorry, not your problem.  Carry on...

JoeCohen:
Good points!
I need to think some more about the extra callback, especially the possibility of an extra email.

JoeCohen:
I will merge it. The only `save` callback -- as I'm sure you knew -- is `cache_content_filter_data`.

PR #1523, opened 2023-06-04
Update signup email info

JoeCohen:
Emphasizes that blocked addresses and one-time addresses will not work.


pellaea:
Looks good.  Thanks,  Joe!

JoeCohen:
Thanks for checking!

PR #1525, opened 2023-06-05
Brakeman 6.0

JoeCohen:
- Updates brakeman gem to  6.0
- Enables Codeclimate brakeman plugin. 
This is not ideal because the the plugin uses v 5.3.1. 
But it's better than nothing; I can't figure out how to make the Github Actions brakeman runner work with v 6.0.


JoeCohen:
The 16 "new" codeclimate issues are actually old issue that are re-reported because brakeman has been restarted in CI.

PR #1526, opened 2023-06-11
Brakeman false positive

JoeCohen:
- Refactors PreferenceController to fix Brakeman false positive.
- Improves PreferenceController test coverage.
- Fixes PreferenceController#no_email AbcSize offense.
- Updates `brakeman.ignore`.

There is no manual test.

### DIFF
--- a/app/controllers/account/preferences_controller.rb
+++ b/app/controllers/account/preferences_controller.rb
@@ -116,7 +116,11 @@ module Account
     def prefs_changed_successfully
       result = false
       if !@user.changed
-         flash_notice(:runtime_no_changes.t)
+        # NOTE: The next line appears to be unreachable
+        # because @user.changed is always truthy. (It's at least `[]`.)
+        # Perhaps `!@user.changed?` was intended, but it breaks tests.
+        # 2023-06-11 JDC
+        flash_notice(:runtime_no_changes.t)
       elsif !@user.errors.empty? || !@user.save
         flash_object_errors(@user)
       else
--- a/app/controllers/account/preferences_controller.rb
+++ b/app/controllers/account/preferences_controller.rb
@@ -27,9 +27,7 @@ module Account
       user = User.safe_find(params[:id])
       return redirect_to("/") unless permitted_user_with_valid_email_type?(user)
 
-      prefix  = "no_email_#{email_type}"
-      success = "#{prefix}_success".to_sym
-      @note   = "#{prefix}_note".to_sym
+      @note = "#{email_msg_prefix}_note".to_sym
       @user.send(email_type_setter, false)
       if @user.save
         flash_notice(success.t(name: @user.unique_text_name))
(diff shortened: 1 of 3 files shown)


app/controllers/account/preferences_controller.rb
```diff
@@ -122,6 +116,10 @@ def update_copyright_holder(legal_name_change = nil)
     def prefs_changed_successfully
       result = false
       if !@user.changed
+        # NOTE: The next line appears to be unreachable
+        # because @user.changed is always truthy. (It's at least `[]`.)
+        # Perhaps `!@user.changed?` was intended, but it breaks tests.
+        # 2023-06-11 JDC
```
JoeCohen, line 122: @nimmolo: I tried to fix this but got (easily) confused. 
Any hints for how to about this. (It's something for another PR.)

@nimmolo, @pellaea:
I was also confused by the code around the other uncovered line. See lines 110-114.
I didn't understand why copyright holder is updated in PreferencesController. 
It doesn't seem to be something that users can change here (as opposed to AccountsController).
```ruby
    def update_copyright_holder(legal_name_change = nil)
      return unless legal_name_change

      Image.update_copyright_holder(*legal_name_change, @user)
    end
```

nimmolo: @JoeCohen - No idea about `changed` vs `changed?`. What tests does it break? (Maybe the tests need fixing!)
Ditto for the `update_copyright_holder` question... I'll try to spin it up as soon as i get Vagrant working with Code again.

I think my "creativity" in this file was probably limited to breaking up the controller. I don't recall trying to change any of this functionality.

pellaea:
Oops!  That's probably my fault.  Weird that "fixing" it would break
tests.  I wouldn't sweat it, though.  I think it'd be okay just to remove
that clause.  Who cares if it says "successfully updated preferences" even
if you submitted the form with no changes?
-Jason

app/controllers/account/preferences_controller.rb
nimmolo: @JoeCohen - It looks like `def update_copyright_holder` is only here because I accidentally left it here when splitting the file! Note that it's also defined in `Account::ProfileController` in exactly the same way. `legal_name_change` happens in `profile` not `preferences`.

`PreferencesController`'s line 20, within `def update`:  `update_copyright_holder(@user.legal_name_change)` is repeated in `ProfileController` line 91, within `def deal_with_possible_profile_changes`. 

In summary, I think you can safely delete anything to do with copyright holder name changes in the `PreferencesController` and its tests.

JoeCohen: @nimmolo: For another PR
`if @user.changed` is always truthy and therefore not useful as a test, so it should be something else perhaps `if !@user.changed?`
When I try `if !@user.changed?`, it breaks part of  Capybara `AccountTest#test_edit_preferences`,
in particular the stuff relating to change of email address.

JoeCohen:
I think we can fix it, but don't want to do it in this PR.
So I'll leave it as-is (including the comment as a reminder).

PR #1528, opened 2023-06-18
Glossary textile

JoeCohen:
- Allows textiled links to Glossary Terms.
These links can take two forms:
  - The unambiguous `_term xxx_` or  `_glossary_term xxx_` (where `xxx` is the integer id of the Term or a string exactly matching the name of the Term) or
  - The ambiguous `xxx`:
    - If `xxx` can be interpreted as a Name, it looks up Names and --if  there are no matches or suggestions -- tries to find a matching Term. 
    - else looks only for Terms. 
    **NOTE**: When `xxx` is all lower case, it looks up only Terms (because all lower case cannot be interpreted as a Name.
- Bonus: Fixes textiled links to SpeciesLists (`spl xxx` or `SpeciesList xxx`).
- Bonus: Fixes a bunch of Rubocop offenses in the touched files

### Manual Test
Paste this into the Textile Sandbox, http://localhost:3000/info/textile_sandbox, then Test and follow the links:
```
| +Textile+ | +rendered+ | +notes+ | +lookup+ | +result when followed+ |
| <code>_Abies_</code> | _Abies_ | ambiguous string which exactly matches a Name. | Name | **Name** ??Abies?? |
| <code>_Aborts_</code> | _Aborts_ | ambiguous string which suggests a Name. | Name | **Name** ??Abies?? with the warning "No names match “Aborts”, maybe you meant this?" |
| <code>_Adnexed_</code> | _Adnexed_ | ambiguous string without Name matches or suggestions, but matching a Glossary Term (exactly). | Name | **Term** Adnexed |
| <code>_term Aborts_</code> | _term Aborts_ | Manually disambiguated link to the Term Aborts. | Name| **Term** Aborts |
| <code>_aborts_</code> | _aborts_ | Automagically disambiguated link to the Term Aborts (It can't be a Name because it's one lower case word.) | Term | **Term** Aborts |
| <code>_term Abies_</code> | _term Abies_ | disambiguated Term link with no matches. | Term | Glossary Index |
| <code>_Howdy doody_</code> | _Howdy doody_ | ambiguous string without Name matches or suggestions and without Term matches. | Name | Glossary Index with 2 error messages: "Can’t find name matching ‘Howdy doody’. Can’t find glossary term matching ‘Howdy doody’." |
| <code>_Howdy Doody_</code> | _Howdy Doody_ | string that can't be a Name, and without Term matches. | Name | Glossary Index with 1 error message: "Can’t find glossary term matching ‘Howdy doody’." |
| <code>_species_list 1_</code> | _species_list 1_ | Compare to the same Textile in the main branch, which renders like <code>{SPECIES_LIST spl 1 }{ 1 }x</code> | List | Species List 1 |
| <code>_Round To Angular Pores_</code> | _Round To Angular Pores_ | more than 3 words; renders as plain, unlinked italics | n.a | n.a. |

```


JoeCohen:
Not ready for prime time (though passing). I noticed some pre-existing issues I'd like to fix.

JoeCohen:
@nwilson-eol, @pellaea 
1. I know you don't have time to review code, but comments on functionality would be welcome. In particular: 
>Does the [Manual Test](https://github.com/MushroomObserver/mushroom-observer/pull/1528#issue-1762324089) described above implement the functionality you had in mind at the Sat. Tech. Meeting?

3. @pellaea. Answer this only if you know the answer off the top of your head. (Otherwise it's not worth your time.)
> What's the function (or and example) of the "prefix" in Textile. See e.g., `textile.rb#convert_other_links_to_tagged_objects! `and `OTHER_LINK_PATTERN`.

pellaea:
Yes!  This sounds awesome!

re: prefix -- This preserves the non-alphanumeric character preceding the
`_link_` expression.  If the link occurs at the beginning of the string
(extremely unlikely!), then it wouldn't be required.  There is probably
also a way to do a "zero width" assertion in the regex that doesn't
actually consume the breaking character before the expression.  I used one
at the end -- `(?!\w)`.  That means "... and not followed by an
alphanumeric", a zero-width negative look-ahead assertion.  But I never
remember how to do the equivalent for *before* the expression.  Is it
called "zero-width look-behind" or something like that?  So instead, I just
let it actually match the character preceding the expression and ensure
that it gets reinserted into the output.  Cheesy but tried-and-true.  The
only drawback is that it actually consumes that character, so you can't
have the expressions you want to match back-to-back -- there needs to be at
least one non-expression character between.  But that is fine in this
particular case.  You can't have `_link1__link2_`.  That wouldn't make
sense.  Instead there needs to be a space or comma or something in between:
`_link1_ _link2_`  In this case "prefix" matches the space between the two
links.  Does that make sense?

app/classes/textile.rb
```diff
@@ -101,13 +102,14 @@ def textilize_without_paragraph(do_object_links = false, sanitize = true)
   # tl::   Do 't' and check for links.
   # tp::   Wrap 't' in a <p> block.
   # tpl::  Wrap 't' in a <p> block AND do links.
-  def textilize(do_object_links = false, sanitize = true)
+  def textilize(do_object_links: false, sanitize: true)
     # This converts the "_object blah_" constructs into "x{OBJECT id label}x".
     # (The "x"s prevent Textile from interpreting the curlies as style info.)
     if do_object_links
-      check_name_links!
```
pellaea, line 108: Thanks for improving these names!

app/classes/textile.rb
```diff
@@ -207,25 +209,30 @@ def self.clear_textile_cache
...
-    (^|\W) (?:\**_+) ([^_]+) (?:_+\**) (?= (?:s|ish|like)? (?:\W|\Z) )
-  /x
-  OTHER_LINK_PATTERN = /
-   (^|\W) (?:_+) ([a-zA-Z]+) \s+ ([^_\s](?:[^_\n]+[^_\s])?) (?:_+) (?!\w)
+    (?<prefix>^|\W)
+    (?:\**_+)
+    (?<formatted_label>[^_]+)
+    (?:_+\**)
+    (?= (?:s|ish|like)?
+    (?:\W|\Z) )
```
pellaea, line 217: I would keep this zero-width, just to be safe.  I think you might be able to test this case with the string `_Agaricus_ _Amanita_` with just a single space between.  I suspect it might not "see" the second link unless you keep this "postfix" zero-width (`?=` instead of `?:`).

app/classes/textile.rb
```diff
@@ -362,6 +383,23 @@ def pre_existing_links_replaced_by_placeholders!(do_object_links)
     saved_links
   end
 
+  IMPLICIT_TERM_PATTERN = /
+    (?<prefix> ^|\W) # prefix
+    (?:_+)
+    (?<id> [\p{Alpha}\-.\ ]+) # alpha chrs, hyphens, periods, spaces
```
pellaea, line 389: Do we care about diacritics?  Does the `{Alpha}` class already handle them?

test/controllers/lookups_controller_test.rb
```diff
@@ -78,7 +80,10 @@ def test_lookup_name
     assert_redirected_to(name_path(names(:agaricus_campestris).id))
 
     get(:lookup_name, params: { id: "Agaricus newname" })
-    assert_redirected_to(names_path)
+    assert_redirected_to(
+      glossary_terms_path,
+      "It should go to Glossary index when there's no Name or Glossary match"
+    )
```
pellaea, line 86: Agreed.  Glossary index is probably cheaper(?) and it will have the effect of "advertising" this new feature whenever someone clicks on a link that doesn't match either.

pellaea (approved):
Thanks for doing this, Joe.

JoeCohen:
@pellaea Thanks very much for looking at this (in more detail than I expected).

app/classes/textile.rb
JoeCohen: 1. \p{Alpha} does handle diacritics.
```ruby
irb(main):002:0> /\p{Alpha}+/ =~ "olé"
=> 0
irb(main):003:0> $LAST_MATCH_INFO
=> #<MatchData "olé">
```
2. Do we care about them? 
It's easy enough to handle them. And I can think of at least potential terms than include them.
Ex: Köhler illumination

JoeCohen:
@pellaea And thanks for the extended explanation of `prefix`. 
For the reasons you suggested, I'm not going to monkey with the regexp's.

JoeCohen:
@pellaea, @nwilson-eol 

### TLDR: Is this UI change okay?
In order to have a simpler UI for textilized glossary terms, `_xxx_` will always lookup a Term if `xxx` is all lower case.
The hitch is this slight change in appearance:
Screenshot 2023-06-25 at 1 41 47 PM
and the fact that it's now an (italicized) link instead of plain italics.

PS: I will add some more questions about specific parts of the code.

---
(Long story, in case you're interested.

Changing only `lookup_name` to link to GlossaryTerm (as discussed in the Tech Meeting) didn't work as well as I hoped. 
Issue:`lookup_name` finds close (as opposed to exact) Name hits. So it redirects to a Name or a Name Index more often than expected. Ex:  https://mushroomobserver.org/lookups/lookup_name/Abort.)

Gemfile
```diff
@@ -102,7 +102,8 @@ gem("unicorn")
 # Use i18n for internationalization
 gem("i18n")
 
-# Enable Textile markup language. See http://redcloth.org/
+# Enable Textile markup language. See https://github.com/jgarber/redcloth,
```
JoeCohen, line 105: Redcloth org is an expired domain. http://redcloth.org/
I wonder if the RedCloth gem is compatible with changes to the Textile standard
(the last change to the gem was over 5 years ago), 
and what might happen going forward.

app/classes/textile.rb
```diff
@@ -301,36 +315,59 @@ def strip_out_sp_cfr_and_sensu(str)
...
+    (?: _+)
+    (?<type>
+      [a-zA-Z]+ # model name
+      (?: _[a-zA-Z]+)? # optionally including underscores
+    )
+    \s+
+    (?<id> [^_\s](?:[^_\n]+[^_\s])?) # id -- integer or string
+    (?: _+)
+
+    (?! (?: \w|<\/[a-z]+>)) # discard if followed by word char or html close tag
```
JoeCohen, line 331: @pellaea, @mo-nathan 
### Is there a better solution?
This line prevents Textile from munging html tags.
Although I don't do it often, I especially wanted to be able to use things like `<code>`.

app/classes/textile.rb
```diff
@@ -362,6 +399,38 @@ def pre_existing_links_replaced_by_placeholders!(do_object_links)
...
+    (?<! x{PROJECT)
+    (?<! x{SPECIES_LIST)
+    (?<! x{COMMENT)
+
+    (?<prefix> ^|\W) # prefix
+    (?: _+)
+    (?<id> [\p{Alpha}\-.\ ]+) # alpha chrs, hyphens, periods, spaces
+    (?: _+)
+
+    (?! (?: \w|<\/[a-z]+>)) # discard if followed by word char or close tag
```
JoeCohen, line 420: See https://github.com/MushroomObserver/mushroom-observer/pull/1528/files#r1241294746

app/classes/textile.rb
```diff
@@ -362,6 +399,38 @@ def pre_existing_links_replaced_by_placeholders!(do_object_links)
...
+  IMPLICIT_TERM_PATTERN = /
+    (?<! x{NAME) # discard match if preceded by an MO object tag
+    (?<! x{GLOSSARY_TERM)
+    (?<! x{OBSERVATION)
+    (?<! x{LOCATION)
+    (?<! x{USER)
+    (?<! x{IMAGE)
+    (?<! x{PROJECT)
+    (?<! x{SPECIES_LIST)
+    (?<! x{COMMENT)
```
JoeCohen, line 413: @pellaea, @mo-nathan 
### Is there a better solution?
These negative lookbehinds prevent further munging of MO tagged objects.
Maybe supplement the negative lookahead at the end of the regexp?

pellaea:
Oh good point, yes, I think that's okay.
-Jason

app/classes/textile.rb
pellaea: I don't think I understand the kinds of cases this guards against.  I think it would stop the second link in this:

```
Blah _observation 123_ _glossary term_ blah.
```
and that doesn't seem right.  What is the intention?

JoeCohen: Fair question. Sorry I didn't explain it better.

TLDR: The intent is to prevent tagged objects from being munged.

The regexp `IMPLICIT_TERM_PATTERN` is used by the `convert_implicit_terms_to_tagged_glossary_terms!` method.
The purpose of that method is to convert implicit glossary links (ex: `_lichen_`).

When the method is called, `self` has been modified. 
In particular, the explicit object links (ex: `_user 252_`) and implicit Name links (ex: `_Lichen_`) have been converted to tagged objects.

Each tagged object takes the form 
`<prefix>x{OBJECT_NAME __label__ }{ id }x`
So each tagged object includes a substring within underscores.

I want `convert_implicit_terms_to_tagged_glossary_terms!` to leave the tagged objects alone -- _not_ convert that substring to a tagged GlossaryTerm object.

Does that make sense?

pellaea: Oh my gosh, I get it now.  I was fooled by the `{x` business.  I never really learned all the fancy new regex notations and thought that was one... but no, it's literal. :)

Right, so I see why it's necessary.  I do think it *may* be better to do a single negative look-behind assertion instead of many:

```
(?<! (?:FOO|BAR|BAZ|etc.) )
```

Maybe.  Also, just a thought... how often do people include literal words like "x{XXXXX" in their mark-up???  Probably not often enough to worry.  In which case this would definitely be shorter/easier (if not as precise):

```
(?<! x{[A-Z_]+)
```

app/classes/textile.rb
```diff
@@ -301,36 +315,59 @@ def strip_out_sp_cfr_and_sensu(str)
...
+      type = $LAST_MATCH_INFO[:type]
+      id = $LAST_MATCH_INFO[:id]
+      matches = OTHER_LINK_TYPES.
+                select { |x| x[0] == type.downcase || x[1] == type.downcase }
+
+      if matches.one?
+        "#{prefix}x{#{matches.first.first.upcase} " \
+        "__#{tagged_object_label(type, id)}__ }{ #{id} }x"
+      else
+        orig
```
pellaea, line 360: I hate to complain, and I may well have been the one to write it this way originally, but...  This loop looks expensive, and more importantly unnecessary.  How about something like this instead:

```
CLEAN_TYPES = {
  "image": "IMAGE",
  "img": "IMAGE",
  "location": "LOCATION",
  "loc": "LOCATION",
  ...
}.freeze
if clean_type = CLEAN_TYPES[type.downcase]
  label = tagged_object_label(type, id)
  "#{prefix}x{#{clean_type}  __#{label}__}{#{id}}x"
else
  orig
end
```

Does that make sense?

app/classes/textile.rb
JoeCohen: Thanks again Jason. I appreciate your time and thoughts.

-  `(?<! x{[A-Z_]+)` is a good idea. But it throws `invalid pattern in look-behind: / (SyntaxError)`.  
The problem is that Ruby does not allow variable length lookbehinds. https://learnbyexample.github.io/Ruby_Regexp/lookarounds.html#variable-length-lookbehind  
-  `(?<! (?:FOO|BAR|BAZ|etc.) )`: I need to think about this. (I've been fighting regexp's for a solid week and am exhausted.)
(I probably can make something similar work because Ruby does allow different length alternatives in lookbehinds, provided that each alternative is fixed length. But my understanding is that will not improve efficiency because "Each alternative is treated as a separate fixed-length lookbehind." https://www.regular-expressions.info/lookaround.html

app/classes/textile.rb
JoeCohen: Complaints are great! I'll look at this tomorrow. Too tired tonight.

app/classes/textile.rb
pellaea: Your week of studying regexen has really paid off!  You have officially surpassed me.  I used to be so good with them.  You've convinced me that your solution is the best available.  (At least without a major refactor.)

pellaea:
Ah!! Good idea.  ...  only I don't think it's going to work, because there
are additional characters between those first 4 uppercase characters and
the label.

Ooooh, but we can change all of those words to be exactly the same length
in reality.  That is, have "CLEAN_TYPE" return "GLOS" instead of
"GLOSSARY_TERM" and so on.  I think all you'd have to do then is create yet
another lookup when converting back after Textile is done.  Something that
converts "GLOS" to "GLOSSARY_TERM", and so on.

It's work, but it may -- emphasize *may* -- improve the render time of some
pages significantly.  And it may not, in which case you're wasting a lot of
time for no good. :(

JoeCohen:
DOH! 
That's a great suggestion. 
I'll:
- leave the code as-is for now given that: it would take a lot of time to implement, it might not noticeably improve render time, and I don't want to add more delay to the PR.
- add your idea as a comment.

app/classes/textile.rb
JoeCohen: @pellaea: Another good idea. Thanks again for the complaint.
Here's what I've done: (pasting the whole thing here so that you get it via email):
```ruby
  MARKED_TYPE_TO_TAGGED_TYPE = {
    comment: "COMMENT",
    glossary_term: "GLOSSARY_TERM",
    image: "IMAGE",
    img: "IMAGE",
    location: "LOCATION",
    loc: "LOCATION",
    name: "NAME",
    term: "GLOSSARY_TERM",
    ob: "OBSERVATION",
    obs: "OBSERVATION",
    observation: "OBSERVATION",
    project: "PROJECT",
    species_list: "SPECIES_LIST",
    spl: "SPECIES_LIST",
    user: "USER"
  }.freeze

  # Convert _object name_ and _object id_ to a textile string.
  def convert_other_links_to_tagged_objects!
    gsub!(OTHER_LINK_PATTERN) do |orig|
      prefix = $LAST_MATCH_INFO[:prefix]
      marked_type = $LAST_MATCH_INFO[:marked_type]
      id = $LAST_MATCH_INFO[:id]

      tagged_type = MARKED_TYPE_TO_TAGGED_TYPE[marked_type.downcase.to_sym]
      next(orig) unless tagged_type

      label = tagged_object_label(marked_type, id)
      "#{prefix}x{#{tagged_type} __#{label}__ }{ #{id} }x"
    end
  end
```

pellaea: Looks great.  (By the way, much better name for the hash than my suggestion. :)

JoeCohen:
I plan to merge and deploy this, but postpone actively publicizing it.

PR #1530, opened 2023-07-03
Fix GlossaryTerm#update

JoeCohen:
Fixes a bug in `GlossaryTerm#update` which updated `<term>.user`. 
That caused `user` to set to the most recent editor of the Term (vs not touching `user`).

NOTE: In the live db, any incorrect `user` attributes should be fixed.

### DIFF
--- a/app/controllers/glossary_terms_controller.rb
+++ b/app/controllers/glossary_terms_controller.rb
@@ -50,7 +50,6 @@ class GlossaryTermsController < ApplicationController
 
     @glossary_term.attributes = params[:glossary_term].
                                 permit(:name, :description)
-    @glossary_term.user = @user
 
     return reload_form("edit") unless @glossary_term.save
 
--- a/test/controllers/glossary_terms_controller_test.rb
+++ b/test/controllers/glossary_terms_controller_test.rb
@@ -247,14 +247,20 @@ class GlossaryTermsControllerTest < FunctionalTestCase
   # ***** update *****
   def test_update
     term = glossary_terms(:conic_glossary_term)
+    creator = term.user
+    user = mary
+    assert_not_equal(user, creator,
+                     "Test needs user who didn't create the term.")
     params = changes_to_conic
-    login
 
+    login(user.login)
     assert_difference("term.versions.count") do
       post(:update, params: params)
     end
     assert_equal(params[:glossary_term][:name], term.reload.name)
     assert_equal(params[:glossary_term][:description], term.description)
+    assert_equal(creator, term.user,
+                 "Editing a Term should not change term.user")
     assert_redirected_to(glossary_term_path(term.id))
   end
 


JoeCohen:
@nwilson-eol @pellaea @nimmolo:
### How should I repair the live db?

1. Add a data migration (see below) to this PR?
2. Run a script (just the `up` part of the migration) on the webserver after deploying this PR (sans migration) 
**Issue**: That would leave the live db out-of-synch with the db snapshot.
3. Same as 2, plus create a new clean snapshot. 
**Issue**: I'm not confident about my ability to create the snapshot.

---

This migration took about 10 seconds on my iMac. 
I'll bet there's a smarter way to do it, e.g., SQL, but I can't figure it out.

```ruby
class FixGlossaryTermUser < ActiveRecord::Migration[6.1]
  def change
    reversible do |dir|
      change_table :glossary_terms do |t|
        # fixes user_id's that are wrong due to a now-fixed bug
        dir.up do
          GlossaryTerm.where.not(version: 1).each do |term|
            original = term.versions.first
            next if term.user_id == original.user_id

            term.update_column(:user_id, original.user_id)
          end
        end

        dir.down do
          GlossaryTerm.where.not(version: 1).each do |term|
            last_version = term.versions.last
            next if term.user_id == last_version.user_id

            term.update_column(:user_id, last_version.user_id)
          end
        end
      end
    end
  end
end
```

pellaea:
I just checked and there are only 27, I think:

```
SELECT count(t.id)
FROM glossary_terms t
JOIN glossary_terms_versions v ON v.glossary_term_id = t.id
WHERE t.version != 1
  AND v.version = 1
  AND t.user_id != v.user_id;
```

Maybe this SQL would do the trick, but with only 27 why not keep your safer, easier-to-read version?

```
UPDATE glossary_terms t 
JOIN glossary_terms_versions v ON v.glossary_term_id = t.id
SET t.user_id = v.user_id
WHERE t.version != 1
  AND v.version = 1
  AND t.user_id != v.user_id;
```

pellaea:
(By the way, is the down migration even necessary?)

JoeCohen:
Thanks Jason!
The number 27 makes sense. Almost all the terms were created by EXD and
other users were reluctant to touch them.
You're right about the down migration. I'll ditch it.
Otherwise, I will wait to hear from Nathan about whether he's OK with a
data migration (as opposed to some other solution.)

JoeCohen:
In my view migrations should focus on schema changes.  Sometimes you need
to manipulate data as part of that, but if you're not changing the schema,
I would not use a migration.  If all you’re doing is fixing bad data due to
some bug, then I'd do it with a script.  At work we usually use rails runner
for one time scripts that we create on a branch that we never merge into
the main branch.  It's something we expect to run multiple times then we
make it a rake task.

JoeCohen:
Thanks Nathan. I appreciate the guidance.
I will use a script (rather than a migration).
*Question: *If the script is in an unmerged branch, is there a way to run
it without switching to the unmerged branch, then switching back to main?
Seems complicated (and maybe unsafe).
(So would it be ok to put this script in this branch, and later -- after
it's run and a snapshot is created -- delete it (in a new PR)?)

JoeCohen:
We just copy the file directly to the server using scp and run it from
there.  The main reason we do it that way is so we don't have to change the
version number of the code running on the server and communicate that to
our customers.  The only reason we put it in the repo is so we can find it
again if needed.

JoeCohen:
Thanks for the added explanation.
I'll try that. (I.e.: remove the migration; create a new branch -- which
will not be merged - with just the script; and -- after deploying PR #1350
-- copy the file directly to the server using scp (after figuring out how
to do that) and run it from there.

PR #1533, opened 2023-07-11
Glossary blame

JoeCohen:
Displays the Creator and Editor on the GlossaryTerm page.
As discussed in the 2023-07-01 Tech Meeting. https://groups.google.com/g/mo-developers/c/qJxCqu3J-u0

### Manual Test
- Run the script found at this link: https://github.com/MushroomObserver/mushroom-observer/blob/e322883554e23c7bcddd7d6f9740c72fcf72cc87/script/db_fix_term_user.rb
(It fixes some errors in the db snapshot, which errors have already been fixed in the live db.)
- View a Glossary Term that you did not create
- Edit it
Expected result: It should list the Creator and Editor(s).

### DIFF
--- a/app/views/glossary_terms/show.html.erb
+++ b/app/views/glossary_terms/show.html.erb
@@ -51,4 +51,10 @@
     <%= :footer_created_at.t(date: @glossary_term.created_at.web_time) %><br/>
     <%= :footer_last_updated_at.t(date: @glossary_term.updated_at.web_time) %>
   </p>
+
+  <div id="glossary_term_authors_editors">
+    <%= panel_block do %>
+      <%= show_authors_and_editors(obj: @glossary_term, user: @user) %>
+    <% end %>
+  </div>
 </div>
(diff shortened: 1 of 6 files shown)


app/views/glossary_terms/show.html.erb
```diff
@@ -51,4 +51,10 @@
     <%= :footer_created_at.t(date: @glossary_term.created_at.web_time) %><br/>
     <%= :footer_last_updated_at.t(date: @glossary_term.updated_at.web_time) %>
   </p>
+
+  <div id="glossary_term_authors_editors">
+    <%= panel_block do %>
+      <%= show_authors_and_editors(obj: @glossary_term, user: @user) %>
+    <% end %>
+  </div>
```
JoeCohen, line 59: I find it always time-consuming to write the view.
But this one was relatively quick, thanks to @nimmolo's helper methods and the pattern he used for Name and Location.

nimmolo: Glad it helped. 

(Of course, I probably changed it for the bootstrap 5 branch, but it should be easy enough to adapt this and the other Glossary Term PR, and keep that branch up with these PRs.)

pellaea:
Also looks fine to me.

PR #1535, opened 2023-07-25
Glossary views

JoeCohen:
Updates Glossary views to:
- reference the new Guidelines
- add some simple instructions to the form


app/views/glossary_terms/_form.html.erb
```diff
@@ -5,13 +5,36 @@
...
+      append: help_block(
+        :p,
+        :form_glossary_name_help.t + " " +
+        link_to(
+          :glossary_term_index_documentation.t,
+          "https://github.com/MushroomObserver/mushroom-observer/blob/main/doc/glossary.md"
+        )
+      )
+    )
+  %>
```
JoeCohen, line 22: @nimmolo:
Off the top of your head, do you see a way to simplify this? 
(It displays fine, so I don't think it's worth your time to dig into it.
Part of the problem is that the link is to an external file (even though it's in the repo) 
in order to make it render with GitHub flavored markdown.)


but the erb is c

nimmolo: This seems right to me. Elsewhere sometimes I define the `help_block` at the top of the page as a variable to shorten up the `text_field_with_label` but it's not necessary.

app/views/glossary_terms/index.html.erb
```diff
@@ -8,8 +8,14 @@ tabs = User.current ? [
...
-<%= content_tag(:div, :glossary_term_index_intro.tp, class: "container-text") %>
-
+<%=
+  content_tag(
+    :div,
+    link_to(:glossary_term_index_documentation.tp,
+    "https://github.com/MushroomObserver/mushroom-observer/blob/main/doc/glossary.md"),
+    class: "container-text"
+  )
+%>
```
nimmolo, line 18: Note that you can use `tag.div(args)` these days, i'm using it a lot on the bootstrap branch.

app/views/glossary_terms/_form.html.erb
nimmolo: Might make more sense to move the file to a deployed folder (i'm finding public/doc for example) and call the file with a relative url:
`public/doc/glossary.md'

nimmolo: Whoops — of course it's an `href` attribute. 
It needs the full path, but you can maybe use something like

```
File.join(Rails.root, "/public/doc/glossary.md")
```

JoeCohen: Thanks for the feedback!
The file will be deployed. I'll give that a try.
But I wonder how it will render. It's written in Github flavored markdown.

app/views/glossary_terms/index.html.erb
JoeCohen: Good point.  re `tag`

nimmolo:
With a relative path you should be able to check the markdown rendering, locally. It should serve from your local repo (instead of the github blob) just like it would live, i would imagine. On Jul 25, 2023, at 8:39 PM, Joseph D. Cohen ***@***.***> wrote:﻿
@JoeCohen commented on this pull request.



In app/views/glossary_terms/index.html.erb:
> @@ -8,8 +8,14 @@ tabs = User.current ? [
 @container = :text_image
 %>
 
-<%= content_tag(:div, :glossary_term_index_intro.tp, class: "container-text") %>
-
+<%=
+  content_tag(
+    :div,
+    link_to(:glossary_term_index_documentation.tp,
+    "https://github.com/MushroomObserver/mushroom-observer/blob/main/doc/glossary.md"),
+    class: "container-text"
+  )
+%>

Good point.  re tag

app/views/glossary_terms/_form.html.erb
JoeCohen: Can't figure out how to link to the deployed file.
So sticking with what I've got for a while.
Maybe a longer term solution is to route everything through the `info` controller.

nimmolo: Joe -  finally got my localhost to boot and now I understand your idea, makes sense, of course!

Github's hosted version of the file displays legible HTML, i.e. it parses the markdown. If _we_ host the file it will just display raw text, as you'd see it in a markdown editor. 

Yes, I agree that the better long-term solution is to just make a static page in the app, not necessarily under the Info controller. 

The main question is db and admin-user-interface design. It would be nice to have a standard UI for editing page content for any page in the Info controller and storing that content in the DB by page ID, so that the page "view" would have no hardcoded HTML but just render the Textile from that db entry, a textile string. 

We'd just write a simple info/edit form with a big Textile textarea, as we would for an observation description, for example — the form for the info page would pull the current text from the db, and the update route would store edits back in the database. The live page would parse the Textile as we do elsewhere.

Obviously it's less of a project to make the Glossary page with raw HTML, like other info pages. But, I understand why you didn't do that, it's not very admin friendly - it's much nicer to edit in Markdown or Textile, isn't it. 

(Fwiw, I found a path that works, so you could theoretically host the file under `/public` in the deployed file, rather than linking back to Github.)

nimmolo: Aha. I think i am getting closer to a solution.

Maybe what would be useful is an Info **model**.
It could be a really simple model at first, with just a title and text. 

Then the info controller would have forms, accessible to admins only, controlling the CRUD of these info pages, and there could be routes for other namespaced InfoPages like names/info, for general info about names, or locations/info

I feel certain there are examples of stuff like this in the wild. 
This is basically what a CMS is, like WordPress. It's a very useful pattern. 
If the route itself is an attribute stored in the db, it seems even possible to dynamically "create" a brand new info page with a new route, through such a UI. Somehow routes.rb would dynamically check each entry in the Info table (in other words, each page) and create routes for each page.

JoeCohen:
I think this is a project for another day.
At the moment I'm trying to figure out how the redcarpet gem
<https://github.com/vmg/redcarpet> works and understand how we use Redcloth
so that we can use redcarpet as a substitute for Redcloth, which is no
longer maintained and has a security warning.
(I'm having trouble getting my head around how to do this.
Maybe I somehow have to trick redcarpet into interpreting MO-flavored
Textile as "standard" Markdown <https://www.markdownguide.org/basic-syntax/>.
Then redcarpet will correctly render the result as html.
If I can do that, then we should be able to render standard markdown as
html, see e.g.,
https://www.mitchcraver.com/2022/04/01/markdown-in-ruby-on-rails-with-redcarpet/
and maybe even extend redcarpet so that we can render Github flavored
markdown as html.)

nimmolo:
I agree — totally outside the scope of this PR, it just struck me as a big improvable.

Let's continue this in Discussions. I will start the discussion now, but hopefully you can flesh out what you're thinking there.

PR #1537, opened 2023-07-29
AJAXify the `ask_observation_question` form

nimmolo:
This PR continues the process of making most of the interaction forms on show observation pages into "modals", so people can ask questions and upload resources without leaving the page.

For easier cutting and pasting, i'm proposing giving the email form its own controller under `Observations::EmailsController`, ~so it can use CRUD actions for `new` and `create`.~  Updated to use refactored JS partials from #1538 

This controller shares a method `can_email_user_question?` with other email actions from the current `EmailsController`, so i'm moving that to a concern and including it in both places. That method will need to handle JS requests from AJAX forms, but once it does it will work for AJAXifying the other forms too.


JoeCohen:
I can't provide detailed feedback: I don't understand AJAX/JS/modals.
From a UI standpoint there seem to be some issues with the above:
- One long-time user says:
> Stuck comment box.
When I go here to reply https://mushroomobserver.org/names/60408?q=1otc0, it shows only half the reply box, it’s stuck at the top portion of the screen with no way to type anything in the box.

See https://mushroomobserver.org/names/28638

- Clicking on parent window destroys the modal. 
I often begin a Comment, type a bunch of text, then realize that I have to search for more MO data to add the Comment. 
If I click on the modal's "parent" window (not sure of the correct terminology), then goodbye modal with everything I've typed so far.

nimmolo:
TODOs:
- make comment and email text persist for on-page interactions (hide, rather than destroy, modal)
- make an icon beside the comment button that opens the comment form link in a new window

PR #1538, opened 2023-07-30
AJAX forms — DRY and keep progress when hiding form before submit

nimmolo:
## New Behavior ##
Per @JoeCohen's suggestion:
It would be nicer if a modal form you're working on doesn't erase what you've written, in the event you need to hide the modal and check something on the page underneath, and then re-open the modal. This PR makes that change for all existing modal forms, ahead of the "email observation question" PR.

#### Affected forms: ####
- Comments
- Collection Numbers
- Herbarium Records
- Sequences
- Namings

## Refactor ##
The first batch of AJAX forms for the `Observation` page were not DRY at all. Trying to change a basic javascript behavior in multiple places makes a refactor seem wise. If there are other JS alterations in the future, it will help.

So this PR refactors most of the AJAX form partials written so far, consolidating dozens that were basically duplicates of a few types, into a few general purpose "JS utility" partials that respond to user input.

- `new` and `edit` js responses (responding to links that open the forms)
  - Display the html form in a modal: `shared/modal_form_show` and `shared/modal_form_bindings`
- `create` and `update` js responses (responding to form submit) 
  - Update the page: `observations/show/section_update` with local vars
  - Show form submit issues via flash messages: `shared/modal_flash_update` and `shared/modal_form_reload`
- This PR starts to make use of Rails `action_name`, a variable that is always available in controllers, instead of passing an `@action` ivar to views
- Modal forms share the same `@title` as the regular HTML form pages. That ivar is now assigned in the controller for these forms, so it doesn't have to be passed as a local to `modal_form_show`

`Naming` forms need different JS responses depending on context. They may: 
- update the title of a matrix box and lightbox image caption, or 
- update the Observation naming table and potentially the title if consensus changed

I was handling this with JS logic in the response, by testing the page HTML to find out what page we're on. Now, it's handled upstream. Each of the "propose naming" buttons send a `context` param. It's added to the form via hidden field, and the controller decides which response to render from that param. Less brittle
#### Scope creep ####
- changes the FASTA link in the `Sequence` form to open in a new tab.

#### Observation modal forms ####
- Go to an existing `Observation`
- Click to add a `Comment`, `CollectionNumber`, `HerbariumRecord`, `Sequence`, or `Naming`
- Type or select some things, but before submitting, hide the modal
- Re-open the modal. It should be as you left it. :)
#### Help Identify modal forms ####
- Go to `Help Identify` page
- Propose a naming from a matrix box
- Cancel the modal and re-open
- Propose a naming from the lightbox
- Cancel the modal and re-open


nimmolo:
Requesting cursory manual test only :)

JoeCohen:
I did some of the Manual Tests and it's all working as advertised.
Thanks for all the extended work on this.

PR #1540, opened 2023-08-03
Start to refactor "tabs" link helpers

nimmolo:
"Tabsets" in most cases are the short menu of links in the top right corner of the page. In the ~130 views that generate them, there are usually only one or two links per set, and they're generated at the top of each view template. But some pages have longer lists, and helpers to generate them. 

The goal of this PR is to get all the tabset code in one `helpers` directory, so it can be refactored. Later commits will move the rest of the tabsets out of the views into new helpers in this directory, and then refactor them.

#### Background ####
In the BS 5 branch i'm aiming for a significant redesign of the page header and navbars. I don't want to recreate the current design, because it would have to be recoded anyway. BS 4-5 has tools for de-cluttering the navigation UI that should feel familiar. 

One component of that is redoing the tabsets. They're currently arrays of anchor tags generated by Rails `link_to`, and changing their presentation across the site would involve adding attributes to every link in every tabset, individually. 

This refactor would instead redefine each link as a hash of attributes `{ link: , text: , class:, **opts }`; those hashes could be used either to generate the same `link_to` links as currently, or to generate more formatted HTML, like list groups and dropdown menus, in the BS5 branch.

### Questions ###
- `show_index_of_objects` seems as if it used to generate links for the tabset, but it does not currently.


pellaea:
Sounds like a good idea all around. Thanks for taking it on,  Nimmo.
- Jason

nimmolo:
Thanks @pellaea. Ironically this is closer to what we had originally, before i started messing with the views (I think Ray coded?). He passed `@links` as arrays of `[[text, url], [text, url]]` and used the helper `create_links` to run them through Rails `link_to` to make the HTML. I'm the one who flattened them into plain links!

Re: the question above, do you recall `show_index_of_objects` generating `@links` at one time? It doesn't now under any circumstance that I can tell, i'm thinking to remove the reference to the `@links` in the notes above the method definition to help unconfuse myself.

nimmolo:
Going to merge at this point to resolve conflicts with the BS5 branch, so future commits will not cause conflicts that branch — this is as far as I got with this on that branch.

Will continue the work on this branch in a second PR

pellaea:
Couldn't say, myself. You know the current code better than me! I wonder
what @links was used for before. Maybe for defining undefined locations?
Beats me! (Sorry, I'm on the road, so I can't check the code myself...)

PR #1543, opened 2023-08-03
Refactor `rss_logs_type_filters` and fix the param parser

nimmolo:
Fixes the way the `RssLogsController` parses the `params[:type]` from the new filter markup using `check_box_with_label` helper. The `params[:type]` is now sent as a hash like this, and needs to be parsed differently. 
`{ observations: "1", names: "1", locations: "0" ... }`
The controller now selects all the keys whose values are "1" and turns it into a string.

Single type params are sent as a string, like before, so there's no change there.
`glossary_terms`

Also renames a couple partials.

Manual Test:
- go to "Latest > Changes by users"
- Change the filters, be sure the RssLogs are filtered correctly
- Click "Make this my default"
- Click again on "Latest > Changes by users" and be sure the default types are applied


nimmolo:
Thankfully #1540 isn't deployed yet, it was a refactor of the Activity Logs "types" filters and broke them.

This PR fixes them. @JoeCohen if you have a moment, could you test that they're working on this branch?

PR #1553, opened 2023-08-13
Send all MO emails via `QueuedEmail`

nimmolo:
Converts all remaining **controller calls to mailers** to go through new `QueuedEmail` classes.

Sorry this is such a doozy, but i was on a roll. With @pellaea's help my understanding of QE vastly improved as I handled more of these, adjusting and improving the previous ones. 

Renames several of the mailers to be hopefully less ambiguous (to me at least).


pellaea:
Thank you!

nimmolo:
Something's wrong here. Going to re-open a slightly stripped down PR

nimmolo:
@pellaea could you have a look at this - just the `QueuedEmail::VerifyAPIKey` code - and see if you notice what's amiss here? There are two related errors that boil down to the fact that `for_user` is not getting defined. It could be a test error, i guess, but i imagine it's in how i'm handling that optional arg. I've spent about 4 hours on it and i'm flummoxed.

- In the VerifyAPIKey request, who's the `for_user` and who's the `user`? Confused because these vars are inconsistently named in related methods (`user` and `other_user` respectively). I'm guessing `for_user` is marked as the sender of the APIKey when it's being sent to someone else. Otherwise it's people creating keys for themselves and there's no `for_user`.
- Disclaimer: I still don't really understand the `QueuedEmail` class, even after reading and rereading the note block in that class. What i'm writing is purely code mimicry and probably as ridiculous as anything ChatGPT would write, with extra typos indicating human origin. (Does `create_email` instantiate an object of the class?) 
- What sort of object (eg `Comment`), sent to a mailer method, needs an associated ID via`get_integer`? All of them? Why does it need that ID integer, when the mailer uses the whole object? And why don't sender/recipient `User` objects need integers, when the mailer refers to the object by ivar?

- I found that Model class names may collide with classes under `QueuedEmail`. Calls to any Model class, even in mailers, may need to specify global namespace like `::User`. Preemptively doing this all over the place cleared up a lot of errors.

nimmolo:
@pellaea hang on, i'm preparing a stripped down PR for troubleshooting. will copy these notes to that PR

test/mailers/queued_email_test.rb
```diff
@@ -231,13 +231,16 @@ def test_author_request_email
...
     QueuedEmail::VerifyAPIKey.create_email(mary, dick, key)
     assert_email(0,
                  flavor: "QueuedEmail::VerifyAPIKey",
-                 from: mary,
-                 to: dick,
+                 from: dick,
+                 to: mary,
                  user: mary.id,
-                 other_user: dick.id,
+                 other_user: disk.id,
```
nimmolo, line 243: @pellaea Why`disk` and not `dick`? 
And shall we go ahead and change this param now to `app_user` while we are thinking about it?

nimmolo: Oh right the `app_user` param is for `user`, not `other_user`. Anyway... let's get those var names as clear as we can while we have a bit of light on the subject...

pellaea:
Because I'm a disk-head, obviously.  Look, I'm trying (and failing) to do
three things at once and my wife and guest came home right in the middle of
it...

test/mailers/queued_email_test.rb
nimmolo: lol. thanks. i'll handle the var renaming then, if i can keep it straight!

nimmolo:
So, re: var names. 

Is it going to mess anything up if i change the var name from `other_user` to `app_user`, as used in the translation string `:email_verify_api_key_intro` and `views/verify_api_key_mailer/build`? Will that invalidate existing translations?

pellaea:
Var names in general shouldn't affect anything, except the special @user =
User.current determines the theme, language, etc.

But also note that the stored variables are of course different.  Since the
sender and receiver are already stored in dedicated columns in the
queued_emails database table, we only needed to store the api key id as an
attached integer/object.  But that means that we need to convert from
user/app_user to recipient/sender and then back, noting that recipient
changes to "to_user".  No changing of variables within VerifyApiKey is
going to avoid at least that amount of confusion.

You might consider -- because I know you like to make sweeping changes! :)
-- changing queued_email.user_id and queued_email.to_user_id columns to
something better.  But that does involve renaming database columns if you
want to do it "right".  I don't have a problem with "to_user" per se, but
then it really should be "from_user" not just plain "user", because plain
"user" suggests false associations like that the "from_user" somehow owns
or created the record.  And I think that's a dangerous and slippery road.
I mean, yeah, sorta, I guess that's right.  But it invites confusion.
Well, it confused me in any case.

But doing so means changing stuff throughout the QueuedEmail tree.

nimmolo:
Thanks. I will keep those column names the same: `user` and `to_user`.  They're clear enough to me, now that I finally understand that part of QE class! Thank you for explaining.

I have two questions to hopefully wrap this up. There are two mailers that don't seem to fit the pattern:
- `WebmasterMailer` doesn't use either sender or receiver column, as i imagine you know. Because this action is important for cases like people having trouble signing in, it does **not** require a User id/object and just sends with a raw email address. On the other hand, this class of email would be nice to have in an eventual message center, and seems equally important for the test site. Can you imagine an efficient way to modify the `QueuedEmail::Webmaster#deliver_mail` so that it checks for a `sender_email` value before refusing to compose an email? Maybe it's just that simple, but I imagine that the class record might not validate without a sender. But wait a minute, many emails have no sender... hmm! (I've currently got it hacked to look up a `User.find_by(email: sender_email)` but that's not right.)
- `DeniedMailer` needs an object (`@user_params`) but it's not an ActiveRecord object. We could leave this case out of the QueuedEmail mechanism, but out of curiosity is there already a way to handle such a thing?

pellaea:
My understanding, reading through queued_email.rb just now, is that neither
user nor to_user need to be non-nil.  And it ... oh! shoot, if user ==
to_user it will raise an error.  We should fix that (line 256 in my copy of
queued_email.rb).  But then it should call deliver_email, which can then be
defined to do anything we like in the subclass WebasterMailer.  Including,
for example, we can store the email address of the "sender" as an attached
string, and access that in deliver_email just like before (when it wasn't
part of the queued emailer).

As for the denied email, honestly, why not just rewrite it as an admin
notification.  It really doesn't need a special class.  That's overkill.
It doesn't need translation or anything, just store the content of the
notification as a plain old string.

(I would also love to see those horribly clunky "begin / object.foo /
rescue StandardError / bar / end" constructs die and ignominious death. The
much more concise and readable "object&.foo || bar" should be more or less
equivalent in most cases.  Certainly in these cases.)

app/models/queued_email/webmaster.rb
```diff
@@ -0,0 +1,35 @@
...
+    def self.create_email(sender_email:, content:, subject: nil)
+      raise("Missing email address!") unless sender_email
+      raise("Missing content!") unless content
+
+      result = create(nil, nil)
+      result.add_string(:sender_email, sender_email)
+      result.add_string(:subject, subject) if subject
+      result.set_note(content)
+      result.finish
+      result
```
nimmolo, line 27: @pellaea does this seem right, using the string `:sender_email`?

test/controllers/herbaria_controller_test.rb
```diff
@@ -405,11 +407,14 @@ def test_create
...
-    assert_includes(email.body.to_s, "Burbank Herbarium")
-    assert_includes(email.body.to_s, herbarium.show_url)
+    assert_equal(count_before + 1, QueuedEmail.count)
+    # FIXME: - can QueuedEmail content be tested in a controller test?
+    #         Or should it be elsewhere?
+    # email = ActionMailer::Base.deliveries.last
+    # assert_equal(katrina.email, email.header["reply_to"].to_s)
+    # assert_match(/new herbarium/i, email.header["subject"].to_s)
+    # assert_includes(email.body.to_s, "Burbank Herbarium")
+    # assert_includes(email.body.to_s, herbarium.show_url)
```
nimmolo, line 417: @pellaea see note here. Seems like all i can test for here is that an email has been queued. No other controller test seems to test the content of the `QueuedEmail`, and only `ObservationsControllerTest` tests that an email has in fact been queued.

app/models/queued_email/add_record_to_herbarium.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def deliver_email
+      # Make sure it hasn't been deleted since email was queued.
+      return unless herbarium_record
+
+      AddHerbariumRecordMailer.build(user, to_user,
+                                     herbarium_record).deliver_now
+    end
+  end
+end
```
nimmolo, line 27: @pellaea, @JoeCohen 

I think Joe had originally sketched out this `QueuedEmail` class, but in the code thus far, had not implemented it, and it had no test yet in `QueuedEmailTest`. 

I added the `reply_to` in the `AddHerbariumRecordMailer`, because it seemed from the way `HerbariaControllerTest` had been written that this should be set by the mailer. It wasn't set before, though, and it was passing tests.

pellaea: Half-finished code sounds more like my style, honesty, but thanks anyway for fixing it!  I will trust that you've got it down now and that your added reply-to is appropriate.

test/controllers/herbaria_controller_test.rb
nimmolo: Thanks. When i binding break here, whether I get the last QE or dump all QE's, all of them are Webmaster emails. I don't get why but at least it does queue another email... 

But i'd rather get it!

Between them, `ApplicationMailerTest` and `QueuedEmailTest` seem like they're covering the email creation, and as i mentioned no other controller test tests the attributes of a `QueuedEmail` in any way. But I agree, it seems like I should be able to get that last instance.

app/models/queued_email/add_record_to_herbarium.rb
nimmolo: update... `reply_to` was not appropriate! Reverted.

app/models/queued_email.rb
```diff
@@ -223,30 +229,24 @@ def send_email
...
-              queued_email_strings.map { |x| "#{x.key}=\"#{x.value}\"" }.
-              join(" ")
+      "from=#{user&.login || "nil"} " \
+      "to=#{to_user&.login || "nil"} " +
+              queued_email_integers.map { |x|
+                "#{x.key}=#{x.value}"
+              }.join(" ") +
+              queued_email_strings.map do |x|
+                "#{x.key}=\"#{x.value}\""
+              end.join(" ")
```
nimmolo, line 239: @pellaea - Here are those statements without the `begin...rescue...end`. Sanity check that this is correct...

app/models/queued_email.rb
```diff
@@ -223,30 +229,24 @@ def send_email
...
-      result = deliver_email
-    end
+    # result = false
+    # if user == to_user
+    #   unless Rails.env.test?
+    #     raise("Skipping email with same sender and recipient: #{user.email}\n")
+    #   end
+    # else
+    result = deliver_email
+    # end
```
nimmolo, line 249: ...and here's the deletion of the check if `user == to_user`. Delivers mail regardless.

pellaea: Actually, I was thinking more along the lines of checking `if user.present? && user == to_user`.  It continues not to make sense to send notifications to a user that they changed something themselves! :)  But if both user and to_user are nil, then it is probably this admin message, and *then* we should send it even though user == to_user.  Does that make sense?

app/models/queued_email.rb
pellaea: That looks right to me.  Hopefully it won't break anything! :)  I think the only way user.login would ever have failed is if user was nil, in which case this should be equivalent.  And user.login will always be non-falsey, so again should be safe.

test/models/comment_test.rb
```diff
@@ -122,15 +124,19 @@ def do_comment_test(chg, obs, user, summary, comment = "")
...
-      "to: #{mail["to"]}, subject: #{mail["subject"]}"
-    end
-    "These emails were sent:\n#{strs.join("\n")}"
+    all_emails = QueuedEmail.all[start..-1]
+    binding.break
+    # strs = ActionMailer::Base.deliveries[start..-1].map do |mail|
+    #   "to: #{mail["to"]}, subject: #{mail["subject"]}"
+    # end
+    # "These emails were sent:\n#{strs.join("\n")}"
   end
```
nimmolo, line 141: After converting Comment emails to use QE, this sub-test (used in many comment tests) fails. I'm going to play with this and see if i can get it to check the recently queued emails for the `to` and the `subject`

nimmolo: Same issue. The only QE returned by `all_emails` is
```ruby
[#<QueuedEmail::Webmaster:0x00007fed92f92bf8
  id: 3379,
  user_id: nil,
  queued: Tue, 15 Aug 2023 18:14:31.000000000 EDT -04:00,
  num_attempts: nil,
  flavor: "QueuedEmail::Webmaster",
  to_user_id: nil>]
```
It seems like the last one should be a `QueuedEmail::CommentAdd`. 
Do queued email records not get added immediately, or something?

nimmolo: Aha - I needed to set 
```ruby
  def setup
    QueuedEmail.queue = true
  end
```
Do i also need to set it to `false` during `teardown`?

nimmolo: Set to false during teardown, yes.

nimmolo:
@pellaea  re:
> As for the denied email, honestly, why not just rewrite it as an admin
notification.  It really doesn't need a special class.  That's overkill.
It doesn't need translation or anything, just store the content of the
notification as a plain old string.

Sorry to be thick. I'm afraid i don't know what an admin notification is, if it's not a Webmaster email and it's not a translation string. Or do you mean... rewrite it as a Webmaster email with the params-printout as the content?

nimmolo:
Ahaa... I think you do mean send a Webmaster email with that as the content. Done.

test/controllers/herbaria_controller_test.rb
nimmolo: Solved. 

The last  `QueuedEmail` _should_ of course be a `Webmaster` email, not an `AddRecordToHerbarium` email, and what we can check for is that the `sender` is set to `katrina`. Done.

test/controllers/api2_controller_test.rb
```diff
@@ -354,6 +354,8 @@ def test_post_user
     assert_equal(CGI.escapeHTML("<p>New API2 Key</p>"), notes.to_s)
   end
 
+  # FIXME: does this test also need to be altered to test for QueuedEmail?
+  # Why does it pass if not?
   def test_post_api_key
```
nimmolo, line 359: This test checks mail deliveries, but `api_key_api.rb` has been changed to use `QueuedEmail` now. 

Why does this test still pass? Is it because `QueuedEmail.queue` is not set to `true` and it sends right away?
Should I leave it as is (you can test more things in the content of an actual `ActionMailer::Base.deliveries`!) or should I also test that it's going into the queue?

test/controllers/herbaria_controller_test.rb
```diff
@@ -405,11 +407,9 @@ def test_create
...
     assert_equal(create_params[:description].strip, herbarium.description)
     assert_empty(herbarium.curators)
-    email = ActionMailer::Base.deliveries.last
-    assert_equal(katrina.email, email.header["reply_to"].to_s)
-    assert_match(/new herbarium/i, email.header["subject"].to_s)
-    assert_includes(email.body.to_s, "Burbank Herbarium")
-    assert_includes(email.body.to_s, herbarium.show_url)
+    assert_equal(count_before + 1, QueuedEmail.count)
+    assert_equal(katrina.id, QueuedEmail.last.user_id)
+    QueuedEmail.queue = false
```
nimmolo, line 412: Hopefully this accomplishes what we want. The controller sends a webmaster email with `katrina` as the `user`.

test/controllers/projects/admin_requests_controller_test.rb
```diff
@@ -10,7 +10,7 @@ def test_post_admin_request
         project_id: eol_project.id,
         email: {
           subject: "Admin request subject",
-          message: "Message for admins"
+          content: "Message for admins"
         }
```
nimmolo, line 14: This has been wrong for awhile (the param name in the form is definitely `content`) but somehow was passing. 

This fix was needed to get it passing now.

test/models/queued_email_test.rb
```diff
@@ -8,10 +8,11 @@ def test_send_email
...
 
-  def test_send_email_no_to_from
-    email = QueuedEmail.new
-    assert_not(email.send_email)
-  end
+  # Actually it does now send with no_to_from, if there's also no user
+  # def test_send_email_no_to_from
+  #   email = QueuedEmail.new
+  #   assert_not(email.send_email)
+  # end
```
nimmolo, line 15: Want to double-check that it's ok to eliminate this test, now that the mailer will actually send without a `to` and a `from`

test/controllers/herbaria_controller_test.rb
nimmolo: Whoops, just found the method `QueuedEmail#dump`. Will assert against that.

test/models/queued_email_test.rb
pellaea: I can't follow all of that!  But this one, yes, I think this test is super unimportant, so it's okay to delete it.

nimmolo: Will delete. 

I had multiple threads going here in the github PR -- that might have gotten confusing on email, sorry

test/controllers/herbaria_controller_test.rb
nimmolo: Restored previous assertions using `QueuedEmail` methods.

test/controllers/api2_controller_test.rb
nimmolo: @pellaea This is my last remaining question on the PR. 

`api2_controller_test#test_post_api_key` is checking for emails in `ActionMailer::Base.deliveries`, and it passes. But the emails are being sent via QueuedEmail now. What's going on, do you think?

nimmolo: Ah, nevermind! It does indeed stop working if i turn on the queue.
Adding a line in the test to keep the queue off for this test, in case it ever gets left on and there are mysterious errors.

PR #1555, opened 2023-08-14
`VerifyAPIKey` troubleshooting

nimmolo:
@pellaea could you have a look at this and see if you notice what's amiss here? There are two related errors that boil down to the fact that `for_user` is not getting defined. It could be a test error, i guess, but i imagine it's in how i'm handling that optional arg. I've spent about 4 hours on it and i'm flummoxed.

- In the VerifyAPIKey request, who's the `for_user` and who's the `user`? Confused because these vars are inconsistently named in related methods (`user` and `other_user` respectively). I'm guessing `for_user` is marked as the sender of the APIKey when it's being sent to someone else. Otherwise it's people creating keys for themselves and there's no `for_user`.
- Disclaimer: I still don't really understand the `QueuedEmail` class, even after reading and rereading the note block in that class. What i'm writing is purely code mimicry and probably as ridiculous as anything ChatGPT would write, with extra typos indicating human origin. (Does `create_email` instantiate an object of the class?) 
- What sort of object (eg `Comment`), sent to a mailer method, needs an associated ID via`get_integer`? All of them? Why does it need that ID integer, when the mailer uses the whole object? And why don't sender/recipient `User` objects need integers, when the mailer refers to the object by ivar?

- I found that Model class names may collide with classes under `QueuedEmail`. Calls to any Model class, even in mailers, may need to specify global namespace like `::User`. Preemptively doing this all over the place cleared up a lot of errors.

### DIFF
--- a/app/classes/api2/api_key_api.rb
+++ b/app/classes/api2/api_key_api.rb
@@ -42,7 +42,7 @@ class API2
     def after_create(api_key)
       return if api_key.verified
 
-      VerifyAPIKeyMailer.build(@for_user, @user, api_key).deliver_now
+      QueuedEmail::VerifyAPIKey.create_email(@for_user, @user, api_key)
     end
 
     def get
--- a/app/models/queued_email/verify_api_key.rb
+++ b/app/models/queued_email/verify_api_key.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+# VerifyAPIKey Email
+class QueuedEmail
+  class VerifyAPIKey < QueuedEmail
+    def api_key
+      get_object(:api_key, ::APIKey)
+    end
+
+    def self.create_email(for_user = nil, user, api_key)
+      raise("Missing api_key!") unless api_key
+
+      result = create(::User.admin, user)
+
+      # for_user ||= user
+      # if other_user
+      result.add_integer(:for_user, user.id) if for_user
+      # else
+      #   result.add_integer(:other_user, recipient.id)
+      # end
+
+      result.add_integer(:api_key, api_key.id)
+      result.finish
+      result
+    end
+
+    def deliver_email
+      # Make sure it hasn't been deleted since email was queued.
+      return unless api_key
+
+      VerifyAPIKeyMailer.build(for_user, user, api_key).deliver_now
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


pellaea:
I just pushed a fix to the original branch.  I'm lazy and will let CI test
it for me.  We'll see if I got it right in a minute...

PS. Yes, the switcheroo from (from_user, user) to (user, other_uesr) is
confusing! The view needs the recipient to be called "user" because @user
is globally special.  We should probably change the queued email code to be
the same.  "other_user" is over obscure, too.  Can you think of a better
name?  "app_user" maybe?  Clearly I was anticipating allowing users to
create additional API keys for themselves via the API, but that was never
implemented, so I think it is safe to "hard-code" it for the one and only
actual extant use case -- i.e., apps creating API keys on behalf of a
user.  So other_user is the user_id of the nominal account for the app.
Remember, each app must itself have an account on MO so that it can own an
API key of its own in order to do read-only requests safely.  The app can
ask the user for their API key so that it can do write/update requests on
behalf of that user.  Thus the confusion of two users for some API
requests.

JoeCohen:
This was over my head. But I want to make sure it won't prevent people from
using Pulk's program that "mirrors" MO Observations to iNat.
Pulk's instructions say:

> You will be prompted to authorize this program to read/write your
> observations on both platforms: at the start, you'll be directed to (1)
> create an "API Key" on MO and supply it;

See https://github.com/JacobPulk/mirror#what-you-will-have-to-do

pellaea:
Nonono, this won't change the behavior of MO at all, this is entirely
refactoring for appearance.  Nimmo (correctly) pointed out that the
existing code is confusing.  So as we .. Nimmo ... wait, no, I've spent
time on this too, so it's "we" dammit! ... convert these emails to use the
QueuedEmail mechanism, we're making some changes to variable names to make
it (hopefully) less obscure.

nimmolo:
@JoeCohen The switch of remaining emailers to `QueuedEmail` is also for two other reasons
- So thenew  test app on BS4 that will hopefully soon be lodging at test.mushroomobserver.org can send emails via the main site URL
- So we can eventually build an in-app "Message Center"

pellaea:
Yes, thank you, Nimmo.  Those reasons, too!

PR #1560, opened 2023-08-17
Remove `disable_link_prefetching` filter

nimmolo:
>This PR should not merge until it is certain that all destructive controller actions are using new Rails CRUD paths, or at least that they do not respond to GET requests. PUT, PATCH, POST and DELETE are OK, because they don't respond to prefetching.

~To check: image remove (multiple controllers) and any nested controller CRUD actions that don't have default action names. `show` and `index` actions with other names are OK.~ Image `attach` and `remove` actions both use PUT.

Update: The routes seem to be OK.

Actually, i just read on MDN that resources are not prefetched unless they're declared in a `<link>` element (not an `<a>`). [Reference](https://developer.mozilla.org/en-US/docs/Glossary/Prefetch#resource_prefetching). Rails Turbolinks has some kind of extra prefetching magic, but we don't use that. I believe this filter should be OK to delete.


nimmolo:
The core of this PR is the deletion of a single method, implemented across dozens of controllers.

My review request is not for code review really, but rather just a sanity check.  "Any objections?" 
(@pellaea indicated via email he thought it should be ok, but i'm double checking here.)

I believe it's good to go.

pellaea:
Oh hell yeah, tear that thing out.  I never did know whether it made the slightest difference.  Probably not.

PR #1563, opened 2023-08-22
Replace `@title` assignments in controllers and views with `add_page_title` and `add_index_title` helpers

nimmolo:
The point of this PR is to make page titles easier to track down and possibly change, in the Bootstrap 5 branch. It's currently a bit mysterious to understand where in the inheritance chain some titles are assigned, because it's an ivar assignment that can occur anywhere. (The `@title` is used both in `<html><head><title>` and in the page content.)

This PR removes assignments of `@title` from any non-mailer controller actions into corresponding view templates, where it is already assigned in most cases. It then replaces the ivar assignment with a new helper `add_page_title(title)` that makes both `content_for(:title)` and `content_for(:document_title)` (the html title). 

In most cases, the title is assigned directly from the translation string, but if there's logic involved, it's moved to a helper method for that controller. The logic for index titles is removed from `ApplicationController` into a an `add_index_title(query, no_hits: nil)` helper.

`@title` is still used in controllers and builders for all mailers.


app/helpers/title_and_tabset_helper.rb
```diff
@@ -12,21 +16,65 @@
...
+  # Special title for new obs default home page query
+  def index_default_title(query)
+    if query.title_args[:type] == :observation &&
+       query.title_args[:order] == :sort_by_rss_log
+      return :query_title_observations_by_activity_log.l
+    end
+
+    query.title
+  end
+
```
nimmolo, line 76: The methods above are the crux of the PR.

PR #1564, opened 2023-08-22
Fix show obs tabset map link

nimmolo:
I believe we discovered and discussed this issue at a meeting a few weeks ago, and then I forgot to fix it. The issue is that it seems the show Observation "Map" link should be a link to "show the current observation’s location on a map". 

It presently links to a "Map of Locations with Observations of {obs.name}”, which is more or less exactly the same as the “Occurrence Map for {obs.name}” linked in the same tabset above.

obs_map

### DIFF
--- a/app/helpers/tabs/general_helper.rb
+++ b/app/helpers/tabs/general_helper.rb
@@ -3,31 +3,27 @@
 module Tabs
   module GeneralHelper
     def coerced_observation_query_link(query)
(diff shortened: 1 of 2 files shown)


nimmolo:
Also fixes helpers for coerced_query_links (when the query returns no hits) on index search result pages:

Screen Shot 2023-08-22 at 2 12 45 PM

PR #1568, opened 2023-08-23
Reorganize show obs tabs for name info

nimmolo:
Move some show obs "tabset" links to a reconfigured name panel

The name panel is now in two columns, for links "on MO" and "on the web"

about_name

### DIFF
--- a/app/views/observations/show/_name_info.html.erb
+++ b/app/views/observations/show/_name_info.html.erb
@@ -3,20 +3,25 @@
   name = observation.name
 %>
 
-<%= content_tag(:p,
-                link_to(:show_name.t(name: name.display_name_brief_authors),
-                        name_path(name.id))) %>
-<%= content_tag(:p, link_to("MyCoPortal", mycoportal_url(name),
-                            target: :_blank)) %>
-<%= content_tag(:p, link_to("Mycobank", mycobank_name_search_url(name),
-                            target: :_blank)) %>
-<%= content_tag(:p, link_to(:show_observation_more_like_this.t,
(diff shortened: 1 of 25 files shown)


app/views/observations/show/_name_info.html.erb
```diff
@@ -1,22 +1,24 @@
...
+<%=
+panel_with_outer_heading(
+  id: "observation_name_info",
+  heading: :show_name.t(name: name.display_name_brief_authors),
+  inner_class: "name-section small"
+) do %>
+
+  <%= tag.div(class: "row") do %>
+    <%= tag.div(class: "col-xs-6") do
+      concat(tag.div("On MO:", class: "font-weight-bold"))
```
pellaea, line 15: These should probably be translated strings.

nimmolo: good catch, thanks! i mocked those up on BS branch and forgot.

What do you think about the box title being "About this taxon", rather than "About {name}"?
Wondering because the "About {name}" wording is repeated for the MO name link below.

pellaea:
You probably shouldn't ask my opinion on UI design!  Whatever you think
looks best and is most understandable at a glance?

app/views/observations/show/_name_info.html.erb
nimmolo: Added the translation strings. 

Here it is with "About this taxon" in case @JoeCohen  wants to weigh in.

Screen Shot 2023-08-23 at 3 40 29 PM

PR #1569, opened 2023-08-24
Refactor `add_sorter` helper methods to work more like `add_tab_set`

nimmolo:
Simplifies the logic of the `sorter` methods to assemble attribute arrays from the raw sorts, and then HTML from the attribute arrays.

Hopefully makes it easier to refactor the HTML, later.

### DIFF
--- a/app/helpers/title_and_tabset_helper.rb
+++ b/app/helpers/title_and_tabset_helper.rb
@@ -205,11 +205,11 @@ module TitleAndTabsetHelper
   end
 
   def create_dropdown_tabs(links)
-    xtrargs = {
+    extra_args = {
       role: "menuitem",
       class: "dropdown-item"
     }
-    create_tabs(links, xtrargs)
+    create_links_to(links, extra_args)
   end
 
   def dropdown_link_options(args = {})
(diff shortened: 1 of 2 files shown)


app/helpers/title_and_tabset_helper.rb
```diff
@@ -239,46 +239,51 @@ def filter_for_type(types, type)
...
-    results << sort_link(:sort_by_reverse.t, query, reverse_by(query, this_by))
+      link_with_query(title, path, **args)
+    end
   end
 
-  def link_or_grayed_text(link_all, this_by, label, query, by)
-    if !link_all && (by.to_s == this_by)
-      [label.t, nil] # just text
-    else
-      sort_link(label.t, query, by)
```
nimmolo, line 274: Note that a link with a `disabled` attribute is not clickable regardless of its `href`, so it's not necessary to send a blank path here.  I noticed the `link_or_grayed_text` method doesn't actually make a blank path — it makes a path to the index, with a `q` param but no `by` param.

I decided to handle it by setting a boolean `active` flag on the `sort_link` array. It's another attribute but it just seems valuable to preserve the correct destination path for the active link, for example if someone tries to "copy link" in the browser.

PR #1571, opened 2023-08-28
Abbreviate location title used in `observation_by_location` queries

nimmolo:
Location strings can be quite long. For example: 
`Observations from between the Adventure Playground and University Ave. in the Berkeley Marina, Berkeley, Alameda Co., California, USA`

Name strings can be similarly long with the authors, etc., but we use `text_name` in the titles. I'd like to do the same for location titles.

This PR abbreviates the location for page titles to just the location, without locality, region or country.
`Observations from between the Adventure Playground and University Ave. in the Berkeley Marina`

I was considering furthermore abbreviating that title to however many words, if it's something like the above, it would be shortened to 
`Observations from between the Adventure Playground...`

But I think i'll leave that to a template helper to do. The only reason to do that would be to fit it in a different part of a template, and this PR is not about changing the templates. 

The argument could be made that the entire truncation should happen in a helper, but the controller is in a position to know if the location string returned is in scientific or postal order, because that depends on the `@user` preference. (That preference also is available in the view via `@user`, but this seems to me like the right place to do it.)

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -315,13 +315,11 @@ class Location < AbstractModel
   # Just the location without locality, region, country
   # Furthermore abbreviates the location if longer than 4 words
   def title_display_name
-    str = if User.current_location_format == "scientific"
-            scientific_name.split(", ").last
-          else
-            name.split(", ").first
-          end
-    str = "#{str.split.first(4).join(" ")}..." if str.split.length > 4
-    str
+    if User.current_location_format == "scientific"
+      scientific_name.split(", ").last
+    else
+      name.split(", ").first
+    end
   end
 
   def display_name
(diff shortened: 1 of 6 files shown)


app/models/location.rb
```diff
@@ -311,6 +311,17 @@ def self.is_unknown?(name)
     false
   end
 
+  # Very abbreviated description of the location for shorter query titles.
+  # Just the location without locality, region, country
+  # Furthermore abbreviates the location if longer than 4 words
+  def title_display_name
+    if User.current_location_format == "scientific"
```
pellaea, line 318: Errr, I'm probably missing something!... but since you're only taking one term, does it matter whether the user's location format is scientific or not?  Wouldn't `name.split(", ").first` be the same either way?

pellaea:
Looks good!

app/models/location.rb
nimmolo: Oh, riiiiiight!

PR #1572, opened 2023-08-28
Change the "Observation Activity" query title

nimmolo:
The home page query title is currently defined as `"Observation Activity"`, in a special method that overrides the query title in a helper for this query only. While this title is admirably brief, the override makes it harder to grasp the query title methods, which are already labyrinthine.

This PR eliminates the override, so this index can use the same composed translation string as other indexes with a `sort_by` order, and allows the title to be defined upstream by the `Query` class as all other index titles are. 

The main change is that the string for an index sorted by `:rss_log` is now `"Recent Activity"`, not `"Time Last Modified"`. So the homepage title effectively changes to `"Observations by Recent Activity"`.

> The context here is that i'm trying to potentially move the title into a different div of the layout in Bootstrap branch, and to do that I may want to get the components of the index titles returned separately by `Query`, for example `{Observations} {of Boletus edulis}`. This would be easier to accomplish without any overrides of the `Query` title methods.

### DIFF
--- a/app/helpers/title_and_tabset_helper.rb
+++ b/app/helpers/title_and_tabset_helper.rb
@@ -61,19 +61,9 @@ module TitleAndTabsetHelper
             elsif query.num_results.zero? && !no_hits.nil?
               no_hits
             else
-              index_default_title(query)
+              query.title
             end
     add_page_title(title)
-  end
-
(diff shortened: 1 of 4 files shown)


pellaea:
This is a great improvement!

PR #1576, opened 2023-08-29
Fix issues in observations show

nimmolo:
Naming submissions now update the `name_info` partial (thanks @JoeCohen for the issue report)

### Manual test ###

Propose a different name, change the consensus, the `name_info` panel should update.

Fixes alignment of the destroy tabset button.
Fixes confusion in calling the `observation_details` partial, used in several forms and the lightbox. (it wasn't displaying.)

Switches to panels with panel-headings on the show obs page.
Refactors `namings_table` to use panel header/footer.
Prepares new button icon helpers with tooltips, now used in `thumbnail_map` and `images` panels. (Text appears on hover, and in screen readers.)

(Next up:  "Edit", "Destroy" and "Add" links to use icons and tooltips.)

obs_show


nimmolo:
@JoeCohen if you have time for a manual test, I believe this fixes the issue you reported.

JoeCohen:
> @JoeCohen if you have time for a manual test, I believe this fixes the issue you reported.
Works like a charm. Thanks so much!

PR #1577, opened 2023-08-31
Tweaked nginx config to allow robots to use api.

pellaea:
The syntax is a bear because nginx doesn't support complex conditionals, but this seems to work.

### DIFF
--- a/config/etc/nginx.conf
+++ b/config/etc/nginx.conf
@@ -55,6 +55,32 @@ http {
              text/javascript application/x-javascript
              application/atom+xml;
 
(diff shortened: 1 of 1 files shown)


pellaea:
Anyone want to have a look at this?  It is already installed on the production server and it works.  I temporarily blocked curl to test it, and sure enough curl was allowed to use the api but nothing else.

JoeCohen:
@pellaea: Thanks for doing this. 
I think you should merge it. (I don't want to take the time to figure out the syntax.)

PR #1580, opened 2023-09-02
Update .gitignore to ignore `checkpoint_stripped.gz`

nimmolo:
It's convenient to keep the snapshot in place when messing with the db.

If any objections, please don't hold back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ coverage
 /config/master.key
 
 /config/credentials/production.key
+checkpoint_stripped.gz


pellaea:
Totally fine.  Notice we already ignore db/checkpoint.gz.  Although it's hidden in db/.gitignore.  Might not be a bad idea to combine all the .gitignore into this root one.  I thought you needed a separate one in each directory, oops!  Oh wow, but there are a ton of them:

```
./public/sitemap/.gitignore
./public/images/.gitignore
./public/.gitignore
./app/assets/javascripts/.gitignore
./test/reports/.gitignore
./db/.gitignore
./config/locales/.gitignore
./config/.gitignore
./.gitignore
./script/.gitignore
```

nimmolo:
Thanks @pellaea. Will merge. 

Can easily revert if @JoeCohen has objections.

PR #1582, opened 2023-09-02
Remove GraphQL and other unused gems

nimmolo:
Removes the whole GraphQL apparatus, plus the gem `rack-cors`.

Removes `jquery-slick-rails`, an unused carousel gem.
Removes `coffee-rails` for unused coffee.js files.
Removes commented-out gems that are obsolete and have been replaced: `therubyracer`, `turbolinks`


JoeCohen (approved):
Looks good! Thanks for all the cleanup and the `Minitest` fix.

PR #1585, opened 2023-09-02
Glossary wikipedia search

JoeCohen:
A super-simple fix for the broken link to a Wikipedia search for a Glossary Term,
which link appears on the Term pages.

Delivers https://www.pivotaltracker.com/story/show/185904998

### Manual test
* Go to any Glossary Term page, e.g., http://localhost:3000/glossary_terms/451
Result: Page should include a link to a Wikipedia search for the Term

### DIFF
--- a/app/views/glossary_terms/show.html.erb
+++ b/app/views/glossary_terms/show.html.erb
@@ -9,10 +9,7 @@ add_tab_set(glossary_term_show_tabs(term: @glossary_term, user: @user))
 <div class="row mt-3">
   <div class="col-sm-8">
     <%= @glossary_term.description.tpl %>
-    <%= tag.p(
-      link_to("Wikipedia search for '#{@glossary_term.name}'",
-              "https://en.wikipedia.org/w/index.php?search=#{@glossary_term.name.tr(" ", "+")}")
-    ) %>
+    <%= tag.p(link_to(*search_tab_for(:Wikipedia, @glossary_term.name))) %>
   </div>
   <% if @glossary_term.thumb_image %>
     <div class="col-sm-4">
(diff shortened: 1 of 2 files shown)


app/views/glossary_terms/show.html.erb
```diff
@@ -9,7 +9,10 @@ add_tab_set(glossary_term_show_tabs(term: @glossary_term, user: @user))
 <div class="row mt-3">
   <div class="col-sm-8">
     <%= @glossary_term.description.tpl %>
-    <%= tag.p(link_to(search_tab_for(:Wikipedia, @glossary_term.name))) %>
+    <%= tag.p(
```
nimmolo, line 12: @JoeCohen `search_tab_for` is a helper I rewrote. It works, but I forgot to splat it! Should be:

`<%= tag.p(link_to(*search_tab_for(:Wikipedia, @glossary_term.name))) %>`

This produces

`"<a id=\"search_link_to_Wikipedia_thing\" href=\"https://en.wikipedia.org/w/index.php?search=thing\">Wikipedia</a>"`

The reason for the tab helpers is to get a handle on the attributes that can then be be fed to nicer button helpers on the Bootstrap 5 branch. Check the helpers/tabs folder: there is one for pretty much every occasion, and it would be helpful for the refactor if we could use them consistently.

Generally speaking I'm using the word "tab" now to refer to these "link attribute arrays" that contain everything a reusable link needs to be printed in whatever format, and identified in integration tests. So a tab produces:
`[:link_text.l, url, { class: "foo", id: "fah" }]`

JoeCohen: Perfect! That's exactly what I was trying to do, but couldn't figure out how.
Thanks for showing me the way.

JoeCohen: I will merge and deploy this because the buggy current display looks so tacky.
In the the future, I'd like to improve the display so that it has something like the Observation "About this taxon" panel,
with an On MO link to a GlossaryTerm sear, and an On the web link to the Wikipedia search.
But that's for a different PR.

PR #1587, opened 2023-09-03
Refactor `_image_thumbnail` partial as `interactive_image` helper

nimmolo:
Awhile back Jason mentioned partials as a reason for the slowness of our page rendering. I dug into it last night.

Learned a lot from these two :
https://scoutapm.com/blog/performance-impact-of-using-ruby-on-rails-view-partials
https://github.com/rails/rails/issues/41452

There is no doubt that rendering a nested partial within a loop is very slow. This PR gets rid of the nested partials in `matrix_box` — there were in fact both `image_thumbnail` and the nested nested `image_vote_links`! — and refactors the `ThumbnailHelper` and `ThumbnailPresenter` to serve the html via tag helpers.

Renaming these `ImageHelper` and `ImagePresenter` since they're not just about thumbnails, they're for any image.

Also, per @JoeCohen 's recommendation, specifying path prefixes for all partial calls everywhere.


nimmolo:
Wow that was tricky. 

## Manual test: ##
Be sure the vote meter updates on voting!

PR #1588, opened 2023-09-03
Open membership, more automation, and better control for projects

mo-nathan:
Delivers these inter-related stories:

- As a project admin, I want to be able to control whether users can automatically add themselves to that project (an "open project") or if they must ask to join or be invited, so I only have to take an explicit action to add a user if I want to.
- As a project member, I want to be able to add an observation to an open project without enabling other members or even admins to change my observations, so I don't essentially enable anyone to change my observations.
- As a project member, I want to be able to leave a project, so I don't get notifications for the project or get my observations automatically added to that project.
- As a user, I want to be able to add myself to any open project, so I can start associating observations to that project without any action from an admin.
- As a project member, I want to be able to have my observations get automatically added to that project, so I don't have to spend the time and effort to add them manually to each project.
- As a project admin, I want to be able to control whether members can add observations to that project, so I can control the time frame when users can contribute to the project.
- As a project admin, I want to be able to access more precise location information that a project member may have obscured from an observation they have added to that project, so I can access more scientifically accurate data.


app/models/project.rb
```diff
@@ -91,6 +95,23 @@ def is_admin?(user)
...
 
+  def can_join?(user)
+    open_membership && !is_member?(user)
+  end
+
+  def can_leave?(user)
+    is_member?(user) && user.id != user_id
+  end
+
+  def check_box_disabled?(obs, user)
```
pellaea, line 106: Would you mind adding a comment to this?  I'm sure I'll eventually figure out what this means as I continue to read through the PR, but I'm sure I'm not the last person who will be confused when seeing this method out of context!

app/models/project.rb
```diff
@@ -91,6 +95,23 @@ def is_admin?(user)
...
+    !accepting_observations || (obs.user != user &&
+                                !is_member?(user))
+  end
+
+  def enabled?
```
pellaea, line 111: This is just a personal preference: `project.enabled?` by itself doesn't tell me sufficiently what it means. I would honestly prefer in-lining this over the convenience method.  `project.open_membership && project.accepting_observations` is completely clear.  I'm sure it would be possible to elaborate a more descriptive method name than "enabled?" but it might end up being something like "open_and_accepting_observations?" which is scarcely shorter than just in-lining it! :)

app/models/project.rb
```diff
@@ -99,12 +120,26 @@ def self.can_edit?(obj, user)
...
+      next if project.open_membership
       return true if group_ids.member?(project.user_group_id) ||
                      group_ids.member?(project.admin_group_id)
     end
     false
   end
 
+  # Check if this user is an admin for a project that includes
+  # this observation.
+  def self.admin_power?(observation, user)
```
pellaea, line 132: Question that I think I know the answer to, but worth asking: This method can result in two db queries (`user.user_groups` and `observation.projects`).  This could get ugly really fast if running it on a long list of observations and/or users.  But I'm guessing it's only being used right now by the observation form?  If so, no worries.

app/models/project.rb
```diff
@@ -99,12 +120,26 @@ def self.can_edit?(obj, user)
...
+    return false unless user
+    return false if observation.projects.empty?
+
+    group_ids = user.user_groups.map(&:id)
```
pellaea, line 136: Don't we get `User#user_group_ids` for free with the association?  Might need to be `user.user_group_ids.to_a`, though, not sure.

app/views/observations/form/_projects.html.erb
```diff
@@ -13,8 +13,7 @@
       <% @projects.each do |project| %>
         <%= check_box_with_label(form: f_p, field: :"id_#{project.id}",
                                  checked: @project_checks[project.id],
-                                 disabled: @observation.user != @user &&
-                                   !project.is_member?(@user),
+                                 disabled: project.check_box_disabled?(@observation, @user),
```
pellaea, line 16: Hmmm.  I thought maybe you named it `Project#check_box_disabled?` because of some Rails automagic that required that method to exist.  But if you're calling it explicitly, you could name it anything, right?

In that case, can we rename this so that it doesn't look like Project knows details of the create observation view?  I think it would also help because the way it is currently named I found myself asking "which check box?" when I first encountered it in Project.

Oh right, it's that one with the complicated logic.  Am I getting this right? -- The box is disabled if the project is not accepting observations, OR if the session user is not the owner of the observation and is not a member of the project...

Darn, I'm feeling dense right about now.  Wait, but it's only getting to the second clause if it is accepting observations, which means that it must be an open project (?) which means only the owner will ever see this check box, right?  Presumably I'm wrong, but what case(s) am I missing?  Can you help me understand the logic here?  I'm thoroughly confused.

Best to write the comment now while it's fresh in your mind! :)   (Than waiting for Nimmo to ask you what the hell you were thinking 10 years from now.  I know this from personal experience. :)

app/views/observations/show/_observation_details.html.erb
```diff
@@ -17,22 +17,16 @@ caption ||= false
 
 <p class="obs-lat-lng" id="observation_lat_lng">
   <%=
-    hidden = "<i>(#{:show_observation_gps_hidden.t})</i>".html_safe
-    if obs.lat &&
-        (!obs.gps_hidden || obs.can_edit?)
+    if obs.reveal_location?
```
pellaea, line 20: Holy moly, this is a lot cleaner than what was here before!!

nimmolo (approved):

app/models/project.rb
mo-nathan: Reasonable.  I used to have a comment about this getting more complicated due to potentially adding date, location, or taxon constraints.  I'm not convinced this is really the direction anymore, so I agree with getting rid of this.

app/models/project.rb
mo-nathan: Yeah, it's only on the observation show page as far as I know.  I've been wondering if there should be some way for admins (or maybe just owners) to download all the location information, but I expect that would go through a different mechanism.

app/models/project.rb
mo-nathan: Good catch.  I was just copying what was in self.can_edit? above, but what you suggest works.  It makes the same DB queries, but it is already defined.  Not clear why it was that way.

app/views/observations/form/_projects.html.erb
mo-nathan: Accepting Observations is separate from Open Membership (one of the realizations I had yesterday).  The logic about it being a different user is from before this PR so I was just getting logic out of the ERB file.  I agree the name sucks.  It was late.  I'll figure out a better name for it.

app/models/project.rb
mo-nathan: Done

app/views/observations/form/_projects.html.erb
mo-nathan: Better name was easy once I flipped the logic around.

app/views/observations/show/_observation_details.html.erb
mo-nathan: I know, right?

JoeCohen:
#### Question: Would you like me to code and commit a change to fix the following?
(It's a one-line change on the app side plus a test. This is an old issue, so I can't use the GitHub Suggestions to suggest code.)

The list of potential members should be limited to verified users. Otherwise we see junk like in this partial screenshot:

Screenshot 2023-09-04 at 8 10 03 AM

mo-nathan:
Sounds great!  Thanks!

mo-nathan:
This is not behaving correctly on my local.  I get no members when I try going to the `Add Members` page for a project that I own.

JoeCohen:
I just fixed it.
https://github.com/MushroomObserver/mushroom-observer/pull/1588/commits/289f13491f566eb112ce35b5c9fbb1af559eb621
(I had written 'Users.where(verified: false)', but what I needed was
'Users.where.not(verified: nil)`.)

mo-nathan:
Ok, that took me some figuring out.  Didn't realize that we're restring to the last 100 folks to login in the list.  Kind of make sense, but it should be documented on the page.

PR #1590, opened 2023-09-04
Restore "Propose name" button

nimmolo:
I had the wrong class on that div.


mo-nathan:
Verified on my local.

PR #1592, opened 2023-09-05
Improvements to Project navigation

mo-nathan:
Added "Your Projects" link and has_member option to the Projects index page.
project.updated_at gets maintained better.
Default sort changed to project.updated_at.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -186,7 +186,7 @@ class ProjectsController < ApplicationController
     end
   end
 
-  # Display list of user's projects, sorted by date.
+  # Display list of projects with a given member, sorted by date.
   def member
     user = find_obj_or_goto_index(
       model: User, obj_id: params[:member].to_s,
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -9,7 +9,8 @@ class ProjectsController < ApplicationController
   # ApplicationController uses this to dispatch #index to a private method
   @index_subaction_param_keys = [
     :pattern,
-    :by
+    :by,
+    :member
   ].freeze
 
   @index_subaction_dispatch_table = {
@@ -183,6 +184,18 @@ class ProjectsController < ApplicationController
       query = create_query(:Project, :pattern_search, pattern: pattern)
       show_selected_projects(query)
     end
+  end
+
+  # Display list of user's projects, sorted by date.
+  def member
+    user = find_obj_or_goto_index(
+      model: User, obj_id: params[:member].to_s,
+      index_path: projects_path
+    )
+    return unless user
+
+    query = create_query(:Project, :all, member: user)
+    show_selected_projects(query)
   end
 
   # Show selected list of projects.
--- a/app/classes/query/modules/joining.rb
+++ b/app/classes/query/modules/joining.rb
@@ -188,7 +188,9 @@ module Query
           users: :user_id,
           rss_logs: :rss_log_id,
           user_groups: :user_group_id,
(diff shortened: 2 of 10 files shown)


mo-nathan:
Builds on open-projects branch/PR.

app/views/observations/show/_observation_details.html.erb
```diff
@@ -17,22 +17,16 @@ caption ||= false
...
-    else
-      obs.display_alt.t
     end
-  %>
+    %>
   <%=
-    if obs.lat && obs.gps_hidden && obs.can_edit?
-      hidden
+    if obs.lat && obs.gps_hidden
+      "<i>(#{:show_observation_gps_hidden.t})</i>".html_safe
```
nimmolo, line 29: @mo-nathan  - I just realized this partial, `observation_details`, is also called as a sub-partial on the obs index, because i added it to the lightbox caption.

So i'm refactoring the parts of the partial used in the lightbox caption as helpers.

Can you check that my refactor here is the same logically? One difference is i've got it returning nothing, instead of a blank `<p>`, if there's no latitude prop on the observation. 

```ruby

  def observation_details_where_gps(obs:)
    return "" unless obs.lat

    gps_display_link = link_to([obs.display_lat_long.t,
                                obs.display_alt.t,
                                "[#{:click_for_map.t}]"].safe_join(" "),
                               map_observation_path(id: obs.id))
    gps_hidden_msg = tag.i("(#{:show_observation_gps_hidden.t})")

    tag.p(class: "obs-where-gps", id: "observation_where_gps") do
      concat(gps_display_link) if !obs.gps_hidden || obs.can_edit?
      concat(gps_hidden_msg) if obs.gps_hidden
    end
  end
```

Please let me know any modifications if needed, and i'll fix my PR. Then you can revert this change in your PR to avoid a merge conflict.

nimmolo: @mo-nathan — Aha. For your merge you'll have to change this helper to use your new Observation method `obs.reveal_location?`. I don't think i can add that method here because it calls your other new method `obs.project_admin?`.

mo-nathan: Thanks for mentioning the conflict.  Resolved.

app/controllers/projects_controller.rb
```diff
@@ -185,6 +186,18 @@ def pattern
     end
   end
 
+  # Display list of user's projects, sorted by date.
+  def has_member
```
JoeCohen, line 190: >Showing 1 of 1 total issue
Rename has_member to member?. NEW
  def has_member
Severity: MinorFound in [app/controllers/projects_controller.rb](https://codeclimate.com/github/MushroomObserver/mushroom-observer/app/controllers/projects_controller.rb?from_sha=70f70bb7&to_sha=00e724a9) by [rubocop](https://docs.codeclimate.com/docs/rubocop)

If you're looking for a method name that doesn't offend Rubocop, maybe something like:
`enrolled_in`
`roster_includes`
`membership_includes`
`enrollments`
`rostered`
`joined`

mo-nathan: Lol!  Jason and I talked about this and I settled on has_member as my favorite option.  I'll figure out something.

JoeCohen:
Or whatever works for you, and tell rubocop to be quiet.

pellaea:
@mo-nathan -- I made a few small refinements.  Tests are all still passing for me.  Looks great!

(Most of the time was spent figuring out how the heck one even invokes the "member" index to start with, haha!  I miss the good old days when I could guess which method would be invoked by mere inspection of the url. :)

PR #1593, opened 2023-09-05
Convert `observation_details` partial to helper

nimmolo:
I forgot. This is another sub-partial used in the obs index, by the lightbox.

Noticed it in the render log output. (lol, I'm the one who added it!)

With this, I believe we're truly done with sub-partials on the index.

### DIFF
--- a/app/helpers/observations_helper.rb
+++ b/app/helpers/observations_helper.rb
@@ -130,4 +130,74 @@ module ObservationsHelper
       )
     ].safe_join
   end
+
+  # The following sections of the observation_details partial are also needed as
+  # part of the lightbox caption, so that was called on the obs_index as a
+  # sub-partial. Here they're converted to helpers to speed up loading of index
+  def observation_details_when_where_who(obs:)
+    [
+      observation_details_when(obs: obs),
+      observation_details_where(obs: obs),
+      observation_details_where_gps(obs: obs),
+      observation_details_who(obs: obs)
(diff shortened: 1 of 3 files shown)


app/helpers/observations_helper.rb
```diff
@@ -130,4 +130,66 @@ def observation_show_image_links(obs:)
...
+                                "[#{:click_for_map.t}]"].safe_join(" "),
+                               map_observation_path(id: obs.id))
+    gps_hidden_msg = tag.i("(#{:show_observation_gps_hidden.t})")
+
+    tag.p(class: "obs-where-gps", id: "observation_where_gps") do
+      concat(gps_display_link) if !obs.gps_hidden || obs.can_edit?
+      concat(gps_hidden_msg) if obs.gps_hidden
+    end
+  end
+
```
nimmolo, line 179: @mo-nathan  This attempts to duplicate your improved logic on the Projects branch.
If this is OK and merges, you'll need to revert the changes on your branch to avoid a merge conflict.
see https://github.com/MushroomObserver/mushroom-observer/pull/1592/files#r1316471674

nimmolo: Ah i understand now. See follow-up notes in my review of your PR. You'll just want to adjust your PR so this method checks `if obs.reveal_location?` instead of `if !obs.gps_hidden || obs.can_edit?`

PR #1598, opened 2023-09-07
Matrix box carousels

nimmolo:
Check it out. Rudimentary!


nimmolo:
Somehow, with this PR, there's a new n + 1 problem on obs indexes now. 
```
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516697) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  Project Exists? (15.6ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_images` ON `projects`.`id` = `project_images`.`project_id` WHERE `project_images`.`image_id` = 1550689 LIMIT 1
  CACHE User Load (0.0ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 455 LIMIT 1
  ImageVote Load (0.7ms)  SELECT `image_votes`.* FROM `image_votes` WHERE `image_votes`.`image_id` = 1550689
  Image Exists? (0.7ms)  SELECT 1 AS one FROM `images` INNER JOIN `observation_images` ON `images`.`id` = `observation_images`.`image_id` WHERE `observation_images`.`observation_id` = 516696 LIMIT 1
  Image Load (0.7ms)  SELECT `images`.* FROM `images` INNER JOIN `observation_images` ON `images`.`id` = `observation_images`.`image_id` WHERE `observation_images`.`observation_id` = 516696
```

app/presenters/matrix_box_presenter.rb
```diff
@@ -128,8 +135,9 @@ def user_to_presenter(user)
     self.where = user.location if user.location
     return unless user.image_id
 
+    # Not user.images because that's every image they've uploaded
     self.image_data = {
-      image: user.image_id,
+      images: [Image.find(user.image_id)],
```
pellaea, line 140: This is a big problem.  This needs to be passed in somehow or it will result in a query for each matrix box.

app/presenters/matrix_box_presenter.rb
```diff
@@ -51,10 +51,17 @@ def rss_log_to_presenter(rss_log)
...
+                 false
+               else
+                 target.show_link_args
+               end
+
     if target&.respond_to?(:thumb_image) && target&.thumb_image
       self.image_data = {
-        image: target.thumb_image,
-        image_link: target.show_link_args,
+        images: target.images,
```
pellaea, line 63: I'm betting these have not been eager-loaded, either, since it doesn't look like you've changed anything in the controllers in this PR.

Example: for activity log, you'll need to adjust the includes at line 89 in app/controllers/rss_logs_controller.rb.  I suspect you will still need the thumbnail (you still want the user to be able to choose which is the first image of their obs users see), but you'll also need to load images:

```
   includes = {
      article: :user,
      glossary_term: :user,
      location: :user,
      name: :user,
      observation: [:location, :name, :user, **:images,**
                    @user ? { thumb_image: :image_votes } : :thumb_image],
      project: :user,
      species_list: [:location, :user]
    }
```

nimmolo: Got it. Thank you. Committing some updates rn -- sorry for the premature review request.

app/presenters/matrix_box_presenter.rb
nimmolo: Changed

app/presenters/matrix_box_presenter.rb
nimmolo: Quite a bit more loading. It now has to be `{ images: [:image_votes, :license, :user] }`

pellaea:
That makes sense, yeah, it's a lot.  Definitely worth checking how this
affects performance!

app/presenters/matrix_box_presenter.rb
nimmolo: I'm not sure I understand why that is the case, though. Or if that is the case. 
This code only executes in the case of a User matrix_box, and ...hmmm. It does seem the ImagePresenter avoids the lookup.

In `ImagePresenter`, there's this:
```ruby
    # Sometimes it's prohibitive to do the extra join to images table,
    # so we only have image_id. It's still possible to render the image with
    # nothing but the image_id. (But not votes, original name, etc.)
    image, image_id = image.is_a?(Image) ? [image, image.id] : [nil, image]
```
But upstream of both these presenters, aren't we already eager-loading the Image associations for whatever model?

pellaea: `Image.find(...)` is always going to do a query, no matter how many images have been eager-loaded elsewhere.

nimmolo:
In terms of manual testing, please make sure the "hot" areas of the matrix_box images make sense and are usable.

There are quite a few of them for such a small area. 
- fullscreen
- left/right carousel nav
- image votes
- obs_link (everywhere not covered by the above)

app/presenters/matrix_box_presenter.rb
nimmolo: by the way, will Bullet tell you if you're over-eager-loading?

I feel like I'm just chucking associations in there willy-nilly and i'd feel better about it if it warned me when it's oversalted.

pellaea: It can in some cases.  I tried to enable it throughout all the tests, but it is too sensitive.  I think it is turned on for all of the integration tests and a handful of others.  But not every test.

I found the best thing to do was (when working on things that were likely to have an effect) watch the development logs.

nimmolo: Roger. I'm finding one maybe un-optimized thing left on obs index that wasn't caught by Bullet:

Are these image votes, or naming votes? Checking the obs index controller...
```
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516707) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (300.3ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516706) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (297.3ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516705) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516705) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (295.2ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516704) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516704) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (298.6ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516703) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516703) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (300.3ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516702) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516702) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (299.6ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516701) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516701) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (299.7ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516700) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516700) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (297.8ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516699) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (299.6ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516698) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516698) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (296.9ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516697) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516697) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (300.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516696) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516696) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (296.7ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516695) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516695) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (304.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516694) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516694) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
etc...
```

nimmolo: Nope, adding `:votes` didn't affect it.  I did:
```ruby
# app/controllers/observations_controller/index.rb line 192:

    def define_index_args(query, args)
      args = { controller: "/observations",
               action: :index,
               matrix: true,
               include: [:name, :location, :user, :rss_log, **:votes,**
                         { images: [:image_votes, :license, :projects, :user] },
                         { thumb_image: :image_votes }] }.merge(args)
```

nimmolo: Yep, there's something funky going on. 
- Why is the obs index even loading `project_images`?
- Immediately after those is a second batch of `images` and `image_votes`.

The reason it's checking `projects` is that Project now participates in defining `can_edit?` for an image. This is really far afield from the business of the Obs index; maybe i'll try maybe eliminating the image captions here. Oh, nope. Captions are used in the lightbox and encoded into every image as a data attribute! Argh. (or rather "Kwargh!" as Rubocop prefers!)

```
          /vagrant/mushroom-observer/app/models/project.rb:115:in `can_edit?'
          /vagrant/mushroom-observer/app/models/image.rb:919:in `can_edit?'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:472:in `editable_by_user?'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:464:in `correct_user_for_object?'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:459:in `check_permission'
          /vagrant/mushroom-observer/app/controllers/application_controller.rb:462:in `check_permission'
          /vagrant/mushroom-observer/app/helpers/image_helper.rb:75:in `show_original_name?'
          /vagrant/mushroom-observer/app/helpers/image_helper.rb:67:in `image_owner_original_name'
          /vagrant/mushroom-observer/app/helpers/image_helper.rb:56:in `image_info'
          /vagrant/mushroom-observer/app/helpers/carousel_helper.rb:73:in `carousel_caption'
```
```
QUERY starting: "\n      SELECT DISTINCT observations.id\n      FROM `observations` \n      ORDER BY observations.log_updated_at DESC, observations.id DESC\n"
  Observation Load (1.9ms)  SELECT `observations`.* FROM `observations` WHERE `observations`.`id` IN (516551, 516707, 516706, 516705, 516704, 516703, 516702, 516701, 516700, 516699, 516698, 516697, 516696, 516695, 516694, 516693, 516692, 516691, 516690, 516689, 516688, 516687, 516686, 516685)
  Name Load (3.0ms)  SELECT `names`.* FROM `names` WHERE `names`.`id` IN (41809, 112489, 112180, 112490, 112491, 112492, 112493, 112494, 112495, 112496, 112497, 112498, 112499, 112500, 112501, 112502, 112503, 112504)
  Location Load (1.1ms)  SELECT `locations`.* FROM `locations` WHERE `locations`.`id` IN (12848, 6042, 27749, 27747, 3097)
  User Load (1.8ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` IN (8496, 455)
  RssLog Load (0.8ms)  SELECT `rss_logs`.* FROM `rss_logs` WHERE `rss_logs`.`id` IN (628633, 628808, 628809, 628810, 628811, 628813, 628815, 628816, 628818, 628820, 628822, 628823, 628825, 628827, 628828, 628830, 628831, 628833, 628835, 628837, 628839, 628841, 628843, 628845)
  Naming Load (5.1ms)  SELECT `namings`.* FROM `namings` WHERE `namings`.`observation_id` IN (516551, 516685, 516686, 516687, 516688, 516689, 516690, 516691, 516692, 516693, 516694, 516695, 516696, 516697, 516698, 516699, 516700, 516701, 516702, 516703, 516704, 516705, 516706, 516707)
  Vote Load (465.4ms)  SELECT `votes`.* FROM `votes` WHERE `votes`.`naming_id` IN (710892, 711066, 711043, 711044, 711045, 711046, 711047, 711048, 711049, 711050, 711051, 711052, 711053, 711054, 711055, 711056, 711057, 711058, 711059, 711060, 711061, 711062, 711063, 711064, 711065)
  ObservationImage Load (0.6ms)  SELECT `observation_images`.* FROM `observation_images` WHERE `observation_images`.`observation_id` IN (516551, 516685, 516686, 516687, 516688, 516689, 516690, 516691, 516692, 516693, 516694, 516695, 516696, 516697, 516698, 516699, 516700, 516701, 516702, 516703, 516704, 516705, 516706, 516707)
  Image Load (0.6ms)  SELECT `images`.* FROM `images` WHERE `images`.`id` IN (1550397, 1550664, 1550665, 1550666, 1550667, 1550668, 1550669, 1550670, 1550671, 1550672, 1550673, 1550674, 1550675, 1550676, 1550677, 1550678, 1550679, 1550680, 1550681, 1550682, 1550683, 1550684, 1550685, 1550686, 1550687, 1550688, 1550689, 1550690, 1550691, 1550692, 1550693, 1550694, 1550695, 1550696, 1550697, 1550698, 1550699, 1550700, 1550701, 1550702, 1550703, 1550704, 1550705, 1550706, 1550707)
  ImageVote Load (0.7ms)  SELECT `image_votes`.* FROM `image_votes` WHERE `image_votes`.`image_id` IN (1550397, 1550664, 1550665, 1550666, 1550667, 1550668, 1550669, 1550670, 1550671, 1550672, 1550673, 1550674, 1550675, 1550676, 1550677, 1550678, 1550679, 1550680, 1550681, 1550682, 1550683, 1550684, 1550685, 1550686, 1550687, 1550688, 1550689, 1550690, 1550691, 1550692, 1550693, 1550694, 1550695, 1550696, 1550697, 1550698, 1550699, 1550700, 1550701, 1550702, 1550703, 1550704, 1550705, 1550706, 1550707)
  License Load (0.4ms)  SELECT `licenses`.* FROM `licenses` WHERE `licenses`.`id` IN (3, 2)
  ProjectImage Load (23.2ms)  SELECT `project_images`.* FROM `project_images` WHERE `project_images`.`image_id` IN (1550397, 1550664, 1550665, 1550666, 1550667, 1550668, 1550669, 1550670, 1550671, 1550672, 1550673, 1550674, 1550675, 1550676, 1550677, 1550678, 1550679, 1550680, 1550681, 1550682, 1550683, 1550684, 1550685, 1550686, 1550687, 1550688, 1550689, 1550690, 1550691, 1550692, 1550693, 1550694, 1550695, 1550696, 1550697, 1550698, 1550699, 1550700, 1550701, 1550702, 1550703, 1550704, 1550705, 1550706, 1550707)
  CACHE User Load (0.0ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` IN (8496, 455)
  Image Load (1.1ms)  SELECT `images`.* FROM `images` WHERE `images`.`id` IN (1550397, 1550664, 1550667, 1550668, 1550671, 1550672, 1550675, 1550676, 1550678, 1550680, 1550682, 1550684, 1550687, 1550688, 1550690, 1550692, 1550693, 1550695, 1550697, 1550699, 1550702, 1550703, 1550705, 1550706)
  ImageVote Load (0.9ms)  SELECT `image_votes`.* FROM `image_votes` WHERE `image_votes`.`image_id` IN (1550397, 1550664, 1550667, 1550668, 1550671, 1550672, 1550675, 1550676, 1550678, 1550680, 1550682, 1550684, 1550687, 1550688, 1550690, 1550692, 1550693, 1550695, 1550697, 1550699, 1550702, 1550703, 1550705, 1550706)
QUERY finished: model=Observation, flavor=all, params={:by=>"rss_log"}, time=0.771511751
  Rendering layout layouts/application.html.erb
  Rendering observations/index.html.erb within layouts/application
  Rendered application/content/_tab_set.html.erb (Duration: 1.9ms | Allocations: 240)
  Rendered application/content/_sorter.html.erb (Duration: 2.4ms | Allocations: 392)
   (262.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516551) AND (`votes`.`user_id` = 8496) AND (`votes`.`favorite` = TRUE)
  Vote Load (467.3ms)  SELECT `votes`.* FROM `votes` WHERE (`votes`.`observation_id` = 516551) AND (`votes`.`user_id` = 8496) AND (`votes`.`favorite` = TRUE) ORDER BY `votes`.`id` ASC LIMIT 1
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516551) AND (`votes`.`user_id` = 8496) AND (`votes`.`favorite` = TRUE)
  Vote Load (413.4ms)  SELECT `votes`.* FROM `votes` WHERE (`votes`.`observation_id` = 516551) AND (`votes`.`user_id` = 8496) AND (`votes`.`favorite` = TRUE)
  Naming Load (0.4ms)  SELECT `namings`.* FROM `namings` WHERE `namings`.`id` = 710892 LIMIT 1
  Name Load (0.6ms)  SELECT `names`.* FROM `names` WHERE `names`.`id` = 25237 LIMIT 1
   (270.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516707) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516707) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (286.1ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516706) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (285.5ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516705) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516705) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (281.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516704) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516704) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (278.8ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516703) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516703) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (273.5ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516702) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516702) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (278.6ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516701) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516701) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (276.2ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516700) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516700) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (287.5ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516699) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (271.2ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516698) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516698) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (274.7ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516697) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516697) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (279.8ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516696) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516696) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (278.5ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516695) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516695) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (279.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516694) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516694) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (280.6ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516693) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516693) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (278.2ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516692) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516692) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (275.9ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516691) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516691) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (275.8ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516690) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516690) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (274.5ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516689) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516689) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (283.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516688) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516688) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (278.4ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516687) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516687) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (264.2ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516686) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516686) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
   (280.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516685) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `votes` WHERE (`votes`.`observation_id` = 516685) AND (`votes`.`user_id` = 455) AND (`votes`.`favorite` = TRUE)
```

nimmolo: Intriguing side-note from my reading at the moment — Turbo and Stimulus offer a way to lazy-load the image caption text from the `ImagesController#show` action `format:turbo_stream` response into a blank lightbox caption, when the lightbox hits that image. That would eliminate having to render the captions ahead of time and get rid of a lot of these includes.

The same is possible for the image voting interface, it can be lazy-loaded too.

pellaea: I assume that lazy-loading still means separate queries for each image, only it does it later, right?  But presumably it may never need to if a user doesn't flip through most of the carousels?  It's a tricky call.  The user experience will include long delays as it loads individual image information.  Or even if it lazy-loads the info for all of the images for one carousel at once, there will still be at least one few-second delay each time the user goes to look at images in a carousel?  That figure -- "few-second" -- may be my unique experience, though, based on poor internet connectivity, and not a fair characterization of most users' experience?

JoeCohen:
Is `obs_link` supposed to be part of the image?
(In my manual testing, only fullscreen, left/right carousel nav, and image votes are hot areas in the image. The consensus id and the obs id link to the image. I'm OK with doing it that way.)

Chrome version 116.0.5845.179 (Official Build) (x86_64)
```ruby
vagrant@ubuntu-focal:/vagrant/mushroom-observer$ git log -1
commit a1eb3c639a42b24f4e70f6f4584be715bc609c52 (HEAD -> nimmo-matrix-box-carousels, origin/nimmo-matrix-box-carousels)
Merge: 237e2d533 d4d57f962
Author: andrew nimmo <andrnimm@fastmail.fm>
Date:   Thu Sep 7 13:19:48 2023 -0700

    Merge branch 'main' into nimmo-matrix-box-carousels
```
> In terms of manual testing, please make sure the "hot" areas of the matrix_box images make sense and are usable.
> 
> There are quite a few of them for such a small area.
> 
> * fullscreen
> * left/right carousel nav
> * image votes
> * obs_link (everywhere not covered by the above)
> 
> It would also be nice to benchmark the load time for this page vs. main. Volunteers encouraged!

app/presenters/matrix_box_presenter.rb
nimmolo: Going to mark the original issue as resolved. The query weirdness was a separate problem arising from the `show_obs_title`, fixed this week.

In this case lazy-loading does mean that if the user doesn't flip through the carousel, they won't load the image.

nimmolo:
@JoeCohen - you're right, the obs_link is not on the carousels.  I agree, it's ok this way.

nimmolo:
@JoeCohen ... aha, i figured out where I'd experimentally disabled it. The obs_link is back working now.

If you have time to check it out, please let me know if you think that's better or not. I do feel like people expect it at this point, but maybe not.

JoeCohen:
It's all good. (The obs_link works fine in the carousel.) 
And it's better this way than jiust outside the carousel.
Very nice!
> @JoeCohen ... aha, i figured out where I'd experimentally disabled it. The obs_link is back working now.
> 
> If you have time to check it out, please let me know if you think that's better or not. I do feel like people expect it at this point, but maybe not.

JoeCohen (approved):
As @nathan says, sweet.

pellaea:
Overall I think this looks great, but since you asked, here is some criticism.  I don't know how to do this in a way that doesn't sound like a litany of negative complaints, sorry.  Best I could do is express them as requests and/or preferences.  Hope that's good enough!

1. First thing I did was watch the logs.  There are a whole bunch of identical simple queries, of two sorts:

```
CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
```

and 

```
CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_obser
vations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observati
ons`.`observation_id` = 515596 LIMIT 1
```

Since they are mostly repeats, they are cached and take < a millisecond, so maybe "who cares?" But this feels like something that could become important later when it may be less easy to catch.  Or maybe it will go away on its own without us ever having to deal with.  Clearly this is not ideal, but maybe not a problem.  For now, just a note.

2. I see it rendering `observations/namings/_propose_button.html.erb` and `observation_views/_mark_as_reviewed.html.erb` each once for each matrix box... but I can't find out how to actually see these buttons! :)  Either these partials are inaccessible in the UI and therefore should be removed, or the UI needs tweaking so that I can find them.  I can't find them even though I _know_ they should be there, haha!  (Also, related, even though the partials take ca. 2 ms to load locally, and even multiplied by 100 this is still only 0.1-0.2 sec, it nevertheless may be worth considering in-lining or whatever the trick is you used to eliminate partials elsewhere that was such a resounding success.)

3. I notice that one big difference using the carousel is that you used to be able to click on the image to get to the observation page.  But now, the only way to do that is by clicking on the mushroom name.  I understand that clicking on the hot areas on left and right will now do something different.  But is it really necessary (for technical or design reasons) to make clicking in the middle do nothing now?  [Oh, I see this is specifically what you were asking re: hot areas.  My experience agrees mostly with Joe's in that there is no hot area in the matrix box image which takes you to the observation page.  The hot areas for the image votes, left and right carousel, and expando icon are all fine.]

4. I notice that when in expando-view, the carousel cycles through all of the observations and you can't tell where one observation ends and the next begins.  (Well, okay, fine, if your window is big enough, you can see the text below the image, and if you pay close attention you can see it there, but it's subtle at best.  This had me totally fooled for a while.  It feels unnatural and surprising to me.)  But the carousels in the main view do not.  I would personally much prefer both expanded and main view carousels behave the same.  And I would personally prefer the carousel showing only one observation at a time, and not wrapping around.  Just my personal preference / vote.  Having them behave the same is far more important than details of how they behave.  (Although, of course, it makes no sense to have the carousels cycle through all of the observations in the main view!!)  The wrap-around thing is even less important to me.

5. FWIW, I preferred the style of the image votes widget in the old version, where it extends all the way across the image and doesn't have rounded corners.

6. I notice also that clicking on the name in the main view takes you to the observation page, but clicking on the name in the expando view takes you to the name page.  These should be the same, I think.  In fact, I can't find any way to click through to the observation page from within the expando view, but maybe I'm missing it?

7. Ah!  I see another difference.  The new carousels enforce all the "cards" being the same size, whereas before we allowed the vertical-oriented images to be taller.  Personally, I'm fine with this, just calling attention to the difference in case it's important to others.  (I remember this being a point of considerable debate in the past!)

pellaea:
PS. re: performance -- I just realized that my laptop blazes compared to the production server, so my back-of-the-envelope calculations leading me to conclude that those superfluous(?) cached queries are inconsequential may not be accurate estimates of real cost.

PPS. I'm running firefox 116.0.3 on ubuntu 5.15.0-82.

nimmolo:
Thank you for all the detail here. I assumed the query stuff was solved by that emergency PR midweek, and I've been focused on the Rails 7 branch, which i'm pleased as peach about.

I haven't tracked down the little queries, but one thing i’m noticing:  calling `count` on an association seems to execute a query every time. Whereas it seems if we already have the association via `includes`, `length` will count the cached records. Did you guys know that?

Try it:
```
rails c --sandbox

obs = Observation.last.includes(:collection_numbers)
Observation Load (0.4ms)  SELECT `observations`.* FROM `observations` ORDER BY `observations`.`id` DESC LIMIT 1

status = (obs.collection_numbers.length > 1) ? "collected" : "not_cool"
"not_cool"

status = (obs.collection_numbers.count > 1) ? "collected" : "not_cool"
CollectionNumber Load (3.1ms)  SELECT `collection_numbers`.* FROM `collection_numbers` INNER JOIN `observation_collection_numbers` ON `collection_numbers`.`id` = `observation_collection_numbers`.`collection_number_id` WHERE `observation_collection_numbers`.`observation_id` = 516707
"not_cool"
```

I can do a separate PR eradicating `count` in these situations right now.

nimmolo:
Pinging @mo-nathan on the `count` issue because his recent visual groups code also contains association counts. 

Note that you do need to call `count` for something like this with a proc. `length` is not a substitute in this case:
    namings.count { |n| n.name == name }.positive?

mo-nathan:
I'm not sure which of my code you're referring to regarding count vs. length, but it wouldn't surprise me if I got it wrong somewhere.  I looked through my last two PRs and didn't see anything obvious.  However, the rule as I understand it is that `length` is Ruby and `count` is SQL.  However, if you do something like `namings.count { |n| n.name == name }.positive?` it does not work well since it is running Rails code on something that should be handled in SQL.  I believe the above will typically create and N+1 issue.  I just tried `Naming.all.count { |n| n.name == 'Russula' }.positive?` and it started spewing tons of SQL (one for each Naming).  However `Naming.where(name: 'Russua').count` just runs one query and quickly gives you the correct answer.  I avoid Ruby procs in queries like the plague exactly because of issues like this.  However, if you need the objects loaded into memory, then it totally make sense to do that in one query that hits SQL and then do any other counting and filter using Ruby to avoid additional queries.

As a side note, all of the above primarily applies to application code.  Test code is a bit different.  There may well be good reason to reconsult the database using `count` if the data may have been changed as a result of running the code in the test.  In that case I lean towards hitting the database since it should really make no difference from a performance level and I'd rather be sure what I'm looking at is based on what got saved to the database since I've been burned way too many times due to caching when writing tests.

Does any of that address the question?

mo-nathan:
Looking at the visual group code, the only place where I see `.count` getting used in application code is:
```
  def image_count(status = true)
    return visual_group_images.count if status.nil? || status == "needs_review"

    if status && status != "excluded"
      return visual_group_images.where(included: true).count
    end

    visual_group_images.where(included: false).count
  end
```
In this case I would argue that `.count` is appropriate because the function doesn't know whether `visual_group_images` has been loaded from SQL or if that data is expected to be used in this transaction.  If it hasn't been used, then using `length` will load all that data into the server before counting it.  Using `count` will just query the database and return a single number.  In the cases I was able to find were `image_count` is called it is being called for all the images in the images in a visual group, but the page is only display a small subset of those image.  If you have a large visual group (like "Non-Diagnostic" in the "Big List" model which have more than 48,000 members), you don't want to load all that data just to get the count.

nimmolo:
Thanks @mo-nathan. 

That example you gave from `observation.rb` is a good one. Thinking what to do about this method:

```ruby
  # Has anyone proposed a given Name yet for this observation?
  def name_been_proposed?(name)
    namings.count { |n| n.name == name }.positive?
  end
```

How about:
```ruby
  # Has anyone proposed a given Name yet for this observation?
  def name_been_proposed?(name)
    namings.where(name: name).present?
  end
```
This seems pretty minimal.

You're right - the hits I found in your visual groups were a false positive, those should stay as they are.

Will add all this to a GitHub Discussion so it's available for reference.

mo-nathan:
For:
```
  # Has anyone proposed a given Name yet for this observation?
  def name_been_proposed?(name)
    namings.count { |n| n.name == name }.positive?
  end
```
I find the alternative code you propsed to be clearer.

Testing it out, it does appear that the namings are already loaded, so your code does generate an extra query over the existing code.  In this case I don't think that really matters since the amount of data is very small.  If it did, I'd lean towards `length` rather than `count` (which also doesn't generate any additional queries in this case).

nimmolo:
Thanks for the explanation.  I will revert that change on the other PR then, because that PR is all about removing extra queries, even small ones.

mo-nathan:
I tried this out and found that on my local o airport wifi, it was noticeably slower both using the developer tools and my own user experience.  I think it went up from about 1.5 seconds to 2.5 seconds for the initial load and then the image loads felt a bit slower as well.  For me the extra time didn't really feel worth it, but I expect I'd adapt to it.  Unfortunately the internet is bad enough here that I can't really get more precise timing data.

nimmolo:
@pellaea 
> 1. First thing I did was watch the logs.  There are a whole bunch of identical simple queries, of two sorts:
I definitely want to sort this out. This was the original thing that i noticed. Why are `projects` even getting loaded on this view?
 
> 2. I see it rendering `observations/namings/_propose_button.html.erb` and `observation_views/_mark_as_reviewed.html.erb` each once for each matrix box... 
Need to fix!

> 3. I notice that one big difference using the carousel is that you used to be able to click on the image to get to the observation page.  
This is restored now on the current branch of this PR.
> 4. I notice that when in expando-view, the carousel cycles through all of the observations and you can't tell where one observation ends and the next begins.  
This was a longstanding request of Alan's. I don't mind it like this (i.e. as he requested), but it's easy to change.
> 5. FWIW, I preferred the style of the image votes widget in the old version, where it extends all the way across the image and doesn't have rounded corners.
Will check this, not sure it's showing like that for me.
> 6. I notice also that clicking on the name in the main view takes you to the observation page, but clicking on the name in the expando view takes you to the name page.  These should be the same, I think.  In fact, I can't find any way to click through to the observation page from within the expando view, but maybe I'm missing it?
Will consider adding a link to the obs in the lightbox.

mo-nathan:
Was able to get some better data.  On main I was seeing the initial page load at just over 2 seconds with subsequent loads at 1.6 seconds.  On this branch the initial load was just over 3 seconds with subsequent loads around 2.6 seconds.

pellaea:
@nimmolo -- I can take a closer look at the queries on the home page tomorrow maybe.  I confess that I get a bit lost now that you've reorganized all of my code so thoroughly! ;)  But I gotta learn what you've done eventually...

Also re: expando box scrolling continuously through all the observations -- I was thinking (for a future PR!) that one way to have the best of all worlds would be to have an animation of some sort "swipe" between images, and then add some sort of obvious spacer between the last image of the observation you're on and the first image of the next observation.  That would be more than enough of a visual clue to keep me happy.  I don't need much, just throw me the occasional crumb. ;)  Just anything to signal to the viewer that they're moving on to another observation.

nimmolo:
@pellaea Just eliminated the partials for the "propose naming button" and "mark as reviewed" toggle. Thanks for catching that!

nimmolo:
I like the idea of the lightbox having some marker between observations.

Possibly easier to implement: a corner overlay on the image with the current obs ID.

pellaea:
Oh, I like the overlay idea, that gives ideas.  The ID is still going to be too subtle for me.  (I know, I know, so picky...)  But... what about a narrow vertical bar overlay?  One bar on the left edge of the first image of any observation, another bar on the right edge of the last image of the observation?  It looks like a black background in the lightbox no matter what the theme (right?), so a white vertical bar that extends above and below the image a bit would be pretty noticeable.  Actually, it could even be restricted to just the text part below the image for that matter, and handled entirely with CSS.

Anyway, not a blocker, just ideas for that frabjous day when we miraculously have nothing better to do.

nimmolo:
@mo-nathan Thank you for testing load time out — and in the airport!

My load time on `<main>` is 25 seconds (!!!) for the first load of the index, and ~6.5 seconds thereafter.
On this branch it's 15 seconds for the first load, and ~6.5s after that. So it seems improved somehow.

Note: I've just eliminated two lurking nested partials that I wasn't aware were being called on the regular index. (They're part of the identify interface, but they DO load on unidentified observations in the lightbox on the regular index.) 

This may have sped up load times. I didn't A/B test before.

nimmolo:
> Oh, I like the overlay idea, that gives ideas. The ID is still going to be too subtle for me. (I know, I know, so picky...) But... what about a narrow vertical bar overlay? 

I like this UX idea a lot.

nimmolo:
@pellaea I'm no longer getting the little extra queries you pointed out on the home page. (I was before).

pellaea:
> @pellaea I'm no longer getting the little extra queries you pointed out on the home page. (I was before).

Right.  I suspect that this is somewhat specific.  See this log based on a (very) old db snapshot (snapshots are too big for me to download easily anymore, so I'm kinda stuck until I find a better internet connection):

```
...
QUERY finished: model=Observation, flavor=all, params={:by=>"rss_log"}, time=0.325337217
  Rendering layout layouts/application.html.erb
  Rendering observations/index.html.erb within layouts/application
  Rendered application/content/_tab_set.html.erb (Duration: 0.7ms | Allocations: 240)
  Rendered application/content/_sorter.html.erb (Duration: 1.1ms | Allocations: 392)
  Synonym Load (2.9ms)  SELECT `synonyms`.* FROM `synonyms` WHERE `synonyms`.`id` = 5702 LIMIT 1
  Name Load (101.4ms)  SELECT `names`.* FROM `names` WHERE `names`.`synonym_id` = 5702
   (195.8ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  CACHE  (0.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`name_id` = 59162
  Project Exists? (8.3ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515597 LIMIT 1
  Project Exists? (4.5ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 515596 LIMIT 1
  Project Exists? (3.3ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 339050 LIMIT 1
  Project Load (5.0ms)  SELECT `projects`.* FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 339050
  Rendered collection of shared/_matrix_box.html.erb within shared/_matrix_table [80 times] (Duration: 3027.3ms | Allocations: 2459592)
  Rendered shared/_form_name_autocomplete.html.erb (Duration: 5.3ms | Allocations: 1427)
  Rendered observations/index.html.erb within layouts/application (Duration: 3060.9ms | Allocations: 2476361)
  Rendered application/app/_head.html.erb (Duration: 369.9ms | Allocations: 54789)
 ```

See how most of the `select count from observations where name_id = 59162` are identical, and the same name?  This is really screwy.  It appears to somehow be precipitated by this synonym thing, wherever that's coming from, then it counts the number of observations of this synonym 14 times.  [This appears to all be coming from a single observation, 505457, Amanita lavendula group.   But there are way more than 14 matrix boxes after that observation.  Why 14 times??  Also, check out that observation on the live site.  The title is fubar.]

And it's also checking for projects for the same observation 515597 a bunch of times.  Again, WTF?  It doesn't appear to be associated with any projects or anything special.  Huh?

It's all happening within the _matrix_box collection somehow.

Bizarre!

**Comment by nimmolo** [comment]
What in the Sam Hill? Investigating the title issue now on obs 505457.

**Comment by nimmolo** [comment]
Ok, I don't seem to get the project cache query nonsense, although i did get it yesterday. 
Are you sure you've pulled the latest version of this PR branch?

**Comment by pellaea** [comment]
I'll have to look into this tomorrow.  I can't reach github right now.  But
yes, I'm sure I pulled the latest before testing it.  I'll see if I can
track down where that stuff is coming from.

**Comment by nimmolo** [comment]
Sounds good. PR is in for the title fix.

**Comment by mo-nathan** [comment]
I woke up way too early and couldn't get back to sleep, so I ran some more tests from my hotel room wifi.  This branch is still consistently slower for me than `main` on the home page.  Approximately 1.75x for the handful of pages I tested.  I also noticed that this branch does a bunch more processing after the initial page load that can persist for up to a minute.  It doesn't look like it's hitting the server at all, but it still makes me a bit nervous once we have bunch of folks hitting the same server.

I also noticed that on the branch there a bunch of addition redundant SQL queries that look like this:
`  CACHE Project Exists? (0.0ms)  SELECT 1 AS one FROM `projects` INNER JOIN `project_observations` ON `projects`.`id` = `project_observations`.`project_id` WHERE `project_observations`.`observation_id` = 528117 LIMIT 1`
It's happening between:
`  Rendered application/content/_sorter.html.erb (Duration: 0.4ms | Allocations: 380)`
and
`  Rendered collection of shared/_matrix_box.html.erb within shared/_matrix_table [144 times] (Duration: 1749.0ms | Allocations: 4321331)`

**Comment by nimmolo** [comment]
@mo-nathan Thanks for testing it again. 

I definitely used to get those extra SQL queries, but be sure you have pulled the latest from this branch. As the latest committer i'm pretty sure i've got the latest version!

I'm getting this just now, in the relevant part of page load you cited:
```
QUERY finished: model=Observation, flavor=all, params={:by=>"rss_log"}, time=0.294038279
  Rendering layout layouts/application.html.erb
  Rendering observations/index.html.erb within layouts/application
  Rendered application/content/_tab_set.html.erb (Duration: 4.3ms | Allocations: 240)
  Rendered application/content/_sorter.html.erb (Duration: 5.1ms | Allocations: 392)
  Rendered collection of shared/_matrix_box.html.erb within shared/_matrix_table [24 times] (Duration: 606.6ms | Allocations: 295303)
  Rendered shared/_form_name_autocomplete.html.erb (Duration: 14.6ms | Allocations: 1450)
  Rendered observations/index.html.erb within layouts/application (Duration: 684.7ms | Allocations: 312086)
  Rendered application/app/_head.html.erb (Duration: 7598.6ms | Allocations: 76531)
```
No extra queries.

Also want to say that despite having worked on it, i'm in no rush to merge this PR. I certainly do not want it to slow the site down that much.

**Comment by nimmolo** [comment]
Ohhh... yep. I got it now. You've got an older DB snapshot, like Jason, I bet. I went back to the index near what's "current" for his snapshot, and yes I do get the weird extra queries.

**Comment by nimmolo** [comment]
Aha. I believe the problem occurs in the lightbox printing the observation title with ALL the bells and whistles including preferred name link, owner naming, deprecated name...

And then there's the observation-details section below it, with so many links there too. 

It was convenient to grab that partial for the lightbox caption, but maybe what we need instead is something like the obs details in plain text with no links, and a single link to the obs.

**Comment by nimmolo** [comment]
To reproduce the issue, go to http://localhost:3000/505457, and then hit "Index" to be sure we're on the same page of the index. Many extra queries will fire. If you examine the lightbox caption for that observation (it's encoded as a data-attribute in each image) you'll find the ID's of the associations that are firing the extra queries. 

This is an issue in the lightbox currently, and it's being multiplied by having more images on the index. I'll try a separate PR to simplify the lightbox caption text, upstream, outside of this carousel PR.

**Comment by nimmolo** [comment]
Experimentally merging the lightbox caption PR into this to check if it helps.

**Comment by pellaea** [comment]
Yes!  Simplify the lightbox caption!  It would totally make sense that's where some of this project-related stuff is coming from.  So glad you found that.

**Comment by mo-nathan** [comment]
I'm still seeing the weird queries on the latest version of `nimmo-matrix-box-carousels`.  It's still slower especially if you count all the loads of the images.  I've attached screenshots.  I would guess the main reasons I'm seeing faster speeds are the M1 chip and not running in any sort of VM.

Screenshot 2023-09-11 at 10 40 08 PM
Screenshot 2023-09-11 at 10 36 57 PM

**Comment by mo-nathan** [comment]
Also as shown in the images, the database I'm running on is from September 7 and I have it configured to display 144 images.

**Comment by nimmolo** [comment]
@mo-nathan Thanks, i'll move the index to these posts and check queries. 

Side question. How can I get on Docker, if that's what youre doing? That's a huge difference.

**Comment by JoeCohen** [comment]
I think Nathan is running Ruby directly on his laptop -- no Virtual Box, no
Docker.

**Comment by nimmolo** [comment]
I am unable to reproduce any weird query issue, and i've paged through the first 15 pages of 150 obs each on the current version of this branch. 

Note that images are expected to keep loading after page load, because in most cases they're lazy loaded. Other assets are getting lazyloaded too, like icons and arrows, and I don't know why that would be but it's probably not different on this branch.

Nathan's using a newer snapshot than I have, though. I'm not sure what's different there. The screenshots don't show the queries

**Comment by pellaea** [comment]
Hey, I just got done with what I was doing tonight.  I'll take a look and
see if I can track this down, because I was able to reproduce it earlier.

**Comment by pellaea** [comment]
I found the problem with the projects.  There are two things that require projects:

1. showing obs lat/long requires observations.projects to determine if the viewer has permission to see it
2. showing the original file name requires images.projects to determine if the viewer has permission to see it (note the need to eager-load both the observations' projects as well as the observations' images' projects -- we were missing the former) 

I fixed it so that it doesn't need to eager-load observations.thumb_image.  It's already eager-loading all of the images, so there's no reason to eager-load the thumbnail twice.

Lastly, I noticed that the carousels really slow down the page a lot.  I debugged this by turning off the matrix boxes completely, then adding it back piece by piece.  The eager-loading alone is only like 1 to 1.5 seconds on my laptop.  But by time I re-enabled all of the carousel stuff it was up to about 3.5 seconds.  This is all 100% rendering time, not database time.  And no partials are involved either.  Just pure, straight ruby.  And lots of it.  This has become very complex.  Not sure how to fix it.

I'm pushing my changes now.

File: app/helpers/lightbox_helper.rb [hunk 3]
```diff
@@ -17,7 +17,7 @@ def lightbox_caption_html(lightbox_data)
     obs_data = lightbox_data[:obs_data]
     html = []
     if obs_data[:id].present?
-      html = lightbox_obs_caption(html, obs_data, lightbox_data[:identify])
+      lightbox_obs_caption(html, obs_data, lightbox_data[:identify])
```
nimmolo, line 20: Not sure this is equivalent

nimmolo:
@pellaea thanks for checking this out so thoroughly. I will merge some of those last tweaks into main tomorrow, after examining more closely - they should probably go independently. 

Question - **do we need the image copyright info on the index lightbox?** I knew this was a doozy of a multiple join. The question is, can we ditch it here? Cache it on a column?

This PR has certainly shed light if nothing else. 

Might be a good opportunity to try out ViewComponents. I can make a PR off this branch turning the obs matrix box partial and helpers into VC's, and we compare speed then.

mo-nathan:
I couldn't get VirtualBox working when I first got my M1.  I tried Docker, but it increased the time it took for tests to run to something like 20 mins.  I tweaked it a bit and it got a bit better, but never back to the level it had been using VirtualBox.  I then ran into this weird RedCloth issue using Docker on M1, so I threw in the towel and just ran it natively.  It's so fast, I haven't looked back.  Another good metric that might be worth comparing is that for me the full test suite runs in under 2 minutes:
```
% rails test
Started with run options --seed 46145

  2307/2307: [==================================================================================================================] 100% Time: 00:01:48, Time: 00:01:48

Finished in 108.31886s
2307 tests, 27162 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Minitest to /Users/nathan/src/mushroom-observer/coverage. 24539 / 26198 LOC (93.67%) covered.
```

mo-nathan:
It looks like Jason's fix got rid of the weird queries.  It's still noticeably slower (2.5 seconds vs. 1.5 seconds).  The most noticeable thing is the image relead when I do a hard reload.  On main it still must be using some sort of cache since it only adds about half a second.  However, on the branch it takes a full minute to load the images for the 144 observations.  All of that is browser time, so I'm a bit more concerned about the server load time, but I'd be ok with at least trying it out.

app/helpers/lightbox_helper.rb
pellaea: By accident it is.  The point is that lightbox_obs_caption just adds stuff to html, it doesn't really return anything.  Apparently `html << blah` returns html as a side-effect, so it's okay that you were replacing html with, well, html.  But it bit me in the foot and confused me mightily when I'd commented out everything in lightbox_obs_caption.  If you really want it to take the form of lightbox_obs_caption returning something, instead of changing an argument in place, then please rewrite it as:

```
html += lightbox_obs_caption(obs_data, blah)
...
def lightbox_obs_caption(obs_data, identify)
  html = []
  if blah
    html << blah
    html << blah
  end
  html << blah
  html << blah
  html
end
```

I think that would be much less confusing, personally.

Tiny point, but it just confused and wasted some of my time last night, and this tiny change would have avoided that.

pellaea:
Very interesting point about image loading.  I thought we were lazy-loading image?

But the observation about the page being almost twice as slow is comparable to my own observation.

I don't think the extra includes (observation -> projects, images -> users, images -> projects) add significantly to the query time.  Initially I ripped out all of the includes that I could, assuming that was going to help.  But surprisingly, even after all of that I was still seeing the same ca. 3.5 second page load.  Maybe a quarter of a second faster without several extra eager-loads?  Not worth the extra complexity required to render different content on some pages than on other pages, in my opinion.  So I just left those two "XXX consider dropping these on indexes" comments, but otherwise reverted as far as I could to Nimmo's original code.  Because it was useful seeing where those projects were required.

nimmolo:
> I couldn't get VirtualBox working when I first got my M1. I tried Docker, but it increased the time it took for tests to run to something like 20 mins. I tweaked it a bit and it got a bit better, but never back to the level it had been using VirtualBox. I then ran into this weird RedCloth issue using Docker on M1, so I threw in the towel and just ran it natively. It's so fast, I haven't looked back. Another good metric that might be worth comparing is that for me the full test suite runs in under 2 minutes:
> 
> ```
> % rails test
> Started with run options --seed 46145
> 
>   2307/2307: [==================================================================================================================] 100% Time: 00:01:48, Time: 00:01:48
> 
> Finished in 108.31886s
> 2307 tests, 27162 assertions, 0 failures, 0 errors, 0 skips
> Coverage report generated for Minitest to /Users/nathan/src/mushroom-observer/coverage. 24539 / 26198 LOC (93.67%) covered.
> ```

That's so fast! Did you just add `rbenv`, the db, bundle gems in the directory and that's it? 

Why am I not doing this? Can we put instructions in the docs? Vagrant crashes on me like 12 times a day and i have to reload (not restart the Rails server!)

app/helpers/lightbox_helper.rb
nimmolo: Thanks. That `=` was definitely a mistake

nimmolo, line 20:
```suggestion
      html << lightbox_obs_caption(html, obs_data, lightbox_data[:identify])
```

nimmolo: Oh wait, no, it was just confusing style. Now i get what you're saying.

pellaea: Good!

(But I do love the `html << lightbox_obs_caption(html)` idea.  I wondered briefly if the universe would implode if you tried to push an Array inside of itself. ;)

nimmolo:
Having these carousels is always going to be a tradeoff between avoiding round-trips to show obs, and slowing down the index itself. The carousels will always be slower, the question is if it's tolerable.

However, maybe this should wait until i can rewrite the image partials as ViewComponents. This should speed up the index quite a lot — then maybe the carousels will not make the higher load time intolerable.

PR #1599, opened 2023-09-07
Make integration test names consistently end with "IntegrationTest"

nimmolo:
So we can easily run them all at once with `rails t -n /IntegrationTest/`

When troubleshooting eager-loading, it's very helpful to run these all at once, because Bullet seems to only run on these tests.

### DIFF
--- a/test/integration/capybara/account_integration_test.rb
+++ b/test/integration/capybara/account_integration_test.rb
@@ -2,7 +2,7 @@
 
 require("test_helper")
 
-class AccountTest < CapybaraIntegrationTestCase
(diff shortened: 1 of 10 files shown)


JoeCohen:
Sorry I didn't respond earlier.
There's no harm in doing this. But it's unnecessary for this subset of tests.
You can specify which tests to run by giving it a path to a directory, which run all tests in that directory and subdirectory.
`rails t test/integration`
See https://guides.rubyonrails.org/v6.1/testing.html#the-rails-test-runner

PR #1602, opened 2023-09-07
Reverted an accidental rubocop "fix" that breaks a cronjob.

pellaea:
Added obnoxious comments to help us remember not to repeat this again!

### DIFF
--- a/app/classes/ip_stats.rb
+++ b/app/classes/ip_stats.rb
@@ -82,7 +82,7 @@ class IpStats
 
     def blocked?(ip)
       populate_blocked_ips unless blocked_ips_current?
-      @@blocked_ips.include?(ip) && @@okay_ips.exclude?(ip)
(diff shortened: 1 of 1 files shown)


JoeCohen:
My apologies.
When I check blame, this happened in one of my PRs on Apr 12 2023.
I remember discussing this with Jason around that time.
I need to dig into this to understand how it got left/put into main.

pellaea:
Oh, no worries, Joe!  It would be all too easy to do.  The comment was
still there at the top of the file, but you wouldn't necessarily notice it
when you fixed the `!include?` statements below.  That should not be an
issue now.  I've added big all-caps comments "DO NOT FIX!" :)

JoeCohen:
Here's what I think happened.
I ran Rubocop locally in way that both

   - ignored `#rubocop:disable Rails/NegateInclude`, and
   - auto"corrected" the offending lines.

I'll dig into this a bit more to see how to improve my workflow (or the
file).

Joe

PR #1603, opened 2023-09-07
Update `login_system.rb` to allow requests from our `ip_stats` script

nimmolo:
The goal is to allow that script to access any page it likes without being logged in.

Question: is `request` available to this class?

### DIFF
--- a/app/classes/login_system.rb
+++ b/app/classes/login_system.rb
@@ -4,35 +4,7 @@
 module LoginSystem
   protected
 
-  # overwrite this if you want to restrict access to only a few actions
-  # or if you want to check if the user has the correct rights
-  # example:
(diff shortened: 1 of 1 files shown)


pellaea:
Why not?

nimmolo:
@pellaea  ahaa, `Admin::Emails::FeaturesControllerTest#test_features` was still failing.

That's because we actually do use `authorize?` in the Admin controller. Maybe you wrote that to facilitate using the login system for my controller refactor, now that i think about it.

This should be good to go, now. Lookin forward to finding out how we stack up against April!

pellaea:
Oooooh, thanks, I'll deploy now.  Thanks!

PR #1604, opened 2023-09-09
Added fix for pathological case…

pellaea:
... in which a user is logged in while sending an account verification email.

### DIFF
--- a/app/models/queued_email/verify_account.rb
+++ b/app/models/queued_email/verify_account.rb
@@ -13,5 +13,12 @@ class QueuedEmail
     def deliver_email
       VerifyAccountMailer.build(to_user).deliver_now
     end
+
+    # AbstractModel#set_user_and_autolog will fill in user with User.current
+    # if we aren't careful.  This may be the simplest (and cheesiest) way to
+    # ensure user stays nil.  This should override the user column.
+    def user
+      nil
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


JoeCohen:
I'm OK with the "cheesy" fix to stop the errors.
Another option?  This is a fairly recent bug (3 weeks?). Can we figure out where it was introduced?

pellaea:
Yeah, I think I know.  Verification emails used to not be queued.  So this
would never have been an issue.  It would have started when Nimmo changed
it into a QueuedEmail.  So, reversion is not an option (that I'm willing to
consider).

JoeCohen:
Thanks! I agree.

pellaea:
And I think I might know sorta why it's happening to start with.  I wonder
if the user is getting impatient and is asking for multiple verification
emails, and accidentally submitting one of those after they've successfully
verified?  I don't understand in detail because I still don't understand
the new routes.  But since the verification email is created in only one
place -- AccountController#create -- I assume reverification must also
somehow be hitting that route, too.  Nowhere do we explicitly set @user =
User.current = nil in that action, so if a user were to visit that route
while logged in, it would result in this error.  There should be no visible
links to that route once the user is logged in, but if they have an old
page open from before logging in, it could potentially happen.

Hmmm.  Does that mean we should actually kick the user out of
AccountController#create if they are already logged in??  Sort of the
opposite of login_required -- login_forbidden? ;)  Maybe send them to the
home page or the welcome page instead?  *That* would be a cleaner way to
solve the problem, maybe.

PR #1605, opened 2023-09-10
Switch `.count` to `.size` where appropriate

nimmolo:
Should resolve a number of mysterious n+1's. 

Discussion in #1598, starting [here](https://github.com/MushroomObserver/mushroom-observer/pull/1598#issuecomment-1712919253)

Please check the changes thoroughly, especially to classes and the model.


nimmolo:
There's a `count` in `shared/_form_list_feedback.rb` that if changed, exposes the need for whichever calling action to eager-load observations. 

But many forms may call this partial, and it's unclear where to add all the `includes`.

app/classes/suggestion.rb
```diff
@@ -68,7 +68,7 @@ def name_with_most_observations(names)
     best_name = nil
     best_count = -1
     names.each do |name|
-      count = name.observations.count
+      count = name.observations.length
```
JoeCohen, line 71: I don't know if `length` is right; I don't know whether this is right. (I don't how where it gets used in user workflow.)
Maybe use `size` here and in other places? Is that bad form?

app/controllers/account/preferences_controller.rb
```diff
@@ -100,7 +100,7 @@ def update_pref(pref, val)
     def update_content_filter(pref, val)
       filter = ContentFilter.find(pref)
       @user.content_filter[pref] =
-        if filter.type == :boolean && filter.prefs_vals.count == 1
+        if filter.type == :boolean && filter.prefs_vals.length == 1
```
JoeCohen, line 103: Though it's not the point of this PR, can you use
`one?`
instead of 
`length == 1`

app/helpers/locations_helper.rb
```diff
@@ -11,7 +11,7 @@ def country_link(country, count = nil)
   # title of a link to Observations at a location, with observation count
   # Observations at this Location(nn)
   def show_obs_link_title_with_count(location)
-    "#{:show_location_observations.t} (#{location.observations.count})"
+    "#{:show_location_observations.t} (#{location.observations.length})"
```
JoeCohen, line 14: Should this be `count` (or `size`)?
In a simple manual test with a breakpoint @ line 14, I get
`  Observation Load (559.3ms)  SELECT `observations`.* FROM `observations` WHERE `observations`.`location_id` = 2437
363
`

`(ruby) location.observations.count
   (192.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`location_id` = 2437
`

`(ruby) location.observations.size
   (200.0ms)  SELECT COUNT(*) FROM `observations` WHERE `observations`.`location_id` = 2437
363
`

So to me it looks like `count` and `size` are 2x faster than `length`.

app/classes/query/modules/high_level_queries.rb
```diff
@@ -112,7 +112,7 @@ def paginate_ids(paginator)
         ids = ids.select { |id| @letters[id] == letter }
       end
     end
-    paginator.num_total = ids.count
+    paginator.num_total = ids.length
     ids[paginator.from..paginator.to] || []
   end
```
nimmolo, line 117: @pellaea is this OK in `Query::HighLevelQueries`? At the point of figuring out the `paginator` and the `@num_results`, the query is done, so it seems `length` is more appropriate. 

It may make no difference currently, but it will probably matter if Query starts using scopes.

app/classes/suggestion.rb
nimmolo: @JoeCohen I don't know either, but I can answer one part of it. According to internet, `size` is an alias of `length`. ([on Medium](https://kevingrow11.medium.com/difference-between-length-size-and-count-array-methods-in-ruby-34f69df8b80b#:~:text=In%20conclusion%2C%20%23length%2C%20%23,the%20arguments%20passed%20into%20it.))([on Ruby in Rails](https://www.rubyinrails.com/2014/01/15/ruby-count-vs-length-vs-size/)).

app/controllers/account/preferences_controller.rb
nimmolo: Yes! I don't know that method. Will alter

app/classes/suggestion.rb
nimmolo: Update: Sorry, `size` is NOT an alias of `length`. Thank you [Stack Overflow](https://stackoverflow.com/a/14794742/3357635)!

nimmolo: However, that's an old answer. In the console, I notice that `length` on an association that has already been fetched only counts the number of cached records, so that may have changed.

nimmolo: After reading through that whole thread, i think I should go with `size`. Thanks for figuring that out.

JoeCohen: You made me look. ​
If we need the records, `length` is better.
Else `size` is better. 

>
length().    [Link](https://api.rubyonrails.org/v6.1/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-length)
Returns the size of the collection calling size on the target. If the collection has been already loaded, length and size are equivalent. If not and you are going to need the records anyway this method will take one less query. Otherwise size is more efficient.

So it looks like `size` is faster here. But I don't know if it's faster enough to be worth worrying about.

nimmolo: From [Roy Lindauer](https://www.roylindauer.com/blog/the-difference-between-length-size-and-count-in-ruby-on-rails.html)
count on a collection will always execute a COUNT(*) SQL statement to get an accurate count value. size returns the size of the collection, but will execute a COUNT(*) SQL statement if it's not already loaded. length returns the size of the collection from memory. If the collection has already been loaded and is in memory length and size are equivalent.

Relying on what's in memory may give you inaccurate results as the database could be modified while you are working on that collection. But making a bunch of expensive database calls to get an accurate count is probably something you want to avoid. There may be times when a count is preferred, but when in doubt it's probably better to use length.

also see diagrams here: https://www.akshaykhot.com/rails-length-size-count-difference/

I'm going to post all this stuff in a GitHub Discussion for reference.

app/helpers/locations_helper.rb
nimmolo: Yes. Thank you.

nimmolo:
```suggestion
    "#{:show_location_observations.t} (#{location.observations.size})"
```
Putting a suggestion here so i can commit it right here.

app/classes/pivotal.rb
```diff
@@ -12,7 +12,7 @@ def get_stories(_verbose = false)
       stories = []
 
       request_stories("", stories)
-      request_stories("&offset=501", stories) if stories.count == 500
+      request_stories("&offset=501", stories) if stories.length == 500
```
nimmolo, line 15:
```suggestion
      request_stories("&offset=501", stories) if stories.size == 500
```

app/classes/query/modules/high_level_queries.rb
nimmolo: Ok i now think `size` may be more appropriate.

app/classes/suggestion.rb
nimmolo, line 71:
```suggestion
      count = name.observations.size
```

app/controllers/account/preferences_controller.rb
nimmolo, line 103:
```suggestion
        if filter.type == :boolean && filter.prefs_vals.one?
```

app/controllers/collection_numbers_controller.rb
```diff
@@ -350,7 +350,7 @@ def figure_out_where_to_go_back_to
       @back_object = Observation.safe_find(@back)
       return if @back_object
 
-      @back_object = if @collection_number.observations.count == 1
+      @back_object = if @collection_number.observations.length == 1
```
nimmolo, line 353:
```suggestion
      @back_object = if @collection_number.observations.one?
```

JoeCohen: Another place for `one?`, which is defined as
```ruby
# File activerecord/lib/active_record/relation.rb, line 282
def one?
  return super if block_given?
  limit_value ? records.one? : size == 1
end
```
https://github.com/rails/rails/blob/56bcc0abd3c9a6b09469e9428f6eea0dd77c2294/activerecord/lib/active_record/relation.rb#L282

Note: I'm fine with not using `one?` here and elsewhere. I'm just putting it out as an option.

app/controllers/herbarium_records_controller.rb
```diff
@@ -413,7 +413,7 @@ def figure_out_where_to_go_back_to
       @back_object = Observation.safe_find(@back)
       return if @back_object
 
-      @back_object = if @herbarium_record.observations.count == 1
+      @back_object = if @herbarium_record.observations.length == 1
```
nimmolo, line 416:
```suggestion
      @back_object = if @herbarium_record.observations.one?
```

app/controllers/collection_numbers_controller.rb
JoeCohen: Oops. Probably no on `size`, stick with `length`, as `observations` is used on the next line.

app/helpers/object_link_helper.rb
```diff
@@ -163,7 +163,7 @@ def description_link(desc)
   end
 
   def observation_herbarium_record_link(obs)
-    count = obs.herbarium_records.count
+    count = obs.herbarium_records.length
```
nimmolo, line 166:
```suggestion
    count = obs.herbarium_records.size
```

app/controllers/names/synonyms_controller.rb
```diff
@@ -155,7 +155,7 @@ def split_off_desynonymized_names(main_name, checks)
       pick_one = other_group.shift
       pick_one.clear_synonym
       other_group.each { |n| pick_one.transfer_synonym(n) }
-      main_name.clear_synonym if main_name.reload.synonyms.count <= 1
+      main_name.clear_synonym if main_name.reload.synonyms.length <= 1
```
nimmolo, line 158:
```suggestion
      main_name.clear_synonym if main_name.reload.synonyms.size <= 1
```

app/helpers/object_link_helper.rb
```diff
@@ -133,7 +133,7 @@ def user_link(user, name = nil, args = {})
   def user_list(title, users = [])
     return safe_empty unless users&.any?
 
-    title = users.count > 1 ? title.to_s.pluralize.to_sym.t : title.t
+    title = users.length > 1 ? title.to_s.pluralize.to_sym.t : title.t
```
nimmolo, line 136:
```suggestion
    title = users.size > 1 ? title.to_s.pluralize.to_sym.t : title.t
```

app/presenters/matrix_box_presenter.rb
```diff
@@ -120,7 +120,7 @@ def user_to_presenter(user)
     # user.contribution and observations.count are just numbers.
     self.detail = "#{:list_users_joined.t}: #{user.created_at.web_date}<br/>
                    #{:list_users_contribution.t}: #{user.contribution}<br/>
-                   #{:Observations.t}: #{user.observations.count}".html_safe
+                   #{:Observations.t}: #{user.observations.length}".html_safe
```
nimmolo, line 123:
```suggestion
                   #{:Observations.t}: #{user.observations.size}".html_safe
```

app/views/account/preferences/_filters.html.erb
```diff
@@ -6,10 +6,10 @@
 
 <% ContentFilter.all.each do |filter| %>
 
-  <% if filter.type == :boolean && filter.prefs_vals.count == 1 %>
+  <% if filter.type == :boolean && filter.prefs_vals.length == 1 %>
```
nimmolo, line 9:
```suggestion
  <% if filter.type == :boolean && filter.prefs_vals.one? %>
```

app/views/account/preferences/_filters.html.erb
```diff
@@ -6,10 +6,10 @@
 
 <% ContentFilter.all.each do |filter| %>
 
-  <% if filter.type == :boolean && filter.prefs_vals.count == 1 %>
+  <% if filter.type == :boolean && filter.prefs_vals.one? %>
     <%= render(partial: "account/preferences/filters/checkbox",
                locals: { filter: filter, f: f }) %>
-  <% elsif filter.type == :boolean && filter.prefs_vals.count > 1 %>
+  <% elsif filter.type == :boolean && filter.prefs_vals.length > 1 %>
```
nimmolo, line 12:
```suggestion
  <% elsif filter.type == :boolean && filter.prefs_vals.size > 1 %>
```

app/views/collection_numbers/_form.html.erb
```diff
@@ -15,7 +15,7 @@ end
 <%= form_with(model: @collection_number, url: url_params,
               local: local, id: "collection_number_form") do |f| %>
 
-  <% if @collection_number.observations.count > 1 %>
+  <% if @collection_number.observations.length > 1 %>
```
nimmolo, line 18:
```suggestion
  <% if @collection_number.observations.size > 1 %>
```

app/views/herbarium_records/_form.html.erb
```diff
@@ -15,7 +15,7 @@ end
 <%= form_with(model: @herbarium_record, url: url_params,
               local: local, id: "herbarium_record_form") do |f| %>
 
-  <% if @herbarium_record.observations.count > 1 %>
+  <% if @herbarium_record.observations.length > 1 %>
```
nimmolo, line 18:
```suggestion
  <% if @herbarium_record.observations.size > 1 %>
```

app/views/names/synonyms/_fields_existing.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# existing synonyms %>
 
-<% if current_synonyms.count > 1 %>
+<% if current_synonyms.length > 1 %>
```
nimmolo, line 3:
```suggestion
<% if current_synonyms.size > 1 %>
```

app/views/observations/show/_collection_numbers.html.erb
```diff
@@ -51,7 +51,7 @@ cn_query = Query.lookup(:CollectionNumber, :all, observations: obs.id)
     </ul>
 
   <% elsif numbers.any? && !can_edit %>
-    <%= numbers.count > 1 ? :Collection_numbers.t : :Collection_number.t %>:
+    <%= numbers.length > 1 ? :Collection_numbers.t : :Collection_number.t %>:
```
nimmolo, line 54:
```suggestion
    <%= numbers.size > 1 ? :Collection_numbers.t : :Collection_number.t %>:
```

app/views/observations/show/_herbarium_records.html.erb
```diff
@@ -53,7 +53,7 @@ query = Query.lookup(:HerbariumRecord, :all, observations: obs.id)
 
   <% elsif records.any? && !can_add %>
     <div>
-      <%= records.count > 1 ? :Herbarium_records.t : :Herbarium_record.t %>:
+      <%= records.length > 1 ? :Herbarium_records.t : :Herbarium_record.t %>:
```
nimmolo, line 56:
```suggestion
      <%= records.size > 1 ? :Herbarium_records.t : :Herbarium_record.t %>:
```

app/views/pivotal/index.html.erb
```diff
@@ -48,8 +48,8 @@
 %>
 
 <div class="container-text">
-  <% if @stories.count > 900 %>
-    <b>Displaying <%= @stories.count %> of 1,000 available Feature Requests. Please <%= link_to 'let the admins know', '/emails/ask_webmaster_question', title: "Contact Us" %> that we're close to the limit.</b>
+  <% if @stories.length > 900 %>
```
nimmolo, line 51:
```suggestion
  <% if @stories.size > 900 %>
```

app/views/pivotal/index.html.erb
```diff
@@ -48,8 +48,8 @@
...
+    <b>Displaying <%= @stories.length %> of 1,000 available Feature Requests. Please <%= link_to 'let the admins know', '/emails/ask_webmaster_question', title: "Contact Us" %> that we're close to the limit.</b>
```
nimmolo, line 52:
```suggestion
    <b>Displaying <%= @stories.size %> of 1,000 available Feature Requests. Please <%= link_to 'let the admins know', '/emails/ask_webmaster_question', title: "Contact Us" %> that we're close to the limit.</b>
```

JoeCohen:
Thank you. A nice, succinct explanation of the differences, plus a link to
a very useful flow chart.

app/controllers/collection_numbers_controller.rb
nimmolo: OK. I went with `one?` - If you get a chance please have another look over everything — i've changed everything in the PR — and check if it seems right.

JoeCohen:
I'm out of time and probably won't be able to return to this PR.
One note of caution: `one?`, `many?`, and `none?` all call `size`. 
So `length == 1` etc. may be preferable to `one?` etc. in cases where `length` is preferable to `size`.

app/classes/query/modules/high_level_queries.rb
pellaea: Oh dear god!  I can't possibly read that entire email thread right now!

I do want to make sure one thing is pointed out in this one case, though -- `ids` here is not a query, it's just a plain old `Array`.  So there's no question of it executing SQL queries or anything if "done wrong".

(Is there in fact any difference between size, count and length on a plain old Array??... oh hold on, strike that question... it's almost certainly answered somewhere in the email thread I just deleted! :)

nimmolo: LOL

PR #1607, opened 2023-09-11
Fix template for obs_title with preferred synonym

nimmolo:
@pellaea This should fix the FUBAR [title](https://github.com/MushroomObserver/mushroom-observer/pull/1598#issuecomment-1713104814) in that [Amanita lavendula group obs](https://mushroomobserver.org/505457?q=Q).

Note: It seems this PR branch prints the owner naming line in cases where current `<main>` does not. (See [504733](https://mushroomobserver.org/504733).) I don't understand why that would be, except maybe... the current helpers bail when the title gets messed up? and then the owner naming line doesn't print? This PR doesn't change anything in the owner naming line, so I really don't understand.

Furthermore, I can't remember which is the expected behavior, hope you can elucidate.

### DIFF
--- a/app/helpers/observations_helper.rb
+++ b/app/helpers/observations_helper.rb
@@ -32,29 +32,47 @@ module ObservationsHelper
   def obs_title_consensus_name_link(name:, owner_naming: nil)
     if name.deprecated &&
        (prefer_name = name.best_preferred_synonym).present?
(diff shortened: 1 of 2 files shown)


nimmolo:
Going to merge, because this is for sure better than what's on `<main>` now.

Will regroup tomorrow and try to get it sorted out better.

PR #1610, opened 2023-09-14
Adjust `MOAutocompleter` to not use jQuery

nimmolo:
## Manual review only ##
To test this PR manually, just open a new obs form and check the behavior of the autocompleted fields.
Or try it on other forms, or check the "Help Identify" page, where the "filter by region" autocomplete works correctly for once. Some autocompleted fields, like the obs image upload "year" autocompletes, are in forms that are not present at page load but created by JS. They also should work!

The goal of this PR was simply rewriting the autocompleter without jQuery, as preparation for Rails 7. All functionality should be exactly as current... except it fixes a number of bugs.

#### Background ####
@pellaea's `MOAutocompleter` is very different from other open source autocompleters. It has custom features such as matching words out of order, as in "scientific" location_format, and is closely configured to work with the responses MO's `AutocompleteController` produces. 

Most importantly, it is MUCH quicker than off-the-shelf autocompleters in dealing with AJAX queries for names and locations, because it calls the db once for all entries starting with the typed letter of the alphabet, and then consults this internal `primer` to refine results as the user types more, rather than making separate server requests. 

Having finally understood this, of course I want to keep it. 

#### New features ####
This PR makes it even easier to implement an autocompleted field in HTML. You just add the data-attribute 

```js
    data: { autocompleter: :location } // or whatever type
```

to an input field. An observer/initializer in `assets/javascripts/main.js` watches the DOM if an element needing to be initialized as an autocompleter has been added. and instantiates the completer. (This initializer is temporary, since Stimulus can handle observing DOM changes in Rails 7.) 

One result of the refactor here is that methods like `turn_into_#{x}_autocompleter` can now be removed from the templates.


app/assets/javascripts/autocomplete.js
```diff
@@ -130,282 +141,296 @@ var MOAutocompleter = function(opts) {
...
+          this.page_down();
+          this.schedule_key(this.page_down);
+          break;
+        default:
+          this.current_row = -1;
+          break;
       }
     }
-    if (this.on_keydown)
-      this.on_keydown(event);
```
nimmolo, line 224: @pellaea can you remember what you meant here? `on_keydown` is not a JS method, it's `onkeydown`. 

There's a series of overrides here but it's not clear to me what these statements do. `this.onkeydown` always fires an event.

```js
if (this.on_keydown)
      this.on_keydown(event);
```

pellaea: My best guess is that this originated in the template I started with.  We don't seem to use these callbacks anywhere, so feel free to tear them out.  (It's not just on_keydown, there's also an on_keypress and on_keyup.)  I think you can simply remove those 6 lines and that's it.

nimmolo: Thank you. 
Do you have any time today to give this a spin and see if this rewrite works as intended?

 It's not quite ready for review but i will mark it as such if that's a possibility.

nimmolo: Ok. I won't mark it for review unless i'm pretty sure it's bug free.

nimmolo: Ready for review, but no rush, safe travels. 
jQuery extension methods are duplicated here in JS until they can be moved into a utility class.

pellaea: At the gate, and need to stay awake... holy moly, this is fast!  Ah, is this what the rest of the world is like???

app/assets/javascripts/autocomplete.js
```diff
@@ -130,282 +140,284 @@ var MOAutocompleter = function(opts) {
...
 
   // Need to prevent these keys from being processed by form.
-  our_keypress: function(event) {
-    var key = event.which == 0 ? event.keyCode : event.which;
-    // jQuery("#log").append("keypress(key=" + key + ", menu_up=" + this.menu_up + ", hot=" + this.is_hot_key(key) + ")<br/>");
-    if (this.on_keypress)
-      this.on_keypress(event);
+  our_keypress: function (event) {
+    const key = event.which == 0 ? event.keyCode : event.which;
+    // this.debug("keypress(key=" + key + ", menu_up=" + this.menu_up + ", hot=" + this.is_hot_key(key) + ")");
```
pellaea, line 262: Thanks for keeping these debugging messages!  They were very helpful once upon a time. :)

pellaea:
Hey Nimmo, I ran out of time.  It looks like the vast majority of changes are one-to-one translations, so this is probably safe.  For some reason github is not letting me check out the branch so I can't test it live.  Maybe a 2FA thing?  Gotta run!

nimmolo:
Thanks for checking the code out. One thing currently broken I found is hot-swapping the autocompleters on `observations/identify`when you flop between clade and region... will try to fix that shortly.

I have a branch off this that makes the function a JS class, also with very few changes, only syntax tweaks.

nimmolo:
Ready for (manual) review.  If anyone is available to hand-test some autocompleters, please do. 

Will merge tomorrow morning, otherwise.

mo-nathan:
Immediately ran into an issue in the observation create workflow.  I clicked "Create Observation", typed "North Falmouth" in the Location box and selected "North Falmouth, Falmouth, Massachusetts, USA".  I then typed "Boletus e" into the Scientific name and selected "Boletus edulis" and hit return.  It then came back with a warning that says "Unknown country ‘North Falmouth’".  Note that my Location Format is set to "Scientific (USA, New York)".  However, the options I was given are in "Postal" format.  This is an important feature to maintain for the professional scientific community.  Roy Halling (formerly of the NYBG, an expert in all things Boletes, and a strong supporter of MO) specifically requested this feature.

nimmolo:
@mo-nathan 
I had "wisely" intuited that a change in the handling of the `format` param could be buggy and pulled it out to its own PR, #1614. But that's where the problem was introduced. Thank you for catching it before any users did.

nimmolo:
Scientific is fixed... workflow described above works as expected, maintaining "scientific" location_format throughout the process. Let me know if anyone wants to manually test again.

mo-nathan:
Hmm, this one doesn't seem to be working (but it's not critical IMO):
`app/views/admin/session/edit.html.erb:                                    autocompleter: :user }) %>`

To get to it, you have to be in admin most and then click "Switch User".  I honestly forgot that this was even a thing.

nimmolo:
Thanks for the thoroughness @mo-nathan. 

It does work for me, though. (well, there's still the VERY annoying Safari pushy-autocompleter).  

Screen Shot 2023-09-18 at 3 06 10 PM

Are you sure it's not just slow... or, could it be that you started typing before the page was loaded. That was a gotcha for me repeatedly in testing.

mo-nathan:
You are correct.  It was just slow.  Working now.

mo-nathan:
Well, not really working.  Sometimes waiting longer matters, but sometimes it never starts working again.  Once it's broken it seems to stay broken.

mo-nathan:
Not clear exactly how long you have to wait, so maybe it's just a matter of waiting long enough, but I'm really not sure.  I did get it into a weird situation where if you type "hollinger" it would show you two accounts (jason and jason_test), but when you select an autocomplete value, it just leaves "hollinger" and then the switch fails.  I've not been able to consistently reproduce it.

nimmolo:
>re: when it's broken it seems to stay broken

That's what i get sometimes too (but not today, just now). I had to reload the page to get it to autocomplete again.

I believe the reason it's an issue in dev but not on production may be that the js is compiled and cached, but on dev it's always reloaded, and so slower to bind?

mo-nathan:
Hmm, and then I got into another weird state where it says "jason <Jason Hollinger>", but when I hit <return> it fails with the error message 'Couldn't find "jason ". Play again?'.

nimmolo:
Yep, i'm getting that too. researching

JoeCohen:
Also note a preexisting bug (which you should probably ignore, just like
Jason & I do):
The Login Name field must contain *only* the login name, rather than both
the login and the user name. (On the webserver the autocompleter supplies
both the login and user name. I haven't had a chance to check the PR.)

nimmolo:
Got it, yes, thanks Joe. 
If you accept an autocomplete, it's been trying to find the user by a string that never corresponds to either a login or an email.

I believe the `update` action just needs to strip out the real name from a string input, if there's no `@`, something like this:

```ruby
@id = params[:id].to_s
if(@id.is_a?(String) && !@id.contains("@")
  @id = @id.split(" <")[0]
```

nimmolo:
Switch user bug should be fixed!

JoeCohen:
> Switch user bug should be fixed!

Thank you!  Now it works for everyone.
(@pellaea and I were too lazy to fix it because we rarely used it and we knew the easy work-around.)

nimmolo:
Deployed

PR #1611, opened 2023-09-16
Project location support

mo-nathan:


app/views/projects/_form.html.erb
```diff
@@ -12,9 +12,12 @@
...
 
-  <%= text_area_with_label(form: f, field: :summary, rows: 20,
+  <%= text_area_with_label(form: f, field: :summary, rows: 5,
                            label: :SUMMARY.t + ":") %>
 
   <%= render(partial: "shared/textilize_help") %>
 
+  <%= text_field_with_label(form: f, field: :where,
+                            label: :form_projects_where.t + ":") %>
+
```
nimmolo, line 22: Just a heads up that there will be a new, easy way to declare an autocomplete field coming up with the Autocompleter refactor PR I hope to merge tonight. 

 I don't know if you would want that for this field, but if you do, the syntax is simple.  You just add `data: { autocompleter: :location }` to the HTML options hash for the `text_field`.

mo-nathan: Nice!  Adding an autocomplete is on my to do list.  I'll check it out.

mo-nathan: Adopted!

JoeCohen:
1. I agree that you should simply accept an Obs whose lat/on meets the Project's constraints regardless whether the Obs Location intersects with the Project Location.
The "weird" case you mention implies that the Observation's lat/lon is outside its Location. That's not a Project-specific problem. So -- if/when we're going to deal with it -- it should be handled elsewhere. (Maybe in both Observations and Locations.) (I suspect that this is not a rare issue.)
2. I agree about triggering a warning if an Obs is edited to move lat/lon outside the Project's constraints.
3. If the agree that there should be a separate interface if the Project constraints change such that Observation(s) no longer meet those constraints. 
But before such a change is accepted, the user who is editing the Project should get a generalized flash warning (something like "This change would result in one or more Observations being outside the Project's constraints.") and a chance to not make the change. 

> I think this ready for review. I have cribbed heavily from Joe's work on the date constraints. Hopefully, I've made it pretty obvious how to add the date constraints in so it's not a big deal.
> 
> I do have a couple of questions that I'd like thoughts on:
> 
> 1. If a new observation has a Location that does not intersect with the Project Location, but has a lat/long that is inside the Project Location, I accept it.  Obviously the person is doing something weird with observation location information, but so it's a bit of an edge case, but do folks think it's worth checking for?
> 2. Right now if you edit the observation and move it's lat/long outside of the project location, it's ok with it.  It seems like this should trigger the same warnings as creation.
> 3. What should happen if the project owner changes the location constraint?  As mentioned on Joe's PR, I think this should probably be handled with a separate interface that highlights observations that don't meet the current constraints which I think is separate work.

mo-nathan:
Thanks!  I think I have things working correctly for updates now as well as creates.

There is some weird JS behavior though.  Specifically, if you try to change just the lat/long (or only interact with text boxes in general), the "Save Edit" button never activates.  The only way I could get it to work was by click a checkbox on and off again.  Is there something we can do to fix this?  I assume it's there to avoid double clicking, but it seems to take effect too early/aggressively.  I also noticed the same issue when the create page gets re-rendered when a constraint is violated.

mo-nathan:
@JoeCohen regarding changing the constraints on a project, I think we can handle the explicit lat/long and date constraints in a single query.  I don't think we can do the same for observations that only have a Location and no lat/long.  That seems ok to me.  Otherwise we end up in N+1 land on the number of observations associated with a project which I think we expect to get at least into the 1000s.  Thoughts?

JoeCohen:
I don't know. Trying to think of a query is making my head hurt.
Does it make any difference if:

   1. We're giving a general notice that the constraint change would put
   some Observations out of bounds, vs
   2. We're in the separate interface for listing the out-of-bounds or
   out-of-date-range Observations?

mo-nathan:
If we're in the UI for the violating observations, then we can deal with the N+1 issue with pagination.  We could also fudge it by not being strict about observations that don't have lat/long.  We might be able to get away with requiring lat/long for observations in projects that have a location.  I'd have to ask Steve Russell and Andy Wilson about that.

Regarding the lat/long constraint is the a better way to express the following:
```
  def count_violations
    return true unless location

    count = observations.where("lat IS NOT NULL").count
    west = location.west
    east = location.east
    count -= observations.where("(lat <= ?) && (lat >= ?)",
                                location.north, location.south).
               where("((`long` >= ?) && (`long` <= ?) && (? <= ?)) ||                                                                   
             (((`long` >= ?) || (`long` <= ?)) && (? > ?))",
                     west, east, west, east, west, east, west, east).count
  end
```

JoeCohen:
How about using the scopes that @nimmolo created; something like
```ruby
  def count_violations
    return true unless location

    observations.in_box(n: location.north, s: location.south, 
                        e: location.east, w: location.west).count
  end
```

> If we're in the UI for the violating observations, then we can deal with the N+1 issue with pagination. We could also fudge it by not being strict about observations that don't have lat/long. We might be able to get away with requiring lat/long for observations in projects that have a location. I'd have to ask Steve Russell and Andy Wilson about that.
> 
> Regarding the lat/long constraint is the a better way to express the following:
> 
> ```
>   def count_violations
>     return true unless location
> 
>     count = observations.where("lat IS NOT NULL").count
>     west = location.west
>     east = location.east
>     count -= observations.where("(lat <= ?) && (lat >= ?)",
>                                 location.north, location.south).
>                where("((`long` >= ?) && (`long` <= ?) && (? <= ?)) ||                                                                   
>              (((`long` >= ?) || (`long` <= ?)) && (? > ?))",
>                      west, east, west, east, west, east, west, east).count
>   end
> ```

mo-nathan:
Nice!  Not exactly equivalent, but I can work with it.

mo-nathan:
Ok, this might be ready to go.  I made it so the Project#show page indicates how many violations there currently are (rather than dealing with figuring out the before and after numbers and throwing up a warning).  I figured the violation warning would eventually have a link to some sort of "project review" page which would list the violating observations.  @JoeCohen To add the date stuff I think you'll just need to hook into Project#violates_constraints? and add another where clause into count_violations.  (And of course add the fields with a migrationm and a UI,..., but I think you already have all that)

app/models/project.rb
```diff
@@ -108,6 +110,28 @@ def user_can_add_observation?(obs, user)
...
+
+    !location.found_here?(obs)
+  end
+
+  def count_violations
+    return 0 unless location
+
+    count = observations.where.not(lat: nil).count
+    count - observations.in_box(n: location.north, s: location.south,
+                                e: location.east, w: location.west).count
```
JoeCohen, line 128: If you're interested, I think there's a way to do this in Rails 6 in a single query, using Arel query that's kind of opaque.
(In Rails 7, it would be much clearer.)
```ruby
    observations.where(
      observations.in_box(n: location.north, s: location.south, 
                          e: location.east, w: location.west).
        arel.constraints.reduce(:and).not
    ).count
```
See https://makandracards.com/makandra/486959-how-to-negate-scope-conditions-in-rails

But unless `count_violations` gets called a lot, I like your code.

PR #1616, opened 2023-09-17
Switch create error to redirect

mo-nathan:
While the previous PR stopped an error from happening it was still confusing since it switched over to an update when it should be creating something new.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -91,7 +91,7 @@ class ProjectsController < ApplicationController
     else
       return create_project(title, admin_name)
     end
-    render(:new, location: new_project_path(q: get_query_param))
+    redirect_to(new_project_path(q: get_query_param))
   end
 
   def update
--- a/test/controllers/projects_controller_test.rb
+++ b/test/controllers/projects_controller_test.rb
@@ -4,7 +4,7 @@ require("test_helper")
 
 class ProjectsControllerTest < FunctionalTestCase
   ##### Helpers (which also assert) ############################################
-  def add_project_helper(title, summary, action = :create, id = nil)
+  def add_project_helper(title, summary)
     params = {
       project: {
         title: title,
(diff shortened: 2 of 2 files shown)


nimmolo:
@mo-nathan The render is recommended by Rails on all CRUD actions because it reduces server load. I switched these wherever possible in the codebase, including innumerable test expectations. 

So I'd like to request that you consider reverting this and just try to remember, this is contemporary expected behavior for `create` and `update`.

JoeCohen:
@nimmo: Thanks for all that work and for the explanation.

mo-nathan:
I made this change because the previous version of things did not work correctly.  The problem was that after adding the @project stuff the create process turned into an update if you used the name of an already existing project.  The behavior was very surprising and in my view wrong.  For example, if you use the name of someone else's project, when the page gets re-render it has data from the other project (summary and other widgets) and when you hit "create" after changing the title, it rejects it since you don't have permission.  If you instead happen to reuse the name of your own project, then change the title, it changes the title of the already existing project and does not create a new one.  As far as I can tell, the redirect behaves much better since it restarts the process.  I think the impact on server load is negligible since it only happens when there is a problem with the first way the form gets filled out.  If you want to propose another solution, I'm happy to test it out, but this is not the thing I actually want to be working on so I don't want to invest more time in it.

nimmolo:
Ah - that makes more sense.  Now i understand what you meant by "the behavior was confusing". I  thought you just weren't used to `render` - it was almost never used for create/update in old MO. 

Do you think this type of validation/switch gotcha is likely to be lurking out ther on other forms?

This behavior seems like it merits a note in the controller so someone like me doesn't revert it out of habit. (Maybe after this PR Joe or I can try to work out why the ivar is switching and fix it, because when the form is AJAXified by Turbo, the same issue may recur. Turbo always calls `render` - part of the motivation for switching hundreds of `redirect`s to `render`s.)

mo-nathan:
Yeah, I agree it's at least worth a comment.  There may well be a way to make this happen using `render`.  Now that I reflect on it, I wonder if the issue is that we shouldn't be using `@project` where we are right now in the `create` method, but instead should be doing `@project = Project.new` only when an error happens.  I just tried something like that, but it's now failing in a new way.  Probably something trivial, but I need to head to bed at the moment.

PR #1619, opened 2023-09-20
Project date range

JoeCohen:
Adds Project start and end dates.
- Project admin sets start and end dates when creating or editing a Project.
- Those dates display on the Project page.
- If a Project is accepting new observations from members, then when a Project member creates an Observation:
  - If the ***current date*** is outside the Project date range, the Project is not automatically selected for the Observation;
  - If the User manually selects the Project and the ***Observation  date*** (`when`) is outside the Project date range, then
    - The User receives a warning (after clicking the Obs Create button) and a chance to correct by:
      - Deselecting the offending Project,
      - Changing the Obs date to within the Project Range
      - Overriding the warning by checking "Ignore Project Dates".
See  [Open Project Stories for MO](https://docs.google.com/document/d/1g8QMR6NzUvM4pjuAuBiI9QGZU_XNail-ocMu120UBxw/edit), Story 1, including comments

Delivers [Pivotal #186018784](https://www.pivotaltracker.com/story/show/186018784)

### Suggested manual test
1. Create a Project whose end (and start) dates are in the past and that’s accepting new observations from members.
2. Click Create Observation
3. Check the box for the project you just created.
4. Click the Create button
Expected Result: Warning including instructions on how to proceed.
5. Click the Create button
Expected Result: Ditto. (You didn’t fix the problem.).
6. Uncheck the project.
7. Click the Create button again.
Expected Result: Observation created.

8. Click Create Observation
9. Check the box for the project.
10. Click the Create button
Expected Result: Warning.
11. Check Ignore Project Dates
12. Click the Create button again.
Expected Result: Observation created && Observation added to project.

Same drill, this time fixing the problem by changing When to be in-range for the Project.
Expected Result: Observation created && Observation added to project.


JoeCohen:
### General Comment
This branch should have the desired functionality.

But it could use improvement, especially in the views, particularly the flash warning and chance to correct. about projects (see the Suggested Manual Test in the [opening PR Comment](https://github.com/MushroomObserver/mushroom-observer/pull/1619#issue-1904010146). 

Other possible improvements include:
* Project admins receive (or can opt to receive) a warning (email) if a user submits an out-of-range Obs.
* Let users create/edit Projects to have no start and/or end. (All existing projects in the db have no start and no end. There's currently no way to create a Project like that’s open on at least one end.)
* Validate start/end dates to prevent start from earlier than end. (Low priority.)
* Convert Project#past, #current, and #future from methods into scoped, ARELized AR queries.
* Move parts of test/integration/capybara/projects_integration_test.rb to controller tests. (Limit integration test to user changing the page, e.g., checking, unchecking, date changes.) I think that will make the tests marginally faster.

JoeCohen:
@mo-nathan:
### Overriding the Project Dates
@mo-nathan’s original suggestion for what happens if a user checks a box for an out-of-range project when creating an Observation included:
>it saves the Observation "if the users] resubmit with the same 'when' value and projects selected."

I did it slightly differently: There’s an “Ignore Project Dates” checkbox. If the user checks that and resubmits, then it saves. 
* I couldn’t figure out how to code @mo-nathan's exact suggestion with only Ruby. (I’d guess there’s a way to do it in js, but that’s beyond me.)
* Using a checkbox makes it slightly more difficult to accidentally add an Observation to an out-of-range Project.

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -86,7 +86,8 @@ def init_project_vars
     @project_checks = {}
     @projects.each do |proj|
       @project_checks[proj.id] = (proj.open_membership &&
-                                  proj.accepting_observations)
+                                  proj.accepting_observations &&
```
JoeCohen, line 89: @mo-nathan 
#### Question: Should we remove `accepting_observations`?
“Project is accepting new observations from members” is largely redundant with date ranges.
It’s hard for me to see a non-edge use case for having both.
I suggest ditching `accepting_observations`.

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -194,6 +195,26 @@ def validate_place_name(params)
...
+
+  def checked_out_of_range_projects
+    checked_proj_check_boxes =
+      params[:project].select { |_, value| value == "1" }.keys
+    return [] if checked_proj_check_boxes.none?
+
+    checked_proj_ids =
+      checked_proj_check_boxes.map { |str| str.gsub("id_", "") }
+    # Get the AR records so that we can call Project methods on them
+    Project.where(id: checked_proj_ids).
```
JoeCohen, line 214: Question: Should I try to turn Project#dates_exclude? into an ARELized, scoped, AR query (instead of a method)?
Maybe I could avoid the filter and just do something like:
`Project.where(id: checked_proj_ids).where(dates_exclude(@observation.when)`

app/models/project.rb
```diff
@@ -315,6 +318,50 @@ def orphan_each_draft(drafts)
...
+  end
+
+  def dates_exclude?(date)
+    !dates_include?(date)
+  end
+
+  def dates_include?(date)
+    starts_no_later_than?(date) && ends_no_earlier_than?(date)
+  end
+
```
JoeCohen, line 338: Should I try to turn these into ARELized, scoped, AR queries.

app/views/observations/form/_projects.html.erb
```diff
@@ -1,15 +1,45 @@
...
+        )}:"
+      ) %>
+      <ul>
+        <% @suspect_checked_projects.each do |proj| %>
+        <%= tag.li("#{proj.title} (#{proj.start_date_str} -
+                    #{proj.end_date_str})") %>
+        <% end %>
+      </ul>
+      <%= tag.p(
+        :form_observations_projects_out_of_range_help.t(button: button_name)
```
JoeCohen, line 19: To avoid implying that we're creating another button, I'll probably change this to `(button_name: button_name)` (here and in `en.txt`)

app/views/projects/show.html.erb
```diff
@@ -18,6 +18,8 @@ add_tab_set(project_show_tabs(project: @project, user: @user))
   <div class="col-md-6">
     <%= panel_block do %>
       <p><b><%= :show_project_created_at.t %>:</b> <%= @project.created_at.web_date %></p>
+      <p><b><%= :show_project_start_date.t %>:</b> <%= @project.start_date %><br/>
+      <b><%= :show_project_end_date.t %>:</b> <%= @project.end_date %></p>
       <% if false %>
         <p><b><%= :show_project_published.t %>:</b> <%= @name_count %></p>
       <% end %>
```
JoeCohen, line 25: @mo-nathan: What's this? Can I delete it?

app/views/projects/_form.html.erb
```diff
@@ -2,6 +2,25 @@
...
+      <%= f.label(:start_date, "#{:form_projects_start_date.t}: ") %>
+      <%= f.date_select(:start_date,
+                        { class: "form-control mb-2" }) %>
+    </div>
+    <div>
+      <%= f.label(:end_date, "#{:form_projects_end_date.t}: ") %>
+      <%= f.date_select(:end_date,
+                        { class: "form-control mb-2" }) %>
+    </div>
+  </div>
```
JoeCohen, line 22: Quick & Dirty.
I need to figure out how to style this with @nimmolo's helpers and css classes.

config/locales/en.txt
```diff
@@ -3261,6 +3270,9 @@
   send_feature_email_success: Delivered feature mail.
   send_feature_email_denied: Only the admin can send feature mail.
 
+  # project general
+  project_nil_date_str: "[:NONE]"
```
JoeCohen, line 3274: @mo-nathan: Is there a better place in this file to put this?

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -194,6 +195,26 @@ def validate_place_name(params)
     success
   end
 
+  def validate_project_checkboxes!
+    return true if params[:project].empty? ||
+                   params[:project][:ignore_proj_dates]
```
JoeCohen, line 200: If `params[:project][:ignore_proj_dates]` is `true` = user has overriden project date restrictions and is adding an out-of-range Observation to a Project.
#### Question
Should the Project admins be notified automatically (or have the option to receive automatic notices)?
IMO yes, but not an immediate priority, and I don't know how this would look as option.
Ideally each admin would have an option for each project, but that's hard.
Easier to have one option that applies to all admins on a project.
Or users could have a user option that applies to all projects that they administer.

app/controllers/observations_controller/form_helpers.rb
mo-nathan: Yes, we should get rid of accepting_observations.  It's now redundant.

app/views/projects/show.html.erb
mo-nathan: Looks like @nimmolo code based on `git blame`.  Looks dead to me.

mo-nathan:
> @mo-nathan:
> 
> ### Overriding the Project Dates
> @mo-nathan’s original suggestion for what happens if a user checks a box for an out-of-range project when creating an Observation included:
> 
> > it saves the Observation "if the users] resubmit with the same 'when' value and projects selected."
> 
> I did it slightly differently: There’s an “Ignore Project Dates” checkbox. If the user checks that and resubmits, then it saves.
> 
> * I couldn’t figure out how to code @mo-nathan's exact suggestion with only Ruby. (I’d guess there’s a way to do it in js, but that’s beyond me.)
> * Using a checkbox makes it slightly more difficult to accidentally add an Observation to an out-of-range Project.

I have dug into it, but I believe the way it gets done (or at least has been done in the past) in pure Rails in other places in our code is by having a hidden attribute that contains the previously specified value.  Nimmo may have changed it, but a similar case is the create observation workflow when an unknown or deprecated name is given.

config/locales/en.txt
mo-nathan: Seems ok.  I'd probably just use `:NONE.t` directly instead of creating a translation specifically for this.

app/controllers/observations_controller/new_and_create.rb
mo-nathan: I'm guessing that 90% of the cases when this happens (if it happens at all) will be a Project admin adding something for some reason.  If some excited user does a bunch of these, I don't think we want to spam the Project admins with a bunch of email.  I think a better solution would be to have a way to indicate and/or search for observations that are outside the official project criteria.

mo-nathan:
In my opinion project start and end dates should not be required.  I'm not sure if they should be opt-in or opt-out, but I think they should be activated/deactivated with a check box.  Whether that gets stored in a separate column in the database or not, I don't have a strong opinion.  I also wonder if it would be better to default to some period of time (a week?) rather than defaulting both dates to today.

app/models/project.rb
nimmolo: Not a bad idea, and they're pretty easy to write.

app/views/projects/show.html.erb
nimmolo: Yes - although it was probably there before in similar form. When i encountered stuff like this in the view sweeps, i often left it there.

mo-nathan:
I also wonder if rather than an end date there should be a time period.  Right now you can create a project that ends before it begins.  At minimum it seems like it should check for that, but I think have a time period after the start time would be better.  I'm thinking you could enter a positive integer number and select days, weeks, and months.  You could add years, but I doubt that would get used much.  The show page should display the end date along with the start date and time period.

mo-nathan:
I'd change "There is a problem with Projects." to "This observation violates the constraints of one or more select Project."

mo-nathan:
I didn't get any complaint when I edited an observation to add it to a project where the observation violates the project constraints.

app/views/projects/_form.html.erb
nimmolo: Looks pretty good. I think your `<div>`s maybe need `class="form-group form-inline"` (or possibly just around the `date_select`'s three fields? Currently my styling reference is the [Bootstrap 3 form](https://getbootstrap.com/docs/3.4/css/#forms) docs.

I have a `select_with_label` but haven't written a `date_select_with_label` yet.

mo-nathan:
The project was still automatically selected when I shifted the project end date to before today.  It did complain about it, but it would be better if it deselected the project even if it was selected in m previous observation, but the date has now passed.

app/controllers/observations_controller/form_helpers.rb
JoeCohen: Done.

app/views/observations/form/_projects.html.erb
JoeCohen: Done.

app/controllers/observations_controller/new_and_create.rb
JoeCohen: I added this to my to-do list.

JoeCohen:
@ mo-nathan:
1. Validating start date not after end is is already on my to-do list.
2. Duration vs. end date: I'll ponder this.  
If I knew js, we could have it both ways: user could pick an end-date which would change the duration, and vice versa.
But without js, I think that for most short-duration projects, e.g., NAMA Forays, even your First Fantastic Fungal Friends Foray Festival!, it's easier for the admin to pick dates rather than a start date + duration. 

> I also wonder if rather than an end date there should be a time period. Right now you can create a project that ends before it begins. At minimum it seems like it should check for that, but I think have a time period after the start time would be better. I'm thinking you could enter a positive integer number and select days, weeks, and months. You could add years, but I doubt that would get used much. The show page should display the end date along with the start date and time period.

JoeCohen:
@nathan: If you like, I'll use a hidden field instead of a checkbox to override project constraints.
(But I'd rather have a checkbox for the reasons I've already given.)

> > @mo-nathan:
> > ### Overriding the Project Dates
> > @mo-nathan’s original suggestion for what happens if a user checks a box for an out-of-range project when creating an Observation included:
> > > it saves the Observation "if the users] resubmit with the same 'when' value and projects selected."
> > 
> > 
> > I did it slightly differently: There’s an “Ignore Project Dates” checkbox. If the user checks that and resubmits, then it saves.
> > 
> > * I couldn’t figure out how to code @mo-nathan's exact suggestion with only Ruby. (I’d guess there’s a way to do it in js, but that’s beyond me.)
> > * Using a checkbox makes it slightly more difficult to accidentally add an Observation to an out-of-range Project.
> 
> I have dug into it, but I believe the way it gets done (or at least has been done in the past) in pure Rails in other places in our code is by having a hidden attribute that contains the previously specified value. Nimmo may have changed it, but a similar case is the create observation workflow when an unknown or deprecated name is given.

app/views/projects/_form.html.erb
JoeCohen: I will leave this for later, as Bootstrap wants a lot of detail that I don't yet understand.

config/locales/en.txt
JoeCohen: I'll use :NONE directly. 
It's easy enough to use a translation in the future if needed.

nimmolo: Incidentally I would love to do a purge of our duplicate translation strings. The proliferation seems unmaintainable, doesn't it? Particularly for translators.

pellaea: The whole point of the `[:tag]` syntax was precisely to keep it maintainable.  This way there is no cost to duplicating strings, and that's what I wanted to encourage.  Who's to say that in Inuit or Cherokee they would use the same phrase as we do in English for this purpose?

nimmolo: Ohhhhhh. I didn't understand this!

mo-nathan:
> @nathan: If you like, I'll use a hidden field instead of a checkbox to override project constraints. (But I'd rather have a checkbox for the reasons I've already given.)
> 
> > > @mo-nathan:
> > > ### Overriding the Project Dates
> > > @mo-nathan’s original suggestion for what happens if a user checks a box for an out-of-range project when creating an Observation included:
> > > > it saves the Observation "if the users] resubmit with the same 'when' value and projects selected."
> > > 
> > > 
> > > I did it slightly differently: There’s an “Ignore Project Dates” checkbox. If the user checks that and resubmits, then it saves.
> > > 
> > > * I couldn’t figure out how to code @mo-nathan's exact suggestion with only Ruby. (I’d guess there’s a way to do it in js, but that’s beyond me.)
> > > * Using a checkbox makes it slightly more difficult to accidentally add an Observation to an out-of-range Project.
> > 
> > 
> > I haven't dug into it, but I believe the way it gets done (or at least has been done in the past) in pure Rails in other places in our code is by having a hidden attribute that contains the previously specified value. Nimmo may have changed it, but a similar case is the create observation workflow when an unknown or deprecated name is given.

Regarding the checkbox vs. hidden field, if it's important to you, I'm fine with the checkbox, it just creates inconsistency with other UIs.  Perhaps those should get changed as well.  Up to you.

I do think it should be a warning (yellow) rather than an error (red).  To me errors mean the user has made a mistake.  In this case they may well have done this intentionally and the system is simply warning them that they have done something surprising.

JoeCohen:
@mo-nathan: How about this for date range:
- Start Date: y, m, d selects (as-is)
- Duration: number select (defaults to 4), period select (days, weeks, years)

Potential issue: the owner has to know the duration. No problem for short projects or longer-range projects that are an integral number of weeks or years. Might be an issue for your foray. Do you know exactly how many days it lasts? 

> In my view it's debatable whether specifying a duration or picking a date is easier. If it had a reasonable default like 4 days I think it would be less clicks more often. We'd need some data to know what the best default is. There's not a strong reason for a NAMA foray, for example, to be exactly 4 days vs. 1 week. So I could do that with two clicks and a selections. In the case of my upcoming foray, the current UI requires 4 clicks and selections. If it had a default of 7 days, it would take 3 clicks, 2 selections and typing two characters. I could also just do 2 clicks and then update the end date if/when it started pushing back. To me the biggest win is that the user can't enter enter invalid data.
> 
> Realistically, I think the date limits are primarily useful for short week long-ish events. If that's accurate, then thinking it over, I'd recommend just asking for a number of days and defaulting to 4 days. I think projects with dates that are more than a month will be rare and I think a lot of forays and mushroom festivals are 4 days-ish (NAMA, MSA, NEMF, Tulleride, etc.).

mo-nathan:
If you want to put in the period select that fine as long as it defaults to days.  I expect most people will use days.  For my own foray, it depends on whether I include the 8th and the 17th or not.  I probably would so I'd put in start date of Dec 8 and give the project 10 days.  If for some reason that's wrong, I can easily fix it.

config/locales/en.txt
pellaea: Thanks, @mo-nathan.  Certainly.  I reuse these simple one-off words all the time, too.

FWIW, I was responding to and pushing back against a perceived desire to simplify more than necessary, and against the worry that duplicates incur maintenance debt.  I felt that needed to be clarified, and share why I designed it the way I did.

In short, I guess, I advocate leaving it the way it is.  I don't think it gains us anything to put in even a few hours' work to remove all of the "trivial" duplicates (the strings that are literally just synonyms of other terms with nothing added or modified).  They do no harm.  It does cost us developers or admins, and they don't even cost translators anything.

mo-nathan: Totally fine with that.  Doesn't seem like a rabbit hole worth going down at the moment.

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -104,6 +104,7 @@ def create
     success = false unless validate_name(params)
     success = false unless validate_place_name(params)
     success = false unless validate_object(@observation)
+    success = false unless validate_project_checkboxes!
```
mo-nathan, line 107: Curious why the '!' at the end?  I think of that as meaning that the method is dangerous like raising an exception like update! or save! or has significant side effects (foo.downcase! vs bar = foo.downcase).  While technically this method changes the object by potentially setting some ivars, that seems like business as usual for a controller.  For example, most CRUD methods set ivars and don't have '!'s.  I guess when I see a token ending in '!' I expect there to be a "safer" method with the same name and no '!'.  There's an interesting discussion here: https://stackoverflow.com/questions/612189/why-are-exclamation-marks-used-in-ruby-methods#612196 that talks about both perspectives.

JoeCohen:
I am switching development to a new branch [jdc-project-add-dates](jdc-project-add-dates).

(I tried merging main into #1619. 
Result: disaster. Too many conflicts, plus too hard to deal with db issues.
Plan: 
- Manually (and grandually) add tests from #1619 to [jdc-project-add-dates](jdc-project-add-dates).
- Use TDD to fix test failures.)

JoeCohen:
Development transferred to PR #1658.

PR #1624, opened 2023-09-21
Serve images locally if they don't get transferred.

pellaea:
In particular, this fixes the rails side.  There's an independent fix on the nginx side.

Here, we just need to make sure the entire `Image` instance is passed all the way in to the point where we calculate the image URL.  At least whenever possible.  There are still some cases (e.g. in textile markup such as `!image 1234!` where we do not have access to the instance and we just fall back on the default source.  But in the standard cases like the activity log and show obs page, we've eager-loaded the images already, so there is no reason not to pass it all the way in.

### DIFF
--- a/app/helpers/image_helper.rb
+++ b/app/helpers/image_helper.rb
@@ -105,6 +105,7 @@ module ImageHelper
                       votes: true) + image_copyright(obs.thumb_image, obs)
   end
 
+  # pass an image instance if possible, to ensure access to fallback image.url
   def original_image_link(image_or_image_id, classes)
     url = if image_or_image_id.is_a?(Image)
             image_or_image_id.url(:original)
--- a/app/helpers/image_helper.rb
+++ b/app/helpers/image_helper.rb
@@ -105,12 +105,22 @@ module ImageHelper
                       votes: true) + image_copyright(obs.thumb_image, obs)
   end
 
-  def original_image_link(image_id, classes)
-    link_to(:image_show_original.t, Image.url(:original, image_id),
+  def original_image_link(image_or_image_id, classes)
+    url = if image_or_image_id.is_a?(Image)
+            image_or_image_id.url(:original)
+          else
+            Image.url(:original, image_or_image_id)
+          end
+    link_to(:image_show_original.t, url,
             { class: classes, target: "_blank", rel: "noopener" })
   end
 
-  def image_exif_link(image_id, classes)
+  def image_exif_link(image_or_image_id, classes)
+    image_id = if image_or_image_id.is_a?(Image)
+                 image_or_image_id.id
+               else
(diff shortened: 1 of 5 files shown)


pellaea:
I'm adding the (tiny) change to the nginx configuration to this PR, too.  It's just one of those strange nginx order things.  Even though the `rewrite` section appears at the end of the file, it still fires before the `try_files` business, because `rewrite` always fires before `try_files` regardless where they occur in the configuration file.  The solution is to shove the order-critical things in `location` blocks, because _those_ get executed in the order they appear.  Nice and easy, ya just gotta know.

nimmolo:
Can we add a comment on the helpers to remind people the reason for these instance/id fallbacks?

pellaea:
Sorry, I haven't written a unit test for this.  It will require temporarily changing `MO.image_precedence` for the space of at least one or two tests.  We need to have it try `:remote1` first then fallback on `:local`, not the other way around.  Then we can set `transferred` to false for some image that is present locally, then make sure the `image.url` for that image is local not remote.  Otherwise, if it prioritizes `:local`, then of course it will serve the local image, that won't test whether it's testing the `transferred` bit correctly on the remote image.  And then, even more important, because I think the `Image::URL` stuff is sufficiently tested already, we need to make sure the URLs that are rendered on the Activity Log and Show Obs pages are correct, too.  Phew.  Beyond the extremely limited scope of time I have available to me.  I tested this in development and it works fine.  So I'm just letting it go for now.  Besides the nginx configuration should catch anything we miss in rails, anyways.  We had a problem this time just because *both* failsafes were failing.

pellaea:
Absolutely, comment away!

app/helpers/image_helper.rb
```diff
@@ -105,12 +105,22 @@ def show_best_image(obs)
                       votes: true) + image_copyright(obs.thumb_image, obs)
   end
 
-  def original_image_link(image_id, classes)
-    link_to(:image_show_original.t, Image.url(:original, image_id),
+  def original_image_link(image_or_image_id, classes)
```
nimmolo, line 108:
```suggestion
  # pass an image instance if possible, to ensure access to fallback URL
  def original_image_link(image_or_image_id, classes)
```

app/helpers/lightbox_helper.rb
```diff
@@ -43,11 +44,11 @@ def caption_obs_title(obs_data)
   end
 
   # links relating to the image object, pre-joined as a div
```
nimmolo, line 46:
```suggestion
  # links relating to the image object, pre-joined as a div
  # pass an image instance if possible, to ensure access to fallback image.url
```

nimmolo:
Comments added

PR #1626, opened 2023-09-23
Update MACOSX_NOTES

nimmolo:
Major gotcha discovered while trying to run Rails natively on Mac OSX.

All is annotated in the readme changed by this pull request.

### DIFF
--- a/MACOSX_NOTES
+++ b/MACOSX_NOTES
@@ -19,6 +19,37 @@
 
 brew install git mysql exiftool libjpeg shared-mime-info openssl imagemagick findutils
 
+# If you haven't done so already, install a recent version of Bash and set it
+# as the default. You can find your installed version like this:
+# bash --version
+
+# Apple only includes Bash 3.2 from 2007 in all versions of Mac OS X even now,
+# because Bash > 4.0 uses GPLv3, and they don't want to support that license.
+# MO's scripts use syntax that requires Bash >= 4.0.
+# Reference: https://medium.com/p/7138bd1066ba
+# https://stackoverflow.com/questions/6047648/associative-arrays-error-declare-a-invalid-option
+# To install the newest Bash alongside older versions:
+# brew install bash
+
+# now check paths of all versions installed
+# which -a bash
+# you should get something like this
+# /usr/local/bin/bash
+# /bin/bash
+# check those versions
+# /usr/local/bin/bash --version
+# /bin/bash --version
+# now check
+# bash --version
+# should be the new one.
+# So far, so good. Now you have make this version the default.
+# sudo vim /etc/shells
+# (add at the bottom of the list: /usr/local/bin/bash and save out)
+# Set the default shell:
+# chsh -s /usr/local/bin/bash
+# Set the default shell for the root user:
+# sudo chsh -s /usr/local/bin/bash
+
 # Configure MySQL
 brew services start mysql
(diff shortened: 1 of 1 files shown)


MACOSX_NOTES
```diff
@@ -19,6 +19,37 @@
...
+# you should get something like this
+# /usr/local/bin/bash
+# /bin/bash
+# check those versions
+# /usr/local/bin/bash --version
+# /bin/bash --version
+# now check
+# bash --version
+# should be the new one.
+# So far, so good. Now you have make this version the default.
```
mo-nathan, line 45: This isn't exactly what's true on my system.  I believe the real key is what `/usr/bin/env bash --version` returns since that's what's in the shebang line of our scripts.  For me, `which bash` returns `/opt/homebrew/bin/bash` and this is the version that gets run by `/usr/bin/env bash`.

MACOSX_NOTES
```diff
@@ -19,6 +19,37 @@
...
+# sudo vim /etc/shells
+# (add at the bottom of the list: /usr/local/bin/bash and save out)
+# Set the default shell:
+# chsh -s /usr/local/bin/bash
+# Set the default shell for the root user:
+# sudo chsh -s /usr/local/bin/bash
```
mo-nathan, line 51: This is not needed.  I use zsh (the default in current MacOSX) and don't have any issues.

nimmolo: Huh. OK, I use zsh too. 

I had believed this was the line that made the script finally work, but i may have been "over-hammering the nail" at that point and not paying close enough attention. In any case -- does this line change the shell app? My impression was it just told it which environment to run in.

MACOSX_NOTES
nimmolo: Note: we have a couple scripts that still run from `#!/bin/bash`: `dump` and `strip_checkpoint`. I don't suppose anyone would run those locally, though.

Others do run from `#!/usr/bin/env bash`.

MACOSX_NOTES
nimmolo: Turned it into a note reference.

nimmolo: @mo-nathan Please feel free to edit further.

mo-nathan: I think `chsh` changes your default shell.  So after running that I don't think you'll be using zsh any more.  Probably the better solution would be to tweak the scripts so they don't care which version of bash they are using.  Looks like you could use $BASH_VERSION to make this decision.  While the whole Apple bash/zsh thing is dumb, it's even dumber than the new version of bash is not backwards compatible with the old version.  I appreciate they are trying to make a point, but breaking the world is not a good way to make new friends.

mo-nathan: Actually looking into it that may not be accurate.  I'm not sure old bash had support for associative arrays.  So maybe it is backwards compatible.  Maybe the scripts should just be converted to ruby or python.

nimmolo: That's right. That's the whole reason for the glitch - Apple's ancient Bash wouldn't make an associative array.

nimmolo: Clarification - of course you're right, i'm in fact running the new bash, not zsh, after running that command. I didn't notice because the prompt style of the updated bash is just like zsh (but not like the old bash), and i didn't check the window frame, which clearly says i'm running bash!

PR #1628, opened 2023-09-23
Open the Obs homepage for not-logged-in visitors

nimmolo:
This is a very belated change that ideally should have happened when we switched homepages. This should finally permit metrics to start happening on the homepage!

Conversely, requires login on Activity Logs, no longer the homepage.


JoeCohen (approved):

nimmolo:
Added tests to make sure it's working

PR #1631, opened 2023-09-24
Split out classes from app/classes/auto_complete.rb

mo-nathan:
so it plays better with the autoloader.

To test:
Load http://localhost:3000/ajax/auto_complete/name/b
Modify some of the rails code
Reload http://localhost:3000/ajax/auto_complete/name/b

On main the above should fail.  On the branch it should succeed.  You can replicate the same behavior by trying to use the autoloader from the Create Observation page instead of loading the AJAX URLs.


nimmolo:
For folder sanity, does it make sense to put these in a new folder`classes/auto_complete` and namespace them:

`class AutoComplete::ByWord < AutoComplete`
`classes/auto_complete/by_word.rb`

nimmolo:
@mo-nathan if you agree with the idea, i'm happy to take care of the folder/namespacing on this PR.

mo-nathan:
Sure.  You're welcome to run with it if you want.  I looked at it a bit and ran into some more subtle issues.  At least some of them are because we don't want multiple classes with the same name, so I would avoid AutoComplete::Location (::Name, ::SpeciesList, etc.).

mo-nathan:
I think I have something working on this branch `njw-autocomplete-rename`.  I need to do some manual testing, but the tests and rubocop seem happy with it.

mo-nathan:
After some manual testing since I didn't see any activity on this branch, I went ahead and merged that other branch into this one.  I also pulled in the latest change from main.

nimmolo:
Ha - oh shoot i was just doing it myself. Looks good! 

For future reference: you just have to call the top level classes with :: to signify top level namespace, within the subclass. Like this:

::Location[:name]

But I prefer your `ForThing` style.

mo-nathan:
I tried the ::Location thing, but the tests were very grumpy about it.  I also think it can give the autoloader conniptions.  It also just a bad idea to have classes with the same name.  We ran into a major but very subtle bug at work a while back where things were working correctly in one context and failing in another because of exactly this problem.

nimmolo (approved):

mo-nathan:
I'm close to 100% coverage.  The only exception is app/classes/auto_complete/for_name_above_genus.rb which has 0% coverage even though I added an autocomplete test for "Fungi".  It's not clear to me if this autocompleter actually gets called from anywhere.  Is this dead code at this point?

nimmolo:
Could be, although this was intended to be the autocompleter that was called on the Help Identify index filter at the top of the page. 

When i get a moment i'll check if the filter is returning "names above genus" without using this autocompleter. If so, we can safely delete it. 

I'm about to head out, but maybe you can check that Help Identify's filter only returns names above genus, and if it does, i'd say go ahead and delete. But i can't imagine how it would do that, without this autocompleter.

mo-nathan:
Figured out a test for it.  I think we're good.  Thanks!

nimmolo:
Ok. Weird, i just checked and it does indeed return only names above Genus. But how?!
LOL somehow i wrote it...

nimmolo:
Ok, it does actually call that autocompleter. The keyword is `:clade`, which routes to `NamesAboveGenus`. 

We could alternatively rename the class `ForClade`. That seems clearer.

mo-nathan:
I'll leave it to you to decide if the clade rename is worthwhile.  Note that the test I wrote uses "Fungi" which is a Kingdom and I don't think we have any clades in the fixtures.  Personally, I'm good with this PR as it is.  I'll let you decide when to merge it at this point.  Happy to look at additional changes if you decide to make them.

nimmolo:
Thanks for all this! I'll merge it as is. 
I'd also need to alter the Autocompleter js if I rename it "clade"... can do that separately.

Hope I get the js integration tests working soon!

PR #1632, opened 2023-09-24
Remove support for IE < 11

nimmolo:
Removes the `single_image_upload.js` script and associated form `observations/form/images_upload/mono.html.erb`.

Removes the helpers `javascript_helper.rb#modern_browser?`, `#can_do_ajax?`, and `#can_do_multifile_upload?` and the very few calls to these.


nimmolo:
The `ObservationFormIntegrationTest#test_post_edit_and_destroy_with_details_and_location` only tests the single image upload form, because we don't have Capybara running with JS yet. 

The sooner we can get that working, the better.

JoeCohen:
If you can suggest some manual tests, I'll try it out.

nimmolo:
@JoeCohen - this would mean getting a test case class working that uses Firefox headless, that we discussed in another thread. We can't test the image uploader without that. 

I believe you could handle it, it's just troubleshooting why the test case won't load properly (Ruby errors).

JoeCohen:
I meant a manual test.

nimmolo:
There's no manual test that will fix the failing integration test on this branch. It needs Capybara working on all cylinders.

JoeCohen:
Gotcha. We were talking about two different things.

PR #1633, opened 2023-09-24
Get Capybara working for 'selenium_headless'

nimmolo:
Having JS-enabled integration tests would allow us to skip a lot of the manual review of these JS PR's. 

The pieces are almost in place. I have (theoretically) set up a Selenium driver to run tests as `selenium_headless`, according to [Capybara docs](https://rubydoc.info/github/teamcapybara/capybara/master#selecting-the-driver) and [examples](https://gist.github.com/vhorb/dec95ecaf6f4899236a610f0d0a0dba9). We have Firefox installed on the VM, too — and, I believe, in CI.

The new `Capybara::Driver` is defined in `test/capybara_integration_test_case.rb`. The way to use it is simply to switch drivers at the top of a test, in a `set_up` block, as shown in this minimal PR. Until now, I haven't been able to get that working. 

Rather than try to get a whole existing integration test running under that driver, I've set up a new `firefox_integration_test.rb`, that simply logs in and tries to click on a button. ~Trying to load this test will throw errors before it even runs.~

~The errors seem to involve the order of includes/requires in `CapybaraIntegrationTestCase`.~

The errors involve AJAX loading elements on the page. Stay tuned, fixes are coming!


nimmolo:
Hold the phone! I believe i've found the solution.

JoeCohen:
For the next few days, I probably will work only on the Project Dates
branch.

   - That branch is proving to be more complex than I imagined;
   - I'm way behind where I need to be; and
   - It's difficult to switch back and forth because of a migration issue.

nimmolo:
Amazing! Thank you Joe. 

I will check on the `webdrivers` dependency - Selenium was strongly recommending using that, the last time I worked on this, so that's a surprise.

nimmolo:
@JoeCohen You gave me the key clue. 
The `webdrivers` gem is no longer required, as of very recently. 
Whew. The tests have to be rejiggered a bit, but they run!

PR #1638, opened 2023-09-26
Autocompleter - tab/return tweaks

nimmolo:
This PR restores some expected `tab`/`return` behavior to the autocompleter, that i probably messed up in the refactor.

- Prevents submitting the form if you hit `return` within an autocompleter.
- `tab` doesn't do anything if the autocompleter is active and nothing is highlighted
- If something is highlighted, `tab` fills the field with that value and waits for further input

Thanks to @mo-nathan for examining it closely and writing it up.

>I've been playing with the auto-completer as part of testing the project location work and I've noticed something I don't think is right.  Specifically, if you go to "Create Observation", click the location, and use the auto-completer from the keyboard and select an item using the arrow keys and then hit <return>, it immediately submits the form.  I expect it to just select that item and fill in the field with it.  This means I either have to remember to hit tab (which has other issues discussed below) or click on the selected item.

>This in turn has made me aware that hitting <tab> doesn't behave the way I expect it to.  I expect it to extend the text to the next ambiguous character, but instead it only adds to the text if an item is selected and regardless it moves the focus to the next item in the form.  This means I can end up with locations like "USA, Mass", when I expected it to extend the text to "USA, Massachusetts".

>This case is clearly trickier since we do more complicated matching than just against the beginning of the string.  A good example at the moment is "USA, Massachusetts, Falmouth" since there is currently a location named "USA, Massachusetts, East Falmouth, Kettle Holes Conservation Area" as well as various names starting with "USA, Massachusetts, Falmouth, North Falmouth".  If you see a way to really solve this situation, that would be great.  However, if not, in my view it would still be better if <tab> didn't do anything if the autocompleter is active and nothing is selected.  If something is selected, I think it's reasonable to fill it in with that text and move to the next item.

### DIFF
--- a/app/assets/javascripts/mo_autocompleter.js
+++ b/app/assets/javascripts/mo_autocompleter.js
@@ -354,6 +354,7 @@ class MOAutocompleter {
           break;
         case EVENT_KEY_RETURN:
         case EVENT_KEY_TAB:
+          event.preventDefault();
           if (this.current_row >= 0)
             this.select_row(this.current_row - this.scroll_offset);
           break;


mo-nathan:
Just look at this and it's a huge improvement.  One more (hopefully) tiny nit: I notice when I first load the Create Observation page that I have to click twice on items with the auto-completer.  I can even click on the location one, then the name one, then the location one again and now the location one is active.

nimmolo:
I'll check that out. The initializer may be waiting on the rest of the page to load. 
Hm.

nimmolo:
@mo-nathan - I'm not getting that behavior, but i will keep checking for it. 
For me locally, the autocompleters on the create obs form seem to be immediately initialized.

mo-nathan:
@nimmolo What browser are you on?  I'm using Chrome 116.0.5845.187.  I also see the problem if I load the Create Observation page, click after the 2023 in the year selector above the Locality and then hit tab.  The Locality field doesn't activate.  If I then hit <shift>-tab> followed by another <tab> it now accepts key presses (and has a blue outline).  It's essentially the same behavior as with clicks.  I can create a video of it if that would help.

nimmolo:
I'm on Safari usually. Appreciate the precise description, that's definitely not what I get. 

Waiting for Chrome to update itself, it won't launch for some reason. 
I'm going to take a break for falafel and come back.

mo-nathan:
I tried it on Safari and you're right it works like a champ.

Issue #1641, opened 2023-09-27
Auto GPS + datetime extraction not working


Lodorenos:
Hi, when using the image uploader I see errors in the console.

![image](https://github.com/MushroomObserver/mushroom-observer/assets/2823835/2f33f712-05a1-4151-a230-50ceca48a53d)

![image](https://github.com/MushroomObserver/mushroom-observer/assets/2823835/91dcdff4-7c98-4b4c-b3e8-1ad060faecf6)

The variable is defined as a constant and assigned a new value afterwards. This fails, preventing the extraction of location and GPS data.

Link to code:
https://github.com/MushroomObserver/mushroom-observer/blob/main/app/assets/javascripts/mo_multi_image_uploader.js#L290

I can create a PR later today, unable to do so now.

nimmolo:
Thanks Lodorenos! 

I got it, will handle now.

nimmolo:
@Lodorenos If you get a chance can you please email me an example image of yours to test with? I think my test images don't have the necessary data. `andrnimm @ fastmail.fm`

nimmolo:
fixed by #1643

PR #1643, opened 2023-09-27
Fix radio containers in `mo_multi_image_uploader.js`

nimmolo:
Fixes incorrect JS that builds and adds "radio" fields to the form in a box.

Also relates to #1641


nimmolo:
Can someone please test creating an obs with inconsistent dates (vs. the image)? I can't seem to find an image that will induce the form to create the dialog box where they have to pick one date or the other — I don't know if it's a setting on my phone, or what.

There should also be a dialog for extracting GPS, I think, but I can't get that to load either.

I'll be glad when I can write JS tests. Hoping the capybara maintainer will answer my SO question.

nimmolo:
Pushing this out because it's better than what's on production currently, but I have no idea if it works, and I'm still interested in manual tests. 

Alternatively, please email me test images with known GPS data and datetimes.

PR #1650, opened 2023-09-29
Project#show cleanup

mo-nathan:
Overhaul of the Project#show page.
Supports background images for Projects.
Moved all the member stuff to the members page and added user thumbnails.
Fixed some lingering image upload bugs in Profile which I used as a model for Projects.
Rearranged a bunch of stuff on the page to make it a bit more sane and more aesthetically pleasing.

I did not add icons or maps or move the violation count.  I don't think these are really needed at this point, but maybe we'll do those at some point.

I do want the numbers to be dynamic, but that's JS code and @nimmolo says it will be better once we have Turbo in Rails 7.


JoeCohen:
The edit project page needs a way to modify the start/end dates.
I think that's my responsibility. What's the best way:
- Merge this PR into #1658 
- Vice versa? (It might take me a half-day or to be ready to merge in that direction.)

PS: I have more time to work on Projects; some non-MO things I'd been through the weekend won't happen. 
(It's the silver lining of COVID.)

JoeCohen (commented):
I like the layout. 
The PR generally looks good (other than issues mentioned in my comments.)

mo-nathan:
> Is the following-expected behavior in dev mode? When I upload an image file locally (on my VM) it uploads a random file.

I saw that behavior as well.  I believe what it's doing is initially grabbing the image from our images server using the same id, but then later it settles back on the image you uploaded.  I didn't dig into it since I don't expect it to be a problem in production and I don't think it's due to anything in my code.

PR #1654, opened 2023-10-01
Update ci_rails.yml to run system tests

nimmolo:
@JoeCohen - thanks for the tip about starting to set this up.

### Requires a browser ###
CI needs [Firefox](https://github.com/browser-actions/setup-firefox) (or any browser, but i'm using FF) to run a system test.

### Run ###
Note: `rails test:all` now works.


nimmolo:
Seems to work! 

Let's go!!!

nimmolo:
Hmm, the system test is not getting run in sequence here. Back to the drawing board, but we're close.

PR #1655, opened 2023-10-01
Ruby 3.1.4 and bundle update

nimmolo:
We can't switch to Ruby 3.2 while we're on Rails 6, but it seems like a good idea to keep it updated.


mo-nathan:
Ugh.  Trying to upgrade to 3.1.4 broke my world.. Can't even get things to work going back to 3.1.2.  I'll need to poke at it in the morning after I sleep.

nimmolo:
Was thinking about this. I wonder if what you need to do is `bundle update` on your machine.
I noticed that our gemfile.lock has hardware-platform-specific versions of a few gems, and if they were missing…?

Sorry, the whole update was super smooth for me. I just changed ruby-version and bundle-update and it worked.

mo-nathan:
I think I got it sorted and I'm not on MacOS Ventura and fully brewed up.  All the tests are passing both on main and on my branch.  I had to wrestle a bit with the mysql2 gem, but I think I finally won.  There's a minor tweak that needs to go in the Gemfile.lock at some point, but I have it on my branch so I think I'll just wait for that to merge.

mo-nathan:
This PR did not get pushed to production so I was surprised when I found that 3.1.4 wasn't installed there.  For the moment I update the .ruby-version on production to stick at 3.1.2.

mo-nathan:
Doing the bundle install for 3.1.4....

mo-nathan:
Ok, 3.1.4 is installed on production and production is flipped over to it.

nimmolo:
Thanks Nathan. On Oct 5, 2023, at 4:38 PM, Nathan Wilson ***@***.***> wrote:﻿
Ok, 3.1.4 is installed on production and production is flipped over to it.

PR #1658, opened 2023-10-02
Add date range to Projects

JoeCohen:
Replaces PR #1619 
Re-delivers [Pivotal #186018784](https://www.pivotaltracker.com/story/show/186018784)

Adds Project start and end dates.
- Project admin sets start and end dates when creating/editing a Project. Those dates can be indefinite.
- Those dates display on the Project page.
- When a Project member creates an Observation:
  - If the ***current date*** is outside the Project date range, the Project is not automatically selected for the Observation;
  - If the User manually selects the Project and the ***Observation  date*** (`when`) is outside the Project date range, then
    - The User receives a warning (after clicking the Obs Create button) and a chance to correct by:
      - Deselecting the offending Project,
      - Changing the Obs date to be within the Project Range
      - Overriding the warning by checking "Ignore Project Constraints".

See
-  PR #1619 for previous discussion
- [Open Project Stories for MO](https://docs.google.com/document/d/1g8QMR6NzUvM4pjuAuBiI9QGZU_XNail-ocMu120UBxw/edit), Story 1, including comments, for more background.


app/controllers/observations_controller/form_helpers.rb
```diff
@@ -84,20 +84,17 @@ def init_project_vars
     @projects = User.current.projects_member(order: :title,
                                              include: :user_group)
     @project_checks = {}
-    @projects.each do |proj|
-      @project_checks[proj.id] = (proj.open_membership &&
```
JoeCohen, line 88: @mo-nathan I didn't understand the rationale for:
- having any projects checked initially
- treating open projects diferently from closed projects

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -84,20 +84,17 @@ def init_project_vars
...
-    end
+    @projects = @projects.union(obs.projects)
     @projects.each do |proj|
-      p = params[:project]
-      @project_checks[proj.id] = p.nil? ? false : p["id_#{proj.id}"] == "1"
+      @project_checks[proj.id] = if params[:project].nil?
+                                   false
+                                 else
+                                   params[:project]["id_#{proj.id}"] == "1"
+                                 end
```
JoeCohen, line 97: Eliminates repeated assignment to `p`.

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -84,20 +84,17 @@ def init_project_vars
...
-                                  proj.accepting_observations)
-    end
   end
 
   def init_project_vars_for_reload(obs)
     init_project_vars
-    obs.projects.each do |proj|
-      @projects << proj unless @projects.include?(proj)
-    end
+    @projects = @projects.union(obs.projects)
```
JoeCohen, line 91: ruby method instead of a rails loop

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -67,9 +67,14 @@ def defaults_from_last_observation_created
     %w[when where location is_collection_location gps_hidden].each do |attr|
       @observation.send("#{attr}=", last_observation.send(attr))
     end
+
     last_observation.projects.each do |project|
-      @project_checks[project.id] = true unless project.open_membership
+      next if project.open_membership ||
```
JoeCohen, line 72: Rationale for different treatment of open vs cosed projects?

app/controllers/observations_controller/form_helpers.rb
```diff
@@ -84,20 +84,25 @@ def init_project_vars
...
-                                  proj.accepting_observations)
-    end
+
+    # FIXME: Why wouold open-membership projects be treated different from
+    # closed-membership projects for initial auto-checking?
+    # And why auto-check non-compliant projects?
+    # @projects.each do |proj|
+    #   @project_checks[proj.id] = proj.open_membership
+    # end
+    # 2023-10-05 jdc
```
JoeCohen, line 94: @mo-nathan: 
Please see the FIXME.
Also note that I changed functionality:
No projects are checked when a user creates an Observation when there's no recently-created Observation.
I didn't understand why any projects wold be checked in that scenario.
I also didn't understand why open- and closed- projects would get different treatment.
If you feel that open-membership projects should be checked in this scenario, I'll fix things up.
I'll need to redo some tests. I should probably also insure that projectgs are not auto-check if constrains are violated.

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -67,9 +67,16 @@ def defaults_from_last_observation_created
...
+
+    # FIXME: Why woold open-membership projects be treated different from
+    # closed-membership projects for check-box persistence?
+    last_observation.projects.where(open_membership: false).
+      find_each do |project|
+        next unless project.current?
+
+        @project_checks[project.id] = true
+      end
+
```
JoeCohen, line 79: @mo-nathan 
See FIXME and comment above about related FIXME.
Note that I did not change the functionality here. (I forget whey. Probably easier because the change broke some tests.)
I would like to change the functionality so that project checkboxes persist identically for open- and closed- membership projects.

app/models/project.rb
```diff
@@ -131,16 +132,22 @@ def violates_location?(obs)
...
 
-    count = observations.where.not(lat: nil).count
-    count - observations.in_box(n: location.north, s: location.south,
-                                e: location.east, w: location.west).count
+    count =
+      out_of_range_observations.or(observations.where.not(lat: nil)).count
+    count - in_range_observations.
+            in_box(n: location.north, s: location.south,
+                   e: location.east, w: location.west).count
   end
```
JoeCohen, line 147: @mo-nathan 
Would you mind double-checking the logic of the revised `count_violations`.

app/views/projects/_form.html.erb
```diff
@@ -2,8 +2,42 @@
...
+      <p>
+        <%= radio_button(
+              "end_date", "fixed", true, checked: @end_date_fixed
+            ) %> <%= :form_projects_date_fixed.t %>
+      </p>
+      <p>
+        <%= radio_button(
+              "end_date", "fixed", false, checked: !@end_date_fixed
+            ) %> <%= :form_projects_date_indefinite.t %>
+      </p>
```
JoeCohen, line 38: @mc-nathan, @nimmolo 
Not a big deal. But is there a way to set up these checkboxes so that the params are:
1. nested under [:project] (rather than being top-level); and
2. names something like `fixed_start` and `fixed-end` with values T/F (or 1/0).

db/schema.rb
```diff
@@ -513,6 +513,8 @@
     t.boolean "open_membership", default: false, null: false
     t.boolean "accepting_observations", default: true, null: false
```
JoeCohen, line 514: @mo-nathan:
I removed all code references to `accepting_observations`.
But I left to column here for now in order to facilitate switching branches.
I plan to remove the `accepting_observations` column after this PR merged.

JoeCohen:
@mo-nathan, @nimmolo Do you have any suggestions how to fix this:
There's some unexpected behavior with a greyed-out, inoperative Create button, and some seemingly incomplete js interaction between that button and other elements.
I'd like to change that, but have no idea what's provoking the behavior or how to fix it.

To replicate:
- Create an Observation with 
  - Locality “McMurdo Station, Antarctica”, 
  - Latitude “-49”, 
  - Longitude “0”
  - checked box for the Project you just created.
Result:
Form reloaded with appropriate warning, 
but with Create buttons grayed out and inoperable, 
and unexpected interaction between the Create button and other parts of the form.
- Click “Ignore Project Constraints” checkbox
Result: Create button no longer grayed out and becomes operative.
- Unclick “Ignore Project Constraints”
Result: Create button remains operative.
- Click Create
Unexpected Result: Observation is created and added to the Project. (Instead, the form should be reloaded.)

JoeCohen:
@mo-nathan. One can create a Project that has no Locality.
Is this expected behavior?

mo-nathan:
Most existing project have no locality and no specific dates.  I expect we will want to continue that.  For example, I expect to make a project at some point related to creating Visual Groups.  Existing project that I would never expect to have a locality:

[Honey bees and fungi](https://mushroomobserver.org/projects/328?q=1pCBR)
[Earth’s Identified Psilocybin Fungi By Locality](https://mushroomobserver.org/projects/325?q=1pCBR)
[Biology Project 1](https://mushroomobserver.org/projects/319?q=1pCBR)
[FunDiS ID and Quality Project](https://mushroomobserver.org/projects/298?q=1pCBR)
[News Articles](https://mushroomobserver.org/projects/213?q=1pCBR)
[Rainbow world of Fungi](https://mushroomobserver.org/projects/25?q=1pCBR)
[Porcini](https://mushroomobserver.org/projects/226?q=1pCQR)
[Xylaria](https://mushroomobserver.org/projects/203?q=1pCQR)
[My Cology is Better than Yours](https://mushroomobserver.org/projects/26?q=1pCQR)
[Exploring Gomphidiaceae](https://mushroomobserver.org/projects/162?q=1pCQR)
[2010/2011 EOL University Species Pages Initiative](https://mushroomobserver.org/projects/7?q=1pCQR)

etc.

mo-nathan:
Regarding the grayed out buttons, yeah, I've noticed that as well.  I assume it has to do with the stuff that went in to avoid double clicking.  It's annoying, but you found the same workaround that I did of just clicking a checkbox.  It's JS so I didn't dig into it.  Would love to see it fixed.

app/controllers/observations_controller/form_helpers.rb
mo-nathan: @JoeCohen My rational is that if a user joins a project and it is currently active and they create a new observation it should by default go into that project.  Imagine someone at a NAMA foray.  They join a project and then immediately make an observation.  They pretty obviously want that observation to be in that project regardless of what they last chose.  Projects with closed membership are a bit different since in my view since there has to be some dialog there and I don't necessarily expect that all observations will be wanted there.  It's also been the historic behavior of those projects, so I didn't think I should change that.  However, I feel more strongly that folks joining open projects should have that project on by default if the constraints are met than I do about the current behavior of closed projects.

JoeCohen:
Thanks. What threw me is the the Projects form says "Locality:(required)"
Do we want to enforce this for new Projects?

mo-nathan:
Ooops.  Copy-pasta.  The (required) wasn't intended.  I think I copied from the Observation Create page where I believe it is currently required (although I've been wondering if we should get rid of that requirement once we start relying more on GPS).

app/views/projects/_form.html.erb
mo-nathan: I just looked at what iNat does (Andy Wilson gave me admin privileges on the most recent NAMA project).  They have one set of radio buttons which allow you to choose between "Any", "Exact" (single data), "Range" (with calendar selector for start and end date), and "Months".  I like the simplicity of "Any" and "Range" being the options rather the concept of "Indefinite".  I can't really think of a reason why I'd want a project that had just a start date or just an end date.  I don't think we have need for the "Exact" or "Months" options.  In this context, I would think if "Any" is chosen, then no date stuff is show on the Project#show page.  If "Range" is chosen, then I'd put the range on top of the project image like the location info.

db/schema.rb
mo-nathan: Makes sense.

app/controllers/observations_controller/new_and_create.rb
mo-nathan: I prefer the new behavior for open projects and I didn't want to change the behavior for existing projects.  Maybe it should really be related to whether the project has constraints.  If it doesn't have constraints, then it uses what was last set.  If it does have constraints then it is automatically selected when the constraints are currently met.  However, this implies getting the location info before the project selection happens.

app/controllers/observations_controller/form_helpers.rb
JoeCohen: Thanks for the explanation. I'll make open projects auto-checked if constrains are met.

app/views/projects/_form.html.erb
JoeCohen: Thanks. 
I like the simplicity of just Any exor Range.
Projects with just a start date or just an end date seem like an edge case, so I won't worry about them.
I'll think through how to do this. 
And It would be good to have UI that's understood by iNat project admins.
I think an actual calendar might require js, so I might go with date pickers for now.

test/controllers/projects_controller_test.rb
```diff
@@ -337,6 +491,7 @@ def test_bad_location
   end
 
   def test_project_save_fail
```
JoeCohen, line 493: @mo-nathan
I'd like to delete this test. 
- It fails as shown below, and I have no idea how to fix it.
- I don't understand its internals and purpose, other than to prove that if it isn't saved, it can't be found.
- Deleting it won't remove any line of coverage.

JoeCohen:
@mo-nathan:
I think this is pretty much ready to go.  I had [one question about deleting a test](https://github.com/MushroomObserver/mushroom-observer/pull/1658/files?show-viewed-files=true&file-filters%5B%5D=#r1349789885).

(I plan to ignore any CI system test failures that I can't reproduce locally. 
Per conversations with @nimmolo, they seem to be sensitive to 
how fast the CI engine executes Ruby vs the js, 
which seems to vary with time of day, phase of moon, etc.
@nimmolo is working useing Cuprise instead of Selenium , which hopefully will improve that.)

test/controllers/projects_controller_test.rb
mo-nathan: Actually deleting it causes line 345 in projects_controller.rb to not be covered.  I was essentially trying to get mocking working reasonable and it looks like I went down the wrong rabbit hole.  I think I have a better approach that I need to test some more and then I'll share it with you.

JoeCohen:
Thanks. I missed the newly missed line. (I was looking at the wrong data.)
I'll be interested to see the new approach.

JoeCohen:
How about https://mocha.jamesmead.org/. I've used it at least once, but
it's not thread safe.

mo-nathan:
I just created https://github.com/MushroomObserver/mushroom-observer/pull/1664 which I think solves the problem pretty well.  I only just saw your comment about Mocha and haven't looked into.

JoeCohen:
@mo-nathan: I'd like to get this merged. But one question about appearances.
The Observations (and Names and Members) buttons appear below and outside the "Summary" panel for any Project lacking Drafts and SpeciesLists:
proj obss outside summary
That seems strange.
Should I move them inside that panel:
proj obss inside summary

mo-nathan:
Sorry for the slow response.  I've been totally distracted the last couple of days.  I don't have a strong opinion about the location of those buttons relative to the summary. It's fine to tuck it into the same panel.  I also don't know what happens when projects have drafts and species lists since I have done any testing of that.  One thing that does look weird to me in the no background state is where the location constraint ends up.  I didn't really think about that case when I was last working on it.  Thanks for checking this out!

mo-nathan:
The CodeClimate violations make me wonder if we should move the location and date calculation stuff out to separate POROs.  The entire concept of a "constraint" could be it's own thing.

JoeCohen:
I'll look at where the location constraint ends up when there's no
background.
I will probably tuck the buttons into the panel.
I haven't manually tested projects with drafts or species lists, but they
would be listed directly under "*Summary* ..." (and above "*Created: *...").
See app/views/projects/show.html.erb.

JoeCohen:
That's a good idea. Sorry I was too lazy to deal with it. I'll see if I can
fix those violations as you suggest.

JoeCohen:
@mo-nathan: 
1. I moved the Observations, Names, and Members buttons.
2. Is this what you meant by "where the location constraint ends up" "in the no background state"?
Screenshot 2023-10-15 at 2 50 16 PM

If so, I can't figure out a better place (at least without creating more spaghetti in the templates).
Any ideas?
3. I moved some code out of ProjectsController to PORO's, and tweaked ProjectController to reduce the number of lines.
4. I moved some code out of Project to Validators. But Project is still > 250 lines. 
Probably more important, the moved code seems awkward. 
I'd appreciate it if you could provide a better pattern.

mo-nathan:
Yes, that's what I meant.  The case where you just had "USA" seemed a bit weird like it was just floating out in space.  If it were lined up with the right side of the Project title and there was less space around it I think it would be better.

JoeCohen:
@mo-nathan:
Potential ways to make short Project Locations look less weird. https://github.com/MushroomObserver/mushroom-observer/pull/1658#issuecomment-1758024727
- Add "Location:" before the Location link
- Move Location and Dates into the summary (and add "Location:" and "Dates:" respectively

Right now the Location is part of the banner, and I don't know how to conditionally right-align it with the title, or even how to calculation the the condition, which depends on the displayed length of the Location name and Project title.

My preference is to leave it as-is and tolerate the "weird" look. (And IMO it looks OK.)

mo-nathan:
Yeah, let's not sweat the constraint placement.

mo-nathan:
Sorry I have been more attentive to this PR.  Thanks for all the work here!  I looked at it a tiny bit this morning and will look more at it this evening.  I did spot a couple issues in my brief looking at it.  Specifically, if you have a project that doesn't have a date range (from before the migration), and then you go to edit it, it doesn't have "Any" selected for the dates and instead thinks I want it to have a range that is just today.  I think that will cause confusion if someone just changes something else like the title and doesn't notice that there is now a date range added.  I would also move the Date Range section below the title and the summary and probably the Locality on the edit project page so it reflects the order stuff shows on the Project Show page.

mo-nathan:
Looks like selecting "Any" for the date range doesn't get saved.

I also think we might want to allow a larger range of years.  I was trying to get a date range that works for my example project, "North Falmouth Fungi", and it has observations dating back to 2010 and the most it would let me go back at a time was 5 years.  Admittedly once "Any" is working correctly, this may not matter.  Another interesting thought might be to be able to automatically select a date range that allows all existing observations in the project to not violate the date constraints.    However, all of this is icing on the cake and probably not worth putting a lot of time into.

JoeCohen:
Thanks for looking at this again.
I've been out of town and dont know if I'll much of a chance to look at
this again until Fri.

JoeCohen:
Starting to work on making this pass.
It might take a while; this branch & main had diverged further than I expected.

JoeCohen:
@mo-nathan 
I think this branch is ready to merge.
(I'd rather deal with potential improvements -- e.g., listing out-of-compliance Observations -- in another branch.

app/models/project.rb
mo-nathan: The changes seem right, but there's a bug in the original implementation.  It doesn't handle cases where observations have a location, but no lat/long.  Essentially it is missing the logic reflected in line 280 from location.rb (last line of found_here?).

app/models/project/date.rb
```diff
@@ -0,0 +1,32 @@
...
+    !future? && !past?
+  end
+
+  # convenience methods for date range display
+  def date_range(format = "%Y-%m-%d")
+    "#{start_date_str(format)} - #{end_date_str(format)}"
+  end
+
+  def start_date_str(format = "%Y-%m-%d")
+    start_date.nil? ? :INDEFINITE.t : start_date.strftime(format)
```
mo-nathan, line 14: Are indefinite dates still a thing?

app/views/projects/_form.html.erb
```diff
@@ -20,9 +17,43 @@
...
+    </div>
+    <div>
+      <p>
+        <%= radio_button(
+              "project", "dates_any", false, checked: !@project_dates_any
+            ) %>
+        <%= "#{:form_projects_range.t} (#{:form_projects_use_date_range.t})" %>
+      </p>
+      <p>
+        <%= radio_button(
```
mo-nathan, line 49: See above comments

test/controllers/projects_controller_test.rb
```diff
@@ -172,6 +219,70 @@ def test_create_project
     assert_equal([rolf], admin_group.users)
   end
 
+  def test_create_project_with_any_dates
+    title = "Project without start or end"
+    start_date = Time.zone.today
+    end_date = start_date
+
+    params = {
```
mo-nathan, line 227: Can't this use `build_params` so it's more DRY.

test/controllers/projects_controller_test.rb
```diff
@@ -172,6 +219,70 @@ def test_create_project
...
+    assert_nil(project.start_date)
+    assert_nil(project.end_date)
+  end
+
+  def test_create_project_end_before_start
+    title = "Backward in Time"
+    summary = "Ends before it starts"
+    start_date = Time.zone.today
+    end_date = start_date - 1.day
+    params = {
```
mo-nathan, line 253: Can't this use `build_params` so it's more DRY.

test/controllers/projects_controller_test.rb
```diff
@@ -197,31 +308,85 @@ def test_edit_project
...
   end
 
-  def test_edit_project_empty_name
-    edit_project_helper("", projects(:eol_project))
-  end
+  def test_update_project_end_before_start
+    proj = projects(:pinned_date_range_project)
+    start_date = Time.zone.today
+    end_date = Time.zone.yesterday
+    params = {
```
mo-nathan, line 364: Again seems very similar to `build_params`.  Might need to make `build_params` a bit smarter, but I think it would be worth it.

test/integration/capybara/observations_integration_test.rb
```diff
@@ -196,4 +196,149 @@ def test_links_to_observations_of
...
+    proj_checkbox = "project_id_#{proj.id}"
+    login(user)
+
+    # create an Observation with Project selected
+    visit(new_observation_path)
+    fill_in(:WHERE.l, with: locations(:unknown_location).name)
+    check(proj_checkbox)
+    first(:button, "Create").click
+
+    # Porve that Project is re-checked for the next Observation
```
mo-nathan, line 215: Typo.

test/models/project_test.rb
```diff
@@ -114,6 +107,57 @@ def test_destroy_orphans_log
     assert_nil(log.reload.target_id)
   end
 
+  def test_dates_current
+    assert(projects(:current_project).current?)
+    assert_not(projects(:past_project).current?)
+    assert_not(projects(:future_project).current?)
+  end
+
+  def test_date_strings
```
mo-nathan, line 116: What is this test testing?

mo-nathan (changes requested):

app/views/projects/_form.html.erb
```diff
@@ -20,9 +17,43 @@
...
+    <div>
+      <%= f.label(:end_date, "#{:form_projects_end_date.t}: ") %>
+      <%= f.date_select(:end_date,
+                        { class: "form-control mb-2",
+                          start_year: Date.today.year - 10,
+                          end_year: Date.today.year + 10 }) %>
+    </div>
+    <div>
+      <p>
+        <%= radio_button(
```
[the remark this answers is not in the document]
nimmolo: @JoeCohen I believe you can use my `radio_button_with_label` helper here.

Bootstrap wants special markup nesting to make the label clickable, and this helper handles it for you.

JoeCohen:
@mo-nathan, @nimmolo 
Thanks for all the new suggestions and changes.
I'll start looking at them now, but probably won't have a big chunk of time until Tues (PST)

JoeCohen:
**Bug**: one can manually create a Project that ends before it starts.
Simply:
- Add Project http://localhost:3000/projects/new
- Pick an End Date earlier than the start date
- Select Range
- Create

JoeCohen (commented):
Fix bug that allows end date before start date.

app/views/observations/form/_projects.html.erb
```diff
@@ -22,6 +22,17 @@
...
+ <div class="mt-3 row" id="observation_projects">
+   <%= fields_for(:project) do |f_p| %>
+      <% if @suspect_checked_projects.any? %>
+        <%= check_box_with_label(
+          form: f_p, field: :ignore_proj_conflicts,
+          label: :form_observations_projects_ignore_project_constraints.t,
+          class: "alert alert-warning"
+        ) %>
+      <% end %>
+</div>
```
mo-nathan, line 34: This indentation confused me.  I created an alternative in https://github.com/MushroomObserver/mushroom-observer/pull/1700 that I find less confusing.  Not 100% it is exactly equivalent, but it seems to work after some light testing.

test/integration/capybara/observations_integration_test.rb
mo-nathan: Add commit for this.

app/views/application/content/_title_and_tab_sets.html.erb
```diff
@@ -36,12 +36,18 @@ left_cols = add_right_tabs ? "col-xs-8 col-lg-7" : "col-xs-12"
...
+          <% end %>
       <% end %>
     <% else %>
       <%= tag.h1(yield(:title), class: "h3", id: "title") -%>
       <% if content_for?(:location) %>
           <%= tag.div(yield(:location), class: "image-title-text") %>
       <% end %>
+      <% if content_for?(:date_range) %>
+          <%= tag.div(yield(:date_range), class: "image-title-text") %>
+      <% end %>
```
JoeCohen, line 50: Duplicates ll. 39-41.
Move it after the conditional.

app/views/projects/_form.html.erb
```diff
@@ -20,9 +17,43 @@
...
+    </div>
+    <div>
+      <p>
+        <%= radio_button_with_label(
+              form: f, field: "dates_any", value: false, checked: !@project_dates_any,
+              label: " #{:form_projects_range.t} (#{:form_projects_use_date_range.t})"
+            ) %>
+      </p>
+      <p>
+        <%= radio_button_with_label(
```
mo-nathan, line 49: This is still not achieving what I was looking for where you can click on the text as well as the button itself.  I created something that I like better on `njw-wrestling-with-radio-buttons`.  In the process I realized that were both radio_with_label and radio_button_with_label in app/helpers/forms_helper.rb and the first actually does more of the right stuff in my view which I largely recreated in radio_button_with_label.  It does seem like we should only have one of these.  However, I still don't like how separated the radio buttons are in radio_with_label and I just ran out of time to push on it further.  I also haven't seen where else that method is called.

nimmolo: @mo-nathan `radio_with_label` and `radio_button_with_label` are intentionally different helpers for UI variations that Bootstrap offers out of the can. 

There is no need to wrestle... all Bootstrap radios with labels (i.e. correctly formatted following their markup examples) have clickable text. The helper you want is `radio_with_label`. (In the markup, the nesting of elements is essential: the wrapping CSS class must be "checkbox", and the input must be inside the label. That's what the `radio_with_label` helper does for you so you don't have to remember.)

The `radio_button_with_label` wraps the **label** in Bootstrap "button" markup (i.e. the CSS class "btn btn-default" — think "submit button") and hides the "radio" circle-checkbox "button".  This is used for the "either-or" array of sorting buttons on index pages.
Screen Shot 2023-12-02 at 2 28 10 PM

The indispensable, timesaving reference, woefully underconsulted on MO, is the Bootstrap docs themselves. These are the three different variations on a radio UI that they offered in version 3:
https://getbootstrap.com/docs/3.4/css/#checkboxes-and-radios
https://getbootstrap.com/docs/3.4/javascript/#buttons-checkbox-radio
https://getbootstrap.com/docs/3.4/components/#input-groups-checkboxes-radios

I think I wrote `radio_with_label` for completeness when i was writing the `checkbox_with_label`, so i'd have it on hand — when you're trying to DRY up hundreds of views you do things like that 😓, but i'm not surprised it isn't called anywhere.

JoeCohen:
Intermittent failure:
>FAIL API2Test#test_deleting_species_lists (169.18s)
        API2 request should have failed, params: {:method=>:delete, :action=>:species_list, :api_key=>"bogus_api_key", :id=>935908374}
        test/models/api2_test.rb:46:in `assert_api_fail'
        test/models/api2_test.rb:3420:in `test_deleting_species_lists'
Failed in CI. See https://github.com/MushroomObserver/mushroom-observer/actions/runs/7095084262/job/19311441216?pr=1658 and https://github.com/MushroomObserver/mushroom-observer/actions/runs/7095084564/job/19311441791?pr=1658
Initially failed locally.
Then I put a breakpoint at test/models/api2_test.rb:3420. After continuing from the breakpoint, it passed.
Then removing the breakpoint, it passes.

mo-nathan:
The problem is that this test is randomly selecting a species list that the users own and there are now a species list that can be deleted by the other user due to some of the fixtures I've introduced.  I got rid of it by locking down the species list being selected test/models/api2_test.rb.  See https://github.com/MushroomObserver/mushroom-observer/pull/1704/files.

app/views/application/content/_title_and_tab_sets.html.erb
JoeCohen: That wouldn't work.

mo-nathan (approved):

app/controllers/projects_controller/validators.rb
```diff
@@ -32,4 +32,12 @@ def valid_dates
     flash_error(:add_project_ends_before_start.t)
     false
   end
+
+  def check_constraint_violations
+    count = @project.count_violations
+    return unless count.positive? && @is_admin
```
JoeCohen, line 38: @mo-nathan: Note: this warning goes only to admins

JoeCohen:
@mo-nathan:
This is ready to merge. Do want to look at it again?

app/controllers/observations_controller/form_helpers.rb
mo-nathan: Initially makes sense to me since I think most people most of the time will want the same selections that they just made for the previous observation.  The reason for making it optional was to avoid changing behavior for closed projects.  However, I now think they should be the same so I think this should just be `proj.current?`.  However, I don't think this has to happen in this PR.

app/controllers/projects_controller/validators.rb
mo-nathan: That seems reasonable and makes me think we should check @is_admin before evaluating @project.count_violations.  So add an initial check of `return unless @is_admin`.

app/models/observation.rb
```diff
@@ -407,27 +407,54 @@ class Observation < AbstractModel
...
+        lambda { |**args|
+          box = Mappable::Box.new(
+            north: args[:n], south: args[:s], east: args[:e], west: args[:w]
+          )
+          return none unless box.valid?
+
+          # resize box by epsilon to create leeway for Float rounding
+          resized_box = box.expand(-0.00001)
+
+          if box.straddles_180_deg?
```
mo-nathan, line 441: Do the current where conditions below actually differ?  Specifically is there any difference between `or.where` and `or`?  Also I think the first case is wrong.  Shouldn't it be long < west AND long > east (since in this case west > east)?

mo-nathan: Tried creating a project called "Pacific Ocean" that includes the entire Pacific Ocean as an example and the ProjectsController#show method fails on line 444.  I think this is a blocker and this needs a test.

mo-nathan (changes requested):
Looks like projects that span the international date line currently fail.

app/controllers/projects_controller.rb
```diff
@@ -46,6 +42,7 @@ def show
   def new
     image_ivars
     @project = Project.new
+    @project_dates_any = false
```
mo-nathan, line 45: I think this should default to `true`.  Currently if you create a project and don't mess with the date settings you get a project that is only good for today.  I think a more natural default is the "Any" state.

app/models/observation.rb
JoeCohen: @mo-nathan Thanks for looking at this again.
I'll look at this and the bug you found (projects that span the international date line currently fail.)

app/controllers/projects_controller.rb
JoeCohen: Fixed: Commit [06035e3](https://github.com/MushroomObserver/mushroom-observer/pull/1658/commits/06035e309b771f9fa35996ef74e4044aa909adbe)

JoeCohen:
@mo-nathan:
I made the requested changes. I don't think it needs another review. 
But if you're so inclined, you can look at just commit [046203f](https://github.com/MushroomObserver/mushroom-observer/pull/1658/commits/046203f3a54d33a01827b3ecaea394576cac4459).
I will probably merge it around noon PDT.

mo-nathan (approved):

PR #1662, opened 2023-10-07
Actually show projects on Observation#show page and add a test for it

mo-nathan:
There was code that was supposed to display an observation's projects, but it didn't work.  This makes it visible.  I wasn't able to get the surrounding "<div>" to render so I got rid of it since it didn't seem to be providing any value.

Is there any good reason to only show projects only if `@user` is truthy?

### DIFF
--- a/app/views/observations/show/_observation_details.html.erb
+++ b/app/views/observations/show/_observation_details.html.erb
@@ -4,12 +4,12 @@
 
 <%= observation_details_when_where_who(obs: obs) %>
 
-<% if @user
-  obs.projects.each do |project| %>
-    <%= "#{:PROJECT.t}: " %>
-    <%= link_to_object(project) %>
-    <br/>
-  <% end
+<%= tag.div(class: "obs-projects", id: "observation_projects") do
+  concat([tag.span("#{:PROJECTS.t}:")), tag.br].safe_join)
+  obs.projects.each do |project|
+    concat(tag.div(link_to_object(project), class: "indent"))
+  end
+end %>
 end %>
 
 <%= tag.p(class: "obs-specimen", id: "observation_specimen_available") do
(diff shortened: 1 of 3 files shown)


app/views/observations/show/_observation_details.html.erb
```diff
@@ -4,12 +4,12 @@
...
-    obs.projects.each do |project|
-      tag.p("#{:PROJECT.t}: #{link_to_object(project)}")
-    end
-  end
+<% if @user
+  obs.projects.each do |project| %>
+    <%= "#{:PROJECT.t}: " %>
+    <%= link_to_object(project) %>
+    <br/>
+  <% end
```
JoeCohen, line 12: A big improvement because it actually works.
But I wish I could figure out the syntax for:
- including blocks wtihin `tag`s,
- including interpolated `link_to_object` in a string in a tag.
Maybe there isn't one.
@nimmo?

JoeCohen (approved):

app/views/observations/show/_observation_details.html.erb
nimmolo, line 12: Yep, I have no idea why we were showing projects only if `@user` — i tried to faithfully reproduce every little thing I didn't understand when re-writing these! Did not always succeed.

I'd definitely like to keep the surrounding div "#observation_projects", and now that I look at it, it's a bit weird to have the "PROJECT:" title repeat on every line when there are multiple projects.

@JoeCohen the way to go is either to `concat` each line that should accumulate (the only option in a loop) or `[array, of, tags].safe_join`

How about this
```suggestion
<%= tag.div(class: "obs-projects", id: "observation_projects") do
  [tag.span("#{:PROJECTS.t}:")), tag.br].safe_join
  obs.projects.each do |project|
    concat(link_to_object(project))
    concat(tag.br)
  end
end %>
```

nimmolo: @JoeCohen I'm not sure i've ever gotten a link interpolated in a string to concat right, but i have thrashed through hundreds of views and maybe there's a solution in there somewhere. But - see suggestion below.

nimmolo: Approved with this change

PR #1663, opened 2023-10-07
Remove `PLATFORMS - arm64-darwin-22`

nimmolo:
I think it's wise to remove this line immediately from `Gemfile.lock` so it doesn't sneak into any branches and disable production. 

It's on `main` now, so it appeared on one of my branches when I merged `main` into it. It doesn't seem very foolproof to have to keep checking that.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -280,7 +280,6 @@ GEM
 
 PLATFORMS
   arm64-darwin-21
-  arm64-darwin-22
   x86_64-darwin-21
   x86_64-linux
 


JoeCohen:
1. I think this is a wise short-term measure to prevent accidentally clobbering the production server.
But when I asked @mo-nathan about that said (in an email in which I forgot to cc mo-developers),
@mo-nathan said:
>I’d actually be interested in trying to reproduce the behavior, but I don’t fully understand what was done.  I’m having a hard time understanding why an additional line in the platforms section of the Gemfile.lock could cause the problem.  I’m wondering if it’s something subtler in the gem configuration or something that’s wrong in the deploy.sh script.  Unfortunately the only way I think we can reproduce it is by potentially breaking production.  My guess is that if we add that line back in, rerun bundle install and restart unicorn, it will just work.  However, I want to do that test sometime when I have some time to really bang on it which is not today (I’m heading to a wedding at the moment).

I still think it's best to remove it to prevent accidentally killing the web site. It can always be restored for test purposes.
2. Is there any magic to removing via `bundle-lock --remove-platform` vs simply deleting the line in `Gemfile.lock`. ()
3. Will it inadvertently reappear? For instance, what happens if you change `Gemfile`, then `bundle install`.

JoeCohen (commented):
See [earlier Comment](https://github.com/MushroomObserver/mushroom-observer/pull/1663#issuecomment-1751807163)

nimmolo:
I believe it can only inadvertently appear if Nathan runs `bundle install`, because he's the one running Ruby with the new architecture. It did not appear for me when I ran `bundle install`.

nimmolo:
Re: 

> 2. Is there any magic to removing via bundle-lock --remove-platform vs simply deleting the line in Gemfile.lock. ()

I don't know. I believe Jason just deleted the line manually?

JoeCohen:
Perhaps a more fail-safe approach would be also changing Gemfile to impose
platform constraints on nokogiri.
nokogiri and its dependency libv8-node are the only gems that have arm64
versions installed.)
ChatGPT claims that one can do
gem 'nokogiri', '~> 1.12.3', platforms: [:x86_64_linux] But maybe it's just
making stuff up.

nimmolo:
Interesting... although I feel like ideally, I wouldn't want to lock the nokogiri version. 

Let's wait if Nathan turns up something.

PR #1664, opened 2023-10-09
Switch better stub pattern to eliminate expectations

mo-nathan:
This replaces a testing pattern using ` Minitest::Mock.new` with one using real objects and just stubbing out the key method needed to access the desired lines for completing coverage.

### DIFF
--- a/test/controllers/projects_controller_test.rb
+++ b/test/controllers/projects_controller_test.rb
@@ -299,22 +299,13 @@ class ProjectsControllerTest < FunctionalTestCase
 
   def test_user_group_save_fail
     title = "Bad User Group"
-    user_group = Minitest::Mock.new
-    add_user_group_expectations(user_group, title)
-    add_user_group_expectations(user_group, "#{title}.admin")
-    UserGroup.stub(:new, user_group) do
-      post_requires_login(:create, build_params(title, title))
-      assert_nil(Project.find_by(title: title))
+    user_group = user_groups(:bolete_users)
+    user_group.stub(:save, false) do
+      UserGroup.stub(:new, user_group) do
+        post_requires_login(:create, build_params(title, title))
+        assert_nil(Project.find_by(title: title))
+      end
     end
-  end
-
-  def add_user_group_expectations(user_group, title)
-    user_group.expect(:save, false)
-    user_group.expect(:name=, title, [String])
-    user_group.expect(:users, [])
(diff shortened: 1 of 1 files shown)


test/controllers/projects_controller_test.rb
```diff
@@ -299,24 +299,15 @@ def test_changing_project_name
 
   def test_user_group_save_fail
```
JoeCohen, line 300: The mocha equivalent is:
```ruby
  def test_u
    title = "Bad User Group"
    UserGroup.any_instance.stubs(:save).returns(false)
    post_requires_login(:create, build_params(title, title))
    assert_nil(Project.find_by(title: title))
  end
```

JoeCohen (approved):
1. Very nice! 
I will use this pattern in multiple places in ProjectsControllerTest.

2. Can also be done with the [`mocha` `any_instance_stubs` method](https://mocha.jamesmead.org/Mocha/ClassMethods.html). See [Comment above](https://github.com/MushroomObserver/mushroom-observer/pull/1664/files#r1350532415).
The `mocha` pattern is easier to write and read. \
We use it a lot (16 x).
But [`mocha` is not thread-safe](https://mocha.jamesmead.org/#thread-safety).

4. Do we care about thread-safety in tests?
If yes, should we try (in another PR) to replace all use of `any_instance_stubs` with your pattern (and hopefully get rid of the gem)?

test/controllers/projects_controller_test.rb
mo-nathan: I'm not a fan of `any_instance` stubs.  You may be able to do the same thing I'm doing with something like:
```
user_group.stubs(:save),returns(true)
UserGroup.stubs(:new).returns(user_group)
```
But I haven't tried it.  I don't know if you then need to "unstub" UserGroup.new.  Personally, I like the nested context a bit more and since I know that just lasts for as long as that context is running.  I think that also gives it at least a fighting chance of being thread safe, but I don't know the ins and outs of that.  As far as I know I don't think we're currently using multi-threading for tests, but I could imagine getting to the point where it was worthwhile.

JoeCohen:
Thanks for the feedback.
I will:
- Use the nested pattern in PR #1658 (as noted above).
- Add a Pivotal Story to try to get rid of any_instance_stubs everywhere.
- Merge this PR (if you haven't already done that).

PR #1666, opened 2023-10-10
Switch system tests to Cuprite from Selenium

nimmolo:


JoeCohen:
Thanks Nimmo. 
it will be nice to stop seeing the CI false negatives.

PR #1667, opened 2023-10-10
Restrict SpeciesList "Remove"

JoeCohen:
Displays "Remove" Observation buttons only to the SpeciesList creator (and in admin mode).

I noticed this morning that the show SpeciesList page has "Remove" Observation buttons visible to all users. 
But they work only for the list owner. (Other users get "Permission denied" if they click the button.

This PR therefore displays the button only to the owner.

### DIFF
--- a/test/controllers/species_lists_controller_test.rb
+++ b/test/controllers/species_lists_controller_test.rb
@@ -274,20 +274,43 @@ class SpeciesListsControllerTest < FunctionalTestCase
     )
   end
 
-  def test_show_species_list
-    sl_id = species_lists(:first_species_list).id
+  def test_show_species_list_non_owner_logged_in
+    list = species_lists(:unknown_species_list)
+    observations = list.observations
+    assert(observations.any?,
+           "Need SpeciesList fixture that has >= 1 Observation")
+    assert_not_equal(rolf, list.user)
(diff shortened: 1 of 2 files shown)


nimmolo (approved):

test/controllers/species_lists_controller_test.rb
```diff
@@ -274,20 +274,43 @@ def test_index_for_project_that_does_not_exist
...
 
-    # Show non-empty list with owner logged in.
-    get(:show,
-        params: { id: species_lists(:unknown_species_list).id })
     assert_template(:show)
     assert_template("comments/_comments_for_object")
+    assert_select(
+      "form:match('action', ?)",
+      %r{/observations/\d+/species_lists/#{list.id}/remove},
+      { count: 2 },
```
mo-nathan, line 311: Seems like it would be better to use `observations.count` rather than the magic constant `2`.

mo-nathan (approved):
One minor comment, but overall it looks good to me.

test/controllers/species_lists_controller_test.rb
JoeCohen: Oops. That's what I meant.

PR #1668, opened 2023-10-14
Replace any_instance.stubs

JoeCohen:
- Replaces `any_instance.stubs` with: 
  - the nested stub pattern of PR #1664; 
  - stubbing one instance of a controller; 
  - another substitute.
- Removes the `mocha` gem (and unique dependencies).

Goals:
- Make these tests thread-safe.
(`any_instance` is a method of the mocha gem, which is not thread-safe. See https://mocha.jamesmead.org#thread-safety. Also note that [minitest mocks are not thread-safe](http://stackoverflow.com/questions/7211086/how-do-i-stub-things-in-minitest).)

Drawbacks:
- Sometimes it's hard for me to figure out a substitute for `any_instance` and/or `stubs`.
- Substitute stubbing is almost always wordier and less transparent than `any instance`.


app/extensions/browser_extensions.rb
```diff
@@ -3,9 +3,12 @@
 require("browser")
 
 module Browser
-  # Prevent `browser` gem from saying that mobile DuckDuckGo browsers are bots
   class Base
     def bot?
+      # Ensure that test requests aren't from bots, unless test uses a known bot
+      return (bot.why? == Browser::Bot::KnownBotsMatcher) if Rails.env.test?
```
JoeCohen, line 9: This lets us drop -- in three places -- stubbing this method.

test/controllers/glossary_terms_controller_test.rb
```diff
@@ -281,17 +281,25 @@ def test_create_invalid_name_with_image
   def test_create_image_save_failure
     login
     # Simulate image.save failure.
-    Image.any_instance.stubs(:save).returns(false)
-    post(:create, params: term_with_image_params)
-
+    image = images(:disconnected_coprinus_comatus_image)
```
JoeCohen, line 284: I'm stubbing a random image. 
Should I do something else.
NOTE: This is getting to the right place in the case, and the assertion passes.

test/controllers/lookups_controller_test.rb
```diff
@@ -126,16 +126,13 @@ def test_lookup_name_matching_only_glossary_term
   end
 
   def test_lookup_name_with_error
-    # Stub a method called by lookup_name in order to provoke an error.
-    LookupsController.any_instance.stubs(:fix_name_matches).
-      raises(RuntimeError)
     login
+    # Stub a method used inside `lookup_name` to provoke an error
```
JoeCohen, line 130: AFAIK, the point of this test was to cover a line in the code and have a reasonable assertion.
The new version does that, BUT the assertion is  looser than in the prior version.

test/controllers/observations_controller_test.rb
```diff
@@ -2850,36 +2852,34 @@ def test_image_upload_when_create_fails
...
+    end
 
-    # Prove that an image was created, but that it is unattached, is in the
+    # Prove that an image was created, but is unattached, is in the
     # @bad_images array, and has not been kept in the @good_images array
     # for attachment later.
-    img = Image.find_by(copyright_holder: "zuul")
+
+    # img = Image.find_by(copyright_holder: "zuul")
+    img = Image.order(updated_at: :asc).last
```
JoeCohen, line 2881: This was the only way I could figure out around the following problem.

Rails stubs `image` per lines 2870-2871, **except that it seems to ignore `copyright_holder: "zuul"`**.
In particular, if I put a breakpoint immediately after `app/controllers/observations_controller/form_helpers.rb` line 153
```ruby
          image = Image.new(args2.permit(permitted_image_args))
```
and examine `image`, I see
```ruby
(rdbg) l    # list command
   ...
   153|           image = Image.new(args2.permit(permitted_image_args))
=> 154|           debugger
   ...
(rdbg) image
#<Image:0x00007f90db78cfd0
 id: 1062212617,
 created_at: Sat, 14 Oct 2023 23:34:38.000000000 EDT -04:00,
 updated_at: Sat, 14 Oct 2023 23:34:38.000000000 EDT -04:00,
 content_type: nil,
 user_id: 241228755,
 when: Sat, 14 Oct 2023,
 notes: "stubbed in test",
 copyright_holder: "Rolf Singer",
 license_id: 982438718,
 num_views: 0,
 last_view: nil,
 width: nil,
 height: nil,
 vote_cache: nil,
 ok_for_export: true,
 original_name: "",
 transferred: false,
 gps_stripped: false,
 diagnostic: true>
```
I know this is the stubbed image because it includes the `notes` I created.
But why does it ignore the `copyright_holder` that I provided?

app/extensions/browser_extensions.rb
mo-nathan: I'd rather we kept the stubbing.

test/controllers/glossary_terms_controller_test.rb
mo-nathan: This seems fine to me.

app/extensions/browser_extensions.rb
JoeCohen: Thanks for looking through this PR.
We were stubbing this method with `.any_instance.stubs`.
I'm not sure I can find a thread-safe alternative, but I will try.

JoeCohen: @mo-nathan:
I still cannot think of a way to stub `bot?` in the three test case files without using `any_instance.stubs`.
Any ideas?

test/controllers/lookups_controller_test.rb
mo-nathan: It's too convoluted for me to sort out what's going on.  If the coverage is the same, then I'm good with it.

test/controllers/observations_controller_test.rb
mo-nathan: So this is a weird one and it took a while for me to hunt it down, but the problem has to do with `init_new_image_var` in app/controllers/observations_controller/form_helpers.rb.  It is calling `Image.new` after it has been stubbed out and then it puts some default values into some of the fields.  The expectation is that these will get overridden by the params in a later call to `Image.new` at line 153 of the same file, but instead it just returns the now modified image.  In essence `Image.new` is getting called multiple times in the process for different reasons.  One potential way through it would be to change `init_new_image_var` so it takes a set of parameters rather than updating then after the call to `Image.new`.

app/extensions/browser_extensions.rb
mo-nathan: I'm not sure exactly what you're looking at, but I looked at the tests in `test/models/browser_extensions_test.rb` and these seems like testing code in other people's libraries.  Why do we care what their library responds with in our testing environment?  I'd be totally cool with changing the test to just `assert(Browser.new("").bot?)` and reverting the code back to the way it was.

JoeCohen: @mo-nathan:
Thanks again for all the time you've spent looking at this and for your comments.
I agree that we should not test other people's libraries.
Here's what I'm looking at.
- main has 3 files with this non-thread-safe stubs of `bot?`
```ruby
    # Treat Rails html requests as coming from non-robots.
    # If it's a bot, controllers often do not serve the expected content.
    # The requester looks like a bot to the `browser` gem because the User Agent
    # in the request is blank. I don't see an easy way to change that. -JDC
    Browser::Bot.any_instance.stubs(:bot?).returns(false)
```
(The three files are: test/application_system_test_case.rb, test/capybara_integration_test_case.rb, test/integration_test_case.rb.)
- I'd like to get rid of those stubs (and remove `mocha`) for the reasons you gave [here](https://github.com/MushroomObserver/mushroom-observer/pull/1664#discussion_r1350611944).  That's the point of this PR.
- If I simply remove those stubs, then (of course) I get a ton of failures and errors
```sh
vagrant@ubuntu-focal:/vagrant/mushroom-observer$ rails test:all
...
  2318/2318: [==================================] 100% Time: 00:02:44, Time: 00:02:44
Finished in 164.91208s
2322 tests, 25912 assertions, 70 failures, 19 errors, 0 skips
```
- My solution was to extend the Browser gem. 
**QUESTION: Can you suggest another solution?**
- I thought I should also unit-test the extension (as opposed to the gem _per se_).
**QUESTION: Is this a bad practice?** (We do it a lot. Seems like a good practice to me. We want to know if updates to the gems break our extensions.)
- Maybe we can discuss this at the tech meeting next Sat. I'm really jammed up this week.

test/controllers/observations_controller_test.rb
mo-nathan: It looks like the above either wasn't understood or wasn't followed up on.  To be more concrete, I think we should try something like:
```
  def init_new_image_var(default_date)
    @new_image = Image.new(when: default_date, license: @user.license, copyright_holder: @user.legal_name)
  end
```
I haven't tested the above, but I think it's point towards something that should work. The idea is to get the stub of Image.new to return what we expect rather than changing what we return in the subsequent lines of code.

mo-nathan:
@JoeCohen Where does this stand?  I just looked at it and it looks like we're in agreement on how to handle `bot?`.  I'm not sure if you understood my previous suggestion regarding Image.new stubbing.  Hopefully my new comment on this topic will make it a bit clearer.

JoeCohen:
Thanks for the follow up and suggestions. i am travelling today & Sat. will
look Sun.

JoeCohen:
@mo-nathan : [Your suggestion](https://github.com/MushroomObserver/mushroom-observer/pull/1668#discussion_r1397865043) let me make progress.
It returns the correct image. So `(assert(img, "Failed to create image")` no longer fails.  (I added messages to the assertions.)
It is (as expected) unattached, so the next line also passes.
But it fails one line further  at:
```ruby
    assert_includes(@controller.instance_variable_get(:@bad_images), img,
                    "Failed to include image in @bad_images")
```
with
```ruby
 FAIL ObservationsControllerTest#test_image_upload_when_process_image_fails (168.69s)
        Failed to include image in @bad_images.
        Expected [] to include #<Image id: 1062212472, created_at: "2023-11-18 09:19:28.000000000 -0500", updated_at: "2023-11-18 09:19:28.000000000 -0500", content_type: nil, user_id: 241228755, when: "2023-11-18", notes: "stubbed in test", copyright_holder: "zuul", license_id: 1, num_views: 0, last_view: nil, width: nil, height: nil, vote_cache: nil, ok_for_export: true, original_name: "", transferred: false, gps_stripped: false, diagnostic: true>.
        test/controllers/observations_controller_test.rb:2884:in `test_image_upload_when_process_image_fails'
```
I will try to dig into this later this weekend.

JoeCohen:
@mo-nathan:
I think this branch should be merged although I couldn't find an acceptable
replacement for the last remaining (triplicate) use of `any_instance.stubs`.
[Your suggestion](https://github.com/MushroomObserver/mushroom-observer/pull/1668#discussion_r1397865043) is incorporated in the test, 
ObservationControllerTest#test_image_upload_when_process_image_fails
and I fixed a final bug in the test.

mo-nathan:
I always forget just how tricky this case is.  I created a solution I prefer to the remaining instances of `any_instance.stubs` on the branch `njw-bot-fix`.  It's not amazingly clean, but it seems a bit more consistent with how we do things elsewhere.

JoeCohen:
@mo-nathan 
Thank you! I hope this didn't spoil part of your trip.
I'm about to head to sleep, but I had a quick look and it makes good sense.
It's not only more consistent with how we do things elsewhere; it lets us get rid of the `mocha` gem.
I'll try to get every merged into main tomorrow (12 Nov) a.m. EST.


> I always forget just how tricky this case is. I created a solution I prefer to the remaining instances of `any_instance.stubs` on the branch `njw-bot-fix`. It's not amazingly clean, but it seems a bit more consistent with how we do things elsewhere.

PR #1672, opened 2023-10-19
Update symbol_extensions.rb

nimmolo:
Work on debugging why it's returning a capitalized localization for a lowercase tag.

Just checking if this breaks anything...


nimmolo:
Closing as stale. This dates to 2023 and the file it touches (`symbol_extensions.rb`) was since renamed to `app/extensions/symbol.rb`, so the branch no longer merges cleanly. The core change (fixing the `:captialize` typo) turned out to have zero functional effect -- `capitalize_result` is checked with a truthiness test, not compared against the symbol's exact spelling, so the misspelling never mattered. The file has also since had substantial thread-safety work done on it. If any of the other small tweaks here (the `args.present?` guard, `test_localize`'s extra param) are ever actually needed, they're trivial to redo fresh against current `symbol.rb`.

PR #1677, opened 2023-10-28
Matrix box carousel preparation without the carousels

nimmolo:
This PR pulls in general improvements from #1598 but does not implement carousels in matrix boxes. I wanted to merge these changes (including Jason's improvements) before it gets too out of sync.

This PR does make the carousel changes available to a future PR via comments, because I'm pretty sure we will want these carousels at some point. I suspect that the n+1 problems may even be cleared up already, but I have been unable to find a recent db to test on.

- Query changes for carousels shelved (e.g. for all images — stored in comments), but important Bullet query optimizations kept.
- Template refactors merged, so a future carousel can be swapped in with minimal template changes.

I'm not sure what it is about the refactor in this PR that surfaced the Bullet optimizations, but Bullet insists on most of these.

#### Question: ####
Does it make sense to make an `observation_matrix_box_image_includes` frozen hash constant in `ApplicationController`, so we don't have to keep fiddling with the queries in the 4 or 5 controllers where observation matrix boxes are displayed?

Yes, it does. I guess that was rhetorical 🐥.

## Manual Test ##
Please test obs index load times vs `main`. This PR adds table joins via eager loading. 
However, it seems unexpectedly **quicker** on my machine, by about 25%.

### Project-level image copyright settings require laybrinthine queries! ###
I again want to draw attention to the fact that Project copyright settings are forcing extra table joins that definitely cost something, and seem only marginally useful (but maybe users really care!). If anyone can think of an acceptable way for matrix boxes **and lightboxes** to skip loading copyright info, please propose. 

However, it's worth recalling: Our long index render times (and the longer times on the Carousels PR) are mostly due to rendering partials, and only ViewComponents are likely to make a dent in these. The actual Index db queries are a very small part of our total render time.

Rails 7 enables lazy-loading template partials like image captions, so that could be a way to keep the pricey caption info but only load it when the image loads in the lightbox.


JoeCohen:
Manual testing: It's fast on my machine, but:
* it doesn't feel faster than main, 
* the numbers look about the same, e.g.
main:
```sh
Started GET "/" for ::1 at 2023-10-28 19:42:17 -0700
...
Completed 200 OK in 1553ms (Views: 495.8ms | ActiveRecord: 419.1ms | Allocations: 1173835)
```
nimmo-matrix-box-improvements:
```sh
Started GET "/" for ::1 at 2023-10-28 19:49:35 -0700
...
Completed 200 OK in 1646ms (Views: 553.4ms | ActiveRecord: 437.9ms | Allocations: 1213037)
```
(I don't think nimmo-matrix-box-improvements is slower. The above is just one sample. They both run about 1555-1655ms.)

nimmolo:
Thanks for checking it out Joe!

PR #1678, opened 2023-10-28
Update MACOSX_NOTES

JoeCohen:
Updates notes on  installing a local native MO in macOSX.
- reformats file as an `.md` for easierreadingandcut-and-pasting
- adds some of @JoeCohen's notes


nimmolo:
@JoeCohen  
- these were @mo-nathan's notes originally, and I think he had planned them as a script rather than a readme. Both seem useful though!
- consider folding in some lines from #1676?  I used `rbenv` and it may help others

MACOSX_NOTES.md
```diff
@@ -0,0 +1,253 @@
...
+if [ ! -f /usr/local/bin/exifautotran ]; then
+    sudo cp script/exifautotran /usr/local/bin/exifautotran
+    sudo chmod 755 /usr/local/bin/exifautotran
+    echo Installed exifautotran script
+else
+    echo exifautotran exists
+fi
+```
+### Install mysql2
+@nimmolo needed to run the following to get the `mysql2`` gem to install.
```
nimmolo, line 215: Probably @mo-nathan ?

MACOSX_NOTES.md
```diff
@@ -0,0 +1,253 @@
...
+    echo Ran migrations on the mo_development database
+fi
+
+rails lang:update
+rm config/credentials.yml.enc
+EDITOR='echo "test_secret: magic" >> ' rails credentials:edit
+```
+
+Hopefully this is not necessary on a fresh clean system,
+but @nimmolo had to run
```
nimmolo, line 249: probably @mo-nathan where you have @nimmolo

MACOSX_NOTES.md
```diff
@@ -0,0 +1,253 @@
...
+### Make sure you have the current version of Ruby.
+```sh
+if ! [[ `ruby --version` =~ `cat .ruby-version` ]]; then
+    echo You need to install version `cat .ruby-version` of ruby
+fi
+```
+
+There are various tools for this (rvm, chruby, rbenv).
+In the past MO used rvm, but recently switched to rbenv.
+@nimmolo used chruby most recently
```
nimmolo, line 131: @mo-nathan rather than @nimmolo. I used `rbenv`

MACOSX_NOTES.md
```diff
@@ -0,0 +1,253 @@
+The following is the beginning of a native MacOSX setup script.
+It is based on the notes @nimmolo took while getting his local Apple M1
```
nimmolo, line 2: ...based on the notes @mo-nathan took

JoeCohen:
@nimmolo: Thanks for looking at this and for the corrections and pointers.
I didn't even notice PR 1676; I've been too absorbed trying for a working MO app (either the VM or native) and non-MO stuff.
I'll take a close look now and try to wrapped it into this PR.

mo-nathan:
Glad you got things working and thanks for recording what you did.  I'm a bit torn about whether this should be a .md file or a script.  We haven't used .mds for much and most of our readme associated files are more like technical logs.  My original goal with making it more script like was so if it was really completed, it would be more reproducible since at least in theory you could just run it.  I'm not sure what the best path forward is here.  Would love to get folks thoughts on this.

JoeCohen:
A script that worked for everyone would be ideal.
I did it as an.md because:

   - I want to install the native app on my laptop (now in progress);
   - It's much easier to grab the shell commands from a .md than from
   comments lines;
   - I don't know what a complete, working script would look like;
   - I figured that the .md was a slight improvement, and that it's easy to
   revert or transition to a script.

Also:

   - The explanations, trouble-shooting, and work-arounds can be very
   helpful, so they should somehow remain in the final version.
   - I'm going to add some more tweaks as soon as I get done installing it
   on my (Intel) MacbookPro

PR #1681, opened 2023-10-29
x86_64-darwin-22

JoeCohen:
Adds x86_64-darwin-22 platform to Gemfile.lock

Appears tohave been added by `bundle install` on
both my iMac and MacBookPro, both of which have
Intel processors and run macOS Ventura 13.6

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -288,6 +288,7 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
   x86_64-darwin-21
+  x86_64-darwin-22
   x86_64-linux
 
 DEPENDENCIES


nimmolo:
Joe - wasn't this the gem associated with crashing the server? I can't remember if it maybe was resolved that this gem was NOT the problem, but maybe check with @mo-nathan and @pellaea before deploying.

JoeCohen:
Excellent question.
My memory is that it was a different platform (arm64-darwin-22), that
@mo-nathan fixed the problem, and that Nathan determined it was a one-time,
inexplicable glitch.
But better safe than sorry.
I'll have a look.

mo-nathan:
The evidence I've seen points to a problem reading the Gemfile.lock file rather than the change to the PLATFORMS section of the file.  The site has been running for quite a while now with new entries in the PLATFORMS section.  The cause of the read error is not clear and as far as I know has not occurred since then.

JoeCohen:
Thanks Nathan.
*Question:*
Are you available now to deal with any issue if I deploy this evening?
if yes, I'm inclined to deploy now.
If no, I'm inclined to revert the commit to main and live with Gemfile.lock
being out of sync on my desktop and laptop until there's a safer time to
deploy.

@nimmolo:

   - As a reminder: `system(cmd)` stopped working after the arm64-darwin-22
   platform was added to Gemfile.lock on the production server. See here
   <https://groups.google.com/g/mo-developers/c/2zW-kbwE0MQ/m/2C6jYytXAwAJ>
   - Jason did an emergency quick fix that included removing that platform
   from Gemfile.lock on the production server.
   - @mo-nathan eventually added that platform via a normal
   deployment process, and the platform continues to be included in the
   production server Gemfile.lock.
   He eventually concluded "All I can think is that something happened on
   the server that somehow prevented the Gemfile.lock from being read by
   process_image and what Jason did restored the ability to read that file.
   It's not a completely satisfying answer and I'll certainly be watching for
   more bad behavior, but I think we're in a reasonable place now." See here
   <https://groups.google.com/g/mo-developers/c/2zW-kbwE0MQ/m/872Mb43LAwAJ>.

mo-nathan:
Yeah, I'm around (and awake) for the next hour or so.

JoeCohen:
Super. I'll deploy now.

JoeCohen:
Deployed.
I'll try creating at least one Observation.

JoeCohen:
Nathan:
Thanks for being available. Everything seems to be running fine.
 I created 3 Observations and don't see any problems.
(The Obss are duplicates, so I'll remove them now.)
Joe

mo-nathan:
👍 I saw them as well.  Looks good!

PR #1682, opened 2023-10-30
Fixes and additions

JoeCohen:
These are fixes and additions based on using the prior version to install MO on
- a MacBookPro 
- with an Intel processor
- running macOSVentura 13.6

### DIFF
--- a/MACOSX_NOTES.md
+++ b/MACOSX_NOTES.md
@@ -37,21 +37,21 @@ Install a bunch of useful stuff from `Homebrew`
 brew install git mysql exiftool libjpeg shared-mime-info openssl imagemagick findutils
 ```
 
-### Install bash
+### Install Bash
 
-If you haven't done so already, install a recent version of `bash` and set
+If you haven't done so already, install a recent version of [Bash](https://www.gnu.org/software/bash/) and set
 it as the default shell. You can find your installed version like this:
 
 ```sh
 bash --version
 ```
 
-Apple includes only `Bash` 3.2 from 2007 in all versions of Mac OS X even now,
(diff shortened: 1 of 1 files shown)


JoeCohen:
Things generally went smoother installing on my laptop, as I had previously installed on my desktop.
The only glitch was that it didn't recognize `rails` as a standalone. I always had to use `bin/rails`.
I didn't mention this in the notes because I suspect that I did something out of order or omitted a step.
I'll try again sometime this week to see if the problem reoccurs.

MACOSX_NOTES.md
```diff
@@ -150,27 +158,44 @@ Add rbenv to zsh/bash so that it loads every time you open a terminal
...
+cp db/macos/database.yml config
+```
+
 Then delete `checkpoint_stripped.gz` from the mushroom-observer directory
 
 ### Run the rest of the mo-dev script
 (See https://github.com/MushroomObserver/developer-startup/blob/main/mo-dev)
 (Both @nimmolo and @JoeCohen did this in pieces.)
-Open a new shell and run
+- Open a new shell
```
JoeCohen, line 197: @mo-nathan @nimmolo:
Why open a new shell here?

nimmolo: No idea

nimmolo: Worth noting that you also don't HAVE to delete checkpoint_stripped.gz anymore. 

It's in `.gitignore` and if you want to reload a db without downloading... there it is?

JoeCohen:
Merging. I don't want this PR to become too long-running. (It's pure documentation and can't affect the webserver. So no harm done if it has gaps or errors.)

PR #1686, opened 2023-11-06
Add a concept of trusting a project for access to GPS data

mo-nathan:
To test you'll need to have a user that is the admin of a project and you'll need another user that is a member of that project and trusts the project.  The member will need an observation whose GPS info is hidden, but is associated with the project.  There is now a button on the show project page that allows members to trust a project.  Once this is setup, the admin user should be able to see the GPS info of the observation with the hidden GPS info.  Other users who are not either the owner of the observation or an admin for this project should not be able to see the GPS info.

Specific places where the GPS info is show include the show observation page, the label output, and reports.  The code for the reports is particularly tricky since it is trying to minimize the number of queries that get run.


JoeCohen:
Suggestion:
Either:
- Make Revoke Trust retroactive, or
- Add a warning or other text indicating that it applies only to future Observations.
(I expected it to be retroactive.)

JoeCohen:
Lat/Lon is missing from a spreadsheet downloaded by the Project Admin after Trust Revoked, but is present in printed Labels.
Here's what I did:
- As me:
  - Created a project Administered by me (in Chrome).
  - Added user OMS to that project.
- Logged in as OMS in a different browser (Firefox). 
  - Trusted the Project.
  - Ceated an Observation with hidden coordinates.
  - Revoked Trust.
- As me, 
  - went to the Project, 
  - clicked Observations, 
  - Download Observations
  - Selected CSV spreadsheet, UTF-8, clicked Download
  Result: spreadsheet without Lat/Lon
  [observations_1oQg6.csv](https://github.com/MushroomObserver/mushroom-observer/files/13326809/observations_1oQg6.csv)
   - Clicked UTF-16, then UTF-8 (in order to un-grey the Print Labels button)
   - Clicked Print Labels
 Result: RTF label that includes Lat/Lon
   Screenshot 2023-11-11 at 4 15 39 PM

I also note that the label rounds lat and lon to 1 decimal place. That seems wrong. (I don't think it should be rounded.) But that's another story.

app/controllers/projects/members_controller.rb
```diff
@@ -110,25 +109,51 @@ def find_member(str)
...
-      when :change_member_status_make_member.l
-        member = :add
+        update_member_status(project, candidate)
+      end
+      return_to_caller(project, params[:target])
+    end
+
+    def update_trust_status(project, candidate)
+      if params[:commit] == :change_member_status_revoke_trust.l
+        flash_notice(:change_member_status_revoke_trust_flash.l)
```
JoeCohen, line 125: Should any of these be `.t` rather than `.l`? (Both here and below.)
(Else why go to the trouble of having a translation string?)

app/controllers/projects/members_controller.rb
```diff
@@ -110,25 +109,51 @@ def find_member(str)
...
+        set_trust(project, candidate, false)
+      elsif params[:commit] == :change_member_status_trust.l
+        flash_notice(:change_member_status_trust_flash.l)
+        set_trust(project, candidate, true)
+      else
+        return false
       end
+      true
```
JoeCohen, line 133: This may be idiosyncratic, but I'd find it easier to read if the return values were included inside each condition. (But that adds a line and a wee bit of repetition.) YMMV
Suggestion:
```ruby
        set_trust(project, candidate, false)
        true
      elsif params[:commit] == :change_member_status_trust.l
        flash_notice(:change_member_status_trust_flash.l)
        set_trust(project, candidate, true)
        true
      else
        false
      end
```

app/views/projects/show.html.erb
```diff
@@ -53,12 +52,29 @@ add_project_banner(@project)
...
                                              candidate: @user.id,
                                              target: :project_index)) %>
-<% elsif @project.can_leave?(@user) %>
-  <%= put_button(name: :show_project_leave.t,
-                 class: "btn btn-default btn-sm",
-                 path: project_member_path(project_id: @project.id,
-                                           candidate: @user.id,
-                                           target: :project_index)) %>
+<% else %>
+  <% if @project.trusted_by?(@user) %>
```
JoeCohen, line 56: Question: Would it make sense to move all the conditionals into helpers or otherwise get them out of the template?

test/controllers/collection_numbers_controller_test.rb
```diff
@@ -49,7 +49,7 @@ def test_index_observation_id_with_one_collection_number
 
     assert_no_flash
     assert_displayed_title(
-      "Collection Numbers attached to #{obs.unique_format_name.t}".
+      "Collection Numbers for #{obs.unique_format_name.t}".
```
JoeCohen, line 52: Question: Is the following a bad practice?
I've been trying (at least when I have the energy) to use translation symbols as much as possible in test expectations, in order to not break the test if I revise the symbol definition.
(For instance, here I would try to use `query_title_for_observation`.)

JoeCohen (changes requested):
Very nice. I like this functionality.
But see my comment [here](https://github.com/MushroomObserver/mushroom-observer/pull/1686#issuecomment-1806946193)
You can ignore my comments about the code, as I didn't spend much time on it.

Question(s):
Is there
- Any functionality that I should try to manually exercise more?
- Any code that I should try to scrutinize?

app/controllers/projects/members_controller.rb
nimmolo: My understanding of this is fuzzy, but it's that `.l` "localize" is generally to be preferred unless the string is itself calling other embedded translation strings, and therefore needs the `.t`. I would appreciate any clarification!

mo-nathan:
@JoeCohen I don't know what you mean by making "Revoke Trust" retroactive.  It's boolean.  Either it is on or it's not.  It defaults to false, but gets explicitly set to true for the owner of each project.  However, now that I think about it, I don't think the owner should trust the project for existing projects since I believe the current user automatically gets access to their own lat/long info.

mo-nathan:
The rounding behavior pre-dated this PR.  We could change it, but that makes me nervous since we're then changing behavior that was created a long time ago.

app/controllers/projects/members_controller.rb
mo-nathan: I used .l since the original code had .l.  As Nimmo points out .t is only needed in the case that there are embedded translation strings and in these case they are kind of being used as constants so .l seemed reasonable to me.  Don't have a strong opinion.

JoeCohen:
What I meant by "retroactive" was: does it apply to Observations that were
earlier added to the project.
In particular, I note that if a user:

   - Trusts a Project
   - adds an Obs with hidden coordinates to the Project
   - Revokes Trust

then the Project admin still sees the coords when viewing the Obs.

So it appeared to me that Revoke Trust is not "retroactive"

JoeCohen:
Thanks for the explanation. Makes sense to leave it as-is.

JoeCohen:
Thanks Nimmo and Nathan.
I didn't understand that about .l and .t.

mo-nathan:
@JoeCohen I was not about to reproduce the case where after revoking trust the admin can see the coordinates.  What you're describing doesn't make sense to me since the trust info is not stored on a per observation basis.  Could it be that the admin user you are using is the owner of the observation?  If that's true, then yes, it will show the coordinates.

app/controllers/projects/members_controller.rb
mo-nathan: Sure.  Rubocop is cool with so that's fine.

JoeCohen:
The Obs owner and Project user are different.
I'll dig into it in a while to try to figure out what's going on.
_____
*Project owner & membership*

[image: Project members.png]

*Member OMS's view of Project. Note that it not Trusted (at the moment).*
[image: Member OMS view of Project.png]

*Project owner's view of Observation. Note that coords are visible.*
[image: Project Owner's view of Obs.png]

test/controllers/collection_numbers_controller_test.rb
mo-nathan: It is better to use translation symbols.  In this case I didn't fix it since that's the way the old code was.  It also wasn't clear that that was the correct symbol to use.  However, after more hunting I eventually found that it happens here: app/classes/query/modules/titles.rb and that the computed value of the symbol is in fact query_title_for_observation.

app/views/projects/show.html.erb
mo-nathan: It would, but I consider that outside the scope of this PR.

JoeCohen:
Re Project admin seeing hidden coordinates after user revokes trust.
It looks like `Observation#can_edit?` is returning the wrong thing, or at least the opposite of what I expect. 
(Ditto for `Observation#project_admin?`)
See below. I will keep digging.
_____
With the project admin (me, User 4468) logged in, 
and a breakpoint in `Observation#reveal_location?`
then when viewing the Obs in question, I see
```ruby
   701|   def reveal_location?
=> 702|     debugger
   703|     !gps_hidden || can_edit? || project_admin?
   704|   end
=>#0    Observation#reveal_location? at ~/mushroom-observer/app/models/observation.rb:702
  #1    block in observation_details_where_gps at ~/mushroom-observer/app/helpers/observations_helper.rb:207
  # and 169 frames (use `bt' command for all frames)
(ruby) User.current
#<User 4468: "Joseph D. Cohen (Joe Cohen)">
(rdbg) self
#<Observation:0x00000001159fa8b0
 id: 516708,
 ...
 user_id: 5485,
...
 lat: 0.458963e2,
 long: -0.918735e2,
 ...
 log_updated_at: Sat, 11 Nov 2023 15:38:27.000000000 PST -08:00>
(ruby) !gps_hidden
false
(ruby) can_edit?
true
(ruby) project_admin?
...
false
```

JoeCohen:
Some Progress. 
I think:
1. You can replicate the bug by making the Project non-open-membership.
2. The underlying problem:
- `ObservationsHelper#observation_details_where_gps` causes gps to be displayed `if obs.reveal_location?`
- `Observation#reveal_location?` returns `true` if (among other things) \<observation>.`can_edit?` is true:
```ruby
  def reveal_location?
    !gps_hidden || can_edit? || project_admin?
  end
```
- `Observation#can_edit?` calls `Project#can_edit?`, rather than seeing if the Obs itself is editable by the current user.
```ruby
  def can_edit?(user = User.current)
    Project.can_edit?(self, user)
  end
```
That seems strange to me. And I don't think it's what's needed in this case; `ObservationHelper` just wants to know if the current user can edit the Obs.

3. I don't know what the fix is.  Would something like this work:
- Delete the current definition of `Observation#can_edit?`.  (So \<observation>.`can_edit?` will call the superclass.)
- Change the calls to `Observation#can_edit?`-- except `reveal_location?` -- to directly call `Project#can_edit?`

mo-nathan:
That is correct behavior as far as I know.  Specifically, before there was the concept of open membership projects, all members and admins were considered to have edit permissions for anything that was part of that project.  The concept being that they had to be invited to be part of the project.  When open_membership is true, the ability for members and admins to edit is turned off.

mo-nathan:
Non-open membership projects essentially implicitly trust the project.  We could change that I suppose.  At minimum it seems like maybe we shouldn't allow such users to even mess with the trust level since it's kind of meaningless.

mo-nathan:
This might do a more intuitive thing:
```
% git diff
diff --git a/app/models/project.rb b/app/models/project.rb
index 28e145c44..86ce27944 100644
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -159,6 +159,7 @@ class Project < AbstractModel
     group_ids = user.user_group_ids
     obj.projects.each do |project|
       next if project.open_membership
+      next unless project.trusted_by?(obj.user)
       return true if group_ids.member?(project.user_group_id) ||
                      group_ids.member?(project.admin_group_id)
     end
diff --git a/db/migrate/20231019202655_update_project_users.rb b/db/migrate/20231019202655_update_project_users.rb
index e8e97677e..752a3bc68 100644
--- a/db/migrate/20231019202655_update_project_users.rb
+++ b/db/migrate/20231019202655_update_project_users.rb
@@ -4,8 +4,8 @@ class UpdateProjectUsers < ActiveRecord::Migration[6.1]
   def up
     Project.find_each do |project|
       project.user_group.users.each do |user|
-        ProjectMember.create(project:, user:,
-                             trusted: project.user_id == user.id)
+        trusted = (project.user_id == user.id) || not project.open_membership
+        ProjectMember.create(project:, user:, trusted:)
       end
     end
   end
```

JoeCohen:
> Non-open membership projects essentially implicitly trust the project. We could change that I suppose. At minimum it seems like maybe we shouldn't allow such users to even mess with the trust level since it's kind of meaningless.

Thanks for the explanation. I agree about not letting users mess with the trust level.

mo-nathan:
My proposed change would make both GPS visibility and edit privileges for admins and members non-open projects depend on the owner trusting the project.  The migration change would default all members of non-open projects to trusting that project.  Should probably make it so new created ProjectMembers trust the Project if it's not an open project.

mo-nathan:
Oof, that broke 18 tests.  My brain is fried.  I'll look at this more tomorrow.

mo-nathan:
So I dug into some of the test failures.  All the ones I've looked at so far are due to objects that are owned by the test user "mary" for tests being run as the user "dick".  It turns out that some of mary's objects are associated with the "bolete_project" (the "detailed_unknown_obs", "minimal_unknown_obs", "unknown_species_list", "in_situ_image", and "turned_over_image") and dick is the only member of the "bolete_users" and "bolete_admins" user groups.  The fact that mary is not a member and hence cannot trust the bolete_project means that these objects have switched from editable to non-editable.  I think simplest thing to do is add mary to bolete_users and have her trust the project.  That reduces the errors from 18 to 13.  I'll have to dig into those.

mo-nathan:
> Lat/Lon is missing from a spreadsheet downloaded by the Project Admin after Trust Revoked, but is present in printed Labels. Here's what I did:
> 
> * As me:
>   
>   * Created a project Administered by me (in Chrome).
>   * Added user OMS to that project.
> * Logged in as OMS in a different browser (Firefox).
>   
>   * Trusted the Project.
>   * Ceated an Observation with hidden coordinates.
>   * Revoked Trust.
> * As me,
>   
>   * went to the Project,
>   * clicked Observations,
>   * Download Observations
>   * Selected CSV spreadsheet, UTF-8, clicked Download
>     Result: spreadsheet without Lat/Lon
>     [observations_1oQg6.csv](https://github.com/MushroomObserver/mushroom-observer/files/13326809/observations_1oQg6.csv)
>   * Clicked UTF-16, then UTF-8 (in order to un-grey the Print Labels button)
>   * Clicked Print Labels
>     Result: RTF label that includes Lat/Lon
>   
>    Screenshot 2023-11-11 at 4 15 39 PM
> 
> I also note that the label rounds lat and lon to 1 decimal place. That seems wrong. (I don't think it should be rounded.) But that's another story.

I don't understand what you want here.  I am simply following what the code currently does.  For reports gps_hidden removes the lat/long.  For labels it rounds to .1.  Change either of these should be handled in a different PR.

mo-nathan:
As I said in the latest commit comment:
    What an adventure!  This is entirely to enable the one line change to
    app/models/project.rb.  This broken many tests (18?) and led me to
    discover that the way the bolete_project was configured was weird
    (user mary had observations but was not a member and hence didn't
    trust the project).  I ended up fixing the bolete_project except that
    it was also being used to test some special cases for a project that
    has a single user.  This led me to create the lone_wolf user and
    lone_wolf_project.  This in turn led me to find what I consider to be
    a bad test in: test/integration/capybara/lurker_integration_test.rb
    since it is dependent on the first observation in the rss_log to not
    have images.  I then found a bug in
    app/classes/query/modules/active_record.rb via
    test/models/query_test.rb where an error gets thrown if the new
    observation, :wolf_fart (English translation of Lycoperdon), has two
    images.  I changed the fixtures to work around the lurker test issue
    and the active_record bug.  The active_record should get fixed,
    but I wasn't able to figure out a satisfactory solution.

I left comments in the code to reproduce the two issues that I worked around by finessing the fixtures.  I really don't like how incredibly complex some of our tests are and how brittle the system is to changes in the fixtures.  It seems to me that you should always be able to add new stuff to the fixtures and not break existing tests.

mo-nathan:
@JoeCohen Is there anything you're still waiting for on this PR?  More manual testing would be great.  I've been buried in the test/fixture issue and haven't done much manual testing.  I think the latest changes will work better for closed projects.

JoeCohen (approved):

JoeCohen:
@mc-nathan There's nothing more I'm waiting on. (A lot of my problem was that i didn't understand the desired functionality.)
I will try to exercise the branch more. 
But I think we should get the pending Project-related PRs merged. (#1658, #1686, #1691)
(Because of the potential for conflict, I'm uncomfortable with having three of them pending. But maybe that's just me. I wanted to add )
I agree about the brittleness of the fixtures and tests. What's the solution? Fewer fixtures + more factories?


> @JoeCohen Is there anything you're still waiting for on this PR? More manual testing would be great. I've been buried in the test/fixture issue and haven't done much manual testing. I think the latest changes will work better for closed projects.

mo-nathan:
Ok, I'd like to do a bit of manual sanity checking before merging, but I'm going to head that way.  Personally, I'm not worried about having multiple project related PRs open, but it would of course be good to get stuff merged.

Regarding fixing the tests, I believe that all of the tests we want to do could be written in a way that allows for additions to fixtures.  In my view our tests are often too complicated.  We are definitely not following best practices for how to write tests.  My recommendation is that we figure out which tests are fragile to fixture changes and work to fix those tests.  I think we could find them by just adding a disconnected new fixture to each fixture file.

We could consider switching to factories.  In my view it is a better way to do tests at scale, but it's not without it's problems.  Given how committed we currently are to fixtures, I don't think it would be worth the effort to switch.

PR #1689, opened 2023-11-10
Reorganize classes pertaining to Mappable objects

nimmolo:
This is a refactor pulled from the Stimulus/Turbo PR, for review by @pellaea, if he has time. 

The point of the refactor is to move all of the derived values of a `MapSet` from instance methods to attributes, so that these derived values will be accessible in Javascript if I send the `MapSet` as JSON to be plotted on a Google map by JS. Otherwise, the JS must be assembled in Ruby (as Jason coded it originally), and that is awkward for Turbo page reloads.

I've kept the attribute names the same as the old methods to ensure minimal code breakage, e.g. `mapset.north_west` is the same. Two of these had to change names: `is_box?` becomes `is_box` as an attribute, it can't have a `?`. Ditto for `is_point?`.

Doing this ahead of the Stimulus/Turbo PR for the sake of troubleshooting.


nimmolo:
Nothing fancy.  Other than name changes, all of the changes are in `classes/mappable/map_set.rb`. 

I'm just moving `MapSet` instance methods to `attr_accessor`s and adding a new method `update_derived_attributes` to update the attributes whenever the box `extents` are changed. This can be during init, or whenever one of the object-adding methods are called on the instance.

PR #1691, opened 2023-11-11
Fix three issue related to editing observations and projects

mo-nathan:
1) Doesn't throw up the warning stuff about constraint violations if the observation is already part of that project.
2) Doesn't automatically try to add the observation to projects if you're editing the observation vs. creating a new one.
3) Don't allow admin editing of objects of users who do not trust a project.  Note that editing of objects is only allowed for closed projects.


JoeCohen (commented):
Thanks for doing this.
I'm fine with it as-is.
There is an edge case that you've probably thought of. I don't know if you want to deal with it:
User edits an Observation that meets the Project constraints so that it no longer meets the project constraints. When I did that, it flashed Success without a warning about failing the constraints.

mo-nathan:
Not going to sweat that edge case.  Not this has to merge after njw-project-members

JoeCohen:
Suggestions (because I didn't understand all implications of Trust).
- When a User clicks Trust the flash should indicate that Admins (or all Members, as the case may be) can edit the User's Observations which belong to the Project. (The flash now just says that admins can view hidden coords: "You now trust admins of this project with hidden GPS data for associated observations.")
- Similar for clicking Revoke Trust
- And in addition to the ephemeral flash, explain the effect of the current trust status on the ShowProject. 
    Something like:
    "You trust Project Admins to edit, and see hidden coordinates of, your Observations that are part of this Project."
    "You trust all Project Members can edit, and see hidden coordinates of, your Observations that are part of this Project."
    "You do not trust Project Admins cannot edit, or see hidden coordinates of, your Observations that are part of this Project."
    "You do not trust Project Admins or Members to edit, or see hidden coordinates of, your Observations that are part of this Project."

mo-nathan:
So as things are currently written if a project is closed objects associated with a project are editable by other members of the project if the owner of that object trusts the project.  I agree that adding comments to that effect makes sense.  However, in thinking more about it (and testing), I found another thing that needs to be fixed.  The main goal of not allowing editing of objects in open projects is to avoid the case where someone just joins a project and can suddenly edit stuff with no review by a project admin or the owner of the object.  However, I've realized this means if an admin wants to edit an observation of someone who trusts the project, they can do so if they switch the project to closed, make the edit, and then switch the project back.  That seems kind of odd.  I think the editing ability needs more explicit control.  I need to think this through more.

JoeCohen:
Would this work:
- When a User joins/is added to a Proj, set trust regardless of Proj type. If it's an Open Proj, set trust to false.
- Now if the Admin changes the Proj type, s/he cannot edit other Users' objects that belong to the Proj because the User doesn't trust the Project.

Maybe I don't grasp something about the desired behavior. 
If the intent is to allow Open Proj Admins (but not other Proj members) to edit other users' Objects, 
then would this work:
- Multiple levels of trust: Untrusted, Admin (only), Trusted. 
- When User joins an Open Project, set Trust to Admin.

mo-nathan:
Once I get the coverage back up on this, I think we can merge it.  It doesn't solve the problem I've described, but it's better to have these fixing in than not.  After that I'll work on cleaning up the concept of trust for projects.

mo-nathan:
@JoeCohen I think this is finally ready to merge.

JoeCohen (approved):

PR #1705, opened 2023-12-10
Queued emails can now have to_user == null, meaning send to webmaster.

pellaea:
This removes a sanity check from `rake email:send` which caused it to silently throw away emails with null recipient.  This used to make sense... sorta... at least as a way to reduce junk error messages.  But now that *all* emails, including those to webmaster, are queued, it means that recipient can legitimately be null (that means the recipient is webmaster).  The obvious solution was simply to remove the sanity check.  Hopefully this doesn't create a whole mess of cronjob error messages!

@nimmolo and @JoeCohen -- The problem here, I believe, is that rake tasks are not unit tested.  The seemingly obvious solution is to move the guts of lib/tasks/email.rb into app/models/queued_email.rb so that it can be unit tested properly.  Then the only thing in lib/tasks/email.rb is `QueuedEmail.send_emails` and really doesn't matter that it's not tested.  Does that sound like a plan?

### DIFF
--- a/lib/tasks/email.rake
+++ b/lib/tasks/email.rake
@@ -20,45 +20,37 @@ namespace :email do
       # Has it been queued (and unchanged) for MO.email_queue_delay or more.
       if e.queued + MO.email_queue_delay.seconds < now
 
-        # Sent successfully.  (Delete it without sending if user isn't local!
-        # This shouldn't happen, but just in case, better safe...)
-        if e.to_user
-          result = nil
-          Rails.root.join("log/email-low-level.log").open("a") do |fh|
-            fh.puts("sending #{e.id.inspect}...")
-            result = e.send_email
-            fh.puts(
-              "sent #{e.id.inspect} = #{result ? result.class.name : "false"}"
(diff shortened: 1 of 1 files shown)


nimmolo:
That sounds like a plan to me.

mo-nathan:
Agreed.  Moving the contents of rake tasks in proper objects that can be unit tested is the right thing.  I also note that there are a bunch of rubocop violations that can be cleaned up quickly by just running `rubocop -A lib/tasks/email.rake`.

JoeCohen:
1. @pellaea. A big thanks for fixing this.
2. I think it is worth getting the fix out now, as the bug makes some
emails disappear.
3. I agree with moving rake tasks to testable objects.

JoeCohen:
It broke some after Tue, Aug 15, 6:35 PM (PDT). (That was the timestamp on
the last merge request we actually received.)

pellaea:
Pushing out quick fix, will address rubocop and tests etc. later.

PR #1716, opened 2023-12-14
Rails 7.1.2

nimmolo:
Like Rails 7.0 (and unlike Turbo/Stimulus), this is not a huge update, and I'm not finding reports of downsides to it in the wild. It's more like a "Rails 6.3.2" - it seems to me the version numbers have as much to do with signaling that Rails is being actively worked on as with any breaking changes. 7.1 does have some nice features, but mostly i want to be very ready for Turbo 8.

I believe this is ready for prime time, it doesn't require any big changes to our codebase.

### TODO: ###
- [x] Fix deprecation warnings
- [x] Make sure configs on `main` are migrated to recommended defaults for 7.0 where applicable in `config/initializers/new_framework_defaults_7_0.rb`
- [x] Implement as many default configs for 7.1 as possible without forcing big refactors

One config we cannot implement yet is 

    Do not treat an `ActionController::Parameters` instance as equal to an equivalent `Hash` by default.

Implementing this requires reworking some tests (or better, adding a test helper) that will convert the hash of params we want to compare into an instance of `ActionController::Parameters`.


config/environments/production.rb
```diff
@@ -79,43 +86,53 @@
...
+  # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for apache
   config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # For nginx
   # If you have no front-end server that supports something like X-Sendfile,
   # just comment this out and Rails will serve the files
 
   # Force all access to the app over SSL, use Strict-Transport-Security,
   # and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
JoeCohen, line 106: ??? Seems like a big change.

JoeCohen: @pellaea: What do you think?

nimmolo: @JoeCohen I don't believe this is much of a change. 

If I'm not mistaken we already default to SSL for all access. If you access the site via http://mushroomobserver.org, it responds with https://mushroomobserver.org as the address.

Also, I think [SSL is necessary to get HTTP/2](https://www.ssl.com/article/an-introduction-to-http2/). As of the Turbo PR, the app will be very slow without HTTP/2.

> Although the standard itself does not require the use of encryption, all major browser implementations (i.e. Firefox [09], Chrome, Safari, Opera, IE, Edge) have decided that they will only support HTTP/2 over TLS.
> 
> In fact, browsers distinguish between clear-text HTTP/2 and HTTP/2 over encrypted TLS as two different protocols. Encrypted HTTP/2 is called h2 and clear-text h2c. As of this writing, none of the major browsers support h2c , which means that TLS encryption is mandatory for a web site to take advantage of HTTP/2’s other advantages.

nimmolo: @pellaea @mo-nathan, do you guys foresee problems forcing SSL, from the API or elsewhere? I feel like we _should_ be forcing SSL, but I don't know all our clients like you guys do.

I'd like to get moving on 7.1, either by trying out `config.force_ssl` in a small PR off `main`, or just merging this PR.

nimmolo: We are now running the `config.force_ssl` since yesterday, without apparent problems. I may merge this tonight if i don't hear from anyone.

PR #1723, opened 2023-12-14
Fix needs review functionality for visual groups

mo-nathan:
You should now be able to see images that can be added to the include or exclude list for a visual group if you go to pages like:

http://localhost:3000/visual_groups/14/edit?status=needs_review


nimmolo:
Looks good. For this visual group on my snapshot, there are "needs review" and "included" images, and no "excluded"... and when I click to "exclude" one, it does show up in "excluded".

nimmolo (approved):

PR #1732, opened 2023-12-19
Tweak layout to remove superfluous space between images and names.

pellaea:
See obs #136617 for an example that really shows the problem this PR proposes to fix.


pellaea:
Oh crap, but this re-orders the sections for mobile.  This is a no-go.

PR #1737, opened 2023-12-21
Observation `show` scopes - move to model, add `strict_loading`

nimmolo:
The `ObservationsController` has two methods defining scopes of eager-loading for a show Observation view: one for logged-out users ("light includes"), and one for logged-in users ("heavy includes").

This PR 
- refactors the two arrays as query scopes on the model
- experimentally adds the newer Rails method `strict_loading` to the scopes. This will raise warnings in tests if any association used in the action is not eager-loaded. It flagged several more associations we weren't loading yet.


nimmolo:
@pellaea @JoeCohen @mo-nathan - Check out this new Rails method `strict_loading`, that you can call on a query scope. It's said to be better than Bullet at alerting us to missing eager-loads, and it has in fact found some missing associations on show Obs. It could be useful in many places.

PR #1739, opened 2023-12-22
Refactor `ConsensusCalculator#initialize` to take observation (not namings) 

nimmolo:
`ConsensusCalculator` is currently initialized by eager-loading the associations of the passed-in `namings`.
It's called by `Observation#calc_consensus`, which has just reloaded the obs. 

If instead `calc_consensus` eager-loads the obs and its associations, the obs can be passed into the `ConsensusCalculator`, saving the second load.


nimmolo:
This doesn't work because `Observation.announce_consensus_change` is still using the stale instance.

On to plan B - the PORO.

PR #1742, opened 2023-12-23
Convert `#observation_namings` to a `div` (from a `turbo-frame`)

### LINKED ISSUE #1741: Clicking on a Proposed Name does not go to the linked Name
Screenshot 2023-12-22 at 11 06 37 AM

## Example 
(iMac, macOS 14.2.1 (23C71), Chrome Version 120.0.6099.109 (Official Build) (x86_64))
- Go to <https://mushroomobserver.org/539639>
- Hover over the proposed name (_Stereum complicatum_)
Result: Browser seems to correctly interpret _Stereum complicatum_ as a hyperlink to <https://mushroomobserver.org/names/3745>
Screenshot 2023-12-22 at 10 51 44 AM

- Click  _Stereum complicatum_ 
Result: 
  - It fails to go to the linked page
  - The Proposed Name section disappears and is replaced by Content Missing after waiting a while (at least 7 seconds for me) 
Screenshot 2023-12-22 at 11 04 53 AM

##PS
The Edit Proposed Name icon does not respond to left-click

##Note:
- Right-clicking _Stereum complication_ or the Edit Proposed Name icon,  then selecting "Open link in new tab" in the pop-up menu works correctly
- The counterpart linkto _Stereum complication_ in the "About this taxon" section works correctly.

nimmolo:
Fixes #1741 

Somehow, the Namings table being a `<turbo-frame>` made ordinary links within it not work, in this case the Name and User links for the naming. Modal links to edit or destroy the naming still worked.

Judging from the console errors, the plain `<a>` links seem to have been intercepted by Turbo.js, and it seems to have tried to replace the table with the returned content from the link (e.g., the show user page!). The result was the naming table was replaced by "Content missing", which is a Turbo error message. 

Turbo will not replace an element if the returned HTML doesn't have anything with a matching DOM ID — in this case, "observation_namings". 

I had it as a `<turbo-frame>` because i was playing around with lazy-loading the entire namings section. Shelved for now.

### DIFF
--- a/app/views/observations/namings/index.erb
+++ b/app/views/observations/namings/index.erb
@@ -1,5 +1,5 @@
 <%=
-tag.turbo_frame(id: "observation_namings") do
+tag.div(id: "observation_namings") do
   render(partial: "observations/namings/table",
          locals: { obs: @observation })
 end %>
--- a/app/views/observations/show/_namings.erb
+++ b/app/views/observations/show/_namings.erb
@@ -3,13 +3,7 @@
 %>
 
 <!--OBSERVATION_NAMINGS / TABLE REFRESHED BY TURBO -->
-<%= tag.turbo_frame(
-  id: "observation_namings",
-  # src: observation_namings_path(observation_id: obs.id),
-  # loading: "lazy"
-  # for lazy loading, don't put anything in the block.
-) do
-
+<%= tag.div(id: "observation_namings") do
   render(partial: "observations/namings/table",
          locals: { obs: obs, query: @query })
 end %>
--- a/test/system/observation_show_system_test.rb
+++ b/test/system/observation_show_system_test.rb
@@ -301,7 +301,7 @@ class ObservationShowSystemTest < ApplicationSystemTestCase
     assert_selector("#title", text: /#{nam.text_name}/)
     # sleep(3)
 
-    # check that there is a vote tally with this naming
+    # check that there is a vote "index" tally with this naming
     within("#observation_namings") do
       assert_link(href: "/observations/#{obs.id}/namings/#{nam.id}/votes")
       click_link(href: "/observations/#{obs.id}/namings/#{nam.id}/votes")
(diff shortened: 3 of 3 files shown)


JoeCohen (approved):
Thanks for the late-night fix! It will be much appreciated by users.
Looks good, works fine.
Seems so simple in hindsight.
I'm glad you know what your doing. It would have taken me forever.

JoeCohen:
PS: I will merge and deploy it.

JoeCohen:
Oh wait! How hard would it be to add a system test?

nimmolo:
Thanks for documenting it Joe, that made it a lot easier to diagnose. 

Ah! I will add a system test now.

(Incidentally, do you think those User and Name links for the naming should open in a separate tab?
That’s a question for another day, or maybe the meeting but I’m curious what you think.)

JoeCohen:
Jason once suggested that all links (or maybe it was all external links)
should open in new tabs.
I agree that's a separate PR.
I've gone back & forth in my mind about that, but lately I've been leaning
to separate tabs.
If the tabs are separate, should we also switch focus to the new tab? If
focus remains the same, it's easy to see that there's a new tab and
switch to it -- if you're on a computer.  But what about mobiles?

JoeCohen:
Hmm. I'm getting this failure locally:
```ruby
joe@Josephs-iMac mushroom-observer % rails test:system
Started with run options --seed 42748

[Screenshot Image]: /Users/joe/mushroom-observer/tmp/capybara/failures_test_identify_index_ui.png                                                                                        ] 44% Time: 00:00:28,  ETA: 00:00:37
 FAIL HelpIdentifySystemTest#test_identify_index_ui (38.82s)
        expected to find css "#modal_obs_247545329_naming" but there were no matches
        test/system/help_identify_system_test.rb:30:in `test_identify_index_ui'
  9/9: 
```

nimmolo:
That's an intermittent failure i've been getting too. 
It results from the modal naming form taking too long to load!
If you try *just* that test again, I bet it passes.

The syntax for a single system test is different unfortunately. You have to specify it by path (and line number).
`rails t test/system/help_identify_system_test.rb:6`

I've been trying to tweak the tests to give these more time to load, but they sometimes drag an unreasonably long time (thankfully less often). Waiting for 15 seconds for four of these forms obviously could increase the test suite time by an entire minute.

Hopefully some of this n+1 work will speed up the naming forms, on the other hand.

JoeCohen:
Thanks!
I ran it by itself. It passed -- some of the time.
BTW this syntax worked for me:
rails t test/system/help_identify_system_test.rb -n test_identify_index_ui

nimmolo:
@JoeCohen - Ok, in the system test, test-loading the name page is not bad. 

But the user page... whew what a doozy.

I've written the test, but the page takes ~30s to load. I think I'd prefer to leave the test commented out and implement it when we can get that page load time down. That is next on my list after the naming table (my current project).

Would that work?

JoeCohen:
Yes, that makes sense. More important to deploy the fix.
Let me know when it's ready to merge and I'll merge and deploy.

nimmolo:
Thanks and thanks for the alternate syntax, i didn't know that one. 

I've been keeping a list of _line numbers of problematic tests_ for weeks! hah

Should be mergeable as soon as CI is done.

PR #1743, opened 2023-12-23
Edit obs: Fix setting `thumb_image_id`

### LINKED ISSUE #1733: Cannot change Default Thumbnail when editing Observation
To replicate:
- Edit any Observation having multiple Images
- Click the radio button next to any Image for which the radio button is not already selected
Note: There is no "Set as Default Thumbnail" button like tthe one that appears when creating an Observation.
That's probably at least part of the problem.
- Save Edits
Result: The thumbnail is unchanged

nimmolo:
Adds system test. 
Fixes #1733

### DIFF
--- a/app/views/observations/form/_good_image.html.erb
+++ b/app/views/observations/form/_good_image.html.erb
@@ -4,7 +4,10 @@
 
   <!-- GOOD_IMAGE -->
   <div class="col-xs-1" style="max-width:30px">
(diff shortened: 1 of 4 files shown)


JoeCohen:
@nimmolo: Thanks for fixing this! I will deploy it.
Sometime in the future, the radio button probably should change to a "Set as Default Thumbnail" button, like the one that appears when you add an Image.

PR #1745, opened 2023-12-25
bundle update 20231225

JoeCohen:
- Updates gems
- Updates version of RuboCop used in CI
- Fixes many auto-correctable RuboCop offenses


JoeCohen:
I'm ignoring the remaining Codeclimate issues as follows:
- possible SQL injection's in [app/classes/eol_data.rb](https://codeclimate.com/github/MushroomObserver/mushroom-observer/app/classes/eol_data.rb?from_sha=84275d46&to_sha=1e0e84c1). Too complicated for me to fix as part of this PR. Probably should refactor the file to use AREL.
- Anonymous arguments forwarding. I cannot duplicate these Rubocop offenses locally (running Rubocop v 1.59.0), and therefore cannot autocorrect them (assuming that they're real offenses). There are too many to manually fix in this PR. (I will report this bug to Rubocop).

PR #1748, opened 2023-12-26
Move the naming table from partials to a helper

nimmolo:
LOL. I'm the one who broke this template up into several partials to make the HTML more comprehensible, and now i'm reuniting everything back in one long file, broken up as methods.

It is certainly questionable if I've made it any more comprehensible!

However, there are two reasons to do this:
- I need to wrangle these n+1's coming from instance methods being called from the template, and sending the helper one argument will help spot where they're happening.
- the goal now is to make it a ViewComponent, and i'm pretty sure that means it will render more quickly. ViewComponent code is structured more like the helpers.


nimmolo:
Bonus: Fixes a bug where the page flash was not updated when the observation associations were CRUDded.

nimmolo:
Bonus 2: Fixes a bug where flash messages for naming problems were not all being printed, and would be leftover to the next page load.

PR #1749, opened 2023-12-27
Prepare for ViewComponents

nimmolo:
- [x] add the gems for ViewComponents 
- [x] add gems for the seemingly-improved EvilMartians defaults
- [x] add base components to the /views/components folder
- [x] rubocop the changes


nimmolo:
Note: This PR reorganizes the views folder. I like the clearer organization, and I think i'd want to keep it even if we get rid of ViewComponents. But I'd like to hear from others.

Although i'm not quite ready to implement this, i'd like to merge it soon, because any work on the views folder in the meantime will get this PR out of sync. I will try to implement it in a few places as soon as I can clear up the eager-loading/N+1 issues.

app/helpers/application_helper.rb
```diff
@@ -165,4 +165,10 @@ def get_next_id(object)
 
     query.result_ids[idx + 1] || query.result_ids[idx - 1]
   end
+
+  # make View components easier to call
+  def component(name, *args, **kwargs, &block)
```
JoeCohen, line 170: Depending on how you and others feel about all-arguments forwarding, 
the following should work here. Your choice:
```suggestion
  def component(name, ...)
```

JoeCohen (commented):
I tried it briefly locally without problems.
It looks like it may take some getting used to the conventions.

nimmolo:
@JoeCohen Thanks for checking it out. 

Other than finding things in `/views` this PR doesn't change anything immediately. It occurs to me that reorganizing the views folder is kind of a separate scope and it's worth evaluating on its own, just to separate mailers from other views. I may do that PR separately, and I could keep this one on hold for when i'm actually ready to write a view component.

Converting re-usable partials and helpers to components does mean getting used to some different conventions for sure. I'm thinking to start with heavily used components where there would be an immediate benefit in performance: images, carousels, and matrix boxes. 

But those would all be separate PRs, one at a time.

PR #1752, opened 2023-12-28
Remove N+1's from `owner_naming_line`

nimmolo:
Derives the owner's preferred naming without db reloads, by moving some Observation instance methods to a new PORO. The same class could be used to reduce N+1's on the naming table in a future PR.

Shaves nearly 50% off page load time!
```
(Views: 203.8ms | ActiveRecord: 2197.0ms | Allocations: 206902) # before
(Views: 99.6ms | ActiveRecord: 1107.7ms | Allocations: 111354)  # after
```
___

I discovered while auditing show obs for N+1's that the page subtitle, where we show the owner's preferred naming, does a bunch of N+1s to get the preference via naming votes. But the votes at this point have already been eager-loaded, and the subtitle's use of `obs.owner_preference` adds a whole second to overall page load time.

Discussing the naming N+1s with @pellaea, i thought it might help to isolate all this naming calculation in a PORO, so that the methods that are currently Observation instance methods would be quarantined and could not accidentally call db queries. He fortunately knew how to sketch out a class that could do that, and i'm still working on that for the next PR. 

But that PORO seemed like the right solution for this n+1 too. They take the same args and both deal only with an observation's namings and votes.


app/models/observation/naming_consensus.rb
```diff
@@ -0,0 +1,49 @@
...
+    private
+
+    # Does observation.user have a single preferred id for this observation?
+    def owner_preference?
+      owner_uniq_favorite_vote&.value&.>= Vote.owner_id_min_confidence
+    end
+
+    def owner_uniq_favorite_name
+      favs = owner_favorite_votes
+      favs[0].naming.name if favs.count == 1
```
JoeCohen, line 37:
```suggestion
      favs[0].naming.name if favs.one?
```
Just a style thing, so take it or leave it. Interesting that there's no cop for this.

app/models/observation/naming_consensus.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def owner_uniq_favorite_vote
+      votes = owner_favorite_votes
+      votes.first if votes.count == 1
```
JoeCohen, line 42:
```suggestion
      votes.first if votes.one?
```

JoeCohen (approved):
Thanks for doing this. 
Although I added Issue #1754 related to owner_naming_line, feel free to ignore it in this PR. It's an old problem that's probably outside the scope of this PR.

nimmolo:
@JoeCohen - Good catch with the naming line not being updated. It's because my code doesn't update it! LOL

I think i'll fix that here because it has the same risk of setting off an N+1, and the same solution.

___

Incidentally - this is what Turbo 8 simplifies. I wanted to mention it here because it might help with getting the hang of how it works. Currently with Turbo, we must expressly send an update action for every DOM element (we can think of)) that should update when circumstances change on the page, as a result of a "live" db change from one of our Turbo forms. 

As you can tell, I didn't think of everything, but it's apparently a common issue. With Turbo 8, we can delete all those special update instructions and just `redirect` to the full page update (not even `render`! We can `redirect` like it's Rail 4 all over again!). Turbo 8 will diff the IDs that have changed, using an algorithm called "morphing" adapted from Svelte and similar frameworks, and sift them out for us, updating only these DOM elements.

Incredibly handy, it would simplify controller and view code quite a bit.

JoeCohen:
@nimmolo:
I started thinking about this again last night. Sorry the following didn't occur to me earlier:
- Maybe simply ditch the owner_naming_line
  - The information it supplies is redundant of the "little eyes" icons in the Proposed Names section.
  - It's expensive in light of that.
  - It's used by only 142 Users. (I no longer use it.) But about a dozen of them are power/important users including @mo-nathan.
If getting rid of it, maybe also replace the small eyes with something less ugly. A heart?♥︎ 
And maybe ditch the big eyes in the Proposed Names. It's redundant.
NOTE: I added this line years ago to try to satisfy a user who constantly complained that the Observation ID should be controlled by the user who created the Observation. He used the line, but kept on complaining.

- Maybe add  a column like `creator_preferred_id`. Then we need not calculate this every time the Observation is displayed. (But would need to be updated when the Namings are updated.)
That's probably how I should have done it in the first place.

nimmolo:
@JoeCohen - These are both interesting ideas. I don't have a preference. 142 is not a lot!

But know that with this PR merged, this line no longer costs any extra lookups. 

The line is now calculated with the new PORO i'm developing for the naming table, and uses associations that were already eager-loaded (even before this PR) but not utilized correctly.

PR #1756, opened 2023-12-28
Really fix `form_name_feedback`

nimmolo:
This fixes a bug reported by Huafang and caused by my PR the other day:`form_name_feedback` is printing double feedback.

To reproduce: 
- Create an obs, propose the name Grifola frondosa, and click "Create". There will be double suggestions:
Screen Shot 2023-12-28 at 4 11 40 PM

- On any obs, propose the name Grifola frondosa, and click "Propose". There will be double suggestions. Pick one, and then commit the form. On show obs reload, it will flash "There are problems with the name..." although this is no longer true.
Screen Shot 2023-12-28 at 4 13 51 PM

This PR also tries to fix what i was trying to fix originally, which is that `form_name_feedback` updates the flash too late for it to be picked up by a Turbo update. The naming error flash messages are now added upstream, either in `NamingParams#name_missing?` or `ObservationsController::Validators#validate_name`. 

Everything seems to work as before, except:

- The current "Missing name" flash message, which was only fired on Naming forms, is now replaced with the "There is a problem with the name", as on Observation forms. It seems to be superfluous because I think it only ever appears alongside "There is a problem".
- Depending if there are `valid_names` offered, `form_name_feedback` sends the flash message "There is a problem..." either as a `warning` or an `error`. But because "Missing name" is a flash error and not a warning, it is always shown as an error. There is currently (on `main`) effectively no difference between the flash messages currently shown in a modal naming form - both "Missing name" and "There is a problem..." are shown as a flash_error, even though they may be registered differently. The same seems to be true on the Observation form, but i haven't tested every case.
- Naming flash messages no longer appear inappropriately on the next page load.

### DIFF
--- a/app/assets/stylesheets/mo/_form_elements.scss
+++ b/app/assets/stylesheets/mo/_form_elements.scss
@@ -112,3 +112,9 @@ a.filter-only {
 .btn.text-danger {
   color: $brand-danger !important;
 }
+
+.name-radio {
+  label {
+    font-weight: normal;
+  } 
+}
--- a/app/classes/naming_params.rb
+++ b/app/classes/naming_params.rb
@@ -44,7 +44,8 @@ class NamingParams
     if @name && @what.match(/\S/)
       false
     else
-      @naming.errors.add(:name, :validate_naming_name_missing.t)
+      @naming.errors.add(:name,
+                         :form_observations_there_is_a_problem_with_name.t)
       true
     end
   end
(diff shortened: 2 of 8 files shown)


nimmolo:
@JoeCohen I think i should go ahead and merge this, it's definitely an improvement. Please let me know if you discover anything wrong, though.

nimmolo:
Oh nevermind, I can't deploy it anyway. 

You can merge and deploy if it seems ok!

PR #1758, opened 2023-12-30
Textile link case

JoeCohen:
Fixes a bug where non-lowercase user markup generated buggy Textile output
> Ex. `_Term gills_` generated garbage (while `_term gills_` correctly generated a link to lookup the GlossaryTerm "gills").

Delivers https://www.pivotaltracker.com/story/show/186064204


JoeCohen:
I will merge and deploy because this includes a big fix.

PR #1762, opened 2024-01-01
Name Form Update

JoeCohen:
- Updates the Name form by clarifying some terms, adding links, and tightening wording.
Hopefully this will avoid some user mistakes when creating MO Names.
- Delivers https://www.pivotaltracker.com/story/show/180451897


JoeCohen:
Merging as it only changes help/instruction text.

PR #1767, opened 2024-01-04
Fix Scientific Name label

JoeCohen:
Fixes a typo in the Name form which I introduced in #1762.


JoeCohen:
I will merge and deploy as soon as CI passes because this fixes a new bug.

PR #1773, opened 2024-01-06
Log in `MRTG` for better stats

nimmolo:
Our MRTG stats are pretty useless because MRTG is not logged in, so when we're testing pretty much any page but the home page, we’re only graphing the speed at which MO says “Please log in". This PR should change that.

I've made a new MO user "mrtg", and this PR changes the "autologin" method to log in requests from "127.0.0.1" as that user so we can re-enable graphs of a show_obs page.

@JoeCohen @mo-nathan The user "mrtg" must be validated by someone with access to "webmaster@mushroomobserver.org", I can't get in.

The test load of show obs is currently disabled, will need to change the `mrtg.cfg` on the server to fix.

### DIFF
--- a/app/controllers/account_controller.rb
+++ b/app/controllers/account_controller.rb
@@ -60,7 +60,7 @@ class AccountController < ApplicationController
       return render(action: :new) unless validate_and_save_new_user!
 
       UserGroup.create_user(@new_user)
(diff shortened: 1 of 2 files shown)


nimmolo:
Okay, i've edited the `mrtg.cfg` and the `show_observation.sh` script, and added a new `show_name.sh` script modelled on the former. Here goes.

PR #1774, opened 2024-01-07
Refactor "Show Name" using carousel, panels and icon links, like "Observation"

nimmolo:
This PR is to update the Name show page to the current style of the Observation show page.
- Views: Implement bootstrap carousel
- Views: Implement bootstrap panels with headings
- Helpers: Refactor MO `panel_block` helper to take a `heading_links` arg (for right-justified link icons, as the "shut-eye" on the Show Obs thumbnail map)
- CSS: Make `panel_heading` bold and smaller everywhere
- Helpers: Create MO `tabs` (reusable link arg arrays) for Name links
- Helpers: Refactor description-related `tabs` to use icons
- Helpers: Add new icons as needed
- Views: Refactor Show Name partials in straight-erb format for better code highlighting (to spot vars)
- Views: Show name partials - Consolidate and inventory ivars, locals and db lookups instigated by the partials, add comments

The partials are expensive here. On `main`, the `names/show/_observation_menu` partial currently runs up to 7 separate slow db loads of Names and Observations/Namings in order to give a count of different subsets of "observations of related taxa/subtaxa" to display in links to queries of these. Although the query is saved (speeding up the loading of these links if anyone clicks on them), it makes Show Name itself *very* slow to load.

This PR makes a first attempt to inventory these lookups and find a way to consolidate them into fewer queries. 
#1775 (soon to be merged here) makes a new PORO to quarantine seven of the `_observation_menu` `select_count` lookups in a single db query and derive the counts by selecting within that query, and reduces page load time by ~33%.

```
Completed 200 OK in 4539ms (Views: 475.0ms | ActiveRecord: 3743.0ms | Allocations: 552513) <- main, repeated load
Completed 200 OK in 3023ms (Views: 194.6ms | ActiveRecord: 2728.6ms | Allocations: 269403) <- #1775, repeated load
```


JoeCohen:
Suggestion: In the About This Taxon panel
Screenshot 2024-01-08 at 6 19 41 AM
- Omit link to Observations of Subtaxa if there aren't any (current behavior).
- Alternatively -- if it won't cause another query -- change it so that it's similar to the links above. Something like:
subtaxa (0)

I personally lean very slightly to the current behavior.
If you prefer the alternative, consider doing it in a different PR. This one is already large.

JoeCohen:
Suggestion:
Increase size of rh column so that links to IF and MB never take > 1 line each. (Current Behavior.)
See screenshot of new behavior (varies by window size).
new show name

JoeCohen:
Looks like there's an extra escape or similar in the Nomenclarture author/display_name. See http://localhost:3000/names/66482
Screenshot 2024-01-08 at 7 13 03 AM

JoeCohen:
Also a similar  escape-type issue in Classification of deprecated Names.
Ex: http://localhost:3000/names/1551
Screenshot 2024-01-08 at 9 54 45 AM

app/controllers/names_controller.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
 class NamesController < ApplicationController
```
JoeCohen, line 4: Suggestion: 
Disable the cop via a comment on this line (and delete `# rubocop:enable Metrics/ClassLength` at the end of the file).
``` ruby
class NamesController < ApplicationController # rubocop:disable Metrics/ClassLength
```

app/controllers/names_controller.rb
```diff
@@ -670,3 +711,4 @@ def permitted_name_params
     params.permit(name: [:author, :citation, :icn_id, :locked, :notes, :rank])
   end
 end
+# rubocop:enable Metrics/ClassLength
```
JoeCohen, line 714:
```suggestion
```

app/views/controllers/names/show.html.erb
```diff
@@ -6,35 +6,32 @@ add_pager_for(@name)
...
-  <div class="col-sm-5">
+  <div class="col-xs-12 col-md-7">
 
     <%= render(partial: "names/show/confident_observations_panel") %>
 
-    <div id="name_previous_export" class="mb-4">
-      <%= show_previous_version(@name) %>
-      <%= export_status_controls(@name) %>
-    </div>
+    <% if !@best_description.blank? %>
```
JoeCohen, line 19:
```suggestion
    <% if @best_description.present? %>
```

nimmolo:
@JoeCohen - Thank you for testing out this PR, and for these two well-documented suggestions, they're easy to follow.

I should have left a note - Maybe preferable to test out #1775 instead. That is a PR off this branch, not an independent PR, and it's where most of the query improvements are, due to the new PORO. It's much faster to load the show_name page — I'd love to hear exactly how much faster it is on whatever machine you have in Oahu.

I kept the two branches separate for debugging, but really they're one package. If it's easier I'll merge them now. I didn't realize til you pointed it out, but I only added the tally to the subtaxa link in #1775.  I believe it doesn't print the "Subtaxa" line if there are no subtaxa (or it prints with (0) like the others). 

I will adjust the columns in Nomenclature and fix the escapes here.

Further dramatic eager-loading improvements are in #1777, more than the PR title would suggest. It affects all `show` actions that have `versions`. Take a peek at the new AR query helper method `strict_loading`, which is a direct and more rigorous alternative to Bullet, and runs before it.

> Suggestion: In the About This Taxon panel Screenshot 2024-01-08 at 6 19 41 AM
> 
> * Omit link to Observations of Subtaxa if there aren't any (current behavior).
> * Alternatively -- if it won't cause another query -- change it so that it's similar to the links above. Something like:
>   subtaxa (0)
> 
> I personally lean to the current behavior. If you prefer the alternative, consider doing it in a different PR. This one is already large.
> 
> PS: Another reason to prefer the current behavior: group/clade Names have no MO subtaxa. So the link would make no sense there and certainly should be omitted. And treating goups/clades different from other MO Names would be another complication. So treat them all the same.

nimmolo:
@JoeCohen - 

IIRC the problem with the current behavior for "Subtaxa" is that the check "if there ARE subtaxa" does one of the slow db lookups. (Every tally in that column does an otherwise-unnecessary observations table lookup, currently, and they are the slowest things on the page!) I will investigate how i can hide it if (0) with the refactor in #1775.

app/views/controllers/names/show/_classification.html.erb
```diff
@@ -3,40 +3,54 @@ approved_name = @name.approved_name
...
-<% end %>
+<%= if approved_name.classification.present? && parents.any?
+  tag.div(class: "mb-2") do
+    ([approved_name] + parents).reverse.each do |n|
+      concat(tag.p do
+        concat("#{rank_as_string(n.rank)}: ")
+        concat(tag.i(link_with_query(n.text_name.t, n.show_link_args)))
+        if n == approved_name && approved_name != @name
+          concat([
+            safe_br, "&nbsp;&nbsp; (= ", tag.i(@name.text_name.t), ")"
```
JoeCohen, line 14:
```suggestion
            safe_br, safe_nbsp, safe_nbsp, " (= ", tag.i(@name.text_name.t), ")"
```
Fixes this issue https://github.com/MushroomObserver/mushroom-observer/pull/1774#issuecomment-1881733372

nimmolo: Thanks!

JoeCohen:
@nimmolo: Re speed.
#1774 is very fast locally. But I think I've never notice speed issues in show_name when running locally.

JoeCohen:
PS: I'm on a MacBookPro, Intel processor, max memory (16 GB 2667 MHz DDR4), Sonoma 14.2.1.

nimmolo:
@JoeCohen My computer is older, but on main, show name for "Grifola frondosa" takes nearly 5 seconds for me locally. 

I still am curious if you will find proportional improvements. If you check your console log, when the page is loading locally, you should see an exact figure in milliseconds at the end of all log messages when the page is rendered.

If you switch between branches without stopping the server, you should find a marked difference between the load times for the page! 

I'm bummed that MRTG is not set up right, but this is the only way we have currently to benchmark improvements for all this work. That's why i'm so curious for exact figures on different machines, it's all we're going to get until I can fix MRTG.

nimmolo:
@JoeCohen 
I moved a couple more links around. The tabs in the top right of the page, for synonym adjustment, deprecation etc. are now icons under "Nomenclature" with hover labels.

These will take some getting used to for the few power users who use them, but i'm trying to make the page less overwhelming and more intuitive for newbies. Let me know if you feel anything would be better elsewhere.

Screen Shot 2024-01-09 at 12 33 51 PM

PR #1783, opened 2024-01-09
Lifeform form method

### LINKED ISSUE #1779: Edit Lifeforms broken
To replicate:
- Goto http://localhost:3000/names/5223
- Next to Lifeform, click Edit (http://localhost:3000/names/5223/lifeforms/edit)
- Make a change (e.g., uncheck Basidiolichen)
- Save
Result: 
>Routing Error
No route matches [POST] "/names/5223/lifeforms"

nimmolo:
When it's a `form_with(url:)` we have to set the `method:` arg explicitly. For edit, that's `put` or `patch`. It defaults to `post`.

@JoeCohen If you can think of a test that submits this form, please write one. I think we just need an integration test that loads and submits the form. I'm going to merge without it though, too many fish to fry right now.

Should fix #1779

### DIFF
--- a/app/views/controllers/names/lifeforms/edit.html.erb
+++ b/app/views/controllers/names/lifeforms/edit.html.erb
@@ -8,7 +8,8 @@ action = { controller: "/names/lifeforms", action: :update,
            id: @name.id, q: get_query_param }
 %>
 
-<%= form_with(url: action, scope: :lifeform, id: "name_lifeform_form") do |f| %>
+<%= form_with(url: action, method: :put, scope: :lifeform,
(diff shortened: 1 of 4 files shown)


nimmolo:
Joe - I think what would work here is adding a `test_name_edit_lifeform` to `names_integration_test`. 
The test would simply have to submit the form and check the result. I will leave it to you if you want to write it, otherwise feel free to merge this and move on. I have a lot on my plate, and many issues to fix.

JoeCohen:
@nimmolo
This does fix the bug, but exposes some other issues.
I'll look into them and add at least an integration test.

PR #1793, opened 2024-01-12
Ruby 3.3.0

nimmolo:
What do you guys think? 

Very few changes required.


JoeCohen:
I think it's a good idea. There should be some, perhaps marginal, performance improvement.
Question: Do the big changes (parsers, JIT) affect us?

nimmolo:
YJIT is supposed to be helpful, but we have to implement it. I don't think the parsers affect us. 

I've installed 3.3.0 on the server so we don't get a crash like last time, if we deploy.

nimmolo:
On Apple M3, with a Rails app running in Docker (only), Ruby 3.3.0 causes a problem with `assets:precompile`, documented here:
https://stackoverflow.com/questions/77725755/segmentation-fault-during-rails-assetsprecompile-on-apple-silicon-m3-with-rub

@mo-nathan If this affects you, I will hold off upgrading Ruby until 3.3.1 comes out, which has the bugfix.

I know you've experimented with docker for developer-startup. But as far as I know, you're running Rails native, not in Docker.

PR #1797, opened 2024-01-12
Create Observation blank Scientific Name

### LINKED ISSUE #1778: Flash error when creating Observation without Name
I created two Observations, leaving Scientific Name blank. (The issue does not occur if I fill in Scientific Name.)
MO correctly creates the Observation, but gives the spurious flash error:
>There is a problem with the name. See message below.
(There is no message below.)
Screenshot 2024-01-07 at 3 54 29 PM

JoeCohen:
Leaving the Scientific Name blank when creating an Observation does not cause a flash error.
- Fixes  #1778.

### DIFF
--- a/app/controllers/observations_controller/validators.rb
+++ b/app/controllers/observations_controller/validators.rb
@@ -23,6 +23,19 @@ module ObservationsController::Validators
   end
 
   def validate_name(params)
+    (success, @what, @name, @names, @valid_names,
+     @parent_deprecated, @suggest_corrections) = resolve_name_ivars(params)
+    if @name
+      @naming.name = @name
+    elsif !success
+      @naming.errors.add(:name,
+                         :form_observations_there_is_a_problem_with_name.t)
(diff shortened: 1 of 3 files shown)


app/controllers/observations_controller/validators.rb
```diff
@@ -23,26 +23,30 @@ def validate_params(params)
...
-    # NOTE: views could be refactored to access properties of the @resolver,
-    # e.g. `@resolver.valid_names`, instead of these ivars.
-    # All but success, @what, @name are only used by form_name_feedback.
     (success, @what, @name, @names, @valid_names,
-     @parent_deprecated, @suggest_corrections) = @resolver.ivar_array
+     @parent_deprecated, @suggest_corrections) = resolve_name_ivars(params)
     if @name
       @naming.name = @name
-    else
+    elsif !success
```
JoeCohen, line 30: @nimmolo: This line is the only substantive change on the app side.

PR #1815, opened 2024-01-17
UTF8 character sets

### LINKED ISSUE #1808: Sequence Notes caron
Certain string values in Sequence Notes cause `Mysql2::Error: Incorrect string value`. See example below.
###Tasks
- [x] Find short string that provokes the error
- [x] Write test (Red)
- [x] Make test Green
- [x] Rubocop
- [x] PR

JoeCohen:
- Changes Articles and Sequences  default character sets  to `utf8mb4`  from `latin1`.  (They the only tables that used a latin charset.)
- Fixes #1808

### DIFF
--- a/db/migrate/20240117180824_latin1_to_utf8.rb
+++ b/db/migrate/20240117180824_latin1_to_utf8.rb
@@ -6,10 +6,5 @@ class Latin1ToUtf8 < ActiveRecord::Migration[7.1]
       execute("ALTER TABLE sequences CONVERT TO CHARACTER SET utf8mb4;")
   end
 
-  def down
-    Article.connection.
-      execute("ALTER TABLE articles CONVERT TO CHARACTER SET latin1;")
-    Sequence.connection.
-      execute("ALTER TABLE sequences CONVERT TO CHARACTER SET latin1;")
-  end
+  def down; end
 end
--- a/db/migrate/20240117180824_latin1_to_utf8.rb
+++ b/db/migrate/20240117180824_latin1_to_utf8.rb
@@ -1,0 +1,15 @@
+class Latin1ToUtf8 < ActiveRecord::Migration[7.1]
+  def up
+    Article.connection.
+      execute("ALTER TABLE articles CONVERT TO CHARACTER SET utf8mb4;")
+    Sequence.connection.
+      execute("ALTER TABLE sequences CONVERT TO CHARACTER SET utf8mb4;")
+  end
+
+  def down
+    Article.connection.
+      execute("ALTER TABLE articles CONVERT TO CHARACTER SET latin1;")
+    Sequence.connection.
+      execute("ALTER TABLE sequences CONVERT TO CHARACTER SET latin1;")
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_12_27_170054) do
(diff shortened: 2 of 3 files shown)


db/migrate/20240117180824_latin1_to_utf8.rb
```diff
@@ -0,0 +1,15 @@
+class Latin1ToUtf8 < ActiveRecord::Migration[7.1]
+  def up
+    Article.connection.
+      execute("ALTER TABLE articles CONVERT TO CHARACTER SET utf8mb4;")
+    Sequence.connection.
+      execute("ALTER TABLE sequences CONVERT TO CHARACTER SET utf8mb4;")
```
JoeCohen, line 6: 1. Should I have separate migrations for Article and Sequence?
2. Is is possible to do this without SQL, for instance using `:options`? I couldn't figure out how.

db/migrate/20240117180824_latin1_to_utf8.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def down
+    Article.connection.
+      execute("ALTER TABLE articles CONVERT TO CHARACTER SET latin1;")
+    Sequence.connection.
+      execute("ALTER TABLE sequences CONVERT TO CHARACTER SET latin1;")
+  end
```
JoeCohen, line 14: I should probably delete the down migration. It was OK for testing purposes.
But it will blow up if either of these tables has a single non-latin character.

db/migrate/20240117180824_latin1_to_utf8.rb
```diff
@@ -0,0 +1,15 @@
+class Latin1ToUtf8 < ActiveRecord::Migration[7.1]
+  def up
```
JoeCohen, line 2: 1. Should I have separate migrations for Article and Sequence?
2. Is is possible to write this migration this without SQL, for instance using `:options`? I couldn't figure out how.

db/migrate/20240117180824_latin1_to_utf8.rb
```diff
@@ -0,0 +1,15 @@
...
+    Article.connection.
+      execute("ALTER TABLE articles CONVERT TO CHARACTER SET utf8mb4;")
```
JoeCohen, line 4: Comment: MySQL recommends utf8mb4.  See https://dev.mysql.com/blog-archive/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/
Maybe I should make it the db default.

db/migrate/20240117180824_latin1_to_utf8.rb
nimmolo: Since this is a hot fix, I vote to do both migrations in one. I don't think it can be done without SQL, and from what i can tell from a quick search this seems like it would be valid SQL for any other db we might use.

db/migrate/20240117180824_latin1_to_utf8.rb
nimmolo: Yes, I vote to make it the db default.

db/migrate/20240117180824_latin1_to_utf8.rb
nimmolo: Sounds right to me.

nimmolo (approved):

db/migrate/20240117180824_latin1_to_utf8.rb
JoeCohen: I'll do that in another PR. 
It's probably safe, but I'd like to think about its consequences.

PR #1820, opened 2024-01-18
Cache the sidebar nav

nimmolo:
Caches the nav partials as two fragments.

User sub-nav, for "logged in user nav on mobile", keyed by user.id. This will cache a different "user" sub-nav per user, so that Joe doesn't get a cached menu with Nathan's name. (Note that this cached version, even if Joe did get it, still wouldn't let him impersonate Nathan, because the controllers access the session user. But it would be disorienting.)
Screen Shot 2024-01-18 at 8 36 03 PM

The rest of the nav is keyed by `user_status_string`, so there is much less to cache. Four different "user stati" generate potentially different nav links more or less throughout the rest of the menu.
- admin_mode
- browser.bot?
- logged_in (cache by user id)
- no_user

### DIFF
--- a/app/assets/stylesheets/mo/_utilities.scss
+++ b/app/assets/stylesheets/mo/_utilities.scss
@@ -321,6 +321,14 @@
   margin-left: 10px !important;
 }
 
+.ml-1 {
+  margin-left: 0.25rem !important;
+}
+
+.ml-2 {
+  margin-left: 0.5rem !important;
+}
+
 .ml-3 {
   margin-left: 1rem !important;
 }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -74,6 +74,21 @@ module ApplicationHelper
                      class: class_names("alert mt-3", alert_class))
   end
 
+  # Returns a string that indicates the current user/logged_in/admin status.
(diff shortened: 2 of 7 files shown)


JoeCohen:
Yes, this could really use reorganization, and I like doing it in small steps.
It''s generally great. I love the idea of a reduced sidebar for anonymous users.

Some comments re News
- Since News was removed from the home page, it's become a less effective way of communicating with users.
- Non logged in users should have a link to the News indiex.
- Maybe remove News from the Index area, and put it at the top (for both logged in and non-logged in users).

JoeCohen (changes requested):
See https://github.com/MushroomObserver/mushroom-observer/pull/1820#issuecomment-1899506725

nimmolo:
@JoeCohen as you probably noticed I made your requested "News" changes on #1820, good idea.

Let me know if you want to play with this caching setup. The relevant test is to log in as two different users in two different browsers, on development, and check that the navbar knows who you are in both cases. (You can only test that by shrinking your browser window, because currently the user menu only displays in the sidebar on mobile.)

It's conceivable we could write a system test that tests that using two different sessions, but i'm more interested to check if the cache makes any difference on the server. I don't think the test environment is set up to use the cache in any case.

PR #1821, opened 2024-01-19
Reorganize sidebar nav a bit

nimmolo:
Part of gradual navbar reorganization in preparation for the Bootstrap 4 navbars. 

My idea is that moving links in small steps is less vexing for users, so this is not intended to be the "goal state" of the navbar, but a midway point. I think we probably want to make Projects more prominent, and Lists less prominent. I feel like the `Indexes` subheading is not organized or worded as intuitively as it could be - but that's for another PR. 

- Adds a nav link for "Latest Observations"
- for logged-out users, hides nav items they can't access, except "Help Identify". Keeping this because I think we want to signal to people that help with ID is a big part of MO even before they sign up.
- Moves "Info" links to the bottom, under "More". They seem to go together.
  - For logged out users, it's not very far from the top. 
  - For logged in users, this is not the highest priority menu.

`Logged in:                          Logged out:                          `
Screen Shot 2024-01-18 at 6 12 42 PM Screen Shot 2024-01-18 at 6 11 15 PM


nimmolo:
@JoeCohen - I modified the above screengrabs. Please have a look and let me know if that seems better. 

Or maybe "News" should go under "More", and that whole "More" section should come after "Observations" (and above Indexes and Lists).

JoeCohen:
> @JoeCohen - I modified the above screengrabs. Please have a look and let me know if that seems better.
> 
> Or maybe "News" should go under "More", and that whole "More" section should come after "Observations" (and above Indexes and Lists).

Looks good!

nimmolo:
Thanks Joe - will merge, and continue work on caching the navbar.

PR #1823, opened 2024-01-19
Sensu lato propagation

### LINKED ISSUE #1817: Sensu lato propagation
Prevent Classification of Name sensu lato from being propagated to subtaxa.
Here's an example why:
The Classification of Pleurotus s.s. is:
>...
Class: Agaricomycetes
**Order: Agaricales
Family: Pleurotaceae**
Genus: Pleurotus

The Classification of Pluerotus s.l. is:
>...
Class: Agaricomycetes
Genus: Pleurotus

The way MO works now, Pleurotus sl has rank genus, author "sensu lato".
If its Classification is propogated to subtaxa, then all the subtaxa lose the Order and Family.

In the long run, we probably need to change the classification scheme. I'm not sure how, and it would involve a lot of work.
So to prevent accidental loss of information int he meantime, I propose simply: don't include Propogate to Subtaxa if the author matches "sensu lato".

JoeCohen:
- Does not show a classification propagation link for Names sensu lato.
- The change prevents potential propogation of an overbroad, imprecise Classification to lower ranks.
- Fixes #1817

### DIFF
--- a/test/models/name_test.rb
+++ b/test/models/name_test.rb
@@ -3284,6 +3284,31 @@ class NameTest < UnitTestCase
     assert_names_equal(names(:stereum), names(:stereum_hirsutum).accepted_genus)
   end
 
+  def test_can_propagate
+    assert(names(:coprinus).can_propagate?,
+           "Classification of genus Names should be propagable")
+
+    assert_false(names(:coprinus_sensu_lato).can_propagate?,
+                 "Classification of Names sensu lato should not be propagable")
+    [:eukarya, :fungi, :ascomycota, :ascomycetes, :agaricales,
+     :agaricaceae].each do |name|
+      assert_false(names(name).can_propagate?,
+                   "Classification of Names above genus should be propagable")
+    end
+    assert_false(
+      names(:amanita_subgenus_lepidella).can_propagate?,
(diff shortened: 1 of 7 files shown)


test/models/name_test.rb
```diff
@@ -3284,6 +3284,31 @@ def test_genus_of_synonym
...
+      "Classification of infra-generic Names should not be propagable"
+    )
+    assert_false(
+      names(:amanita_boudieri_var_beillei).can_propagate?,
+      "Classification of infra-specific Names should not be propagable"
+    )
+    assert_false(
+      names(:boletus_edulis_group).can_propagate?,
+      "Classification of group or clade Names should not be propagable"
+    )
```
JoeCohen, line 3309: Dry and simply this. Note type in original should be "should _not_ be propagated"
Something like:
```ruby
    [<all the bad name symbols>].each do |name|
      assert_false(names(name).can_propagate?,
                   "Classification of Names of rank {#name.ran} should not propagate")
    end
```
Might need to fixed pluralization of "genus" and "subgenus":
```ruby
irb(main):002> "genus".pluralize
=> "genus"
irb(main):006> "subgenus".pluralize
=> "subgenus"
```

PR #1833, opened 2024-01-19
Hotfix for broken sidebar links

### LINKED ISSUE #1832: "Your Observations" is broken
Reported by users: When I click on "Your Observations" (ie. to see my observations) I get "no matching observations found." 
Some bug with "Your Lists"

JoeCohen:
- Caches application/sidebar/observations and application/sidebar/species_lists at the user instance because they contain links at the user instance to Your Observations and Your Lists
- Fixes #1832 
- May need follow-up fine tuning
- It's highly desirable to find a way to test this stuff to avoid future breakage. 
- 
But I will deploy as soon as CI tests are green because:
- the website is partly broken
- it would take too for me to figure out how to fine tune (within those 2 partials)
- it would take me too long to figure out how to test it

### DIFF
--- a/app/views/controllers/application/_sidebar.html.erb
+++ b/app/views/controllers/application/_sidebar.html.erb
@@ -37,16 +37,32 @@ classes = {
         <%= render(partial: "application/sidebar/login",
                     locals: { classes: classes }) %>
       <% end %>
+    <% end %>
 
-      <%= render(partial: "application/sidebar/observations",
-                  locals: { classes: classes }) %>
+    <%# Cache customized for the user instance because of Your Observations.%>
+    <% cache(@user) do %>
+      <% if @user %>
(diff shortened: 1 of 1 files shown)


nimmolo:
@JoeCohen - The issue here was cache keys, which need to be unique within the template. 

When you call `cache(thing)`, `thing` is evaluated as a string, even if it's an instance. So `cache(@user)` becomes something like `users/1-20240119185828000000` within the cache key, which is longer: `views/application/_sidebar:then_a_long_digest_string_like_for_other_assets/users/1-20240119185828000000/en`. 

Yet despite how long this key is, it will be generated the same if called for a second cache on the same instance in the same template. The way to work with that is to pass an array of things to use as the key, like `cache([user_status_string, "login"])`

Issue #1841, opened 2024-01-20
Names with Observations "All Names" doesn't work


JoeCohen:
The link on https://mushroomobserver.org/names?with_observations=true to All Names goes to https://mushroomobserver.org/names?with_observations=true
A link to All Names was added to this page in deploy-2023-08-26-00-16 4d125b5ffde2671c82ddb9bec20f9babfd2ef348
and has since been modified. But it has always gone to the wrong url.

PR #1843, opened 2024-01-20
All Names link

JoeCohen:
- Fixes the link to `All Names` on the Names With Observation page (fixing Issue #1841)
- Adds tests and new helpers

### DIFF
--- a/app/helpers/tabs/names_helper.rb
+++ b/app/helpers/tabs/names_helper.rb
@@ -101,24 +101,7 @@ module Tabs
        { class: tab_id(__method__.to_s), target: :_blank, rel: :noopener }]
     end
 
-    def name_correct_spelling_tab(name)
-      [name.correct_spelling.display_name.l,
-       add_query_param(name_path(name.correct_spelling_id)),
-       { class: tab_id(__method__.to_s) }]
-    end
-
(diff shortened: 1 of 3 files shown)


app/helpers/tabs/names_helper.rb
```diff
@@ -101,24 +101,7 @@ def mycobank_basic_search_tab
        { class: tab_id(__method__.to_s), target: :_blank, rel: :noopener }]
     end
 
-    def name_correct_spelling_tab(name)
```
JoeCohen, line 104: Dead Code

app/helpers/tabs/names_helper.rb
```diff
@@ -234,6 +217,21 @@ def descriptions_of_these_names_tab(query)
...
+    def all_names_index_tabs(query:)
+      [
+        new_name_tab,
+        all_names_tab(query),
+        coerced_observation_query_tab(query)
+      ].reject(&:empty?)
+    end
+
+    def all_names_tab(query)
+      return if query&.flavor == :all || query&.flavor&.empty?
```
JoeCohen, line 229: @nimmolo: Does this make sense? Is the 2nd condition over-cautious?

PR #1850, opened 2024-01-24
Fix caching sidebar links where query is relevant

### LINKED ISSUE #1849: Query getting lost on SpeciesListsController#show
- Go to any Species List
- Click AddTo/RemoveFrom Another List
result: Flash Error
Screenshot 2024-01-23 at 3 38 54 PM

I think this is because the query param is not getting passed in, when the page is accessed (from where? -AN).

It could result from the species lists nav link in the sidebar being cached with a stale query. It's cached per user, but not the query record.

nimmolo:
The sidebar obs and list links need to incorporate the current query, to work properly. They're currently only cached on the current user, whose query may have changed. 

Solutions: they could either (1) not be cached, or (2) cache with the current `QueryRecord` as part of the key. 

I'm going to try the former. But here's a question: how can I easily get the currently relevant `QueryRecord`? We don't seem to have a helper that gets the record instance. And does the query in the link rather use the `query_from_session` or the `passed_query`, or try one then the other?

#1849

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -954,6 +954,15 @@ class ApplicationController < ActionController::Base
   end
   helper_method :passed_query
 
+  # TODO: If we're going to cache user stuff that depends on their present q,
+  # we'll need a helper to make the current QueryRecord (not just the id)
+  # available to templates as an ApplicationController ivar. Something like:
+  #
+  # def current_query_record
+  #   current_query = passed_query || query_from_session # could both be nil!
+  #   current_query_record = current_query&.record || "no_query"
+  # end
+
   # Return query parameter(s) necessary to pass query information along to
   # the next request. *NOTE*: This method is available to views.
   def query_params(query = nil)
@@ -1006,8 +1015,6 @@ class ApplicationController < ActionController::Base
   helper_method :get_query_param
 
   # NOTE: these two methods add q: param to urls built from controllers/actions.
-  # Seem to be dodgy with Rails routes path helpers. If encountering problems,
-  # try redirect_to(whatever_objects_path(q: get_query_param)) instead.
   def redirect_with_query(args, query = nil)
     redirect_to(add_query_param(args, query))
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -946,6 +946,7 @@ class ApplicationController < ActionController::Base
 
     Query.safe_find(id)
   end
+  helper_method :query_from_session
 
   # Get instance of Query which is being passed to subsequent pages.
   def passed_query
--- a/app/helpers/tabs/species_lists_helper.rb
+++ b/app/helpers/tabs/species_lists_helper.rb
@@ -5,20 +5,20 @@ module Tabs
     # Moved download link into species_list_logged_in_show_tabs and
     # nixed user: kwarg
     # Can't access this page unless logged in as of 2023
-    def species_list_show_tabs(list:)
-      tabs = species_list_logged_in_show_tabs(list)
+    def species_list_show_tabs(list:, query: nil)
+      tabs = species_list_logged_in_show_tabs(list, query)
       return tabs unless check_permission(list)
 
       tabs += species_list_user_show_tabs(list)
       tabs
     end
 
-    def species_list_logged_in_show_tabs(list)
+    def species_list_logged_in_show_tabs(list, query = nil)
       [
         species_list_download_tab(list),
         species_list_set_source_tab(list),
         clone_species_list_tab(list),
-        species_list_add_remove_from_another_list_tab(list)
+        species_list_add_remove_from_another_list_tab(list, query)
       ]
     end
 
@@ -57,10 +57,10 @@ module Tabs
          help: :species_list_show_set_source_help.l }]
     end
 
-    def species_list_add_remove_from_another_list_tab(list)
+    def species_list_add_remove_from_another_list_tab(list, query = nil)
       [:species_list_show_add_remove_from_another_list.t,
        add_query_param(
-         edit_species_list_observations_path(species_list: list.id)
+         edit_species_list_observations_path(species_list: list.id), query
        ),
        { class: tab_id(__method__.to_s) }]
     end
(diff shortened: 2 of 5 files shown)


pellaea:
I don't remember how it looks up the currently relevant query record.
Should be in ApplicationController.

nimmolo:
@JoeCohen This should fix the SpeciesList missing query issue. It disables the cache on those menu items, until I can figure out a reliable way to get the current QR.

JoeCohen:
This isn't working for me. I still get that flash error with this sequence:
- Home page
-  click Your Lists in left navbar
-  click any of my lists
-  click Add/Remove from Another List
result: same flash as above

nimmolo:
Ahaa. This issue is likely not related to caching. (My detective skills must be off today!)

I checked out branch `jdc-1778-create-observation-name-blank`, which predates all the caching, and the same workflow produces the same issue. 

I'll keep the work on this PR, though, because it remains true that these cached links in the sidebar will _necessarily_ be stale if the query changes, so they should either not be cached, or keyed to the QR. The change thus far remains desirable, then.

But we need further diagnosis to determine why the flash message, or the stale query, is showing up on this add/remove form. It may be that the controller doesn't `pass_query_params` when it should, or something else.

JoeCohen:
I will also look at this. I will start by trying to find when the bug shows
up.
There's certainly a q param that's gone missing at some point. (Not sure
where.)
At the end of the day, it's flashing because

   - app/controllers/species_lists/observations_controller.rb#find_obs_query_or_redirect
   is unable to find an Observation Query (one was created in at least
   app/controllers/species_lists_controller#show).because
   - app/controllers/application_controller.rb is returns nil when it calls
      dealphabetize_q_param because
         - at that point there is no q param

JoeCohen:
Here's where it breaks:
>commit e761cedf1ed3193ad08096d21d45b0a634f61cb9 (HEAD, tag: deploy-2023-08-13-18-12)
Merge: 7331010c4 eb5e8ba65
Author: andrew nimmo <andrnimm@fastmail.fm>
Date:   Sun Aug 13 02:59:41 2023 -0700
  >  Merge pull request #1551 from MushroomObserver/nimmo-refactor-tabs-helpers
  >  Move app partials to views/application

JoeCohen:
Oh duh. Github's #1551 doesn't show the changes that were merged into that PR from #1549.

JoeCohen:
@nimmolo:
TLDR: I've got it.  I'll write some tests, make the needed changes and push to this branch.

Long story:
PR #1547 dropped a query argument as folllows.
That PR changed `app/views/species_lists/show.html.erb` from
```ruby
 if @user
    tabs += [
    …
      link_to(:species_list_show_add_remove_from_another_list.t,
              edit_species_list_observations_path(
                species_list: @species_list.id, q: get_query_param(@query))
              )
    ]
  end
```
[Note: this is the only explicit call to @query in the changed part of file at that time]
to
```ruby
  links = species_list_show_links(list: @species_list)
  tabs = create_tabs(links)
```
with `app/helpers/tabs/species_lists_helper.rb` (5b31be4)
having
```ruby
   def species_list_show_links(list:)
      links = species_list_logged_in_show_links(list)
      …
    end

  def species_list_logged_in_show_links(list)
      [
        …
        [:species_list_show_add_remove_from_another_list.t,
         add_query_param(
           edit_species_list_observations_path(species_list: list.id)
         ),
         …
      ]
    end
```
So we need to change the last line to restore a query argument.

nimmolo:
@JoeCohen - just catching up here. Does this mean that `add_query_param` isn't working in the navbar? 

Note that the `logged_in_show_links` is trying to add it via `add_query_param`, where it previously did `q: get_query_param`. 

I've been treating these as interchangeable (and actually converting the latter to the former in many places)!

nimmolo:
The sidebar link to "Your lists" does _not_ include a q, either. Is it possible that's where it's getting lost?

In fact it seems it never had one, so maybe that means it doesn't need it.  
Even before I broke up `left_nav` into the partials in #1150, it was like this:
```ruby
<% unless @user.nil? %>
    <%= link_to(:app_your_lists.t,
                species_list_species_lists_by_user_path(id: @user.id),
                { class: "list-group-item indent" } ) %>
  <% end%>
```

JoeCohen:
@nimmolo: I think the sidebar link is OK.
The relevant query is created in SpeciesListsControllers#show
```ruby
    @query = create_query(:Observation, :in_species_list,
                          by: :name, species_list: @species_list)
```

nimmolo:
Aha. Maybe the most useful change would be to make it so `add_query_param` checks for that `@query` ivar, and uses it if exists?

nimmolo:
Not sure at all. 
~Let's maybe just revert to use `get_query_param`, but add a comment why it's like that - that `show` generates the ivar.~
Aha, no. `add_query_param` can take a query as a param. Let's do this maybe:
```ruby
    def species_list_add_remove_from_another_list_tab(list)
      [:species_list_show_add_remove_from_another_list.t,
       add_query_param(
         edit_species_list_observations_path(species_list: list.id), @query # generated in #show
       ),
       { class: tab_id(__method__.to_s) }]
    end

```
 
I'm sure i've converted many links from `get_query_param` to `add_query_param`, and although I do remember noticing and looking up the difference between `get_query_param(@query)` and `get_query_param`, I couldn't swear that i haven't overlooked it at some point. At least if we find more non-functioning links, we'll know what the cause was, now.

JoeCohen:
@nimmolo: That would work, but
we shouldn't have an ivar in a helper method.
So we've also have to add an optional `query` parameter to  `species_list_add_remove_from_another_list_tab` nand to its caller, `species_list_logged_in_show_tabs` and ,
and call species_list_logged_in_show_tabs (the relevant top-level helper method) with @query

I think the simplest safe fix is a one-line definition `@query_params[:q]` in SpeciesListsControllers#show.

So to start things off, I'll do that and push it to this branch.
Then you can change things if you like.

JoeCohen:
@nimollo: 
I ended up doing what you suggested (with modifications): supply a `query` parameter to `species_list_add_remove_from_another_list_tab`.
CI is running now.
My idea (@query_params[:q] in SpeciesListsControllers#show) won't work because `q` doesn't exist at that point.

PR #1858, opened 2024-01-26
Clean up `autologin` multiple lookups of user

nimmolo:
The `autologin` before_filter has been doing multiple lookups of the same user for ages:

```
Started GET "/" for ::1 at 2024-01-26 12:25:44 -0800
  ActiveRecord::SchemaMigration Load (0.2ms)  SELECT `schema_migrations`.`version` FROM `schema_migrations` ORDER BY `schema_migrations`.`version` ASC
Processing by ObservationsController#index as HTML
START AUTOLOGIN BEFORE FILTER
  User Load (0.3ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 7238 LIMIT 1
  ↳ app/models/abstract_model.rb:154:in `safe_find'
  User Load (0.4ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 7238 LIMIT 1
  ↳ app/models/user.rb:413:in `reload'
user=1 robot=N ip=::1
  TRANSACTION (0.2ms)  BEGIN
  ↳ app/models/user.rb:1145:in `login_already_taken?'
  User Load (31.6ms)  SELECT `users`.* FROM `users` WHERE `users`.`login` = 'Joe Namath' LIMIT 1
  ↳ app/models/user.rb:1145:in `login_already_taken?'
  User Update (0.4ms)  UPDATE `users` SET `users`.`updated_at` = '2024-01-26 20:25:45', `users`.`last_activity` = '2024-01-26 20:25:44' WHERE `users`.`id` = 7238
  ↳ app/controllers/application_controller.rb:423:in `track_last_time_user_made_a_request'
  TRANSACTION (4.4ms)  COMMIT
  ↳ app/controllers/application_controller.rb:423:in `track_last_time_user_made_a_request'
END AUTOLOGIN BEFORE FILTER
```

This PR gets rid of the reload of the (freshly loaded!) `@user`,  and changes `User` validation to only validate the login/password if it's a new record, or they're changing. That saves the `login_already_taken?` check, which is the slower lookup.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -354,10 +354,13 @@ class ApplicationController < ActionController::Base
     User.current = nil
   end
 
+  # Save a lookup of the mrtg stats "user".
+  MRTG_USER_ID = 164054
+
   def try_user_autologin(user_from_session)
     if Rails.env.production? && request.remote_ip == "127.0.0.1"
       # Request from the server itself, MRTG needs to log in to test page loads.
-      login_valid_user(User.find_by(login: "mrtg"))
+      login_valid_user(User.find(id: MRTG_USER_ID))
     elsif user_verified_and_allowed?(user = user_from_session)
       # User was already logged in.
       refresh_logged_in_user_instance(user)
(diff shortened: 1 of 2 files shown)


nimmolo:
@JoeCohen This affects your refactor of `autologin` from 8 years ago. 

Do you happen to remember why you were reloading the @user right after loading it?

PR #1887, opened 2024-02-07
Revert "Turn off user site stats for the day"

mo-nathan:
Reverts MushroomObserver/mushroom-observer#1842

I need the user summary page to be working.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -29,7 +29,7 @@ class UsersController < ApplicationController
   def show
     store_location
     id = params[:id].to_s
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):

nimmolo:
Maybe this is a good time to ask -- 
should I make the stats section a lazy loaded panel that people have to click on to reveal?

In a separate PR of course... this one is fine. I just dread clicking on any user! The load time is heinous. 

The other idea I had was to use the cache the user stats for all users, for the week or maybe the day.  Since we now have the `refresh caches` script working, thanks to @pellaea, that could be easier to implement.

PR #1889, opened 2024-02-10
Use panels in show descriptions

nimmolo:
Purely graphic changes in the show pages for NameDescription and LocationDescription. 
- Uses the `panel_block` helper as used in show obs, for consistency and hopefully improved legibility. (screengrab below)
- Makes more use of icon CRUD links (details below)
- DRY these two pages, using generalized description helpers
- Move the top right "tabset" (action links for the present description) to icons in the top panel, or a link to the parent 

Screen Shot 2024-02-09 at 11 07 31 PM

___

Questions: 
- Could the top panel get a title bar like "Details" below, and if so is that title OK?
- Should the "Okay for export"/"Review status" stuff go in a panel footer of that top panel? (screengrab below)
- Shall I move some of the top-right "Tab links" for crud actions to the top panel as icons? (Edit, destroy, clone, merge, adjust permissions, publish)
 
Screen Shot 2024-02-09 at 11 08 06 PM

Screen Shot 2024-02-10 at 1 01 16 AM

Screen Shot 2024-02-10 at 2 05 44 AM


pellaea:
FWIW, I really don't care about name descriptions, and frankly wouldn't
vote against ripping them out altogether in the interests of simplifying
the code base.  Do you have my support, at least, for doing whatever you
think is right re: name descs.

(But obviously others may have differing opinions.)

JoeCohen (approved):
1. My very slight preferences, but do whatever you prefer.
- No title bar in the top panel. (That's just more noise on the page and in the template file.)
- If you decide to add a title bar, "Details" is fine.
- I'm okay with icons if help appears when you mouse over. Icons are cleaner than the textual link, but the meaning of some of the icons is obscure. (I cannot suggest better icons.) In the mockup, the icons are kind of small when they're in the title bar.
2. I'm opposed to ripping out the Descriptions. But @mo-nathan and I breifly discussed ways to simplify them (e.g., combining some of the field). That's a very low priority at the moment.
3. I obvously can't try this pr locallly because of my problems with the db.

mo-nathan:
I agree with Joe on all his points.  I'm going to try to get my local working today, but we'll see how it goes.  From my perspective improving Name Descriptions is really important for the Visual Group concept.  Once we figure out how to become the obvious choice for mushroom foray events, I believe the next thing we need to do is develop our community of folks writing name descriptions.  This is an area where iNat is pretty lacking and while you can often find stuff on the web with a Google search the content is very inconsistently written and often not appropriate for most of our community.  We can certainly leverage tools like Wikipedia, but I don't think we should just grab that content automatically.  iNat does that and it is often really bad.

nimmolo:
Heard. 

I am aiming to make all icons bigger and more legible once we get to update Bootstrap - but it's an entirely different icon library and i feel like it's not worth futzing with too much now. 

I will try to make the mouseover "tooltip" labels bigger in the meantime. They are hard to read for me.

PR #1890, opened 2024-02-10
Getting Ruby 3.3.0 working

mo-nathan:
Had to:
uncomment redis gem
run `bundle update`

Then after running `rails db:migrate` there were changes to db/schema.rb that I don't fully understand.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem("requestjs-rails")
 # turbo for partial page updates
 gem("turbo-rails")
 # redis for combining actioncable broadcasts with turbo_stream
-# gem("redis", "~> 4.0")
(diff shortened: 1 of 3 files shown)


nimmolo:
@mo-nathan - 
- I notice that your bundle update is updating `turbo-rails` here, which is brand new and a huge update, it likely has breaking changes. You'll need to lock it at version 1.5.x for now.
- We don't use `redis` yet, and it does not seem to be a dependency of any other gem, either. I believe it should stay commented out.

nimmolo:
@mo-nathan did you get everything working? should we close this PR?

mo-nathan:
Yeah, I think we're good here.  Thanks for the help!

PR #1891, opened 2024-02-10
Add collation for project_members to db/schema.rb

mo-nathan:


JoeCohen:
Would it work and (if so) would it better to do this as a migration?

mo-nathan:
I don't believe it really should be done with a migration.  Our production system and my local system already have the right collation.  I'm not sure why Nimmo and you (I think) have the wrong collation for that table.  There must have been something weird going on when you ran the migration for that table originally.  I'm pretty sure if you restored you database from a production snapshot or recreated the database using the current schema.rb it would just work.

JoeCohen:
The collation might actually have been right in my db but I manually
screwed it up.
Hard for me to tell right now because my local environment is effectively
toast.

Issue #1904, opened 2024-02-11
"Show Images" doesn't work correctly from observations page with query parameter


mo-nathan:
To reproduce:
Go to https://mushroomobserver.org/names/907
Click on "this name" on the right hand side
See observations of Hydnellum aurantiacum
Click on "Show Images"
See the most recent images on the website.  I expect to see all images of Hydnellum aurantiacum.

JoeCohen:
@nimmolo Perhaps another caching issue?

nimmolo:
At the moment, caching is only implemented on Observation indexes (the matrix_boxes) and the left navbar. 

This should help demystify: caching has to be manually implemented in a template or helper with the command `cache`, so you can search the code to find everything we're caching. Here's what it does:

`cache(:object) do html_helper end` will cache the html fragment generated by `html_helper` under a specially generated key in the cache database. The key is how the cache knows if it's fresh or not: it's a string that contains the current template directory, an identifier for the `object` if it's an AR object, and a generated "digest" (random-seeming string that will be changed if the template is changed, the object is updated, or rails is rebooted).

nimmolo:
Question: 

Can someone please explain to me what the point of these image links really is? I get that they _seem_ to make sense, but I'm trying to raise a question here: on a deeper level, they do not make sense to me, and are conceptually at odds with MO. On MO, as in life, we don't have observations of a species (unless we have conclusively keyed out everything), we have observations that _seem to belong to_ a species.

Image indexes look just like Observation indexes, but without the accompanying data. I don't see a legitimate non-admin use case for querying images without observation data. Any connection between an Image and a Name exists entirely through the contextual process of Naming an Observation (which has a place and time), and I would argue that a Mushroom Observer image divorced from its accompanying observation data is borderline meaningless, _except for the machine learning project_. 

When someone wants "images" of a name, what they really want is "Observations" of a name. We don't have "images of a name".  I'm proposing that all the "Images of..." links are adequately addressed by "Observations of..." links, and we should retire all the "images of" links.

mo-nathan:
Fair question.  The reason I want to be able to click on this link is to see all images in MO that are associated with observations that currently have this name.  The problem with just seeing the observations is that it will only show you one image for each observation, but many of the observations have multiple images.  If I am trying to get an idea of what the community has thought a name means, I find it more meaningful to see all the images rather than just the chosen thumbnails.

mo-nathan:
The fix looks good to me.

[...]

PR #1910, opened 2024-02-14
Add index keys to images controller

nimmolo:
"fixes" #1904. 

I believe the `q` and `id` query params may have been overlooked in Joe's DRY refactor of the Index actions.

Could use a test, but i'm not going to write it.


mo-nathan:
Thanks!  This looks good.

PR #1912, opened 2024-02-14
msql2 0.5.6

JoeCohen:
Bumps mysql2 version

I needed this because I could not install 0.5.5.
I got:
```ruby
bundle install
...
Installing mysql2 0.5.5 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
```


mo-nathan (approved):
Works fine for me.

Issue #1907, opened 2024-02-11
Name deprecations do not unbold names in cached matrix boxes on observation indexes


JoeCohen:
Sometimes light (correct), but sometimes bold (wrong).
Examples: 
- https://mushroomobserver.org/observations?pattern=name%3A%22Trichia+decipiens%22
- goto https://mushroomobserver.org/names/21206 
In the "About this taxon" panel, click "this name"

nimmolo:
@JoeCohen @pellaea 
To be honest, I don't understand what governs the boldness of name display. Is it `Textile.register_name`?

nimmolo:
Ah, I don't believe it's a Textile issue. It seems "Trichia decipiens" may be a very recently deprecated name, because in my db snapshot, it is accepted. If that's the case, then previously cached matrix_boxes will have it bolded, and matrix boxes pulled after the deprecation will have it not bolded. 

If it's worth fixing, we need to alter the controller action for name deprecations so that it "touches" all associated observations. In other words, the `deprecate` action needs to query all obs where the name is the consensus id, and `touch`, which changes the `updated_at` column on the obs.

For an observation matrix box, the cache only keeps track of the AR records of **the observation, its namings and votes**. Changes to other records associated with the observation will not currently invalidate the cache of the matrix box. 

In other words, anything else in the matrix box that we want to keep updated (like some kind of location name change, or user "login" change) will need to be altered so that the update actions explicitly `touch` the observation.

This may sound like a pain in the butt, but nothing has sped up our home page more than this caching. Maintaining that improvement will require changes like this.

nimmolo:
Thanks for checking it out, @mo-nathan. But heads up: invalidating the entire cache is much costlier than touching 1000 records in terms of server load. That's what I'm learning from all these experiments to speed up the site. The "way to just invalidate the relevant cache entries" is to `touch` them, it's not an expensive operation.

I want to be sure everyone has an appreciation of HOW MANY lookups and renders the cache is saving us. If you haven't examined your dev log, i will do a little TED-talk about it at the next meeting with show and tell.

nimmolo:
@JoeCohen - good question. 
Yes is the answer. All names whose status has changed will need to go into one batch in the query, and then they get `touched` in a single batch operation. 

That's why I nominate you to handle the PR 😁, you're more aware of Name logic. I also think everyone needs some practice writing cache- and query-friendly code.

Check the `Names::SynonymsController`: 
[...]

PR #1913, opened 2024-02-16
Disable cache for matrix boxes

nimmolo:
(Temporary) This PR immediately turns off caching of all `matrix_box`es.   

Observation `matrix_box` caches were not getting properly invalidated upon update of associated records, as pointed out by Joe in #1907 (`Name` deprecations not immediately propagated to cached `Observation` partials). 
___
The relevant AR changes that should cause a `matrix_box` refresh are not limited to `Name` deprecations:
- `Name` - could be affected by Author changes, too
- `Location` - changes to the `name`
- `User` - a change to someone's `login` handle should be immediately propagated

I've had a long, constructive discussion with Jason today about it, and we have a solution that doesn't involve manually `touch`ing the obs when the association is updated. We will need to alter the caching strategy for `matrix_box`, implementing "Russian Doll" fragment caching for the nested records.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1491,12 +1491,16 @@ class ApplicationController < ActionController::Base
 
   # If caching, only uncached objects need to eager_load the includes
   def objects_with_only_needed_eager_loads(query, include)
-    user = User.current ? "logged_in" : "no_user"
-    locale = I18n.locale
+    # user = User.current ? "logged_in" : "no_user"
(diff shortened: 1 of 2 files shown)


mo-nathan (approved):
Thanks!  The stuff in app/controllers/application_controller.rb is the key that I couldn't find yesterday.  Great to hear about the improved approach you and Jason have come up with.  I will do my best to review it whenever it's ready so we can get the cache going again.

PR #1917, opened 2024-02-17
Enhance Name external links

### LINKED ISSUE #1884: Add iNat & Wikipedia links to RH column of Name "About this taxon" panel

JoeCohen:
- Adds additional external search links to Name `About this taxon` panel
- Reorganizes the panel:
  - Adds a **Research Links** caption, indenting external search links under that caption.

While working on the NEMF checklist, I've been wishing I had links like this in MO. So I finally added them. Some (e.g. MushroomExpert) are also useful for beginners.
(I also have an idea about adding these and similar links to the Observation pages. I'll write that up as a discussion.)

### DIFF
--- a/app/helpers/object_link_helper.rb
+++ b/app/helpers/object_link_helper.rb
@@ -193,7 +193,7 @@ module ObjectLinkHelper
     links = users.map { |u| user_link(u, u.legal_name) }
     # interpolating would require inefficient #sanitize
     # or dangerous #html_safe
-    title + ": " + links.safe_join(", ")
+    title + ": " + links.safe_join(", ") # rubocop:disable Style/StringConcatenation
   end
 
   # Wrap object's name in link to the object, return nil if no object
--- a/app/helpers/object_link_helper.rb
+++ b/app/helpers/object_link_helper.rb
@@ -69,9 +69,12 @@ module ObjectLinkHelper
   end
 
   def mushroomexpert_name_search_url(name)
-    # Google it because mushroomexpert has no internal name search
-    "https://www.google.com/search?q=" \
-    "#{name.text_name.tr(" ", "+")}:www.mushroomexpert.com"
+    # Use DuckDuckGo see https://github.com/MushroomObserver/mushroom-observer/issues/1884#issuecomment-1950137454
+    name_string = name.text_name.
+                  sub(/ (group|clade)$/, "").
+                  tr(" ", "+")
+    "https://duckduckgo.com/?q=site%3Amushroomexpert.com+" \
+    "%22#{name_string}%22&ia=web"
   end
 
   def mycoguide_name_search_url(name)
--- a/app/helpers/object_link_helper.rb
+++ b/app/helpers/object_link_helper.rb
@@ -73,6 +73,16 @@ module ObjectLinkHelper
     "https://www.gbif.org/species/search?q=#{name.sensu_stricto}"
   end
 
+  def google_name_search_url(name)
+    if name.rank == "Group"
+      # require quoted name ss, optional group/clade/complex for best results
+      "https://www.google.com/search?q=%2B%22#{name.sensu_stricto}%22+" \
+      "%28group+OR+Clade+OR+Complex%29&"
+    else
+      "https://www.google.com/search?q=%2B%22#{name.sensu_stricto}%22"
+    end
+  end
+
   def inat_name_search_url(name)
     # omit `group`, else there are no hits
     "https://www.inaturalist.org/search?q=#{name.sensu_stricto}"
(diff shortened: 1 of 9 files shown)


app/helpers/object_link_helper.rb
```diff
@@ -60,15 +60,42 @@ def name_link(name, str = nil)
...
   # url for IF record
   def index_fungorum_record_url(record_id)
     "http://www.indexfungorum.org/Names/NamesRecord.asp?RecordID=#{record_id}"
   end
 
-  # url for Index Fungorum search. This is a general search.
-  # IF lacks an entry point that includes the name to be searched.
-  def index_fungorum_basic_search_url
-    "http://www.indexfungorum.org/Names/Names.asp"
+  # Use web search because IF internal search uses js form rather than a url
```
JoeCohen, line 80: @nimmolo: What do you think of the DuckDuckGo search results?
In particular, would other search engines yield a more useful result?
(DDG has a pop-up the first time, and may show some ads before the first real result. But other search engines also do something obnoxious.)

nimmolo: @JoeCohen I tried Google's "advanced search" syntax which is a little different - and it also returned no results for your example! Good idea trying DDG, seems like the way to go.

nimmolo: I would maybe consider the "tab" pattern for these, as in app/helpers/tabs. 

I've been using those to give links identifying classes for system tests. The pattern is an array: 
`if_link_tab = [link_text, url, html_options]` - there are examples that auto-generate a class name. 
then `link_to(*if_link_tab)` in the template.

mo-nathan:
Should there also be a vanilla "Google Search"?  Should there be a way to select different search engines (Bing, DuckDuckGo, etc.)?

mo-nathan:
The searches for "group"s doesn't work well.  Compare /names/58294 vs. /names/3914.  Maybe we shouldn't include " group" in group searches?

I'm not familiar with MycoGuide and it doesn't seem to have info at the species or even genus level (searches for "Tricholoma equestre" and "Tricholoma" are not at all compelling).  Why did this make the cut?

mo-nathan (approved):
Overall looks good and is a great improvement!  I added a few suggestions after playing around with it bit.  I leave it to @JoeCohen to decide which of those suggestions are worth implementing.

app/helpers/object_link_helper.rb
JoeCohen: @nimmolo: Credit for the DuckDuckGo idea goes to @mo-nathan. I would never have thought of it.
Bing also returns decent results, but includes a Chat thingy that's more active, and therefore mroe annoying, than DuckDuckGo's thingy.

JoeCohen: @nimmolo: Re "tab" pattern. I think that's what I'm doing (and it sure makes it easier to write the code). 
Maybe I'm misunderstanding the pattern or your comment.

JoeCohen:
> I think the "Google Images" link should go under "Research Links". It seems weird where it is now on the left side of that panel just above "Occurrence Map".

@mo-nathan: You're right. I moved the link.

JoeCohen:
@mo-nathan: TLDR: I removed the link to MyCoGuide.
Back story:
It's Patrick Leacock's (the [NAMA Voucher Committee](https://namyco.org/about/committe) Coordinator) site. It's useful for the NEMF name work.
I agree it's not complete or strong enough for present purposes. 
(Might be useful in the future for other purposes, e.g., for Observations near Chicago.)
>I'm not familiar with MycoGuide and it doesn't seem to have info at the species or even genus level (searches for "Tricholoma equestre" and "Tricholoma" are not at all compelling). Why did this make the cut?

app/helpers/object_link_helper.rb
nimmolo: Oh, yes. I missed the rest of what you were doing... 
I was looking at this on my phone and only caught the new "url" helpers in object_link_helper.

JoeCohen:
@mo-nathan: I added a vanilla Google Search. It would be cool to have a way to select different search engines, but I don't want to deal with that now. I think it would require one of:
- Adding individual links for a each of a bunch of engines. I don't like this at all because of the clutter.
- Something like a drop-down menu to select the engine. This is desireable in the long run. But I think this requires a form or js, neither of which I have time to deal with now. 
> Should there also be a vanilla "Google Search"? Should there be a way to select different search engines (Bing, DuckDuckGo, etc.)?

mo-nathan:
Thanks!  Looks good.  I wonder if there is someway to provide a link that uses the users preferred search engine.  I did a bit of search around and didn't find anything that seemed right.  I don't think the `<search>` tag or `<input type="search">` do what I'm thinking of.  There's probably some JavaScript thing that does something reasonable.

PR #1921, opened 2024-02-19
Added something to remove old unverified user accounts.

pellaea:
This just deletes users that are unverified and were created more than a month ago.

And the user-groups associated with them, and removes them from the "all users" group.

It assumes that there are no other references because they are unverified and shouldn't be able to create anything.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -737,6 +737,21 @@ class User < AbstractModel
       split(",").map(&:squish).compact_blank
   end
 
+  def self.cull_unverified_users(dry_run: false)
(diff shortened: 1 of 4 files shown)


JoeCohen:
@pellaea: Thanks so much for doing this. It will be nice to clean these out (current about 25% of the User's).
Would you have 5 minutes to update the documentation at the top of `script/refresh_caches`?
(I'd do it myself, but I don't understand all of the things that are run.)

pellaea:
Thanks, Joe, yes, I will update the comments.  Meant to but forgot.

mo-nathan:
Again this is great!

PR #1926, opened 2024-02-20
Fix broken database references

pellaea:
Adds yet another script to be run periodically, maybe weekly?, that checks for internal references within the database and corrects them.  For example, it checks for api_keys owned by nonexistent users and deletes them.

Specifically, it checks every single `belongs_to` reflection in the application, including polymorphic ones.  If it finds a record that belongs to a nonexistent record there are four options:

1. delete the record (most cases)
2. set the reference to null (e.g., users.image_id -- no need to delete the user, but we don't want it to point to a record that doesn't exist, so set it to null, which is permissible)
3. set the reference to zero (e.g., name_versions.user_id -- the user that made the change has apparently been deleted and orphaned this record; we don't want to delete the record, so instead set user_id to zero = admin just to hold the place)
4. just whine about it! -- that is, send an email to webmaster, e.g., there are a bunch of images with invalid licenses -- we should probably figure out why and fix this by hand

It also does a directory of app/models to create a list of all `belongs_to` reflections, and complains if the script is missing any.  This way when we inevitably create a new reflection and forget to update this script it should alert us within a week.

### DIFF
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -149,7 +149,6 @@ require("fastimage")
 #  license_history::    Accounting history of any license changes
 #                       (started using April 2012).
 #  quality::            Quality (e.g., :low, :medium, :high).
-#  reviewer::           User that reviewed it.
 #  num_views::          Number of times normal-size image has been viewed.
 #  last_view::          Last time normal-size image was viewed.
 #  transferred::        Has this image been successfully transferred to the
@@ -251,7 +250,6 @@ class Image < AbstractModel # rubocop:disable Metrics/ClassLength
 
   belongs_to :user
   belongs_to :license
-  belongs_to :reviewer, class_name: "User"
 
   has_many :copyright_changes, as: :target,
                                dependent: :destroy,
--- a/script/check_for_broken_references
+++ b/script/check_for_broken_references
@@ -1,0 +1,268 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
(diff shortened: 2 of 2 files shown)


mo-nathan:
This is great!  Thanks for doing this.

pellaea:
Thanks, Nathan.  What I neglected to mention is that we presently have
literally tens of thousands of broken records in the database.  I'd just
started by checking to see if there were any, with the thought that I'd fix
any that I found by hand.  I had no idea how widespread it was!  This
clearly needs a periodic task to keep an eye on it.  Granted most of the
existing broken records are harmless or would only cause rare crashes (e.g.
show_user on a user with a bogus image or location).  But still, might as
well clear all that cruft out.

mo-nathan:
Not surprised at all.  I know I found a bunch of weirdness when I dug into the image data for the ML stuff in 2022.  I'm trying to remember if anyone actually followed through on cleaning up what I found there in production.  I remember there being a bunch of orphaned images that probably should just get deleted as well as database inconsistency.

pellaea:
My guess is "no", because it's so hard to decide what to do with the
problems once we find them.  We always get bogged down at that point.  I
still have a script that's been waiting for a couple years for me to
convince myself that how it cleans up teh mistakes it finds is correct.
(Name.propagate_generic_classifications)

It's worth noting that *this* PR will not remove orphaned images.  I mean,
they *could* still be referred to in comments, descriptions or on external
sites, for example.  Not sure what the "right" way to deal with those is.

PR #1927, opened 2024-02-20
Switch server from unicorn to puma

nimmolo:
This PR is to help switch the production server from unicorn to puma, **without** turning on threading.
___
It follows the only complete guide I've found, but it's pretty old: 
[Migrating from unicorn to puma with no downtime](https://www.t27duck.com/posts/1-migrating-from-unicorn-to-puma-with-no-downtime)

Current preferred way is to run Puma via `systemd`, not `init.d`
Note that there is no file `/etc/systemd/system/unicorn.service` on production, what we have is `/etc/init.d/unicorn`
- [x] Add gem `puma` in this branch
- [x] Configure `/config/puma.rb` in this branch
- [ ] Configure `/config/puma-dev.rb` in this branch (in case Jason wants to run it locally)
- [x] Configure systemd service `/config/etc/puma.service` and copy to `/etc/systemd/system/puma.service` on production 
- [x] Alter our `/deploy.sh` in this branch _not_ to reload Unicorn
- [x] Modify `config/etc/nginx.conf`in this branch to point to new Puma socket instead of Unicorn 
Check example of `upstream` and `proxy_pass` [configs here](https://github.com/puma/puma/blob/master/docs/nginx.md)
```
  upstream puma {
    # fail_timeout=0 means we always retry an upstream even if it failed
    # to return a good HTTP response (in case the Puma master nukes a
    # single worker for timing out)
    server unix:/var/web/mo/tmp/sockets/puma.sock fail_timeout=0;
  }

# also remove or fix proxy_pass: 

    location @app {
      ...
      # this is where puma is listening
      proxy_pass http://puma;
```

Deploy this PR. Then:
- [x] Copy `/config/etc/nginx.conf` to `/etc/nginx/nginx.conf` 
- [x] `sudo systemctl start puma.service` so it's running alongside Unicorn
- [x] `sudo systemctl status puma.service` be sure it's running
- [x] `sudo systemctl reload nginx` (to listen to the Puma socket)
- [x] `sudo systemctl stop unicorn.service` (or maybe `sudo uni stop`, since there's no such thing as `unicorn.service`)
- [x] `sudo systemctl disable unicorn.service` (or maybe remove the init file in `/etc/init.d/unicorn`)
- [x] `sudo systemctl enable puma.service `
___
Reference:
[Benefits of Switching from Unicorn to Puma](https://tomkadwill.com/benefits-of-switching-from-unicorn-to-puma) 
[Puma docs](https://github.com/puma/puma/tree/master/docs): [Configuration](https://github.com/puma/puma/blob/master/docs/nginx.md) and [Deployment](https://github.com/puma/puma/blob/master/docs/deployment.md)
[How Do I Know Whether My Rails App Is Thread-safe or Not?](https://bearmetal.eu/theden/how-do-i-know-whether-my-rails-app-is-thread-safe-or-not/)
Do we want to consider using [Capistrano](https://github.com/capistrano/capistrano) for deployment?


mo-nathan:
Was just thinking about this due to the outage this morning.  Great to see progress here!

PR #1934, opened 2024-02-21
Remove class attributes from API

pellaea:
I don't think API was actually thread-unsafe, but why not remove its dependence on class attributes anyway?  Easy enough.


pellaea:
The coverage issue must be that this code is slightly longer, but the coverage is identical, so the percentage goes down infinitesimally.  I'm going to ignore it.

PR #1936, opened 2024-02-21
Remove class attributes from reports

pellaea:
And another one...


JoeCohen (approved):
@pellaea Thanks for doing this. 
May I merge it and deploy?

pellaea:
Definitely, thanks, Joe.

PR #1944, opened 2024-02-22
`run_level.rb` - convert class var to class ivar

nimmolo:
This `@@class_var` is not used outside the class, so it seems safe to convert to an ivar.


JoeCohen:
@nimmolo: I hadn't reviewed this because of the push test failure. 
I often just re-run the failing test in Github when I see a CI failure because of a system test. But Github isn't giving me that option for this failure.
Is that something you can try?
It would be nice to get your work into main.

PR #1954, opened 2024-02-22
IF search bad Name

### LINKED ISSUE #1925: Var. Research Links
These appear to be broken, at least in the following example:
https://mushroomobserver.org/names/113529
It's a variety, so maybe that's the issue.

JoeCohen:
One-line fix (plus a test) of a bug in the Nomenclature "IF web searcjh url. See #1925 
Fixes #1925

### DIFF
--- a/app/models/name/format.rb
+++ b/app/models/name/format.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Name::Format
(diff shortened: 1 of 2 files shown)


JoeCohen:
Deployed  deploy-2024-02-22-23-56 -> deploy-2024-02-22-23-56

PR #1966, opened 2024-02-27
Reverting a few changes from commit ca33c12c39d12237d115e51f54d07dd09…

pellaea:
…1969a97 that might have caused a performance drain.


JoeCohen:
Thanks Jason!

pellaea:
Nimmo and I are banging our heads trying to figure out what could possibly
be causing this.  Note that this just solves the main page load time, not
show obs.  And we have no idea why.  But at least we've got production back
to running the latest code.

PR #1972, opened 2024-02-28
Generalize runners

pellaea:
We used to have two runners: `script/run_script` and `script/run_rake`.  We just added `bin/puma`, sort of another (highly specialized) runner.  These are all hard-coded for `rvm`.

This PR removes all of these, and replaces them with a single `bin/run` which can run any of the three sorts of commands, using an environment variable (tentatively called `$RUBY_MANAGER`) to decide how to run it.

Note that the `rbenv` version is untested.  @nimmolo -- would you mind looking at that and/or testing it for me?

This should not affect dev env, because I don't think anyone uses any of these runners in dev.  But if you do, you will have to set the $RUBY_MANAGER env variable locally.  There is no default.


nimmolo (approved):

JoeCohen (approved):
Itrust you guys. 
While I understand what you said in the PR, the code itself is beyond me without a lot of effort.

mo-nathan:
It seems fine.  However, I guess I don't really see theh point.  It seems like we're just using chruby now in production and we should just stick with it.  I don't see RUBY_MANAGER getting set to anything else in the code base.  Am I missing some important use case?

mo-nathan (approved):

pellaea:
Thanks, @mo-nathan.  Yeah, it really only affects production, which will only use one of these ruby managers.  We settled on chruby because Nimmo and I couldn't get rbenv to work.  But given all the switching around during this process, it seemed like a good idea to generalize this.  If and when we have to switch again, this will hopefully prove at least somewhat useful! :)

PR #1975, opened 2024-02-29
Switch to `dartsass-rails` and `foreman`, `bin/dev`

nimmolo:
Our gem `dartsass-sprockets` works with `sprockets-rails`, this allows us to switch to `propshaft`.

> [!IMPORTANT]
> **Developers need to change the way we start the local server**
>
> First, `gem install foreman` — devs must install locally, it [should not be](https://github.com/ddollar/foreman/wiki/Don%27t-Bundle-Foreman) bundled in the Gemfile. 
>
> Then, instead of `rails s`, you run `./bin/dev`. That's it. 
>
> Explanation: `bin/dev` is a script that runs `rails s` and starts `foreman`. Unlike `sassc-rails`, `dartsass-rails` does not automatically watch our assets/stylesheets folders for changes. The "watching" is performed by the gem "foreman". 

> [!NOTE]
> In our CI, before running any tests that interact with the browser, build the Sass files by running:
>
> `bundle exec rails dartsass:build`


nimmolo:
@mo-nathan This may clear up the finicky solid queue situation in dev. We could try it either before or after the field slips PR.

If we try it after, we need to add this to the `Procfile.dev`
```solidqueueworker: bundle exec rake solid_queue:start
```

PR #1983, opened 2024-03-01
Update to Rails 7.1.3 and bundle update

nimmolo:
Also removes some version pegs that are obsolete.

Biggest change may be the `rack` version bump. It was held back by Unicorn.


mo-nathan:
Looks good!

nimmolo:
Here goes...

PR #1987, opened 2024-03-01
Comment out `on_worker_boot` block in `config/puma.rb`

nimmolo:
Theoretically obsolete.


nimmolo:
Closing as stale/superseded. The `on_worker_boot` hook this comments out has since been rewritten entirely -- it now calls `ActiveRecord::Base.connection_handler.clear_all_connections!` with a comment documenting that it's a deliberate fix for a real production crash (`Trilogy::SynchronizationError` when SolidCache's background expiry thread hits a connection inherited from before the fork). So the hook isn't obsolete, it's load-bearing -- merging this would mean re-deleting a fix that was added back after this PR was opened.

PR #1988, opened 2024-03-01
Store `SiteData` in a `UserStats` table

nimmolo:
The scope of this PR is to speed up the Show User page by caching the info it's querying (currently on every page load) in a new table. 

#### `UserStats`:
Each current metric has its own column in the new table, and the tallies are kept up to date with an existing callback (`update_contribution`) that updates the user's contribution scores on every (relevant) CRUD action. The callback has been modified to _also_ increment the relevant column in the UserStats table.

A new cron job refreshes the table so it doesn't get out of whack, using bulk upserts developed with @pellaea that refresh all user data for all users at once, in a single transaction. This is also the method that will need to be run to populate the table initially on the server, and for developers locally.

Per conversation with @mo-nathan, the plan is to reproduce current scoring exactly, out of consideration for MO users being possibly invested in scoring changes. 

In the future we can propose (and solicit proposals for) modifications to scoring via public discussions on MO. If that requires new columns and different scoring, we can handle it then. 

#### `Checklist`:
The show user page also shows tallies of `x species in y genera` from the user's "Life List", and it's also currently queried and calculated on page load. These tallies will have their own column in the `UserStats` table for quick access. @pellaea has refactored the query so it doesn't do n+1, and we're working on a query so a `Checklist` class method can provide access to bulk life list stats broken down for all users, in a single (double) query. Tallies would be updated by the cron job (but probably not the callback.)

#### `SiteData`:
For overall site data, we can similarly store this in its own table, but that's a separate PR.


mo-nathan:
Just took a quick glance at this.  I'd still consider collapsing the translations data unless it's a pain for some reason.  I don't see much value in break it out by language and it has issues if new languages get added.

pellaea:
+1

nimmolo:
Yea - maybe i'm the only one who likes it! 

But I do like the existing breakdown, because i'm clicking on the user to learn something about them as a person... Language says a lot sometimes.

Thoughts?

JoeCohen:
I also prefer the simpler approach.

pellaea:
I prefer breaking it out by language, but I don't think it's worth the
trouble.

mo-nathan:
I can see the information being interesting.  However, I don't think it makes sense to have to do a database migration to add a new language.  However, I guess it already requires a code release, so it doesn't really matter to me.  Just seems like it's more overhead if you have to remember to add a column to the appropriate table.  If it were me, I wouldn't bother, but if you want to do it, then I guess it's ok.  If you're going to do it, then document it somewhere.  I think there might already be something written about adding a new language which would be a good place for this, but i don't know that for sure.  I think Jason did most (all?) the language stuff.

pellaea:
Actually doesn't require a PR or migration to add a new language.  (It does
require inserting a record into the DB by hand, though.)

mo-nathan:
I was thinking of the <lang>.txt file that needs to get added to the code base.

pellaea:
Not to be argumentative, but only en.txt is checked in, all the rest is
kept in the DB.  `rake lang:update` creates the .yml and .txt files from
the database.

mo-nathan:
Oh, ok.  That's ringing a bell now.  I've worked in other systems where each language had it's own file and I think that's a pretty common practice, but it doesn't support the whizzy interface you created.

pellaea:
Yeah, I bet that's the typical way it's done.

nimmolo:
Ok. Thanks for all the feedback.  FWIW, the code breaking out the translation strings by language is _already written_; so for that aspect, at least, storing those figures in columns is not much more work. But I agree that remembering to add a migration to this table is a potential banana peel.

I wanted to ask about the alternative I put in my revised post above — I think those participating in this thread via email probably missed it. What do you guys think about this idea:

In addition to the one column of `translation_strings_versions` with the total count of translation strings, another column that's simply an array of the langs they've contributed to.  I could use it to display the langs as "flag badges" beside the display of the user's total translation count. 

I'm looking for some way to signal someone's language fluency, and that would satisfy the urge without storing a bunch more stuff; it also wouldn't necessitate new columns. 

I just feel like for non-English-native speakers, it's kinda key to be able to check if somebody is "your people", that you could maybe email them in Polish about something. (Of course, it's also true that anybody can write their languages in their profile.)

pellaea:
I think if you're going to store languages you should use the simpler
normalized table instead.

nimmolo:
@pellaea - That latter suggestion was not to store the whole language contribution counts, it was just to store an extra column with the locale abbreviations like this:

    user_stats.languages = "zh,es,pt"

pellaea:
@nimmolo - Understood, but it just seems like the complexity is growing
unnecessarily.  The normalized table can handle this kind of data better
and without adding any complexity.

nimmolo:
@pellaea's contributions to this PR have vastly sped up the queries that find user_data, both sitewide and per-user. Could not have done it without his SQL expertise.

One row of this table gives us everything we need to print all the current stats for a user.

Use:
```ruby
@user_stats = UserStats.find_by(user_id: @show_user.id)
@user_stats.names
@user_stats.observations
@user_stats.translation_strings
```

Notes: 
- Moved `user.bonuses` to the `user_stats` table... this is currently a serialized array column on User, but it seems to belong here, and the bonuses are printed in the table.
- Kept the language string breakdown, i guess because it already exists, i find it interesting, and it worked with the translation_string query refactor. Stored in a serialized hash column `languages` — to me, at least, this makes sense for this use case (throwaway derived data, keys not set in stone).
- Stores the *counts*, but not the actual taxa, of the user's "life list" (total taxa, species, genera) in another serialized hash column, `checklist` (also disposable data, may want to count other things in the future)

nimmolo:
@pellaea I put the user stats sitewide stats refresher in `refresh_caches`, but i'm not sure if that's the right frequency.

Note: devs will need to run `UserStats.refresh_all_user_stats` after this merges, to populate the table. Whoever deploys this will also have to run it on the production server.

JoeCohen:
I played around with this and it looks good.
Suggestion:
I am uncomfortable (on the SiteData page) calling MO users "Members". I suggest "Users". ("Member" has some legal overtones.)

nimmolo:
@JoeCohen Thanks for checking it out. I agree about "Members".

I haven't changed the SiteData page here though, that's for another PR. 
This PR only affects "show user".

JoeCohen:
When I tried the branch about 10 minutes ago, it included Members in Site
Stats.
But now I cannot reproduce that.
Was I picking up something from someone else's cache?

JoeCohen:
Sorry. I was looking at the wrong tab.
This branch does show Members.
Screenshot 2024-03-15 at 1 20 08 PM

nimmolo:
Good catch @JoeCohen 
Merging now...

PR #1989, opened 2024-03-02
Simplify and refactor `SiteData`, removing unused code

nimmolo:
Trying to get a handle on the confusing logic of this class, prior to writing a table to store the data.

1. There's a bunch of unused code for two planned but never implemented contribution metrics: `observations_with_voucher` and `observations_without_voucher`.  These cases required SiteData to account for much more complex logic. The logic somehow *was* implemented, but is always bypassed in code execution. I'm getting rid of it; it makes the code harder to read and refactor. Those fields can be rebuilt when the code here is simpler, if desired.
2. There is code for getting all data for all users, and for the whole site, that is never implemented, thankfully, because it crashes if run locally.
3. The main queries are composed in strings of raw SQL. This works, but I'm going to try to rewrite the whole thing in AR.


app/classes/site_data.rb
```diff
@@ -120,63 +116,36 @@ class SiteData
...
+  # FIELD_STATE_PROCS = {
+  #   observations_with_voucher: lambda do |obs|
+  #     obs.specimen && obs.notes.to_s.length >= 10 &&
+  #       obs.thumb_image_id.to_i.positive?
+  #   end,
+  #   observations_without_voucher: lambda do |obs|
+  #     !(obs.specimen && obs.notes.to_s.length >= 10 &&
+  #       obs.thumb_image_id.to_i.positive?)
+  #   end
+  # }.freeze
```
pellaea, line 149: Feel free to erase these completely, they'd be easy to reconstruct anyway.

app/classes/site_data.rb
```diff
@@ -204,15 +173,15 @@ def self.update_contribution(mode, obj, user_id = nil, num = 1)
...
   end
 
-  def self.calc_impact(weight, mode, obj, field)
+  def self.calc_impact(weight, mode)
     case mode
     when :del
       -weight
     when :chg
-      get_weight_change(obj, field)
+      0 # get_weight_change(obj, field)
```
pellaea, line 184: What is this case for?  Do we ever actually use it?

app/classes/site_data.rb
```diff
@@ -246,17 +217,11 @@ def self.get_applicable_field(obj)
...
-      end
-      old_field = get_applicable_field(obj_copy)
-    end
-    FIELD_WEIGHTS[new_field] - FIELD_WEIGHTS[old_field]
-  end
+  # This makes no sense without field procs!
+  # def self.get_weight_change(obj, new_field)
+  #   old_field = new_field
+  #   FIELD_WEIGHTS[new_field] - FIELD_WEIGHTS[old_field]
+  # end
```
pellaea, line 224: Good riddance!!!!

app/classes/site_data.rb
```diff
@@ -401,45 +354,52 @@ def load_field_counts(field, user_id = nil)
...
+    # else
+    #   @user_id = nil
+    #   users = User.all
+    # end
+
+    @user_id = id.to_i
+    user = User.find(id)
 
     # Prime @user_data structure.
     @user_data = {}
```
pellaea, line 407: This is fishy.  Either this is never called on multiple users, in which case @user_data doesn't need to be a hash, or it *is* called on multiple users, in which case shouldn't this is `@user_data ||= {}`?  As it is, it clobbers the hash each time and sticks a single user in it.

pellaea (commented):
Looking so much better!

app/controllers/info_controller.rb
```diff
@@ -58,10 +58,6 @@ def site_stats
     store_location
     @site_data = SiteData.new.get_site_data
 
-    # Add some extra stats.
-    @site_data[:observed_taxa] = Observation.distinct.count(:name_id)
-    @site_data[:listed_taxa] = Name.count
-
```
nimmolo, line 64: Moved into `SiteData`

PR #1990, opened 2024-03-03
Switch to conventionally named version tables

nimmolo:
Turns out the default behavior of the gem is to name the columns like "location_versions", which is what i believe we want. 
We've been overriding the default to make them "locations_versions". 

This PR removes the overrides and migrates to the default table names.

~I think this will break old migrations. I'll need to update them (?!)~


pellaea:
FWIW, I couldn't find any other places that need to change.

pellaea:
FWIW, I couldn't find any other places that need to change.

Oh, and we don't need to change old migrations, because if you rewind it will also change the names of the tables back so they will work.  They would only break if you tried to cherry pick one old migration and run that out of order on the modern state of the database.  But you would never do that.

nimmolo:
Check the test results here. The old migrations do not work on CI on this branch...

When I change the models to NOT override the default table name, it seems to immediately change the schema at the point of model instantiation, and the database doesn't find the old table names.

It is very strange!

nimmolo:
in other words, I think maybe this plug-in generates the tables without a migration, so there's no initial migration that the earlier migrations, modify those tables can refer to

pellaea:
But you need to write a migration!  Otherwise when you deploy this we will strand all of the old versions tables!

nimmolo:
right that's what's confusing. If I write the migration which I have a migration here it doesn't find the old tables. 

Maybe i need two quick PRs. 

Because if I change the set up of 'act_as_versioned' for say the Name model, to not use the custom table name, 'names_versions', but to use the default name, 'name_versions' it immediately sets up that table without a migration!

pellaea:
How did schema.rb change?

nimmolo:
That's what i'm saying. That is the mystery.

The gem seems to have changed the schema directly. I started writing a migration but I never ran it, because I noticed Schema had already changed.

pellaea:
That doesn't make a lick of sense.  How could it possibly know the name of
the old tables????

Okay, when exactly did... oh nevermind, I'll pull the PR myself and pay
close attention to when the tables miraculously get renamed.

JoeCohen:
1. I'm curious why we're doing this.
2. Would it make any sense to declare acts_as_versioned bankruptcy and switch to a maintained versioning gem. 
I'm pretty sure we'd have to abandon all the old versions and start from scratch. The last time I looked a this there wasn't a path from acts_as_versioned to a newer gem.
(I also wonder if acts_as_versioned is thread safe.)

pellaea:
You're right, Joe.  The reason we kept acts_as_versioned alive was because
it was so much work to migrate the old tables to the new ones.  Surely
that's important, right?  We aren't actually considering removing all
editing changes ever made on the site?

PR #2006, opened 2024-03-07
Fixed bug in "prev" / "next" on obs page.

pellaea:
I don't know how to test this.  There is already a test of this functionality.  The problem is the old code relied on `redirect_to` returning something truthy.  But this is not reliable apparently.  But reliable enough that tests were passing.

### DIFF
--- a/app/controllers/observations_controller/show.rb
+++ b/app/controllers/observations_controller/show.rb
@@ -22,11 +22,9 @@ module ObservationsController::Show
   def show
     pass_query_params
     store_location
-    case params[:flow]
-    when "next"
-      redirect_to_next_object(:next, Observation, params[:id]) and return
-    when "prev"
-      redirect_to_next_object(:prev, Observation, params[:id]) and return
+    if params[:flow].present?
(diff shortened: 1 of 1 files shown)


nimmolo:
Does redirect_to_next_object have a fallback if the object at :id has been deleted? (asking from phone, sorry). That was the issue, right?

pellaea:
Nope, that wasn't the issue.  The object existed.  The code in question was:

```
  def show
   pass_query_params
   store_location
   case params[:flow]
   when "next"
     redirect_to_next_object(:next, Observation, params[:id]) and return
   when "prev"
     redirect_to_next_object(:prev, Observation, params[:id]) and return
   end

   check_if_user_wants_to_make_their_votes_public
   check_if_user_wants_to_change_thumbnail_size
   ...
```

See how it is supposed to return after the redirect?  Usually it does.
Like in the tests it seems to do so every time.  But in the wild, it would
occasionally fail to do so, and then attempt to both redirect and the later
render the obs page, and that's when it failed.

-Jason

PR #2015, opened 2024-03-09
Observations map limits

JoeCohen:
- Limits the number of points + locations that are mapped to 10,000
See [MO Tech Meeting Notes 2024--02-22](https://groups.google.com/g/mo-developers/c/jaPO9ZwHbxg)

We get 1-2 Digital Ocean downtime notices/day that turn out to be caused by logged-in users' mapping all Observations. It's bad to have the site be nonresponsive, even for relatively short periods. 
And I'd rather not spend time needlessly dealing with those notices. See also [@mo-nathan's Slack post](https://mushroomobserver.slack.com/archives/C06FHR54EPQ/p1709574475483929) ("We should figure [out] a process for having someone look into these failures after they happen."

### DIFF
--- a/app/views/controllers/observations/maps/index.html.erb
+++ b/app/views/controllers/observations/maps/index.html.erb
@@ -5,6 +5,7 @@ add_tab_set(observation_maps_tabs(query: @query))
 
 objects = @observations
 objects += @locations if @locations.any?
+objects = objects.first(10_000)
 %>
 
 <%= make_map(objects: objects, # CollapsibleCollection


JoeCohen:
@nimmolo, @pellaea: 
- Is there a less expensive way of doing this than my one-liner?
- Can this be tested without creating >10K Observations (or other objects)?

nimmolo:
@JoeCohen 
This seems like this will truncate the query results in the view, rather than limiting the results and producing a quicker query.

I would think the controller actions that initiate these queries would be the place to limit the results.
`Locations::MapsController#index`
`Observations::MapsController#index`
`Names::MapsController#index`

pellaea:
Yes, I agree with @nimmolo.  For example, in `Observations::MapsController#index` you could limit the number of observations after the `restrict_query_to_box` statement on line 13.  I think you can add it in the `args` of `find_locations_matching_observations` line 39.  Try just adding `limit: 10000`.  Same thing with the other two controllers.  Pass the additional argument `limit: 10000` into `@query.results` or `@query.select_rows`.

Another thing to consider -- do we care *which* 10000 objects it chooses?  We can set the order in the same place: `order: "created_at DESC"`.  We could even randomize!  (For the record, I don't care for code that produces random results -- it makes it so much harder to test!)

JoeCohen:
Closed in favor of #2018

JoeCohen:
Thank you both. That's exactly the kind of advice I needed.
I've closed this PR in favor of Nathan's #2018, which limits the results
returned by the query.

PR #2018, opened 2024-03-09
Limit the number of observations that can get mapped

mo-nathan:
Also includes a fix for config/puma.rb so it should work better with local systems.


nimmolo:
Looks good, but I feel like locations/maps and names/maps may need the same limits. I've noticed these showing up in the logs too.

Thanks for these better puma defaults.

Issue #1844, opened 2024-01-21
Constraint Violations


JoeCohen:
A page, linked from the Project page, which
- Lists, in abbreviated form, the Observations which violate the Project's Constraints 
- For each violation, highlights the parameter(s) that caused the violation
- Allows removal via checkbox

Wire frame: (updated 2024-03-09) (needs submit button)
<img width="905" alt="Screenshot 2024-03-09 at 7 53 43 AM" src="https://github.com/MushroomObserver/mushroom-observer/assets/45460/2ad918cf-87cd-4072-8ebb-fb04708ec45f">

### Tasklist
- [x] Test and create link on Project page to violations page
- [x] Test violations index
- [x] Create violations page content
- [x] Fix Project page Observations button going to wrong place (goes to all Observations, rather than Observations for that project.)
- [x] Rearrange violations table (obs name, number, user, under "Constraints")
- [x] highlight individual parameter violations with css
- [x] Ensure lon styling proper if Proj.west > Proj.east
- [x] Highlighted Constraints button instead of flash warning for Show project page
- [x] Refactor Project#violates_date_range? using (start_date..end_date).cover?(observation.when)
- [x] Better top part of page? Title Constraint Violations., next line: Project: <project name>
- [x] Test that compliant Obs not listed in violations index
- [x] Hide lat/lon for unauthorized users
- [x] Add conditional checkboxes for Obss; condition: current user is project admin || created obs
- [x] #2022
- [x] Add `update` action
- [x] submit button(s)? Maybe "Remove checked Observations" conditioned on their being removable observations; Show Project; Maybe just "Submit"
- [x] Add instructions. (e.g., "Check Observations to be removed from Project."
- [x] Try moving app/helpers/table_helper.rb#violation_headers et seq. to a different helper
- [x] Font size same as Name, Observation stuff
- [x] If can't chack anything, hide help_text, change button
- [x] PR
- [x] ? Project vs Observation
- [x] ? Better name for "Violations"?
- [x] ? Better namespace for "Violations"?
- [x] ?`edit` vs `index`
- [x] ? how much code to include in form
- [x] ? font size same as left navbar
- [x] ? bg highlite: is there a style that works: with any User theme?  with screen width <= 766 px?
- [x] Change violations edit to index per https://github.com/MushroomObserver/mushroom-observer/pull/2021#discussion_r1519142331
analogous to 
```ruby
    observation_namings GET  /observations/:observation_id/namings(.:format)
[...]

PR #2021, opened 2024-03-11
Show Project Constraint Violations

JoeCohen:
- Adds a highlighted Constraint Violations button instead of flashing each time the Project is displayed.
- Lists a Project's Constraint Violations;
- Highlights the Violations' cause(s);
- Lets Project admins (and Observation creators) remove offending Observations from the Project.
- Delivers Issue #1844
- Improves Project.rb code re violations

Sample partial screenshots
Project (with highlighted Violations button)
Screenshot 2024-03-14 at 5 07 17 PM

Violations -- trusted Admin's view:
Screenshot 2024-03-10 at 8 37 38 PM

(Untrusted) non-member views:
Screenshot 2024-03-10 at 8 44 05 PM


JoeCohen:
@mo-nathan, @nimmolo 
Should this instead be under Observations? (The Violations are all Observations.)

JoeCohen:
@mo-nathan, @nimmolo 
Is there a better (and reasonable short) name than "Violations"?

JoeCohen:
@mo-nathan, @nimmolo
Should the actions be in the parent controller (Projects) rather than Projects::Violations?

JoeCohen:
@nimmolo, @mo-nathan 
What should be the action names? Because there was a form involved I settled on edit/update. Alternatives include `index`, or something(s) unique.

app/controllers/projects/violations_controller.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# Show and remove non-compliant Observations from a Project
+module Projects
+  # Actions
+  # -------
+  # edit (get)
+  # update (patch)
```
JoeCohen, line 8: @nimmolo, @mo-nathan 
What should be the action names? Because there was a form involved I settled on edit/update. Alternatives include `index`, or something(s) unique.

app/assets/stylesheets/mo/_utilities.scss
```diff
@@ -609,6 +609,10 @@
   text-indent: 0;
 }
 
+.violation-highlight {
+  background-color: orange;
+}
+
```
JoeCohen, line 615: @nimmolo: How can I style the Constraints Violation button so that it works with:
- any User theme? 
- screen widths <= 766 px?
What I've done here isn't satisfactory.

app/views/controllers/projects/show.html.erb
```diff
@@ -1,11 +1,18 @@
 <%
 add_project_banner(@project)
 @container = :wide
+violations_count = @project.count_violations
+violations_button_class =
+  if violations_count.zero?
+    "btn btn-default btn-lg"
+  else
+    "btn btn-warning btn-lg"
+  end
```
JoeCohen, line 10: @nimmolo See [earlier question about styling this button](https://github.com/MushroomObserver/mushroom-observer/pull/2021/files#r1519139794).

app/views/controllers/projects/violations/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+<div class="smaller">
+
+<%= help_block(:div, violations_help_text(@project)) %>
+
+<%=
+  form_with(
+    model: @project, url: project_violations_path, method: :post
+  ) do |form|
+%>
```
JoeCohen, line 18: @nimmolo: Should stuff above be included in the form?

app/views/controllers/projects/violations/edit.html.erb
```diff
@@ -0,0 +1,31 @@
+<%
+  add_page_title(:violations_edit_title.t)
+  @container = :wide
+%>
+
+<h4 class="pb-3">
+  <%= :PROJECT.l + ": " + link_to_object(@project).t %>
+<h4>
+
+<div class="smaller">
```
JoeCohen, line 10: @nimmolo: How do I make the font size for the rest of the page -- including the footer -- the same as the font size on all our other pages?
What I did here is unsatisfactory.

JoeCohen:
I will deal with coverage issues tomorrow.

app/controllers/projects/violations_controller.rb
nimmolo: The way I've been doing (re-doing) these, the page that shows the list of all (of the violators) would be an `index` action. 

It's debatable of course — we have usually thought of these as "edit" forms in the past, but Rails CRUD `edit` routes require an ID, and for me it feels like more of a filtered `index`. I might call this a "Project::ObservationsController".

If you have buttons or checkboxes with submit, the `form` should submit to an `update` action, though, in my opinion.

JoeCohen: @nimmolo: Thanks for the insightful feedback. I'll conform to how you're doing it. 
It is a kind of filtered index, but the template differs from a normal index.

JoeCohen:
Converted to a Draft PR until I finish moving things around as suggested by @nimmolo at https://github.com/MushroomObserver/mushroom-observer/pull/2021#discussion_r1519142331

app/controllers/projects/violations_controller.rb
JoeCohen: @nimmolo I'm still thinking about how to do this?
Can you point me at an example where you used index in a namespaced controller?
Also, I don't see a way around requiring an `:id` (or equivalent); the page doesn't list globally all violations,  but rather all the violations of one project.

nimmolo: Oh - yes, even when you do it as an index, the CRUD route will need a `project_id`. 

(A good example of using an index with links for new/edit/create/update is the Namings table in show observation - that has its own route, Observation::NamingsController#index. If you go to /observations/:observation_id/namings, you should get the table all by itself.)

I was thinking, in the event you were making this an `#edit` action instead of an index, the Rails "resource route" would want an `observation_id` by default, because it would assume it's a page for editing a single observation belonging to a Project (like editing a Description belonging to a Name).  

In a sense maybe it is, but to me this is sort of more "bulk editing" project observations; it still seems like an index with a direct link to an update action that can handle multiple observation ids. 

Anyway i don't feel too strongly about it being an index.

app/assets/stylesheets/mo/_utilities.scss
nimmolo: Thinking about this. Need a bit of time to delve into those CSS variables

app/views/controllers/projects/violations/edit.html.erb
nimmolo: I kind of like the help block being outside the form element. 

I'd give the form an HTML ID though, for testing, maybe "project_violations_form", and that should allow you to skip the inner div#violations (not sure)

app/assets/stylesheets/mo/_utilities.scss
JoeCohen: @nimmolo: I've re-thought this, instead of changing the background, I'll use a colored border.
- It should be easier to find a non-enauseating color;
- It should be easier to find a color that works with all themes; 
- Perhaps it will solve the problem of the theme not working at small screen widths.

nimmolo: That sounds good... i think you can use "border-danger" and it should contrast in all themes, but i'm not sure.

app/views/controllers/projects/show.html.erb
nimmolo: This seems right, though, to me. What's happening with it at smaller screens that's not good?

(Sorry I have migrations on my UserStats branch that are making it kind of involved to switch branches right now. Should be done soon!)

app/views/controllers/projects/violations/edit.html.erb
nimmolo: Hmm. IIRC many panels on other pages use a class like "name-text" to make the text smaller, and the footers you're thinking of probably have a class like that on them too. 

It could also be there's a text size set on `panel` itself, but i don't think so.

app/assets/stylesheets/mo/_utilities.scss
JoeCohen: Thanks for the suggestion. I'll have to dig into the styling some more.
"border-danger" doesn't seem to work on its own.
I finally ended up changing the text color.
Resolving conversation.

app/views/controllers/projects/violations/edit.html.erb
JoeCohen: Thanks for suggesting that I `id` the form rather than the div. It's done and I will resolve the conversation.

app/views/controllers/projects/violations/edit.html.erb
JoeCohen: Everything on the page -- not just what's included in `app/views/controllers/projects/violations/index.html.erb` was too big. (That's why I made the div `smaller`.) For example, the Edit Translations footer  is too large.
Screenshot 2024-03-14 at 7 41 38 AM

I'll try to figure out why that happens.

JoeCohen: Problem: I messed up the `<h4>` closing tag -- I omitted the slash. 
Interesting that there wasn't a notice about this in VS Code or the Chrome developer's console.
In any event, it's fixed. So I will be Resolving coversation.

app/views/controllers/projects/violations/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<%
+  add_page_title(:violations_edit_title.t)
+  @container = :wide
+%>
+
+<h4>
+  <%= :PROJECT.l + ": " + link_to_object(@project).t %>
```
JoeCohen, line 7: @nimmolo:
`link_to_object` doesn't work -- there's no href.  I think I saw a similar issue raised by @nathan. 
Here the line renders as:
```html
 <a>First Fantastic Fungal Friends Foray Festival!</a>
```
In this PR, I'll probably replace the link with the project name. (I no longer need the link; Submit/Show Project button makes the link redundant.)

app/views/controllers/projects/violations/edit.html.erb
nimmolo: I know what you mean. I like using the tag helpers, like `tag.h4`, partly for that reason. There's no linting for erb, but there is for pure Ruby.

app/views/controllers/projects/violations/index.html.erb
JoeCohen: Conversation resolved. I removed link_to_object, used `tag.h4` instead of html

db/schema.rb
```diff
@@ -23,7 +23,7 @@
 
   create_table "articles", id: :integer, charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
     t.string "title"
-    t.text "body", size: :medium
+    t.text "body", size: :long
```
mo-nathan, line 26: What's the source of this change (and similar ones below)?  I'm not seeing it in any migration or in current production.

app/helpers/projects_helper.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+  end
+
+  def violations_help_text(project)
+    :form_violations_help.l if project.violations_removable_by_current_user?
+  end
+
+  def violations_submit_text(project)
+    if project.violations_removable_by_current_user?
+      :SUBMIT.l
```
mo-nathan, line 38: I think it would be better if this said "Remove Selected" since that's really the action it will take.  There's a comment at the top but that's pretty remote from the button.

mo-nathan: Also, if there are no violations it would be better if it said "Show Project".

mo-nathan (changes requested):
I think the db/schema.rb issue should get cleaned up.  Up to you if you want to follow through with my comments on the "submit" button text.

mo-nathan:
I was playing with it and I wanted to get straight back to the project from the constraint page without clicking "Submit".  I'm thinking the "Project: North Falmouth Fungi" bit at the top of the page should be a link back to the Project#show page.

db/schema.rb
JoeCohen: Resolving conversation.

mo-nathan (approved):

Issue #1947, opened 2024-02-22
MO Field Slip Concept


mo-nathan:
### Discussed in https://github.com/MushroomObserver/mushroom-observer/discussions/1896

Add FieldSlip class with CRUD.  FieldSlips can be associated with at most one Observation.  Can be associated with a Project. I don't think so since Observations can be associated with multiple Projects.  You really want at most one "special" project for a VoucherSlip.
- Project "code", e.g., NEMF2024, TMF-24, etc.
- FieldSlip "code" usually associated with Project, but CRUD allows any unique ID.  Code always has a trailing number that can be left zero padded, preceeded by a "-".  The part before the "-" is the "project code".
- FieldSlip can exist without an Observation, but then can create an Observation from that FieldSlip.
- FieldSlip code "GET" should first go to the Observation if there is one, then to FieldSlip show page if it already exists, or the FieldSlip create page if it doesn't exist.
- FieldSlip create page should derive the Project if possible from the code.

mo-nathan:
- [ ] Create model with observation and project association.
- [ ] Add foray code to projects.

mo-nathan:
Settled on FieldSlip as the model name.

PR #2024, opened 2024-03-12
Field Slips

mo-nathan:
This provides the key functionality needed for Field Slips.  See #1947 for the issue and #1896 for a more detailed discussion.

This PR delivers the functionality where there is a route that accepts a field slip code.  The /qr/:code route is provided for simplicity and brevity.  To test you probably want to update some projects with field slip prefixes.  For example, if you're working with a recent database dump you could add the prefix '6F' to the [First Fantastic Fungal Friends Foray Festival!](http://localhost:3000/projects/336) or 'NFAL' to [North Falmouth Fungi](http://localhost:3000/projects/334?q=1pkXE).  You can then exercise the new functionality with URLs like:

localhost:3000/qr/6F-0001
localhost:3000/qr/NFAL-0001

The #1947 has a checklist of functionality that I added as part of the PR to get to the point where it feels like a cohesive whole.

If you want to try it starting from a QR code, you can create QR codes here: https://www.the-qrcode-generator.com/


app/views/field_slips/_form.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <h2><%= pluralize(field_slip.errors.count, "error") %> prohibited this field_slip from being saved:</h2>
+
+      <ul>
+        <% field_slip.errors.each do |error| %>
+          <li><%= error.full_message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
```
nimmolo, line 13: I think in this situation we've been using `<div class="alert alert-danger">` (instead of `<div style="color: red">`)

app/views/field_slips/_form.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  <div>
+    <%= form.label :project_id, style: "display: block" %>
+    <%= form.number_field :project_id %>
+  </div>
+
+  <div>
+    <%= form.label :code, style: "display: block" %>
+    <%= form.text_field :code %>
+  </div>
+
```
nimmolo, line 28: The helper `text_field_with_label` should produce the Bootstrap markup here that we use elsewhere. 

We may need a new `number_field_with_label` helper to do that for number fields.

app/views/field_slips/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<p style="color: green"><%= notice %></p>
+
+<%= render @field_slip %>
+
+<div>
+  <%= link_to "Edit this field slip", edit_field_slip_path(@field_slip) %> |
+  <%= link_to "Back to field slips", field_slips_path %>
+
+  <%= button_to "Destroy this field slip", @field_slip, method: :delete %>
+</div>
```
nimmolo, line 10: Idk if you'd want the index field slips to also show these "edit/destroy" links per field slip, but it might be possible to pass a "context" local, and put those icon-links, plus a "back to index" link `if context == :show`... Something like that.

Can provide syntax for the icon links if you want to do it that way.

mo-nathan:
@nimmolo Thanks for the comments.  Happy to follow your suggestions.  Most of the code is actually from `rails generate scaffold`.  I was just trying to get something working as a starting point.  Did you see my comment in #general in Slack?  I'm trying to figure out how to make the matrix_box in app/views/field_slips/_field_slip.html.erb look reasonable instead of ultra skinny.  I'm guessing there's some reasonably simple way to do that.  I played around with it a bit but I'm hopeless with front end stuff.

app/views/field_slips/_field_slip.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <%= link_to_object(field_slip.project) %>
+    <br/>
+    <strong>Code:</strong>
+    <%= field_slip.code %>
+  </p>
+
+
+  <%= render(partial: "shared/matrix_box",
+             locals: { object: field_slip.observation }) %>
+</div>
```
nimmolo, line 16: I think this is all you need here. (The row width might need adjusting.)
This is from collection_numbers#show
```ruby
<ul class="row list-unstyled mt-3">
  <%= render(partial: "shared/matrix_box", locals: { columns: "col-xs-12" },
             collection: @collection_number.observations, as: :object) %>
</ul>
```

mo-nathan: Found a solution based on the herbarium record that is working well.

app/views/field_slips/_form.html.erb
mo-nathan: Adopted.  Thanks!

app/views/field_slips/show.html.erb
mo-nathan: I don't think this is needed so I'm not going to sweat it.

test/system/field_slips_test.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "application_system_test_case"
+
+class FieldSlipsTest < ApplicationSystemTestCase
+  setup do
+    @field_slip = field_slips(:one)
```
JoeCohen, line 7:
```suggestion
    @field_slip = field_slips(:field_slip_one)
```
Fixes 
```ruby
ERROR FieldSlipsTest#test_visiting_the_index (0.75s)
Minitest::UnexpectedError:         StandardError: No fixture named 'one' found for fixture set 'field_slips'
            test/system/field_slips_test.rb:7:in `block in <class:FieldSlipsTest>'
```

mo-nathan: Weird.  The tests pass fine for me locally without this change which I totally don't understand.  Thanks for fixing it.

mo-nathan: Oh, I thought you pushed it already.  I just did the same change locally and have pushed it.

mo-nathan:
I don't understand the errors about :SpeciesList not being a language tag.  I don't see where it ever gets used, but I added it and we'll see if it resolves the issue.

JoeCohen:
`localhost:3000/qr/6F-0001` Last Observation is a surprise.
I expected it to show my last Observation.
Instead it's showing the Obs at the top of the Activity Feed, which:
- was created 2013-06-23, 
- updated 2024-03-20 06:34:14 PDT (-0700)
- created by Vlad Lekach (laddycans)
- for which I proposed a Name
- is the local copy of https://mushroomobserver.org/137413
Screenshot 2024-03-23 at 12 46 18 PM

JoeCohen:
My "fix" just exposed more issues. Sorry. (I forgot to re-run the system tests.)

mo-nathan:
Looks like I didn't remember how to run the tests in test/system/.

app/views/field_slips/new.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>New field slip</h1>
```
JoeCohen, line 1: 1. Consider a different title; I think many (most?) people will confuse a physical slip with a FieldSlip object, and wonder why they're making a "new" slip when they have one in front of them. 
But I don't have a great suggestion. Maybe "Link Observation to Field Slip".
2. Whatever the title, use a "translation". (Even if you don't anticipate use in forays outside the USA, some foray participants -- particularly at NAMA -- may not be native English speakers.)

app/helpers/field_slips_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module FieldSlipsHelper
+  def last_observation
+    return unless User.current
+
+    ObservationView.last(User.current)
+  end
+end
```
nimmolo, line 9: This is potentially a generally useful helper, but just remember that all helpers are in a global "helper namespace". If it's just for field slips, you may want a more specific name.

nimmolo:
> `localhost:3000/qr/6F-0001` Last Observation is a surprise. I expected it to show my last Observation. Instead it's showing the Obs at the top of the Activity Feed, which:
> 
> * was created 2013-06-23,
> * updated 2024-03-20 06:34:14 PDT (-0700)
> * created by Vlad Lekach (laddycans)
> * for which I proposed a Name
> * is the local copy of https://mushroomobserver.org/137413

From checking through the new method in `ObservationView#last_observation`, I believe @mo-nathan has it pulling the _last viewed observation by the current user_, and that may not be the expected obs.

JoeCohen:
@nimmolo is right. field_slip#new is showing the last observation viewed by the current_user.

JoeCohen:
The `qr` url with Create Observation creates two Observations.
The first includes the field slip #, BUT omits images, the Project, Namings.
Screenshot 2024-03-23 at 2 12 56 PM

The second is a more-or-less normal Observation, but omits the Field Slip.
Screenshot 2024-03-23 at 2 14 49 PM

app/views/field_slips/_form.html.erb
nimmolo: Glad it helped. I think you can drop the wrapping `<div>`s too. The Bootstrap form element classes will handle it.

JoeCohen:
PS: There's no log for the first Observation.

mo-nathan:
@JoeCohen Very interesting!  I know I manually tested after just giving a location and it seemed to work, but I didn't try adding images.  I'll look at it after I get back from a walk.

JoeCohen:
I just did the same thing, this time by scanning a QR code and not adding images.
Very slick and great to see it in action.
Unfortunately, same result: 2 Observations. 1st has FieldSlip but nothing else; 2nd has everything but the FieldSlip#.

JoeCohen:
@mo-nathan: I think it's a bug in constraint violations (or a bad interaction with that).
If I create an Observation that complies with the Project constraints, the above problem doesn't happen.
Each example I created above had a constraint violation, and I overrode the project constraints.

JoeCohen:
I pushed a short (failing) integration test exposing the bug. You can run it with:
`rails t test/integration/capybara/field_slips_integration_test.rb`
Sorry I didn't get any further.

If -- without using Field Slips -- I create an Observation, fill in the fields with something that violates project constraints, and click Create,
it: 
- redisplays the create_observation form with a flash warning ("This observation violates the constraints of one or more selected projects")
- does NOT create an Observation until I again click Create.

But if I scan a field slip, click Create New Observation, fill in constraint-violating info, then
- it redisplays the create_observation form, gives me the flash warning,
- BUT immediately creates an Observation (apparently including the FieldSlip, but ignoring namings, images).

app/helpers/field_slips_helper.rb
mo-nathan: I think this could be helpful in other places and it is not specific to FieldSlips.  I evolved it to create ObservationView.last(User) which might be a better standard to use than having it as a helper.  I don't have a strong opinion here.

app/views/field_slips/new.html.erb
mo-nathan: Interesting thought.  Thanks for catching the missing translation.  This naming came from the scaffold.  I generally need to fix up titles on the various pages.

mo-nathan:
To do:

- [ ] Fix double observation bug
- [ ] Fix titles (include a rename for the "New field slip" header)
- [ ] Fix system tests

mo-nathan:
> > `localhost:3000/qr/6F-0001` Last Observation is a surprise. I expected it to show my last Observation. Instead it's showing the Obs at the top of the Activity Feed, which:
> > 
> > * was created 2013-06-23,
> > * updated 2024-03-20 06:34:14 PDT (-0700)
> > * created by Vlad Lekach (laddycans)
> > * for which I proposed a Name
> > * is the local copy of https://mushroomobserver.org/137413
> 
> From checking through the new method in `ObservationView#last_observation`, I believe @mo-nathan has it pulling the _last viewed observation by the current user_, and that may not be the expected obs.

This is a really interesting case.  I can see an argument for limiting it to observations the current user created, but I think there is a legitimate use case where someone like a project admin might want to create a field slip object that the user creating the observation didn't know to create.  I'm hoping the use of the last observation feature is more of a rare power user thing and if they legitimately want to create a field slip for someone else's observation I think they should have a way to do that (e.g., a project admin sees an observation that should have been included that has a photo of the field slip, but the field slip link is on the observation).  Maybe I should add a comment on the page making it clearer that this is the last observation you viewed and if you view another observation in another window and reload the page or rescan the same QR code after visiting the observation you care about, then it will update.  Does that seem reasonable?

mo-nathan:
> I pushed a short (failing) integration test exposing the bug. You can run it with: `rails t test/integration/capybara/field_slips_integration_test.rb` Sorry I didn't get any further.
> 
> If -- without using Field Slips -- I create an Observation, fill in the fields with something that violates project constraints, and click Create, it:
> 
> * redisplays the create_observation form with a flash warning ("This observation violates the constraints of one or more selected projects")
> * does NOT create an Observation until I again click Create.
> 
> But if I scan a field slip, click Create New Observation, fill in constraint-violating info, then
> 
> * it redisplays the create_observation form, gives me the flash warning,
> * BUT immediately creates an Observation (apparently including the FieldSlip, but ignoring namings, images).

Thanks for the test!  Turns out the fix was quite simple.  I was ignore the success flag.  I also need to pass the field_code param to the next page.

JoeCohen:
I banged on field slips for a while. 
A possible issue: inconsistent relation among FieldSlip, Observation, Project.
Here's what I did:
- Added a FS prefix to an existing project. (In my case I added `SARCO` to http://localhost:3000/projects/337.)
Screenshot 2024-03-24 at 10 07 42 AM

- Created an Observation, **which was not in the Project**
- Pasted http://localhost:3000/qr/SARCO-0001 into the address bar, selected Last Observation (the one I just created). 

Result:
- Created a FieldSlip, correctly relating it to the Observation and the Project.
Screenshot 2024-03-24 at 10 12 24 AM

- Updated the Observation, adding the FieldSlip #, but not the Project
Screenshot 2024-03-24 at 10 18 17 AM

- Did not add the Observation to the Project.

app/controllers/observations_controller/new_and_create.rb
```diff
@@ -102,6 +112,7 @@ def create
     success = false if @name && !@vote.value.nil? && !validate_object(@vote)
     success = false if @bad_images != []
     success = false if success && !save_observation(@observation)
+    update_field_slip(@observation, params[:field_code]) if success
```
JoeCohen, line 115: Suggestions:
1. Put this line with the other things that get executed `if success`. (Result: one fewer conditional.)
2. Consider a guard clase.

The net result would be something like this:
```ruby
    success = false if success && !save_observation(@observation)
    return reload_new_form(params.dig(:naming, :reasons)) unless success

    # Once observation is saved we can save everything else.
    @observation.log(:log_observation_created)
    update_field_slip(@observation, params[:field_code])
    ...
    redirect_to_next_page
```

nimmolo:
Attempted some improved markup for the form on #2063 - it's a PR off this branch

app/controllers/observations_controller/new_and_create.rb
mo-nathan: Nice!

mo-nathan:
> I banged on field slips for a while. A possible issue: inconsistent relation among FieldSlip, Observation, Project. Here's what I did:
> 
> * Added a FS prefix to an existing project. (In my case I added `SARCO` to http://localhost:3000/projects/337.)
> 
> Screenshot 2024-03-24 at 10 07 42 AM
> * Created an Observation, **which was not in the Project**
> * Pasted http://localhost:3000/qr/SARCO-0001 into the address bar, selected Last Observation (the one I just created).
> 
> Result:
> 
> * Created a FieldSlip, correctly relating it to the Observation and the Project:
> 
> Screenshot 2024-03-24 at 10 12 24 AM
> * Updated the Observation, adding the FieldSlip #, but not the Project:
> 
> Screenshot 2024-03-24 at 10 18 17 AM
> * Did not add the Observation to the Project.

Interesting case.  I see what you're driving at.  Howevver, I am imagining that observations associated with field slips that are not associated with the same project to be reasonable common in the long run.  The case I'm thinking of are folks who go to events that are using MO field slips and walk away with a bunch of them.  In this case it seems reasonable to associate the field slip with the project (due to the prefix) and this could be useful once we have forms associated with projects, but if the observation doesn't meet the constraints of the project, then it's legit for it not to be added.  I guess it would be fine to add it to the project if it meets the contraints, but not if it doesn't with appropriate flash messages.

mo-nathan:
Resolved last observation getting added to project.

PR #2028, opened 2024-03-15
Paginate Contributors index

nimmolo:
This page loads very slowly because the results are vast, and not paginated.
Here, it's paginated. 

Adds a new query flavor for `User` `by_contribution` so we can call `show_index_of_objects` to get this query order.

### DIFF
--- a/app/classes/query/user_with_contribution.rb
+++ b/app/classes/query/user_with_contribution.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class Query::UserWithContribution < Query::UserBase
+  def initialize_flavor
+    where << "users.contribution > 0"
+    super
+  end
+end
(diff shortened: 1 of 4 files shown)


JoeCohen:
I assume that the boldface number (see below) is the ranking
Screenshot 2024-03-17 at 4 21 12 PM
Suggestions:
1. Start at 1 rather than 0. 
2. The number should be the absolute rank, rather than the rank on the particular page. (I.e., right now each page starts with 0.)

JoeCohen (changes requested):
I assume that the boldface number (see below) is the ranking
Screenshot 2024-03-17 at 4 21 12 PM
Suggestions:
1. Start at 1 rather than 0. 
2. The number should be the absolute rank, rather than the rank on the particular page. (I.e., right now each page starts with 0.)

Issue #1897, opened 2024-02-11
MO Species List Generator is Broken.  Fix it or turn it off.


mo-nathan:
The default species list in MO is horribly broken.  It wouldn't surprise me if it's been this way for years or if it has become progressively worse, but it is embarrassingly bad at the moment.  For example,

Go to https://mushroomobserver.org/species_lists/name_lister/new
Click "Abortiporus".
See appropriate names show up in the middle panel.
Double click "biennis" in the middle panel.
See it get added to the "Selected" panel.
Now click on another genus in the "Genera" panel.
See the appropriate species show up after the names added by clicking on "Abortiporus".
However, you cannot now select the newly added species in any rational way.
Double clicking on entries in the "Species" panel seems to add stuff based on what should be in that panel, but it also duplicates previous selections.

<img width="1071" alt="Screenshot 2024-02-11 at 6 46 09 AM" src="https://github.com/MushroomObserver/mushroom-observer/assets/3211730/cbbf8d75-55c3-4cb8-9e7f-ccf115d7ac34">

We need to either fix this or turn this way of creating species lists.  There is another species list creation UI that you can get to by clicking "Create Species List" from the broken one.  Weirdly you can't get there using the URL that shows up after you click that button (https://mushroomobserver.org/species_lists/name_lister).  I expect there are some missing data/parameters that need to get tracked down.

nimmolo:
@mo-nathan Thanks for documenting this, this issue writeup is pretty thorough. "Broken" is an imprecise term though (I wouldn't bring a car with a flat tire to a mechanic and say "my car is broken"...) — it sounds to me like it's "mostly working" :) 

Seems the issue is that:
- the third column is behaving like the other two: it's clickable. Whereas it should be an inert (generated) list of names that are NOT clickable. Is that right? 

I personally have never used the NameLister. If it's worth keeping and we're going to get into fixing it, let's get a solid description of what the expected behavior is. (All the logic for this UI is in `app/javascript/controllers/name-list_controller.js`, incidentally. I feel like the issue is in `drawColumn`.)

I agree that our other "Create" link, the one that's more like a "select your observations for a list" link, should be the prominent one. I remember we discussed this at a meeting more than a year ago, and then I forgot to act on it. (My feeling is still that the term "Species List" is misleading: it's an Observation List, and the NameLister is more like a "draft" of an observation list.)

PR #2029, opened 2024-03-15
Fix "name lister" js

nimmolo:
Hopefully restores original functionality!
#1897

### DIFF
--- a/app/javascript/controllers/name-list_controller.js
+++ b/app/javascript/controllers/name-list_controller.js
@@ -386,7 +386,7 @@ export default class extends Controller {
 
       if (x > 0) {
         author_el.classList.add("normal")
-        author_el.innerHTML = this.escapeHTML(name.substr(x + 1))
+        author_el.innerHTML = this.escapeHTML(" " + name.substr(x + 1))
         name = name.substr(0, x)
       }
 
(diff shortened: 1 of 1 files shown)


nimmolo:
@mo-nathan If you get a chance, please give this a spin and let me know if it behaves correctly.

mo-nathan:
Seems better after merging in the latest changes on main and running the usual update stuff.  However, the names are weird.  They appear to be coming from the test fixtures for some reason rather than mo_development as I would expect.  It's possible it's coming from mo_test which gets loaded with the test fixture, but I dropped that database and the weird names continue to show up.  It also doesn't make sense to me why the middle panel lists all the species for the genera that have been selected so far.

nimmolo:
- Weird names: On dev, these come from a static file Jason created long ago to test the name lister — they aren't intended to be up to date. On production, the names are generated via a nightly script from the live db.
- I'll ask Jason, but I think it always listed the species names of all selected Genera in the middle panel.

mo-nathan:
Interesting.  I'm guessing this is coming from test/reports/name_list_data.js.  At this point I don't like the behavior in the center panel, but at least it seems to work reasonably.

mo-nathan (approved):

nimmolo:
Oh. Maybe i get it - you're expecting the middle column to ONLY show species of the _currently selected_ Genus...

That was confusing me. Just changed it, this should do it.

mo-nathan:
Yes, that's exactly what I meant.  Awesome!  Thanks for fixing this.

PR #2032, opened 2024-03-17
Update schema.rb based on fresh pull from prod

mo-nathan:
These are the changes Rails made to the db/schema.rb after copying the latest production dump and running `rails db:migrate` on the latest main branch.  No migrations were run.  So this indicates that what we have in production is different from what the db/schema.rb currently specifies.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -496,7 +496,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_03_09_072017) do
     t.integer "project_id", null: false
   end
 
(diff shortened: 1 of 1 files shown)


mo-nathan:
@nimmo Can you load a current database snapshot into your local?  I think something is funky in your local database that's causing these changes to show up.  There might be some difference in how the migrations get interpreted in your local system vs. what happens in production.

JoeCohen (approved):

JoeCohen:
Merging per conversation with @mo-nathan

PR #2036, opened 2024-03-17
Fix collection number link

mo-nathan:
To reproduce:
Go to an observation page with a collection number (e.g., https://mushroomobserver.org/observations/489292)
Note that if you hover over the collection number link the URL is really weird (e.g., https://mushroomobserver.org/collection_numbers/9848?q=%23%3CQuery%3A%3ACollectionNumberAll%3A0x00007f4ee4c74748%3E)
This PR fixed that by alphabetizing the id of the query rather than encoding the query as above.
However, even after that fix, if you click on the link it does not in fact go where it claims it will, but instead the name of the collection number is replaced by "Content missing".
Weirder still "Content missing" is not in our code base.

### DIFF
--- a/app/helpers/tabs/collection_numbers_helper.rb
+++ b/app/helpers/tabs/collection_numbers_helper.rb
@@ -55,9 +55,10 @@ module Tabs
 
     def show_collection_number_tab(c_n, obs)
       cn_query = Query.lookup(:CollectionNumber, :all, observations: obs.id)
+      q_id = cn_query.id ? cn_query.id.alphabetize : nil
 
       [tag.i(c_n.format_name.t),
-       collection_number_path(id: c_n.id, q: cn_query),
+       collection_number_path(id: c_n.id, q: q_id),
        { class: "#{tab_id(__method__.to_s)}_#{c_n.id}" }]
     end
 
--- a/app/helpers/tabs/collection_numbers_helper.rb
+++ b/app/helpers/tabs/collection_numbers_helper.rb
@@ -55,12 +55,12 @@ module Tabs
 
     def show_collection_number_tab(c_n, obs)
       cn_query = Query.lookup(:CollectionNumber, :all, observations: obs.id)
-      q_id = cn_query.id ? cn_query.id.alphabetize : nil
+      q_id = cn_query.id&.alphabetize
 
       [tag.i(c_n.format_name.t),
        collection_number_path(id: c_n.id, q: q_id),
        { class: "#{tab_id(__method__.to_s)}_#{c_n.id}",
-         data: { turbo_frame: "_top" }}]
+         data: { turbo_frame: "_top" } }]
     end
 
     def collection_number_mod_tabs(c_n)
--- a/app/helpers/tabs/collection_numbers_helper.rb
+++ b/app/helpers/tabs/collection_numbers_helper.rb
@@ -59,7 +59,8 @@ module Tabs
 
       [tag.i(c_n.format_name.t),
        collection_number_path(id: c_n.id, q: q_id),
-       { class: "#{tab_id(__method__.to_s)}_#{c_n.id}" }]
+       { class: "#{tab_id(__method__.to_s)}_#{c_n.id}",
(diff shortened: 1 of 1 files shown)


JoeCohen:
The second Collection Number link at http://localhost:3000/489292 works fine.
Does this offer any clues about the first link:

- Right click the frist link and open in a new tab.
- Result: It does go to http://localhost:3000/collection_numbers/9848 in a new tab.
That page has the title `Collection Number 'Paul Sadowski PS1'` and shows 13 Observations.
Is the number is parsed correctly? Did the user reuse the number?

JoeCohen:
Content MIssing. Perhaps this from Bing Chat:
>The **"Content Missing"** error you're encountering with Turbo Frames in your Rails application can be resolved. Let's address it step by step:
1. **Cause of the Error**:
   - When using `turbo_frame_tag`, you wrapped your page content inside a frame.
   - After submitting a new task, you see the "Content Missing" message.
   - The issue arises because the expected `<turbo-frame>` is not present in the response.
2. **Solution**:
   - You can specify that the page should **redirect** instead of rendering the "Content Missing" error.
   - To achieve this, add a **turbo visit control tag** to your page. There are two ways to do this:
     a. In your `show.html.erb` file, add the following meta tag:
        ```html
        <meta name="turbo-visit-control" content="reload">
        ```
        [This instructs Turbo to reload the page instead of showing the error](https://stackoverflow.com/questions/75738570/getting-a-turbo-frame-error-of-content-missing)¹.

     b. Alternatively, you can use a helper method:
        ```erb
        <%= turbo_page_requires_reload %>
        ```
3. **Break Out of the Frame**:
   - If you want a link to break out of the Turbo frame, explicitly add the following data attribute:
     ```erb
     <%= link_to "Cancel", "#", data: { turbo_frame: "_top" }, class: "cancel-button" %>
     ```
     This ensures that the link doesn't stay within the frame and avoids the "Content Missing" issue¹.
...
>
Source: Conversation with Bing, 3/[1](https://stackoverflow.com/questions/75738570/getting-a-turbo-frame-error-of-content-missing)7/2024
([1](https://stackoverflow.com/questions/75738570/getting-a-turbo-frame-error-of-content-missing)) Getting a Turbo Frame Error of "Content Missing". https://stackoverflow.com/questions/75738570/getting-a-turbo-frame-error-of-content-missing.
(2) Common Mistakes — WickedWhims. https://wickedwhimsmod.com/installation-mistakes.
(3) Troubleshooting Runs – Turborepo. https://turbo.build/repo/docs/troubleshooting.
(4) undefined. https://turbo.hotwired.dev/reference/attributes.

mo-nathan:
Nice!  That seemed to work.  Pushed the fix to the PR.

nimmolo:
Aha. I believe this issue started with the upgrade to Turbo v2. There will be a few more, most likely.

I think a better solution here is to put `data-turbo="false"` on the link. (or in Ruby, `data: { turbo: false }` in the attributes)

**Simple explanation:**
If a link is inside a `<turbo-frame>`, Rails expects the destination to provide a `turbo_frame` response that will replace the HTML of *that section of the page*, and there is no fallback. If the response doesn't provide that, Turbo replaces the HTML of the `<turbo-frame>` with "Content missing". 

So the deeper problem is my use of `turbo_frame` for the "collection numbers" section of Show Obs (in my initial Turbo build out, pre-2). I believe it's not necessary anymore, but I'd have to check through the controller because it may require some refactoring.

Turbo 2 does not require turbo-frame tags. It will update the relevant section of the page if that's the only thing that changed.

nimmolo:
EDIT: nope, I think it's more correct the way you have it now with `"_top"`. 
This is a new attribute, for me, but now I know. Thanks Bing!

nimmolo:
Heads up i just discovered "fungarium records" has the same issue.  

I'm going to try a separate PR replacing these `turbo-frame` tags with `div`s and removing the special data-attribute, I think that will fix it more durably with less gotchas.

mo-nathan:
Sounds good.  Just saw this after deploying my fix.

PR #2037, opened 2024-03-18
Host jQuery locally

nimmolo:
This is to shut up the people who have installed browser extensions blocking jquery and then tell us that our images don't load, mysteriously.

I'm reluctant to do this because it locks our version of a frequently changing library, and forces us to manually update the library. jQuery also will load more slowly from our server than from the CDN — if a user has ever downloaded jQuery from the same CDN, they don't even have to load it at all, it will be in the browser cache. 

At a certain point we're going to have to tell people to stop blocking public JS repos, or... sorry, Charlie. But maybe this is not so bad in this case, because we're aiming to eliminate the jQuery dependency with Bootstrap 5. Hopefully it's just for a few months.

### DIFF
--- a/config/importmap.rb
+++ b/config/importmap.rb
@@ -6,8 +6,7 @@
 
 # jQuery is tricky with importmap: must use cdn.jsdelivr.net, NOT jspm
 # (which is for React and similar), or else `$` will not be available globally
-pin "jquery", to: "https://cdn.jsdelivr.net/npm/jquery@3.7.1/dist/jquery.js",
-              preload: true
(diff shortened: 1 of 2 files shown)


JoeCohen (commented):
I'm not knowledgeable enough to add anything meaningful to this.
Does hosting jquery increase the load on our system?
@mo-nathan: Is it a good practice to rely on 3rd party services for core functions?

mo-nathan:
This seems reasonable to me.  Technically, yes, it increases the load on our system, but the jQuery file is less than 300K and I expect it would be aggressively cached.

PR #2038, opened 2024-03-18
Show obs - convert `turbo-frame` tags to `div`

nimmolo:
Lower-maintenance solution for turbo updates on the show obs page:

We don't have to use `<turbo-frame>` tags for "surgical" updates to the DOM. `<div>` with an `id` is enough.

Switching to `<div>`s also means we don't need any special data attributes on nested `<a>` links. That's because links within a `turbo-frame` expect a response that replaces the `turbo-frame` they are nested within, and not a new page. (#2036) It seems like a gotcha that will confuse future developers. 

Note: there's a separate existing issue here, and I don't know how or if it's related: #2039
Update: it wasn't related, but this PR fixes the issue. The obs id wasn't getting passed to the re-rendered `new` HR/CN links, so the links didn't work.

### DIFF
--- a/app/helpers/tabs/collection_numbers_helper.rb
+++ b/app/helpers/tabs/collection_numbers_helper.rb
@@ -59,8 +59,7 @@ module Tabs
 
       [tag.i(c_n.format_name.t),
        collection_number_path(id: c_n.id, q: q_id),
-       { class: "#{tab_id(__method__.to_s)}_#{c_n.id}",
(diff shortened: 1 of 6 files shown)


nimmolo:
@mo-nathan Should I maybe merge this, and sort out the existing delete/add refresh issue in a separate PR?

mo-nathan:
Looks good to me.  I was not able to reproduce the behavior reported in #2039.

Issue #2042, opened 2024-03-19
POST logout


JoeCohen:
Use POST (rather than GET) for logout.

JoeCohen:
See this email chain
Nathan Wilson <nathan@collectivesource.com> Tue, Mar 19, 2024 at 4:54 AM
>Yeah, I essentially did all that already and it doesn't work. link_to by default creates GETs these days and the `method:` option does not work. Like I said there's a turbo-ish way to do it or we could use button_to and fix the CSS. I'm leaning toward button_to and fixing the CSS, but I'm hopeless at fixing CSS.

-----
 On Tue, Mar 19, 2024 at 7:31AM Joseph Cohen <jdcohenesq@gmail.com> wrote:
>Bing Co-Pilot says the following. I don't understand all of it and it doesn't apply directly to our code as we don't have a SessionsController.
But it may hint at the right solution
To modify the logout action in a Ruby on Rails application to use an HTML POST request instead of a GET request, follow these steps:
>>1. Update Routes: In your config/routes.rb file, define a custom route for the logout action. You can use the map method to create a named route for
logging out. For example:
`map.log_out 'logout', controller: 'sessions', action: 'destroy'`
This route maps the URL path /logout to the destroy action in the SessionsController.
>>2. Link to Logout: In your view (e.g., a navigation bar or a dedicated logout page), create a link to log out using the newly defined route. You can use the `link_to` helper method:
  `<%= link_to "Sign out", log_out_url %>`
This will generate an anchor tag with the appropriate URL for logging out.
>>3. Controller Action: In your SessionsController, ensure that the destroy action handles the POST request. You can use the method option to specify the
HTTP method:
  ```ruby 
    def destroy
      # Your logout logic here
      # ...
    end
```
>>By default, Rails uses GET for logout actions. By specifying the method as DELETE, you ensure that the logout action uses a POST request.
Remember to adjust the above steps according to your specific application structure and requirements. Using a POST request for logout is a good practice for security reasons, as it prevents accidental logouts due to browser prefetching or cached URLs.

---

PR #2043, opened 2024-03-19
POST logout

JoeCohen:
- Uses POST (instead of GET) for `logout` action. 
This prevents eager loading and may solve the issue of some users being spontaneously logged out.
It requires rendering `Logout` as a button (rather than a link).

See #2042

#### TODO
- [x] Fix tests to use logout title, or preferably href, instead of funky class.
- [x] Review stylesheets

### Suggest Manual Test
- On a large screen, hover over the Logout icon for a while to see if it causes eager loading or a spontaneous logout. Then click the link to ensure that it does log you out.
- Do the equivalent on a small screen.

### DIFF
--- a/app/assets/stylesheets/mo/_form_elements.scss
+++ b/app/assets/stylesheets/mo/_form_elements.scss
@@ -116,5 +116,12 @@ a.filter-only {
 .name-radio {
   label {
     font-weight: normal;
-  } 
+  }
+}
+
+.btn-glyphicon-only {
+  background: none;
+  border: none;
+  padding: none;
+  top: 5px;
 }
--- a/app/assets/stylesheets/mo/_form_elements.scss
+++ b/app/assets/stylesheets/mo/_form_elements.scss
@@ -118,10 +118,3 @@ a.filter-only {
     font-weight: normal;
   }
 }
-
-.btn-glyphicon-only {
-  background: none;
-  border: none;
-  padding: none;
-  top: 5px;
-}
--- a/app/views/controllers/account/welcome.html.erb
+++ b/app/views/controllers/account/welcome.html.erb
@@ -6,7 +6,7 @@ add_page_title(account_welcome_title(@user))
   <%= :welcome_note.tp %>
   <%= button_to(:app_logout.t,
                 account_logout_path,
-                { class: "btn", id: "nav_user_logout_link" }) %>
+                { class: "btn btn-default", id: "nav_user_logout_link" }) %>
 <% else %>
   <%= :welcome_no_user_note.tp %>
 <% end %>
--- a/app/views/controllers/account/welcome.html.erb
+++ b/app/views/controllers/account/welcome.html.erb
@@ -4,7 +4,9 @@ add_page_title(account_welcome_title(@user))
 
 <% if @user %>
   <%= :welcome_note.tp %>
-  <%= link_to(:welcome_logout_link.t, account_logout_path) %>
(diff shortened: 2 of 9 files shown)


JoeCohen:
@mo-nathan & @nimmolo : 
A very short PR to prevent eager load of `logout`, based on @mo-nathan's email. See #2042.
I think the logout link looks OK in the top navbar. 
Can the css be improved?

app/assets/stylesheets/mo/_form_elements.scss
```diff
@@ -116,5 +116,12 @@ a.filter-only {
 .name-radio {
   label {
     font-weight: normal;
-  } 
+  }
+}
+
+.btn-glyphicon-only {
```
JoeCohen, line 122: @nimmolo 
- Is there a better name for this style?
- Can its substance be improved?
- Does it belong in this file or in app/assets/stylesheets/mo/_top_nav.scss?

app/views/controllers/application/sidebar/_user.html.erb
```diff
@@ -2,18 +2,13 @@
   [
     tag.i("", class: "glyphicon glyphicon-user"),
     tag.span(h(@user.login), class: "ml-2"),
-    tag.span(class: "pull-right") do
-      link_to(:app_logout.t, account_logout_path, id: "nav_user_logout_link")
+    tag.span(class: "pull-left") do
+      button_to(:app_logout.t,
+                account_logout_path,
+                { class: "btn", id: "nav_user_logout_link" })
     end
```
JoeCohen, line 9: Should this be centered or otherwise moved?

app/assets/stylesheets/mo/_form_elements.scss
nimmolo: @JoeCohen - I changed some things to make it resemble the current links. 

The class will be also useful in Bootstrap 4-5, it's `btn-link`. 
It's for when the element must be a button, but you want it to resemble a link. 
I agree that the top nav logout might be better as a button, and i'm looking to rethink all that stuff soon (before upgrading to BS4) but it might make that simpler to keep it as is for now.

app/views/controllers/application/sidebar/_user.html.erb
nimmolo: This one currently flushes right, but it doesn't really work with long usernames > "chaelthomas".
I propose knocking it down to the next line... committed that change, let me know if that works.

nimmolo: (Trying to avoid the situation where it currently smushes the username on top of the word "logout")

app/assets/stylesheets/mo/_form_elements.scss
nimmolo: (I think `.btn-glyphicon-only` is not necessary with these changes)

app/views/controllers/account/welcome.html.erb
```diff
@@ -4,7 +4,9 @@ add_page_title(account_welcome_title(@user))
 
 <% if @user %>
   <%= :welcome_note.tp %>
-  <%= link_to(:welcome_logout_link.t, account_logout_path) %>
+  <%= button_to(:app_logout.t,
+                account_logout_path,
+                { class: "btn", id: "nav_user_logout_link" }) %>
```
nimmolo, line 9: iirc with BS everywhere you use `btn` you have to also use a coloration, like `btn-default`.

PR #2045, opened 2024-03-19
Remove the vote-table modal if it exists, on vote, to force a refresh

### LINKED ISSUE #1994: Confidence vote table is not updating on vote
user pinonbistro says:

> Confidence vote is not working.  See https://mushroomobserver.org/468517  where I said I'd call it Cortinarius trivialis group and the confidence is only 57%.  It change promising on Cortinarius sp. to 38%.

(I don't understand what the score should be.)

nimmolo:
- Fixes #1994 Confidence vote table is not updating on vote
- Simplifies updates of show obs page to take advantage of Turbo 8 morphing (!)
___
### Updating the table in the modal ###
The table wasn't refreshing because we're using `modal_link_to`, which _re-opens_ a modal _if it exists_.  This behavior is helpful and expected in the case of a form in progress, but here it's preventing reload of the modal content.

Just needed to add something to the Stimulus `naming-vote_controller` so that whenever the user casts a vote, we delete the `modal_naming_votes_#{naming.id}` modal if it's in the DOM.

**To test:**
- Open the vote table modal, close it
- Change your vote on a naming
- Open the vote table again, the tally should be updated

### Refreshing the page more simply ###
An update to the Namings table is a case of a Turbo partial page refresh: the namings table should be updated, as well as possibly the title of the obs, and the name info panel, but nothing else on the page should change, and the page should not "jump" back to the top - it should retain scroll position. 

Before Turbo 8 the DOM elements had to be explicitly changed in a turbo_stream template. 

Turbo 8 allows us to simply redirect to the same action for a refresh. Turbo detects that it's the same page, loads the HTML of the page without reloading Rails, detects the differences in DOM elements (having IDs) with a diffing algorithm, and "morphs" changed DOM elements into the page, all without a page refresh. 

This means we don't need to write a special turbo_stream updating template, and in some cases we don't even need any separate Turbo response. We can just redirect to the page, as we used to do in Rails 4.

To maintain scroll position, i had to add a `meta` tag to the `head` element.

**To test:**
- Downvote an existing naming
- Propose a new naming and give it a high vote so it becomes the new "consensus" name
- The naming table, title, name panel and flash should all refresh


JoeCohen (dismissed):
I did both manual tests and it works fine.
Looked at the changed files, but am still way behind on understanding js etc., so I have nothing to add.

JoeCohen (commented):
The tests consistently fail locally with this failure, which seems relevant to the fixes in this PR:
```ruby
joe@Josephs-iMac mushroom-observer % rails test:all
...
[Screenshot Image]: /Users/joe/mushroom-observer/tmp/capybara/failures_test_identify_index_naming_and_vote_ui.png
 FAIL HelpIdentifySystemTest#test_identify_index_naming_and_vote_ui (167.08s)
        expected to find css "#modal_obs_247545329_naming" but there were no matches
        test/system/help_identify_system_test.rb:30:in `test_identify_index_naming_and_vote_ui'

...
```
Fails even when run on its own.
```ruby
% rails t test/system/help_identify_system_test.rb:30
Started with run options --seed 25344

[Screenshot Image]: /Users/joe/mushroom-observer/tmp/capybara/failures_test_identify_index_naming_and_vote_ui.png
 FAIL HelpIdentifySystemTest#test_identify_index_naming_and_vote_ui (15.50s)
        expected to find css "#modal_obs_247545329_naming" but there were no matches
        test/system/help_identify_system_test.rb:30:in `test_identify_index_naming_and_vote_ui'
```

Issue #2048, opened 2024-03-21
Hovering over the Admin Mode button turns it on (similar to logout)


nimmolo:
- Hover over the button in the top bar for a second
- Refresh the page
- You're in admin mode

PR #2049, opened 2024-03-21
Change "admin_mode_path" from a GET to a POST

nimmolo:
#2048 - Prevent accidentally turning on Admin mode by hovering over the Admin button.

Changes the action from `show` to `create` in Admin::SessionController.

### DIFF
--- a/app/controllers/admin/session_controller.rb
+++ b/app/controllers/admin/session_controller.rb
@@ -15,7 +15,7 @@ module Admin
     before_action :login_required
 
     # The route to turn admin mode on or off. Takes params.
-    def show
+    def create
       if params[:turn_on]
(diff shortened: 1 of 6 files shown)


JoeCohen:
@nimmolo: Thanks for doing this. 

1. It does work to prevent accidentally entering/leaving admin mode, which is great.
2. When explicitly entering/leaving admin mode, the page you end up on is sort of random.
It appears to depend on where your cursor has been. Maybe it's what was last eager loaded.
Perhaps deal with that in a separate PR?

nimmolo:
That's true, huh. I imagine it is the last eager loaded page you end up on, and I can't really imagine how we could override that... this preloading is a new twist.

PR #2053, opened 2024-03-22
Bundle update (`Gemfile.lock`)

nimmolo:
Update current gems, including rubocop


nimmolo:
@JoeCohen if you approve, please feel free to merge and deploy in the morning.

JoeCohen:
Done.
Thanks for the update.
Question: How often should we be doing this.

nimmolo:
Good question - i don't really know. I sometimes run it by accident, and then i notice there are a bunch of updates.

PR #2054, opened 2024-03-22
Try to quiet SASS deprecation warnings on deploy in production

nimmolo:
Adds a config attempting to silence the VERY numerous SASS deprecation warnings that now appear during deploy on production. This config is undocumented and it's likely it will just be ignored, but it is part of the dart-sass API.

If this doesn't work and bothers anyone, we have three options:
- switch to gem `dartsass-rails`, it does offer the `--quiet-deps` option
- submit a PR on the gem `dartsass-sprockets` that will implement this config
- Update to Boostrap 4. The deprecations are because BS3 uses outdated SASS syntax within the gem.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -95,6 +95,11 @@ MushroomObserver::Application.configure do
   # Generate digests for assets URLs
   config.assets.digest = true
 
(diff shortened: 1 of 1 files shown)


nimmolo:
Whoa. It worked.

PR #2055, opened 2024-03-22
Rework show user and contribs layouts

nimmolo:
This PR rearranges the user layout into panels, and puts the user's images in a carousel. Todo:
- [x] Move top tab links into user info panel
  - [x] Observations by user
  - [x] Comments by user (?)
  - [x] Email user (requires a CRUD route for modal form)
- [x] New `Users::EmailsController`

It also reworks the "Contributors" index into a matrix_box grid. Hopefully more lively! The "How we estimate" is here a collapsible info panel.
- [x] Redo legend info "tables" as `table` elements
- [ ] Move "Site Stats" link into panel


JoeCohen (commented):
1. The layout is good. (But I'm personally not a fan of blowing up photos because they often to lose resolution and look worse. However, it's your choice.)
2. See suggested code changes.
3. Maybe add some explanation of how the thumbs are selected? I think most users don't realize that they can override other people's favorites.)

nimmolo:
@JoeCohen - thanks, i was wondering how the current code would not stomp the user's favorites as it was. Committed those changes.

The photo carousel for user obs photos is the same size as in show observation... ??

JoeCohen:
Yes, the photo carousel size is the same.
You should just ignore my comment about big images.

PR #2063, opened 2024-03-24
Field Slips - Adjust form templates

nimmolo:
Uses simplified thumbnail panels for "current"/"last" with buttons to use them, and deletes some unnecessary divs, etc.

Order of links:
- New: Create, Keep, Use Last
- Edit: Keep, Use Last, Create

The "keep" and "use last" panels will display side by side if both exist, or just one if there's only one.

screen_shot_2024-03-24_at_3 57 38_pm

Please give this a spin and feel free to merge (into `njw-field-slips`) if you like it, or let me know about button placement.


mo-nathan:
I like this faciility for other pages, but I don't think we want to add this much complexity to the Record Field Slip page and we really want "Create New Observation" to be the most obvious choice so new users don't get confused/distracted.  If anything, I'd rather the "Use Last Observation" was more hidden so only power users would know it's there.

nimmolo:
Revised with simpler panel, no matrix_box

mo-nathan (approved):

PR #2068, opened 2024-03-26
QR code on Field Slip show page

mo-nathan:
Mostly to demonstrate the rqrcode gem.  We may want to put the QR code on this page, but mostly it's planned to get added to a PDF output for printing slips.

### DIFF
--- a/app/views/controllers/field_slips/show.html.erb
+++ b/app/views/controllers/field_slips/show.html.erb
@@ -6,6 +6,10 @@
 
 <h1><%= "#{:FIELD_SLIP.t}: #{@field_slip.code}" %></h1>
 
+<div>
+  <%= @field_slip.qr_code(field_slip_url(@field_slip)).html_safe %>
+</div>
+
 <%= render @field_slip %>
 
 <div>
--- a/app/views/controllers/field_slips/show.html.erb
+++ b/app/views/controllers/field_slips/show.html.erb
@@ -1,10 +1,8 @@
-<% add_page_title(@field_slip.code) %>
+<% add_page_title("#{:FIELD_SLIP.t}: #{@field_slip.code}") %>
 
 <% if notice %>
   <p class="alert alert-success"><%= notice %></p>
 <% end %>
-
-<h1><%= "#{:FIELD_SLIP.t}: #{@field_slip.code}" %></h1>
 
 <div>
   <%= @field_slip.qr_code(field_slip_url(@field_slip)).html_safe %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -384,8 +384,8 @@ MushroomObserver::Application.routes.draw do # rubocop:todo Metrics/BlockLength
                                      as: "glossary_term_versions")
 
   # ----- Field Slip Records: standard actions --------------------------------
-  resources :field_slips
   get("qr/:id", to: "field_slips#show", id: /.*[^\d.-].*/)
+  resources :field_slips
 
   # ----- Herbaria: standard actions -------------------------------------------
   namespace :herbaria do
--- a/Gemfile
+++ b/Gemfile
@@ -111,6 +111,9 @@ gem("fastimage")
 # for detecting file type of uploaded images
 gem("mimemagic")
 
+# QR code support for field slips
+gem("rqrcode")
+
 # for creating zip files
 # RubyZip 3.0 is coming!
 # **********************
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -98,6 +98,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
+    chunky_png (1.4.0)
     concurrent-ruby (1.2.3)
     connection_pool (2.4.1)
     crack (1.0.0)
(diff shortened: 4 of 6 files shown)


JoeCohen:
Suggestion: Omit the page title on field_slip#show. It duplicates what's in the displayed field slip.
I don't have answers to your questions, except I don't see an easy way to do the routing you want.
@nimmolo may have answers to the first two.

JoeCohen (commented):
See my earlier comment about page title.

JoeCohen:
@mo-nathan. Re routes.  
I think this gets what you want, except for the`:id` constraints, but you can check for that in the controller.
(When I added `id: /.*[^\d.-].*/`, the tests blow up. Without the constraint, everything passes.)
In routes.rb:
```ruby
  match "field_slips", to: "field_slips#update", via: [:patch, :put]
  delete("/field_slips", to: "field_slips#destroy")
  get("/qr/:id", to: "field_slips#show", as: :field_slip)
  resources :field_slips, only: [:index, :new, :create, :edit]
```
This yields the following routes:
```ruby
% rails routes | egrep "\s/(field_slip|qr)"                               
    field_slips GET       /field_slips(.:format)           field_slips#index
                POST      /field_slips(.:format)           field_slips#create
 new_field_slip GET       /field_slips/new(.:format)       field_slips#new
edit_field_slip GET       /field_slips/:id/edit(.:format)  field_slips#edit
                PATCH|PUT /field_slips(.:format)           field_slips#update
                DELETE    /field_slips(.:format)           field_slips#destroy
     field_slip GET       /qr/:id(.:format)                field_slips#show
```

nimmolo:
It might be better to keep a `field_slip_qr_path` helper separate from a `field_slip_path`. (Or maybe i don't yet understand why you'd want the `/qr` path as the default). 

In any case there's no way to get Rails to generate that special path automatically with/alongside the other resources. 

What I noticed in my most recent `routes` PR is that the named routes are used more than once, and differentiated by method, so:
- `field_slip_path` — `put` is for `:update` and `get` is for `:show` 
- `field_slips_path` — `post` is for `:create` and `get` is for `:index`. 

If you do `resources :field_slips, except: :show` and try to call something else `field_slip_path`, it will skip the `:update` path too, which you probably don't want.

config/routes.rb
```diff
@@ -384,8 +384,8 @@ def route_actions_hash
                                      as: "glossary_term_versions")
 
   # ----- Field Slip Records: standard actions --------------------------------
-  resources :field_slips
   get("qr/:id", to: "field_slips#show", id: /.*[^\d.-].*/)
```
nimmolo, line 387: I'd do something like 
```ruby
get("qr/:id", to: "field_slips#show", id: /.*[^\d.-].*/, as: "field_slip_qr")
```
I don't think you need the `id` part, either.

app/views/controllers/field_slips/show.html.erb
```diff
@@ -1,10 +1,12 @@
...
+<% add_page_title("#{:FIELD_SLIP.t}: #{@field_slip.code}") %>
 
 <% if notice %>
   <p class="alert alert-success"><%= notice %></p>
 <% end %>
 
-<h1><%= "#{:FIELD_SLIP.t}: #{@field_slip.code}" %></h1>
+<div>
+  <%= @field_slip.qr_code(field_slip_url(@field_slip)).html_safe %>
+</div>
```
nimmolo, line 9: You probably will need `html_safe` if it's rendering as `svg`... that's potentially unsafe html.

mo-nathan:
> Suggestion: Omit the page title on field_slip#show. It duplicates what's in the displayed field slip. I don't have answers to your questions, except I don't see an easy way to do the routing you want. @nimmolo may have answers to the first two.

Yeah, that's already in #2067 which I just merged.

mo-nathan:
> It might be better to keep a `field_slip_qr_path` helper separate from a `field_slip_path`. (Or maybe i don't yet understand why you'd want the `/qr` path as the default).
> 
> In any case there's no way to get Rails to generate that special path automatically with/alongside the other resources.
> 
> What I noticed in my most recent `routes` PR is that the named routes are used more than once, and differentiated by method, so:
> 
> * `field_slip_path` — `put` is for `:update` and `get` is for `:show`
> * `field_slips_path` — `post` is for `:create` and `get` is for `:index`.
> 
> If you do `resources :field_slips, except: :show` and try to call something else `field_slip_path`, it will mess up the `:update` path too, which you probably don't want.

The motivation for /qr/ rather than /field_slip/ is to make the QR code simpler.  Realistically it doesn't really matter so I don't think it's worth the added complexity.  Thanks for the ideas!

app/views/controllers/field_slips/show.html.erb
mo-nathan: Thanks!

mo-nathan:
> Suggestion: Omit the page title on field_slip#show. It duplicates what's in the displayed field slip. I don't have answers to your questions, except I don't see an easy way to do the routing you want. @nimmolo may have answers to the first two.

I don't think I understand what you're getting at on field_slip#show unless it's something I pushed merged in a bit later yesterday and just merged into main from #2067.  As far as I can tell the only place the field slip number is given now is in the title.

JoeCohen:
@mo-nathan: Re field slip duplication of field slip number. Here's what I see (in this branch, in main, and on the webserver).
Should I be seeing something different?
Screenshot 2024-03-28 at 5 27 00 AM

mo-nathan:
That is what is currently in production.  Did you do a pull on the branch and main?  On this branch and main I currently get:
Screenshot 2024-03-28 at 7 23 26 AM

mo-nathan:
On main I currently get:

JoeCohen:
I must have been looking at the wrong branch. As you said, the FS # only appears once in #show.

JoeCohen (approved):

Issue #2066, opened 2024-03-26
Name Version Display


JoeCohen:
Showing back versions of Names should show all versioned fields. 
User Igor Safonov tried to figure out what had changed in a Name he created, but was having issues. He said:
> I compare versions often enough for location pages because it is possible to visualize two versions in the same format in two separate windows by quickly toggling between the two via the Alt-Tab command and noticing the differences.  Unfortunately, this is not (or no longer) possible for name pages because the previous version just gives you only a view of the nomenclature "sub-window", not the whole page.  That's why I asked for your help hoping that you can tell easily by some means only available to you as the website admin. :-)

PR #2069, opened 2024-03-26
Add notes to Show Name Version

nimmolo:
Igor requested more info on the Show Name Version page.  #2066

It should be possible to reuse partials from `show_name`, if that template has everything we want (which it should by definition). The "notes" are the only versioned info I could tell that's missing. (what else, maybe lifeform?)

Screen Shot 2024-03-26 at 2 41 05 PM


JoeCohen (approved):
Thanks for doing this. It looks good.

nimmolo:
PR now includes a refactor of the classification and lifeform panels, to make them easier to reuse, and of the Versions panel, to make it a proper panel.

PR #2070, opened 2024-03-27
Fix a cron error message, couple other things

nimmolo:
Refresh cron jobs should return an array of messages, not joined. `UserStats` was joining the messages.

Also - missed a few changes in last PR.
- `panel_block` - use `class_names` to concat css classes
- versions table - use `.table-hover`


JoeCohen:
@nimmolo: Thanks for doing this. I had wondered about an apparent increase in error messages.
We probably should have a protocol fro dealing with error messages sent to webmaster, especially from the chron job.
Maybe something to discuss at the next tech meeting.

PR #2079, opened 2024-03-30
Initial implementation of SolidQueue for running ActiveJobs

mo-nathan:
To test:
1. Run the migrations.
2. Start `bundle exec rake solid_queue:start` in a new shell.
3 .Enter the Rails console in another shell.
4. In the console, run `FieldSlipJob.perform_later(:test_arg)`.
5. `cat tmp/fs-*.txt` should show `[:test_arg]`.

After you're done test you probably want to run `rm tmp/fs-*.txt`.

To actually get this working in production we'll need run the process in step 2 in the background.  Also note that this does inflate the logs a bit since it polls the queue fairly often.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -205,3 +205,5 @@ group :production do
   # https://newrelic.com/
   gem("newrelic_rpm")
 end
+
+gem "solid_queue", "~> 0.3.0"
--- a/app/jobs/field_slip_job.rb
+++ b/app/jobs/field_slip_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class FieldSlipJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    filename = "tmp/fs-#{Time.now.to_i}.txt"
+    File.write(filename, args.to_s)
+    filename
+  end
+end
--- a/config/solid_queue.yml
+++ b/config/solid_queue.yml
@@ -1,0 +1,18 @@
+# default: &default
+#   dispatchers:
+#     - polling_interval: 1
+#       batch_size: 500
+#   workers:
+#     - queues: "*"
+#       threads: 5
+#       processes: 1
+#       polling_interval: 0.1
+#
+# development:
+#  <<: *default
+#
+# test:
+#  <<: *default
+#
+# production:
+#  <<: *default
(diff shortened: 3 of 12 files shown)


JoeCohen (approved):
Thanks for doing this!
I followed the procedure listed in the PR summary. It went fine.
I know nothing about SolidQueue or ActiveJobs, so cannot comment further.
(I'll have to learn about them soon.)

config/solid_queue.yml
```diff
@@ -0,0 +1,18 @@
...
+#       polling_interval: 0.1
+#
+# development:
+#  <<: *default
+#
+# test:
+#  <<: *default
+#
+# production:
+#  <<: *default
```
nimmolo, line 18: Aha - here's that reference to threads. You probably found this, but from what I could tell reading about it yesterday, if you pass no config, it doesn't default to threading.

Gemfile
```diff
@@ -205,3 +205,5 @@ group :production do
   # https://newrelic.com/
   gem("newrelic_rpm")
 end
+
+gem "solid_queue", "~> 0.3.0"
```
nimmolo, line 209: Two questions:
- Is that version lock intentional, or just the result of running `bundle add`
- Could we move this up to around line 45 (or wherever seems to make sense)? 

(I just went through the Gemfile last week to try to organize it a bit, and that helped me find a bunch of redundant and unnecessary gem declarations. I tried to group things sort of by functionality.)

app/jobs/field_slip_job.rb
```diff
@@ -0,0 +1,11 @@
...
+
+class FieldSlipJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    filename = "tmp/fs-#{Time.now.to_i}.txt"
+    File.write(filename, args.to_s)
+    filename
+  end
+end
```
nimmolo, line 11: Yay - i'm glad you have a job to give it right away.

Gemfile
mo-nathan: `bundle add` as you suspected and sure.

config/solid_queue.yml
mo-nathan: I'm also a lot less worried about threads in the job queue.  I find that the natural way to write jobs is to assuming that you get everything you know from the arguments you're given.  It certainly possible to get bitten by singleton variables like `User.current`, but it wouldn't be meaningful in a job unless you set it in the job which is kind of a silly thing to do.

mo-nathan:
Thanks for the comments this draft PR.  I've encorporated the changes into #2081.  I'm going to close this PR.

PR #2083, opened 2024-03-31
Fix MISSING REFLECTIONS

mo-nathan:
I think this is what is now wanted.  I guess this makes sense.  Seems a bit like unnecessary overhead to me.


pellaea:
Actually, check_for_broken_references looks for belongs_to reflections.
For example, field slips must belong to a project and a user?  (Sorry, I've
been out doing fieldwork and haven't been able to keep up with your recent
work on field slips!)

At some point in the relatively recent past Nimmo added all of the missing
associations for all of the other models.  He can probably provide a better
defense of the policy.  But I think they are pretty lazy these days, and if
they aren't used, they incur very little cost.

But if you would prefer, feel free to alter check_for_broken_references.  I
think all we'd need to do is add these two associations  to the reflections
hash around lines 26-40, since they won't be picked up automatically by
searching for belongs_to relationships.  My intent here was to make it
automatically pick up new associations without requiring us to remember to
do anything.

mo-nathan:
It already had the belongs_to relationships.  I added the corresponding has_manys (although nothing uses them), but that didn't fix the issue.  I finally got them to go away by adding the rows to script/check_for_broken_references.  I don't understand the purpose here well enough to have more opinion about it.

JoeCohen (commented):
This looks technically correct.
But what would you like to have happen in this scenario:
A collection from a foray ends up in a herbarium with its paper field slip.
The foray project gets destroyed.
Should a herbarium worker be able to scan the field slip and get directed to the Obs?

mo-nathan:
@JoeCohen Thanks for pointing that case out.  I think I've fixed it up so it will work better in that case.  As you suggest I would want the field slip to continue to work.

PR #2085, opened 2024-04-02
Location#merge fix

### LINKED ISSUE #2084: Location#merge incomplete
`Location#merge` fails to update `herbarium.location_id`.
That causes a broken reference for any Herbarium which used the merged Location.

### Tasks
- [x] Write a test exposing the bug.
- [x] Fix the bug.
- [x] Double check to make sure to make sure that merge isn't neglecting another association.
- [x] Suggestion: include the `id` of the non-surviving Location in the admin note which is added to Location.notes

JoeCohen:
Prevents certain broken references, i.e. references to Locations that are deleted via a merge: 
- Updates herbarium.location if the location has been merged.
- Ditto for project.location

Fixes the bug described in #2084

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -697,18 +697,10 @@ class Location < AbstractModel
       obs.save
     end
 
-    # Move species lists over.
-    SpeciesList.where(location_id: old_loc.id).find_each do |spl|
-      spl.update_attribute(:location, self)
-    end
-
-    # Update any users who call this location their primary location.
-    User.where(location_id: old_loc.id).find_each do |user|
-      user.update_attribute(:location, self)
-    end
-
-    Herbarium.where(location_id: old_loc.id).find_each do |herbarium|
-      herbarium.update_attribute(:location, self)
+    [Herbarium, Project, SpeciesList, User].each do |klass|
+      klass.where(location_id: old_loc.id).find_each do |obj|
+        obj.update_attribute(:location, self)
+      end
     end
 
     # Move over any interest in the old name.
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -105,6 +105,7 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
   has_many :comments,  as: :target, dependent: :destroy, inverse_of: :target
   has_many :interests, as: :target, dependent: :destroy, inverse_of: :target
   has_many :observations
+  has_many :projects
   has_many :species_lists
   has_many :herbaria     # should be at most one, but nothing preventing more
   has_many :users        # via profile location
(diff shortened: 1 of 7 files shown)


JoeCohen:
@mo-nathan, @nimmolo, @pellaea :
NOTE: `Location#merge` is a hidden trap for the unwary when adding a `location` to an existing model or when adding a new model that has a `location.`
### Question
Is there a way to either
- Reorganize the code to avoid this trap?
- Detect if an existing or new model has fallen into the trap and warn the coder?

mo-nathan:
By "trap" I think you mean that it's easy to not realize that Location#merge needs to be aware of any relationship to a Location object in case that object is deleted/merged, right?  If that's what you mean, you could at least look at Location.reflection.  However, in this case it would still have missed Project since Project has a belongs_to relationship with Locations, but this relationship is not expressed in location.rb with a corresponding has_many.  If you wanted to be through along the lines of check_for_broken_references, you could go through all the reflections of all the models.  Getting all the models is a bit tricky due to Rails lazy loading.  Googling around a bit, it seems like you might have to do some eager loading or file searching in the models directory.  If you do want to go through all the models, you could also (at least in theory) determine if for every belongs_to there is corresponding has_many or other relationship.

pellaea (approved):
Looks good at a glance.  I trust you!  Sorry about the trouble you apparently must have had with albion and the API test!

pellaea:
My two cents: check_for_broken_references kinda accomplishes the "detect if
something has fallen into the trap" thing.  Say you add another reference
to location and forget to deal with it in Location.merge.  Then when such a
referenced location is merged, check_for_broken_references will let us
know.  Somewhat cryptically, I admit, but it will point us in the right
direction, in any case.

JoeCohen:
Thanks Nathan!.
1. That's exactly what I meant by ""trap for the unwary". E.g., We've
neglected Herbaria locations for years. And the Project location_id
column was added, there weren't corresponding changes to Location merge.
2. I just pushed a test that finds the models via file searching. But my
test isn't clever about reflection or other relations. At the moment it
seems like too much trouble for now. Probably -- as @pellaea observes --
script/check_for_broken_references is enough to do what we need.

pellaea:
I should back-pedal a bit.  Even my script relies on a belongs_to
relationship being declared somewhere, which doesn't necessarily have to
happen.

(Incidentally, my script is slightly better than grepping for "belongs_to",
but it still does rely on directory listing of app/models to find the full
list of model classes!  Once it has a list of model classes, it uses a
documented ActiveRecord trick for enumerating the reflections, so that's
safe and stable.)

nimmolo:
So then...how about just adding at least this to `Location.rb` — the one that we know about?
```ruby
has_many :projects
```
No sense putting it off til we have the exhaustive list of missing reflections.

Issue #2080, opened 2024-03-30
Error showing non-existent User


JoeCohen:
Showing a non-existent user throws an Error.
To reproduce:
- `/user/1234567890`
- result: See [Action Controller_ Exception caught.pdf](https://github.com/MushroomObserver/mushroom-observer/files/14812447/Action.Controller_.Exception.caught.pdf)

We probably should go to the user index with the following flash error. (I think -- but not 100% certain -- we used to do that.)
`runtime_object_not_found: "Sorry, the [type] you tried to display (id #[id]) does not exist.  Someone may have deleted it or merged it into another."`

JoeCohen:
@nimmolo: Off the top of your head, do you think this one is related to recent changes to site & user stats? (Cf. #2078)
- If not, I'll dig into it.
- If so, could you see if you can fix it?

nimmolo:
@JoeCohen — It's definitely related somehow. I notice from your PDF that the error is thrown in `Checklist`, which i also worked on. The `raise` error line there on line 39 is a Joe line — I was wondering about that. Maybe that should throw to the user index instead? 

It seems like it shouldn't ever be hitting that line though, for a nonexistent user...  What I think should happen is that it should hit `find_user!` in `UsersController#show`, which is basically my eager-loading version of `find_or_goto_index`. (In various controllers, I've been rewriting the guard line in the `show` action to perform the functionality of `find_or_goto_index`, but allow for eager-loading, using a model scope. Maybe there's some way to pass the scope to `find_or_goto_index`?)

The other question is, it seems like we should probably be creating a blank `UserStats` for every new user. Does that seem right?

JoeCohen:
@nimmolo 

1. Nathan deserves credit for line 39, which irignally used `fail` instead of `raise error`. https://github.com/MushroomObserver/mushroom-observer/blame/113d1064d99f8e7bb75a911755324c4495c2fef2/app/classes/checklist.rb#L31. I just rubocopped it.
2. I'm also puzzled why it hits that line; `find_or_goto_index` feels like it should display the index and cease processing `create`. But it doesn't. Instead it returns `nil`, with the result that @show_user is set to `nil`, and `create` goes on its merry way with a nil @show_user.
3. A blank UserStats for new users seems right. Would it work instead (or should this happen additionally) to make the default UserStat#checklist an empty string, with `null: false`?

nimmolo:
@JoeCohen 
I'm out the rest of the day, but:

[...]

Issue #2078, opened 2024-03-30
nil user_stats for fresh User


JoeCohen:
Lately `/user/nnnnnn` throws errors for brand-new users. Example below.
This is reproducible locally:
- logout
- create a new account
- verify the account (find the verification email via `grep verify log/development.log`, go to the indicated url.)
- Show the new user.
 
-----
Cron Daemon to webmaster  Mar 29, 2024, 11:08 AM [PDT] (19 hours ago)
```ruby
INFO -- : Started GET "/users/170060" for 149.40.50.109 at 2024-03-29 18:07:52 +0000
INFO -- : Processing by UsersController#show as HTML
INFO -- :   Parameters: {"id"=>"170060"}
WARN -- : user=170060 robot=N ip=149.40.50.109
INFO -- :   Rendered layout controllers/layouts/application.html.erb (Duration: 11.0ms | Allocations: 2610)
INFO -- : Completed 500 Internal Server Error in 683ms (ActiveRecord: 645.5ms | Allocations: 9432)
E, [2024-03-29T18:07:53.652091 #540829] ERROR -- :   
ActionView::Template::Error (undefined method `checklist' for nil):
    12:     nil
    13:   end
    14: footer =
    15:   if @user_stats.checklist[:species].to_i > 0
    16:     [link_to(checklist_path(id: @[show_user.id](http://show_user.id/))) do
    17:        tag.strong(:app_life_list.t)
    18:      end, ": ",

app/views/controllers/users/show/_profile.erb:15
app/views/controllers/users/show.html.erb:15
app/controllers/application_controller.rb:226:in `catch_errors_and_log_request_stats'
```

JoeCohen:
Note that the `user_stats` table allows null values for  the `languages`, `bonuses`, and `checklist` columns.

JoeCohen:
@nimmolo: Off the top of your head, do you think this is related to recent changes to site & user stats?
- If not, I'll dig into it.
- If so, could you see if you can fix it?

PR #2086, opened 2024-04-03
Create `UserStats` on create Account, fix show nonexistent user

nimmolo:
Seems like a blank `UserStats` record should get created on account creation: It's also necessary for the show user page, which a new user might try to access (before the cron that would otherwise create a missing UserStats object has run). This should solve the issue in #2078

Also fixes the redirect in UsersController#show, for a nonexistent user, #2080

Adds three tests.


JoeCohen (approved):
Thanks for fixing these problems and for the tests.
The file changes look right on, but I haven't test driven it.

PR #2087, opened 2024-04-03
Show field code on observation form when present

mo-nathan:


JoeCohen (commented):
@mo-nathan  This functioned as intended.

#### Should the `field_code` field be editable vs just a label?

Edge case: User scans paper slip _n_, modfies the `field_code` field to be _n + x_;  creates a FieldSlip and Observation. Now scanning paper slip _n + x_ (which may be held by a different user) may have a suprise results

mo-nathan:
In my view, the field code on the Create Observation page should not be editable for exactly the reason you said.  However, we need some way to fix things, so I think allowing it to be edited on the FieldSlip#edit page is reasonable.  We could restrict that to admins.  I don't have strong opinion about whether the risk that reduces is worth the effort needed.

JoeCohen (approved):

PR #2091, opened 2024-04-10
Exclude vague locations from `Mappable::CollapsibleCollectionOfObjects`

nimmolo:
### Cleans up maps by getting rid of overly vague location boxes ###

Currently on the site, we exclude observations whose location is "Earth" on maps, because they mess up the interface. But an observation with a large associated Location like "USA" also obscures more specific observation data on a map, whenever _the vague location gets grouped with the more specific locations_. (See discussion #2088.)

One way to fix this might be by grouping large locations together so they don't hide the precise ones. But it's hard to control the order of grouping (it would require calculating each location's surface area from lat/lng) and ordering by this derived value. But big boxes would still make underlying boxes difficult to click.

@pellaea proposed just excluding all big locations from mapsets, because 
- they are relatively few
- they are nearly useless, and 
- they will always obscure better data. 
I believe this is the right idea. 

### First try: 1º lat/lng max box size ###

I tried excluding locations whose east-west longitude difference, or north-south latitude difference, was more than 1º either direction, and it immediately improved the legibility of the test map I used for debugging, "observations of name 5066".

Before:
`mushroomobserver.org/names/5066/map`
Screen Shot 2024-04-10 at 4 34 54 PM

After:
`localhost:3000/names/5066/map`
Screen Shot 2024-04-10 at 4 33 36 PM

However, 1º is not the same everywhere on the globe. Using a "1º rule" obscures some locations closer to the poles that may be considered marginally useful. 

For example, observations at Denali National Park and Peace River Area currently appear on the live site:

`mushroomobserver.org/names/5066/map`
Screen Shot 2024-04-10 at 4 39 56 PM

But when limiting areas to "1º" in either direction on this branch, the observations at these locations are not shown:

`localhost:3000/names/5066/map`
Screen Shot 2024-04-10 at 4 38 27 PM

I'm going to search for a better formula for areas of these boxes so we're doing orange-to-orange comparisons of geometric area.

### DIFF
--- a/app/classes/mappable/box_methods.rb
+++ b/app/classes/mappable/box_methods.rb
@@ -66,12 +66,13 @@ module Mappable
       west <= east ? east - west : east - west + 360
     end
 
-    # Arbitrary test for whether box is "vague" (i.e., covers a large area).
-    # Uses formula for the area of a patch of a spherical earth where R = 6400.
-    # area = Rˆ2 * (long2 - long1) * (sin(lat2) - sin(lat1))
-    # NOTE: Must convert degrees to radians.
+    # Arbitrary test for whether a box covers too large an area to be useful on
+    # a map with other boxes. Large boxes can obscure more precise locations.
+    # NOTE: Formula for `the area of a patch of a sphere`, lat/lng in radians:
+    #   area = Rˆ2 * (long2 - long1) * (sin(lat2) - sin(lat1))
+    # Earth R rounded to 6372km
     def vague?
-      area = 6400 * 6400 * east_west_distance.to_radians *
+      area = 6372 * 6372 * east_west_distance.to_radians *
              (Math.sin(north.to_radians) - Math.sin(south.to_radians))
 
       area.to_i.abs > 24_000
--- a/app/classes/mappable/box_methods.rb
+++ b/app/classes/mappable/box_methods.rb
@@ -81,10 +81,14 @@ module Mappable
       geometric_area.to_i.abs > 24_000
     end
 
-    # Is a given lat/long coordinate within or close to the bounding box?
+    # Determines if a given lat/long coordinate is within, or close to, a
+    # bounding box. Method is used to decide if an obs lat/lng is "dubious"
+    # with respect to the observation's assigned Location.
+    # NOTE: This is fairly strict. Larger delta makes more sense in rural areas.
     def lat_long_close?(lat, long)
-      delta_lat = north_south_distance * 0.20
-      delta_long = east_west_distance * 0.20
+      delta = 0.20
+      delta_lat = north_south_distance * delta
+      delta_long = east_west_distance * delta
       return false if lat > north + delta_lat
       return false if lat < south - delta_lat
 
--- a/app/classes/mappable/box_methods.rb
+++ b/app/classes/mappable/box_methods.rb
@@ -66,9 +66,19 @@ module Mappable
       west <= east ? east - west : east - west + 360
     end
 
+    # Converts degrees to radians.
+    def to_radians(degrees)
+      degrees * Math::PI / 180.0
+    end
+
     # Arbitrary test for whether box is "vague" (i.e., covers a large area).
+    # Uses formula for the area of a patch of a spherical earth where R = 6400.
+    # Rˆ2 * (long2 - long1) * (sin(lat2) - sin(lat1))
     def vague?
-      north_south_distance > 1 || east_west_distance > 1
+      area = 6400 * 6400 * to_radians(east_west_distance) *
+             (Math.sin(to_radians(north)) - Math.sin(to_radians(south)))
+
+      area.to_i.abs > 24_000
     end
 
     # Is a given lat/long coordinate within or close to the bounding box?
--- a/app/classes/mappable/box_methods.rb
+++ b/app/classes/mappable/box_methods.rb
@@ -66,6 +66,11 @@ module Mappable
       west <= east ? east - west : east - west + 360
     end
 
+    # Arbitrary test for whether box is "vague" (i.e., covers a large area).
+    def vague?
+      north_south_distance > 1 || east_west_distance > 1
+    end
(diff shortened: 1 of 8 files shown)


nimmolo:
Ok - I've updated the PR to calculate a consistent area when trying to determine if a location is small enough to be considered "not vague". It uses the proper formula for the area of a spherical patch, converting degrees to radians.

My arbitrary area is 24,000 sq. km. This includes "Peace River" but excludes "Denali". 
Doubling this to 48,000 sq. km will include "Denali", but also "Massachusetts". 

Feel free to help me decide what the area should be!

app/classes/mappable/box_methods.rb
```diff
@@ -66,6 +66,21 @@ def east_west_distance
       west <= east ? east - west : east - west + 360
     end
 
+    # Converts degrees to radians.
+    def to_radians(degrees)
```
JoeCohen, line 70: Would it make sense to instead extend Numeric to include a `radians` method?
Then you could write `degrees.radians`.

app/classes/mappable/collapsible_collection_of_objects.rb
```diff
@@ -106,28 +106,43 @@ def round_number(num, prec)
...
+    # if GPS points or useful locations are grouped with huge nonspecific areas.
+    def mappable_location?(is_collection, obj)
+      return true unless is_collection
+
+      if obj.location?
+        !obj.vague? && !Location.is_unknown?(obj.name)
+      elsif obj.observation?
+        !obj.location&.vague? && !Location.is_unknown?(obj.location&.name)
+      end
+    end
```
JoeCohen, line 121: If obj can only be a Location or an Observation, then 
- consider `else` instead of `elsif` 
- and/or consider adding an `else` that does some error handling or just returns `false`.

test/models/collapsible_map_test.rb
```diff
@@ -394,6 +394,18 @@ def test_mapping_one_location
...
 
+  # CollapsibleCollection should ignore large locations and observations with
+  # vague locations like "California", when mapping multiple locations.
+  # (:california is currently one of the locations for a "Boletus edulis" obs.)
+  def test_mapping_collections_ignores_vague_locations
+    bol = observations(:boletus_edulis_obs)
+    obss = Observation.where(text_name: bol.text_name).to_a
+    coll = Mappable::CollapsibleCollectionOfObjects.new(obss)
+    coll.mapsets.each do |mapset|
+      assert_not(mapset.underlying_locations.include?(locations(:california)))
```
JoeCohen, line 405: Would this work and do you like it:
```suggestion
assert(mapset.underlying_locations.none?(locations(:california)))
```

test/models/collapsible_map_test.rb
```diff
@@ -394,6 +394,18 @@ def test_mapping_one_location
     assert_obj_arrays_equal([loc], mapset.underlying_locations)
   end
 
+  # CollapsibleCollection should ignore large locations and observations with
+  # vague locations like "California", when mapping multiple locations.
+  # (:california is currently one of the locations for a "Boletus edulis" obs.)
+  def test_mapping_collections_ignores_vague_locations
```
JoeCohen, line 400: Question: I don't understand Mappable well enough (i.e., my understanding is little to none) to answer the following:
**Does this test prove anything?**
In particular, if this test were part of the current main branch, would it pass?

JoeCohen (commented):
I'm out of town till Tues, and unfortunately don't have time to drive this.
Please go ahead without me.

app/classes/mappable/box_methods.rb
```diff
@@ -66,6 +66,21 @@ def east_west_distance
...
+      degrees * Math::PI / 180.0
+    end
+
+    # Arbitrary test for whether box is "vague" (i.e., covers a large area).
+    # Uses formula for the area of a patch of a spherical earth where R = 6400.
+    # Rˆ2 * (long2 - long1) * (sin(lat2) - sin(lat1))
+    def vague?
```
JoeCohen, line 77: Will this ever get called on an "undefined" Location?
If so, will `north_south_distance` and `east_west_distance` throw Errors?

JoeCohen:
Thanks for thinking this through and doing the PR.

test/models/collapsible_map_test.rb
nimmolo: Good question. It would not pass. `locations(:california)` is one of the locations of a "Boletus edulis" observation in the fixtures, so it's included in the set of observations passed to `CollapsibleCollection` and also returned by it.

test/models/collapsible_map_test.rb
nimmolo: I generally like anything that teaches me a new syntax, and i was not aware that was equivalent.  But... hmm... i find that syntax a little harder to understand, I reckon.

app/classes/mappable/box_methods.rb
nimmolo: I like this idea. Will investigate. (Hopefully we already extend `Numeric`)

app/classes/mappable/box_methods.rb
nimmolo: I don't believe so.

nimmolo:
Added the following message to show obs, under the location, if the location is vague:
> This observation's location will not appear on maps of this taxon because it is too vague.

If the current user is the observation owner, it also says this:
> If you would like it to be included in maps, choose a more precise location.

Also added a test that these messages appear if appropriate.

app/classes/mappable/collapsible_collection_of_objects.rb
nimmolo: Revised

test/models/collapsible_map_test.rb
nimmolo: Ah. Rails adds the method `exclude?` to Enumerable and that seems just right. Edited!

PR #2092, opened 2024-04-11
Make map behavior more predictable

nimmolo:
- **All maps**: Make markers and boxes the same red color (!)
- **Occurrence maps**: Display _location boxes_ only, when obs get aggregated by `CollapsibleCollection`. 
Reason: There is a problem with my recent "improvement" on main, for Joanne's Santa Cruz Island observations. Two very close points that were previously aggregated into "USA" (and thus hard to detect) are now being aggregated into a box that is so small it is invisible at a zoomed-out view. Switching to displaying the associated Location boxes is the easiest way to bump up the box size to something noticeable. Links to the individual obs are still accessible via the box caption. 
- My goal is still to move to Google's `MarkerClusterer`, but I feel like this stopgap is much improved over what's currently on main.

Screen Shot 2024-04-11 at 4 36 31 PM

- **Single Observation map**: When there is GPS point data for an obs, favor the point, even if it's outside the bounds of the associated Location. (Increase delta for "location dubiousness" to 2.0, from 0.2). This is part of a general move towards making GPS point data definitive, and Location associations secondary.
- **Single Observation map**: Improve default zoom for points (on main, these are too zoomed in to be useful)

### DIFF
--- a/app/classes/mappable/box.rb
+++ b/app/classes/mappable/box.rb
@@ -1,7 +1,16 @@
 # frozen_string_literal: true
 
-# rectangle on the surface of the earth, with borders: north, south, east, west
-# used mostly (exclusively?) by model scopes
+# Box represents a geographic bounding box, a rectangle on the surface of the
+# earth, with borders: north, south, east, west
+# Used mostly by model scopes, and for area comparisons.
+#
+#  == Instance methods
+#
+#  valid?::              Return true if box is valid.
+#  straddles_180_deg?::  Return true if box straddles 180 degrees.
+#  expand(delta_lat, delta_lng = nil)::  Return a new box with edges expanded
+#                                        by delta (optional delta_lng)
+
 module Mappable
   class Box
     attr_reader :north, :south, :east, :west
@@ -13,6 +22,8 @@ module Mappable
       @west = west
     end
 
+    include Mappable::BoxMethods
+
     def valid?
       args_in_bounds? && south <= north
     end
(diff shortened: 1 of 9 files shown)


nimmolo:
The change in display for Observations with "point" data maybe requires explanation. When people create an observation:

- We ask people for a GPS point _and_ a Location (a database entry with a box)
- We ask people to create a Location entry if there isn't one for their GPS point, but we also default to the last location used, which may be wildly incorrect (see observation 419427, part of a foray in Idaho where the GPS on the obs is 50 miles from the associated Location.) Creating a location is time consuming and we can safely assume many people skip this step, as this user obviously did.
- We call the Location dubious if their GPS point lies outside a narrow delta from the Location box bounds. When this happens, we don't display the GPS point for that obs. We display the location, as if that were definitive!

Of course, it's the GPS point that should probably be considered definitive, not whatever Location the observation is associated with (even though GPS is not 100% reliable either). I feel like this all adds unnecessary confusion and effort to the create obs workflow, and will get in the way of simplifying the process of creating an observation.

In remote or rural places, and especially on public land, the boundary of the common-sense "postal location" is often some distance away from the coordinates of the obs. But together with the GPS, it's likely "good enough" to identify the observation location. 

Particularly if we let Google look up postal addresses in the future, we'd probably want to let the postal address be "the location", with no further ado. So this PR loosens the requirement for `lat_long_close?`. 

It's GPS data we want to encourage people to provide, not "better" entries in the Location table, in my opinion.

JoeCohen:
Maybe use a different symbol instead of a box if the the box represents only a combination of (a) any ungeolocated Observations and (b) any Observations with hidden geolcation.
Possibilities include: a circle, a different color/shape pin.

nimmolo:
@JoeCohen - good catch re: points in Klickitat County.

I just pushed a change where it lets points through if the number of objects is below the max, i.e. if there's no chance the points will be auto-aggregated and potentially become invisible, like Joanne's in Santa Cruz Island on the larger occurrence map. 

This is a definite improvement that I would have missed. Your example in Klickitat now shows points.

- On this PR, if the map has enough objects to contain _auto-collapsed boxes_, there's no way to get points at all. This is intentional, per Joanne's example.
- Unfortunately there's no logic to flag what an auto-aggregated box contains currently. It's conceivable, but we may soon retire the auto-collapsed boxes.
- As of the last meeting, my goal is to try switching to sending ONLY point data to GMaps, and let the "MarkerClusterer" do the aggregating. Observations with no GPS would send their associated location's center point, as we discussed. 
 
This will only work if Google somehow provides separate captions for coinciding points, because there will be plenty of these. If that's not possible, we're stuck with boxes for the long term, and I can invest more time into differentiating them.

nimmolo:
@JoeCohen The question in #2088 is my two long illustrated posts at the bottom, regarding maps for single observations. This PR changes the default to favor the geolocation over the associated MO location in all cases.

Issue #2093, opened 2024-04-12
Observation has_location


JoeCohen:
Observation pattern search `has_location` variable is broken.
Example:
This pattern:
`Sarcosphaera coronaria region:"Klickitat Co.\, Washington\, USA" has_locations:yes`
yields
http://localhost:3000/observations?pattern=Sarcosphaera+coronaria+region%3A%22Klickitat+Co.%5C%2C+Washington%5C%2C+USA%22+has_location%3Ayes
which return 7 Observations, the last 2 of which lack geolocations.

nimmolo:
@JoeCohen I don't think that parameter checks for lat/lng
```ruby
# app/classes/query/observation_base.rb
    def initialize_has_location_parameter
      add_boolean_condition(
        "observations.location_id IS NOT NULL",
        "observations.location_id IS NULL",
        params[:has_location]
      )
    end
```

nimmolo:
Checking through that file, we don't yet have a search param for "has_lat_long". You could probably write one pretty easily, though.

JoeCohen:
@nimmolo Thanks for looking at the code ... somthinhing I should have done.
The in-app documentation says
Thanks for checking. That's a surprise to me.
The [in-app documentation](https://mushroomobserver.org/info/search_bar_help) says

> **has_location**: Are the Observation’s Latitude and Longitude fields filled in? 

I'm probably the one who misdocumented it.

nimmolo:
I'm starting a PR to add this search param.

The misunderstanding arises from an inaccurate translation string:
```
  observation_term_has_location: Are the [:OBSERVATION]'s [:LATITUDE] and [:LONGITUDE] fields filled in?
```
It should be something like:
```
  observation_term_has_location: Is the [:OBSERVATION] associated with a defined [:LOCATION]?
```

nimmolo:
Whoops, cross posting! Just getting your comment. PR coming...

PR #2095, opened 2024-04-12
Replace observation search param `has_location` with `has_public_lat_lng`, fix search instructions

nimmolo:
Until now, we don't have a parameter to search for obs with geolocation. We do however have a query param `has_location` which is currently not very useful. Addresses #2093. 

Now incorporates #2096, renaming `long` to `lng` everywhere, because it's tangentially relevant to this PR.

#### Questions: ####
- should a query like this also check for `gps_hidden`? I believe so.
I had thought not, because on my checkpoint the lat/lng seems to have been stripped where `gps_hidden`. But Joe pointed out that the lat/lng is saved on the obs even when `gps_hidden`, on production.
- should it instead be called `has_public_geolocation`?


nimmolo:
Please vote for a name for this param:
- `gps_public`
- `has_geolocation`
- `has_public_geolocation`
- `has_lat_lng`
- `has_public_lat_lng`

pellaea:
My 2¢ = `has_public_lat_long` because that is least ambiguous.

nimmolo:
Per email conversation with @pellaea it seems the `has_location` param is only useful for finding undefined Locations.  That query returns only one observation in the entire database with an undefined location currently (and it's from the town I live in!)

nimmolo:
@JoeCohen should we retire `has_location`?

JoeCohen:
param name:
I'm good with whatever others like. If it were up to me I'd use something without "has" (will avoid one rubocop issue).
maybe `with_public_geolocation`

I'm OK with retiring `has_location`.

nimmolo:
@mo-nathan is the `has_location` query param useful for projects, or anything you're working on? 

In case you missed the discussion on this, `has_location` only checks for an associated location, which of course could be "Alabama, USA", so it's not so useful, except for finding obs with undefined location. 

The current use people seem to have for something like this, however, is trying to find observations that "have a precise and public GPS lat/long". Until now, the search instructions for the `has_location` param erroneously suggested this was what it did.

mo-nathan (approved):
Looks good to me.

JoeCohen (approved):

PR #2096, opened 2024-04-12
Switch Observation column `long` to `lng` everywhere

nimmolo:
If everyone likes the new query param `has_geolocation` in #2095, I'd like to take this opportunity to merge this change  immediately afterwards, before deploying. 

The point of this PR is to help avoid confusion when dealing with the `observation.long` attribute. Google Maps is the main use for the attribute, and GMaps only accepts a `lng` property. It's a perpetual gotcha to remember making this switch. Keeping the param name the same everywhere (except in translation string labels) should eliminate the mental overhead.

Fortunately the API handles this property as `longitude` and only converts to Observation model attributes when creating/updating an observation, so this is a non-breaking change for the API.


JoeCohen (approved):
I'm OK with this.

PR #2099, opened 2024-04-15
Remove default values for UserStats created_at and updated_at

mo-nathan:
The original UserStats migration was using Time.zone.now as the default for the created_at and updated_at fields.  This creates instability with db/schema.rb since any time the migration gets run it will use a different value.  This will create a maintenance burden for developers.  This PR eliminates that need since I see no evidence that it is required despite the title of this commit: https://github.com/MushroomObserver/mushroom-observer/commit/43d8757914c204d4c084e2d82bc620075f40d0e9.  If it required (please provide documentation), then we should set it to some explicit date like '2024-01-01 00:00:00' rather than a dynamic value like Time.zone.now.


JoeCohen (approved):

nimmolo (approved):

nimmolo:
Thanks for doing this.

PR #2104, opened 2024-04-17
Add Stimulus for user feedback showing FieldSlip PDF job status

nimmolo:
New Stimulus controller to give on-page feedback on the progress of generating PDFs.

For returning data from the `FieldSlipJobTracker`, this adds a new AJAX endpoint: a tiny `FieldSlipJobTrackersController` with just a `show` action. It returns the JSON for time, pages, and status.

New intended workflow
- User creates a PDF with x number of pages
- The `Projects::FieldSlipsController` `:create` action responds with a `turbo_stream` that prepends a new row (because they're most recent first) to the table of field slip PDF jobs, instead of redirecting to the Project `:show` page
- The new row is "noticed" by the Stimulus controller, which starts a timer to check for updates
- Stimulus checks the new `FieldSlipJobTrackersController` `:show` endpoint for the current status of the tracker every second
- The `:show` endpoint responds with a `turbo_stream` that updates the row, until the `status` is `done`


nimmolo:
@mo-nathan — If you can add the "pages" data to the tracker, i'll add it to the controller here.

nimmolo:
New approach - use Turbo!

This use case is actually a good demonstration of how Stimulus can be simpler with Turbo, and how it requires a simpler way of thinking about JS.

Instead of writing the Stimulus controller to explicitly update the `targets` in each row, our `FieldSlipJobTrackersController` can just return a `turbo_stream` response with the refreshed HTML for the entire row (identified by DOM id), and Turbo will update the row without us handling all the HTML in JS.

nimmolo:
@mo-nathan i'm trying to play with this, have started solid_queue locally. 
But i'm blanking on this step, do I need to create this project?
 
3. Go to a project with a field slip prefix.

nimmolo:
Nevermind! got it.

nimmolo:
Workflow description updated above. 

Current issue: it's not updating old jobs. Somehow they're done, but the status attribute is not getting updated when that happens:

```
irb(main):011> fj = FieldSlipJobTracker.find(4)
irb(main):011> fj.attributes
=> 
{"id"=>4,
 "start"=>54,
 "count"=>6,
 "prefix"=>"PJ1",
 "status"=>"Starting",
 "created_at"=>Wed, 17 Apr 2024 17:46:07.150271000 EDT -04:00,
 "updated_at"=>Wed, 17 Apr 2024 17:46:07.150271000 EDT -04:00}
irb(main):012> fj.done
  TRANSACTION (0.2ms)  SAVEPOINT active_record_1
  FieldSlipJobTracker Update (0.6ms)  UPDATE `field_slip_job_trackers` SET `field_slip_job_trackers`.`status` = 3, `field_slip_job_trackers`.`updated_at` = '2024-04-17 22:20:11.809895' WHERE `field_slip_job_trackers`.`id` = 4
  TRANSACTION (0.1ms)  RELEASE SAVEPOINT active_record_1
=> true
```

JoeCohen:
@nimolo
- If you have a recent db snapshot, you can use /projects/341 or /projects/334
(FWIW, I've been using the uncleaned recent snapshots, rather than the Oct. clean snapshot. 
The disadvantage is that you can't login as someone else.
My workaround is a second account that uses a different email.)
- Else just create a new project (including a Field Slip Prefix)

nimmolo:
Polling the endpoint is making a LOT of requests for me locally because my jobs aren't getting marked as "done", and although this is a development bug, I figure in the case of any glitch, we don't want to tie up someone's browser with zombie requests like this.

I believe the preferred way to do this, rather than polling the endpoint, is to make the `FieldSlipJobTracker` model [broadcastable](https://github.com/hotwired/turbo-rails/blob/main/app/models/concerns/turbo/broadcastable.rb). Turbo broadcasts are something I wanted to try anyway, and they were created for exactly this situation.

app/views/controllers/projects/field_slips/new.html.erb
```diff
@@ -1,41 +1,39 @@
 <%
 add_page_title(:field_slips_for_project_title.t(title: @project.title))
+@container = :text_image
```
mo-nathan, line 3: What does this do?

app/controllers/field_slip_job_trackers_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class FieldSlipJobTrackersController < ApplicationController
+  before_action :login_required
+
+  # This is only a JSON endpoint describing the status of a particular tracker.
+  def show
+    # should we raise an error if the tracker is not found?
```
mo-nathan, line 8: Seems reasonable to raise an error, but where would it show up?

nimmolo: See Slack for a possible new strategy using ActionCable broadcasts.

mo-nathan:
The new row isn't showing up until I refresh the page.  And I'm seeing bunch of tiny lines with text show up at the bottom of the page.
Screenshot 2024-04-18 at 9 22 50 AM

nimmolo:
Yes - I missed an ID change. Updated!

mo-nathan:
That gets the row in, but it still doesn't update every second.  I just ran a batch of 30 and it showed the initial row, but then sat there without doing any updates.  When I eventually (> 10 seconds) refreshed the page, it reported that it is Done and took a total 5.9 seconds.

nimmolo:
Yeah, that's what's confusing me too. I thought maybe the tracker instances were not getting freshly retrieved, so I tried reloading the records in the controller, and that didn't work either, so then i debugged and checked the records directly in the console. 

Whenever I did that, they hadn't budged past "Starting".

mo-nathan:
If they aren't getting past "Starting" in the database, that means SolidQueue is not running.  Remember you have to restart the SolidQueue process if you change code at all even in a completely unrelated part of the system.  I'm not sure why that makes it not work, but it's very different from working with the rails server.

nimmolo:
Ahaa. Interesting. Thanks!

I wonder if Solid Queue can work with the [`foreman`](https://github.com/ddollar/foreman) gem. Ok - [yes it does](https://www.bigbinary.com/blog/migrating-to-solid-queue-from-sidekiq#3-starting-solid-queue). 

Foreman is a gem that runs a script with a proc to start and stop processes and compile assets. It is the default way to run Rails dev servers currently. The reason I didn't switch to it yet was because it requires us to use a different command to start the dev server: `/bin/dev` instead of `rails s`. I didn't want to confuse people. 

But check that link out when you get a chance, (you too @JoeCohen and @pellaea) and see if you think switching to foreman is a tolerable option to simplify running the dev server. 

A small bonus is that it should clear up our scss deprecation warnings on CI.

JoeCohen:
foreman is above my pay grade. I didn't understand anything.
I do worry about moving away from the Rails main stream.

nimmolo:
Whew! Debugged the stimulus. 

It turns out it works with `async`, without making too many requests. The "too many" part came from my bad syntax with `setInterval` - we should only pass the function name, not the executed function results: in other words we should pass `setInterval(this.do_thing, 1000)` not `setInterval(this.do_thing(), 1000)`. 

The other trouble was that we can't call a class function (`this.do_thing`) within `setInterval`, because we lose the context of `this`. It is possible within an IIFE as we have here, but there's no need. Our function is very short. 

There's also a new Turbo partial that updates the row. Rather than trying to cram the `render(turbo_stream: turbo_stream.replace(partial: "thing", locals: hash))` into the controller, i just put that in a partial that calls the other partial, as I've done elsewhere. That's the easiest syntax that I've gotten to work.

nimmolo:
I think all this needs now is `pages` for the tracker.

mo-nathan:
Sounds great! Thanks! I’ll check it out in the morning and add the pages
bit.

nimmolo:
The remaining errors seem to relate to the changes in how the `create` action works now.

I'll wait for the `pages` attribute before trying to sort it out.

PR #2107, opened 2024-04-20
Live comment updates with ActionCable

nimmolo:
This PR lays the groundwork for live updates for any object via `turbo_stream`. 

The first place I'm implementing them is Comments. Comments seem like a low-stakes place to test and get it right — it's not that they are so useful, but they're easy to test. 

This PR is intended to clear a path for other uses of live updates on the app. The ones I'm more interested in are namings, user notifications, images (for image transformations that are run as a job) and observations, for the home page. These may require a bit of refactoring and renaming some templates.

Incredibly, it took only a few lines of code, but the configurations took quite some time to figure out. 

Comments was the easiest to do because the comments panel has the same structure everywhere, and uses conventionally named partials and ivars.

#### Trying it out locally: prepare your local server ####
1. [Install `redis`](https://redis.io/docs/latest/operate/oss_and_stack/install/install-redis/) server software locally. 
Linux: `sudo apt-get install redis` / Mac: `brew install redis`. You may also have to start the server: `redis-server`
2. Install the gems on this branch, `bundle install` (installs `puma` locally)
3. In one terminal window, run `./bin/dev` or otherwise start Solid Queue - creates and updates are run as background jobs.
4. In another terminal window, run `rails server`
5. In a third terminal window, test your redis server connection. Run `rails console --sandbox`, then
```ruby
redis = Redis.new
redis.ping
```
You should get a "PONG". If so, you can `exit`

6. In two different browsers, say, Firefox and Chrome, or in a regular window and an "incognito" window, log in as two different users. Visit the same observation.
7. Start posting comments on the observation in either browser.
8. The comments should appear pretty quickly in the other browser without a page refresh, even if you have a comment form open.

Notes:
- This will require installing redis on production


mo-nathan:
Ran into a problem testing based on the above instructions.  After install redis, I think you have to start the server with `redis-server` in another shell.  This could probably get added to the bin/dev stuff.  Without it running I wan't able to get the "PONG" described above.  Once it was running it seems to work.

mo-nathan:
Looking over the code, it's pretty slick.  Do you know if ActionCable is smart enough to only send messages to browsers that are currently looking at a relevant comment or is it essentially sending to everyone and letting the JS decide if it's relevant or not?  The first approach seems a bit less reliable, but the second approach sounds like potentially a lot of traffic.

nimmolo:
Smart enough to only send messages to browsers that are currently looking at a relevant observation - yes. 
But it's Turbo. The "subscription" part comes from us putting a `turbo_stream_from(channel_name)` in the template.

nimmolo:
(updated my answer on Github with much more detail, it unfortunately won't come through in the email thread.)

config/puma.rb
```diff
@@ -24,7 +24,13 @@
...
+    begin
+      ActiveRecord::Base.connection.disconnect!
+    rescue StandardError
+      ActiveRecord::ConnectionNotEstablished
+    end
+    ActiveRecord::Base.establish_connection(
+      YAML.load_file("#{app_path}/config/database.yml")[rails_env]["primary"]
+    )
   end
 end
```
nimmolo, line 36: @mo-nathan I can't figure out these test failures. CI is getting "invalid authenticity token".

These tests do not fail locally and they're only failing on CI "push" - note that CI "pull request" passes. They are mostly not system tests and so they don't seem related to solid queue and there are no changes in test environment configs.

There is a change however here in `config.puma.rb`, to run puma as the test server. This was necessary to get the system tests to use Puma (and thereby have jobs/cable avaliable), otherwise the server wasn't set up on port 3000. (See also below in `test/application_system_test_case.rb:28` `Capybara.server = :puma`)

nimmolo: Fixed it. It just needed to not have the port set, in test environment.

nimmolo:
> Ran into a problem testing based on the above instructions. After install redis, I think you have to start the server with `redis-server` in another shell. This could probably get added to the bin/dev stuff. Without it running I wan't able to get the "PONG" described above. Once it was running it seems to work.

Added to the instrux

nimmolo:
Now deployed on `test.mushroomobserver.org`. Not working yet.

Steps taken so far:
- [Install redis](https://redis.io/docs/latest/operate/oss_and_stack/install/install-stack/linux/)
- `redis-server`
- `sudo sysctl vm.overcommit_memory=1`

nimmolo:
This is now working in a "production" environment at `test.mushroomobserver.org`. You can log in as yourself in a regular and "incognito"/"private" window and check it out.

Be careful if testing... any comments you post as tests will also post to the production database.

nimmolo:
### Issues
- Comments are created with no controls, until the page is refreshed. 
  - This is because the current user is not set when a comment is created via broadcast
  - Controls are shown via `check_permission(comment)` - maybe refactor to send local `controls` separately
  - Action cable doesn't access the user as `User.current` but as `connection.current_user`

nimmolo:
Recent changes:
- Moved broadcast methods to the controller, from the model. There seems to be an emerging consensus that broadcasting from the model is too automagical and a confusing violation of MVC separation of concerns. (The reason it is possible is because the model knows about CRUD actions and is therefore in a position to broadcast without intermediaries.)
- Added system tests with two sessions, to check broadcasts

Issue #1945, opened 2024-02-22
Locations with hidden GPS


mo-nathan:
Use case: Foray on private land at a NAMA foray.

Locations get a new boolean value "hidden".
Only the owner (or an MO admin) can change the hidden boolean.  By default it is false.
When an observation is associated with a location, if the location is hidden, then the observation's GPS hidden boolean gets set to false.
Hidden bit cannot be changed if there are associated observations.
The coordinates of a hidden location are automatically rounded to the nearest 1/10 of a degree that makes the area larger.
Being the owner of a hidden location does not in and of itself grant the ability to see the GPS of associated observations.

nimmolo:
This makes sense, i think.

mo-nathan:
Yes.  I don't think it's worth maintaining enough state to do anything different than that.  I think it's a rare edge case with a reasonable way to address.

mo-nathan:
I'm wondering if only MO admins should be able to set a location to hidden if there are already observations associated with it.  Unintentionally hiding a location with a lot of existing observations could make a big mess.

PR #2121, opened 2024-04-26
Hidden locations

mo-nathan:
To test:
1) Create a new Location and set it hidden (checkbox at the top of the edit form)
2) Note that lat/long values cannot be more precise than 0.1 degrees.
3) Associate an observation with lat/long with this location.
4) Note that the observation is forced to have it's GPS hidden.

See #1945 for more notes/discussion.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -552,7 +552,9 @@ class LocationsController < ApplicationController
 
   def permitted_location_params
     params.require(:location).
-      permit(:display_name, :north, :west, :east, :south, :high, :low, :notes)
+      permit(:display_name,
+             :north, :west, :east, :south, :high, :low,
+             :notes, :hidden)
   end
 end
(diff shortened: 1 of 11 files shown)


nimmolo (approved):

JoeCohen (approved):
Here's one interesting hypothetical:
A hidden Location with Observations is merged into a non-hidden Location. (Note that only Admins can do this.)
- The Observations' GPS will become public. 
- Should this merger even be allowed?

mo-nathan:
> Here's one interesting hypothetical: A hidden Location with Observations is merged into a non-hidden Location. (Note that only Admins can do this.)
> 
> * The Observations' GPS will become public.
> * Should this merger even be allowed?

Yeah, I thought about it a bit and ended up feeling we shouldn't limit what admins can do.  Putting in a warning might be worthwhile, but it feels pretty edge casey to me.

PR #2124, opened 2024-05-01
Adjust script/extract_log to handle non-utf8 sequences.

pellaea:
The "real" problem is why are there non-utf8 sequences in the log???  But be that as it may, code should never crash, as a general rule, and I can at least fix that...

### DIFF
--- a/script/extract_log
+++ b/script/extract_log
@@ -46,6 +46,7 @@ end
 
 file = "#{app_root}/log/production.log"
 File.open(file).readlines.each do |line|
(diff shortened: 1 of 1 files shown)


mo-nathan:
Thanks for fixing this.  Given all the monkeys banging on all the keyboards, I can see non-utf8 sequences ending up in the log.  I can specifically imagine monkeys with bad intent throwing binary data where it shouldn't go.

pellaea:
:)  Yep, I'm not overly concerned.

PR #2129, opened 2024-05-03
Bundle update

JoeCohen:
On general principles:  update often


JoeCohen:
@nimmolo Will this break anything you have planned?

nimmolo:
I don't think so! Thanks for asking.

nimmolo:
@JoeCohen 
Ok if I merge the "puma in dev" branch?  This bundle update does not seem to update Puma, so it doesn't seem to matter which goes first.

PR #2130, opened 2024-05-04
Image scripts in Ruby

nimmolo:
This is an experiment to redo the `process_image`, `retransfer_images` and `rotate_image` scripts in ruby.

They share a bunch of methods.


nimmolo:
Doing the "job" part separately from the "ruby" part now.

PR #2132, opened 2024-05-04
Enable PDFs with One Field Slip Per Page

mo-nathan:
Field Slip PDF generation page now has a checkbox to generate PDFs with a single slip per page rather than 6 which is better for US letter paper.  The single slip per page option is preferred by at least some print shops.


JoeCohen (changes requested):
- "duff" and "Other" overlap.
Screenshot 2024-05-04 at 10 48 26 AM

- I don't know if the print shop would want you to add cut lines.

JoeCohen:
PS: Should "Habit" be aligned with Odor, Substrate, Plants?

JoeCohen:
Question: Is "Other Codes" where participants should put iNat ID?
Maybe **in a different PR** let admins customize part of the slip???

mo-nathan:
Good catch on the Substrate.  It's actually "dung" that's getting word wrapped.  I don't understand why it's word wrapping only in the single page mode, but I think I can fix it.  The Habit is intentional to give folks more room to write down nearby plant species.  It used to be where the "Photo" checkbox is that Rick wanted.  "Other Codes" serves two purposes.  1) Folks can write the iNat code there.  2) John Plischke can use it for his voucher labels.  I couldn't figure out a better way to handle both those cases.  If there's iNat code and John Plischke wants to put his voucher label there, then he can either copy the iNat code to somewhere else or he can assume that we already have it in the database since I think it will have gone through the recorder at that point.  I don't see a great way to give admins a lot of flexibility in the slip.  The space if very constrained and I don't really want to create a specialized form editor that is expected to give people flexibility while still looking decent and being easy to correlate back to the form.  I know Anne mentioned putting specific Amanita stuff on the form, but I haven't talked to her about exactly what she's looking for.  I'm guessing that's going to force the form up to a 4/page format which could be a time to provide more flexibility.

JoeCohen (approved):

app/views/controllers/projects/field_slips/new.html.erb
```diff
@@ -9,12 +9,13 @@
...
-    <%= :field_slips_for_project_max_pages.t(max: @page_max) %>
+    <%= :field_slips_max_for_project.t(max: @field_slip_max) %>
   </div>
   <%= form_with(url: action, data: { turbo: true },
                 id: "project_field_slip_form", class: "my-3") do |f| %>
-    <%= number_field_with_label(form: f, field: :pages, inline: true,
-                                value: @page_max.zero? ? "0" : "1") %>
+    <%= number_field_with_label(form: f, field: :field_slips, inline: true,
+                                value: @field_slip_max.zero? ? "0" : "6") %>
+    <%= check_box_with_label(form: f, field: :one_per_page, label: :field_slips_one_per_page.t) %>
```
nimmolo, line 18: Got here late, but this looks good

nimmolo (commented):
👍

Issue #1950, opened 2024-02-22
Field Slip Forms


mo-nathan:
Allow creation of forms for Field Slip data.  These forms would be associated with a Project and would be used on the create/edit page for an associated Field Slip.  Should support strings and checkboxes (habitat).

mo-nathan:
I've been thinking about this more and I think this is important for the recorder role at mushroom events.  The recorder should be able to scan the QR code and immediately go to a page that let's them transcribe any data that's on the physical field slips.  This is actually kind of different from the observation per se.  You still want to link to an existing observation and ideally the data is the same, but I can see them being out of sync due to human error.

mo-nathan:
I think project admins for a field slip should go to the edit field slip page which should have all the fields from a field slip.  It would be cool if it was a show & edit page.  Not sure those are still fashionable.

mo-nathan:
Planning on reusing the concept of a notes template from users/observations only for projects/field slips.  The template would by default have the fields for the current field slip.  By default it would sync with the notes for the associated observation.  The edit form page should have a way to add images to the associated observation and create the observation if it doesn't already exist.  Should completing the field slip form create the observation?  Might not have any images yet.  Pretty sure the answer is yes.  Would want to allow the collector to create the observation/add to the observation.  Should provide a way to pull data from another MO (or iNat) obsie.  Admin or field slip owner can move the field slip to point at a later obsie.

mo-nathan:
Plan:

- [ ] Add note template field to field slip model
- [ ] Provide note template UI for field slip edit page
- [ ] Populate from observation if available
- [ ] Push changes from field slip to observation if possible
- [ ] Push changes from observation to field slip

mo-nathan:
Having looked a bit at the idea of using notes templates as the primary approach for field slips, I don't think it makes sense outside of the "Notes" field.  All the other fields have corresponding MO objects (users, locations, names) that would be reasonable to optionally associate with the field slip data if it exists.  I'm thinking through what sort of autocomplete options to provide.  Does it make sense to involve the create location and name workflow if they enter values that aren't already in MO?  I think we have to allow that for people since we cannot assume that the collector or identifier have MO accounts.  It would also be helpful for the identifier initials and the locations to be easy to find, but maybe that's a follow on feature.

mo-nathan:
[...]

PR #2139, opened 2024-05-15
Add notes to field slips

mo-nathan:
Related to #1950.


mo-nathan:
Going a different direction

PR #2141, opened 2024-05-16
Field Slip system test adjustments

nimmolo:
Refactors the `field_slips_test` as integration tests. Should run faster and be just as effective.

#### Background: 
With system tests, they're slow and we only really need them if we're testing JS.


nimmolo:
Update: #2107 passes now, so i don't consider this necessary. 
It does however have some things we may want to keep:

- Adds identifying selector classes to the important links on Field Slips, so **the system test never has to use `xpath`**. xpath selectors tend to be brittle if HTML is reorganized.
- The test runs 30-50% faster (3-5s for me locally, and similar difference in CI) when all strung together, no matter how many times I re-run it. It's a small gain here, but i think does illustrate that there's a material difference when consolidating a system test workflow.
- Parens for method args is MO's Rubocop style across the whole app. I feel like if we're going to have a style...

nimmolo:
@mo-nathan — I just converted your original short tests to integration tests. I think this preserves your original intent. I think we only really need to write system tests when there's JS we want to test.

Note the addition of selector classes to the templates, so we don't have to use XPATH selectors in our tests - much easier!

nimmolo:
@mo-nathan I am realizing this will affect your #2160 Field Slip Forms PR. 

I believe it is an improvement on the form HTML, particularly the selectors, and makes the tests less brittle. If you like, I'll be happy to check #2160 over and reconcile these changes with that PR.

mo-nathan (approved):
Looks good to me!

nimmolo:
Ok thanks, I will merge. Just say the word and I'll either reconcile it with #2160, or offer suggestions - I realize this creates extra work on a long-running PR of yours and that you're probably not wanting any added confusion with it.

PR #2150, opened 2024-05-19
iNat Import

### LINKED ISSUE #1955: iNat import
This task embodies the simplest acceptable functionality for NEMF 2024

### Tasks
- [ ] Add MO project to Obs if it meets project criteria && user is a member
- [ ] Perhaps change redirect from Observations to Observations by current user.
- [ ] Rubocop

### Completed
- [x] iNat observations/photos with restrictive licenses (e.g. nonderivs, Options: Don't import; require MO user consent).
- [x] Remove import all option from form
- [x] Fix conflicts with main.
- [x] #2341 
- [x] Different flash if obs ##s missing
- [x] #2219 
- [x] #2246 - [x] Import single iNat Observation (in foreground, unauthenticated)
- [x] Isolate call to MO API instead of conditioning images on non-test environment.
- [x] Change API Key to something temporary that I can delete after fixing environment variables
- [x] Add production MO API Key for webmaster
- [x] ?? Slime molds: Better solution than `iconic_taxon_name: "Protozoa"`? probably too complicated:  ("Slime molds" are polyphyletic).

JoeCohen:
PR aims to implement the minimum acceptable functionality for NEMF for importing iNat observations
- Allows an MO user to import a list of that User's iNat observation(s) to MO via the iNatImport form.
- Implements the [Basic Inat Import milestone](https://github.com/MushroomObserver/mushroom-observer/milestone/3)

### Note
- It uses an inefficient way of giving users feedback about jobs. I didn't know how to do the js-Turbo 🪄 voodoo. I think that's enough for NEMF because I removed the UI for importing all a user's observations.
- It doesn't rate limit. I think that's OK for NEMF for the the same reason.

### Questions for reviewers
- Is more functionality needed?
- Appearance
- General organization
- Better names for classes, methods, variables, ex: ImportedInatObs
- Any ideas about shortening/dividing long classes/methods
- Any ideas about fixing ABC violations
- Any other criticism is welcome.

### Manual Test Protocol
- Create a sacrificial iNat observation, making sure it has lat/lng. I suggest also selecting Obscured Geoprivacy.
- In MO: Make sure SolidQueue is running, Create Observation, Import from iNat http://localhost:3000/observations/inat_imports/new
- Fill in the fields, using the iNat observation # of the sacrificial iNat obs, and Submit
Result: redirects to Observation Index (to be change to Inat Import Job Tracker)
- Wait 10 seconds, then refresh screen
Expected result: Index includes a new Observation
- Goto the new Observation. and review it. The Location should make sense. If the iNat obs had Obscured Geoprivacy, then MO coordinates should be hidden from the public.
- Goto/refresh the sacrificial iNat Obs
Expected result: 
Description includes imported to MO and date of the import; 
Observation Fields include Mushroom Observer URL linking to the newly created **Local** observation. 
(In the production environment, it links to the live db.)
- Delete the sacrificial iNat Obs


nimmolo:
Did you try adding the key to Rails credentials yet? It's very easy and the steps for decoding aand re-encoding the credentials are documented online.

JoeCohen:
Thanks for the follow up.
No. it looks easy. But for my case, I need something
additional/diferent which I was able to do without using credentials.
I might refactor later.

mo-nathan:
It's working well.  It doesn't seem like it should need me to type my iNat username every time.  Seems like this could be saved in the MO database after the first authentication cycle.

mo-nathan:
Since the Turbo/JS magic isn't updating the page, I think the page should say something about manually refreshing the page.  Maybe after the "Status" if the status isn't "Done", it could say "Refresh to monitor the status" or something along those lines.

JoeCohen:
How about instead changing the name of the button?
> Since the Turbo/JS magic isn't updating the page, I think the page should say something about manually refreshing the page. Maybe after the "Status" if the status isn't "Done", it could say "Refresh to monitor the status" or something along those lines.

app/jobs/inat_import_job.rb
```diff
@@ -0,0 +1,408 @@
...
+    update_names_and_proposals
+    add_inat_sequences
+    add_snapshot_of_import_comment
+    # NOTE: update field slip 2024-09-09 jdc
+    # https://github.com/MushroomObserver/mushroom-observer/issues/2380
+    update_inat_observation
+    increment_imported_count
+  end
+
+  def create_observation
```
mo-nathan, line 152: For this ABC metric violation, I consider implementing something like `create_hash` on the the `Inat::Obs` that returns a hash which you could then add in the other stuff with a merge and pass the resulting to `create` rather than the big long list of key value pairs.

app/jobs/inat_import_job.rb
```diff
@@ -0,0 +1,408 @@
...
+  end
+
+  def add_snapshot_of_import_comment
+    params = { target: @observation, user: inat_manager,
+               summary: "#{:inat_data_comment.t} #{@observation.created_at}",
+               comment: import_snapshot }
+    Comment.create(params)
+  end
+
+  def import_snapshot
```
mo-nathan, line 345: For this ABC metric violationm I'd move most of this into the `InatObs` object as a `comment` method or some such.  That would probably go a long way towards addressing this since you'd no longer be doing all those dereferences.

app/controllers/observations/inat_imports_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+    REDIRECT_URI =
+      "http://localhost:3000/observations/inat_imports/authorization_response"
+    # iNat's id for the MO application
+    APP_ID = Rails.application.credentials.inat.id
+    # The iNat API. Not called here, but reference in tests and ActiveJob
+    API_BASE = "https://api.inaturalist.org/v1"
+
+    def new; end
+
+    def create
```
mo-nathan, line 50: For this ABC metrics violation, I would start with moving all the error checking into it's own method so it could be a single lline in this method.

app/classes/inat/obs.rb
```diff
@@ -0,0 +1,463 @@
...
+#
+class Inat
+  class Obs
+    def initialize(imported_inat_obs_data)
+      @obs = JSON.parse(imported_inat_obs_data, symbolize_names: true)
+    end
+
+    ########## iNat attributes
+
+    def inat_description
```
mo-nathan, line 63: I'd start by loking for a way of getting rid of this pattern,  This seems very repetitive.

mo-nathan (changes requested):

JoeCohen:
@mo-nathan: Thanks so much for looking at this and for all the suggestions.
I'll work on them tomorrow (after dealing with the new conflicts).

JoeCohen:
@mo-nathan Here's what I did to the tracker page. I don't have strong feelings about this, so am happy to make any other changes you'd like. Some options:
- Change the button name to  "Refresh to monitor the status", "Click to update status" or anything else.
- A flash  like "Import started. Click the Update button to update status"when the user (first) lands on the page.
- Additional text like what you proposed.

Screenshot 2024-09-17 at 2 37 22 PM

> Since the Turbo/JS magic isn't updating the page, I think the page should say something about manually refreshing the page. Maybe after the "Status" if the status isn't "Done", it could say "Refresh to monitor the status" or something along those lines.

JoeCohen:
Saving and auto-filling the iNat username is implemented as of Commiit #[276338c](https://github.com/MushroomObserver/mushroom-observer/pull/2150/commits/276338c2033d4e8a948cf795fe952de096f55060)

PR #2154, opened 2024-05-25
Licenses controller

### LINKED ISSUE #2151: Add LicensesController
- Update MO's Creative Common licenses to v.4
- Add new licenses if appropriate

### License Controller TODO
- [x] Index
  - [x] Links
- [x] Show License
  - [x] Admin Only
  - [x] Links to: Add, Edit, Destroy if unused
- [x] New
  - [x] `Deprecated?` checkbox
  - [x] Links
- [x] Create
  - [x] Validate
- [x] Edit
  - [x] Links
- [x] Update
- [x] Destroy
- [x] Link in Admin Menu (lh navbar) to Index
- [x] Rubocop
- [x] Update documentation
  - [x] Translations
  - [x] Readmes
- [x] Update badges https://creativecommons.org/mission/downloads/
  - [x] Fix badges for existing licenses. They're not broken. It's a problem with my local settings. 
  - [x] Add badges and forms for new licenses. Ex: app/views/controllers/shared/_form_ccbyncsa25.erb
- [x] Consider new unicode symbols? https://creativecommons.org/2020/03/18/the-unicode-standard-now-includes-cc-license-symbols/ **Won't do**
- [x] Validate uniqueness of display_name, url
- [x] Flash validation errors.
- [x] params.deprecated: `true`/`false` or `1`/`0`
- [x] QUESTIONS
  - [x] no `created_at`, many lack `updated_at`
- [x] Add `created_at`
- [x] single, parametized, license badge partial to replace app/views/controllers/shared/_form_ccbyncsa25.erb, et al. See https://licensebuttons.net/
- [x] Revise index display. See https://github.com/MushroomObserver/mushroom-observer/pull/2154#discussion_r1623392075; also include "Default" if default license
- [x] Revise show display. https://github.com/MushroomObserver/mushroom-observer/pull/2154#issuecomment-2143816910; also include "Default" if default license
- [x] Revise form, showing Default if default license
- [x] Coverage of License model and Licenses controller
- [x] Rubocop
- [ ] Merge & Deploy
- [ ] Follow up with #2151

JoeCohen:
- Adds an admin-only UI with standard CRUD actions for Licenses.
- Updates and rewrites License portion of `/info/how_to_use` to include new licenses

#### Some Suggested Manual Tests
- go into Admin mode
- click License in the sidebar
- view various Licenses
- add a new License, then destroy it
- show an existing License. Note lack of a Destroy link.
- Edit some Licenses to: blank out an attribute; to duplicate another License's attribute.


app/views/controllers/shared/_form_ccby40.erb
```diff
@@ -0,0 +1,9 @@
+<%=
+tag.div(id: "license") do
+  link_to("https://creativecommons.org/licenses/by/4.0/", rel: "license") do
+    image_tag("https://i.creativecommons.org/l/by/4.0/88x31.png",
+              alt: "Creative Commons Attribution License",
+              class: "border-none")
+  end
+end
+%>
```
JoeCohen, line 9: Template for license badge. 
This and the following templates anticipate adding these licenses to the db.

test/integration/capybara/licenses_integration_test.rb
```diff
@@ -0,0 +1,14 @@
...
+# Test things that are untestable in integration tests
+class LicensesIntegrationTest < CapybaraIntegrationTestCase
+  def test_link_to_licenses
+    login(users(:admin))
+    assert_no_selector(:link, text: :LICENSES.l)
+
+    first("button", text: "Turn on Admin Mode").click
+    assert_selector(:link, text: :LICENSES.l)
+  end
+end
```
JoeCohen, line 14: Can I do this in a functional test:
- Prove that when you're in admin mode, the sidebar has a `Licenses` link.

app/models/license.rb
```diff
@@ -45,15 +48,19 @@ class License < AbstractModel
   has_many :name_descriptions
```
JoeCohen, line 48: @nwilson-eol , @pellaea:
### Should I add a License `created_at` column?
Seems weird that there isn't one, but that (obviously) doesn't break anything.

(PS: 3 of the Licenses have a updated_at == nil. That also seems weird. 
2 will be changed (shortly after this PR is deployed) when they are deprecated. The third (publicdomain) won't be changed.)

app/views/controllers/licenses/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<%
+  @container = :wide
+  add_page_title(:index_license_header.t)
+  add_tab_set(license_index_tabs)
+
+  flash_error(@error) if @error && @objects.empty?
+%>
+
+<%= paginate_block(@pages) do %>
```
mo-nathan, line 9: Not a blocker, but I'd rather see this as a table than as what appears to be essentially a CSV.

app/views/controllers/shared/_form_ccby40.erb
mo-nathan: It seems odd to me that there are partials for each license if we are expecting admins to add licenses   Perhaps the license object can be parameterized with the CC.org links so there can be a single generic partial.

app/models/license.rb
mo-nathan: It does seem odd that there isn't a created_at.  I note that true of the nearby "interests" table as well.  Not sure what the thinking there was other than maybe trying to save some space?  At this point I'd rather it was all consistent with both a created_at and an updated_at.

mo-nathan:
The "form_name" field seems a bit odd from a naming perspective.  Putting "Form" in the show form doesn't make sense to me.  I'm guessing the "form_name" has to do with how it gets used either to generate URLs or maybe for the form submission?  Didn't dig in deeper, but it seems worth at least renaming to "short_name" or something along those lines.

app/views/controllers/licenses/index.html.erb
mo-nathan: I'd also include the "display_name" in the table since that's more readable than the "form_name".

mo-nathan:
Overall this seems like really good progress on getting this more easily manageable for admins.  Thanks for your work on it!

JoeCohen:
@nathan: Thanks for looking at this and for all your suggestions.

app/models/license.rb
JoeCohen: Thanks! I will:
- Add a `created_at` column as a migration in this PR.
- Insure that `created_at` and `updated_at` are present for all Licenses by -- after this PR is merged and Licenses are updated via the new UI -- making any needed fixes on the back end.

app/views/controllers/shared/_form_ccby40.erb
JoeCohen: Thanks for the suggestion. I'll investigate way(s) of having a generic partial.

JoeCohen: @mo-nathan This turns out to be very easy; CC has changed urls so that we don't even have to parametize the license object. In fact, I'm going to ditch the partial and put the code in the show template.

PR #2175, opened 2024-06-10
Fix `project_messages` alert/checkbox HTML in obs form

nimmolo:
Sorry - I should have noticed this on the original PR. This doesn't change much, it just moves some things around (invisibly).

- Eliminates duplicate DOM id `observation_projects`, which is invalid HTML
- Moves `flash_warning` from the view to the controller, so flash messages are available to all rendering methods, in the case the obs form gets Turbo-ized. (Generally, flash needs to happen in the controller if we ever want to flash inside a "modal" form or use something like `ViewComponent`, but I think it's also just better practice. I'm aiming to gradually move all `flash` calls to the respective controllers when possible.)
- On the form, moves the whole message and checkbox into the `#observation_projects` div, and move the checkbox inside the `.alert` div, as it is for `form-name-feedback` et. al.

### DIFF
--- a/app/controllers/observations_controller/validators.rb
+++ b/app/controllers/observations_controller/validators.rb
@@ -65,6 +65,9 @@ module ObservationsController::Validators
 
     @suspect_checked_projects = checked_project_conflicts -
                                 @observation.projects
+    if @suspect_checked_projects.any?
+      flash_warning(:form_observations_there_is_a_problem_with_projects.t)
+    end
     @suspect_checked_projects.empty?
   end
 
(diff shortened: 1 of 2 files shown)


JoeCohen (approved):
Thanks for the fixes & improvements.

nimmolo:
Thanks @JoeCohen - will merge then.

nimmolo:
Oops - I'm on the train and cannot SSH to deploy on production. If @JoeCohen or @mo-nathan can easily deploy, please do; otherwise i'll do it later tonight.

Just wanted to give a heads up, because we've had bad experiences with my merged but undeployed code before!

JoeCohen:
I'll deploy it now.

JoeCohen:
Deployed.

nimmolo:
Thanks @JoeCohen

PR #2176, opened 2024-06-11
Update display of license dates

JoeCohen:
- Fixes bug in created_at display
- Gets fixtures a bit more in sync with db
  - adds created_at to fixtures
  - removes nil updated at
- Updates show template
- Simplifies helper
- Deletes now uneeded test


JoeCohen:
This PR includes a bug fix, so I will merge and deploy it now.

PR #2183, opened 2024-06-18
Observation form phase 2a: matching MO Locations

nimmolo:
This "phase 2a" PR adds some JS and a query scope that tries to guess a Location as soon as we get a lat/lng. 

It adds a JS listener so that _changes to the lat/lng fields, either from user input or image autofill, call up a new autocompleter class_, `ForLocationContaining`. This class queries the db to return all MO locations _encompassing_ the lat/lng given, in order of ascending area. The JS then populates the autocomplete options of the Location input with these locations, and auto-fills the `place_name` and hidden `location_id` fields with the first (smallest fitting) location.

"phase 2b" would move the UI for creating a custom location to a pop-up modal, available from the "create_obs" form. It would require any undefined `place_name` to be created as a db Location _prior_ to saving the obs, rather than offering this optionally, after creating the obs, as is currently the case. (It would still not require a Location however — it would only block creating an obs with an undefined place name.)

"phase 2c" is intended to supplement the suggestions in 2a with a call to Google's reverse geolocation service, in the event that MO does not have any matching locations.


nimmolo:
- [ ] Actually fill the fields `place_name` and `location_id` if the autocompleter gets a result. Currently just fills out options
- [ ] Handle swapping autocompleter back to `location` if lat/lngs are removed
- [ ] Needs tests and maybe a Location fixture that encompasses our geotagged image (or... a separate geotagged2.jpg)

nimmolo:
This is now deployed on the test server, in case you want to check it there.

nimmolo:
Issues discovered by @pellaea 
- [ ] Map point does not move when new image EXIF selected
- [ ] Mystery options appear in autocompleter when moving point manually on map
- [ ] Keyboard up/down within autocompleted `location_containing` does not highlight current row
- [ ] Locations other than the smallest location_containing are not selectable, or selection reverts

nimmolo:
This passes tests for me locally now. Idk what's up with CI.

PR #2184, opened 2024-06-18
Refactor autocompleters to use query params

nimmolo:
This is groundwork for the location suggestions autocompleter, and any further autocompleter work. Want to be sure this works and is merged separately.


nimmolo:
@JoeCohen I think this works and causes no changes, but you've been noting typeahead probs on the obs form branch. Wanted to ask you for a quick test (not a code review).

PR #2193, opened 2024-06-23
Fix `nil?`checks for associated records: check the ID only

nimmolo:
Housekeeping PR that came up when examining very old code involved with the feature I'm working on, auto-suggesting locations. 

Even though it's convenient and looks more elegant, it's inefficient to nil-check an associated record directly, unless it's been eager loaded — apparently, it will instantiate the record. 

Bad habit:
```ruby
@observation.location.nil?
```
Good habit:
```ruby
@observation.location_id.nil?
```

### DIFF
--- a/app/controllers/herbarium_records_controller.rb
+++ b/app/controllers/herbarium_records_controller.rb
@@ -358,7 +358,7 @@ class HerbariumRecordsController < ApplicationController
     if name.blank?
       flash_error(:create_herbarium_record_missing_herbarium_name.t)
       false
-    elsif !@herbarium_record.herbarium.nil?
+    elsif !@herbarium_record.herbarium_id.nil?
       true
     elsif name != @user.personal_herbarium_name || @user.personal_herbarium
       flash_warning(:create_herbarium_separately.t)
--- a/app/controllers/observations_controller/edit_and_update.rb
+++ b/app/controllers/observations_controller/edit_and_update.rb
@@ -194,7 +194,7 @@ module ObservationsController::EditAndUpdate
   end
 
   def redirect_to_observation_or_create_location
-    if @observation.location.nil?
+    if @observation.location_id.nil?
       redirect_with_query(new_location_path(where: @observation.place_name,
                                             set_observation: @observation.id))
     else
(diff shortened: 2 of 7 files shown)


nimmolo:
Also - heads up, very unexpected behavior with fixtures: 
Giving a fixture a `nil` association apparently creates a bogus fixture ID for the `nil` record — but no fixture. 

Example:
```yaml
# includes observations that violate date constraints
# members: roy
nowhere_2023_09_project:
  <<: *DEFAULTS
  user: dick
  admin_group: dick_only
  user_group: falmouth_2023_09_users
  start_date: 2023-09-01
  end_date: 2023-09-30
  location: nil
  observations: brett_woods_2023_09_obs, falmouth_2023_09_obs, falmouth_2022_obs, nybg_2023_09_obs # roy (non-admin) owns nybg_2023_09_obs
``` 
```ruby
project = projects(:nowhere_2023_09_project)
debugger
```
```
(rdbg) project.location
nil
(rdbg) project.location_id
221495974
(rdbg) Location.find(221495974)
eval error: Couldn't find Location with 'id'=221495974
  /Users/nimmo/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/activerecord-7.1.3.4/lib/active_record/core.rb:253:in `find'
  (rdbg)//Users/nimmo/Documents/GitHub/developer-startup/mushroom-observer/app/helpers/projects_helper.rb:1:in `styled_obs_lat'
nil
```
In this case, the default `project` has no associated `location`, so it's not necessary to set `location: nil`, i just commented it out.

mo-nathan (approved):
Looks good.

JoeCohen:
@nimmolo: Thanks for your work on this, and for alerting us to the inefficiencies and the fixture issue.
Some questions/comments:
- Should these be reported to the Rails team:
  - The unexpected fixture behavior seems like a bug. It's a least a trap for the unwary that should be documented.
  - Instantiation of the record seems like at least an opportunity to speed up Rails, if not a bug. 
- Should we update other fixture sets to common out unnecessary `nil` associations? I sometimes add an explicit `nil` to indicate that I don't want the association changed, or to override a default fixture value. If I'm using explicit fixture `nil`s, it seems like I must nil-check the associated record directly (because `<association>.id.nil?` will get the wrong result in tests).
- Seems like sometimes I must do what you describe as a bad habit. For me it's complicated and confusing to figure out when I should nil-check the id vs. when I should (must) nil-check the association. As a result, I'll probably always nil-check the association directly; I'd rather be slower than buggy.

PR #2195, opened 2024-06-24
Obs form: accept autocompleted `:location_id` and `:name_id`

nimmolo:
If we have a match (with an ID) for location or name, this changes the controller to deal with the selected record id, bypassing some of the code that otherwise tries to match a submitted string to a record. 

For names, though, it does not bypass the code that checks if the name is deprecated or ambiguous.


nimmolo:
Doing these separately.

PR #2212, opened 2024-07-05
Location.contains Scope 

JoeCohen:
Adds a new Location scope.
`Location.contains_box(n: a, s: b, e: c, w: d)` returns all Locations which bound the defined rectangle.

This is a small, but necessary step, in dealing with iNat places as part of importing iNat observations.

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -243,8 +243,8 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
             # Location straddles 180
             #   Location 100% wrap; necessarily straddles w/e
             and(Location[:west] == Location[:east] - 360).
-            #   Location < 100% wrap-around
-            or((Location[:west].gt(Location[:east])). # rubocop:disable Style/RedundantParentheses
+            #  Location < 100% wrap-around
+            or(Location[:west].gt(Location[:east]).
               and(Location[:west] <= shrunk_w).
               and(Location[:east] >= shrunk_e)))
           end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -212,56 +212,20 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
             args => {n:, s:, e:, w:}
           end
 
-          shrunk_n = n - ROUND_ERROR
-          shrunk_s = s + ROUND_ERROR
-          shrunk_e = e - ROUND_ERROR
-          shrunk_w = w + ROUND_ERROR
+          # Correct for possible floating point rounding
+          # Shift points by 180 degrees to simplify east/west comparisons
+          # (Avoids issues with locations straddling 180)
+          shrunk_n = n - ROUND_ERROR + 180
+          shrunk_s = s + ROUND_ERROR + 180
+          shrunk_e = e - ROUND_ERROR + 180
+          shrunk_w = w + ROUND_ERROR + 180
 
-          # Location straddles n & s
           where(
             Location[:south].lteq(shrunk_s).
-              and(Location[:north].gteq(shrunk_n)).
-
-            # and Location east/west straddle e & w
-            and(
-              # Location east/west do not straddle 180 degrees
-              Location[:west].lteq(Location[:east]).
-              and(Location[:west].lteq(shrunk_w)).
-                # TODO: Fix or understand the next line. It seems backwards.
-                and(Location[:east].lteq(shrunk_e))
-=begin
-                .
-              or(
-                # Location east/west straddle 180 degrees
-                Location[:west].gt(Location[:east]).
-                  and(Location[:west].gteq(shrunk_w)).
-                  and(Location[:east].lteq(shrunk_e))
-              )
-=end
-            )
+            and(Location[:north].lteq(shrunk_n)).
+            and((Location[:west] + 180) <= shrunk_w).
+            and((Location[:east] + 180) >= shrunk_e)
           )
-            # and(Location.where(Location[:west].lteq(w + ROUND_ERROR))).
-            # and(Location.where(Location[:east].gteq(e - ROUND_ERROR)))
-
-
-=begin
-            # Location straddles e/w
-            and(
-              # containing location does not straddle 180
-              Location[:west].lteq(Location[:east]).
-                and(Location[:west].lteq(w + ROUND_ERROR)).
-                and(Location[:east].gteq(e - ROUND_ERROR))
-              # containing location west > east
-              # or(Location[:west].gteq(w - ROUND_ERROR).
-              #  and(Location[:east].lteq(e + ROUND_ERROR)))
-
-=begin
-              # containing location straddles 180 deg
-              or(Location[:west].gt(Location[:east]).
-                and(
-                  Location[:west].lteq(w + ROUND_ERROR).or(Location[:east].gteq(e - ROUND_ERROR))
-                ))
-=end
         }
 
   # Let attached observations update their cache if these fields changed.
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -150,7 +150,7 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     end
   end
 
-  ROUND_ERROR = 0.00001
+  ROUND_ERROR = 0.0005
 
   # NOTE: To improve Coveralls display, do not use one-line stabby lambda scopes
   scope :name_includes,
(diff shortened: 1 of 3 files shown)


app/models/location.rb
```diff
@@ -199,6 +201,57 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
...
+                  and(Location[:north].gteq(shrunk_n)).
+            # Location straddles 180
+            #   Location 100% wrap; necessarily straddles w/e
+            and(Location[:west] == Location[:east] - 360).
+            #   Location < 100% wrap-around
+            or((Location[:west].gt(Location[:east])). # rubocop:disable Style/RedundantParentheses
+              and(Location[:west] <= shrunk_w).
+              and(Location[:east] >= shrunk_e)))
+          end
+        }
```
JoeCohen, line 253: @nimmolo: This is complex because the conditions that work for Location.contains(point)
```ruby
              Location[:west].lteq(lng).and(Location[:east].gteq(lng)).or(
                Location[:west].gteq(lng).and(Location[:east].lteq(lng))
```
do not work in 2 dimensions. For instance, Alaska would not contains most places inside Alaska.

@nimmolo & @mo-nathan: I would appreciate anything that would simplify this mess, or even just make it easier to understand or read.

test/fixtures/locations.yml
```diff
@@ -258,7 +258,10 @@ perkatkun:
   north: 71.154802
   west: -179.809370
   south: 71.104900
-  east: -179.613214
+  # NOTE: jdc 2024-07-04
+  # Next line causes failures. Due to AR/Ruby truncation/rounding??
+  # east: -179.613214
+  east: -179.613
```
JoeCohen, line 264: @mo-nathan: Can you shed any light on this? 
When I had `east: -179.613214`:
1. If I put a breakpoint anywhere in the tests and do an AR query for this fixture's `east`, then the result is rounded to 6 digits plus the minus sign. Ex:
```ruby 
=> 631|     debugger
   632|     # loc doesn't straddle 180
   633|     #   potential br (bounding rectangle) entirely to "left" of loc
   634|     do_contains_box(loc: albion, external_loc: perkatkun,
   635|                     regions: [california, earth])
=>#0    LocationTest#test_scope_contains_box at ~/mushroom-observer/test/models/location_test.rb:631
  #1    block in run (3 levels) at ~/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/minitest-5.23.1/lib/minitest/test.rb:95
  # and 25 frames (use `bt' command for all frames)
(ruby) perkatkun.east
-179.613
```
I don't understand why, or what implications that might have outside of this PR.

2. Tests fail, and I don't know how to fix them. For instance, increasing `ROUND_ERROR` didn't work.
My change to the fixture makes the tests pass, but it dodges the problem rather than solving it.

JoeCohen: What should I do if `Location.contains` called with missing/bad arguments:
1. Currently I think it throws an error? This alerts us to the problem.
2. Return Location.none
3. Other?

app/models/location.rb
nimmolo: Looks pretty good, not sure how it could be improvable. Ideas:
- Maybe let's split the one scope into two: a `contains_point` and a `contains_box`. That will help legibility at least and shorten the scope.

test/fixtures/locations.yml
nimmolo: I vote error.

mo-nathan:
@JoeCohen These values are getting rounded in the database.
```
mysql> update locations set east = -179.61345 where id = 732613413;
Query OK, 1 row affected (0.00 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> select east from locations where id = 732613413;
+----------+
| east     |
+----------+
| -179.613 |
+----------+
1 row in set (0.00 sec)

```

Is there reason to believe this is new behavior?

JoeCohen:
@mo-nathan Thanks for the rounding information! That's helpful.
I have no reason to believe the db behavior is new. 
I'm probably doing something that -- for the first time -- depends on more precision. 
I'll see if I can fix that.

nimmolo:
Note that there's rounding happening in the JS too, which eventually populates the DB. 

Consider making an app-wide Rails constant for the number of decimal places we round to. Once we establish that, I can send that to the JS for consistency.

JoeCohen:
@nimmolo: app-wide rounding constant seems like a good idea. 
I think it should be in a different PR because (a) I don't want to stack too much stuff into this one; (b) I don't know how to do it or where to put it. Probably it should be derived from the db. I'm guessing (based on experimentation) that it's 4 digits to the right of decimal point for positive numbers, 3 for negative numbers. I also don't understand whether the db rounds it or just truncates it.

nimmolo:
If the db were doing the rounding, i'm pretty sure it would be in our schema. It's not the db doing this though - it's our Location-saving methods that round the float. 

I'm not aware of any code that rounds negative numbers differently - that's worth checking on. I'm also not aware of any truncation!

But there is more than one place in the code that rounds lat/lngs, so let's not forget to fix this error-prone situation. (I guess we could also change the schema through a migration - maybe the db is the place to do this.)

If we do a constant that is something I would put in a config with our other MO constants. 

Let's get advice from
@mo-nathan - he my have a preference.

nimmolo:
Per your communication on Slack, I went ahead and changed to `contains_point` on my PR, which I hope to merge today.

mo-nathan:
@nimmolo What evidence do you have that the fixtures are getting affected by any of our code?  I did look at this and could not find anywhere that was actually being called during a test run that rounding the values when running the test that uses the location fixture that Joe mentioned.  I then found articles on the web about MySQL rounding FLOAT type values and confirmed that with the test I gave above (that uses no Ruby at all).

nimmolo:
Oh - it sounds like i misperceived the situation. I have no evidence of fixtures being affected.

My concern is more structural, and I wanted to raise a flag before I forgot about it again. I've just been working on location code for a few months and realizing that:
- we round in several separate places in the code
- it's currently up to each method to round lat/lngs however it likes
- until recently they were rounded inconsistently

I recently changed all the float rounds to 4 places (some of mine from a couple years ago had been 6 places), and it occurred to me to use a constant, but i didn't go that far.

mo-nathan:
I agree that being consistent about the rounding would make sense.  I did some more playing around with the lat/lng and on the production server we round to 6 decimal digits.  This mean that somewhere around Kansas the longitude resolution switches from about 300' to about 30'.  This can cause weirdnesses like this: https://mushroomobserver.org/locations/29140 (this is the house where I grew up).  This seems wrong, but I'm not sure exactly what "right" is.  That observation was created by searching Google for that address and then saving.  While I was defining it, it looked like this:
Screenshot 2024-07-06 at 3 47 03 PM

It would at least be better if it saved as:
Screenshot 2024-07-06 at 3 50 37 PM

rather than
Screenshot 2024-07-06 at 3 51 25 PM

mo-nathan:
It seems like it would be be reasonable to support 7 digits (rather than 6) and then round off to the nearest 1/10,000s of degree (~30ft).  This would not lose any existing data, but would start to treat the west coast the same as the east coast.

mo-nathan:
Note that my suggested solution involves a data migration to store higher precision floats in general.  We might consider switching to the DECIMAL type.  Not sure whether Rails likes that, but it seems like the right number format based on the MySQL documentation.

JoeCohen:
Re rounding:
1. I was wrong about negative numbers. It's a total of 6 digits. As @mo-nathan points out, this leads to less precision in (-100..100). 
2. I agree with @mo-nathan that the ideal solution (for Location) is decimal north, south, east, west. (Maybe also altitude.)
Otherwise there are surprises. Simple example:
```ruby
(ruby) Location.where(Location[:east].eq(albion.east)).include?(albion)
false
```
@nimmo: Above implies that until/unless we switch to decimal, `scope_contains_point` should have have something to compensate for db rounding, and should have some model unit tests. I'll have a look at your PR later today.

JoeCohen:
PS: @mo-nathan: Rails is okay with :decimal column type. https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_column
Question: Would we need to recast the existing db entries for affected columns?

mo-nathan:
Probably the best option would be to have a migration that creates a set of
new decimal columns, copies the date to the new columns, drops the old
columns, then renames the new columns to the old names.  Obviously we’d
have to test it, but at least in theory that should do it.

nimmolo:
Hold on... let's not forget about Geometry columns! 

I feel like that's the migration we want here. Almost totally forgot about that!

nimmolo:
Potentially interesting: [Rails - How to migrate code from float to decimal](https://stackoverflow.com/questions/13255987/ruby-on-rails-how-to-migrate-code-from-float-to-decimal)

PR #2214, opened 2024-07-05
Prevent download of all Observations

JoeCohen:
Redirects to Observation#index with flash error if User tries to download all Observations.

- Downloading all Observations is so expensive that it causes a Downtime Alert. See [Example](https://mushroomobserver.slack.com/archives/C06FHR54EPQ/p1719681356634849). I did it accidentally, but more often it's deliberate.
- Users can get the same data using fewer resources, via the CSV files.

### Simple Manual Test
- Goto the home page
- Click `Download Observations`
Expected result: Redirects to home page with a flash error.
- Click the link in the flash.
Expected result: Opens a new window scrolled to the`CSV files` section of [README_API.md](https://github.com/MushroomObserver/mushroom-observer/blob/main/README_API.md)

### DIFF
--- a/app/controllers/observations/downloads_controller.rb
+++ b/app/controllers/observations/downloads_controller.rb
@@ -6,6 +6,8 @@ module Observations
 
     def new
       @query = find_or_create_query(:Observation, by: params[:by])
+      return too_many_results if too_many_results?
+
       query_params_set(@query)
     end
 
@@ -30,6 +32,15 @@ module Observations
     end
 
     private
+
+    def too_many_results
+      flash_error(:download_observations_too_many_results.t)
+      redirect_to(observations_path)
+    end
+
+    def too_many_results?
+      @query.flavor == :all
+    end
 
     def download_observations_switch
       if params[:commit] == :CANCEL.l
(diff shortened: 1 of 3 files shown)


app/controllers/observations/downloads_controller.rb
```diff
@@ -31,6 +33,15 @@ def print_labels
 
     private
 
+    def too_many_results
+      flash_error(:download_observations_too_many_results.t)
+      redirect_to(observations_path)
+    end
+
+    def too_many_results?
+      @query.flavor == :all
```
JoeCohen, line 42: Should I expand this? Examples:
- A fixed upper limit
- Max of a fixed upper limit || all the current user's Observations.

config/locales/en.txt
```diff
@@ -1479,6 +1479,7 @@
   download_observations_utf16: UTF-16 (best for Windows 7 and up)
   download_observations_print_labels_header: Or you can print labels for the selected observations
   download_observations_print_labels: Print Labels
+  download_observations_too_many_results: You are trying to download too many Observations. Please see <a href="https://github.com/MushroomObserver/mushroom-observer/blob/main/README_API.md#csv-files" target="_new">CSV Files</a> for an alternative.
```
JoeCohen, line 1482: Any comments about this error message?
- I don't want to tell users to email webmaster, because I don't want to spend time dealing with their messages.
- The link in the message goes to the file which describes the API. There's some danger in encouraging random visitors to use the API. So maybe we just say "You are trying to download too many Observations."

PR #2230, opened 2024-07-08
Create/Edit Obs: new create location flow

nimmolo:
The create or edit observation form should now either
- find an MO Location from a lat/lng (either manually entered, or from an image exif data), but if it does not find one that fits... 
- populate the autocompleter with options from Google's matching locations, in ascending area.
- offer to "Create location" with Google's bounding box data autofilled in hidden fields, placing an interactive box in the map that the user can alter, thereby updating the hidden fields.
- offer to lock that box down ("Use this location") so the user can click a point within it to set the lat/lng, if they haven't done so already.

On save, 
- the Location should be created before the Observation, and the Observation attached to the Location. 
- Problems with the Location should throw flash errors in the Obs form and prevent validation.
- Trying to create a Location with the name of an existing MO Location should find and attach the existing Location to the obs, and ignore whatever the user was trying to (re)define the Location as.


JoeCohen:
if this is possible **and easy* modify new location names as follows.
(If not eay (but possible) do it in the future.)
Screenshot 2024-07-18 at 6 17 10 PM
With the above input, it correctly found the boundaries of Costilla County, Colorado. (The photo did not have location info.)
But it doesn't get the location name perfect.
Since it figured out the bounding box, it seems like there should be enough information to get the name MO style. (Costilla Co., Colorado, USA.)

JoeCohen:
I also don't understand why the CI tests aren't running.
Screenshot 2024-07-18 at 6 25 11 PM

nimmolo:
Thanks @JoeCohen - i didn't even check that, since it's using the same foramatter as the lat/lng matcher, which does format locations correctly for MO. 

It's very unexpected that the country component is missing - it's assembling the string from separate address components that are assumed to be in the object returned. Will check.

JoeCohen:
Another "nice to have if easy" and seeming bug.
Update the Locality window after this sequence: enter something in that window, Create Location, bounding box found.
Example:
- Create Observation 
- Fill in Locality with "Harrietsburg, New York, USA"
- Click Create Location
Result: It finds a box for the Harrietstown (there's no such place as Harrietsburg), but does not update the Locality window. 
Screenshot 2024-07-19 at 4 54 02 AM

It would be nice to automatically update that window.
- Click in the Locality window
Result: double municipality: "Harrietstown, Harrietstown, Franklin Co., New York, USA"
Maybe one is due to the township, the other to the cemetery?

JoeCohen:
Bug: The changes to Observation break the /location/new `Find on Map` button.
Ex: 
- On this branch
  - Click Create Location (or goto http://localhost:3000/locations/new)
  - Enter "Slater, Iowa, USA" in the Locality field 
  - Click Find on Map
Result: Radio silence: No bounding box, default geolocation fields unchanged.
Screenshot 2024-07-19 at 5 22 15 AM

- Same thing on the webserver. 
Result: It finds and displays a bounding box, fills in geolocation.
Screenshot 2024-07-19 at 5 25 49 AM

nimmolo:
@JoeCohen Thank you for all these! Very helpful. I'm checking what I think is fixed here, feel free to un-check and this will serve as the TODO list:

- [x] Location form "Find on map" working
- [x] Updates "Locality" name field on Google place match
- [x] County abbreviated

PR #2236, opened 2024-07-14
Fix date behavior in the Project edit form

mo-nathan:
To reproduce:

1) Edit a project
2) Select a date range, click the "Range" radio button, and save.
3) Re-edit the project
Expected behavior: New date range is shown.
Current behavior: Date is set to today

There were two issues.  1) The function `date_select_opts` only looked for the `when` field in the object.  2) The same function always expected to get the object from an explicitly provided `object:` option.  The fix uses the `field:` option and tried to look up the object from any associated form.  If it doesn't find one, then it uses the `object:` option.


JoeCohen (approved):
I tested it manually and (of course) it works.
Nothing in the code jumped out at me as wrong (but I don't understand the helper methods and don't have time to dig into them today).

PR #2256, opened 2024-07-26
Only accept known type names to avoid code injection warnings

mo-nathan:


test/models/comment_test.rb
```diff
@@ -11,6 +11,13 @@ def teardown
     QueuedEmail.queue = false
   end
 
+  def test_find_object_for_all_types
+    Comment.all_types.each do |type|
+      assert(AbstractModel.find_object(type.to_s, type.first.id),
+             "Unable to use find_object to find #{type}")
+    end
+  end
```
JoeCohen, line 19: @mo-nathan: thanks for doing this. It Looks good.
Would it make sense to also have an assertion that goes the other way. Sort of like
```suggestion
    assert_nil(AbstractModel.find_object(Account, 1))
  end
```

JoeCohen (approved):

test/models/comment_test.rb
mo-nathan: It wouldn't hurt, but I don't know that it's necessary.  I'm merging while the build is green since there seems to be some instability around some of the recent JS/system tests.

PR #2258, opened 2024-07-26
Observation form: steps

nimmolo:
Simple stepped form.

Screen Shot 2024-07-26 at 7 27 00 PM
Screen Shot 2024-07-26 at 9 43 25 AM
Screen Shot 2024-07-26 at 9 44 56 AM


nimmolo:
@JoeCohen This is deployed on test server now

JoeCohen:
I'd like to have the ID stuff (but not the Notes in the first tab.
I think of photos, Location/lat-lng, and ID as quasi mandatory
Notes and Additions as completely optional.
So First tab: what you have now plus ScientificName and Confidence
Second tab: Notes
Third tab: as-is

nimmolo:
Thanks. It's more of a space consideration, putting naming and specimen stuff in the second step. For me the form page is simply too long, and I want to break it up and narrow the user's focus to one thing at a time. 

I'm signalling the second step is mandatory by not providing a commit button on the first step. 

I would have actually preferred to make it a four-step form, splitting images and location, but they're too tightly coupled!

Think of the first two steps as mandatory!

JoeCohen:
Locally, Location type-ahead seems broken.
I dragged an Image, it gave me a Location I didn't like. I wanted to change the Location. But it's ignoring my input to the Location type-ahead window. 
Screenshot 2024-07-26 at 1 37 16 PM

JoeCohen:
Is there a way to make the map stuff a modal? That way you could put the ID
stuff on the right side of the first tab.

JoeCohen:
How hard would it be to avoid losing coverage? See the lines marked "New" here:
https://coveralls.io/builds/68885088/source?filename=app%2Fhelpers%2Fcontent_helper.rb#L217

JoeCohen:
Locally, there's no Next button on the first tab

JoeCohen:
Test server hangs (I waited for 5 minutes) when I click the Create button.

nimmolo:
Do you mean there's no `Create` button in the first tab? There should be a `Next` and no `Create`, that is intentional. 

Nothing has changed on the Location typeahead, so if there are bugs they'll probably be in `main`.

nimmolo:
Modal was the original idea, but is a no-go for the map because of the impossibility of passing the Google info back and forth between the autocompleter, Turbo and the modal. I had to rule that out a couple weeks ago, I think i mentioned on Slack.

JoeCohen:
> Do you mean there's no `Create` button in the first tab? There should be a `Next` and no `Create`, that is intentional.
Hmm. Now I am getting a Next button. (That's what I meant.) You should ignore this. 
> Nothing has changed on the Location typeahead, so if there are bugs they'll probably be in `main`.
Ditto. Now the type-ahead works.

nimmolo:
> How hard would it be to avoid losing coverage? See the lines marked "New" here: https://coveralls.io/builds/68885088/source?filename=app%2Fhelpers%2Fcontent_helper.rb#L217

These are just tab-navigation helper methods i built while i was thinking about them, and able to remember how to code them correctly. They were used on an early version of the tabbed form, and ditched, but I believe they will be useful in the future for info pages. No sense deleting them, and no sense testing them yet, as far as I can tell.

nimmolo:
Ugh. I've had it with these system tests. 

What fails on CI passes locally. Blech!

JoeCohen:
> Ugh. I've had it with these system tests.
> 
> What fails on CI passes locally. Blech!

FWIW, I essentially ignore them. I don´t run them locally. If they fail in CI, I ignore that.

nimmolo:
Interesting. Thanks for telling me. 

Unfortunately, they're the only thing we have (other than us — but frequently no one is available) for testing the JS UI, so when I'm working on the UI I do have to get them passing. And it's true that they catch a lot of bugs. But so time consuming.

nimmolo:
@JoeCohen i'm just hiding the resolved issues you brought up, so as not to confuse @mo-nathan in case he has time to review.

PR #2297, opened 2024-08-08
Observe todo file's Exclude's

JoeCohen:
RuboCop was ignoring some of the todo file's Exclude's. This PR fixes most of that
- Applies the workaround suggested [here](https://github.com/rubocop/rubocop/issues/13091#issuecomment-2275752003) and [here](https://github.com/rubocop/rubocop/issues/9325#issue-777465916)
- regenerates RuboCop todo file
- removes now-unnecessary `disable`'s in `app/helpers/forms_helper.rb`
- inlines the remaing `disable`; using the inline gives us one fewer place to change in case we want to remove this temporararily or permanently.


JoeCohen:
@nimmolo: This mostly solves the problem we discussed. 
But I still needed to explicitly disable  `Metrics/ModuleLength` in `forms_helper.rb`.

nimmolo (approved):

nimmolo:
Thanks for doing this, Joe

PR #2300, opened 2024-08-09
Location: Improve Riparius handling

nimmolo:
This should make the autocompleter less eager to override a manually entered place name when there is also a lat/lng from a photo. 

It will still however retrieve Google's version of the place name formatted for MO, so for example "Riparius, New York" becomes "Riparius, Chester, Warren Co., New York, USA". 

It also hopefully fixes a JS error where the bounding box hidden fields stopped being updated by the user manually resizing the box.

### DIFF
--- a/app/javascript/controllers/autocompleter_controller.js
+++ b/app/javascript/controllers/autocompleter_controller.js
@@ -215,6 +215,13 @@ export default class extends Controller {
     }
     if (type == undefined) { return; }
 
+    let location = false;
+    if (detail.hasOwnProperty("request_params") &&
+      detail.request_params.hasOwnProperty("lat") &&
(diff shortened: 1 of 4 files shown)


nimmolo:
@mo-nathan I thiiiink this smooths out most of the weirdness. If you have time, and are able to return your local db to a Riparius-agnostic state, check it out and let me know how it goes. 

It shouldn't keep overriding the place with "Chester" when you delete the place name and try to type "Riparius..". It should just offer you a "Create locality" button, and when you click that, it should fill it with Google's version of the place. You can then edit the borders. 

There may be other issues but i think this is at least an improvement. There are so many interdependent factors and ux triggers.

PR #2304, opened 2024-08-12
Fix Run BLAST links

### LINKED ISSUE #2303: Fix BLAST links
They appear to be broken.
Sample Obs: https://mushroomobserver.org/205345

JoeCohen:
- Changes the HTML method for BLAST queries.
- Uses `Post` instead of `Put`. `Put` no longer works; NCBI botches the `QUERY` param when the method is `Put`.
- Delivers #2303


JoeCohen:
Deployed. deploy-2024-08-12-15-40

PR #2315, opened 2024-08-18
More legible tooltips for all icons

nimmolo:
Because I didn't "rtfm", I've never implemented tooltips correctly on MO. We've apparently been getting by on browser-default HTML `title` attributes the whole time.

This PR puts a tooltip activator in Stimulus (where it needs to be now) and should make the icon "tooltips" more legible. The main benefit is that the text appears much more quickly on hover than browser "title" attributes did, so the tips are more likely to be helpful to users.

Before:
Screen Shot 2024-08-18 at 2 57 03 PM

After:
Screen Shot 2024-08-18 at 2 57 24 PM

The one that doesn't display nicely is "Turn Admin Mode On", but I trust that admins know what that button does at this point. The tooltip JS library is different in BS 4, so there doesn't seem to be much point spending an hour troubleshooting this private use case.

Screen Shot 2024-08-18 at 3 03 11 PM

### DIFF
--- a/app/helpers/link_helper.rb
+++ b/app/helpers/link_helper.rb
@@ -81,16 +81,17 @@ module LinkHelper
   # Icon link with optional active state. (Tooltip title must be swapped in JS.)
   # Now also accepts active state options: active_icon, active_content
   # NOTE: Takes same args as link_to, e.g. *edit_description_tab(desc, type)
-  # icon_link_to(text, path, **args)
+  # icon_link_to(text, path, **args). Can also print a button_to.
   def icon_link_to(text = nil, path = nil, options = {}, &block)
     return unless text
 
-    link = block ? text : path # because positional
+    link_path = block ? text : path # because positional
     content = block ? capture(&block) : text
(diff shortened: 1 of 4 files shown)


nimmolo:
@JoeCohen @mo-nathan 
Not necessary to review this, but I know we have talked about better icon legibility in the past. I believe this is an improvement.

JoeCohen:
Yes, judging by the screenshots you attached.
Thanks for doing this.

nimmolo:
PR also standardizes the `icon_link_to` helper so that it can print an `button_to` for non-`GET` links. (Rails supposedly can build a `link_to` a `PUT` route, but in our app at least, it requires a `button_to` to accomplish this. The resulting HTML in both cases needs to be a hidden `form` with a `button` element, not an `a` link.)

JoeCohen:
I'll never remember that. 😢

nimmolo:
Don't worry, it'll ring a bell if your link doesn't work. 
Also - that's why i write comments! :)

PR #2337, opened 2024-08-26
Pattern search controllers

nimmolo:
- This PR creates a new “pattern search” form for each model (e.g. `Comment`, `Name`). It lives at a new endpoint, `names/search/new`, with a new controller, e.g. `Names::PatternSearchController`. The forms are different per model because the keywords available in `Query` for each model are different. (Mostly just for name and observation.)
- Even though these are separate new form endpoints, we can eventually use a single modal form for pattern searches. When the user selects the model they’re searching, Stimulus calls the model’s controller endpoint to get the search form. Turbo then populates the search modal with the form. (Notice here how Turbo facilitates keeping responses in Ruby (the regular controller response) while making them appear on the page via Javascript — i.e. without separate page reloads.)
- In each form, there’s a field for each search keyword (param). The form commits to the same controller’s create action and sends the params as a regular param hash.

Conceptually it may seem straightforward to initiate a query from the form's keyword params, but they’re not ready yet: the `PatternSearch` class does validation and translating for both the values and the param keys. It translates human input values into values that are executable in AR/SQL, and translates our pattern search keywords and the actual `Query` class param names. 

So I considered two ways of handling these search params in the create action:

1. Translate the HTML param string into the type of MO pattern search string that an expert user would enter, if they knew how.
  - Reassemble the params into a query string, the way they came in: `?pattern=&date=2024-08-24&id=2345`
  - Then do a bunch of string manipulation to change this string into the type of pattern string expected by `PatternSearch.new`. Change the `&` into spaces, escape the commas in addresses, re-encode the whole string as a single pattern param, and send it to the index action at `names`, where if everything’s ok, we show the results.
2. Assemble the params directly into a sanitized, corrected `query_params` hash and initiate a query from the `create` action, then send the query to the names `index` action. Doing this without duplicating functionality would involve a new type of object in the `PatternSearch` class that could validate the native query params and instantiate a query from that hash of params, rather than parsing the string (which is what it’s set up to do currently — there’s no direct “hash outlet” to plug into).

The way i’m handling it is the first one. I actually prefer the directness of #2, and may try it in the future, but it made my head swim thinking about all the validation and error handling. #1 is a roundabout way of doing things: putting params back into a string and altering the string rather than keeping a clean param hash, but it leverages a lot of work that the `PatternSearch` class already does. The values for the params are human input, and require translation to be valid `query_params`, so even if we used the native `Query` class param keys, we’d have to translate and validate the values. 

I intend to merge this PR after testing, but before building the modal search form. That way we will be able to test each of these forms at the _un-advertised_ endpoints, and be sure they're working, before changing anything about the search UI.

### DIFF
--- a/app/controllers/comments/pattern_search_controller.rb
+++ b/app/controllers/comments/pattern_search_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Comments pattern search form.
+#
+# Route: `new_comment_pattern_search_path`
+# Only one action here. Call namespaced controller actions with a hash like
+# `{ controller: "/comments/pattern_search", action: :create }`
+module Comments
+  class PatternSearchController < ApplicationController
+    before_action :login_required
+
+    def new
+    end
+  end
+end
--- a/app/controllers/glossary_terms/pattern_search_controller.rb
+++ b/app/controllers/glossary_terms/pattern_search_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# GlossaryTerms pattern search form.
+#
+# Route: `new_glossary_term_search_path`
+# Only one action here. Call namespaced controller actions with a hash like
+# `{ controller: "/glossary_terms/pattern_search", action: :create }`
+module GlossaryTerms
+  class PatternSearchController < ApplicationController
+    before_action :login_required
+
+    def new
+    end
+  end
+end
--- a/app/controllers/herbaria/pattern_search_controller.rb
+++ b/app/controllers/herbaria/pattern_search_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Herbaria pattern search form.
+#
+# Route: `new_herbarium_pattern_search_path`
+# Only one action here. Call namespaced controller actions with a hash like
+# `{ controller: "/herbaria/pattern_search", action: :create }`
+module Herbaria
+  class PatternSearchController < ApplicationController
+    before_action :login_required
+
+    def new
+    end
+  end
+end
--- a/app/controllers/herbarium_records/pattern_search_controller.rb
+++ b/app/controllers/herbarium_records/pattern_search_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Herbarium record pattern search form.
+#
+# Route: `new_herbarium_record_pattern_search_path`
+# Only one action here. Call namespaced controller actions with a hash like
+# `{ controller: "/herbarium_records/pattern_search", action: :create }`
+module HerbariumRecords
+  class PatternSearchController < ApplicationController
+    before_action :login_required
+
+    def new
+    end
+  end
+end
--- a/app/controllers/locations/pattern_search_controller.rb
+++ b/app/controllers/locations/pattern_search_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Locations pattern search form.
+#
+# Route: `new_location_search_path`
+# Only one action here. Call namespaced controller actions with a hash like
+# `{ controller: "/locations/pattern_search", action: :create }`
+module Locations
+  class PatternSearchController < ApplicationController
+    before_action :login_required
+
+    def new
+    end
+  end
+end
--- a/app/controllers/names/pattern_search_controller.rb
+++ b/app/controllers/names/pattern_search_controller.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+# Names pattern search form.
+#
+# Route: `new_name_search_path`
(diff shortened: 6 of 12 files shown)


nimmolo:
@mo-nathan Please let me know if this plan sounds OK to you, I ran it by Joe already.

nimmolo:
Faceted search form preview:
Screen Shot 2024-09-04 at 8 10 35 PM

nimmolo:
Reminder: must run when deploying
```ruby
TranslationString.rename_tags({ observation_term_date: :observation_term_when })
TranslationString.rename_tags({ search_term_date: :search_term_when })
```

nimmolo:
This PR keeps the forms hidden - they're only for testers, available at the unadvertised URLs:

localhost:3000/observations/search/new
localhost:3000/names/search/new

mo-nathan:
Shouldn't the Name search have a spots for a substring of the name like "Amanita"?  I also think it would be helpful to be able to search by author (e.g., "Murrill".  It would also be helpful to control whether the search is specific to a particular part of the Name.  For example, if I search for "Amanita" in the regular search box the first result is "Agaricus campestris" for so reason I don't understand.  I would expect it to return names that include the string "Amanita" first.  If no such names exist, then searching for synonyms or other fields that are on the page for that Name.  If there was an option for including synonyms as well as other content that would be great.

A pet peeve I have about our search (and search in general) is when you get results but it isn't clear why that is a result.  I recognize that this not the simplest thing to do, but I think it is a reasonable goal.

mo-nathan:
Hit a bug observations/search/new:
Screenshot 2024-09-22 at 7 23 07 AM

NoMethodError in Observations::FiltersController#create
undefined method `each' for nil
app/controllers/concerns/filterable.rb:59:in `block in remove_invalid_field_combinations'
app/controllers/concerns/filterable.rb:56:in `each'
app/controllers/concerns/filterable.rb:56:in `remove_invalid_field_combinations'
app/controllers/concerns/filterable.rb:43:in `sift_and_restructure_form_params'
app/controllers/concerns/filterable.rb:32:in `formatted_pattern_search_string'
app/controllers/observations/filters_controller.rb:58:in `set_pattern_string'
app/controllers/observations/filters_controller.rb:24:in `create'
app/controllers/application_controller.rb:227:in `catch_errors_and_log_request_stats'

nimmolo:
@mo-nathan Thanks for checking it out. I know you're busy with mobile.

- **Has author**, **Author**, **Include synonyms**:  I think what you missed is that the _fields are collapsed_ — these options are in the Names form already. (Or did you mean the obs search form?) Screengrabs, description and reason for collapsing are written up above in this thread. We may need the word "more" by the collapse caret, that might be more necessary here than in the create obs form. Or not... check it out with fresh eyes and let me know. 
- **Searching by substring** -  On Names you can currently search for "Amanita" and check `Include subtaxa?`, and that will get you the search you describe. (I had a "pattern" field at the top of both forms before, and removed it because I wasn't sure it did anything the subfields didn't already do. I still can't think of a use case on Observations that isn't covered by the subfields. It might still be useful, for some other reason, but let's try to think of one.)
- Thanks for the bug report.
- I don't know how we'd provide users with flash messages about the reasoning behind the search results, this would seem to require oracular incantations that i haven't learned at my level of apprenticeship. Useful though. Reasonable goal... for a different PR?

nimmolo:
@JoeCohen — I would like to hear your feedback before I write the tests for this.

@mo-nathan — Prefilled fields and selects are working again. Getting together code for the global toggle

JoeCohen:
Random suggestions:
### Observations
- Top level search form heading, e.g., Observation Search or Name Search (or even Search for Observations ...)
- Return to the form if there's an error or there are no hits. This gives users a chance to improve the search without losing what they've already entered. **IMO this is a blocker**.
- A nice-to-have would be a map where users could create/adjust a box instead of having to enter the coordinates.
- Arrange the NSEW fields to mimic their relative locations and how they appear in Locations
```
    N
  W   E
    S
```
- Limit form width on big screens. On my screen the form is almost 20" wide.
- Hide the old pattern search form, especially its Search button. 
- "Must exactly match the entire Locality field." should be visible on the form without clicking the help link. Otherwise the results can be unexpected, especially for new users.
- I get this error: `Filter terms come after the bare search pattern, “east:-121.297 north:45.6991 south:44.7226 west:-122.314 has_notes:yes”. Maybe you forgot to enclose the value for :region in double quotes?` 
for this search 
`http://localhost:3000/?pattern=name%3A%22Boletus+edulis%22+include_subtaxa%3Ano+notes%3AAbies+region%3AOregon%5C%2C+USA+east%3A-121.297+north%3A45.6991+south%3A44.7226+west%3A-122.314+has_notes%3Ayes`

### Names
- See above. Many comments also apply to Names
- Needs a Name field

Out of time & energy

PR #2338, opened 2024-08-26
Fix multiple autocomplete (on List form)

nimmolo:
The MO autocompleter is able to autocomplete multiple records in the same input, one after the other, if the user uses the expected "separator" (like a `\n` or `,`). Multiple autocomplete is used on the "new SpeciesList" form for a list of names. I unintentionally broke this at some point, this re-enables it.

This also (newly) gets "multiple-record" autocompleters to store the array of record ids of the matching records. That took quite a bit of refactoring of the Stimulus for autocompleters.

### DIFF
--- a/app/classes/auto_complete.rb
+++ b/app/classes/auto_complete.rb
@@ -46,6 +46,15 @@ class AutoComplete
     matches
   end
 
+  # returns an array of ONE { name:, id: } object. Uses `exact_match` which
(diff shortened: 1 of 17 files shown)


nimmolo:
TODO:
- [ ] Add autocomplete unit tests for `first_matching_record` / `exact_match`
- [ ] Add a system test for multiple autocomplete (`Name`)

PR #2344, opened 2024-09-01
Quick create from field slip

mo-nathan:
If you create a brand new fields slip (e.g., http://localhost:3000/qr/ATEST-00101) and you provide a location name and an id, the "Quick Create" button should create the observation without going through the regular "Create Observation" workflow.  If either of those are not provided or not already in the database, then you'll get directed to the regular "Create Observation" workflow.

The purpose of this is to speed up workflow for foray recorders who typically have 100s if not 1000s of field slips to process.


JoeCohen (approved):
Looks good.

PR #2346, opened 2024-09-02
Enables recorders to go directly to the field slip edit page

mo-nathan:
Recorders are currently defined as project admins during the project's date range.  To test:

1) Create a project whose current date range includes todays date (including projects with no date range).
2) Go to a QR code URL for an existing observation that uses the field slip prefix for that project.
3) This PR should take you directly to the field slip edit page rather than the corresponding observation (where you used to have to do two more clicks to get to the field slip edit page).

When the current time is outside the project's date range (or if the current user isn't a project admin), the QR code should take you to the corresponding observation.

This PR also switches back to saying "Create New Observation" on the edit field slip page vs. "Add Images and Other Data" to contrast with "Quick Create" when you are creating a new field slip.

### DIFF
--- a/app/controllers/field_slips_controller.rb
+++ b/app/controllers/field_slips_controller.rb
@@ -22,7 +22,11 @@ class FieldSlipsController < ApplicationController
       obs = @field_slip&.observation
     end
     if @field_slip
-      redirect_to(observation_url(id: obs.id)) if obs
+      if foray_recorder?
+        redirect_to(edit_field_slip_url(id: @field_slip.id)) if obs
+      else
+        redirect_to(observation_url(id: obs.id)) if obs
+      end
     else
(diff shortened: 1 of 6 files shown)


app/controllers/field_slips_controller.rb
```diff
@@ -115,10 +119,17 @@ def destroy
...
+    return false unless project
+
+    return project.is_admin?(User.current) && project.happening?
+  end
+
   def html_create
     if params[:commit] == :field_slip_quick_create_obs.t
       quick_create_observation
-    elsif params[:commit] == :field_slip_create_obs.t
+    elsif params[:commit] == :field_slip_add_images.t
```
mo-nathan, line 132: This is actually a fix on the previous PR.

JoeCohen (approved):
I ran the suggested manual tests.

PR #2347, opened 2024-09-02
Fix `id_of_nested_field` string extension

nimmolo:
This method is to translate a field name like `observation[name]` to a field_id `observation_name`, or a field name like `observation[naming][vote]` to the field_id `observation_naming_vote`.  It should leave a field name like `location` as is.

The simple gsub for the brackets will generate a trailing underscore that needs to be removed, if there are brackets. But if it's being run on all field names, the brackets may not be in the name, so it needs to check the last character before chomping.

`.chop` -> `.chomp("_")`


mo-nathan:
Tested it out and it fixes the broken test.  Thanks!

PR #2350, opened 2024-09-03
Have grep work on non-alphanumeric characters

mo-nathan:


JoeCohen (approved):
Interesting. Thanks for the fix!

PR #2352, opened 2024-09-03
Switch Location name cache to use Rails caching

nimmolo:
Changes `Location#location_exists?` to cache the existing location names via the environment's cache, instead of storing them in the class variable `@@location_cache`, which is flagged by Rubocop for thread safety, even if it is not an actual danger. 

The environment cache is set in `production` to `:solid_cache`. 

In `development` it's set to `:memory_store` if the server is run with `rails dev:cache`; otherwise there's no caching. 

As of this PR, in `test` the server runs with `:null_store`, which is as if there is a cache, but it is always expired.  The `Location` method will simply requery the cached names if it doesn't find them cached, so the method will return values either way.

> [!NOTE]
> This PR does not require any change to development db setups.
> There's currently no method to invalidate this cache, say if a new location is added, but this makes it easier to invalidate and potentially could we make it longer lasting.


app/models/location.rb
```diff
@@ -593,25 +593,25 @@ def self.countries_by_count
     CountryCounter.new.countries_by_count
   end
 
-  @@location_cache = nil
+  def self.location_name_cache
```
JoeCohen, line 596: Projects have Locations. Do those need to be taken into account?

nimmolo: Good question. For now, if the create project form allows people to sneak out without defining the location, yes, we should search for undefined strings in Projects. It's a tiny query too. 

With the new Location UI though, I'm trying to channel people to easily create locations in any context, and eventually seal off the exits so they cannot litter the db with undefined location strings that an admin must then deal with. When this is done, we can maybe delete all this code (here and elsewhere) that looks for stray "location" strings. 

But I haven't implemented the new Location UI for either `Lists` or `Projects` yet, nor have i disabled the ability to create objects with undefined "locations". 

So, "yes".

mo-nathan:
Why is this change being made?  I don’t understand the logic behind replacing a working in memory cache of locations with something that requires solid queue for the site to function and in my view inserts a lot more mechanism into a pretty simple concept.  Personally I would rather not have to start up solid queue every time I switch to do MO development.  I often shutdown MO tasks so I can start up the mobile development environments which are very resource intensive.

nimmolo:
Because of the class var, which is flagged by Rubocop as not thread safe. If you know otherwise, i'm all ears.

mo-nathan:
I believe caches are generally thread safe and I don't believe the class
variable in this case matters.  The problem with threading is if two
threads are trying to write *different* things to the same place in
memory.  For example, with User.current, the two sessions each want to set
it to the user associated with that session which are typically different.
Only one of them "wins" and ends up overwriting the value for the session
that happened to get their first.  Then when the first thread tries to read
User.current, it gets the wrong value.  In the case of a cache, both
threads are trying to write the same thing so it doesn't matter which of
them "wins".

app/models/location.rb
JoeCohen: Based on manual testing, you can create a Project without a Location.
@mo-nathan: Seems that we shouldn't allow that.

mo-nathan: There are expected to be Projects without Locations.  The vast majority (193 vs. 11) of existing Projects do not have a Location.  Projects for forays with locations are only one use case.

JoeCohen: Oops. Completely forgot that obvious fact; I'm too focused on Field Slips.

nimmolo:
Altered the PR so it does not require `SolidCache` in `test`. Rewrote the description. It now uses `:memory_store` in `test`.

It seems to me it is still worth doing this PR, because... we are indeed creating a "[low-level cache](https://edgeguides.rubyonrails.org/caching_with_rails.html#low-level-caching)" in Rails terminology, and this is the Rails way to cache query results like this. Rubocop is going to continue to flag all our class variables from here on out; I am persuaded they're a bad pattern even if in this case it's very unlikely to cause a snafu. The location cache will certainly change, and different instances may be the ones updating it.

nimmolo:
@mo-nathan Do you reckon this resolves it well enough to merge?

app/models/location.rb
nimmolo: Update: Projects do not have a `where` column, or any column for location strings. (Whew!) 

No need to check for orphaned strings there.

PR #2364, opened 2024-09-06
Add project add flash to quick create workflow

mo-nathan:


mo-nathan:
I don't understand what Coveralls is complaining about here with respect to this PR.  All the changes are covered when I run things locally and the stuff that shows in the Coveralls UI does not appear to be relevant.  I'm going to go ahead and merge this.

PR #2370, opened 2024-09-07
Add observation counts to project checklists

mo-nathan:
To test, view the Names tab/button from a project page.  Each name should be followed by a count of the number of times that name has been observed.

### DIFF
--- a/app/views/controllers/checklists/show.html.erb
+++ b/app/views/controllers/checklists/show.html.erb
@@ -8,6 +8,7 @@ add_tab_set(checklist_show_tabs(user: @show_user,
                                  project: @project,
                                  list: @species_list))
 names = @data.taxa
+counts = @data.counts if @data.show_counts
 %>
 
 <div class="my-4">
@@ -19,7 +20,12 @@ names = @data.taxa
 <%= panel_block(inner_class: "checklist") do %>
   <ul class="list-unstyled">
     <% names.each do |name| %>
-      <li><i><%= link_to(name[0], name_path(name[1])) %></i></li>
+      <li>
+        <i><%= link_to(name[0], name_path(name[1])) %></i>
+        <% if counts %>
+          (<%= counts[name[0]] %>)
+        <% end %>
+      </li>
     <% end %>
   </ul>
 <% end %>
--- a/app/classes/checklist.rb
+++ b/app/classes/checklist.rb
@@ -54,6 +54,8 @@ class Checklist
       raise("Expected Project instance, got #{project.inspect}.")
     end
 
+    def show_counts = true
+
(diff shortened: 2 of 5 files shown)


app/views/controllers/checklists/show.html.erb
```diff
@@ -19,7 +20,12 @@ names = @data.taxa
 <%= panel_block(inner_class: "checklist") do %>
   <ul class="list-unstyled">
     <% names.each do |name| %>
-      <li><i><%= link_to(name[0], name_path(name[1])) %></i></li>
+      <li>
+        <i><%= link_to(name[0], name_path(name[1])) %></i>
```
JoeCohen, line 24: Would it be more useful to link to an index of the Project's Observations of the Name?

JoeCohen (approved):
Looks good.

nimmolo:
@mo-nathan Would you be open to namespacing this PORO as `Checklist::ProjectCounter`?

I faced a steep learning curve in this "Classes" directory just trying to understand what the heck everything did, and my feeling is that namespacing makes the scope of use for these POROs ultra-clear. I did quite a bit of tidying up to subdirectory-ize these POROs so the folder wasn't such a grab bag.

app/views/controllers/checklists/show.html.erb
mo-nathan: It would be, but I'm not sure we have an existing URL that filters observations based on project membership and name.  If there is one, I'd be happy to use it.

mo-nathan:
> @mo-nathan Would you be open to namespacing this PORO as `Checklist::ProjectCounter`?
> 
> I faced a steep learning curve in this "Classes" directory just trying to understand what the heck everything did, and my feeling is that namespacing makes the scope of use for these POROs ultra-clear. I did quite a bit of tidying up to subdirectory-ize these POROs so the folder wasn't such a grab bag.

Done.

app/views/controllers/checklists/show.html.erb
nimmolo: Hmm, maybe i can supply from PatternSearch.
I'd also suggest using the `<i>` tag inside the link, so this:
```ruby
<%= tag.li do 
      link_to(observations_path(pattern: "project:#{@project.id} name:#{name.id") do 
        tag.i(name[0]
      end
    end %>
```
I'm pretty sure that will get you where you need to go.

nimmolo: Aside: It might be cool in the future to refactor this param (or do an alternate) so we could pass stuff like 
```observations_path(filter: { project: @project.id, name: name.id })```

PR #2374, opened 2024-09-07
More useful `Checklist` links for "Observations of name in ___"

nimmolo:
@mo-nathan you may have missed my idea for @JoeCohen's [suggestion](https://github.com/MushroomObserver/mushroom-observer/pull/2370#discussion_r1748683894) on your last PR, or it may not have worked at first. I realized this needed more logic to handle the various types of checklists.

Here's a solution - gives a link to "an index of the Project's (or User's, or Species List's) Observations of the Name. The URL is made using a `PatternSearch` filtered index of Obs, which I only just recently learned how to make. Example:
```ruby
observations_path(pattern: "project:#{@project.id} name:#{name[0]")
```

Aside: It might be cool in the future to refactor this param (or do an alternate like `filter`) so we could pass stuff like this, rather than a string for the `pattern` param.
```ruby
observations_path(filter: { project: @project.id, name: name[0] })
```

### DIFF
--- a/app/helpers/checklist_helper.rb
+++ b/app/helpers/checklist_helper.rb
@@ -10,7 +10,7 @@ module ChecklistHelper
     tag.li { link_to(link) { content } }
   end
 
-  def checklist_name_link_path(name, project, list)
+  def checklist_name_link_path(name, user, project, list)
     if user
       observations_path(pattern: "user:#{user.id} name:#{name[1]}")
     elsif project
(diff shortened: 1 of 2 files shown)


mo-nathan:
This doesn't seem all that urgent to me, but it would be nice.  Let me know when it's ready for review.

nimmolo:
@JoeCohen No need to review the PR, but you may want to note how to make that link for "Observations of Name in Project"

JoeCohen (approved):

nimmolo:
@mo-nathan This also is just waiting on your ok, or i can merge without.

PR #2375, opened 2024-09-07
Replace deprecated arel extensions == and !=

### LINKED ISSUE #2371: Fix deprecated arel_extensions
arel_extensions: `==` is now deprecated and will be removed soon. Use `.eq` instead.  
arel_extensions: `!=` is now deprecated and will be removed soon. Use `.not_eq` instead.

JoeCohen:
Replaces deprecated arel extensions `==`and `!=` with `.eq` and `.not_eq` per the test deprecation warning.

I used this regexp to find instances of the deprecated expressions:
```ruby
[A-Z]\w+\[:\w+\]( )(=|!)=
```

Unfortunately, the warning continues to appear even when there are no instances of these operators.
But this at least takes care of all known offenders

### DIFF
--- a/app/models/name.rb
+++ b/app/models/name.rb
@@ -350,25 +350,24 @@ class Name < AbstractModel
 
   # enum definitions for use by simple_enum gem
   # Do not change the integer associated with a value
-  enum rank:
-        {
-          Form: 1,
-          Variety: 2,
-          Subspecies: 3,
-          Species: 4,
-          Stirps: 5,
-          Subsection: 6,
-          Section: 7,
-          Subgenus: 8,
-          Genus: 9,
-          Family: 10,
-          Order: 11,
-          Class: 12,
-          Phylum: 13,
-          Kingdom: 14,
-          Domain: 15,
-          Group: 16 # used for both "group" and "clade"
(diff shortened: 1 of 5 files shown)


app/models/name.rb
```diff
@@ -350,25 +350,24 @@ class Name < AbstractModel
...
-          Genus: 9,
-          Family: 10,
-          Order: 11,
-          Class: 12,
-          Phylum: 13,
-          Kingdom: 14,
-          Domain: 15,
-          Group: 16 # used for both "group" and "clade"
-        }
+  enum :rank, {
```
JoeCohen, line 353: @nimmolo @mo-nathan @pellaea: Heads-up.
This change was automatic --without notifying me-- I guess via the combination of VSCode and RuboCop.
That's disturbing and probably reason for closer scrutiny of PRs from users of VSCode (@nimmolo, @JoeCohen)
But the change is okay: "Defining enums with keyword arguments syntax is deprecated and will be removed in Rails 8.0. Positional arguments should be used instead"
See https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsenumsyntax

PR #2376, opened 2024-09-08
Make FieldSlip.location smarter

mo-nathan:
This changes the way the field slip form location value is computed.  Rather than just being the value for any associated observation, if there is no such value, then it uses the last location used by this user for a field slip associated with the same project.  If that doesn't produce a value, then it uses any location set for the project.

### DIFF
--- a/app/models/field_slip.rb
+++ b/app/models/field_slip.rb
@@ -70,21 +70,32 @@ class FieldSlip < AbstractModel
   end
 
   def location
-    result = observation&.place_name || users_last_location
+    @location ||= calc_location
+  end
+
+  def location_name
+    location&.display_name
+  end
+
+  def location_id
+    location&.id
+  end
+
+  def calc_location
+    result = observation&.location || users_last_location
     return result if result
 
-    loc = project&.location
-    loc&.display_name || ""
+    project&.location
   end
 
   def users_last_location
     user = User.current
-    return unless user
+    return nil unless user
 
     field_slip = user.field_slips.where(project:).
                  order(updated_at: :desc).last
     obs = field_slip&.observation
-    obs&.place_name
+    obs&.location
   end
 
   def collector
--- a/app/models/field_slip.rb
+++ b/app/models/field_slip.rb
@@ -70,7 +70,21 @@ class FieldSlip < AbstractModel
   end
 
   def location
-    observation&.place_name || ""
+    result = observation&.place_name || users_last_location
+    return result if result
+
+    loc = project&.location
+    loc&.display_name || ""
+  end
+
+  def users_last_location
+    user = User.current
+    return unless user
+
+    field_slip = user.field_slips.where(project:).
+                 order(updated_at: :desc).last
+    obs = field_slip&.observation
+    obs&.place_name
   end
 
   def collector
--- a/app/views/controllers/field_slips/_form.html.erb
+++ b/app/views/controllers/field_slips/_form.html.erb
@@ -17,7 +17,7 @@
 <% if field_slip.code %>
   <%= form_with(model: field_slip) do |form| %>
     <% if field_slip.projects %>
-      <%= select_with_label(form: form, field: :project_id, inline: true,
+      <%= select_with_label(form:, field: :project_id, inline: true,
           label: :Project.t + ":", options: field_slip.projects) %>
     <% end %>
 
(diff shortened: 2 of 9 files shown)


app/models/field_slip.rb
```diff
@@ -70,7 +70,21 @@ def field_value(field)
   end
 
   def location
-    observation&.place_name || ""
+    result = observation&.place_name || users_last_location
```
nimmolo, line 73: Instead of `observation&.place_name`, how about `observation&.location` (and below)? That would allow you to also prefill the `location_id` in the autocompleter by sending a `hidden_id:` arg, giving the user that "verified" checkmark, although it may be getting that just fine on this branch.

With the new autocompleters I would expect we are getting actual `Location` records saved on all observations (to be verified...).

JoeCohen (approved):
Looks good as is. 
[@nimmolo's suggestion](https://github.com/MushroomObserver/mushroom-observer/pull/2376/files#r1749338320) could be an improvement, but I didn't have time to try to understand it.

app/models/field_slip.rb
nimmolo: @JoeCohen here's some context: 

The autocompleter is a lookup device that tries to match a string to a `Location` instance, but in the case of a preselected Location, _**we already have the instance**_. In the template, we can send 
- the `Location.name` to the autocompleter as the text `value`, and 
- the `Location.id` as the `hidden_id`.

The MO app has innumerable duplicate lookups of things by text string. My feeling is that if we are going to improve our provision of geographic UIs, we need to move away from thinking of "locations" as strings we pass around and parse, to passing around MO `Locations`,  instances of geographic data, which we usually already have. 

As it is, MO's code strikes me as a library full of librarians who refuse to use the Dewey decimal system, and only request books from each other by title and author, even when they have the code right in front of them. There's been very little consideration of the cost of these flurries of lookups.

mo-nathan: I think I see what you're driving at.  However, note that `Location.name` does not do the Postal/Scientific name thing which is why I'm using `place_name`.  There is `Location.display_name` that I could use instead.  Getting the `hidden_id` working would involve creating a FieldSlip.hidden_id, right?  It's not clear to me how that would work since the only use of `hidden_id` I see in the current code base is in `app/javascript/controllers/autocompleter_controller.js`.  What I was hoping you were saying is that if I passed the autocompleter a Location rather than a string that it would take care of the rest.  However, when I tried that I got this:
Screenshot 2024-09-08 at 7 35 01 PM

nimmolo: Yes - very close. When building out the autocompleters though, I found it was not always predictable what the identifiers of the two fields should be — like in Observation, for the Location autocompleter, the text field name has to be `place_name` (the visible field of the autocompleter) but i'm expecting the param `location_id` for the hidden field. 

That's why, to keep the autocompleter helper as versatile as needed (not guessing too much about the field names) the two field names and their values can be passed separately (and both are optional). 

To do what we're talking about here, you'd pass:
```ruby
autocompleter_field(form:, field:, value: @location.display_name, 
                    hidden_value: @location.id, hidden_name: #{see note})
```
So: The `hidden_value` saves the AJAX lookup from the text input when the form loads. 

Note: The name of the hidden field `hidden_name` should be whatever `param` you're handling it with in the receiving controller action. In this case, you're not handling it yet, and it can be whatever you like, but fyi the helper defaults to calling it `#{model_type}_id`, e.g., `location_id`. The only controller actions that handle this param currently are the obs create and edit actions, thus saving the second lookup. But i'm looking to roll that out more widely.

nimmolo: For reference, here's the relevant part of the HTML the autocompleter generates, the two inputs. The second is the hidden "matched record ids" field:
```html
<input placeholder="Start typing..." autocomplete="off" data-autocompleter-target="input" data-map-target="placeInput" data-action="..." class="form-control" type="text" value="USA, California, Alameda Co., Berkeley" name="observation[place_name]" id="observation_place_name">
<input value="742" data-autocompleter-target="hidden" data-map-target="locationId" data-north="37.9067" data-south="37.8359" data-east="-122.234" data-west="-122.327" class="form-control" readonly="readonly" autocomplete="off" type="hidden" name="observation[location_id]" id="observation_location_id">
```

nimmolo: @mo-nathan Updated my (overlong) response for clarity above. 

Sorry for the TMI, i guess it's a case of "in case you're interested this is how it works"

nimmolo (approved):

PR #2377, opened 2024-09-09
Field slip qr reader input

nimmolo:
This sketch of a QR code form lives at `field_slips/qr_reader/new`

- The field _focuses_ on form load, thanks to a new Stimulus controller for QR inputs that we can use for any kind of input that should accept a QR, if we need to.
- The form commits _on change_ of the input (also thanks to Stimulus), which I think is what we want for the reader sending the whole code at once. But that means you can't type into the input to test it. What you can do is paste a whole URL in , like `http://localhost:3000/info/search_bar_help`
This does the redirect as expected. Also, pasting in valid FieldSlip codes works.
- The form template is Turbo-ready (`.erb` not `.html.erb`), so it could be made to appear in a modal popup window from a button on an appropriate page
- The missing translation tag was named by AI, but it sounds official


nimmolo:
@mo-nathan Looks good. 

You had requested that it "submit on scan" but i notice you've disabled that in the Stimulus controller. 

Note that there is no `submit` button in the form. I haven't tested it with a QR reader, though, and I bet you have, so disregard if it submits anyway. (Or better yet let me know what the QR reader does!)

mo-nathan:
The submit seems to happen (at least in my browser) upon entering a carriage return which the scanner does automatically.

nimmolo:
Ah - that reminds me... slight sidetrack here.

I was really wanting to disable submit-on-return for the complicated forms like create Observation, and possibly overall. I have unintentionally submitted that form about 700 times while testing, because carriage returns are a valid input UI for selecting a value the autocompleters, and it's hard to keep track of where you are in the form sometimes.

This unpredictability is really counterproductive (at least to me). What do you think?

mo-nathan:
Yeah, I've seen the problem as well and would be in favor of turning it off most places as long as it's easy to turn on case by case.  The new QR code page definitely needs it.  I also like it for the search UI.

nimmolo:
Ok. That makes sense. 

At least with the new expanded search UI, though, it may be again confusing because of the autocompleters...

mo-nathan:
Are you still planning on there being a really simple search where the user just types something in and hits return?  I know it won't do a real Google like search, but I think the faceted search form could be overwhelming to a newbie who just wants to see some Amanitas or whatever.

nimmolo:
Oh yea. Definitely. 

The faceted form is just  what will appear if you click “Advanced Search”.

PR #2391, opened 2024-09-12
Add iNat to production credentials

JoeCohen:
Adds id & secret for an iNat "app" for which iNat calls the webserver with the authentication code. 
(Other environments use a different iNat "app" for which iNat calls http://localhost:3000.)


nimmolo (approved):

nimmolo:
Seems right to me. I decrypted locally and it seems A-OK.

JoeCohen:
@nimmolo: Thanks for indulging me.

PR #2392, opened 2024-09-12
Field slip subnote revisions

mo-nathan:
Also added the query_count gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -147,8 +147,6 @@ gem("prawn-manual_builder")
 # https://stackoverflow.com/a/23125762/3357635
 gem("puma")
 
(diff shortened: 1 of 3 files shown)


nimmolo:
@mo-nathan note that Rails 7.2 [gives a query count in the log without the gem](https://www.bigbinary.com/blog/rails-8-adds-sql-queries-count-to-template-rendering-logs). 

Maybe we just want to bump up to 7.2...

mo-nathan:
I don't want to open the potential can of worms involved in switching to 7.2 right now.  Happy to try it after NEMF.  I'll remove query_count.

PR #2396, opened 2024-09-15
Finalize NEMF field slips

mo-nathan:
Couple more tweaks for the NEMF field slip print run.


JoeCohen (approved):
Did the color logos print poorly if printing in black & white?
(I liked the color.)

mo-nathan:
Honestly this is entirely an optics thing.  It feels odd to have just the MO logo be in color.  There is also a general belief that black and white copies are cheaper than color copies, but I don't know for sure that that is still true.  Weirdly for the order at Staples I said "black and white", but they printed it with a color logo anyway.  If figured the simplest thing to do was just make the logo black and white and let Staples do whatever they want to do.

PR #2406, opened 2024-09-22
Comment to avoid future confusion

mo-nathan:


app/classes/api2/model_api.rb
```diff
@@ -61,6 +61,8 @@ def build_object
       validate_create_params!(params)
       obj = before_create(params) ||
             model.create(params)
+      # This generally means so validation was hit.
```
JoeCohen, line 64: I don't understand "so validation was hit".

mo-nathan: Typo.  I meant 'a validation error was hit'.  Updated.

Issue #2397, opened 2024-09-15
Clean up for Basic iNat Import


JoeCohen:
### TASKS
- [ ] #2387
- [ ] Delete unneeded success flash
- [ ] Delete JobTracker footer
- [ ] #2372 
- [ ] Fix CodeClimate offenses
- [ ] deep Rubocop (without inheriting from .rubocop_todo

JoeCohen:
Copilot query for handling iNat request errors:

> Ruby on Rails. Job perform calls import_requested_observations, which calls next_page, which calls "RestClient::Request.execute(
      method: :get, url: "#{API_BASE}/observations?#{query}", headers: headers
    )". How can I deal with this error by returning from nex_page and saving the RestClient error message in perform?

PR #2416, opened 2024-09-24
Inat::Obs class length

JoeCohen:
- Responds to part of #2397 by fixing the ClassLength offense via:
  - responding to @mo-nathan's suggestion to eliminate a repetitive pattern, and
  - using end-less method definitions.

There is no manual test.


JoeCohen:
### Replaced by #2418

PR #2418, opened 2024-09-25
Hash lookup for Inat::Obs attributes

JoeCohen:
- Uses hash lookup to access `Inat::Obs` attributes.
- Removes needless, repetitive pattern of defined methods.
Ex: `inat_obs[:id]` instead of `inat_obs.inat_id`
- Fixes `Inat::Obs` class length metric offense.
- Responds to [@mo-nathan suggestion](https://github.com/MushroomObserver/mushroom-observer/pull/2150#discussion_r1762039634).
- Replaces #2416 

There is no manual test.

### DIFF
--- a/app/classes/inat/obs.rb
+++ b/app/classes/inat/obs.rb
@@ -73,16 +73,14 @@ class Inat
       @obs[:ofvs]
     end
 
-    def inat_taxon
-      @obs[:taxon]
-    end
-
+    # NOTE: Fixes the ABC count of `snapshot` because
+    # inat_taxon_name is one fewer Branch than self[:taxon][:name]
     def inat_taxon_name
-      inat_taxon[:name]
+      self[:taxon][:name]
     end
 
     def inat_taxon_rank
-      inat_taxon[:rank]
+      self[:taxon][:rank]
     end
 
     ########## MO attributes
@@ -304,15 +302,15 @@ class Inat
         insert_rank_between_species_and_final_epithet
       elsif complex?
         # iNat doesn't include "complex" in the name, MO does
-        "#{inat_taxon[:name]} complex"
+        "#{inat_taxon_name} complex"
       else
-        inat_taxon[:name]
+        inat_taxon_name
       end
     end
 
     def infrageneric?
       %w[subgenus section subsection stirps series subseries].
-        include?(inat_taxon[:rank])
+        include?(inat_taxon_rank)
     end
 
     def prepend_genus_and_rank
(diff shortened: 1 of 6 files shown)


nimmolo:
Joe - I believe there is a better, more standard Railsy way to do all this, and there's an example on my PR:

Define your inat_obs as an activemodel object. This gives you the accessor methods without repetitive declarations, and other goodies besides. You won't have to access by hash key. 

It's a "model not backed by a db table" and it's a very convenient thing.

JoeCohen:
Thanks! Which pr has an example?

nimmolo:
#2337 — which as of yesterday does not require any migrations to check out, so it won't mess up your local db!

JoeCohen:
@nimmolo: TLDR: I like the idea of a model not backed by a db table, but will postpone it till after NEMF.

1. I want a good window, before NEMF, for live shakedown tests of iNat import.
2. What I have now works, so I'd prefer to postpone refactors until after NEMF.
3. The hash stuff is convenient because it differentiates the raw iNat data from methods which manipulates or map that data. But that might indicate that Inat::Obs has too many concerns, and therefore that I should create another class. That's something that I certainly don't want to attempt prior to NEMF.
4. The hash stuff also works with all levels of the iNat data. I don't know about a model not backed by a db table, and don't want to take time right now find an answer.

> Joe - I believe there is a better, more standard Railsy way to do all this, and there's an example on my PR:
> 
> Define your `inat_obs` as an `ActiveModel::Model` object, rather than a bare PORO.
> 
> This gives you the accessor methods without repetitive declarations, and potentially other goodies besides, like validation (although i didn't use that). You won't have to access your attributes by hash key. It makes it work like a model instance (because it is).
> 
> It's a "model not backed by a db table" and it's a [very convenient thing](https://testdouble.com/insights/rails-forms-non-active-record-objects). This is what people do for form data where the are assembling attributes that may go into some other model (like an `observation`, in your case) but they want the attributes of the form accessible like a model, too.
> 
> I've known about this for years but never understand quite how to use it, but now that I did, I realize how easy it is and will be using it elsewhere.
> 
> Check ObservationFilter at `app/models/observation_filter.rb` and NameFilter, which are based on a model SearchFilter that inherits from ActiveModel. It's at `app/models/search_filter.rb`.
> 
> ```ruby
> class SearchFilter
>   include ActiveModel::Model
>   include ActiveModel::Attributes
> 
>   attribute :pattern, :string
> ```

PR #2419, opened 2024-09-25
Notes link to iNat when recording Field Slip

JoeCohen:
- Adds a checkbox to the field_slip form. (See below.)
- Links notes[:Other_Codes] to iNat if the user checks that box
Screenshot 2024-09-30 at 7 20 37 AM
Screenshot 2024-09-30 at 7 30 52 AM

### Manual Test
- http://localhost:3000/field_slips/qr_reader/new
- enter a valid Field Slip Code
- complete the form, including Location, ID, and a number in Other Codes,
- check the box
- Quick Create
Expected Result: It shows a new Observation whose Notes[:Other_Codes] is a link to an iNat observation.

### DIFF
--- a/app/controllers/field_slips_controller.rb
+++ b/app/controllers/field_slips_controller.rb
@@ -205,9 +205,16 @@ class FieldSlipsController < ApplicationController
     notes[:Collector] = collector
     notes[:Field_Slip_ID] = field_slip_id
     notes[:Field_Slip_ID_By] = field_slip_id_by
-    notes[:Other_Codes] = params[:field_slip][:other_codes]
+    notes[:Other_Codes] = other_codes
     update_notes_fields(notes)
     notes
+  end
+
+  def other_codes
+    codes = params[:field_slip][:other_codes]
+    return codes unless params[:field_slip][:inat] == "1"
+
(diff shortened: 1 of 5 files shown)


app/controllers/field_slips_controller.rb
```diff
@@ -219,13 +226,9 @@ def update_notes_fields(notes)
...
 
-  def collector
-    user_str(params[:field_slip][:collector])
-  end
+  def collector = user_str(params[:field_slip][:collector])
 
-  def field_slip_id_by
-    user_str(params[:field_slip][:field_slip_id_by])
-  end
+  def field_slip_id_by = user_str(params[:field_slip][:field_slip_id_by])
```
nimmolo, line 231: This method `user_str` that parses `Jason Hollinger <jason>` to `Jason Hollinger` is repeated at several points in the code, including in pattern search. Can we please make this either
- A method of `User`
- A method of `StringExtensions`
and use that method everywhere in MO?

JoeCohen: I agree ti would be good to keep things DRY, but I think it should be in a different PR.
I'm trying to keep this PR tiny so that there's a shot at getting it deployed in the next 24 hours.

PR #2426, opened 2024-09-27
Switch obs search param from `date:` to `when:` 

nimmolo:
Changes the observation search keyword from `date:` to `when:`, and changes the translation string keys that relate to this term. The point of doing it here is to move the migration out of the pattern search PR to main, so it's easier for people to test that PR without messing up their db.
 
The reason for the change is that these search params map directly onto the attributes of a new `Filter` model used in the new search form, and `date` is a reserved accessor method for any model. 

But the cool thing about doing it this way is that these params are all translated; so for people who are already in the habit of typing `date:`, it will still work — it's "the English translation of the param 'when'". Session search strings with `date` will be silently updated to `when`, without the user having to do anything. Search bar help will reflect the new term.

### DIFF
--- a/.github/workflows/ci_rails.yml
+++ b/.github/workflows/ci_rails.yml
@@ -66,9 +66,6 @@ jobs:
           bundle exec rake db:schema:load
           bundle exec rake db:fixtures:load
 
-      - name: Run migrations
-        run: bundle exec rake db:migrate
-
       - name: Update translation files
         run: bundle exec rake lang:update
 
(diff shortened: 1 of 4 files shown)


nimmolo:
The "migration" involved here just renames the translation strings in the db. It does not change the schema, so loading the schema doesn't run the change.

Until now, CI did not run pending migrations. I guess we haven't had migrations in the past like this that only altered the db values, but did not change the schema. This adds `rake db:migrate` to `ci.yml` so that migrations like this will run and the tests can pass.

PR #2433, opened 2024-09-29
Attempt immediate send of verification email

mo-nathan:


nimmolo:
@mo-nathan   Does this work locally? (iirc local does not send emails...) How can we test?

I'm fine with putting it on production and trying it, then reverting if not.

mo-nathan:
I have no reason to believe it won't work.  As you mentioned locally doesn't actually queue email, but it does print the email into the server log which this does.

nimmolo:
It seems to me you've done it the way QueuedEmail is set up. Let's deploy it and test.

PR #2436, opened 2024-09-29
Add location mapping columns to locations and observations

nimmolo:
### Location: `box_area`, `center_lat`, `center_lng`
Previously the area of a location was accessible by the method `box_area`, now renamed `calculate_area`. But having the value cached in a column allows querying for things like `Location.where(box_area: ..MO.obs_location_max_area)`, and more easily excluding observation locations that are too vague. It also allows sorting locations by area, with a new sort button added on this PR.

### Observation: `location_lat`, `location_lng`
Caches the location `center_lat`/`center_lng` on the obs, for simpler querying of obs in a given area.
___
### Deployment
The column values must be manually populated on production after deployment with `Location.update_box_area_column`, but they are recalculated on save through a callback in the Location model.

### Goals
This change facilitates assigning mappable **center points** to Observations _without GPS_, where the associated Location area is not unreasonably big.


nimmolo:
Added a sort to the locations index by `box area` ⛰🏝🏕

PR #2449, opened 2024-10-03
Change naming log message to "Name proposed by user: name"

nimmolo:
...from "Proposed name created by user: name". 

Screen Shot 2024-10-03 at 3 58 55 PM

The current phrasing is easier to program, but drives me batty every time I read it.


JoeCohen (approved):
Yes, this is better.
(I haven't tested it locally because I'm having some software issues.)

Issue #2425, opened 2024-09-27
Simplify `Query` flavors


nimmolo:
From a discussion with @pellaea. Simplifying `Query` is a longtime goal, that will also make search queries easier to maintain and more readable.

1. Change internal uses of flavors, converting each usage one at a time:
from, e.g., `Query::ObservationByUser` to `Query::ObservationAll`
```ruby
# Turn this:
lookup_query(:Observation, :by_user, user)

# Into this:
lookup_query(:Observation, :all, user: user)
```
That can happen piecemeal, one query at a time, without breaking anything.  There will be a few cases that require additional work:
```ruby
advanced_search
needs_naming?
pattern_search
```
I think we might be able to get away with simply moving the additional parameters into base.  For example, pattern_search has the parameter "pattern".  It just does a subtle variation on "names" and "locations".  (Weird, I could swear pattern search used to also search the observation notes and comments!)  Anyway, the point is that in these cases where base doesn't have an exact equivalent, figure out how to implement something sufficiently similar with existing base parameters and/or add new base parameters to handle them.

2. Fill out the PatternSearch translator tables so that *every* base query parameter can be translated to and from search pattern parameter(s).  I think it makes most sense to maintain one-for-one correspondence.  There can be overlap.  For example, there is a "notes_has" query param, but the search pattern may have a single all-purpose "pattern" parameter that searches notes, comments, names, etc.  That's fine.  Just fill both out: add the all-purpose "pattern" parameter to Query::ObservationBase, and add the more specific "notes_has" to PatternSearch.

3. Provide a reverse translator to `PatternSearch` to convert `Query` instance parameters into a `PatternSearch` hash and/or string of parameters.  Once this is done, we should be able to convert *any* `Query` instance into a `PatternSearch` and vice versa.

4. Do whatever needs to be done to go back and forth between `PatternSearch` and your search form.  Sorry, I just don't have the capacity right now to dive into new code to see what that would involve.  But I would hope it would just involve some simple generic "escapers" and "unescapers".  After all, the values accepted by `PatternSearch` are already human-readable.

#### Notes

- Yes, all the subclasses of `ObservationBase` are flavors.

- Makes sense to stick with whatever is minimal to stabilize the existing search form PR.

re: how to cover the bare search word in existing pattern search -- Yes, I would make a new `Query::ObservationBase` parameter to cover that so that it is explicit and handled just like all the other search terms.

[...]

PR #2451, opened 2024-10-04
Start to refactor Query flavors

nimmolo:
Starts to move all Query flavors into `:all` using expanded params, to facilitate the storing of queries as a simple hash of params. Addresses #2425. 

This PR does the easier ones.


nimmolo:
@JoeCohen I'm noticing that `SequencesController#index` does not handle pattern searches, or index params via the dispatcher in `ApplicationController` like other indexes. This means that pattern searches for sequences will not work, `param[:pattern]` is ignored by the index.

I assume there's a reason for that, but I wanted to check with you in case that's unintentional.

nimmolo:
Oh - we don't offer Sequence pattern search. Interesting that the query can handle the pattern param.

JoeCohen:
It simply hadn't occurred to me to offer Sequence pattern search.
It's a poor way to search for DNA patterns. Might be useful for text or other parts of an MO Sequence.
IMO low priority.

nimmolo:
This is ready to go, but i can hold until after NEMF if people prefer.

I did this refactor according to @pellaea's advice, and Query is pretty well tested, but I don't want to be the one messing up NEMF.

PR #2454, opened 2024-10-04
improve location fixtures

nimmolo:
This gives all location fixtures a `box_area`, `center_lat` and `center_lng`.

### DIFF
--- a/test/fixtures/locations.yml
+++ b/test/fixtures/locations.yml
@@ -17,8 +17,6 @@ DEFAULTS: &DEFAULTS
   west: -1
   east: 1
   south: -1
-  high: 1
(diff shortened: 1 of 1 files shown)


nimmolo:
@JoeCohen I think this is good to merge. Have a look and let me know what you think, I noticed you moved it into drafts.

The location fixtures that give me pause are the ones that have an area equal to the Earth. I feel like these should get much more specific locations that better reflect how they're used in the tests. But that doesn't have to be on this PR.

JoeCohen:
will look in 20 minutes. moving it to a draft was an accident

JoeCohen (approved):

PR #2470, opened 2024-10-07
Sort out current `rss_logs` fixture situation 

nimmolo:
This PR makes some minimal changes to the obs/rss_log fixtures to get them on a stable footing:
- adds a test to be sure all Observation fixtures have an associated RssLog fixture
- gives default values for `created_at`, `updated_at`, and `log_updated_at` for observation fixtures
- adds a method to look up a fixture's commonly used name (e.g. "roy") by its id, useful for error messages in loops where you're not sure which fixture is throwing the error
- Replaces `rss_log_id: null` values for two observation fixtures: `deprecated_name_obs` and `reused_observation`

Even these few changes (most likely the change in default dates) seems to have thrown off two integration tests regarding project constraints. I'm not very familiar with these tests, but after reading through them and debugging, i decided they still basically seem to work as intended if i simply adjust the expectations for the new reality of the fixtures. 

@mo-nathan and @JoeCohen — if you have a moment, amid everything else you're doing, please review the tests to check that these are still testing what they need to be testing. If not, this can wait til after NEMF.

### DIFF
--- a/app/extensions/fixture_set_extensions.rb
+++ b/app/extensions/fixture_set_extensions.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class ActiveRecord::FixtureSet
+  # Reverse lookup of fixture label (how we refer to fixtures) by id.
+  def self.reverse_lookup(table, id)
+    ActiveRecord::FixtureSet.all_loaded_fixtures[table.to_s].
+      fixtures.each_key do |key|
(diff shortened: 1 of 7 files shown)


nimmolo:
@JoeCohen I assume you've checked this out by now.

I'm going to merge it in a few minutes, because it only changes fixtures and tests, and I realized it could not possibly mess up the server for NEMF.

PR #2484, opened 2024-10-12
Location contains edges

JoeCohen:
- Fixes contains_lng? to make Location contain its E/W edges
- Fixes bug found by @mo-nathan during NEMF


JoeCohen:
Merging this immediately because it's a hot fix and a trivial change in the code.

PR #2486, opened 2024-10-16
Fix date handling on Field Slip form

mo-nathan:


app/controllers/field_slips_controller.rb
```diff
@@ -146,16 +131,23 @@ def html_create
...
     # Must have valid name and location
     location = Location.place_name_to_location(place_name)
     flash_error(:field_slip_quick_no_location.t) unless location
-    name = Name.find_by(text_name: params[:field_slip][:field_slip_name])
+    name = Name.find_by(text_name: fs_params[:field_slip_name])
     flash_error(:field_slip_quick_no_name.t) unless name
     notes = field_slip_notes.compact_blank!
-
-    obs = Observation.build_observation(location, name, notes)
+    date = nil
```
mo-nathan, line 144: This is the interesting bit for this PR.

app/controllers/field_slips_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class FieldSlipsController < ApplicationController
+  include Show
```
mo-nathan, line 4: Tried moving `def show` into the show.rb file, but got Rubocop complaints.  Not sure what that's all about.

PR #2491, opened 2024-10-18
Fix collector name on the edit field slip page

mo-nathan:
To test:

1) Create a field slip (e.g., enter "NEMF-123" into the QR reader page on your local) with a collector that is not an MO user.
2) Save it.
3) Edit it.
4) On main it will insert your user name.  On the branch it will have the non user name you inserted.

### DIFF
--- a/app/models/field_slip.rb
+++ b/app/models/field_slip.rb
@@ -99,7 +99,7 @@ class FieldSlip < AbstractModel
   end
 
   def collector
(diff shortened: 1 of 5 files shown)


JoeCohen (approved):
Code looks good.
I think you should deploy if all tests pass. 
(I always worry about the effect of adding more fixtures.)

PR #2493, opened 2024-10-18
Allow quick create with no name

mo-nathan:
Create a new field slip
Ensure that no name is specified
Click 'Quick Create'
On main it should take you to the Create Observation page with a flash error
On this branch it should just create the observation with the name "Fungi"

### DIFF
--- a/app/controllers/field_slips_controller.rb
+++ b/app/controllers/field_slips_controller.rb
@@ -131,7 +131,6 @@ class FieldSlipsController < ApplicationController
     location = Location.place_name_to_location(place_name)
     flash_error(:field_slip_quick_no_location.t) unless location
     name = Name.find_by(text_name: fs_params[:field_slip_name])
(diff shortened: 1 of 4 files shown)


mo-nathan:
Also moved the save edit/create button up on the observation form.

mo-nathan:
@JoeCohen I'm a bit nervous about moving the edit/create button for observations just because I'm moving user's cheese, but I like it a lot better on the. mobile app.  Thoughts?

JoeCohen (approved):
Code is great.
I like the `Submit Edit` on top where you put it, because it's always visible to me; my work normal work flow would be fold sections when I'm done with them. 
Perhaps @nimmolo had a reason for preferring it further down. E.g., if users scroll down instead of folding sections, then the button will tend to disappear.
Maybe there should be multiple buttons.

nimmolo (approved):

PR #2502, opened 2024-11-04
Avoid Earth

### LINKED ISSUE #2501: Observation with geoloc should use smallest containing Location instead of Earth
If an Observation has a geoloc then Observation.location should be the Location with minimum bounding rectangle which contains obs.lat and obs.lng, instead of Earth.
> Nathan Wilson Sun, Nov 3, 9:06 AM (23 hours ago)
Sounds great!
>> On Sun, Nov 3, 2024 at 11:10 Joseph Cohen <jdcohenesq@gmail.com> wrote:
... I also don't like all the Obss have geoloc, but have lat/ng. These mostly (all? 99%) are coming from the API. 
Simple solution?:
>> - Observation before_save callback (pseudo-code)
>> - where = minimal_bounding_rectangle(lat, lgn)  where == Earth && lat.present?; 
Possibly limit this to Observations being created via the API. ...

### Tasks
- [x] API Test of posting Obs with lat/lng & Earth
- [x] Check case of lat/lng not contained in Earth (currently 89, -89, 179, -179)
- [ ] Resolve Earth fixture vs db discrepancy
- [x] Possibly deal with weirdly incorrectly defined Locations.

JoeCohen:
Uses the smallest Location containing lat/lng instead of Earth, for Observations.

(Explanation: Many recent Observations have lat/lng with Location Earth. Most (all? 99%) of them some from the mobile app or otherwise from the API. This frustrates mobile app users. It also stops the Observation from appearing on maps of its taxon (the location is too vague). This PR aims to improve things by preferring the smallest reasonable Location to `Earth` if an Observation has a geolocation.)

The PR adds:
- a Location scope, `with_minimum_bounding_box_containing_point`
- an Observation before_save callback, `prefer_minimum_bounding_box_to_earth`
- corresponding code in the Observation API

### Using the website
- [Create Observation](http://localhost:3000/observations/new)
  - Locality: `Earth`
  - Latitude: `46.438`
  - Longitude: `-121.84`
  - `Create`
  
**Expected result: (new) Observation with Location: Cispus Environmental Learning Center**

### Using the API
- Enter the following on the unix command line, substituting a valid API Key for `xxx`
```zsh
curl -X POST "http://localhost:3000/api2/observations?api_key=xxx&location=Earth&latitude=46.438&longitude=-121.847&date=2024-11-05"
```
- The response should be something like
```xml
<?xml version="1.0" encoding="UTF-8"?>
<response xmlns="http://localhost:3000/response.xsd">
  ...
  <results number="1">
    <result id="565807" type="observation"/>
  </results>
  ...
</response>
```
- View the Observation which has the `id` displayed in the above response.

**Expected result: Observation has Location: Cispus Environmental Learning Center**

### DIFF
--- a/app/classes/api2/observation_api.rb
+++ b/app/classes/api2/observation_api.rb
@@ -372,14 +372,14 @@ class API2
       @latitude  = parse(:latitude, :latitude)
       @longitude = parse(:longitude, :longitude)
       @altitude  = parse(:altitude, :altitude)
-      prefer_minimum_bounding_rectangle_to_earth!
+      prefer_minimum_bounding_box_to_earth!
     end
 
-    def prefer_minimum_bounding_rectangle_to_earth!
+    def prefer_minimum_bounding_box_to_earth!
       return unless Location.is_unknown?(@location) &&
                     @latitude.present? && @longitude.present?
 
-      @location = Location.with_minimum_bounding_rectangle_containing_point(
+      @location = Location.with_minimum_bounding_box_containing_point(
         lat: @latitude, lng: @longitude
       ).name
     end
--- a/app/classes/api2/observation_api.rb
+++ b/app/classes/api2/observation_api.rb
@@ -372,6 +372,16 @@ class API2
       @latitude  = parse(:latitude, :latitude)
       @longitude = parse(:longitude, :longitude)
       @altitude  = parse(:altitude, :altitude)
+      prefer_minimum_bounding_rectangle_to_earth!
+    end
+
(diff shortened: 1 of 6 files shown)


nimmolo:
Allow "Earth" with lat/lng? Heck no. 

I don't feel we should allow Earth as an obs location, ever.

JoeCohen:
> Allow "Earth" with lat/lng? Heck no.
> 
> I don't feel we should allow Earth as an obs location, ever.

Thanks @nimmolo. I agree about "ever", but that's for another PR.

PR #2520, opened 2024-11-10
Add user quotas for showing original images

pellaea:
This was my first experience using AI to help write code.  Dang!  It really works!  I feel almost superfluous!

So, this adds a count to the users table that gets reset every day. If you go over the limit it pops up an error message (javascript alert).

When you click on "Show Original" it checks first if it's one of the images that's being served from the cloud.  Remember, even though we transfer all originals to google cloud as soon as we get them (once a week? the cronjob is on the image server), we only *delete* them periodically by hand, like 100,000 at a time or whatever.  Then we update the nginx configuration to start serving the deleted images from the cloud. I've added this threshold to config/consts.rb.  This needs to be kept in sync with the nginx configuration and the image server whenever we delete more original images from the image server to free up space!

Next, it checks if the image is already cached locally.  Then it checks the user quota.  Then it finally launches the job to download the image to local cache.

Meanwhile a stimulus controller polls every second to see when a file appears.

When the image is available (regardless where it is), it opens the original image in a new tab.

That's it.

I have not written tests yet, that's up next.  (Actually, without credentials or my own private test cloud storage service, I've only been able to half-test it in development, too.)

Questions:

1. How many originals per day per user? -- I think simply requiring login is enough and quota is unnecessary, so I've set it high (1000 I think).  Thoughts?

2. Is polling for the file every second too much?  How often is safe?  I guess we should check how expensive the request is, perhaps stripping some of the unnecessary filters from that action.

3. I'm not sure whether the file will appear atomically.  If not, we should download it to a temp file then move it atomically into place.  Does anyone happen to know how google-cloud-storage gem works in this respect?

4. Also, should we worry about multiple users trying to download the same image simultaneously?  I suspect this will happen exceedingly rarely because: a) recent observations will have their images still on the image server and so skip this whole mess, b) it will only take a fraction of a second to download the image from the cloud(?), so what are the odds two requests will overlap?, c) I can't think of any really bad side-effects even if it were to happen.  Won't it just overwrite the file with itself, no harm done?


mo-nathan:
Awesome!  Thanks for the work on this.  I know the feeling about using AI.  However, I've come to realize that it actually let's me focus on what actually matters which is setting the goal/intention for the code change.  If it can handle the labor of writing the code to realize my intention, that's nothing but a win.

Here are my takes on your questions:

1. I was thinking 100 images/user and 1000 for the whole site per day.  I have a hard time imagining a human doing it manually 100 times in a day or it happening 1000 times in a day across the whole site and it not being something that they should ask for our help with.  However, a 1000 per user per day is OK with me.
2. Polling every second makes sense to me (and is what I do for the field slip generator).  A second is a long time for a computer.  The file check itself should be a matter of milliseconds.
3. I would use the temp file approach.  It's simply a better more reliable design.
4. I would not worry about multiple simultaneous downloads.  The worst I can imagine is the first write happens.  The polling code tells the user "your file is ready".  The second write happens (to a temp file by the way) and then the new copy gets swapped in.  I really can't see the file system screwing that up.  Even if it did, they would just try again and it would fine.

mo-nathan:
I scanned the PR and it looks reasonable to me.  The one thing that I didn't spot that I think would be really helpful is some sort of usage tracker.  It would be really handy to know how much it's getting used, what maximum numbers we're seeing, and how often the limit gets hit.

pellaea:
Thanks for looking at it, Nathan.  I changed it to download to a tempfile as agreed.  What did you have in mind for stats?  Maybe just a quick line in the log file?

pellaea:
Okay, so maybe you're thinking about a new table that logs image id, user
id and time for each request?

mo-nathan:
Yeah, something like that.  Might also want a Boolean to mark if it is
successful or if it was blocked.  We may never get a block but if we do
we’ll want to know.

pellaea:
I think this might be ready to go.

mo-nathan:
I told CodeClimate to chill out about those issues.

app/controllers/images/originals_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      if on_image_server?
+        render(json: { status: "ready", url: @image.original_url })
+      elsif already_cached?
+        render(json: { status: "ready", url: @image.cached_original_url })
+      elsif already_loading?
+        render(json: { status: "loading" })
+      elsif user_maxed_out? || site_maxed_out?
+        render(json: { status: "maxed_out" })
+      else
+        # Save job for tester.
```
mo-nathan, line 41: I don't understand this comment.

app/controllers/images/originals_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      # Because this is only called from stimulus within valid pages, we
+      # can assume the user is already logged in and the image is correct...
+      # unless a malicious actor is calling this directly, in which case we
+      # don't really care if errors are handled gracefully.
+      @user = User.current = session_user or raise("login required")
+      @image = Image.safe_find(params[:id]) or raise("image not found")
+      log_request(@user, @image)
+
+      respond_to do |format|
+        format.html { redirect_to(@image.cached_original_url) }
```
mo-nathan, line 19: I think another way to avoid the complaint about this line would be to create a route for it that generates the right URL - something like `local_original_image` and then use `local_orginal_image_path(@image.id)`.

app/javascript/controllers/image_loader_controller.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  async load(event) {
+    event.preventDefault()
+
+    if (this.isPolling) return
+
+    const url = this.linkTarget.href
+    this.originalText ||= this.linkTarget.textContent
+    this.isPolling = true
```
mo-nathan, line 22: Seems like this should at least be just below line 18.  Ideally there would be a way to make test/set really atomic, but I don't know how to do that in JavaScript (but I imagine there must be some way).

mo-nathan (approved):
Just a few minor comments.  I leave it to you to decide if they warrant more work.

app/controllers/images/originals_controller.rb
pellaea: Would `# Save job for unit tests via assigns(:job).` make more sense?  Or is there a better way to check in the unit tests whether a job was queued?

app/controllers/images/originals_controller.rb
pellaea: Interesting, I didn't know routes could do that, will look into...

app/javascript/controllers/image_loader_controller.js
pellaea: I think this fires on the user clicking the link.  The only way we could get a race condition here is if the user could click ultra fast!  I'll move the line up, but I'm not gonna sweat figuring out how to make groups of statements atomic in JS.  (I might ask google anyway out of curiosity... it may be Atomics.compareExchange...)

mo-nathan: Yeah, it seems unlikely to be a problem.  It's just seems like at least setting it as close as possible makes sense and reduces the chance of someone later coming along and introducing something that causes enough of a delay that it could become a problem.  It's the kind of thing that would be really hard to debug later.

app/controllers/images/originals_controller.rb
mo-nathan: I don't really see the connection to the unit tests.  I must be missing something if `perform_later` is not intended to run in production.

pellaea: I want to test to make sure it actually queued a job in the unit test.  You're right, it's not intended to happen immediately, which is why it's perform_later.  (Actually, we wouldn't mind it happening immediately, but I assume one uses perform_later to ensure that it is done asynchronously, and that perform_now may literally block the execution thread?  But my intuition is frequently way off when it comes to rails!, so it's worth asking...)

But with perform_now it definitely doesn't happen immediately, so there's no way for the unit test to know by the *effects* of the job.  Therefore the unit test needs to be able to test for the existence of the (unperformed) job, right?  I don't care how.  This just seemed the easiest way that I personally knew how to do.

app/controllers/images/originals_controller.rb
pellaea: I'm struggling with this.  This is close:

```
resources :images, only: [:show] do
  member do
    get("cached_original", to: static("/orig_cache/#{id}.jpg")
  end
end
```

But I can't figure out how to get the id in there.  (And ensure that id is an integer!  Otherwise our security "fix" ends up making it *less* secure!)  And recommendations?

config/initializers/google_cloud_storage.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require "google/cloud/storage"
+require "json"
+
+file = Rails.root.join(
+  "config/credentials/mo-image-archive-service-account.json"
+)
+# For some reason the autotester on github will still try to load this file
+# even though it should presumably be in test mode.
+if File.exist?(file) && !Rails.env.test?
```
JoeCohen, line 11: Would you be happier with a guard clause
```ruby
return if !File.exist?(file) || Rails.env.test?
```

test/jobs/image_loader_job_test.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def teardown
+    FileUtils.rm_rf(DIR)
+  end
+
+  test "image download works" do
+    rolf_count = users(:rolf).original_image_quota
+    admin_count = User.admin.original_image_quota
+    with_stubs do
+      ImageLoaderJob.perform_now(@test_image.id, users(:rolf).id)
```
JoeCohen, line 58: Question: I thought tests always perform_now. Is that correct?

JoeCohen (approved):
I don't really have anything to contribute.

test/jobs/image_loader_job_test.rb
pellaea: Ah, this is something I actually know something about!  No.  It was essentially random before, but now I have it configured

```
config.active_job.queue_adapter = :test
```

Which means the jobs do not run unless you explicitly queue them as "perform_now".

It may be the case that they were set to start immediately before, but if so, they were running asynchronously resulting in random behavior in the tests that really makes testing a nightmare.  It was acting like the jobs were persisting from one test to another in some cases, which is just crazy because of stubs changing.  I have no idea what the expected behavior in such a case would be!!

I use `ImageLoaderJob.perform_now` to test the `perform` method in test/jobs/image_loader_job.test.rb, and that seems to be stable.

config/initializers/google_cloud_storage.rb
pellaea: I wouldn't be any happier, but if everyone else would be, I can change it!  There's something odd about returning from a config file... :)

app/javascript/controllers/image_loader_controller.js
pellaea: Done.

PR #2526, opened 2024-11-15
Allows note fields to have '+'s without Textile badness

mo-nathan:
To test:
Enter '+photo' into a data field (but not the last data field).
The branch renders it correctly.
Current production does something like:

<ins>Other</ins>: <ins>photo</ins>
<ins>+Trees/Shrubs</ins>: pitch pine

See https://mushroomobserver.org/564569 for a live example of the issue.


JoeCohen (approved):
Looks good. No other edge case occurs to me.
Is it worth adding a test to cover line 256?

mo-nathan:
Test added for the issue being addressed and it provides coverage.

PR #2549, opened 2024-11-24
Change how checklists handle synonym and deprecated names

mo-nathan:
Example checklists include http://localhost:3000/checklist?id=1, http://localhost:3000/checklist?project_id=352, and http://localhost:3000/checklist?species_list_id=19.  Synonyms are no longer folded together, but instead marked with "+".  Deprecated names are marked with "*".


JoeCohen (approved):
Thanks for dealing with this.
The results are great. (I tried it on a bunch of Lists, didn't review the code.)
`*`, `+` plus footnotes worked out great as a way of explaining what the suer is seeing.

PR #2551, opened 2024-11-30
Better banners

mo-nathan:
Banners should now get updated correctly for all users when you change them in the "Change Site Banner" page as an admin.
Once a user dismisses a banner it should remain dismissed until an admin changes the banner.
Old banners that get changed using this mechanism should remain available in the database in the "banners" table.
The en.txt file does not get changed when the banner changes.


mo-nathan:
Keep this a draft until Ive implemented "un-dismiss" icon.

JoeCohen (approved):
Thanks for doing this. Very helpful.

- I tried it locally and it works smoothly. I did not review the code.
- I wouldn't mind if the banner re-appeared after a user logged out then logged in.
- Can `app/models/language_exporter.rb#update_export_file` be covered with reasonable effort?

app/assets/stylesheets/mo/_links_buttons_alerts.scss
```diff
@@ -35,6 +35,39 @@
...
+#show-banner-container {
+  width: 100%;          /* Full width */
+  position: relative;   /* Keep it in document flow */
+  background-color: #f8f9fa; /* Optional background for visibility */
+  padding: 5px 0;      /* Add padding to give space */
+  text-align: right;   /* Center the chevron horizontally */
+  z-index: 1;           /* Ensure it's above content if needed */
+  display: none;        /* Initially hidden */
+}
+
```
nimmolo, line 49: I would suggest using CSS classes rather than ids for styling, and adding a CSS ID supplementally for JS identification if necessary. (With stimulus, even that is probably not necessary, because the root "controlled" element is always accessible in the stimulus controller.) 

I would recommend making this change even though the banner seems like it can be safely presumed to be unique on the page. In whatever future development contingency where there may be two, it will cause difficult to debug issues.

Another suggestion: use BS utility classes on the element, and remove some of this specificity in the class.
`<div class="show_banner_container position-relative w-100 py-2 text-right">`. You can also add the class `d-none` but it has `!important` so you'd want to modify your stimulus controller to add or remove the `d-none` class.

I will add these suggestions again below on the files i think they'd be applicable.

app/assets/stylesheets/mo/_links_buttons_alerts.scss
```diff
@@ -35,6 +35,39 @@
...
+  .glyphicon-chevron-down {
+     display: block;
+   }
+}
+
+#show-banner-icon.bottom {
+  top: auto;    /* Reset top position */
+  bottom: 20px; /* Place the chevron back at the bottom */
+}
+
```
nimmolo, line 70: here: `.show_banner_icon` instead of the CSS ID, and you can use the utitlity class `d-block text-green` on the element.
```scss
.show_banner_icon {
  font-size: 1em;
  cursor: pointer;

  &:hover {
    color: darkgreen !important;      /* Change color on hover */
  }
    etc...
```

app/views/controllers/admin/banners/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1><%= :change_banner_title.t %></h1>
+
+<%= form_with model: @banner, url: admin_banners_path, local: true, method: :post do |f| %>
+  <div>
+    <%= f.label :message %><br>
+    <%= f.text_area :message, rows: 5, cols: 120 %>
+  </div>
+  <%= f.submit :banner_update.t %>
```
nimmolo, line 8: No div needed - please do use the MO form helpers here, that will help keep all our markup consistent during Bootstrap upgrades. 
```suggestion
  <%= text_area_with_label(form: f, field: :message, rows: 5) %>
  <%= submit_button(form: f, button: :banner_update.t) %>
```
Also, our Ruby style is parens for method calls, which I'd prefer be consistently used also in ERB for legibility.

app/views/controllers/application/app/_banners.html.erb
```diff
@@ -9,28 +9,26 @@
...
-      <%= banner_text %>
+      <p><%= Banner.current.message.t %></p>
     </div>
-  <% end %>
 
+    <div id="show-banner-container">
+      <div data-banner-target="showButton" id="show-banner-icon" style="display: block;">
+        <span class="glyphicon glyphicon-chevron-down" aria-hidden="true"></span>
+      </div>
+    </div>
```
nimmolo, line 31: Incorporating the changes suggested above:
```suggestion
    <div class="show_banner_container position-relative w-100 py-2 text-right" id="show-banner-container">
      <div data-banner-target="showButton" id="show-banner-icon" style="display: block;">
        <span class="glyphicon glyphicon-chevron-down" aria-hidden="true"></span>
      </div>
    </div>
```

app/javascript/controllers/banner_controller.js
```diff
@@ -1,20 +1,54 @@
...
-      cookie_name = "hideBanner2=" + time + "; ";
+  isBannerDismissed() {
+    const version = this.dismissButtonTarget.dataset.version;
+    return document.cookie
+      .split("; ")
+      .some((cookie) => cookie === `dismissed_banner_version=${version}`);
+  }
+
+  hideBanner() {
+    this.bannerTarget.style.display = "none";
```
nimmolo, line 40: This is fine, but fwiw I've been adding and removing Bootstrap CSS classes to do this. I like it because it allows other behavior to be attached to the class in the future, for example fade-outs, etc.

To do it that way, assuming the element had the class "d-block" on it to begin with:
```suggestion
    this.bannerTarget.classList.remove("d-block");
    this.bannerTarget.classList.add("d-none");
```

app/javascript/controllers/banner_controller.js
```diff
@@ -1,20 +1,54 @@
...
+  }
 
-    date.setTime(date.getTime() + (expiresInDays * 24 * 60 * 60 * 1000));
-    const expiresText = "expires=" + date.toUTCString() + "; ";
+  showBanner() {
+    this.bannerTarget.style.display = "block";
```
nimmolo, line 44:
```suggestion
    this.bannerTarget.classList.remove('d-none');
    this.bannerTarget.classList.add('d-block');
```

app/javascript/controllers/banner_controller.js
```diff
@@ -1,20 +1,54 @@
...
+  }
+
+  hideShowButton() {
+    document.getElementById("show-banner-container").style.display = "none";
```
nimmolo, line 48: I have discovered it's unreliable to use `document.getElementById` in stimulus, i strongly recommend changing this.

The way to go is to make the container a Stimulus target. But in this case I'd make it even simpler. I'd make the container the root element of the controller, eliminate the wrapping div, and just call `this.element` to access the container.

app/javascript/controllers/banner_controller.js
```diff
@@ -1,20 +1,54 @@
...
+  }
 
-    document.cookie =
-      cookie_name + expiresText + "samesite=lax;path=/";
+  showShowButton() {
+    document.getElementById("show-banner-container").style.display = "block";
```
nimmolo, line 52: Also here: either `this.containerTarget` or `this.element`, depending on which div you put the controller on.

app/views/controllers/application/app/_banners.html.erb
```diff
@@ -9,28 +9,26 @@
...
+	aria-label="Close"
+	data-version="<%= Banner.current.version %>">
+        <span class="glyphicon glyphicon-chevron-up" aria-hidden="true"></span>
       </button>
-      <%= banner_text %>
+      <p><%= Banner.current.message.t %></p>
     </div>
-  <% end %>
 
+    <div id="show-banner-container">
```
nimmolo, line 27: See stimulus recommendation below.
```suggestion
    <div id="show-banner-container" data-banner-target="container">
```

app/javascript/controllers/banner_controller.js
nimmolo, line 52:
```suggestion
    this.containerTarget.classList.remove('d-none');
    this.containerTarget.classList.add('d-block');
```

app/javascript/controllers/banner_controller.js
nimmolo, line 48: I have discovered it's unreliable to use `document.getElementById` in stimulus, i strongly recommend changing this.

The way to go is to make the container a Stimulus target, and just call `this.containerTarget` to access the container.

```suggestion
    this.containerTarget.classList.remove('d-block');
    this.containerTarget.classList.add('d-none');
```

app/views/controllers/application/app/_banners.html.erb
```diff
@@ -9,28 +9,26 @@
...
+      <div data-banner-target="showButton" id="show-banner-icon" style="display: block;">
```
nimmolo, line 28:
```suggestion
      <div data-banner-target="showButton" class="show_banner_icon d-block text-green">
```

app/assets/stylesheets/mo/_links_buttons_alerts.scss
mo-nathan: Thanks!  This code is larger straight from ChatGPT.  The suggestions you made make sense to me.  I'll have see if I can get it to work.

app/views/controllers/application/app/_banners.html.erb
mo-nathan: Interesting.  Why do you use underscores for class names?

mo-nathan:
@nimmolo in order to get your changes to the Stimulus controller to work (in particular `this.containerTarget...`), I had to add "container" to the targets.

I also cleaned up some of the SCSS change I had made, but found I still needed to do the color changes for the chevron.  Apparently "text-green" doesn't work for this.

mo-nathan:
> Thanks for doing this. Very helpful.
> 
> * I tried it locally and it works smoothly. I did not review the code.
> * I wouldn't mind if the banner re-appeared after a user logged out then logged in.
> * Can `app/models/language_exporter.rb#update_export_file` be covered with reasonable effort?

I don't see a reason to implement the banner reappear after a user logs out and then logs back in.  If they want to see the banner there's a chevron for doing that now.

It's not really relevant to this PR, but I added a trivial test for update_export_file.  If there is other stuff that Coveralls is complaining about I'm pretty sure it's due to lines getting deleted.

PR #2552, opened 2024-12-01
Fungarium record MyCoPortal link

### LINKED ISSUE #2471: Fungarium specimen hotlink
Automatic hotlink both the MO Fungarium Record and Observation to the official (remote) Fungarium's specimen record (for MO Fungarium records with an accession number). 
- The MO Fungarium Record should link to a search for the remote fungarium's record for that accession number.
- Ditto for the MO Observation page.

### TODO
- [x] Obtain list of Fungaria searchable by MyCoPortal
- [x] Add Herbarium#mycoportable? method.
- [x] Hotlink on appropriate MO Fungarium Record pages
- [x] Replace MCP_COLLECTIONS hash with Herbaria table `mycoportal-db` numeric field
  - [x] add new field
  - [x] update our code to get db from table
- [x] Hotlink on appropriate Observation pages 
- [x] Herbarium show:  display `mycoportal_db`
- [x] Herbarium form: display `mycoportal_db`, instructions for finding db number
- [x] Herbarium update/save
- [x] **Bug** MO Fungarium record page is missing a link to the record. Ex: http://localhost:3000/herbarium_records/44948
- [x] In Herbarium form, help, the link to mcp should open in a new tab. Can't do this because the link is created by a xlation string
- [x] Rubocop
- [ ] PR
- [ ] Respond to suggestions
- [ ] Merge, Deploy, Close issue and switch attention to immediate post-deploy tasks #2555

>[MO] Question From pinonbistro@msn.com
2 messages
webmaster@mushroomobserver.org <webmaster@mushroomobserver.org>	Sun, Oct 6, 2024 at 4:55 PM
Reply-To: pinonbistro@msn.com
To: webmaster@mushroomobserver.org
(from User #4674 Terri Clements/Donna Fulton(pinonbistro))
I use MO as my master list for AZ specimens, collected by myself and others, that are vouchered and sequenced.  MO allows me to add both fungarium and GenBank accession numbers to MO obs created by others.  This is great but while the GenBank link is a hot link to the record on GenBank the fungarium record is not linked by a hot link to the fungarium.  The MycoPortal link is a hot link to the fungarium record but for some reason only the MO obs owner can add the MycoPortal link. It would seem to make sense to allow others to add the MP link.
>
>Thanks,
>
>Terri
webmaster mushroomobserver.org <webmaster@mushroomobserver.org>	Sun, Oct 6, 2024 at 7:27 PM
To: pinonbistro@msn.com
Bcc: "webmaster mushroomobserver.org" <webmaster@mushroomobserver.org>
Terri:
Good point. Thanks for the suggestion! 
I agree that more people should be able to create a hot link to the fungarium.
I'll create a ticket for this. I won't be able to look at it further for at least a week.
(The Northeast Mycological Foray, which is happening next weekend, is using MO Field Slips and is recording foray collections in MO; the MO developers are focused on making that happen with minimum glitches.)
Best,
Joe

JoeCohen:
Links to the fungarium's collection record on MyCoPortal.
- Delivers #2471, which is inspired by a requested from MO User pinonbistro;
- Adds a `mycoportal_db` integer column to the Herbarium table;
- Allows populating that column via the Herbarium form; 
- For MO Fungarium Records, for MO Fungaria with `mycoportal_db`, links to the fungarium's collection record (in MCP), **based on the Accession Number**:
  - on the MO Fungarium Record page; and
  - on the MO Observation page.

### Suggested Manual Test
- Enter Admin Mode;
- Edit an MO institutional Herbarium by filling in the `MyCoPortal Collection ID`;
Important:
  - I suggest the [Robert L. Gilbertson Mycological Herbarium](http://localhost:3000/herbaria/407) or [University of Tennessee ](http://localhost:3000/herbaria/66/edit) because many of their MO Herbarium Records have an accurate Accession Number.
  - To find the Collection ID, click the help icon (a question mark) to the right of `MyCoPortal Collection ID`
- Exit Admin Mode;
- Click "This fungarium contains _nnnn_  fungarium record(s).";
- Click on an Accession Number;
- Click the link which takes the form "_herbarium abbrev_ Collection Record".
Expected result:  Opens a new tab with the MyCoPortal collection record corresponding to that Accession Number.
(in many cases the result is "Your query did not return any results." because the MO Accession Number is inaccurate.)

### Notes
After deployment
- I plan to immediately fill in `MyCoPortal Collection ID` for some MO Fungaria;
- I plan to add a Article soon after deployment.
- I may directly contact some users who have many MO Fungarium Records to get feedback.

### DIFF
--- a/app/models/herbarium.rb
+++ b/app/models/herbarium.rb
@@ -317,7 +317,7 @@ class Herbarium < AbstractModel
   def mcp_url(accession)
     base_url = "https://www.mycoportal.org/portal/collections/list.php"
     search_params =
-      { catnum: strip_leading_code(accession), db: MCP_COLLECTIONS[code],
+      { catnum: strip_leading_code(accession), db: mycoportal_db,
         includeothercatnum: 1 }
 
     "#{base_url}?#{search_params.to_query}"
--- a/app/models/herbarium.rb
+++ b/app/models/herbarium.rb
@@ -32,6 +32,7 @@
 #  curator?(user)::         Check if a User is a curator.
 #  add_curator(user)::      Add User as a curator unless already is one.
 #  delete_curator(user)::   Remove User from curators.
+#  mcp_searchable?          Is this herbarium searchable via MyCoPortal?
 #  sort_name::              Stripped-down version of name for sorting.
 #  merge(other_herbarium):: merge other_herbarium into this one
 #
@@ -306,5 +307,9 @@ class Herbarium < AbstractModel
 
   def self.find_by_name_with_wildcards(str)
     find_using_wildcards("name", str)
+  end
+
+  def mcp_searchable?
+    MCP_COLLECTIONS.include?(code)
   end
 end
--- a/app/models/herbarium.rb
+++ b/app/models/herbarium.rb
@@ -32,7 +32,8 @@
 #  curator?(user)::         Check if a User is a curator.
 #  add_curator(user)::      Add User as a curator unless already is one.
 #  delete_curator(user)::   Remove User from curators.
-#  mcp_searchable?          Is this herbarium searchable via MyCoPortal?
+#  web_searchable?::        Are its digital records searchable via the internet?
+#  mcp_searchable?          Are its digital records searchable via MyCoPortal?
 #  sort_name::              Stripped-down version of name for sorting.
 #  merge(other_herbarium):: merge other_herbarium into this one
 #
@@ -308,6 +309,10 @@ class Herbarium < AbstractModel
 
   def self.find_by_name_with_wildcards(str)
     find_using_wildcards("name", str)
+  end
+
+  def web_searchable?
+    mcp_searchable? || mycoportal_db.present?
   end
 
   def mcp_searchable?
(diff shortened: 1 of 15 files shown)


app/models/herbarium.rb
```diff
@@ -55,6 +58,148 @@ class Herbarium < AbstractModel
   # Used by create/edit form.
   attr_accessor :place_name, :personal, :personal_user_name
 
+  # Herbaria whose collections are searchable via MyCoPortal.
+  # https://www.mycoportal.org/portal/collections/index.php
+  # rubocop:disable Layout/LineLength
+  MCP_COLLECTIONS = {
```
mo-nathan, line 64: Seems like this should be a Set rather than a Hash since you aren't using the value for anything.  I also wonder if this shouldn't be in the database somewhere since I expect this list to change over time and as written this would require a code change to add or remove entries.

mo-nathan (approved):
Overall seems reasonable.  I do think it would be better to put the MCP_COLLECTIONS data into the database rather than a giant hash in herbarium.rb, but that could be for further down the road.

app/models/herbarium.rb
JoeCohen: Thanks for the review and comment!
Good point about putting that hash into the db. I agree it's the way to go. (I had thought about that before.)
But I want to do it as a separate PR. 
It might be possible to populate/update it (semi?) automatically with the collection numbers, depending on what I learn from Scott Bates. (See emails of today.) I need to think more about all this.

PR #2554, opened 2024-12-01
Paginating FieldSlip#index

mo-nathan:
Go to the /fieldslips page and play with the new pagination.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -159,6 +159,9 @@ gem("prawn-manual_builder")
 # https://stackoverflow.com/a/23125762/3357635
 gem("puma")
 
+# Popular pagination gem
+gem("kaminari")
+
 ########## Development, Testing, and Analysis ##################################
 group :test, :development do
   # https://github.com/ruby/debug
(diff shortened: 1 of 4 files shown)


mo-nathan:
I'm debating whether to just submit this as is or if I should include allowing FiledSlips to be filtered by project and put a link to such URLs on the Project#show page.

nimmolo:
Are you sure you can't just use our MO pagination? 
We don't need any gems to paginate an index.

mo-nathan:
In the last MO Tech Meeting, @pellaea recommended finding whatever the standard way is to do pagination in Rails these days and do that.  Otherwise we need to add all the plumbing to Query to get that mechanism to work.  It ended up being less than 10 lines of code and it was only that much because I prefer having pagination on both the top and the bottom of index pages.  The gem I added appears to be well known, widespread, and very actively maintained.  It can do a lot more than I have it doing, but even what it's doing seems great to me.  Happy to hear other arguments, but this makes a lot more sense to me than delving into Query for this.

PR #2561, opened 2024-12-05
Fix Observation Download

### LINKED ISSUE #2553: Fix Download Observations
To replicate:
- Observation pattern search for a reasonable number of Obss. Ex: https://mushroomobserver.org/observations?pattern=user%3A4468+date%3A2024-10 (@JoeCohen's Obss for 2024-10)
- Click `Download Observations`
Expected result: "Downloads" displayed Obss
Actual result: Flashes "You are trying to download too many Observations. Please see [CSV Files](https://github.com/MushroomObserver/mushroom-observer/blob/main/README_API.md#csv-files) for an alternative." because it's trying to download all Obss, rather than just the displayed Obss.

JoeCohen:
- Allows Observation Download if the number of Observatios downloaded is reasonable, or if in admin_mode.
- Delivers #2553

### DIFF
--- a/app/controllers/observations/downloads_controller.rb
+++ b/app/controllers/observations/downloads_controller.rb
@@ -4,8 +4,6 @@ module Observations
   class DownloadsController < ApplicationController
     before_action :login_required
 
-    MAX_DOWNLOADS = 120
-
     def new
       @query = find_or_create_query(:Observation, by: params[:by])
       return too_many_results if too_many_results?
@@ -41,7 +39,7 @@ module Observations
     end
 
     def too_many_results?
-      !in_admin_mode? && @query.num_results > MAX_DOWNLOADS
+      !in_admin_mode? && @query.num_results > MO.max_downloads
     end
 
     def download_observations_switch
--- a/app/controllers/observations/downloads_controller.rb
+++ b/app/controllers/observations/downloads_controller.rb
@@ -4,6 +4,8 @@ module Observations
   class DownloadsController < ApplicationController
     before_action :login_required
 
+    MAX_DOWNLOADS = 120
+
     def new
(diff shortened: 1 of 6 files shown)


JoeCohen:
Questions:
- What's the right limit on # of Observations to download?
- @pellaea suggests stubbing `MAX_DOWNLOADS` instead of creating that many Obss in the tests.  See #2553
What's best?

app/controllers/observations/downloads_controller.rb
```diff
@@ -4,6 +4,8 @@ module Observations
   class DownloadsController < ApplicationController
     before_action :login_required
 
+    MAX_DOWNLOADS = 120
```
pellaea, line 7: 120 seems way too low.  Surely there's some compromise between half a million and 120?  It was originally designed to be able to render a million in a few seconds, so surely 1000 wouldn't be too many?  Has anyone actually tested the performance of this method?

JoeCohen:
Good question.
1. I'm not aware of performance tests.
BUT
2. Downloading all Observations was bringing the site to its knees.

JoeCohen:
I'll bump the number. Let me give it some thought. I'll probably do it
tomorrow.

pellaea:
Fair enough. Maybe start with 100, push it out, test it a few times and
look at the log to see how long 100 takes.  Then bump it to 1000.  Rinse
and repeat until it pushes against what you think is acceptable.

Long term, it probably makes sense to convert it into an async job, right?
Probably pretty easy to do, at this point, too.  We've got a number of
examples of async jobs -- field slips, inat import, original image
retrieval.  Probably just need some sort of job tracker table in the
database and move a few files and tests around.

JoeCohen:
Good suggestions.
I agree that a job is the long-range solution.

PR #2564, opened 2024-12-07
Break up the `ApplicationController` into modules

nimmolo:
I think this will help make `ApplicationController` less daunting to study and refactor.

It's been a bit murky to me, keeping straight what is done by `Query` and what is done by `ApplicationController` *with* the Query.


JoeCohen:
👍 for making `ApplicationController` less of a god object (or at least more polytheistic).
Based on the module names, these seem like good splits.
Could some of the modules be moved to other controllers (maybe in another PR), e.g., `name_validation`.
I will postpone further review until it's passing tests.

app/controllers/application_controller/flash_notices.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  # Get a copy of the errors.  Return as String.
+  def flash_get_notices
+    # Maybe there is a cleaner way to do this.  session[:notice] should
+    # already be html_safe, but the substring marks it as unsafe. Maybe there
+    # is a way to test if it's html_safe before, and if so, then it should be
+    # okay to remove the first character without making it html_unsafe??
+    # rubocop:disable Rails/OutputSafety
+    session[:notice].to_s[1..].html_safe
+    # rubocop:enable Rails/OutputSafety
```
JoeCohen, line 59:
```suggestion
    session[:notice].to_s[1..].html_safe # rubocop:disable Rails/OutputSafety
```

app/controllers/application_controller/flash_notices.rb
```diff
@@ -0,0 +1,135 @@
...
+      :flash_notices?, :flash_get_notices, :flash_notice_level,
+      :flash_clear, :flash_notice, :flash_warning, :flash_error
+    )
+  end
+
+  ##############################################################################
+  #
+  #  :section: Error handling
+  #
+  #  NOTE: MO doesn't use built-in Rails `flash`, i.e. session[:flash].
```
JoeCohen, line 16: @mo-nathan, @pellaea: I didn't realize that we rolled our own flash. 
I'm curious. Is there an short explanation?

app/controllers/application_controller/indexes.rb
```diff
@@ -0,0 +1,380 @@
...
+  # rubocop:disable Rails/FindEach
+  def apply_content_filters(query)
+    filters = users_content_filters || {}
+    @any_content_filters_applied = false
+    ContentFilter.all.each do |fltr|
+      apply_one_content_filter(fltr, query, filters[fltr.sym])
+    end
+  end
+  # rubocop:enable Rails/FindEach
+
```
JoeCohen, line 106: 1. @mo-nathan had long ago suggested leaving a comment  explaining why a cop is disabled.
2. No need to disable the cop for the whole method.
3. If just one line is affected, I like to disable cops with a an in-line comment, which results in two fewer lines.
```suggestion
  def apply_content_filters(query)
    filters = users_content_filters || {}
    @any_content_filters_applied = false
    # disable cop because ContentFilter is not an ActiveRecord model
    ContentFilter.all.each do |fltr| # rubocop:disable Rails/FindEach
      apply_one_content_filter(fltr, query, filters[fltr.sym])
    end
  end
```

app/controllers/application_controller/queries.rb
```diff
@@ -0,0 +1,437 @@
...
+    if path.include?("?") # Does path already have a query string?
+      "#{path}&q=#{query_param}" # add query_param to existing query string
+    else
+      "#{path}?q=#{query_param}" # create a query string comprising query_param
+    end
+  end
+
+  # Allows us to add query to a path helper:
+  #   object_path(@object, q: get_query_param)
+  def get_query_param(query = nil)
```
JoeCohen, line 106: Outside the scope of this PR, 
but how about (globally) renaming this to something like
`query_id_string` or `query_id_alphabetized` or `q_param`?
("param" is overloaded, "get" is redundant)

app/controllers/application_controller/queries.rb
```diff
@@ -0,0 +1,437 @@
...
+  end
+
+  def tweak_up(value, amount, max)
+    [max, value.to_f + amount].min
+  end
+
+  def tweak_down(value, amount, min)
+    [min, value.to_f - amount].max
+  end
+
```
JoeCohen, line 307: Feels like these belong somewhere else, maybe in Query or their own class?

app/controllers/application_controller/queries.rb
```diff
@@ -0,0 +1,437 @@
+# frozen_string_literal: true
+
+# see application_controller.rb
+# rubocop:disable Metrics/ModuleLength
+module ApplicationController::Queries
```
JoeCohen, line 5:
```suggestion
module ApplicationController::Queries # rubocop:disable Metrics/ModuleLength
```
So you don't have to disable it at the end. 2 fewer physical lines in the file.

app/controllers/application_controller/queries.rb
```diff
@@ -0,0 +1,437 @@
...
+                                               type: object.type_tag))
+    { object: object, id: id, query: query }
+  end
+
+  def no_more_objects_in_given_direction(object, id, query)
+    flash_error(:runtime_no_more_search_objects.t(type: object.type_tag))
+    { object: object, id: id, query: query }
+  end
+end
+# rubocop:enable Metrics/ModuleLength
```
JoeCohen, line 437: See comment at top of file.

JoeCohen (approved):
Nicely done.
My comments are all stylistic; take them or leave them.

app/controllers/application_controller/flash_notices.rb
nimmolo: @JoeCohen I was surprised too, that's my note I believe, from a few years ago. 

After looking into this vs [the documentation](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html), my impression was that the main difference between this and the "Rails way" is the naming of the session attribute itself, otherwise I remember feeling that as far as the Rails flash docs provide guidance, it's mostly roll your own anyway, and seems to work pretty much the same.  

But you may notice differences I missed.

app/controllers/application_controller/queries.rb
nimmolo: Agree it could be elsewhere. However, I think we will not need to tweak bounding boxes here when we switch to AR rather than Query. Our scopes include the tweak, IIRC, so it may be going away soon, along with a lot of this module.

app/controllers/application_controller/queries.rb
nimmolo: I like the inline cop disabling, but Rubocop doesn't seem to allow inline disabling class definitions.

(Also - note that because these are "magic" comments, they don't count towards the class line length. Comments and blank lines don't count.)

PR #2570, opened 2024-12-08
Improve API formatting

mo-nathan:
Specifically only add paragraph tags if the value has newlines.


mo-nathan:
Should `tl_for_api` replace `tpl_nodiv` everywhere?

JoeCohen (approved):

PR #2572, opened 2024-12-09
Revise MCP auto-linking

### LINKED ISSUE #2571: Revise mcp_searchable
Revise `Herbarium#mcp_searchable?` to rely on a csv file (below) supplied by MyCoPortal rather than on the inline `MCP_COLLECTIONS` hash which I created.

### Tasks
- [x] Find a place for the file.
- [x] Get MCP db id from the file instead of a Herbarium column.
- [x] Revise `mcp_searchable?` logic
- [x] Revise link url logic
- [x] Remove `MCP_COLLECTIONS`
- [x] Review herbarium show
- [x] Review herbarium form
- [x] Review show observation
- [x] Review FungariumRecord show
- [x] Review FungariumRecord form
- [x] adjust mcp-collid logic to account for official abbrv != `coll acryonym`
Exs: TENN-F, NMC-FUNGI, UBC-Obs
- [x] Remove the Herbarium#mycoportal_db column
- [x] Manual test 
  - [x] existing links work
  - [x] new links
- [x] Rubocop
- [x] PR #2572
- [ ] Respond to PR comments

[mycoportal-collections.csv](https://github.com/user-attachments/files/18053821/mycoportal-collections.csv)

JoeCohen:
- Replaces hand-coded hash with file received from MCP as source of truth about MCP collection existence & ids
- Revises logic for creating links from Observations/Fungarium Records to MyCoPortal 
- Removes now-uneeded mcp_db column


JoeCohen:
@mo-nathan You had suggested moving the `MCP_COLLECTIONS` out of `herbaria.rb` and into the db.
This PR removes it from `herbaria.rb`, and replaces it with a file received from Andy Miller.
I think that periodically updating that file would be a relatively easy way to keep things up to date.
Is there a better approach?

public/mycoportal-collections.csv
```diff
@@ -0,0 +1,167 @@
+collid,coll acryonym,coll name
```
JoeCohen, line 1: @mo-nathan 
1. You suggested moving the MCP_COLLECTIONS out of herbaria.rb and into the db.
This PR removes it from herbaria.rb, and replaces it this file received from Andy Miller.
I think that periodically updating that file would be a relatively easy way to keep things up to date.
Is there a better approach?
2. if a periodically update file is the right approach, where should the file live?

app/models/herbarium.rb
```diff
@@ -58,147 +64,16 @@ class Herbarium < AbstractModel
...
+  def self.mcp_collections
+    @mcp_collections ||=
+      begin
+        csv_path = Rails.public_path.join("mycoportal-collections.csv")
+        CSV.foreach(csv_path,
+                    headers: true).each_with_object([]) do |row, collections|
+          collections << row.to_h
+        end
+      end
+  end
```
JoeCohen, line 76: @mo-nathan, @nimmolo:
1. Does this method make sense? (I'm still uncomfortable with some aspects of Ruby.)
2. Is there a more performant way to deal with `mcp_collections`?

JoeCohen:
@mo-nathan:
You suggested moving the MCP_COLLECTIONS out of herbaria.rb and into the db.
This PR removes it from herbaria.rb, and replaces with `public/mycoportal_collections.json`, which I downloaded using the MyCoPortal API. See below. I think that periodically updating that file would be a relatively easy way to keep things up to date.


### Questions:
- Is there a better approach?
- If a periodically updated file is the right approach, where should the file live?
- Should the file be trimmed? The file is 332KB. Each herbarium has about 31 fields. I'm currently using only two of them. But I can see using more, for instance `fulldesciription` on the herbarium show page.

Curl:
```
curl -X 'GET' \
  'https://www.mycoportal.org/portal/api/v2/collection?collectionType=preservedSpecimens&limit=500&offset=0' \
  -H 'accept: application/json'
```
Request URL
```
https://www.mycoportal.org/portal/api/v2/collection?collectionType=preservedSpecimens&limit=500&offset=0
```

app/models/herbarium.rb
```diff
@@ -58,147 +63,26 @@ class Herbarium < AbstractModel
...
+            CollID: collection["CollID"]
+          }
+        end
+      end
+  end
+
+  # wrap the class method
+  def mcp_collections
+    self.class.mcp_collections
+  end
```
JoeCohen, line 85: @mo-nathan, @nimmolo, @pellaea 
@mo-nathan, @nimmolo:

### Questions:

- Do these method make sense? (I'm still uncomfortable with some (pretty basic) aspects of Ruby.)
- Is there a more performant way to deal with `mcp_collections`?

PR #2573, opened 2024-12-10
Standardize paginated index methods

nimmolo:
The idea of this PR is to standardize the methods used to build index queries in different controllers, so that we can build a set of alternate methods that use AR instead of Query. Then, we can switch over one controller at a time.

Things to note:
- The `index` method is again defined in each controller, instead of implicitly in the `ApplicationController::Index`
  - It simply delegates to `build_index_from_query` (the old `index` method defined in `ApplicationController`)
  - When we figure out a `build_index_with_active_record`, the idea is to have each index instead delegate to that, when it's ready
- `index_subaction_param_keys` and `index_subaction_dispatch_table` are changed from ivars to methods
- Many repetitive declarations are DRY'd up to the `ApplicationController::Index`

I'm not expecting a code review, these changes are vast. What would be good is a spot check of any indexes you care about, just to be sure they're behaving.

Note that param names have been standardized. I believe the only URL that's changed that would be linked from the outside is `herbaria_path(flavor: :all)`, which is now just `herbaria_path`. Fortunately the index action will simply ignore the `flavor` param if anyone sends it, so old links will still work. `herbaria_path(flavor: :nonpersonal)` has changed to `herbaria_path(nonpersonal: true)`, so if there are external links to that, maybe we should add a redirect in routes.


app/controllers/application_controller/indexes.rb
```diff
@@ -1,48 +1,170 @@
...
+      # If we have the param but get a blank query we should bail.
+      if query.blank?
+        bail = true
+        break
+      end
+      # If we have a query, display it.
+      return filtered_index(query, display_opts)
     end
+    return if bail
+
```
JoeCohen, line 32:
```suggestion
    index_active_params.each do |subaction|
      next if params[subaction].blank?

      query, display_opts = send(index_param_method_or_default(subaction))
      # Some actions may redirect instead of returning a query.
      # If we have the param but get a blank query we should bail.
      return if query.blank?

      # If we have a query, display it.
      return filtered_index(query, display_opts)
    end
```

JoeCohen (changes requested):
I had no blowups or unexpected results after spending some time clicking through various indices.
There's too much code for me to review right now. (But I did have one suggestion.)

app/controllers/application_controller/indexes.rb
nimmolo: Ah - thanks for the suggestion. That's what I had originally, but Rubocop `Lint/NonLocalExitFromIterator` didn't like it. It said "Non-local exit from iterator, without return value. `next`, `break`, `Array#find`, `Array#any?`, etc. is preferred."

I tried switching it to `break`, but then the tests fail with "AbstractController::DoubleRenderError". That's when i tried the `bail` thing. But I'm just now tweaking your version to explicitly return `nil` and that's working, and not offending Rubocop. 

Committing that now.  Thank you! The `bail` was bugging me.

JoeCohen:
Thanks for the feedback. I would have tried it myself; then I would have
gotten the same offense and would have found the correction.

mo-nathan (approved):
Played around with various index pages and didn't spot anything odd.  Try to read through at least some of it, but got lost pretty quickly.  Thanks for all the work on this!

PR #2591, opened 2024-12-20
Project show page improvements

### LINKED ISSUE #2513: Add ability to filter project observations by location
As a user viewing a project, I should be able to filter the names based on relevant locations so I can share such lists with land owners/managers.

mo-nathan:
Switch to more of a tab style for the different Project pages.
Add a Field Slips tab.
Add a Summary tab.


app/helpers/checklist_helper.rb
```diff
@@ -12,14 +12,16 @@ def checklist_name_link(taxon:, user:, project:, list:, data:)
...
                "user:#{user.id}"
              elsif project
                "project:#{project.id}"
              elsif list
                "list:#{list.id}"
              end
+    prefix += " location:#{location.id}" if location
     return name_path(name_id) unless prefix
 
     observations_path(pattern: "#{prefix} name:#{name_id} " \
```
nimmolo, line 27: You may want to test to be sure this works. 

Presently, as far as I know, the controllers will not yet correctly process multiple filtering params. (That's what i'm working towards currently; it's definitely the way it should work IMO.)

The way you would currently create a working link that filters for, say, `project`, `location` and `name` using Query is maybe easier to grasp by the examples in the `names_helper.rb`, where it's broken out in separate methods. I'll put it all together here, just as an example. The following produces the "alphabetized" q param, but it's the only way that works for now... i believe. (Do check though!)
```ruby
 def checklist_name_link_path(name_id, params)
    user, project, location, list = params
    args = params.compact_blank
    query = Query.lookup(:Observation, :all, **args)
    add_query_param(observations_path, query)
end

mo-nathan: The checklist controller doesn't use Query, but does handles it directly in the controller and the Checklist class.  See line 23 in checklists_controller.rb and line 51 in app/classes/checklist.rb.

mo-nathan: There is also a test for it: ChecklistsControllerTest#test_checklist_for_project_location.

nimmolo: Oh good. OK

nimmolo: Mmmm... still not sure. I'm just talking about that link out to the observations index... 

I don't think your link will produce the expected index of observations filtered by (e.g.) NEMF, Ann Arbor and Suillus tomentosus (all three).

mo-nathan: Ah, in that case it is using the pattern search syntax which seems to work.  Currently those links go to a link like this:

https://mushroomobserver.org/observations?pattern=project%3A352+name%3A3801+include_synonyms%3Afalse+include_subtaxa%3Afalse

which is wrong since it shows observations for multiple locations.  The change I make now uses likes like this:

http://localhost:3000/observations?pattern=project%3A352+location%3A29270+name%3A3801+include_synonyms%3Afalse+include_subtaxa%3Afalse

which is definitely better (3 obsies that are only for that location vs. 9 for all of the locations in the project).

nimmolo: Yes, ok. Pattern search syntax is currently the other way, I forgot about that.

But just know that I plan to pull the plug on that syntax, and have PatternSearch use plain params as above. I think that's the goal we all want to get to!

nimmolo: I totally missed the word `pattern` in your `observations_path`!!!

nimmolo: Another consideration i am remembering here — it looks like i'm definitely going to be able to sweep through the code and replace all those links that use Query currently and generate `?q=23k4ju4` links — these are numerous on obs, name and location pages, to turn them into `/observations?project=342...` etc.

If you use the currently common "filtered index link maker" in the suggestion above, it'll just make it easier for me to `grep` around to find all the examples of `Query.lookup_and_save` or `Query.lookup` in the app. Not really a big deal in this case, because I will probably remember this outlier, but...

That's really what i'm thinking about when i'm being a stickler for using the same methods everywhere — saving myself work. I've many times swept through the app to standardize method calls, and then had to track down the quirky cases where we didn't use the same calls as the bulk of cases, in order to make overall changes to methods. This change to using regular index links with flat params will be a big one, and i think will make everyone happier!

Issue #2594, opened 2024-12-21
Lat/long behaving badly in Create Observation page


mo-nathan:
1. Go to Create Observation
2. Clear out any location string
3. Enter the following as lat/long: 41.63, -70.62 (which is in North Falmouth)
4. Note that the map updates to somewhere south of Nantucket (which is wrong)
5. Note that if you move the pin, it updates the lat/long to something around 41.1, -70.1 (which is south of Nantucket)
6. If you move the pin to North Falmouth, it updates the lat/long to something close to 41.63/-70.62.
7. Now select “North Falmouth” from the “New Locality” drop down.
8. If you now try to edit the latitude to 41.63, it will again jump to somewhere south of Nantucket.
9. For me, it updates the lat/long to 41.1008/-70.1083.

mo-nathan:
Looks like this got fixed.

PR #2597, opened 2024-12-22
Stop assuming lat and lng are in degree, minute, second

mo-nathan:
Addresses #2594.

Old bad behavior:
1. Go to Create Observation
2. Clear out any location string
3. Enter the following as lat/long: 41.63, -70.62 (which is in North Falmouth)
4. Note that the map updates to somewhere south of Nantucket (which is wrong) [Fixed on branch so the rest doesn't happen]
5. Note that if you move the pin, it updates the lat/long to something around 41.1, -70.1 (which is south of Nantucket)
6. If you move the pin to North Falmouth, it updates the lat/long to something close to 41.63/-70.62.
7. Now select “North Falmouth” from the “New Locality” drop down.
8. If you now try to edit the latitude to 41.63, it will again jump to somewhere south of Nantucket.
9. For me, it updates the lat/long to 41.1008/-70.1083.

### DIFF
--- a/app/javascript/controllers/geocode_controller.js
+++ b/app/javascript/controllers/geocode_controller.js
@@ -320,6 +320,11 @@ export default class extends Controller {
       this.getElevationTarget.disabled = true
   }
 
+  // Using a regular expression
+  isValidDecimal(str) {
+    return /^-?\d*\.?\d+$/.test(str);
+  }
+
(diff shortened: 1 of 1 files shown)


nimmolo (approved):
If this works IRL it certainly looks right as code. Thanks for tackling it!

PR #2598, opened 2024-12-24
Upgrade to Ruby 3.3.2

nimmolo:
Ruby is currently at 3.3.6, with 3.4 on the way. I figure minor version upgrades are the way to go, one by one.

So i'm planning to do this, then 3.3.4, then 3.3.6 in as short a time as possible.

PR #2601, opened 2024-12-24
Rails 7.2.2.1

mo-nathan:
Latest version of Rails 7.2.

@nimmolo Noticed that your PR includes an incomplete upgrade from 7.1.3 to 7.1.5.1.  I created the branch rails-7-1-5-1 which includes your changes, my changes (bundle update), and does all the updates for 7.1.5.1.

PR #2600, opened 2024-12-24
Bundle update ruby 3.3.2

mo-nathan:

PR #2602, opened 2024-12-25
Rails 7.1.5.1 upgrade

mo-nathan:


JoeCohen (commented):
I'm good with this, but:
- am confused by the various pending PRs which upgrade Rails & Ruby.
- wonder about 7.1.4

mo-nathan:
There is no reason to go to 7.1.4.  The documented upgrade process involves going directly to the latest patch version so going directly from 7.1.3 to 7.1.5.1 is reasonable.  See https://guides.rubyonrails.org/upgrading_ruby_on_rails.html.  It states:

"The process should go as follows:

Write tests and make sure they pass.
Move to the latest patch version after your current version.
Fix tests and deprecated features.
Move to the latest patch version of the next minor version."

1) Nimmo's PR, #2598, takes us Ruby 3.3.2.
2) My PR, #2600, builds on that by doing a `bundle update`.
3) This PR builds on that and upgrades Rails to 7.1.5.1 ("latest patch version after your current version").
4) My other PR, #2601, goes to 7.2.2.1 (per the last line in the above process).  However, this generates a bunch of deprecation warnings which we may want to fix before making that jump.  However, that's not technically required to go to 7.2.2.1, but it will be if we try to go to Rails 8.0 anytime soon.  On the other hand, if we don't fix them, then it creates a bunch of noise when running tests or starting the server.

At this point, I think we should probably merge this PR and deploy it to production.  Upgrading Ruby is a bit of pain, but we need to do that not matter way.  The bundle update and 7.1.5.1 upgrades are just changes to the Gemfile.  Then we just close #2598 and #2600.

nimmolo:
@mo-nathan it doesn't seem to me this PR is the way to go, because 3.3.2 is not our target Ruby version. 3.3.6 is where I'm trying to get to, before updating everything and letting things get settled. 

I feel like doing bundle updates each time (3.3.2, 3.3.4, 3.3.6) would be adding unnecessary variables to these PRs. 

Do you believe otherwise? I thought minimal was the way to go with stuff like this.

PR #2606, opened 2024-12-27
Remove references to mysql_native_password

mo-nathan:


README_MACOSX_NOTES.md
```diff
@@ -132,7 +132,7 @@ FLUSH PRIVILEGES;
 exit;
 # Setting it to null above seems necessary. Then:
 mysql -u root
-ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';
+ALTER USER 'root'@'localhost' IDENTIFIED BY 'root';
```
JoeCohen, line 135: I suggest removing the entire section starting with 
`@JoeCohen had problems setting the MySQL root password as follows:`
through
`7. Restart MySQL
brew services start mysql`

JoeCohen (approved):

JoeCohen:
Thanks for doing this.

README_MACOSX_NOTES.md
```diff
@@ -243,6 +207,10 @@ rake db:drop
 mysql -u root -p < db/initialize.sql
 ```
 
+NOTE: 2024-12-27 @JoeCohen: For me (but not others) `rake db:drop` threw an error
+because `config/database.yml` was missing.
+Please contact us if that is the case.
```
JoeCohen, line 212: @mo-nathan:
I didn't add more about this because `config/database.yml` is outside version control for confidentiality reasons. 
I also didn't add what we later did in mysql because I don't know if would publish confidential info.

mo-nathan: @JoeCohen We should add a `config/database.yml-template` file to address this, but that can happen in another PR.

mo-nathan (commented):
Looks good to me!

PR #2608, opened 2024-12-28
Backfill original_name

JoeCohen:
This is a script to ensure that images imported from iNat photos have a unique `original_name`.
- Adds the iNat `photo_id` and `uuid` to the `original_name` of Images which were imported from iNat photos prior to Fri Dec 13 15:03:37 2024 -0800. (Later imports already have the desired `original_name`.)
- The point is enabling efficient syncing/updating of Observations.
- This is a run-once script which can be deleted after it's run.

NOTE: Some `original_names` cannot be fixed. Either:
- The photo has been deleted from iNat. 
- The entire iNat observation has been deleted.

### DIFF
--- a/scripts/backfill_inat_img_orig_fname.rb
+++ b/scripts/backfill_inat_img_orig_fname.rb
@@ -14,7 +14,7 @@ require_relative "../config/environment"
 class BackfillInatImgOrigFname
   CUTOFF = Time.new(2024, 12, 13, 15, 3, 37, "-08:00")
 
-  def self.run
+  def self.run # rubocop:disable Metrics/AbcSize
     obss = Observation.where(source: "mo_inat_import").
            where(Observation[:created_at].lt(CUTOFF))
 
--- a/scripts/backfill_inat_img_orig_fname.rb
+++ b/scripts/backfill_inat_img_orig_fname.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require_relative "../config/environment"
+
+# Ensure that images imported from iNat have a unique original_name.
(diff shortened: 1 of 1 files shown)


JoeCohen:
This PR isn't a thing of beauty, but it works -- I manually tested it locally -- and will be used only once.
I therefore prefer to avoid investing time time to make it pretty.
It takes about 10 minutes to run, primarily because it hits the iNat API no more than once per second, per iNat's request to "eep it to 60 requests per minute". (There are about 600 relevant Observations, and I'm hitting the iNat API ~once per Observation.)

### Question
Should this be turned into a Job?

JoeCohen:
#### QUESTION
Is it worth the effort to turn this into a background Job?

This isn't a thing of beauty, but 
- It works (I tested it manually in the dev environment).
- It will be used only once.

I therefore prefer to avoid investing time to make it pretty.

It takes about 10 minutes to run, primarily because iNat asks: "keep it to 60 requests per minute", I'm hitting that API once per O)bservation, and there about 600 relevant Obss.

nimmolo:
Sounds like no need to make it a job. Good to go, IMO.

JoeCohen:
Thanks for looking at it and the comments.

mo-nathan:
This is a one time thing, right?  If so, then I wouldn't merge it into main, but rather copy it over to the server and run it.  You can leave it on this branch as a record of what was done, but I don't think there is value in merging it into main.

JoeCohen:
Thanks. Will do.

JoeCohen:
Closing this per @mo-nathan's suggestion.

PR #2618, opened 2024-12-31
Get some defaults/deprecations ready for Rails 7.2

nimmolo:
Getting this right will make migration to 7.2 easier. 
Deals with most deprecation warnings outside of the `enum`s.
___
Code changes outside of configs:
- `YAML` is no longer the default `coder` for serialized columns, it's recommended to use explicit declarations. The original Rails plan as of 7.0 was to make `JSON` the default, because most apps in the wild probably want that, so we would have had to declare ours as `YAML` in that case. However, Rails dropped that plan and now the recommended default value is `nil` with `coder` declared on each attribute. Seems easier to go along with this convention and just declare `coder: YAML` on our few serialized columns. (Ours all use YAML.)
- `form_with(model: @instance, url: fallback)` will currently work even if `@instance` is `nil`, but it will stop working in 8.0 and is deprecated in 7.2. Modifies our Naming Votes helper method to check if the vote instance exists.


JoeCohen:
Would it make sense to split this into 2 prs?

nimmolo:
@JoeCohen I finally figured out the issue with the form.

PR #2621, opened 2025-01-01
Restore `QueryTest` ordering checks

nimmolo:
Restores the ordering checks in QueryTest, and adds much more complete analogies between AR and Query.

Adds new `default_scope` ordering for those models that currently have filterable indexes. 

@JoeCohen - Note this introduces a significant potential gotcha — all AR scopes of these indexable models will be magically "pre"-ordered according to the index's default sort order going forward. This default order will not always be what you want - if your scope needs a different order, you must call `reorder` instead of `order`, or `unscoped`.  You can override the scope, but don't forget this!

Also, a few of our maintenance queries like name caches and classification updates now have to `unscope` their initial queries, particularly when joining, to get viable results. I learned this through trial and error. 
___

I dimly remember I had removed the checks for query order when converting the test expectations from SQL to ActiveRecord, because getting the ordering of the AR to correspond to all the logic in Query seemed like a ton of work. Well, it was! A full week, pretty grody.

But it turns out this test class is the best reference we have going forward for equivalent AR scopes and Query results. As I'm converting indexes to use AR scopes, I found I needed these tests to compare against, to be sure I was getting it right. Turning on ordering revealed 84 failing tests, that tested against AR scopes!

Bonus: fixes several broken index ordering options, particularly in Species Lists.

Note: the idea here is not to optimize these Queries, but rather to make it so that we have accurate tests of what they produce. The project is to use the AR replacements for these queries in the indexes.


test/models/query_test.rb
```diff
@@ -1160,43 +1161,49 @@ def test_collection_number_all
...
+              where(CollectionNumber[:name].matches("%123a%").
+                    or(CollectionNumber[:number].matches("%123a%"))).
+              sort_by(&:format_name)
+    assert_query(expects, :CollectionNumber, pattern: "123a")
   end
 
   def test_comment_all
-    expect = Comment.all.reverse
-    assert_query(expect, :Comment)
+    expects = Comment.order(Comment[:created_at].desc, Comment[:id].desc).uniq
```
JoeCohen, line 1183: I'm curious whether you hit a case where you needed two sort params.
Also about `uniq`. 
Seems like `Comment.order(id: :desc)` should be enough.

JoeCohen: Never mind.  Looks like you did this to make the `expects` more consistent across different tests.
Makes sense.

test/models/query_test.rb
```diff
@@ -1160,43 +1161,49 @@ def test_collection_number_all
...
   end
 
   def test_collection_number_pattern_search
-    expect = CollectionNumber.
-             where(CollectionNumber[:name].matches("%Singer%").
-                   or(CollectionNumber[:number].matches("%Singer%"))).
-             sort_by(&:format_name)
-    assert_query(expect, :CollectionNumber, pattern: "Singer")
+    expects = CollectionNumber.
+              where(CollectionNumber[:name].matches("%Singer%").
```
JoeCohen, line 1170: Curious about `
matches("%Singer%")` vs 
`matches(/Singer/)`

JoeCohen (approved):
Wow! a ton of work, greatly improved tests.
Feel free to ignore my questions.

test/models/query_test.rb
nimmolo: @JoeCohen I think you can use `distinct` in all test scopes _except_ polymorphic associations. 

For Comment I had to call `uniq` for any of the scopes joining to the `target`, to get it to execute the query right away, or else i got a `polymorphic join cannot be eager loaded` error.

test/models/query_test.rb
nimmolo: Good question. I think it can also just be `matches("Singer")`.

nimmolo:
@JoeCohen I'm thinking of moving these verbose ordering clauses into the models as a `default_scope` where possible. I didn't know this was a thing until I stumbled on an incomplete implementation of it in `FieldSlip`.

This is not just to make these tests less verbose, but also to make it less likely to trip me up when calling the scope instead of `create_query` from a controller action. This will guarantee each model's scopes get the default sort order that Query currently applies.

Note: you can call `resort(updated_at: foo)` on a scope chain to reset the ordering at any point. I still believe cases where we would use this are far less numerous than the cases where `default_scope` would help.

test/models/query_test.rb
nimmolo: Ha! `distinct` works here too.

The current state of the PR is kind of a rolling record of "things I tried", cleaning it up now.

test/models/query_test.rb
nimmolo: Nope, it has to be `matches("%Singer%")`. The other two won't work.

The slashes won't work in AR, and the plain quoted string apparently doesn't mean the same thing.

nimmolo:
This branch now provokes 1750 deprecation warnings per test run.

nimmolo:
Arel deprecation warnings now down to 99.

app/classes/query/modules/lookup_names.rb
```diff
@@ -86,16 +86,16 @@ def find_exact_name_matches(vals)
   def find_matching_names(name)
     parse = Name.parse_name(name)
     name2 = parse ? parse.search_name : Name.clean_incoming_string(name)
-    matches = Name.where(search_name: name2) if parse&.author.present?
-    matches = Name.where(text_name: name2) if matches.empty?
+    matches = Name.unscoped.where(search_name: name2) if parse&.author.present?
+    matches = Name.unscoped.where(text_name: name2) if matches.empty?
```
mo-nathan, line 90: Is there a functional reason for adding `.unscoped` here?  If the only motivation is performance have you actually tested the performance of the query with and without?  I'm guessing this is unnecessary and just makes the code more complicated.

app/models/project.rb
```diff
@@ -448,8 +455,8 @@ def out_of_range_observations
     elsif end_date.nil?
       observations.where(Observation[:when] < start_date)
     else
-      observations.where(Observation[:when] > end_date).
-        or(observations.where(Observation[:when] < start_date))
+      observations.reorder("").where(Observation[:when] > end_date).
```
mo-nathan, line 458: `reorder("")` strikes me as a weird thing to do here and an unneeded complication.

test/controllers/api2_controller_test.rb
```diff
@@ -141,7 +147,7 @@ def test_post_minimal_observation
     params = { api_key: api_keys(:rolfs_api_key).key, location: "Earth" }
     post(:observations, params: params)
     assert_no_api_errors
-    obs = Observation.last
+    obs = Observation.unscoped.last
```
mo-nathan, line 150: I think this is dangerous.  If this eliminates the `ORDER BY` then you are subject to whatever ordering MySQL decides to use which changes depending on the amount of data in the given database table.  Specifically you are not guaranteed to get the Observation with the highest `id`.  If that is intended, then that should be done explicitly.  I'm sure we've been fast and lose about this in our tests, but if you're going to be explicit about it, then I would be correctly explicit about it.

mo-nathan (changes requested):
I think we need a consistent policy regarding the use of scopes.  I have been burned multiple times by using unordered queries and find the idea of explicitly turning off ordering with `unscoped` and `reorder("")` to be dubious at best.  Also the usage of these tools seems inconsistent across this PR.  The only reason I could imagine turning off ordering being important would be performance considerations.  In over 30 years of working with SQL, I have only very rarely seen a situation where ordering the results has any real performance impact and when I have, the correct solution has been to add an index for that ordering, not abandoning the ordering.  An example that I could imagine being an issue for us are Names where the default sorting is now the string `sort_name`.  A bit of quick testing shows that this causes a query returning values of every name in the system to go from 0.03 seconds to 0.11 seconds.  However, the reality is that we shouldn’t be querying for all the Names very often if at all so that impact is going to be negligible.  If it is an issue, then add an index.  Just to prove the point, I added an index on `sort_name` in my load database, and it dropped that query to 0.04 seconds which is a tiny price to pay for reliably consistent ordering.

nimmolo:
Did you see my notes? I don't like adding these `unscoped` calls at all... it had absolutely nothing to do with performance. The reason i turned off ordering in most of these cases was to shut up the 1700 Arel deprecation warnings! 

In some cases though, the reason was that the query didn't return the right results, in which case explicit ordering seems like the right way to go.

mo-nathan:
I saw your note on #2633 and I don't understand them.  It would be helpful if you gave more context than "my notes".

mo-nathan:
I just removed the two example uses of `.unscoped` given in the note in #2633 and I see no deprecation warnings.

```
nathan@Nathans-MacBook-Pro mushroom-observer % rails t
Started with run options --seed 31498

WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"00:01:28

  2674/2674: [==============================================================] 100% Time: 00:02:16, Time: 00:02:16

Finished in 136.55136s
2674 tests, 29407 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Minitest to /Users/nathan/src/mushroom-observer/coverage.
Line Coverage: 93.51% (26806 / 28667)
nathan@Nathans-MacBook-Pro mushroom-observer % git grep 'matches = Name.unscoped.where'                       
app/classes/query/modules/lookup_names.rb:    matches = Name.unscoped.where(search_name: name2) if parse&.author.present?
app/classes/query/modules/lookup_names.rb:    matches = Name.unscoped.where(text_name: name2) if matches.empty?
nathan@Nathans-MacBook-Pro mushroom-observer % emacs app/classes/query/modules/lookup_names.rb
nathan@Nathans-MacBook-Pro mushroom-observer % rails t
Started with run options --seed 53976

WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"00:01:39

  2674/2674: [==============================================================] 100% Time: 00:02:13, Time: 00:02:13

Finished in 133.86660s
2674 tests, 29407 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Minitest to /Users/nathan/src/mushroom-observer/coverage.
Line Coverage: 93.51% (26806 / 28667)
nathan@Nathans-MacBook-Pro mushroom-observer % git diff
diff --git a/app/classes/query/modules/lookup_names.rb b/app/classes/query/modules/lookup_names.rb
index 27cc925c7..ce71b0f31 100644
--- a/app/classes/query/modules/lookup_names.rb
+++ b/app/classes/query/modules/lookup_names.rb
@@ -86,8 +86,8 @@ module Query::Modules::LookupNames
   def find_matching_names(name)
     parse = Name.parse_name(name)
     name2 = parse ? parse.search_name : Name.clean_incoming_string(name)
-    matches = Name.unscoped.where(search_name: name2) if parse&.author.present?
-    matches = Name.unscoped.where(text_name: name2) if matches.empty?
+    matches = Name.where(search_name: name2) if parse&.author.present?
+    matches = Name.where(text_name: name2) if matches.empty?
     matches.map { |name3| minimal_name_data(name3) }
   end
 ```

**Comment by nimmolo** [comment]
@mo-nathan LOL. That's because I just pushed out this morning's update to `arel_extensions` gem, which silences all the warnings. There were 1700 of them, I keep saying!

I'm busy removing all these `unscoped`s right now.

**Comment by mo-nathan** [comment]
Ok, thanks.

File: app/models/name/taxonomy.rb [hunk 6]
```diff
@@ -468,9 +468,9 @@ def best_matches_from_array(names, includes = [])
     # Now allows includes, for batch lookup of Naming email interested parties
     # GOTCHA: `search_name` cannot be used as a field in this AR where clause
     def batch_lookup_all_matches(name_or_names, includes = [])
-      Name.includes(includes).where(Name[:search_name].in(name_or_names)).
-        or(Name.where(Name[:text_name].in(name_or_names))).
-        with_correct_spelling
+      Name.unscoped.where(Name[:search_name].in(name_or_names)).
+        or(Name.unscoped.where(Name[:text_name].in(name_or_names))).
+        with_correct_spelling.includes(includes)
```
nimmolo, line 473: I believe this is an appropriate use of `.unscoped`, as is the one below. 

These are one-off queries that should not be otherwise scoped and do not rely on order.

nimmolo:
@mo-nathan If you have the bandwidth, please do look it over now.

I got rid of almost all those `unscoped`s. There are a few cases where they seem appropriate and necessary, as long as we're going to have default ordering. Believe me, I considered scrapping it.

PR #2626, opened 2025-01-02
Enum deprecation solution attempts

nimmolo:
Adds a new zero value for `Name.rank` and `Observation.source` to avoid doing migrations to change these enum values. Modifies `Name.all_ranks` to skip the new 0 "unused" value. Also adds new zero `:Unstarted` value to `FieldSLipJobTracker.status`, because it seemed suitable.

Adds a migration to decrement all non-null existing integer values for the following columns:
```
users.image_size
users.thumbnail_size
users.location_format
users.votes_anonymous
users.hide_authors
users.keep_filenames

location_descriptions.review_status

name_descriptions.review_status
name_descriptions.source_type

project_members.trust_level
```


db/schema.rb
```diff
@@ -909,13 +909,11 @@
     t.text "alert"
     t.boolean "email_locations_admin", default: false
     t.boolean "email_names_admin", default: false
-    t.integer "thumbnail_size", default: 1
     t.integer "image_size", default: 5
     t.string "default_rss_type", limit: 40, default: "all"
     t.integer "votes_anonymous", default: 1
     t.integer "location_format", default: 1
```
mo-nathan, line 915: I suspect the change in the meaning of the default here and in other locations is a major source of the test failures you're seeing.

db/schema.rb
```diff
@@ -929,6 +927,8 @@
     t.boolean "no_emails", default: false, null: false
     t.string "inat_username"
     t.integer "original_image_quota", default: 0
+    t.integer "hide_authors"
```
mo-nathan, line 930: Is losing the `null: false` here intentional?  I'm guessing this causes the default to be `nil` rather than `0`.

mo-nathan (changes requested):
Overall I like the approach.  Do you know how long the migration takes on a current database?  It's great that you found what appears to be a single SQL statement for these updates.  I was worried that you'd have to take a much slower approach.

Given the level of errors and the untested nature of the migrations at this point, I did not look at the branch on my own local.  However, I did spot some issues that I think are causing at least some of the test failures.

db/schema.rb
nimmolo: Wow, good catch. I didn't change these intentionally!

nimmolo:
Ugh. This PR is nonsense; I misread the deprecation warning. 

We don't need zero values. We just need to declare the enum with a symbol.

On MO, this leads to other warnings about conflicting enum value magic method names, in which case we need to prevent Rails from creating those.

It's about this:
```ruby
#  bad
enum status: { open: 1, closed: 2 }, _prefix: true
# good
enum :status, { open: 1, closed: 2 }, prefix: true, instance_methods: false
```

Issue #2630, opened 2025-01-03
Fix location merge workflow


mo-nathan:
From Igor:

Here is an example of what I am doing:

From the main activity page, select "Location" in the search window and press "search"
The above action gets me to https://mushroomobserver.org/locations
There is a list of 7 undefined locations on that page (in the right column)
One of these undefined locations is Eberwhite Nature Area, Washtenaw Co., Michigan, USA (with 2 observations)
However, this location already exists as https://mushroomobserver.org/locations/22737 (with 8 observations), so the undefined location is a DUPLICATE
I want to merge the above undefined duplicate record into the record of #22737. In the past, the merge action used to generate a merge request, wherein I would enter the reason for merging and send the request to MO Admins (Jason / Joseph).  The end result would be a single location, i.e. #22737, with 10 observations in it, and the undefined duplicate would be nuked.
I am clicking on the undefined Eberwhite record and that gets me to https://mushroomobserver.org/observations?where=Eberwhite+Nature+Area%2C+Washtenaw+Co.%2C+Michigan%2C+USA
From there I click on "Define this Location" and that gets me to https://mushroomobserver.org/locations/new?q=1xuyL
However, that last page is NOT populated with ANYTHING, so that is where I think the problem is in the code.  Something there is not working because in the past all the fields would usually be populated with the info from the UNDEFINED location, which I would change to what already exists in the defined location to initiate the merge action.  Ergo, in my understanding, https://mushroomobserver.org/locations/new?q=1xuyL is just a blank page that is NOT LINKED TO ANYTHING , it's basically a page to create a location from, not the page that would initiate the merging action.
Similarly, if I search for "Eberwhite", I get to https://mushroomobserver.org/locations?pattern=eberwhite.  When I click on the undefined "Eberwhite" with 2 observations, I get to https://mushroomobserver.org/observations?where=Eberwhite+Nature+Area%2C+Washtenaw+Co.%2C+Michigan%2C+USA, which is the same page as in step (7), and from there on, it's the same problem as above.

Anyway, that's all for now.  But the bottom line is whatever undefined duplicate locations are out there, they cannot be merged with the existing ones.  The same is true for novel undefined locations -- you can create a new record by going through some of the above steps, but in the end the undefined location is still where with its observations, and the newly created record lacks any observations...  I hope this makes sense now.

Let me know if you need any more details from me because I cannot think of anything else that can help you with troubleshooting.

mo-nathan:
[...]

PR #2631, opened 2025-01-03
Fix for location merge issue

mo-nathan:
This fixes the issue described in #2630, but this is just a quick fix which I don't think is the approach we really want to take.

### DIFF
--- a/app/helpers/tabs/observations_helper.rb
+++ b/app/helpers/tabs/observations_helper.rb
@@ -145,14 +145,24 @@ module Tabs
     # these are from the observations form
     def define_location_tab(query)
       [:list_observations_location_define.l,
-       add_query_param(new_location_path(where: query.params[:user_where])),
+       add_query_param(new_location_path(where: where_param(query.params))),
        { class: tab_id(__method__.to_s) }]
+    end
+
+    # Hack to use the :locations param if it's present and the :user_where
+    # param is missing.
+    def where_param(params)
+      result = params[:user_where]
+      return result if result
+      return params[:locations][0] if params.include?(:locations)
+
+      ""
     end
 
     def assign_undefined_location_tab(query)
       [:list_observations_location_merge.l,
        add_query_param(matching_locations_for_observations_path(
-                         where: query.params[:user_where]
+                         where: where_param(query.params)
                        )),
        { class: tab_id(__method__.to_s) }]
     end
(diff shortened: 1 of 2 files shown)


app/helpers/tabs/observations_helper.rb
```diff
@@ -145,14 +145,20 @@ def observations_at_where_tabs(query)
...
+       add_query_param(new_location_path(where: where_param(query.params))),
        { class: tab_id(__method__.to_s) }]
     end
 
+    # Hack to use the :locations param if it's present and the :user_where
+    # param is missing.
+    def where_param(params)
+      params[:user_where] || params[:locations][0]
+    end
+
```
nimmolo, line 157: It looks like Igor's example page (that has the `define_location_tab`) had the **incoming** `where` param you need:
`https://mushroomobserver.org/observations?where=Eberwhite+Nature+Area%2C+Washtenaw+Co.%2C+Michigan%2C+USA`

Does `params[:user_where] || params[:where]` not work? IIRC `user_where` would only come from an Advanced search.

nimmolo: Oh, i notice you're using this below too.

nimmolo: OH. I get it. There's a confusion here between `query.params` and ACTUAL params. Hang on, i'm going to commit what I believe is a fix.

app/helpers/tabs/observations_helper.rb
```diff
@@ -145,14 +145,20 @@ def observations_at_where_tabs(query)
...
-       add_query_param(new_location_path(where: query.params[:user_where])),
+       add_query_param(new_location_path(where: where_param(query.params))),
        { class: tab_id(__method__.to_s) }]
     end
 
+    # Hack to use the :locations param if it's present and the :user_where
+    # param is missing.
+    def where_param(params)
+      params[:user_where] || params[:locations][0]
+    end
```
nimmolo, line 156:
```suggestion
    def where_param(query_params)
      query_params[:user_where] || params[:where]
    end
```

app/helpers/tabs/observations_helper.rb
nimmolo: Fix seems to work.

mo-nathan:
The coveralls failures aren't real.

PR #2633, opened 2025-01-04
Address some of the AREL deprecation warnings

mo-nathan:


nimmolo:
The other method that is throwing a ton of warnings is `batch_lookup_all_matches` in `models/name/taxonomy.rb`

nimmolo:
I merged this already into #2621, and took your lead of examining all the methods throwing these for reasons they could make Arel grumpy.

These primarily seemed to result from `or` clauses or joins where two scopes had incompatible ordering conditions, because of the new default ordering scopes on that PR. I examined the AR and added `reorder("")` or `unscoped` to these where it seemed right, and the warnings are down from 1700 to 100. 

Note that help is on the way, there's a [PR](https://github.com/Faveod/arel-extensions/pull/126) on `arel_extensions` gem to stop propagation of these warnings.

mo-nathan:
I don't understand your explanation.  I'll have to dig into it.

Issue #2640, opened 2025-01-06
Write cron job to maintain `Observation.center_lat` and `lng`


nimmolo:
`Location.update_box_area_and_center_columns` is intended to be run as a weekly cron job.

SolidQueue is now capable of running cron jobs. Let's write one.

PR #2641, opened 2025-01-06
Use SolidQueue to run a cron job repopulating Location and Observation center_lat/center_lng

nimmolo:
~Not sure if this config is phrased right, need to do more research. But apparently~ that's all this takes: a config and a command. **Update**: [yes, it is that simple](https://github.com/rails/solid_queue/issues/356#issuecomment-2572744366).

Also, I'm not sure we need to run it weekly. Monthly might be A-OK. There was some question initially about whether our callbacks would populate these correctly... we should check how many are currently missing in the db.

Responds to #2640

### DIFF
--- a/app/jobs/update_box_area_and_center_columns_job.rb
+++ b/app/jobs/update_box_area_and_center_columns_job.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class UpdateBoxAreaAndCenterColumnsJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    log("Starting UpdateBoxAreaAndCenterColumnsJob.perform")
+    # This should be zero, but count just in case.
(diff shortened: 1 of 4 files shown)


nimmolo:
@JoeCohen @mo-nathan 
We could run this job silently as I initially set it up, but now it occurs to me, we'd probably want a real job that counts what it's updating and logs a report, right?

nimmolo:
I don't think it's something that we would review, it's that we'd want a log to know what's been happening in case something went awry.

These columns are the key to our being able to switch from obs locations as boxes to obs locations as the center points of boxes that are under a given threshold. Once we're displaying locations like that, we are going to want to know what's happening with these columns.

nimmolo:
Considering how many obs need updating since our original migration (>20 per day, i'm estimating), and the fact that people will probably want to find recent obs reflected on maps, i've changed the frequency of this job to "nightly".

PR #2642, opened 2025-01-06
Suppress warning in the one test where we call to_time

mo-nathan:
Let's see if CI passes and if the code passes for all the other devs.  If so, we should just merge this.

### DIFF
--- a/test/controllers/admin/blocked_ips_controller_test.rb
+++ b/test/controllers/admin/blocked_ips_controller_test.rb
@@ -5,6 +5,7 @@ require("test_helper")
 module Admin
   class BlockedIpsControllerTest < FunctionalTestCase
     def test_blocked_ips
+      ActiveSupport.to_time_preserves_timezone = true
       new_ip = "5.4.3.2"
       IpStats.remove_blocked_ips([new_ip])
       # make sure there is an API key logged to test that part of view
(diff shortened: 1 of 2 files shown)


JoeCohen:
@mo-nathan: 
I suggest commenting out the failing assertion in `test/jobs/inat_import_job_test.rb:209`.
It fails because the test stubbing is screwed up. 
I'm working on it at the moment, but don't know if I will finish today. 
I don't think we should let it hold up other PRs.  It's "just" a test thing.

nimmolo:
@mo-nathan @JoeCohen 
The question here is the global config, though, isn't it? 
Rails 7.2 wants us to consider opting in to this config in application.rb.

JoeCohen (commented):
It passes for me locally.
(with `ActiveSupport.to_time_preserves_timezone = true` in `test_blocked_ips`).

mo-nathan:
I think they are only suggesting setting it if you're using it.  It makes no difference for us outside of exactly that one test since we don't use it production code.  I see no reason to not just ride with whatever the Rails maintainers think is the default since we really don't care.

nimmolo:
Still sounds like it should probably be in config/application.rb to prevent a deprecation warning in case we use to_time in some other test in the future.

mo-nathan:
I think Nimmo and I are aligned on this approach.  Haven't heard from @JoeCohen.  I'll wait until this evening or tomorrow morning for more feedback.  I'll go ahead and merge if I don't hear anything.

JoeCohen:
Makes sense to me. (Sorry. I thought I had responded earlier.)

PR #2644, opened 2025-01-10
Project aliases

mo-nathan:
This implements a feature where project admins can define a set of "aliases" for users and locations.  E.g., `NJW` represents the user `nathan` or `Walk 10` represents the location `USA, Massachusetts, Wellfleet, Marconi Beach`.  The aliases only work in the context of the project (specifically when filling in field slip forms).

Most of the UI is on the Members and Locations tabs for a project.  From there you should be able to create, edit, and deleted aliases for either users or locations.  The primary UI uses turbo.  The HTML equivalent pages are still available just in case.


app/controllers/projects/project_aliases_controller.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module Projects
+  class ProjectAliasesController < ApplicationController
```
nimmolo, line 4: I think it will make the routes come out as desired if you just call this `Projects::AliasesController` (or `module Projects class AliasesController`). 

For example, the controller for NameDescriptions is `Names::DescriptionsController` even though the model is `NameDescription` (as this is `ProjectAlias`).

app/views/controllers/projects/project_aliases/_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  </div>
+
+  <div class="field">
+    <%= form.label :target_type %>
+    <%= form.select :target_type, ['Location', 'User'] %>
+  </div>
+
+  <div class="field">
+    <%= form.label :target_id %>
+    <%= form.number_field :target_id %>
```
nimmolo, line 30: For style consistency, remember these form helpers are just Ruby methods, and even though Rubocop doesn't lint .erb (yet) the MO style in all cases, including erb, is parentheses for method args.

config/routes.rb
```diff
@@ -660,6 +660,7 @@ def route_actions_hash
                           controller: "projects/locations"
     resources :members, only: [:new, :create, :edit, :update, :index],
                         controller: "projects/members", param: :candidate
+    resources :project_aliases, controller: "projects/project_aliases"
```
nimmolo, line 663: I think this should be `resources :aliases, controller "projects/aliases"` according to what i'm proposing above. Note that it **_doesn't_** mean that you have to namespace of the model(!) - this threw me for a long time.

app/views/controllers/projects/project_aliases/edit.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Editing Project Alias</h1>
+
+<%= render 'form', project_alias: @project_alias, url: project_project_aliases_path(project_id: @project_alias.project_id, id: @project_alias.id) %>
+
+<%
+  project_id = @project_alias.project_id
+%>
+<%= link_to 'Show', project_project_alias_path(project_id:, id: @project_alias.id) %> |
+<%= link_to 'Back', project_project_aliases_path(project_id:) %>
```
nimmolo, line 9: Ditto here, we use parentheses for `render` and `link_to`, all helper methods.

app/views/controllers/projects/project_aliases/_form.html.erb
mo-nathan: This was a side effect of using a generator.  Not intentional.  Thanks for catching it.

mo-nathan:
Thanks Nimmo!  This is much better.  More work to go, but this feels like a good first cut.

app/views/controllers/projects/project_aliases/_form.html.erb
nimmolo: I figured... Just now i got curious if we can configure our generators to follow our rubocop config, but i haven't found a way yet. That would be convenient, right? Maybe I'm not using the right keywords.

config/routes.rb
nimmolo: Remember when i spent probably 3 months trying to get `module Name class Description` to work? 2019 i think. Sheesh. What a weird obsession... Good times.

app/views/controllers/projects/project_aliases/_form.html.erb
mo-nathan: Yeah, that would be great.  Did you see this: https://guides.rubyonrails.org/generators.html ?

mo-nathan:
- [ ] Add Project header
- [ ] Clean up the form
- [ ] Apply the aliases when they get saved to the field slip
- [ ] Tests

JoeCohen:
Will this PR:
- accommodate uploading a list of foray registrants, with aliases (if any) to be completed later? (If the foray organizers ask for it on the registration form, the upload could include MO login. Maybe also iNat login.)
- allow for a future (pie-in-the-sky) PR which does type-ahead in the Collector, ID By, and Location fields of the Field Slip form?

mo-nathan:
@JoeCohen This PR is just going to do the basic CRUD for ProjectAliases and apply them to values in project field slips when they get saved.  Once we're sure they are reasonable, then adding a facility for bulk creation of users with aliases for a specific project makes sense.  I'm not sure about whether it's worth doing the type ahead bit since the point is to make these aliases short enough that they are really easy to type.  For example, if the walks at "1", "2", ... "15" and the users are "NJW", "JDC", and "NS" is there any need for type ahead?

JoeCohen:
I was thinking about people not entering the the alias on the Field Slip:

   - Field slip has "Collector: Richard Bingly"
   - Recorder types something like "Richard Bingl" and type ahead shows
   "Richard Bingle".

Also thinking about adding aliases on the fly.

   - Recorder types Collector name as it appears (or seems to appear).
   - Type ahead shows nothing
   - Recorder consults a reliable list of participants, finds that
   participant, adds an alias.

nimmolo (dismissed):

test/models/project_alias_test.rb
```diff
@@ -0,0 +1,16 @@
...
+class ProjectAliasTest < ActiveSupport::TestCase
+  test "valid user aliases" do
+    pa = Project.find_by(target_type: "User")
+    assert_equal("User", pa.target_type)
+    assert_equal(User, pa.target.class)
+  end
+
+  test "valid location aliases" do
+    pa = Project.find_by(target_type: "Location")
+    assert_equal(Location, pa.target.class)
```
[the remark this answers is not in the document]
mo-nathan: Not clear what you mean by this, but I note it's "Outdated" so maybe I already fixed it.

app/views/controllers/projects/aliases/_form.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        <% project_alias.errors.full_messages.each do |message| %>
+          <li><%= message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
+  <div class="field">
+    <%= text_field_with_label(form:, field: :name, label: "#{:Name.t}:", inline: true) %>
+  </div>
```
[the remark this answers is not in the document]
mo-nathan: Removed.

app/views/controllers/projects/aliases/edit.html.erb
```diff
@@ -0,0 +1,12 @@
+<%
+  add_project_banner(@project_alias.project)
+  @container = :wide
```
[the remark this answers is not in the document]
mo-nathan: Done.

app/views/controllers/projects/aliases/_form.html.erb
```diff
@@ -0,0 +1,50 @@
...
+                            field: :target_type,
+                            label: "#{:project_alias_type.t}:",
+                            options: [[:USER.l, :user], [:LOCATION.l, :location]],
+                            inline: true,
+                            selected: type,
+                            data: { autocompleter_target: "select",
+                                    action: "autocompleter#swap" }) %>
+    </div>
+
+    <%= autocompleter_hidden_field(form:, type:) %>
```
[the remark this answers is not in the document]
mo-nathan: I'm using:

`<%= autocompleter_hidden_field(form:, type:, hidden_value: project_alias.target_id) %>`

which seems to be working fine.  If you think there's still an issue, let me know how to see it.

app/views/controllers/projects/aliases/_form.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    <%= form.text_field(
+      :term, value: value, class: "form-control",
+      data: { autocompleter_target: "input" }
+    ) %>
+    <%= autocompleter_dropdown %>
+  <% end %>
+
+  <div class="actions">
+    <%= form.submit %>
+  </div>
```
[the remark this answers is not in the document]
mo-nathan: Thanks!  Done.

mo-nathan:
@JoeCohen I don't understand your "Richard Bingl" case.  Are there two users?  Or are you saying that the completed names shows somehow?  For example if I type "NJW" somewhere it shows "Nathan Wilson".  My worry about adding complexity here is that my own experience and the experience from Jess Evans is that the UI is already too slow.  I've been thinking about how to speed it up.  It might work out since one of my thoughts is to send the browser the info all the aliases when the page loads so the browser side could display the full name/login without a round trip to the server.  Nimmo and I did look at this at the current autocompleter is not behaving as expected since it does a server round trip every time you add a letter.  I think this is the main source of slowness on this form.

Regarding adding "aliases on the fly", that's a good idea.  I need to think about how that would work in the UI.  I am thinking about listing the aliases on the Members and Locations tabs.  I've realized that users and locations can and may often have more than one alias.

JoeCohen:
"Richard Bing" case. 
- Field Slip hard copy has  the Collector's full name, rather than an alias.
- It's not very legible, and appears to be "Richard Bingly"
- Recorder discovers the true alias/full name upon typing the first 12 (or fewer) characters of "Richard Bingly"

mo-nathan:
I think the described "Richard Bing" behavior is already implemented now that the default user format is "Full Name (login)".  However, the problem this creates is that now you cannot autocomplete using the login.  For example Bill Yule is "Bill (boletebill)".  To get the autocompleter to find him you need to type "Bill (bol" and it is unlikely that that info is going to be on the field slip.  Aliases at least help with this since we could create aliases for him for "BY" and "Bill Yule".  It turns out that his login already operates as an alias.  I do think the translation of an alias should show up in the UI via JavaScript, but I haven't implemented that yet.  Is there more to your suggestion that I'm missing?

app/javascript/controllers/autocompleter_controller.js
```diff
@@ -212,6 +212,7 @@ export default class extends Controller {
   // to refresh the primer (as with location_containing a changed lat/lng).
   //
   swap({ detail }) {
+    console.log("swap called");
```
nimmolo, line 215: Note that this Stimulus controller has a method `verbose`.
If you switch your `console.log("swap called")` to `this.verbose("swap called"), and then uncomment the actual method body of `verbose`, it will log a whole trace of what it's executing (as defined by wherever Jason and I left lines that say 
```js 
this.verbose("Say something");
```

JoeCohen:
I don't think this is blocker. 
It fills in the Field Slip with the plain text of an aliased User's login, rather than a link anchored by that text.
Ex:
For NAMA 2024, I created an alias `Nathan`.

Screenshot 2025-02-17 at 4 01 00 PM
Then created a Field Slip, filling in the ID By field with `Nathan`.
Result:
The `Field Slip ID By` field has `nathan`. I was hoping for `_user nathan_`.

JoeCohen (approved):
See my comment about user alias in the Field Slip.

app/views/controllers/projects/aliases/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      <tr>
+        <td><%= project_alias.name %></td>
+        <td><%= project_alias.target_type %></td>
+        <td><%= link_to(project_alias.target.try(:name), project_alias.target) %></td>
+        <td>
+          <%= link_to(:EDIT.t, edit_project_alias_path(project_id:, id: project_alias.id)) %>
+          <%= link_to(:DELETE.t, project_alias_path(project_id:, id: project_alias.id),
+	              data: { turbo_method: :delete, turbo_confirm: :are_you_sure.t }) %>
+        </td>
+      </tr>
```
nimmolo, line 30: Maybe worth checking out our `TableHelper`, that predates me.

You can iterate over your `@project aliases` in the controller or the template, generating a simple array of content. Then call `make_table` with this array and it renders all that HTML.

The reason I've been haphazardly trying to convert layouts to use it, is because it will be helpful to have the HTML centralized for the bootstrap upgrade. Applying styling can happen in the helper, rather than tracking down every table in the layouts.

app/models/project_alias.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ProjectAlias < AbstractModel
```
nimmolo, line 3: Maybe add a comment block here that explains the purpose of this model for newbies, like 
```ruby
# Project admins can define a set of "aliases" for users and locations to facilitate data entry in the field. 
# E.g., "NJW" can be made to represent the user `nathan`, or "Walk 10" for the location
# `USA, Massachusetts, Wellfleet, Marconi Beach`. 
# The aliases only work in the context of the project 
# (specifically when filling in field slip forms).

mo-nathan: Done.

app/views/controllers/projects/aliases/index.html.erb
mo-nathan: Done.

PR #2648, opened 2025-01-12
Replace `default_scope` with `scope :index_order`

nimmolo:
This was a relatively quick fix, but may also make these scopes useful for the indexes.


nimmolo:
Removes many dozens of `reorder` and the few `unscoped` calls also, now redundant.

Issue #2646, opened 2025-01-11
Showing nonexistent obs


JoeCohen:
Works fine locally, displaying a flash error "Sorry, the observation you tried to display (id #nnnnnn) does not exist."
BUT throws an error on the webserver.
Example:
http://localhost:3000/569282
vs
https://mushroomobserver.org/569282
e.g.:
```ruby
INFO -- : Started GET "/569282?q=1oaSB" for 50.107.155.36 at 2025-01-11 16:11:41 +0000
INFO -- : Processing by ObservationsController#show as HTML
INFO -- :   Parameters: {"q"=>"1oaSB", "id"=>"569282"}
WARN -- : user=667 robot=N ip=50.107.155.36
INFO -- :   Rendered layout controllers/layouts/application.html.erb (Duration: 13.8ms | GC: 0.1ms)
INFO -- : Completed 500 Internal Server Error in 387ms (ActiveRecord: 340.3ms (37 queries, 2 cached) | GC: 1.0ms)
E, [2025-01-11T16:11:42.072393 #4039156] ERROR -- :   
ActionView::Template::Error (undefined method `legal_name' for nil):

Causes:
NoMethodError (undefined method `legal_name' for nil)
    15:                                 presenter.image_link_method))
    16:   end
    17:   concat(lightbox_link(presenter.lightbox_data))
    18:   concat(carousel_caption(image, object, presenter))
    19: end
    20: %>

app/helpers/images_helper.rb:103:in `show_image_copyright?'
app/helpers/images_helper.rb:89:in `image_copyright'
app/helpers/images_helper.rb:65:in `image_info'
app/helpers/carousel_helper.rb:7:in `carousel_caption'
app/views/controllers/shared/_carousel_item.erb:18
app/views/controllers/shared/_carousel_item.erb:10
app/views/controllers/shared/_carousel.erb:22
app/views/controllers/shared/_carousel.erb:21:in `each'
app/views/controllers/shared/_carousel.erb:21:in `each_with_index'
app/views/controllers/shared/_carousel.erb:21
app/views/controllers/shared/_carousel.erb:20
app/views/controllers/shared/_carousel.erb:18
app/views/controllers/observations/show.html.erb:29
app/views/controllers/observations/show.html.erb:28
app/controllers/application_controller.rb:218:in `catch_errors_and_log_request_stats'
```

JoeCohen:
Looks like observation 569282 was somehow created without a user.

Nathan Wilson
[...]

PR #2650, opened 2025-01-12
Un-owned Observation Error

JoeCohen:
Prevents server from throwing an Error when a user shows an Observation whose user_id == nil.

Delivers #2646

### DIFF
--- a/app/helpers/images_helper.rb
+++ b/app/helpers/images_helper.rb
@@ -100,7 +100,7 @@ module ImagesHelper
   def show_image_copyright?(image, object)
     object.type_tag != :observation ||
       (object.type_tag == :observation &&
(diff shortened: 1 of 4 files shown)


mo-nathan (changes requested):
I'd add at least one test for this.  You should be able to create an image and an observation with no user and see if they can be rendered.

JoeCohen:
Thanks. Good point. I will add those.

JoeCohen:
@mo-nathan: I added tests.

(BTW, the chron job doesn't null image.user of Images of the disabled user's Observations.
Rather it deletes unattached Images, leaving the other Images unchanged.)

mo-nathan (approved):

PR #2653, opened 2025-01-14
Move a few more things around in Query

nimmolo:
Just moving methods around in included modules.


nimmolo:
The ABC violations are not new, they're just moved to a new file. I'm about to rewrite the methods for AR so i'm not going to spend time on them here.

PR #2658, opened 2025-01-15
Fix identify UI for proposing or voting on names

nimmolo:
Several fixes here. 
- Vote `:value` param was not properly nested (under `:vote`) when the user had not already voted. `form_with` needed `scope: :vote` in that case. It already gets `model: @vote` it they have voted.
- "Propose a name" link was hard to read in the lightbox view
- Turbo now explicitly `remove`s the modals (either vote progress spinner or propose form). Previously it was depending on the `updated` event being fired, but it does not get fired because the elements are `replace`d by Turbo, not `update`d.
- Somehow on local I seem to have lost the CSS that shows/hides the location format that is not the user's preference. This gives the CSS a fallback value.

### DIFF
--- a/app/helpers/lightbox_helper.rb
+++ b/app/helpers/lightbox_helper.rb
@@ -35,10 +35,10 @@ module LightboxHelper
   def lightbox_obs_caption(obs, identify)
     html = []
 
-    html << caption_identify_ui(obs: obs) if identify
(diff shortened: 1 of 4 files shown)


nimmolo:
@mo-nathan if this does the trick please feel free to merge and deploy

PR #2659, opened 2025-01-16
Restate polymorphic associations explicitly to allow joining

nimmolo:
Explained [here](https://dev.to/anakbns/polymorphic-joins-in-active-record-2hoj) and here in the paragraph "[Define the association with a scope](https://veelenga.github.io/joining-polymorphic-associations/)".

Currently, both these throw an error:
```ruby
Comment.joins(:location).where(target_id: location.id)
Comment.joins(:target).where(target_id: location.id, target_type: "Location")

#<ActiveRecord::EagerLoadPolymorphicError: Cannot eagerly load the polymorphic association :target>
```
Adding explicit `belongs_to` associations with a scope and foreign_key for each `target` polymorphic relationship allows us to write:
```ruby
Comment.joins(:location).where(target_id: location.id)
```
and is far more efficient (explained in the second link) than this:
```ruby
Comment.includes(:locations).where(target_id: location.id, target_type: "Location")
```

### DIFF
--- a/app/classes/api2/comment_api.rb
+++ b/app/classes/api2/comment_api.rb
@@ -28,13 +28,13 @@ class API2
     end
 
     def query_params
-      @target = parse(:object, :target, limit: Comment.all_types, help: 1)
+      @target = parse(:object, :target, limit: Comment::ALL_TYPES, help: 1)
       {
         where: sql_id_condition,
         created_at: parse_range(:time, :created_at),
         updated_at: parse_range(:time, :updated_at),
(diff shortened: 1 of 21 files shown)


JoeCohen (approved):
Thanks for doing this and for the references.

nimmolo:
Refactoring this to DRY, i discovered some mixed use of class methods and class constants to define arrays of `types` in various models. Since these are all constant arrays of model types, I regularized them all as `CLASS_CONSTANT = %w[this that].freeze`.

PR #2664, opened 2025-01-18
`AutoComplete::ForUser` — restrict to verified users

nimmolo:
Also adds User class method `unique_text_name` for combined name attributes, because the instance method is no longer available in AutoComplete when we use AR to select the columns we need (they're not instances of User anymore).

### DIFF
--- a/app/classes/auto_complete/for_user.rb
+++ b/app/classes/auto_complete/for_user.rb
@@ -27,7 +27,7 @@ class AutoComplete::ForUser < AutoComplete::ByString
       name = user.unique_text_name
       user = user.attributes.symbolize_keys
       user[:name] = name
-      user.except(:login, :bonuses) # idk why this is getting bonuses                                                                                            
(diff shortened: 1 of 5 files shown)


app/classes/auto_complete/for_user.rb
```diff
@@ -28,10 +28,10 @@ def matches_array(users)
       user[:name] = if user[:name].empty?
                       user[:login]
                     else
-                      "#{user[:login]} <#{user[:name]}>"
+                      User.unique_text_name(user)
                     end
```
mo-nathan, line 32: Why are you doing this rather than `user.unique_text_name`?

mo-nathan: Proposed changes here: https://github.com/MushroomObserver/mushroom-observer/pull/2665

nimmolo: Thank you

PR #2665, opened 2025-01-19
Update handling of unique_text_name

mo-nathan:
I think this is equivalent and allows us to get rid of the new redundant class function in user.rb.


nimmolo:
Slick! For some reason i was stuck on "not having user instances" at some point in the loop, but now I forget where.

PR #2669, opened 2025-01-22
Quick fix for advanced search user search

mo-nathan:
This is a bit slap dash and the parsing of the new name format would be better done by the parser in the Project alias PR.  There should also be tests for this.  In the process of testing this, I noticed that even if I select my name via autocomplete (now "Nathan Wilson (nathan)" it still returns all observation for users whose login contains "nathan" which it shouldn't.  However, I think that gets into getting query to use :user_id rather than :user which didn't work when I tried it naively.  I don't know how deep that rabbit hole is.  The behavior in this PR seems to return us to the old behavior.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -430,6 +430,31 @@ class User < AbstractModel # rubocop:disable Metrics/ClassLength
     unique_text_name
   end
 
+  def self.lookup_unique_text_name(str)
+    return nil unless str
+
+    user = nil
+    login = nil
+    if (match = str.match(/\(([^(]+)\)$/))
+      login = match[1]
+      user = find_name_match(User.where(login:), str)
+    end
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -430,6 +430,31 @@ def format_name
...
+  end
+
+  def self.find_name_match(users, str)
+    return users.first if users.count == 1
+
+    users.find_each do |user|
+      return user if user.unique_text_name == str
+    end
+  end
+
```
nimmolo, line 457: Is this basically the method from your Project Alias PR?

nimmolo (approved):
Left a question...

app/models/user.rb
nimmolo: Nvm. now that i'm not on a phone, i can see that it is.

PR #2676, opened 2025-01-25
Move PORO tests to new test/classes folder

nimmolo:
Easier to find PORO tests this way, for me, and hopefully for others too.

The test/models folder has accumulated a lot of non-model stuff in it over the years; it's been the home for any unit test that didn't fit elsewhere. 

But it's about to get even more crowded. This also breaks up QueryTest into 20 modules, one for each of the different models, as I'm adding tests for uncovered params.
___
Bonus: renames the PORO `AutoComplete` to `Autocomplete`


JoeCohen (approved):

nimmolo:
I went ahead and split up QueryTest here, so i can track the diffs on the next PR. They would be hidden by the split otherwise.

@JoeCohen - I think we should reconsider our setting for the cop `Naming:VariableNumber`. The setting we're using, `EnforcedStyle: normalcase`, invalidates fixture names like this:
```ruby
comments(:minimal_unknown_obs_comment_1)
```
But the style it's preferring is less legible, to me:
```ruby
comments(:minimal_unknown_obs_comment1)
```
Let's consider changing the cop to `EnforcedStyle: snake_case`.

PR #2692, opened 2025-02-01
Tab PORO Proposal

mo-nathan:
This is an attempt to get away from using the pattern `tab_id(__method__.to_s)` in our code.  Currently this pattern is only really used to enable some of our test to find things in the DOM.  The current code allows all tests to pass with a generic string returned by `tab_id()`.

It introduces a couple of POROs to handle the concept of "tabs" (`ModelTab` and `CoercedQueryTab`).  Note that the `alt_title` option for `ModelTab` is a hard override used to allow all the existing tests to pass.  In at least most cases ("nonpersonal_herbaria_index" might be an exception), I would look at fixing the test to avoid the use of the `alt_title`.

If this approach is agreed upon, `tab_id` should be removed and existing cases should be switched over to this approach.

### DIFF
--- a/app/classes/coerced_query_tab.rb
+++ b/app/classes/coerced_query_tab.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+class CoercedQueryTab
+  def initialize(query, model, html_options: {})
+    @query = query
+    @model = model
+    @html_options = html_options
+    @html_options[:class] = html_class unless @html_options.include?(:class)
+  end
+
+  def tab
+    [:show_objects.t(type: @model.type_tag),
+     { controller: @model.show_controller,
+       action: @model.index_action,
+       q: @query.id.alphabetize },
+     @html_options]
+  end
+
+  private
+
+  def html_class
+    "coerced_#{@model.name.underscore}_query_link"
+  end
+end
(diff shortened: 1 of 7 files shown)


JoeCohen (approved):
The explanation makes sense.
I looked at the code, but don't understand the details.

nimmolo:
I LOVE IT. I wonder if you need the nesting under `html_options`: iirc it's only going to be `class` `id` and `icon`. (Do some have data-attributes in the tab array? Ah, i can just look in helpers/tabs... how nice! Searching for `data:` in the entire front-end codebase would be inefficient. Yes there are a few with data attributes.)

One request then: make 'em both subclasses of a module `Tab` to keep them together in app/classes. `Tab::CoercedQuery` and `Tab::Model`. (In general, I'm always going to request this naming convention for POROs. I have a strong memory of how bewildering it was for me trying to learn the POROs at first because they were not grouped any logical way in the classes folder when i started; i dreaded even opening this directory because of the learning curve.)

mo-nathan:
@nimmolo I was also going to proposed something "link"-ish rather than "tab"-ish for this and I did consider having a base class so I'll move that direction.  I just wanted to get my first draft up and working to get your input before going too far down the rabbit hole.  Since all seem to be in agreement that this is the right direction for this, I'll work on getting everything using this approach.

mo-nathan:
Closing this in favor of a branch name that refers to "internal link" rather than "tab".

PR #2695, opened 2025-02-02
Switch from raw "tabs" to InternalLink POROs

mo-nathan:
This PR affects nearly every page on the site since it changes how "tabs" are handled.  In most cases so called "tabs" are really links to other pages and other websites.  When I started working on it, I thought they were all going to be internal links hence the name of the base class "InternalLink".  However, I have since learned that this are also getting used for some external websites.  We need to discuss the proper naming of these.  However, the primary goal of this PR is to significant reduce the use of the conceptually private system variable `__method__`.


nimmolo:
general_helpers.rb is where i'll be replacing the coerced query tabs with a different thing, so the InternalLink subclass will be different too. I checked the most recent commit and I think i'll be able to redo these easily after your PR merges.

mo-nathan:
Sounds good.  Just so you're aware, I also think we should do something a bit different in app/helpers/tabs/names_helper.rb.  See the function called `external_name_tab`.  It should probably be an`ExternalLink` class, but it has some commonality with the `InternalLink` classes.  Not sure what the best names are for these things.

mo-nathan:
There really isn't any difference between "internal" and "external" links.  Both come in as URLs.  The only really interesting thing that `external_name_tab` does is add `{ target: :_blank, rel: :noopener }`.  I think I'm going to finish what I'm currently doing with InternalLink and get that done.  Then we can talk about the next phase.

mo-nathan:
Also just ran across all the stuff in `app/helpers/link_helper.rb` like `destroy_button` and `edit_button` which seems to duplicate what is going on with some of the stuff in `app/helpers/tab/`.

nimmolo:
@mo-nathan Seems that `coerced_query_tab` is now using `InternalLink::Model`. Is anything using `InternalLink::CoercedQuery`? I couldn't find it scrolling through all the changes, but i may have missed it.

Also, you've got a front-end note to add a User autocompleter in `views/controllers/descriptions/authors/show.rb`, that's fine with me, just checking if you want that or a `# NOTE` comment.

mo-nathan:
CoercedQuery is being used here:
```
app/helpers/tabs/observations_helper.rb:      [InternalLink::CoercedQuery.new(query, Location).tab,
app/helpers/tabs/observations_helper.rb:       InternalLink::CoercedQuery.new(query, Name).tab,
app/helpers/tabs/observations_helper.rb:       InternalLink::CoercedQuery.new(query, Image).tab]
```
I initial converted `coerced_query_tab` to use InternalLink::Model because that was the easiest way match to the original behavior.  However, it could use CoercedQuery and now that I look more closely at it, that's probably reasonable.

Regarding `views/controllers/descriptions/authors/show.rb`, I didn't sweat it because it doesn't work on the site as it currently stands and it's another rabbit hole.

nimmolo:
Sounds good. No big deal either way, I plan to immediately replace `coerced_query_tab` and `InternalLink::CoercedQuery` with my next PR, but helpful to know where they are.

nimmolo:
It's erroring on cases with a nonexistent `q` param.

app/controllers/names/maps_controller.rb
```diff
@@ -12,6 +12,7 @@ def show
       return unless @name
 
       @query = create_query(:Observation, names: @name.id)
+      @query.save
```
nimmolo, line 15: Check `ApplicationController::Queries#find_or_create_query(model_symbol, args = {})`

This seems to be what other controllers are calling instead of these two lines.

nimmolo: Yes, that's what `controllers/locations/maps_controller.rb` is doing. 
I think that's the way to go here for now, until we revisit these methods.

nimmolo, line 15:
```suggestion
      @query = find_or_create_query(:Observation, names: @name.id)
```
Check `ApplicationController::Queries#find_or_create_query(model_symbol, args = {})`

This seems to be what other controllers are calling instead of these two lines. Yes, that's what `controllers/locations/maps_controller.rb` and `controllers/observations/maps_controller.rb` are doing. I think that's the way to go here for now, until we revisit these methods.

nimmolo (changes requested):

PR #2700, opened 2025-02-05
Quick fix to tame the validate_value query explosion

mo-nathan:


mo-nathan:
We're not going this way...

PR #2701, opened 2025-02-05
Try to fix validation explosive recursions

nimmolo:


test/classes/api2/images_test.rb
```diff
@@ -156,32 +156,37 @@ def test_getting_images
...
-      name: agaricus, images: [agaricus_img], thumb_image: agaricus_img,
-      user: rolf
-    )
-
-    assert_api_pass(
-      method: :get, action: :image, name: "Agaricus", include_subtaxa: "yes"
-    )
-    assert_api_results(imgs << agaricus_img)
-  end
+  # FIXME: IMO this test should fail. This should be considered an invalid use
```
mo-nathan, line 159: Why is this marked as "FIXME"?

mo-nathan:
I think that's a false negative from Coveralls.

JoeCohen (commented):
@mo-nathan: Thanks for the changes! The code makes sense.
I've run it locally with some troublesome requests, and the request completes.
@nimmolo: Still lots of queries, but that could be for another PR
- Observations of Agaricus, sorted by confidence.
`ActiveRecord: 20985.6ms (7063 queries, 2297 cached)`
- Observations of Psilocybe, sorted by confidence.
`ActiveRecord: 12032.1ms (2890 queries, 764 cached)`

mo-nathan:
Most of this PR is now about getting rid of the fake duplicate, Lentinellus ursinus Kuhner.

However, the meat of the change is all in app/classes/query/modules/validation.rb.  The fixes I made were to ensure that the the synonym and subtaxa stuff gets applied to all the ids rather than one at a time and that it gets run in all appropriate cases.  I also eliminated a uniq that was causing a test to fail, but I don't think was needed any more.

mo-nathan (approved):

nimmolo:
Thanks for the changes! 

@JoeCohen's query counts there still seem very high, don't they? @nathan is that what you're getting? Is this somehow reversing your improvements?

PR #2703, opened 2025-02-05
Speed up Lookup::Names

mo-nathan:
To test from a database dump from Feb 5:
```
% rails c
mushroom-observer(dev)> ids = QueryRecord.find("1yzec".dealphabetize).description.split(",#")[1...]
mushroom-observer(dev)> Lookup::Names.new(ids, include_synonyms: true, include_subtaxa: true).ids.count
```

On current main this will generate over 500 SQL queries.  On the branch it will generate 5.

### DIFF
--- a/app/classes/lookup/names.rb
+++ b/app/classes/lookup/names.rb
@@ -47,6 +47,18 @@ class Lookup::Names < Lookup
 
   # Matches for the given vals, from the db.
   def original_matches
+    @original_matches ||= if numeric_list?(@vals)
+                            numeric_matches
+                          else
+                            map_matches
+                          end
+  end
+
+  def numeric_matches
+    Name.where(id: @vals).distinct.select(*minimal_name_columns)
+  end
+
+  def map_matches
     @original_matches ||= @vals.map do |val|
       if val.is_a?(@model)
         val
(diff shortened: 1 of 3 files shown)


app/classes/lookup/names.rb
```diff
@@ -60,6 +72,20 @@ def original_matches
...
   end
 
+  def numeric_list?(list)
+    list.all? do |item|
+      case item
+      when Numeric
+        true
+      when String
+        # Handles integers, floats, scientific notation, and negative numbers
+        item.match?(/\A-?\d*\.?\d+(?:[eE][-+]?\d+)?\z/)
```
nimmolo, line 82:
```suggestion
item.match?(/^-?\d+$/)
```
Since this is supposed to be a list of IDs, i do not think it should match "floats, scientific notation, and negative numbers", right?

nimmolo:
Amazing how much this reduces queries!

app/classes/lookup/names.rb
mo-nathan: Thanks!  Updated.

app/classes/lookup/names.rb
```diff
@@ -60,6 +72,20 @@ def original_matches
...
   end
 
+  def numeric_list?(list)
+    list.all? do |item|
+      case item
+      when Numeric
+        true
+      when String
+        # Handles integers only
+        item.match?(/^-?\d+$/)
```
JoeCohen, line 82: Do you want to match negative numbers? (Nimmo says it's "fed straight to Name.where(id: @vals)")

PR #2706, opened 2025-02-06
`MinimalLocation`/`Observation` to `ActiveModel `

nimmolo:
Moves these to inherit from `ActiveModel` along with their sibling, `Mappable::Box`

https://guides.rubyonrails.org/active_model_basics.html
https://guides.rubyonrails.org/active_record_validations.html#custom-validators

The most noticeable thing this changes is that you now have to instantiate them with attribute kwargs, as you would if you called `Observation.new(**atts)`, rather than just sending the args in the right order.


nimmolo:
@mo-nathan I guess my guiding consideration re: whether `ActiveModel` would be helpful is "will the object need validation?". A `Mappable::Box` clearly responds to validation, but for example the `Mappable::MapSet` and `CollapsibleCollection` are basically POROs for deriving a set of data about some given inputs, like the `ConsensusCalculator` for Observation naming votes, and don't seem to be meaningfully validatable. 

These two seem to be somewhere in between, so i'm not sure, but I'm leaning towards this pattern for them.

app/controllers/observations/maps_controller.rb
```diff
@@ -42,19 +43,24 @@ def find_locations_matching_observations
...
-        @query.select_rows(args).map do |id, lat, long, gps_hidden, loc_id|
-          locations[loc_id.to_i] = nil if loc_id.present?
-          lat = long = nil if gps_hidden == 1
-          Mappable::MinimalObservation.new(id, lat, long, loc_id)
+        @query.select_all(args).map do |obs|
+          obs.symbolize_keys!
+          # Setting this to nil is a flag to look up the location. Because we
+          # selected obs attributes not instances, we can't call `obs.location`
+          # and we shouldn't anyway. Getting the Locations in bulk is quicker.
+          locations[obs[:location_id].to_i] = nil if obs[:location_id].present?
```
mo-nathan, line 51: Minor point, but I think it would clearer if in this function `locations` was `location_ids` and handled with a Set and then `eager_load_corresponding_locations` returned the dictionary.  As written `locations` is kind of playing a double role and getting updated as a side effect in `eager_load_corresponding_locations`.

mo-nathan (approved):
Overall this seems reasonable.  For me it's a bit surprising to have an object in between a PORO and an ActiveRecord, but now that I understand it, it seems reasonable.  I think it's a difficult call to say whether future Rails devs would find this more or less intuitive.  In my own experience I think this is the first time I've come across an object that is derived from ActiveModel and not from ActiveRecord.  Before reviewing this I would have found it a bit of a head scratcher.  It might be worth putting a comment in on each `include ActiveModel::Model` line noting that it is NOT inheriting from `ActiveRecord`.  I'm guessing that at least 80% of Rails developers are like me.  I just checked the Rails projects I oversee at work and none of them mention `ActiveModel::Model`.

JoeCohen (approved):
Looks like what you called "model not backed by a db table" and suggested for one class for Inat imports. It's more useful in this PR than in Inat Imports. (I'm still on the fence about using it there.)

nimmolo:
@mo-nathan @JoeCohen Thanks for the review. 

Check PR #2337 to see where ActiveModel is really useful. It  allows me to pull all the fields of the search form into a single validatable object. I can then access that object in model-y ways, to rearrange the params for a Query. Very smooth and intuitve. 

That's the way i have read ActiveModel is most often used in the wild - for form objects. I think for example it could simplify dealing with our create obs form. It gives you something to validate that doesn't have to have the same validations as the object you're preparing it for, because of the difference between human input and DB values.

Having said that, the model classes on the Pattern Search Form PR are not as easy to read, because they derive their attributes from the PatternSearch classes. I may refactor that whole thing once I understand this better.

app/controllers/observations/maps_controller.rb
nimmolo: @mo-nathan Can you show me what you mean?

nimmolo: Reading up on Ruby `Set`. I don't remember ever coming across this class, incredibly.

nimmolo: Hmm, why don't we use a Set in the name validation refactor?

mo-nathan: I was in favor of using a Set when I thought we had to do things iteratively.  Now that we can get all the data in a single SQL query, it's not really worth it.  In general Sets are great.  They are way more efficient than lists or dictionary and folks are often aware of them or don't think to use them.

PR #2708, opened 2025-02-08
Quick patch to avoid expanding name lists

mo-nathan:
This goes back to the `.uniq` approach I proposed last week, but with an extra guardrail that allows all the tests to pass.  @nimmolo has a PR that should help even more, but I don't see any danger in pushing this quick fix out before that work settles.

### DIFF
--- a/app/classes/query/modules/validation.rb
+++ b/app/classes/query/modules/validation.rb
@@ -28,9 +28,7 @@ module Query::Modules::Validation
   def validate_value(param_type, param, val)
     if param_type.is_a?(Array)
       result = array_validate(param, val, param_type.first).flatten
-      if positive_integers?(result)
-        result = result.uniq
-      end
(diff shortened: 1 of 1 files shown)


app/classes/query/modules/validation.rb
```diff
@@ -27,7 +27,11 @@ def check_for_unexpected_params(old_params)
 
   def validate_value(param_type, param, val)
     if param_type.is_a?(Array)
-      array_validate(param, val, param_type.first).flatten
+      result = array_validate(param, val, param_type.first).flatten
```
JoeCohen, line 30: Is `result` ever an array of numerical strings, e.g., `["123", "145", ...]`?
If yes, should it be handled the same as an array of integers?

JoeCohen (approved):

PR #2709, opened 2025-02-08
Reduce the query load for searches

mo-nathan:
Without this change, searching for "Amanita" on our current database runs over 1800 queries.  With this change, it runs 42.


mo-nathan:
Coveralls is complaining about the logic from a line that had a trialing if that is never true in the tests so there is no actual change to the coverage in this file.  I also the idea of have queries generate titles is going away.

PR #2715, opened 2025-02-11
`Query`: convert "coercions" to subqueries

nimmolo:
This PR converts our current "coerced query" code to straightforward subqueries.

These are the queries that convert a query on one model to an analogous query on a related model, to produce the following results:
- Show Locations [of these Observations by user: nathan]
- Show Images [of these Names in clade: 'Petigera']
- Show Names [found at Locations of these Observations of Amanita jacksonii]

In the UI these links are very commonly visited by users. They just say "Show Locations", "Show Names", "Map Observations", etc. but under the hood they do some of the most complicated query juggling on the site.

This PR makes the queries for these links easier to transfer to AR from SQL. 

It also greatly reduces the number of params that our convertible Query classes must handle. No longer do Image and Sequence queries need to handle all the Observation params; they only need to handle an `:observation_query` param that holds the conditions for the subquery, and they instantiate an Observation query to generate the subquery SQL (or soon, AR).


mo-nathan (dismissed):
While there are a number of files that were deleted (718 vs. 727), there are more lines missed (1933 vs. 1921).  I haven't had the time to figure out what the missed lines are.  I'll try to figure that out tomorrow.

JoeCohen:
Coveralls is again functioning for me to show newly missed lines.

nimmolo:
Reviewed top uncovered lines and added tests, deleted some unused leftover code.

mo-nathan:
> While there are a number of files that were deleted (718 vs. 727), there are more lines missed (1933 vs. 1921). I haven't had the time to figure out what the missed lines are. I'll try to figure that out tomorrow.

Thanks for the additional tests!  For the record we're now at 1911 missed lines which is better than main.  It's interesting to me that Coveralls gave the thumbs up since the percentage of coverage has gone down slightly (93.48% on main, 93.45% on the branch).  The Coveralls check claims that the percentage has increased.  I noticed on the Project Alias PR that it seemed to not be comparing to the coverage on main.  I suspect that whatever happened recently in Coveralls may have lost track of previous coverage numbers at least for older PRs.

app/classes/api2/image_api.rb
```diff
@@ -60,8 +60,13 @@ def query_params
         with_votes: parse(:boolean, :has_votes),
         quality: parse_range(:quality, :quality),
         confidence: parse_range(:confidence, :confidence),
-        ok_for_export: parse(:boolean, :ok_for_export)
-      }.merge(parse_names_parameters)
+        ok_for_export: parse(:boolean, :ok_for_export),
+        observation_query: parse_observation_query_parameters.compact
```
mo-nathan, line 64: Is there a reason you're calling `.compact` twice?

app/helpers/tabs/species_lists_helper.rb
```diff
@@ -94,6 +94,46 @@ def destroy_species_list_tab(list)
...
+    def species_list_observations_tabs(list, query)
+      [species_list_observations_tab(query),
+       species_list_observations_locations_tab(list),
+       species_list_observations_names_tab(list),
+       species_list_observations_images_tab(list),
+       species_list_observations_checklist_tab(list),
+       species_list_observations_map_tab(query)]
+    end
+
+    def species_list_observations_tab(query)
```
mo-nathan, line 106: Shouldn't this use an InternalLink?

app/helpers/tabs/species_lists_helper.rb
```diff
@@ -94,6 +94,46 @@ def destroy_species_list_tab(list)
...
+      related_images_tab(:Observation, species_list_obs_query(list))
+    end
+
+    def species_list_observations_checklist_tab(list)
+      [:app_checklist.t, checklist_path(species_list_id: list.id)]
+    end
+
+    def species_list_observations_map_tab(query)
+      [:show_object.t(type: :map),
+       add_query_param(map_observations_path, query)]
```
mo-nathan, line 134: InternalLink?

app/classes/query.rb
```diff
@@ -304,6 +304,14 @@ def self.lookup(*)
     Query::Base.lookup(*)
   end
 
+  def self.current_or_related_query(*)
```
mo-nathan, line 307: Why is Query a module and not a class?  The documentation at the top of this page goes even further and claims that it's a "model".  This may be outside the remit of this PR, but I'd like to understand what should be fixed here.  At minimum it seems like we should update the documentation so it's accurate.  I don't see an obvious reason why Query should not be a class, so maybe that's the right direction.

app/views/controllers/locations/index.html.erb
```diff
@@ -25,42 +25,43 @@ flash_error(@error) if @error && @known_pages.empty? && @undef_pages.empty?
...
-      <div class="h4">
-        <%= :list_place_names_known.t %>
-        <%= :list_place_names_known_order.t if @default_orders %>
-      </div>
-      <%= paginate_block(@known_pages) do %>
-        <div class="list-group">
-          <% @known_data.each do |location| %>
-            <div class="list-group-item">
-              <%= location_link(location.name.t, location,
+    <div class="h4">
```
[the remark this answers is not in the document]
nimmolo: Oh my gosh... It's VSCode! 

It's started doing this automatic formatting with ERB and i haven't figured out how to disable it. If I absentmindedly save without remembering to command-p "Save without Formatting" it does this instantly and if I then close the file I can't undo easily. 

I prefer the tab-liberally approach.

app/classes/query.rb
nimmolo: Good question, this is still Jason's setup as far as I know. 

I agree it's a class, and furthermore I'm preparing to have it inherit from `ActiveModel` and adapt the validations. (That's what that little `Mappable:MinimalLocation` refactor last week was practice for. `Query` is pretty much the exact use case that the whole `ActiveModel::API` was designed around.)

So regardless, it may be changing soon. I'll try making it a class, but if it takes a lot of time i'm inclined to keep it this way until i can switch it to ActiveModel and attributes.

nimmolo: Made it a class

app/views/controllers/locations/index.html.erb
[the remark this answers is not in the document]
nimmolo: reverted to tabs

app/helpers/tabs/species_lists_helper.rb
nimmolo: done, but please check

app/helpers/tabs/species_lists_helper.rb
```diff
@@ -94,6 +94,46 @@ def destroy_species_list_tab(list)
...
+    def species_list_observations_names_tab(list)
+      related_names_tab(:Observation, species_list_obs_query(list))
+    end
+
+    def species_list_observations_images_tab(list)
+      related_images_tab(:Observation, species_list_obs_query(list))
+    end
+
+    def species_list_observations_checklist_tab(list)
+      [:app_checklist.t, checklist_path(species_list_id: list.id)]
```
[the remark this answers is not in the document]
nimmolo: done, but please check

app/helpers/tabs/species_lists_helper.rb
nimmolo: done, but please check

app/classes/api2/image_api.rb
nimmolo: good catch, fixed

nimmolo:
I was uncomfortable having the "user content filter" code, which may add new params after validation of sent params, mixed in with the validation code. 

So I moved it to another module. It does its own iteration through the params to be sure it isn't contradicting intentionally sent and validated params. It also must re-instantiate Query objects for each subquery, to check if the filter is applicable (which the Query object knows, but the params do not). 

This latter duplicates the instantiation of the Query objects during validation. It doesn't seem too costly, and it seems to pay off in code readability and separation of concerns.  

A better alternative might be to have the validator save each subquery's Query object as an added param, and for the filter checking method to then remove these objects when it's done with them. This would save the second instantiation (of the PORO), but params with Query objects would then be in a temporarily unexpected (but i think, harmless) state, coming straight out of the validator.

PR #2717, opened 2025-02-13
QueryRecord - convert `:description` `to_json` serialization

nimmolo:
There’s an issue with QueryRecords currently, and it becomes more important with the coercion-conversion PR.

Our `QueryRecord` uses a custom serializer to encode the `Query` params in the `:description` attribute, which is declared in schema as `text` and not Rails-serialized.

The custom `serialize` method in `Query::Modules::Serialization` can serialize arrays and values, but not nested hashes. That means that the current `:in_box` query param, which is a hash of `{ north:, south:, east:, west: }` is not getting saved in any QueryRecords. With my upcoming PR replacing coercions with subqueries, it means the filtering subquery is not saved.

We can solve this pretty easily by storing the query params using `to_json` instead. 

~Converting existing QRs in a reversible migration turns out to be harder, because once you declare the column `serialized`, it is possible to extract the existing string value from the column and save a Rails-serialized value back to it, but for the `down` migration I don't believe we can write a custom-serialized but not Rails-serialized string directly back to the column.~

We could try a one-way migration if we really want to ensure continuity.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_02_13_084440) do
+ActiveRecord::Schema[7.2].define(version: 2024_12_09_041049) do
   create_table "api_keys", id: :integer, charset: "utf8mb3", force: :cascade do |t|
     t.datetime "created_at", precision: nil
     t.datetime "last_used", precision: nil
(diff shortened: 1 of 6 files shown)


nimmolo:
I have an idea why the one test `SpeciesListsControllerTest#test_set_source` is failing. 

It seems that a new QR (and new QR.id) is being created when we're expecting it to stay the same.
That implies that some equality check isn't working.

JoeCohen (approved):
I agree with [@mo-nathan's Slack post about how to deploy this](https://mushroomobserver.slack.com/archives/C0889H2DYPN/p1739449043987999).
So it shouldn't be merged until we're ready to deploy.

nimmolo:
@JoeCohen Are you available to try this migration locally? 

I think it works now, and worst case scenario you just have to delete all your query records locally.

JoeCohen:
In about 10 minutes. I'll ping you when I'm available.
Do you really need the migration?

JoeCohen:
OK. I'm ready to try it. Should I just run rails db:migrate

db/schema.rb
```diff
@@ -929,6 +927,8 @@
     t.boolean "no_emails", default: false, null: false
     t.string "inat_username"
     t.integer "original_image_quota", default: 0
+    t.integer "hide_authors"
+    t.integer "thumbnail_size"
```
mo-nathan, line 931: What caused these changes?  I'm guessing these aren't intentional.

nimmolo: Nope they sure aren't. No idea... will revert.

nimmolo: Reverted.

nimmolo:
Removed Rails automatic serialization from the `:description` column, and went with calling `to_json` before storing the description, so we can still access the string as a `String` for comparisons.

PR #2718, opened 2025-02-13
bundle update for 2025-02-13

mo-nathan:
Includes "Bump net-imap from 0.5.5 to 0.5.6".

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -65,7 +65,7 @@ GEM
     ansi (1.5.0)
     arel-helpers (2.16.0)
       activerecord (>= 3.1.0, < 8.1)
-    arel_extensions (2.3.2)
+    arel_extensions (2.3.3)
       activerecord (>= 6.0)
     ast (2.4.2)
     autoprefixer-rails (10.4.19.0)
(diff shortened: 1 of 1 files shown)


mo-nathan:
Also picks up the recent rack vulnerability mentioned by dependabot:

remote: GitHub found 3 vulnerabilities on MushroomObserver/mushroom-observer's default branch (3 moderate). To find out more, visit:
remote:      https://github.com/MushroomObserver/mushroom-observer/security/dependabot

I've dismissed those three issue since one is related to Bootstrap that we're (slowly) working on, another is the net-imap, and the third is for rack < 3.1.10.

JoeCohen (approved):
CI is failing because [Coveralls has been down since about Feb 12, 2025 - 12:54 PST](https://status.coveralls.io/?utm_source=embed).

We could make CI pass although Coveralls is down. But I'm not sure we should do that.

The fix is: In `.github/workflows/ci_rails.yml` add `fail-on-error: false` to the Coveralls step. See https://status.coveralls.io/?utm_source=embed. 
This would "avoid CI failure when upload fails **_due to any errors_**."  (Emphasis added.)
That strikes me as heavy handed for our situation (but could make sense for other organizations).

PR #2720, opened 2025-02-16
Suggested changes. Untested!

nimmolo:
Hi Joe - 
I simplified the partial by removing unused data attributes, but of course feel free to re-add the data-targets if you will want Stimulus to do anything with them separately. (Didn't seem necessary, though.)


JoeCohen:
@nimmolo: Unfortunately the Stimulus controller is still not disconnecting when the job is done. See below.
Nonetheless -- because it's still an improvement over main -- I will merge your PR #2720 into #2640, 
then merge #2640 into main and deploy.

After the job is done, I get this:
```erb
<div id="status_698" data-controller="inat-import-job" data-endpoint="/inat_imports/1/job_trackers/698" data-stimulus="inat-import-job-connected">
  <div id="current_status_698" data-inat-import-job-target="current" data-status="Done">
```

PR #2702, opened 2025-02-05
InatImportJobTracker auto updates

JoeCohen:
- Uses Stimulus/Turbo to update remaining changeable elements of job tracker show page.
- Delivers #2460

### Suggested Manual Test
- in iNat,  create a bare-bones test observation: something with a fungal name, no photos needed.
- Make sure SolidQueue is running in your local server, and Import the test obs to MO
Expected Result is like this:
> ### **iNat Import Tracker**
>
> **Status**: Done
**Imported**: 1 of 1 observations
>
> **Started**: 2025-02-08 10:23:35 -0800
**Elapsed Time**: 0:00:10
**Ended**: 2025-02-08 10:23:45 -0800

- Clean up
  - **IMPORTANT**:  Clean up or Delete the test iNat obs. To clean up:
    - Go to the iNat obs, 
    - Scroll down until Observation Fields are visible.
    - click on "Mushroom Observer URL:" (the title, not the url)
    - Delete 
  - Delete the MO Observation if you want to repeat the test

PR #2723, opened 2025-02-18
Terminate Field Slip Status Requests

JoeCohen:
- Fixes how the js controller stops sending GET requests when the job is done.
- Contains only one commit, authored by @nimmo, (cherry picked from commit 70155d93435ed052a9a4baf77f554d60a84113ba) and effectively moved from #2702 so that it can be considered separately.

### DIFF
--- a/app/javascript/controllers/field-slip-job_controller.js
+++ b/app/javascript/controllers/field-slip-job_controller.js
@@ -36,8 +36,6 @@ export default class extends Controller {
   }
 
   // Every second, send a get request to find out the status of the PDF.
-  // NOTE: Can't call a class function from `setInterval` because it resets
-  // the context of `this`
   start_timer_sending_requests() {
     if (this.status_id != "3") {
       // Set the intervalId to the interval so we can clear it later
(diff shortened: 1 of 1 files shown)


mo-nathan:
Looks reasonable to me.  Wonder if we have any way to test it.

JoeCohen:
@nimmolo: Like @mo-nathan said [here](https://github.com/MushroomObserver/mushroom-observer/pull/2702#issuecomment-2665562641), I also don't understand what this fixes in Field Slips.
In `main` it stops sending  `GET "/field_slip_job_trackers"` requests as soon as the Field Slip pdf is complete.

nimmolo:
Sounds like this PR is not necessary. There must be something different about the Stimulus that I didn't notice when comparing to iNatImport Stimulus.

Oh i know. It's the difference between update and replace.

nimmolo:
Actually I do think this PR is still worth it, for two reasons.
- It removes an inaccurate comment about calling `this` within `setInterval`
- It contains an improved code pattern (for use in future templates) than the current version. Currently the "pdf finished" code clears the interval and removes the name of the controller from the  "data-stimulus" attribute, but does not actually disconnect the controller; that's handled in the `disconnect` function. I think the "pdf finished" code should simply `clearInterval`, this will create less confusion for future coders, including me!

JoeCohen:
@nimmolo. 
- Could you have another look at this. When I tested it manually, the tracker sent GET requests forever. And it appears (but probably isn't the case) that the job never finishes. (Perhaps I messed up the branch.)
Screenshot 2025-02-20 at 6 53 04 PM
- Is there a way -- other than a system test -- to test this?

JoeCohen:
Closing this PR because it does not work when tested manually.

PR #2730, opened 2025-02-20
Convert API to send ids through Query's `:ids` param

nimmolo:
API has been allowing "arrays of ranges" for its `id` param, and since this is not something currently supported or validated by Query, API has a special method that sends ready-made SQL defining the ranges as greater-than/less-than statements, via a special "back-door" where callers can send SQL to Query's `:where` param. Aside from a single use in the controllers, the API classes sending `:ids` are the only users of that back door.

To convert Query to AR I want to close this back door. I considered writing an "arrays of ranges" handler in Query, but then I realized that there seems to be no realistic use case for ranges of ids, and Query already has `:ids` params for all these classes. @JoeCohen @pellaea and @mo-nathan agree.

Rather than maintain this (as far as we know) unused flexibility, i'm changing API to send these ids through "the front door". API here accepts only flat arrays of ids for its `id` param.


nimmolo:
One improvement on this PR — the ids are looked up and validated as existing ids.

PR #2740, opened 2025-02-22
RuboCop dual config

JoeCohen:
- Adds 2nd RuboCop config file, used only by CodeClimate, which loads plugins with `require`
- Tells CodeClimate GitHub App to use that file to configure RuboCop
- Changes local (default) RuboCop config file to load extensions with `plugins`
- Fixes CodeClimate CI error which was caused by using `plugins`. Ex: https://codeclimate.com/github/MushroomObserver/mushroom-observer/builds/11148
- Result:
  - We can use RuboCop >= 1.72 locally
  - CodeClimate CI works


JoeCohen:
@mo-nathan & @nimmolo 
This is an alternative to freezing the RuboCop version.
I'd mostly like your opinion whether it's the better alternative.

JoeCohen:
Advantage:
- We can use RuboCop >= 1.72, so we get bug fixes as well as newer cops.
Disadvantage:
- The local Rubocop version is different from the Codeclimate version.

JoeCohen:
Converted back to draft. I wasn't paying attention to the Gem files/

JoeCohen:
I added:

- an Issue to CodeClimate https://github.com/codeclimate/codeclimate-rubocop/issues/376
- an Issue to RuboCop https://github.com/rubocop/rubocop/issues/13934

I also contact CodeClimate directly via their support form.

JoeCohen:
Closing. Replaced by #2828.

PR #2741, opened 2025-02-22
Scope & param naming `with_` to `has_` 

nimmolo:
>  There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

yes, Martin Fowler: this PR undoes a previous Nimmo renaming. I'm going to find the older PR and read the reasons for doing it in the first place. _Update: no overwhelming reasons, in retrospect._

Reasons to revert are not imperative either, but:
- both the API and PatternSearch use the `:has_thing` naming convention, and they are much harder to rename. Reverting the Query params and scopes to similar naming means less mental overhead.
- `:has_` makes just as much sense and is a little shorter. 
- i believe it will simplify the Query work. If i can get the names in line, there's an opportunity to slim down query by having most params hand off directly to scopes of the same name.  

The general goal is "less to think about" and in the short term, this goes maybe counter to the goal, for anyone who's been using the current scopes. But in the long term, having these names conform to each other could make params less confusing to remember.

#### Boolean-ish scopes:
Most of these Query params and scopes take Boolean-ish values to modify the meaning. For example, `has_classification(false)` means `has_no_classification`. So we don't actually need the `without/has_no_` scopes.

Question:
- Is it convenient to have a `has_no_classification`scope?
- Or would it be less confusing to eliminate the negative scopes, and have all callers use **the same signature as the Query param:** `has_classification: false`/`has_classification(false)`. This might help make Query syntax feel familiar, which is one of the main goals of the Query refactor.

> [!IMPORTANT]
> Need to call script `update_user_content_filter` on the server if deploying.


JoeCohen (commented):
I favor less mental load.
But which will create less mental load in the long run:
- Scope names resemble Query params.
- Scope names resemble AR Queries.

IDK, and am fine with whatever you decide. Maybe in the real long run it doesn't matter.

I prefer eliminating the negative scopes.

nimmolo:
Done and script run

PR #2747, opened 2025-02-23
Reduce Inat Import emails

### LINKED ISSUE #2673: Reduce iNat Import emails
Reduce or silence emails sent to the importing User during imports. 
Import currently generates at least one email/Observation to the importing user. 
The Import generates some objects associated with the Obs. The Create callback (e.g. for a Comment) will generate an email if the object has a different owner than the Observation.

~~The proposed solution: (1) have those objects generated by the user, and (2) if they should be protected change by the user (e.g., the Snapshot Comment), immediately change the owner directly in the db. (in order to avoid callbacks) to webmaster~~

After discussion with Nathan, decided to simply make the importing user the creator of the Snapshot.

### Tasks
- [x] Test no new emails generated (will fail).
- [ ] Test Snapshot comment owned by importing user.
- [ ] Manual test.  (Might need to be live. Do emails get generated in dev mode?)
  - [ ] Create test iNat Obs for jdc
  - [ ] As OMS, propose an id
  - [ ] Import it.
- [ ] Draft PR
- [ ] RuboCop
- [ ] Ready PR

JoeCohen:
Sends fewer emails to a user who imports an Inat observation.
- Does not send an email about the Snapshot Comment (because that Comment is now created by the user instead of the iNat Import Manager/Webmaster).

### Suggested Manual Test
- In iNat, create a bare-bones test observation with a fungal Identification. (Ex: https://www.inaturalist.org/observations/253424654. You don't need a photo or a sequence.)
- empty `log/email-debug.log`
- in MO, import the test iNat observation you just created.
- when the import is done, search `log/email-debug.log` for `QueuedEmail::CommentAdd from=MO Webmaster to`
Expected result: There should be no occurrences of that string.


mo-nathan (approved):
Looks good to me.

PR #2749, opened 2025-02-24
Script to update query records

nimmolo:
This PR also does a bit of refactoring and cleanup of Query.

Running the new `script/update_query_records.rb "true"` on production will do a dry run checking for query records that have the obsolete filter param keys `with_images` and `with_specimen`. Pass "false" to actually update the records.

None of the other changed `with_` query params seem to be present on production, only `with_images`. `with_specimen` is just there for safety.

Alternatively we could run a regex to replace any `with_` param in the description string, but it doesn't seem to be necessary.

### DIFF
--- a/script/update_query_records.rb
+++ b/script/update_query_records.rb
@@ -49,7 +49,7 @@ class UpdateQueryRecords
       if hsh.key?(:with_specimen)
         hsh[:has_specimen] = hsh.delete(:with_specimen)
       end
-      entries << { id: query_record_id, description: Query.serialize(hsh) }
+      entries << { id: query_record_id, description: hsh.to_json }
     end
     entries
   end
--- a/script/update_query_records.rb
+++ b/script/update_query_records.rb
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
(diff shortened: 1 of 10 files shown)


script/update_query_records.rb
```diff
@@ -0,0 +1,59 @@
...
+    to match the new keys in the `Query::Filter` class.
+
+  PARAMETERS::
+
+    --help     Print this message.
+
+HELP
+
+class UpdateQueryRecords
+  def self.update_query_records(dry_run)
```
JoeCohen, line 31:
```suggestion
  def self.update_query_records(dry_run:)
```
Consider using a kwarg for consistency with other methods. Examples:
`UserStats#refresh_all_user_stats`,  `Synonym::make_sure_all_referenced_synonyms_exist`, etc.
Cf. https://docs.rubocop.org/rubocop/cops_style.html#styleoptionalbooleanparameter

JoeCohen (commented):
Sorry for the belated suggestion. 
Also is it worth writing a test for `update_query_records`? IDK

nimmolo:
@JoeCohen Thanks! 
I took these suggestions and made a couple more important alterations.

PR #2753, opened 2025-02-25
Fix title for filtered query when no cached instance found (undefined Location)

nimmolo:


nimmolo:
Merging hot fix because it throws an error on production.

Issue #2752, opened 2025-02-25
Undefined Location link


JoeCohen:
Clicking an Undefined Location (on the Location  Index page) throws an Error
Ex (using `database-20250223.gz`)

#### NoMethodError in ObservationsController#index
undefined method `title_display_name' for nil

```ruby
Extracted source (around line #77):
              
  def set_at_location_title
    location = find_cached_parameter_instance(Location, :location)
    @title_args[:location] = location.title_display_name
    location
  end


Rails.root: /Users/joe/mushroom-observer

Application Trace | Framework Trace | Full Trace
app/classes/query/modules/associations.rb:77:in `set_at_location_title'
app/classes/query/modules/associations.rb:72:in `initialize_locations_parameter'
app/classes/query/observations.rb:98:in `initialize_association_parameters'
app/classes/query/observations.rb:71:in `initialize_flavor'
app/classes/query/modules/initialization.rb:21:in `initialize_query'
app/classes/query/modules/low_level_queries.rb:29:in `select_values'
app/classes/query/modules/high_level_queries.rb:56:in `result_ids'
app/classes/query/modules/high_level_queries.rb:36:in `num_results'
app/controllers/application_controller/indexes.rb:241:in `set_index_view_ivars'
app/controllers/application_controller/indexes.rb:227:in `show_index_setup'
app/controllers/application_controller/indexes.rb:211:in `show_index_of_objects'
app/controllers/application_controller/indexes.rb:139:in `filtered_index'
app/controllers/application_controller/indexes.rb:50:in `block in build_index_with_query'
app/controllers/application_controller/indexes.rb:39:in `each'
app/controllers/application_controller/indexes.rb:39:in `build_index_with_query'
app/controllers/observations_controller/index.rb:7:in `index'
app/controllers/application_controller.rb:218:in `catch_errors_and_log_request_stats'
```
#### Request
Parameters:
```ruby
{"where"=>"Oakfield, Halifax, Nova Scotia, Canada"}
```

nimmolo:
Got it. I think i can fix

nimmolo:
Maybe fixed by #2753. Needs a test though.

JoeCohen:
Delivered: #2753, #2756

PR #2756, opened 2025-02-26
Test Index Undefined Location

JoeCohen:
Adds a test to prevent reversion of Issue #2752
Also see PR #2753 (a hotfix for the issue)


JoeCohen:
@nimmo: This is simple, but I think it's sufficient to prevent reversion.
(I ran it on the deploy just prior to the hotfix, and it Errors there.)

nimmolo (approved):

nimmolo:
Thanks for doing this. Will work as long as we don't add an "Oakfield, Halifax, Nova Scotia, Canada" fixture i guess!

PR #2766, opened 2025-03-02
Various Project Alias fixes

mo-nathan:
- Filter AliasController#index by the associated project and order by name
- Better error handling for #create
- Actions work correctly from #index as well as member and locations tab
- Avoid #show in favor of #index for ProjectAliases
- Eager load target for ProjectAlias
- Use destroy_button rather than link_to
- Add more ProjectAlias validations
- Turbo update for #index
- Use format_name for #show
- Fix tests


app/views/controllers/projects/aliases/show.html.erb
```diff
@@ -16,7 +16,7 @@
 
 <p>
   <strong>Target:</strong>
```
JoeCohen, line 18: If you want to use our stylesheets instead of html tag:
```suggestion
  <span class="font-weight-bold">Target:</span>
```

mo-nathan: Done.  Hadn't updated that template.  It's not terribly relevant any more, but I updated it just the same.  I did notice that we have a bunch of other `<strong>`s in the code base, but that's clearly outside the scope of this PR.

JoeCohen:
As you've seen, alphabetic order puts `Walk 10` before `Walk 2`. I don't know an easy way to improve that, and anything harder probably isn't worth the trouble.

JoeCohen (approved):
Very nice! 
I banged on it locally, and everything looked fine.
(I still don't understand Turbo & Stimulus. Maybe I'll catch up one day.)

mo-nathan:
> As you've seen, alphabetic order puts `Walk 10` before `Walk 2`. I don't know an easy way to improve that, and anything harder probably isn't worth the trouble.

Yeah, not interested in changing that.  I think it will be obvious enough to folks since it's a common problem with sorting in English (and many other languages as far as I know).  If someone really cares for their project, they can always add 0 padding.

mo-nathan:
> Very nice! I banged on it locally, and everything looked fine. (I still don't understand Turbo & Stimulus. Maybe I'll catch up one day.)

Thanks!  Yeah, I think I get at least the basics of Turbo.  I still have a question or two, but I feel like I can use it for new stuff.  Stimulus is still a mystery to me, but since I avoid doing complicated things in JS it hasn't been a real issue for me yet.

nimmolo (approved):

Issue #2767, opened 2025-03-02
Revise Provisional Names


mo-nathan:
There is a new consensus for how to handle provisional names.  As I understand it we should be displaying provisional names like this:

_Pleurotus sp. 'pulmonarius-PNW02'_

We need to switch to this format for the output of all provisional names.  For input (search, naming, name creation) we want to be more permissive and support not only that format, but also various common alternative formats.  Desired input support includes:

Not having the "sp. ", e.g. `Pleurotus 'pulmonarius-PNW02'`
Using double quotes instead of single quotes: e.g. `Pleurotus sp. "pulmonarius-PNW02"`
Using no quotes: e.g., `Pleurotus pulmonarius-PNW02`
Spaces instead of "-"s: e.g., `Pleurotus sp. 'pulmonarius PNW02'`
Having "sp-" embedded in the provisional name, e.g., `Pleurotus 'sp-pulmonarius-PNW02'`
Having no epithet: e.g., `Pleurotus sp. 'PNW02'`
And all reasonable combinations of the above.

The two cases that I'm aware of that create potential ambiguity are dropping the epithet, e.g., `Pleurotus 'PNW02'`, and combining spaces with no quotes, e.g., `Pleurotus pulmonarius PNW02`.

In the case of dropped epithets, the input string could match multiple provisional names, e.g., _Pleurotus sp. 'pulmonarius-PNW02'_, _Pleurotus sp. 'ostreatus-PNW02'_ , or _Pleurotus 'PNW02'_.  For searching all potentially matching names will be used.  For cases where exactly one name is expected such as apply a name, then if there is a version with no epithet, it wins.  However, if there is ambiguity with multiple epithets, then an error will be generated that will list the different options.

In the case of spaces with no quotes, the issue is ambiguity with author citations.  For all cases, if there is one or more match in the database, then search will show all options.  For naming, if there is exactly one match, then it will be used.  If there are no matches and a name is bing created, then it will be interpreted as an author citation.  For example, if there is no existing match for _Pleurotus sp. 'PNW02'_ and a user creates a new observation with the name string `Pleurotus PNW02`, then a new genus name will be created with the name _Pleurotus_ and the author PNW02.  Hopefully, cases like this will be very rare and they are easily avoided by simply adding quotes or `sp.` when the name is entered.

JoeCohen:
👍 I agree with all of that.

Some edge-y cases that may present special issues:
[...]

PR #2778, opened 2025-03-07
Add fix_names.rb script

mo-nathan:
Adds script for ensuring that the various name strings in a Name objects are self consistent and consistent with our standard parsing rules.  This should be run against our production database to tidy up around 450 issues discovered as part of digging into supporting a new format for provisional names.  This script will also be used to support that transition once `Name.parse_name` is updated to support the new format.

### DIFF
--- a/app/models/name/notify.rb
+++ b/app/models/name/notify.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
 module Name::Notify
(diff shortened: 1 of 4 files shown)


mo-nathan:
Related to #2767

mo-nathan:
Coveralls lies.

PR #2797, opened 2025-03-14
Simpler index titles, with filter captions

nimmolo:
The aim of this PR is to remove index titling from Query's responsibilities, and put it in `ApplicationController`'s `TitleHelper`.

- It refactors the `#title` code generating index titles to simply print the name of the model being queried, in all cases, e.g. "Projects". 
- On the page just below the title, it composes a `#filters` caption describing the filters applied to the index, taken from `@query`. It tries to make the filters comprehensible without investing too much in clever semantic logic that requires translation: each param name gets a translation string, and when the values are ids, we look up titles for the corresponding records.
- The filter values are truncated if there are more than three values at any param, or if the param value string is longer than 100 chars. If the params have been truncated, it renders a toggle, which reveals the full list of param values, via Stimulus.


JoeCohen:
I started to bang on this. I like the the UI.
I did this:
`http://localhost:3000/observations?pattern=xerocomellus+region%3A%22Oregon%5C%2C+USA%22`
Nice
Clicked "Show Names"
Got the following title. Is it what you intended:
> ### Names
>
> of observations: {
regions: Oregon, USA
names: xerocomellus, with synonyms, with subtaxa
}

nimmolo:
@JoeCohen Yes — I'm trying to make it clear that the filters are on the observations, not the names.

We could run it on one line, but it could go on and on, and the above seemed clearer.

> of observations: { regions: Oregon, USA, names: xerocomellus, with synonyms, with subtaxa }

JoeCohen:
Thanks. Makes sense. I just wanted to make sure it's what you meant.

nimmolo:
Coverage is good here, as far as I can tell. Coveralls is flagging old issues.

app/classes/api2/herbarium_record_api.rb
```diff
@@ -40,11 +40,11 @@ def query_params
         herbaria: parse_array(:herbarium, :herbarium, as: :id),
         observations: parse_array(:observation, :observation, as: :id),
         has_notes: parse(:boolean, :has_notes),
-        initial_dets: parse(:string, :initial_det, help: 1),
-        accession_numbers: parse(:string, :accession_number, help: 1),
+        initial_det: parse(:string, :initial_det, help: 1),
+        accession: parse(:string, :accession_number, help: 1),
```
mo-nathan, line 44: Is this changing the parameters for the API?  Looking over the last couple weeks of logs, I do see some references to these these fields.  I'm not sure yet if they are due to API calls, but in general changing parameters without providing backwards compatibility or changing the version number of the API is a bad practice.

mo-nathan: Did a bit more testing and I think the API parameters for GET and POST are not changed.  See test/classes/api2/herbarium_records_test.rb.  However, it does seem weird that the GET returns:
            "initial_determination": "Tricholoma arvernense",
but the POST seems to want:
      initial_det: @initial_det,

However, that's out of scope for this PR.

app/helpers/title_helper.rb
```diff
@@ -0,0 +1,266 @@
...
+      tag.h5(naming, id: "owner_naming")
+    end
+  end
+
+  # contents of the <title> in html <head>
+  def title_tag_contents(title, action: controller.action_name)
+    if title.present?
+      title.strip_html.unescape_html # removes tags and special chars
+    elsif TranslationString.where(tag: "title_for_#{action}").present?
+      :"title_for_#{action}".t
```
mo-nathan, line 40: Not covered.

app/helpers/title_helper.rb
```diff
@@ -0,0 +1,266 @@
...
+    lookup_class = "Lookup::#{PARAM_LOOKUPS[param]}".constantize
+    joined_vals = lookup_class.new(lookups).titles.join(", ")
+    return joined_vals unless truncate
+
+    truncate_joined_string(joined_vals, ids)
+  end
+
+  def truncate_joined_string(joined_vals, ids)
+    if joined_vals.length > 100
+      joined_vals = "#{joined_vals[0...97]}..."
```
mo-nathan, line 260: Not covered.

app/helpers/title_helper.rb
```diff
@@ -0,0 +1,266 @@
...
+    elsif ids.length > CAPTION_TRUNCATE
+      joined_vals += ", ..."
```
mo-nathan, line 262: Not covered.

app/helpers/title_tab_set_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+    kwargs[:class] = class_names(kwargs[:class], extra_args[:class])
+    # merge in other args from extra_args (will overwrite keys!)
+    kwargs&.merge(extra_args&.except(:class))
+  end
+
+  # New style dropdown tabsets take array of tabs as hash of args,
+  #   { name:, link:, class:, id:, etc. }
+  #   not fully-formed `link_to` or `link_with_query`
+  def add_dropdown_tab_set(tabs:, title: :LINKS.t)
+    content_for(:dropdown_tab_set) do
```
mo-nathan, line 93: Not covered.

app/helpers/title_tab_set_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+      render(partial: "application/content/dropdown_tab_set",
+             locals: { title: title, links: create_dropdown_links(tabs) })
+    end
+  end
+
+  def create_dropdown_links(tabs)
+    extra_args = {
```
mo-nathan, line 100: Not covered.

app/helpers/title_tab_set_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+      role: "menuitem",
+      class: "dropdown-item"
+    }
+    make_tab_links(tabs, extra_args)
+  end
+
+  def dropdown_link_options(args = {})
+    args&.except(:name, :link, :button, :class) # prolly delete name and link
```
mo-nathan, line 108: Not covered.

app/models/herbarium_record.rb
```diff
@@ -72,16 +72,16 @@ class HerbariumRecord < AbstractModel
   scope :notes_has,
         ->(str) { search_columns(HerbariumRecord[:notes], str) }
 
-  scope :initial_dets, lambda { |val|
+  scope :initial_det, lambda { |val|
     exact_match_condition(HerbariumRecord[:initial_det], val)
   }
   scope :initial_det_has,
         ->(str) { search_columns(HerbariumRecord[:initial_det], str) }
```
mo-nathan, line 79: Not part of PR, but interesting that this isn't covered. Haven't checked on main.

app/helpers/title_tab_set_helper.rb
mo-nathan: Also not covered on main.

app/helpers/title_tab_set_helper.rb
mo-nathan: Also not covered on main.

app/models/herbarium_record.rb
mo-nathan: Not covered on main.

app/helpers/title_helper.rb
nimmolo: Ah! Thanks

app/helpers/title_helper.rb
nimmolo: Ok, yes. Will need to send a long query

app/helpers/title_tab_set_helper.rb
nimmolo: These two are my Bootstrap 4 helpers that have been waiting patiently for me for 4 years. 

I wouldn't call them dead code - they lived for a year or so on the BS4 branch, but they haven't been reincarnated yet.

app/models/herbarium_record.rb
nimmolo: Yeah, this scope is of questionable utility, but its analogue is offered on the API. I can add a test there.

app/helpers/title_tab_set_helper.rb
nimmolo: The comment notes they were made using my "new" tab pattern, lol!

app/models/herbarium_record.rb
nimmolo: This scope is already tested in `tests/query/herbarium_records_test.rb` right here:
```ruby
  def test_herbarium_record_initial_det
    expects = [herbarium_records(:field_museum_record)]
    scope = HerbariumRecord.index_order.initial_det("Lichen")
    assert_query_scope(expects, scope, :HerbariumRecord, initial_det: "Lichen")
    assert_query_scope(expects, scope, :HerbariumRecord, initial_det: "lichen")
  end
```
It seems unlikely that Coveralls is wrong, but in this case lo and behold...?

app/helpers/title_helper.rb
nimmolo: Just want to say that writing this test improved the helper. I had forgotten to add the ellipsis after truncated text vals (looked-up text results already get it) and writing the expectation made me notice.

app/helpers/title_helper.rb
nimmolo: Added test for this too

app/models/herbarium_record.rb
nimmolo: I added a second test that covers this scope, in `test/api2`

app/helpers/title_helper.rb
nimmolo: This was dead code. Removed, along with translation strings, a fixture and tests.

app/helpers/title_tab_set_helper.rb
nimmolo: After review, saying goodbye to these. They're simple enough to rebuild.

Will refactor tab_set/context_nav all at once when the time comes.

nimmolo:
Also fixes an ancient issue i inadvertently caused where the click area for "tab" links went far beyond the link.

PR #2798, opened 2025-03-14
Spiders Begone!

mo-nathan:
Try going to a page greater than 10 in an observation index page while logged out.

### DIFF
--- a/app/controllers/application_controller/indexes.rb
+++ b/app/controllers/application_controller/indexes.rb
@@ -57,6 +57,15 @@ module ApplicationController::Indexes # rubocop:disable Metrics/ModuleLength
     filtered_index(new_query, display_opts)
   end
 
+  def check_for_spider_block
+    trace_tests
+    return unless !User.current && params[:page].to_i > 10
+
+    Rails.logger.warn(:runtime_spiders_begone.t)
+    render(json: :runtime_spiders_begone.t,
+           status: :forbidden)
+  end
+
   # It's not always the controller_name, e.g. ContributorsController -> User
   def controller_model_name
     controller_name.classify
--- a/app/controllers/application_controller/indexes.rb
+++ b/app/controllers/application_controller/indexes.rb
@@ -58,7 +58,6 @@ module ApplicationController::Indexes # rubocop:disable Metrics/ModuleLength
   end
 
   def check_for_spider_block
-    trace_tests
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller/indexes.rb
```diff
@@ -57,6 +57,15 @@ def build_index_with_query
     filtered_index(new_query, display_opts)
   end
 
+  def check_for_spider_block
+    trace_tests
+    return unless !User.current && params[:page].to_i > 10
+
+    Rails.logger.warn(:runtime_spiders_begone.t)
+    render(json: :runtime_spiders_begone.t,
+           status: :forbidden)
```
mo-nathan, line 66: I was tempted to use 418 (I’m a teapot) rather than 403 (Forbidden), but I resisted the urge…

app/controllers/application_controller/indexes.rb
```diff
@@ -57,6 +57,15 @@ def build_index_with_query
     filtered_index(new_query, display_opts)
   end
 
+  def check_for_spider_block
+    trace_tests
```
JoeCohen, line 61: Regardless whether I am logged in, `http://localhost:3000/` throws
```ruby
NameError in ObservationsController#index
undefined local variable or method `trace_tests' for an instance of ObservationsController
Did you mean?
track_translations

  def check_for_spider_block
    trace_tests
    return unless !User.current && params[:page].to_i > 10

    Rails.logger.warn(:runtime_spiders_begone.t)

app/controllers/application_controller/indexes.rb:61:in `check_for_spider_block'
app/controllers/observations_controller/index.rb:7:in `index'
app/controllers/application_controller.rb:218:in `catch_errors_and_log_request_stats'
Request
Parameters:
```

JoeCohen: PS: I cannot parse the comment before `test/test_helper.rb#trace_tests`:
```ruby
# Function for creating a log the tests called that
# somehow call this function.
```

mo-nathan: Oh, I removed the reference to it.

JoeCohen (approved):

app/controllers/application_controller/indexes.rb
nimmolo: I'm in favor of 418, feel free :)

PR #2801, opened 2025-03-16
Create default for description source_type in link builder

nimmolo:
On my db snapshot there are descriptions where `source_type` is `nil`. 

`nil` will cause an error in the link builder, where it can't find a translation string. 
Screen Shot 2025-03-16 at 1 24 36 AM

When I checked the live site, the source type for the example [117805](https://mushroomobserver.org/117805) was `public`. 
Screen Shot 2025-03-16 at 1 26 20 AM
Maybe there's a script that updates these, but to be safe I'm proposing giving the link builder a default. Public?


JoeCohen:
Will try to look at this later today. I usually get confused with `source_type:` and `public:`, writeability, and which one is the default.
I'm also confused about your snapshot. 
In the live db:
```ruby
mushroom-observer(prod)> NameDescription.where(source_type: nil).count
=> 0
```
and the same for `database-20250315.gz` snapshot
```ruby
mushroom-observer(dev)> NameDescription.where(source_type: nil).count
  NameDescription Count (5.4ms)  SELECT COUNT(*) FROM `name_descriptions` WHERE `name_descriptions`.`source_type` IS NULL
=> 0
```

JoeCohen (approved):
Looks good to me.
Thanks for finding and fixing the issue.

test/models/description_test.rb
```diff
@@ -204,9 +204,9 @@ def test_user_sourced_description_with_unknown_user
     desc = name_descriptions(:peltigera_user_desc)
     desc.update(user_id: nil)
 
-    assert(desc.text_name.start_with?("?'s "),
+    assert(desc.text_name.start_with?("Public"),
            "text_name of user sourced Description with unknown user should " \
-           "start_with \"?'s\" ")
+           "start_with Public ")
```
nimmolo, line 209: @JoeCohen Does this test express what we wanted to happen, or just what we thought would happen? 

Note that it fails on this branch because if the user is unknown, it prints "Public".

JoeCohen: IDK. Maybe @mo-nathan has a better idea.

PR #2813, opened 2025-03-17
Exclude misspellings if synonyms are excluded.

mo-nathan:


test/controllers/observations_controller/observations_controller_show_test.rb
```diff
@@ -688,7 +688,7 @@ def test_prev_and_next_observation_with_fancy_query
       :Observation, names: { lookup: n2.id, include_synonyms: false },
                     by: :name
     )
-    assert_equal(2, query.num_results)
+    assert_equal(1, query.num_results)
```
nimmolo, line 691: The comment above this line should be maybe altered. I believe it's indicating that a misspelling should get you the correct spelling and the misspelling.

nimmolo (changes requested):

test/controllers/observations_controller/observations_controller_show_test.rb
mo-nathan: Done.

JoeCohen (approved):
Side issue: There are ~10 misspellings whose synonym is nil.
Is that a problem? 
Should I fix it manually in the live db?
```ruby
mushroom-observer(dev)> Name.where.not(correct_spelling_id: nil).where(synonym_id: nil).count
  Name Count (68.2ms)  SELECT COUNT(*) FROM `names` WHERE (`names`.`correct_spelling_id` IS NOT NULL) AND (`names`.`synonym_id` IS NULL)
=> 10
```

mo-nathan:
I don't think that's a problem.  Here's an example, https://mushroomobserver.org/names/1560.  There would be no reason for there to be a synonym in that case.  I assume the others are similar.

PR #2815, opened 2025-03-18
Rename extensions files to agree with class names

nimmolo:
This should fix the Zeitwerk error, if I understand it correctly.

File names must agree strictly with the classes they define.


nimmolo:
Codeclimate issues in extensions are longstanding, that's a separate PR.

PR #2819, opened 2025-03-18
Add parens to remaining `require` and `require_relative` calls

nimmolo:
On MO, there are currently 
- 448 total uses of `require("file")` and `require_relative("path/to/file")` with parentheses
- 56 total uses `require "file"` and `require_relative "path/to/file"` without parentheses 

This PR standardizes these around the parentheses usage.


JoeCohen:
I'm fine with this if it's something you & @mo-nathan like.
Note, however, that the RuboCop (and Rails 8) default is that macros do not require (but are allowed to have) parens around arguments. I.e., they are leaving it up to each developer.
There's no way to force macros to omit parens around arguments.

JoeCohen (commented):
I'm OK with this (despite prefering to omit parens) if it's something you & @mo-nathan like.
Note, however, that the RuboCop (and Rails 8) default is that macros do not require (but are allowed to have) parens around arguments. I.e., Rails and RuboCop leave it up to each individual developer.

nimmolo:
@JoeCohen OK - Makes sense. So `require` is a "macro" and not a "method"?  

For the sake of consistency, I'll merge though.  Having them all one way or the other makes it a bit easier to find the requires. It's easy enough to change us to the other style via regex search in VSCode if we reconsider later.

JoeCohen:
Based on RuboCop's behavior `require` is a macro. (Otherwise it would have complained about missing parens.)
Good point about consistency making it easier to find things.

PR #2823, opened 2025-03-20
where...proposed

### LINKED ISSUE #2822: About this taxon, xxx proposed
The "xxx proposed" counts/links in the About this taxon panel of the Name page are still wrong.
Ex: http://localhost:3000/names/63203 (_Grandinia curvispora_) shows 
> other taxa, this taxon proposed (1)
any taxon, this name proposed (1)

This Name has never been proposed:
```ruby
mushroom-observer(dev)> name = Name.find(63203)
...
mushroom-observer(dev)> Naming.where(name: name).count
  Naming Count (124.3ms)  SELECT COUNT(*) FROM `namings` WHERE `namings`.`name_id` = 63203
=> 0
``` 

And there are no Observations of other taxa where this taxon was proposed.
```ruby
mushroom-observer(dev)> name = Name.find(63203)
...
mushroom-observer(dev)> synonyms = name.synonyms
...
mushroom-observer(dev)> Observation.joins(:namings).where.not(name: synonyms).where(namings: { name: synonyms } )
  Observation Load (154.1ms)  SELECT `observations`.* FROM `observations` INNER JOIN `namings` ON `namings`.`observation_id` = `observations`.`id` WHERE (`observations`.`name_id` NOT IN (63203, 51158, 63201, 63202)) AND (`namings`.`name_id` IN (63203, 51158, 63201, 63202)) /* loading for pp */ LIMIT 11
=> []
```

JoeCohen:
- Fixes bugs in Name page, About this taxon panel:
  - other taxa, this taxon proposed 
  - any taxon, this name proposed
- Delivers #2822

### Manual Test
Compare
- webserver https://mushroomobserver.org/names/63203, 
(incorrect counts, and links get `No matching observations found.`
> other taxa, this taxon proposed (1)
any taxon, this name proposed (1)

with
-  this branch http://localhost:3000/names/63203 
(correct counts, without links)
> other taxa, this taxon proposed (0)
any taxon, this name proposed (0)

### DIFF
--- a/app/models/name/observations.rb
+++ b/app/models/name/observations.rb
@@ -38,12 +38,12 @@ class Name
 
     # "Observations of other taxa, this taxon proposed"
     def where_taxon_proposed
-      @all.reject { |obs| obs&.name_id == @name.id }
+      @all.reject { |obs| obs&.name_id.in?(@name_ids) }
     end
 
     # "Observations of any taxon, this name proposed"
     def where_name_proposed
-      @all
+      @all.where(namings: { name_id: @name.id })
     end
 
     def has_images # rubocop:disable Naming/PredicateName
--- a/test/models/name_test.rb
+++ b/test/models/name_test.rb
@@ -3728,6 +3728,35 @@ class NameTest < UnitTestCase
   end
 
   # ----------------------------------------------------
+  #  Name:Observations
+
+  def test_where_proposed
+    name = names(:lactarius_alpigenes)
+    assert(name.deprecated?, "Test needs deprecated Name")
+    assert_blank(Naming.where(name: name),
+                 "Test needs name without Namings")
+    assert_blank(Observation.where(name: name),
+                 "Test needs name without Observations")
+    approved_name = name.approved_name
+    # Test must have an Obs of a synonym
+    obs = Observation.create(name: approved_name, user: rolf)
+    # and synonym must be proposed
+    Naming.create(observation: obs, name: obs.name, user: obs.user)
+
+    obss = Name::Observations.new(name)
+
+    assert_blank(
+      obss.where_name_proposed,
+      "`where_name_proposed` should be blank if only a synonym was proposed"
+    )
+    assert_blank(
+      obss.where_taxon_proposed,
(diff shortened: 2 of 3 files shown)


app/models/name/observations.rb
```diff
@@ -38,12 +38,12 @@ def of_taxon_any_name
...
     # "Observations of other taxa, this taxon proposed"
     def where_taxon_proposed
-      @all.reject { |obs| obs&.name_id == @name.id }
+      @all.reject { |obs| obs&.name_id.in?(@name_ids) }
     end
 
     # "Observations of any taxon, this name proposed"
     def where_name_proposed
-      @all
+      @all.where(namings: { name_id: @name.id })
```
JoeCohen, line 46: @nimmolo We need the info from `namings`, not just from the obss.
I hope this will have little or no effect on performance. I think the `namings` have been eager loaded. Is that right?

test/models/name_test.rb
```diff
@@ -3727,6 +3727,35 @@ def test_more_brief_authors
     assert_equal("(A et al.) D et al.", name.send(:brief_author))
   end
 
+  # ----------------------------------------------------
+  #  Name:Observations
+
+  def test_where_proposed
```
JoeCohen, line 3733: Is there a better file, or a better location in this file, for this test?

app/models/name/observations.rb
nimmolo: That's right - check the definition of `all` above. Thanks for figuring this out!

test/models/name_test.rb
nimmolo: There may be a Name::ObservationsTest?

nimmolo: Nope. But the module tests could be broken out like that. NameTest seems ripe for a refactor. In another PR.

nimmolo: Oh there is already a name_parse test and a name_sorter test. This could be in its own file, I'd lean that way.

JoeCohen: Thanks for looking and for the advice. I moved it to its own file.

PR #2824, opened 2025-03-20
Provisional name rank designators

mo-nathan:
This is intended to be a middle step in supporting the latest provisional name format proposed by Steve Russell.  See step 2 of https://mushroomobserver.org/articles/45.  The code on this branch should support names like:

Amanita sp. "MA-01"
Gen. "Mycena-42" pura
etc.


mo-nathan:
Running fix_names.rb is revealing some issues that need to get sorted before I can take this out of draft.

app/models/name/format.rb
```diff
@@ -136,14 +136,11 @@ def brief_author
...
         gsub(/__\*?\*? [^_*]* \s (#{Name::Parse::ANY_NAME_ABBR}) \s \*?\*?__/ox,
-             ' \1 ').
-        # (this part should be unnecessary)
-        # Because "group" was removed by the 1st gsub above,
-        # tack it back on (if it was part of display_name)
-        gsub(/__\*?\*? [^_*]* \*?\*?__/x, " ").
-        concat(group_suffix(name))
+             ' \1 '). # Remove internal author
+        gsub(/\*?\*?__\*?\*?/, ""). # Remove textile ornamentation
+        concat(group_suffix(name)) # Readd group suffix
```
JoeCohen, line 143: typo "Read"

app/models/name/parse.rb
```diff
@@ -201,18 +217,22 @@ def match?(str)
   GROUP_CHUNK     = /\s (?<group_wd>#{GROUP_ABBR}) \b/x
 
   # matches to ranks that are included in the name proper
-  # subspecies is not included because it's the catchall default
-  RANK_START_MATCHER = /^(f|sect|stirps|subg|subsect|v)/i
+  RANK_START_MATCHER = /^(f|sect|stirps|subg|subsect|v|sp|ssp|subsp|s)/i
 
   # convert rank start_match to standard form of rank
   # subspecies is not included because it's the catchall default
```
JoeCohen, line 223: `# subspecies is not included because it's the catchall default` seems obsolete

app/models/name/parse.rb
```diff
@@ -201,18 +217,22 @@ def match?(str)
...
   STANDARD_SECONDARY_RANKS = {
     f: "f.",
     sect: "sect.",
+    section: "sect.",
```
JoeCohen, line 227: Do you need to also include other unabbreviated ranks as keys?

app/models/name/parse.rb
```diff
@@ -427,60 +485,59 @@ def make_sure_ranks_ordered_right!(prev_rank, next_rank)
...
-    elsif words[-3] == words[-1]
-      [
-        format_name(words[0..-3].join(" "), deprecated),
-        author,
-        words[-2],
-        format_name(words[-1], deprecated)
-      ].join(" ")
-    else
-      format_name(name, deprecated) + " " + author
+    return format_name(name, deprecated) if author.blank?
```
JoeCohen, line 503: There was an MO debate years ago whether MO autonyms should have authors. Reasonable arguments on both sides.
I think Scott Redhead said no authors and I've tried to standardize them that way. 
The vast majority of existing MO autonyms do not have authors.

Looking over the the discussion now, I think that binomial autonyms should not have authors. 
In other autonym combinations, author is optional.

JoeCohen (approved):
I think you should merge & deploy. 
As you said, it’s “just” step 1. But it’s a big step, really improves standardizing the names, and doesn’t break anything that works (except maybe [searches with quoted epithets](https://mushroomobserver.slack.com/archives/C08FRUK0Z8T/p1743350868742249)., and that’s probably something that would need changing on the search side).

JoeCohen (approved):
Reminder: You had mentioned the possibility using `unicode_normalize`.

mo-nathan:
Thanks for the reminder.  Done.

app/models/name/parse.rb
mo-nathan: That isn't the case according to the code currently in production.

AlanRockefeller: I don't think provisional names need authors.

mo-nathan:
Both CodeClimate and Coveralls are being flaky.  Merging without their approval.

app/models/name/parse.rb
mo-nathan: I think it's debatable.  I could see two different teams unintentionally using the same provisional name especially if you take into account how we're planning on normalizing the format.  Given that provisionals are even looser than published names, I think the chances of collision are high.  Of course I don't know how likely any collision is these days given the increased speed of communication between researchers.  However, I don't see that allowing authors hurts anything and it's already implemented.  Whether folks use them are not is up to them.

PR #2830, opened 2025-03-22
Update form-exif_controller.js to be more tolerant of camera date data

nimmolo:
Addresses #2831

Extend the form-exif Stimulus controller to parse nonstandard EXIF datetime data formats found in photos by DaveW and Anna Durkin, and log console messages if format still doesn't parse, rather than JS errors that arose from the code assuming the field and format.

Adds more helpful error messages to the browser `console.log` in case the format is not something we can handle. 
___

Dave's lab camera records a different separator between date and time, `"T"`, that is fairly common in code but is not EXIF spec. We had it coded to check for a space `" "`. Furthermore, we were prepared for EXIF date separators using `":"` but Dave's camera is using the common `"-"` -- common outside of EXIF, that is. 

There actually are two EXIF standard field names, `"DateTimeDigitized"` and `"DateTimeOriginal"`, but Anna's camera is not using either, and the only date field her camera is using, `"ICC Profile Date"`, is for the date a digital camera was calibrated to a color profile. The date values for this field in the photos she sent are different, though, so this must be the field it's using.  

(Phil Harvey of `exiftool` says these cameras are [doing it wrong](https://exiftool.org/forum/index.php?topic=8701.0), but we can't do much about that.)


nimmolo:
Note to @mo-nathan and @JoeCohen, please keep your browser console open and let me know what messages you get in the log in the case that some image's date isn't parsed right.

JoeCohen (approved):
Works fine for me without any error messages.
I glanced at the js, but am hopeless there.

app/javascript/controllers/form-exif_controller.js
```diff
@@ -143,6 +137,58 @@ export default class extends Controller {
...
+  // EXIF date parsing logic
+  // Confusingly, some cameras seem to incorrectly implement the EXIF standard 
+  // using other date field name conventions, and non-standard datetime formats.
+  // So we can't be too sure what we'll get. For example:
+  // exif_data["DateTimeDigitized", "DateTimeOriginal"] - legit, first preferred
+  // exif_data["ICC Profile Date"] - incorrect per the standard, but encountered
+  // in photos with no other datetime data where it seemed to be a created date.
+  // Note the difference in date/time separators:
+  //   {description: "2025:03:09 16:46:41.560", correct for EXIF
+  //    value: "2025-03-09T16:46:41.560"}, non-standard
```
mo-nathan, line 149: This is a very widely used ISO standard format.  I'm curious why your code doesn't use the standard JavaScript function Date.parse() (https://www.w3schools.com/jsref/jsref_parse.asp). I don't know if it handles the EXIF standard you mentioned but it seems like it would be a reasonable thing to use for the ISO standard format.

mo-nathan:
I'm encountering:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'description')
    at form-exif_controller…911cb941c.js:154:27
    at Array.find (<anonymous>)
    at t.parseExifDate (form-exif_controller…911cb941c.js:153:43)
    at t.populateExifDate (form-exif_controller…911cb941c.js:123:34)
    at t.populateExifData (form-exif_controller…8911cb941c.js:58:10)
    at _image.onload (form-exif_controller…8911cb941c.js:49:12)
  ```
 when upload any of the attached three images.  Two of them are old and probably have little or no EXIF data.  The other is a much more recent image, but may have had it's EXIF data purged somewhere along the way.

app/javascript/controllers/form-exif_controller.js
nimmolo: Thanks for checking it out. `Date.parse` unfortunately can't parse the EXIF standard format used in nearly every camera, `"2025:03:09 16:46:41.560"`, I tried. There's a new class `Temporal`, but it's not widely accepted enough for me to even check if it works. 

The other thing i didn't like about Date is that `Date.prototype.getMonth("June")` returns 5, as if that is ever helpful 🤪 (Date wouldn't even parse its own dates correctly, by that logic) so i felt like it wouldn't save us many lines or be as readable. ISO only differs from EXIF by the separators, so once we get separators, we can just take the digits as is.

The EXIF standard date format turns out to be very clear and explicit, despite what i said when i first inspected these images, and seems to be used most of the time. Thankfully Dave's camera's use of ISO seems to be an outlier. Anna's phone is just plain doing it wrong, but at least it has a date in there!

nimmolo: I will change that comment from `non-standard` to `ISO, incorrect for EXIF`

nimmolo:
@mo-nathan  - The sulfur shelf image seems to have had its EXIF data scrubbed. In this case, our script is going to pull the value Anna Durkin's camera uses, which is definitely incorrect (Jan 1, 2022 00:00:00).

I improved the object-"digging" safety that hunts for the value.

nimmolo:
The other two images at least don't error anymore.

PR #2832, opened 2025-03-23
RuboCop 1.74 string-concatenation autocorrects

JoeCohen:
Fixes offenses unsafe autocorrectable offenses. Part 1 of 3. 
Broken into 3 parts for ease of review. This part deals only with [Style/StringConcatenation](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylestringconcatenation)
For each putative offense:
- Accepts the fix if safe, or disables the cop if unsafe
- Makes follow-on changes to avoid new offenses. (Ex: removes `.to_s` inside string interpolation, divides long lines.)
- Updates .rubocop_todo


app/models/name/parse.rb
```diff
@@ -462,7 +462,7 @@ def format_autonym(name, author, _rank, deprecated)
         format_name(words[-1], deprecated)
       ].join(" ")
     else
-      format_name(name, deprecated) + " " + author
+      "#{format_name(name, deprecated)} #{author}"
```
JoeCohen, line 465: @mo-nathan will any of the changes to this file mess up things yo're working on?

JoeCohen:
@mo-nathan I had a very low level of concern about this PR, tagging you in only one file.
Nonetheless, it would be good to have a 2nd set of eyes on this.

app/models/name/parse.rb
mo-nathan: Don't think so.  Happy to resolve any conflicts if they show up.

mo-nathan (approved):
Looks good!  Thanks.

mo-nathan:
I agree this one is ready to go.  Let me know if you want some help with the conflicts.

PR #2835, opened 2025-03-23
Remove "create checklist from image query" code

nimmolo:
Apparently the UX for this was already removed, because deleting these methods doesn't flunk any tests.


nimmolo:
Apparently the UX for this was already removed, because deleting these methods doesn't flunk any tests.

PR #2837, opened 2025-03-23
Rename Query param[:by] as [:order_by]

nimmolo:
On the AR branch, I'm thinking to move the ordering logic into scopes, with the current param[:by] delegating to a new AbstractModel scope `.by(ordering_method)` as the dispatcher to more specific ordering scopes in each appropriate model. This is part of a goal of having Query do as little logic under the hood as possible, with most of its methods accessible to any scope in AR.

However, the name `:by` seems a little newbie-ambiguous in this case, since most of our records are "by" a user. The meaning of `:order_by(ordering_method)` is clearer.

### DIFF
--- a/app/classes/api2/model_api.rb
+++ b/app/classes/api2/model_api.rb
@@ -38,7 +38,7 @@ class API2
     def build_query
       params = query_params
       params.remove_nils!
-      params[:by] = :id
+      params[:order_by] = :id
       Query.lookup(model.name.to_sym, params)
     rescue RuntimeError => e
       raise(QueryError.new(e))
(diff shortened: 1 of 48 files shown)


JoeCohen:
You've probably already considered this.
But -- if not -- how about plain [:order]

nimmolo:
@JoeCohen `order` is ActiveRecord's method, reserved. 

It's what ultimately gets called by this, but this param takes shorthand args (like `:owners_quality`) that allow the scope to do some logic, like this:

```ruby
# AbstractModel::Scopes
  scope :order_by,
        ->(method) { send(:"order_by_#{method}") }

# Image::Scopes
  scope :order_by_owners_quality, {
    joins(:image_votes).
      where(ImageVote[:user_id].eq(Image[:user_id])).
      order(ImageVote[:value].desc)
  }
```

PR #2838, opened 2025-03-24
RuboCop 1.74 Unsafe Correctables 3/3 

JoeCohen:
Fixes unsafe correctable offenses. Part 3 of 3. 
Broken into 3 parts for ease of review. This part includes cops linked below. 
For each putative offense:
- Accepts the fix if safe, or disables the cop if unsafe
- Makes follow-on changes to avoid new offenses. (Ex: removes `.to_s` inside string interpolation, divides long lines.)
- Updates .rubocop_todo

#### Cops in this PR
Check the linked documentation for comments about cop safety.
> [Rails/Blank](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsblank)
[Rails/Output](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsoutput)
[Rails/RakeEnvironment](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsrakeenvironment)
[Rails/RedundantActiveRecordAllMethod](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsredundantactiverecordallmethod)
[Rails/ResponseParsedBody](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsresponseparsedbody)
[Rails/RootPathnameMethods](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsrootpathnamemethods)
[Rails/SelectMap](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsselectmap)
[Style/BitwisePredicate](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylebitwisepredicate)
[Style/CombinableLoops](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylecombinableloops)
[Style/FileNull](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylefilenull)[Style/SlicingWithRange](https://docs.rubocop.org/rubocop/1.74/cops_style.html#styleslicingwithrange)
[Style/HashEachMethods](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylehasheachmethods)
[Style/SafeNavigation](https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylesafenavigation)

### DIFF
--- a/app/extensions/string.rb
+++ b/app/extensions/string.rb
@@ -739,10 +739,6 @@ class String
 
   ### Misc Utilities ###
   #
-  def print_thing(thing)
-    print("#{self}: #{thing.class}: #{thing}\n")
-  end
-
   def to_boolean
     ActiveRecord::Type::Boolean.new.cast(self)
   end
--- a/app/extensions/string.rb
+++ b/app/extensions/string.rb
@@ -738,7 +738,12 @@ class String
   end
 
   ### Misc Utilities ###
-  #
+
+  # Disable cop because it seems like we really want to priint, not just log
+  def print_thing(thing)
+    print("#{self}: #{thing.class}: #{thing}\n") # rubocop:disable Rails/Output
+  end
+
   def to_boolean
     ActiveRecord::Type::Boolean.new.cast(self)
   end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -662,7 +662,10 @@ class Image < AbstractModel # rubocop:disable Metrics/ClassLength
   # to the :image field.  Returns true if the file is successfully saved.
   def save_to_temp_file
     result = true
-    unless upload_temp_file.present?
+    # disable cop; In theory upload_temp_file could be
+    # a blank non-empty string (e.g. " ").
+    # Don't fix unless...present?; it isn't broken.
+    unless upload_temp_file.present? # rubocop:disable Rails/Blank
 
       # Image is supplied in a input stream.  This can happen in a variety of
       # cases, including during testing, and also when the image comes in as
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -662,10 +662,7 @@ class Image < AbstractModel # rubocop:disable Metrics/ClassLength
   # to the :image field.  Returns true if the file is successfully saved.
   def save_to_temp_file
     result = true
-    # disable cop; In theory upload_temp_file could be
-    # a blank non-empty string (e.g. " ").
-    # Don't fix unless...present?; it isn't broken.
-    unless upload_temp_file.present? # rubocop:disable Rails/Blank
+    if upload_temp_file.blank?
 
       # Image is supplied in a input stream.  This can happen in a variety of
       # cases, including during testing, and also when the image comes in as
--- a/app/models/name/spelling.rb
+++ b/app/models/name/spelling.rb
@@ -111,8 +111,14 @@ module Name::Spelling
     ############################################################################
 
     # Guess correct name of partial string.
-    # This method should be private.
-    # See https://www.pivotaltracker.com/story/show/176098819
+    # NOTE: jdc 20250324 (copied from pivotaltracker)
+    # guess_word and guess_with_errors should be private.
+    # They were intended to be private
+    # and had been placed after a call to private.
+    # But that call was ineffective, so it was removed.
+    # https://docs.rubocop.org/rubocop/1.0/cops_lint.html#lintuselessaccessmodifier
+    # Furthermore, they can't be privatized because they are tested directly.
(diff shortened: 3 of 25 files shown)


app/extensions/string.rb
```diff
@@ -739,10 +739,6 @@ def md5sum
 
   ### Misc Utilities ###
   #
-  def print_thing(thing)
```
JoeCohen, line 742: `print_thing` isn't called anywhere.

app/models/name/spelling.rb
```diff
@@ -35,7 +35,7 @@ def suggest_alternate_spellings(str)
         results = guess_word("", words.first)
         (2..num).each do |i|
           next unless results.any?
-          next unless (i & 1).zero?
+          next unless i.even?
```
JoeCohen, line 38: @nimmolo : Is my change okay?
RuboCop wanted me to change `(i & 1).zero?` to `i.nobits?(1)`. https://docs.rubocop.org/rubocop/1.74/cops_style.html#stylebitwisepredicate
But both seem like extra mental gymnastics.

config/initializers/monkey_patches.rb
```diff
@@ -6,7 +6,7 @@
 require("action_view/log_subscriber")
 
 # Require all Ruby files in the core_extensions directory
-Dir[Rails.root.join("lib/core_extensions/**/*.rb")].each { |f| require f }
+Rails.root.glob("lib/core_extensions/**/*.rb").each { |f| require f }
```
JoeCohen, line 9: @nimmolo Please double-check

lib/tasks/lang.rake
```diff
@@ -78,7 +78,7 @@ namespace :lang do
   task setup: [:environment, :login, :verbose, :safe_mode]
 
   desc "Log in user for import tasks."
-  task(:login) do
+  task(login: :environment) do
```
JoeCohen, line 81: @mo-nathan @pellaea 
I added `:environment` as a dependency to 4 tasks.
- Question: Are these tasks ever used? 

- I don't understand the tasks.
But without this dependency, I got
```ruby
joe@iMac mushroom-observer % rake lang:login_admin
rake aborted!
NameError: uninitialized constant User (NameError)

    User.current = User.find(0)
    ^^^^
/Users/joe/mushroom-observer/lib/tasks/lang.rake:91:in `block (2 levels) in <top (required)>'
Tasks: TOP => lang:login_admin
(See full trace by running task with --trace)
```

JoeCohen:
@mo-nathan & @nimmolo:
I tagged you in comments where I had special questions.

lib/tasks/lang.rake
mo-nathan: I for sure don't intentionally use them.  However, they might get called indirectly.  In my view our translation stuff is kinda of wonky so I would move carefully.

app/models/name/spelling.rb
nimmolo: Thanks for asking. I don't understand what this is doing, though.

app/extensions/string.rb
nimmolo: Seems useful (and intended) for debugging, so it wouldn't stay in the code base. But i didn't know about it.

app/models/name/spelling.rb
mo-nathan: It's doing a bitwise and of i against the number 1 and then testing if that equals zero.  This will be the case if and only if i is even.  Why it was written the way it is I have no idea.  `git blame` current points @nimmolo , but I'm guessing it was copied from somewhere else.

app/models/image.rb
```diff
@@ -662,7 +662,10 @@ def validate_image_name
   # to the :image field.  Returns true if the file is successfully saved.
   def save_to_temp_file
     result = true
-    unless upload_temp_file.present?
+    # disable cop; In theory upload_temp_file could be
+    # a blank non-empty string (e.g. " ").
+    # Don't fix unless...present?; it isn't broken.
+    unless upload_temp_file.present? # rubocop:disable Rails/Blank
```
mo-nathan, line 668: `present?` is literally defined as `!blank?` (see https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/object/blank.rb).  I would go with Rubocop's recommendation here.  For " " `blank?` is true and `present?` is false.  If you want to allow filenames that include `" "` you would have to explicitly test for that.  However, that seems so edge casey that I think it's a bad idea to consider in our code.  If you know this is actually a likely name, then I'd reject it or rename it to something reasonable.

mo-nathan (approved):
Made a few minor points, but overall this looks great!

PR #2843, opened 2025-03-27
Normalize author characters

JoeCohen:
Composes unicode characters in Name.author prior to validation and saving/updating.

- Fixes issue where author containing uncomposed unicode characters (example below)
  - failed validation, and
  - was hard to impossible to fix.

Example:
```ruby
mushroom-observer(dev)> Name.find(63893).author
=> "V. Kučera, Lizoň & Tomšovský"
Name.find(63893).author.unicode_normalized?
=> false
```

### Suggested Manual Test
- Edit (just open for editing) one of the Names in the [Name Issues spreadsheet](https://docs.google.com/spreadsheets/d/1XMgkxDg4m_sUv65rQd6F7NSiLDjuZG9iQFXLvk4w_Rk/edit?gid=1264169157#gid=1264169157) which appears to be failing because it has diacriticals.
- Don't enter any changes.
- Save edits
Expected Result: `Successfully updated name` 
(Although you didn't enter any changes, the Name gets updated because the author changed.)
- Edit (just open for editing) the same Name.
- Don't enter any changes.
- Save edits
Expected Result: `No changes made.` 
(No changes made because the author was fixed first time around.(


JoeCohen:
@mo-nathan 
- This is the simplest thing I could think of to deal with the issue.
- It's reasonable to also do this in name/parse.rb especially if you are later using regexp's on author.
- Maybe (separate PR) the validator should be moved to AbstractModel and run on all text-ish fields.
- IDK which coverage dropped (and -0.1% is relatively big for us). But I don't how the drop could be caused by this PR.

JoeCohen:
I hadn't thought about how this might affect sort and order

mo-nathan (approved):
Looks good to me!

PR #2845, opened 2025-03-27
Rename pagination stuff

nimmolo:
Something about the way `ApplicationController` instantiates only the current page of results for the query is failing on my Query AR branch, and that has led me to confront my confusion about how this works. The ivar `@pages` is part of it. Until today thought it contained query results, and I can't remember the last time I really checked what it contains.

It's just a little package of data about current pagination. It gets populated by `ApplicationController` using the current query, but that's different. It's also used by `Query` to figure out which results to return. The name `@pages` is admirably brief, but since the object contains `@pagination_data` and not `@pages_of_results` i'm renaming it here. Also renaming the class `MOPaginator` that it is an instance of to `PaginationData`.

I also thought `paginate_block` et al. did a bunch of magical manipulation of the query, but it just adds a pagination nav on the top and bottom of the results. Yes, that's a valid meaning of "paginate", but in the context of a query "paginate" means "return results subdivided by a given page". Here, it's renamed `pagination_nav`.

This PR hopefully disambiguates what has confused me for years, and maybe demystifies these methods for others. If not, hopefully it's at least as clear.


nimmolo:
@mo-nathan That's a good suggestion.  I am realizing my issue is also with the name "MOPaginator" - as the comment block at the top of the class acknowledges, the name implies this is more than a data object, i.e. a class that does or retrieves stuff, which it is not. 

Shall I rename that object "PaginationData"? (or "MOPaginationData")

JoeCohen (approved):

PR #2851, opened 2025-03-29
User specific iNat Import API Key

### LINKED ISSUE #2849: Import with User's API key
Import using the importing User's MO API key

- [ ] Search iNatImportJob for uses of iNat manager's API Key
- [ ] Add assertions to iNatImportJobTest, maybe InatImportControllerTest
- [ ] In InatImport find or create API Key for the user
- [ ] Fix places in INatImportJob which used webmaster key
- [ ] Will this enable greater use of MO API
  - [ ] Create obs
  - [ ] add images

JoeCohen:
Employ a special-purpose, personal MO APIKey for iNat Imports.
- Creates the key if it doesn't already exist.
- Uses the key for all API2 calls during the import.
  - Imports Image, adding it to Observation in a single step, thereby avoiding 2 writes to the db and some emails.
- All objects created during an iNat import are now owned by an MO User other than `webmaster`.
- Removes last vestiges of  `inat_manager`.

### Suggested Manual Test
- Delete any of your [MO API keys](http://localhost:3000/account/api_keys) whose Notes == `inat import`
- Create an iNat observation which has a “Provisional Species Name” Observation Field, with a simple provisional name -- _genus “sp”_  (including the quotes) which does not exist in MO.  Ex: `Mycena "skamaniac"`
See https://www.inaturalist.org/observations/253297232
- Start Solid Queue is running: `bundle exec rake solid_queue:start`
- Import the iNat observation to MO. 
- When importing is Done, click `Results`. (It should display the new Observation with a new provisional Name.)
- Go the Name page (click the Name), scroll to the bottom
Expected Result: The Name should have been Created By you.
(In main, it would be Created and Updated by webmaster.) 
main: Screenshot 2025-04-05 at 2 17 28 PM

- Go to your [MO API keys](http://localhost:3000/account/api_keys). http://localhost:3000/account/api_keys
Expected Result: You should have a key, created today, with Notes `inat import`.


app/jobs/inat_import_job.rb
```diff
@@ -255,42 +260,35 @@ def add_inat_images(inat_obs_photos)
...
-        original_name: photo.original_name
+      next unless Rails.env.test?
+
+      # NOTE: 2025-04-01 jdc
+      # A hack faking the side effect of the above API call
+      # The API adds the imported image to the MO observation
+      image = api.results.first
+      ObservationImage.create!(
+        observation: @observation,
+        image: image
```
JoeCohen, line 271: I can't figure out how to stub, mock, or otherwise fake this side effect in the test (`test/jobs/inat_import_job_test.rb`).

`Inat::PhotoImporter` wraps an API call.
The API imports an image and **adds it to an MO Observation**.
The tests do mock `Inat::PhotoImporter` and return a value that's sufficient for testing.
When the mock is created the Observation doesn't exist, so I don't know how tell the mock which Observation needs to have the image added.

The hacky code in InatImportJob was the best I could do.
When that code is run both the Observation and the (mock) Image are knowm.

mo-nathan:
Was able to confirm that it works from local.

app/jobs/inat_import_job.rb
mo-nathan: Start looking at this, but need some more time.

mo-nathan: Ok, I found the issue.  The primary issue was that you weren't actually providing the system with an image to connect to the observation.  I have version of the code on njw-inat-job-stub-fix that removes the dependency on Rails.env.test? and all the tests pass.  This required removing some of the `Inat::PhotoImporter.stub`s.  I actually wonder if all of those should get removed, but I didn't go that far.  The key new function is `image_for_stubs` which provides an actual image in the stub response to the iNat get calls.  I spread this around liberally.  What I didn't do was refactor the new stub calls to keep them DRY.  Finally, I ran into an issue where `imported_img.original_name` was returning `nil` rather than the expected value.  This turns out to be due to the user `rolf` having the keep_filenames preference set to `toss`.  I don't know how important that bit of the test is to you, so I just covered ith with a simple guard condition.  So in summary, before merging that PR into this PR, I recommend:
1. Reviewin all the stubbings of `Inat::PhotoImporter.stub` and see if you can replace them with lower level stubs to the iNat get calls.  I think this will conceptually at least simplify things and it is a better place to do the stubbing.
2. Look at refactoring the iNat get calls into one or two helpful functions.  You have started down this road with `stub_inat_photo_requests`, but there seem to be two different places that the stored iNat responses expect their images to come from (`https://inaturalist-open-data.s3.amazonaws.com/photos` and `https://static.inaturalist.org/photos/`).  Not sure why, but it's what I observed while working on the tests.
3. Review the conditional I put on `imported_img.original_name` and figure out what you really want there.  Maybe you want to use an other users, maybe you want to not have that assert, maybe something else.

mo-nathan (approved):
Looks good!

PR #2852, opened 2025-03-29
Refactor Report not to use User.current

nimmolo:
Does not change functionality.

The only UI client for `Report` is the `DownloadsController`. Changes the callers to send `@user`, and changes `Report` to expect a user or default to nil (as before).

Passes user to `ProjectTweaker` and `ProjectLatLngs`, where it is also used to get private lat/lngs.

Alters tests accordingly. (User-related functions already tested.)

### DIFF
--- a/test/classes/report_test.rb
+++ b/test/classes/report_test.rb
@@ -224,11 +224,10 @@ class ReportTest < UnitTestCase
     ]
     do_csv_test(Report::Fundis, obs, expect, &:id)
 
-    User.current = mary
     expect[11] = "34.1622"
     expect[12] = "-118.3521"
     expect[13] = nil
-    do_csv_test(Report::Fundis, obs, expect, &:id)
+    do_csv_test(Report::Fundis, obs, expect, user: users(:mary), &:id)
   end
 
   def test_raw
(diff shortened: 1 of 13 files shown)


nimmolo:
Note this passes tests and maintains coverage, but Coveralls is down for service tonight, so the test run is showing as a fail, even though it's a pass.

test/classes/report_test.rb
```diff
@@ -540,17 +538,17 @@ def test_project_tweaker_report
...
     table = CSV.parse(body, col_sep: report_type.separator)
     assert_equal(query.num_results + 1, table.count)
     idx = query.results.sort_by(&block).index(obs)
     assert_equal(expect, table[idx + 1])
   end
 
-  def report_body(report_type, query)
-    report = report_type.new(query: query)
+  def report_body(report_type, query, user: nil)
+    report = report_type.new(query:, user:)
```
JoeCohen, line 551: Should this be the following? No reply necessary:
`report = report_type.new(query: query, user:)`

JoeCohen (commented):
Glanced at code quickly. Had only one comment. If that looks ok to you I think you should merge.

test/classes/report_test.rb
nimmolo: I think i understand your question: only `:user` was a kwarg incoming, whereas `query` was an anonymous arg. But with the Ruby "kwarg" forwarding syntax, if the kwarg name matches a defined variable it will implicitly take that value if none given, even if the var was not originally a kwarg.

PR #2859, opened 2025-03-31
First Query conversion: Comments

nimmolo:
This PR stages the new modules supporting the AR Query classes, and converts one model, Comments. The new classes inherit from a new base class, `base_new.rb`.

Coverage of not-yet-used methods in the new Query `ScopeModule`s will temporarily dip until the other classes and their tests are converted. Check #2856 for eventual coverage.

### DIFF
--- a/app/classes/query/base_new.rb
+++ b/app/classes/query/base_new.rb
@@ -25,14 +25,15 @@ class Query::BaseNew
 
   delegate :parameter_declarations, to: :class
 
-  # can use has_attribute? here.
-  # only called in ApplicationController::Queries#apply_one_content_filter
+  # Could use has_attribute? here, but it doesn't exist yet for ActiveModel.
+  # Only called in ApplicationController::Queries#apply_one_content_filter
   def self.takes_parameter?(key)
     parameter_declarations.key?(key)
   end
 
   delegate :takes_parameter?, to: :class
 
+  # :id_in_set must be moved to the last position so it can reorder results.
   def self.scope_parameters
     excepts = [:id_in_set, :preference_filter]
     @scope_parameters = parameter_declarations.except(*excepts).keys +
@@ -78,15 +79,15 @@ class Query::BaseNew
   # more efficient however than using a Rails-serialized column and comparing
   # the parsed hashes (in whatever order), because when a column is serialized
   # you can't use SQL on the column value, you have to compare parsed instances.
+  # was params.sort.to_h.merge(model: model.name).to_json
   def serialize
-    # params.sort.to_h.merge(model: model.name).to_json
     attributes.compact.sort.to_h.merge(model: model.name).to_json
   end
 
   def record
     # This errors out if @record is not set since it
     # cannot find Query.get_record.  If you copy the
-    # above definition of get_record into the same scope
+    # definition of get_record into the same scope
     # as this method and get rid of "Query." it works,
     # but that is not a great solution.
     # You can trigger the issue which is
--- a/app/classes/query/base_new.rb
+++ b/app/classes/query/base_new.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+# base class for Query searches
+class Query::BaseNew
+  include ActiveModel::API
+  include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+
+  include Query::Modules::ClassMethods
+  include Query::ScopeModules::ResultAccessors
+  include Query::ScopeModules::Initialization
+  include Query::ScopeModules::SequenceOperators
+  include Query::Modules::Validation
+
+  validates_with Query::Modules::Validator
+
+  # "clean up" and reassign attributes before validation
+  before_validation :clean_and_validate_params
+
+  attr_writer :record
+
+  def self.parameter_declarations
+    { order_by: :string }
+  end
+
+  delegate :parameter_declarations, to: :class
+
+  # can use has_attribute? here.
+  # only called in ApplicationController::Queries#apply_one_content_filter
+  def self.takes_parameter?(key)
+    parameter_declarations.key?(key)
+  end
+
+  delegate :takes_parameter?, to: :class
+
+  def self.scope_parameters
+    excepts = [:id_in_set, :preference_filter]
+    @scope_parameters = parameter_declarations.except(*excepts).keys +
+                        [:id_in_set]
+  end
+
+  delegate :scope_parameters, to: :class
+
+  def self.content_filter_parameters
+    filters = Query::Filter.all
+    @content_filter_parameters ||= filters.each_with_object({}) do |f, p|
+      p[f.sym] = f.type
+    end.freeze
+  end
+
+  delegate :content_filter_parameters, to: :class
+
+  # Can the current class be called as a subquery of the target Query class?
+  def relatable?(target)
+    self.class.related?(target, model.name.to_sym)
+  end
+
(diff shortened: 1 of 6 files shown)


app/classes/query/base_new.rb
```diff
@@ -0,0 +1,105 @@
...
+  def self.takes_parameter?(key)
+    parameter_declarations.key?(key)
+  end
+
+  delegate :takes_parameter?, to: :class
+
+  def self.scope_parameters
+    excepts = [:id_in_set, :preference_filter]
+    @scope_parameters = parameter_declarations.except(*excepts).keys +
+                        [:id_in_set]
```
JoeCohen, line 39: Reads like you are first removing `:id_in_set` as a valid param, then adding it back.

app/classes/query/base_new.rb
```diff
@@ -0,0 +1,105 @@
...
+  # Prepares the query params, adding the model, for saving to a QueryRecord.
+  # The :description column is accessed not just to recompose a query, but to
+  # identify existing query records that match current params. That's why the
+  # keys are sorted here before being stored as strings in to_json - because
+  # when matching a serialized hash, strings must match exactly. This is
+  # more efficient however than using a Rails-serialized column and comparing
+  # the parsed hashes (in whatever order), because when a column is serialized
+  # you can't use SQL on the column value, you have to compare parsed instances.
+  def serialize
+    # params.sort.to_h.merge(model: model.name).to_json
```
JoeCohen, line 82: Consider moving this line into the prior comment.

app/classes/query/base_new.rb
```diff
@@ -0,0 +1,105 @@
...
+    attributes.compact.sort.to_h.merge(model: model.name).to_json
+  end
+
+  def record
+    # This errors out if @record is not set since it
+    # cannot find Query.get_record.  If you copy the
+    # above definition of get_record into the same scope
```
JoeCohen, line 89: "above definition of get_record"?

app/classes/query/scope_modules/initialization.rb
```diff
@@ -0,0 +1,223 @@
...
+      if subquery_params.present?
+        @scopes = @scopes.send(:"#{model.name.downcase}_query",
+                               **subquery_params)
+      end
+    end
+  end
+
+  # Current types requested on the RssLog page. Defaults to :all.
+  def rss_logs_requested_types
+    types = [:observation, :name, :location]
```
JoeCohen, line 94: Do you also need Species Lists, Projects, Glossary, News?
Screenshot 2025-03-31 at 7 04 12 AM

app/classes/query/scope_modules/sequence_operators.rb
```diff
@@ -0,0 +1,85 @@
...
+  # Move to first place.
+  def first
+    new_self = self
+    id = new_self.result_ids.first
+    if id.positive?
+      @current_id = id
+    else
+      new_self = nil
+    end
+    new_self
```
JoeCohen, line 43: Will this work?  Regardless, do you need safe navigation in `.result_ids.first`
```suggestion
    if result_ids.first.positive?
      @current_id = id
    else
      nil
    end
```

app/classes/query/scope_modules/sequence_operators.rb
```diff
@@ -0,0 +1,85 @@
...
+  def last
+    new_self = self
+    id = new_self.result_ids.last
+    if id.positive?
+      @current_id = id
+    else
+      new_self = nil
+    end
+    new_self
+  end
```
JoeCohen, line 84: See comment about `def first`

app/classes/query/base_new.rb
nimmolo: That's right. 

Oh, wait... I think this is an artifact of when i was trying to `compact` the params _other than_  `id_in_set`: that is the only param that should preserve an empty array as its value, because it forces empty results. 

I'll review and check if i can remove this, thanks.

app/classes/query/scope_modules/initialization.rb
```diff
@@ -0,0 +1,223 @@
...
+  def sendable_params
+    sendable = params.slice(*scope_parameters)
+    return sendable unless model == RssLog
+
+    sendable.except(*content_filter_parameters.keys)
+  end
+
+  # Most name queries are filtered to remove misspellings.
+  def filter_misspellings_for_name_queries
+    return if model != Name || !params[:misspellings].nil?
```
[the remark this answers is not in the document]
nimmolo: That's handled elsewhere, in the scope. 

This is just the default of the scope chain, at the very end: if there was no `:misspellings` param (either way) in the request, then add one.

app/classes/query/scope_modules/sequence_operators.rb
nimmolo: Good question. This is very old code and I didn't want to touch it. 

With almost all of Jason's logic, when i found apparent quirks like this i tend to discover after messing with it for 3 hours that i cannot improve on it. That happened so many times i got scared to waste time trying.

Here I'd bet that the object needs to be cloned for something else to work, but idk.

app/classes/query/base_new.rb
nimmolo: Ah, good catch. I actually need to delete that. 
That's the version of the same line, from before I added the `ActiveModel::API`.

app/classes/query/base_new.rb
nimmolo: Oh i know. It's to force order. `:id_in_set` must come last because it reorders results, but to ensure that it unsets any previous order it must be last in the chain.

JoeCohen (commented):
Left some nit-picky comments (because I don't have a good grasp of he big picture).
I played with Comment pattern searches a bit and results seem the same as main.

app/classes/query/scope_modules/initialization.rb
nimmolo: Added

app/classes/query/scope_modules/sequence_operators.rb
nimmolo: I'm leaving this alone. Feel free to try something afterwards though.

app/classes/query/scope_modules/initialization.rb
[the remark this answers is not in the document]
nimmolo: Renamed

app/classes/query/scope_modules/initialization.rb
```diff
@@ -0,0 +1,223 @@
...
+    @initialized = true
+    @scopes      = model
+    initialize_scopes
+    @last_query = sql
+  end
+
+  def sql
+    initialize_query unless initialized?
+
+    @sql = scopes.all.to_sql
```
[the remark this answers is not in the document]
nimmolo: Done

app/classes/query/base_new.rb
nimmolo: Edited

app/classes/query/base_new.rb
nimmolo: Moved it up, in case needed later

app/classes/query/base_new.rb
nimmolo: Added comment explaining

app/classes/query/scope_modules/initialization.rb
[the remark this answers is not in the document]
nimmolo: Comment added

Issue #2865, opened 2025-04-01
UI for Observation#inat_id


JoeCohen:
From Slack posts 2025-04-01
> [04:17](https://mushroomobserver.slack.com/archives/D2AAX55NK/p1743506229836919)
nathan
 Is there a way to create a link between an existing MO obs and an iNat obs?
[06:07](https://mushroomobserver.slack.com/archives/D2AAX55NK/p1743512868767339)
jdcohenesq
 TLDR: Probably not something which will work or you (other than adding something to the Observation notes).
What scenario do you have in mind?
The following exist, but probably aren’t what you want.
From MO to iNat:
Pulk’s Mirror script simply adds text in this form to the Observation#note: “---\nMirrored on iNaturalist as observation 116782486 on May 12, 2022.”
Inat sets Observation#inat_id to the id of the iNat obs and sets Observation#sourcce to mo_inat_import. Result: Observation Details says “Imported yyyy-mm-dd from iNat #nnnnnn.” Ex: https://mushroomobserver.org/572691.
If you set Observation#inat_id  and Observation#source is not mo_inat_import , then Observation Details says “Copied to iNaturalist as [iNat #1234567](https://inaturalist.org/observations/1234567)”/ That probably isn’t what you want. I was anticipating using Obseration#id for exports.
From iNat to MO;
Add a “MO URL” Observation field whose value is the url of the MO obs. This was/is done by: iNatImport, the (defunct) iNat program which imported MO Obss; Pulk’s Mirror script. It can be done manually. It could be done ia the iNat API
Notes:
InatImport checks inat_id in order to avoid re-importing an iNat obs.  app/controllers/inat_imports_controller/validators.rb#fresh_import?.
InatImport also checks the “MO URL” Observation Field for the same purpose.
There’s no UI to set/modify inat_id . I plan to let the API do that.
...
[08:33](https://mushroomobserver.slack.com/archives/D2AAX55NK/p1743521598602969)
nathan
 Do the use case is are these two observations: https://mushroomobserver.org/572696 and https://www.inaturalist.org/observations/267915292.  Aubrey who runs [mushroommonday.com](http://mushroommonday.com/) just posted this blog post: https://mushroommonday.substack.com/p/mushroom-monday-da6 and I want to use it as an opportunity to mention the MO/iNat import feature.  I’m planning on mentioning your article about the importer, but I’d like to describe how to link two existing observations.
  > iNaturalistiNaturalist
[...]

PR #2867, opened 2025-04-01
Move external links

nimmolo:
#### Add column `:base_url` to `ExternalSite`
- Used to validate if associated `ExternalLink`s fit the site.

#### Adds new class `FormatURL`
- Validates and "corrects" URL format for single URLs, and can reformat a provided `url` against a provided `base_url` 

#### Moves external links into the Observation Details box under Notes, per discussion on #2865.
- Adds date created beside the link.

Screen Shot 2025-04-01 at 4 50 41 PM

#### Adds a bunch of new ExternalLinkController tests. 
These have been turned off since the CRUD refactor. Adds minimal HTML templates so the controller tests can work.


mo-nathan:
When no links are present it always says "MyCoPortal: [+]".
Screenshot 2025-04-02 at 7 42 24 AM

mo-nathan:
I also think it we need to be able to add more than one external link.  I think I saw an observation that there was one ever one in our database, but that just seems like more of a sign of lack of sophistication in our user base.  The obvious use case is an observation that has an entry in MyCoPortal and iNat.  If we are successful this should become common.

mo-nathan (changes requested):
The interface should not say "MyCoPortal" when there is no link.
The interface should allow me to add both a "MyCoPortal" link and an "iNat" link and ideally other links that may be relevant.

JoeCohen:
> The interface should not say "MyCoPortal" when there is no link. 

I don't understand this comment, and probably disagree.
- If we want users to always have the ability to add a link to MyCoPortal (see below), then MyCoPortal [+] is the right thing. Ditto for iNat. See your other comment below
- It's the same UI as for Fungarium Records and Sequences which appear directly above.
It's simple and IMO intuitive. 
@nathan: What other UI do you suggest. See your comment below.

- "Fungarium Records +" should appear only if observation#specimen == true. I don't understand how there can be a Fungarium Record if there's no specimen. This probably belongs in another PR.
- Eventually Users should **not** be able to add MyCoPortal links to the MyCoPortal MUOB counterpart. [Example](https://www.mycoportal.org/portal/collections/individual/index.php?occid=8188734&clid=0). Rather, this should happen automatically and users should not be able to modify it.
- MyCoPortal can have counterparts which are not MUOB records. A typical case will be a herbarium record for a herbarium which uses MyCoPortal. For instance this [UBC MyCoPortal record](https://www.mycoportal.org/portal/collections/individual/index.php?occid=13380191&clid=0) likely is a counterpart of [MO 352150](https://mushroomobserver.org/352150). 
- MyCoPortal could have more than one non-MUOB counterpart to a single MO record. Examples: Collections could be split and go to multiple herbaria. Observations may also appear in  other MyCoPortal databases, like FunDiS or General Observation and Personal Collections(MP). This implies that we should allow multiple MyCoPortal links per MO Observation.

> The interface should allow me to add both a "MyCoPortal" link and an "iNat" link and ideally other links that may be relevant.

Agree. What UI do you suggest?

JoeCohen (commented):
1. Consider limiting this PR to simply moving the location of the external links. IMO that would be an improvement and avoid mission creep.
2. If not so limited, could all/some of the error handling be moved to model validations?
3. The UI should have more documentation. I don't know how the UI works/should work (other than adding a MyCoPortal link). For instance, how can someone add an External Site (e.g. MyCoMap). 
4. I'm happy to play with the UI, as time allows, once all tests are passing.

nimmolo:
Clarifications: 
- The one-link limitation: The current limitation is that you can only create one link **per ExternalSite**.
- `external_sites` is a database table with exactly one entry (MyCoPortal). I suspect this db table could be dropped, and `EXTERNAL_SITES` could just be a site constant hash. Other properties could be added to each key in the hash, that give us configuration details about how that link should behave, automatic URL builders by ID for the UI, etc.
- The "add link" UI currently shows you the names of all the sites you can link to, with a plus sign, that's why it says `MyCoPortal`. Because there's only one, it's confusing. When we add iNat, the current UI would say `External Links: iNaturalist [+]  MyCoPortal [+]`.
- However I agree this UI could be simpler. I'll adjust it so it's just a `[+]` and the form itself gives you a `select` that allows you to pick which site, graying out options where you already have a link.

nimmolo:
PR now prints a plain `[+]` to add an `ExternalLink`, and offers a `select` element in the form to pick the site. Currently only offers `MyCoPortal`.

Not done in this PR: adding iNaturalist to `ExternalSites`.

nimmolo:
@JoeCohen Currently admins have to manually add the `ExternalSite` to the db, there's no interface for it. 

However, I feel like making this a site constant hash is the way to go. 
If everybody agrees, that would be my next PR.

- eliminate the `has_one` association to `ExternalSite`
- add a migration to make `external_site_id` an enum whose values map to the keys of the `EXTERNAL_SITES` constant.
- write a site constant like this:
```ruby
EXTERNAL_SITES = {
  mycoportal: {
    site_url: "https://www.mycoportal.org",
    url_validations: [
      "some_regex_string_we_can_save_to_the_db",
      "an_alternate_valid_regex_we_should_check"
    ],
    some_other_property: "foo"
  },
  inaturalist: {
    site_url: "https://inaturalist.org",
    url_validations: [
      "some_regex_string_we_can_save_to_the_db",
      "an_alternate_valid_regex_we_should_check"
    ],
    some_other_property: "foo",
    a_custom_inat_property: "fah"
  },
}
```

nimmolo:
@JoeCohen I like the idea of model validations, will move the url validator there.

JoeCohen:
Re validators. My thought was:
The controller tries to `create` or `update`
The model validates everything possible using `validate` or `validates` callbacks, each of which creates an error msg upon failure.
The controller just reloads the form if the `create`/`update` update fails.

JoeCohen:
Re External Sites.
In the long run there should be a way to add one via a UI, maybe limited to admins.
I think that implies that ExternalSite should be a model.
IDK if an EXTERNAL_SITES constant hash helps us get there.

nimmolo:
@mo-nathan @JoeCohen 
I can move our `EXTERNAL_SITES` to a site constant hash in this PR, if you guys want that. 
This will make it easier to add special properties per site (easier than adding db columns).

However, there's a consideration I didn't notice before. 
`ExternalSite` has a `project_id` column because it `belongs_to :project`.
`Project` however does not have an `external_site_id` column.

I'd make `:external_site` an enum for `ExternalLink`, but to preserve the links between ExternalSites and Projects maybe we'd want to add an `:external_site` enum to `Project` as well?

Check out this method. If a user is a member of the right project, they can add external links to other users' observations. This could be rewritten as a Project class method to check the enum instead.

```ruby
# external_site.rb
  # Either the observer, an admin, or a member of a Project associated with
  # a specific ExternalSite can add an external link to an Observation.
  def self.sites_user_can_add_links_to(user, obs, admin: false)
    return [] unless obs && user

    obs_site_ids = obs.external_links.map(&:external_site_id)
    if (user == obs.user) || admin
      where.not(id: obs_site_ids)
    else
      where(project: user.projects_member).where.not(id: obs_site_ids)
    end
  end
```

nimmolo:
@JoeCohen I don't know if we really need an admin UI for something that is this easy:
```ruby
> rails console
> ExternalSite.create(name: "iNaturalist", project_id: 456)
```

JoeCohen:
I'm not getting my head everything in (and proposed to be in) this PR. 
So I'll just bak off. Sorry.

mo-nathan:
Arguments against moving external site type information from the database into the code:

1) In my view this does not make it "easier to add special properties per site".  Migrations are very easy to create and they are expected by most Rails developers.  Hiding the external site configuration information inside the code makes it essentially impossible to some day add a UI for adding new types of external sites.  This could be particularly relevant for Names where I could see us wanting to add links to places like Wikipedia, Amanitaceae.org, or whatever new site comes along that provides useful and interesting data about names.

2) In my view all code changes should have tests that fail without the change and succeed with the change.  Storing what really is data in the code makes this weird and surprising.  It also means rather than just adding more data to the database we actually have to make a code change and deploy it.  I have been burned by that problem many times.  We happen to have a relatively efficient deployment process, but it still requires a PR etc.

3) Assuming this data is going to be available by from the UI, I would expect there to be a one call that tells you the type information about external sites and another for getting the external sites possibly filtered by the type of the external site.  The closer the API is the database the better.  It creates more consistency and folks using data dumps will be better able to understand the content.

4) We already have a table for this information so why would be change that?

JoeCohen:
@nimmolo @nathan Is something like this possible:
An ordinary user cannot delete an ExternalLink to iNat if that link was created by iNatImportJob?
(Could that be implemented by making the ExternalLink.user = 0 ?)

mo-nathan:
> @nimmolo @nathan Is something like this possible: An ordinary user cannot delete an ExternalLink to iNat if that link was created by iNatImportJob? (Could that be implemented by making the ExternalLink.user = 0 ?)

Possible, sure.  Wise, more debatable.  Why would you want that?

JoeCohen:
Don't want to lose the data about the ultimate source of the obs.

mo-nathan:
Recording the data source should probably be handled differently than just having the link.  I would love it if the logs worked better.  It's on the list to fix some of the persistent issues, but that's probably a fair amount of work.  However, getting a note about the import into the log should be hard and seems like a really good idea to me.  My concern is if the observation in iNat gets deleted, or maybe changed to something better, or something gets twisted around with the import due to some bug, I'd rather allow folks to edit the link than require them to go through the admins.

PR #2869, opened 2025-04-03
Single quotes rather than double quotes for provisional names

mo-nathan:
All places where we used to use double quotes for provisional names, you should now be able to use single quotes.  In most cases if you use double quotes, they should get mapped to single quotes.


app/models/name/parse.rb
```diff
@@ -122,7 +122,7 @@ def match?(str)
     #{ANY_AUTHOR_ABBR} |
     van\s | d[eu]\s |
     [A-ZÀÁÂÃÄÅÆÇĐÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞČŚŠ(] |
```
JoeCohen, line 124: Line 124. Can/Should the list of letters be replaced by `/\p{Latin}/u`

mo-nathan: Perhaps, but I consider that outside the scope of this PR.

PR #2874, opened 2025-04-04
Ignore brakeman Name Format Validation warning

JoeCohen:
Updates Brakeman configuration file to ignore this warning: 

[Format Validation](https://brakemanscanner.org/docs/warning_types/format_validation/) Insufficient validation for author using `/[\p{Alpha}\.]( *)\z/`. Use \A and \z as anchors near line 374

This validation is concerned only with the termination of `author`, and therefor doesn't use `\A`.
Another validation checks the entirety of `author` using `\A ... \z`.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,13 +1,35 @@
 {
   "ignored_warnings": [
     {
+      "warning_type": "Format Validation",
+      "warning_code": 30,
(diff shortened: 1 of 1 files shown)


JoeCohen:
Merging without deploying. It's just a test config file having no effect on production.

PR #2882, opened 2025-04-07
Cover ERBs

JoeCohen:
Include `.erb` (and builder) files in coverage report


mo-nathan (approved):
Looks good.  It's a bit surprising that this actually increased our coverage percentage, but that's a good thing.  Still some ERBs that we need to work on getting covered, but I was afraid it was going to cause a big drop.  The warnings are annoying.  I was wondering if we could suppress them by putting a footer comment at the bottom of each file.  I don't know how annoying it will be that the report will be 2 to 3 lines off, but as long as everyone is aware, I think we can adapt.

JoeCohen:
@mo-nathan What do you mean by "a footer comment"?
I tried adding `<%# >` at the bottom of couple of files, but it didn't stop the excess coverage warnings.
I looked for, but couldn't find, another workaround.
I'm OK ignoring the warnings.
But there are a ton of them.
@nimmolo: What do you think?

mo-nathan:
Yeah, that’s what a meant.  Sounds like it doesn’t work.  I assume just add
some blank lines at the end doesn’t work.

JoeCohen:
Unfortunately, extra blank lines don't work.

JoeCohen:
Another possibility: monkey patch SimpleCov.
There's an old (2013) [simplecov-ruby Issue ](https://github.com/simplecov-ruby/simplecov/issues/1057) which suggests code to do that, but doesn't actually submit a PR. The gist of the suggestion is:
```ruby
# spec/support/simplecov_warnings_patch.rb
# frozen_string_literal: true

module SimpleCov
  class SourceFile
    # this is needed for SimpleCov Issue#56
    # it also shows up for views when we `enable_coverage_for_eval`
    # patching here so that we can turn this off via ENV VAR
    def coverage_exceeding_source_warn
      return unless ENV['SHOW_EXCESS_LINE_WARNING']

      message = "Warning: coverage data provided by Coverage [#{coverage_data['lines'].size}] "
      message += "exceeds number of lines in #{filename} [#{src.size}]"
      warn message
    end
  end
end
```
I personally wouldn't mind turning off those warning completely. I see them occasionally but never pay attention.

JoeCohen:
I don't know how to do that. In particular, where (in what file?) should that code go?

JoeCohen:
@mo-nathan Thanks for the monkeypatch. 
@nimmolo I'd like to mere this so that we get notified about any future loss of coverage in views.

PR #2883, opened 2025-04-07
Try to fix AR in checklist.rb

nimmolo:
For reference, this was the original SQL:
```ruby
    synonyms = Name.connection.select_rows(%(
    SELECT GROUP_CONCAT(n.id),
      MIN(CONCAT(n.deprecated, ',', n.text_name, ',', n.id, ',', n.rank))
    FROM names n
    GROUP BY IF(synonym_id, synonym_id, -id);
    ))
```

I don't know what the `-id` means in the last line (specifically the `-`).

This PR is trying this rewrite and it runs locally without error where the other one failed. 
But i'm not sure it's right:
```ruby
    synonyms = Name.connection.select_rows(
      Name.select(
        Arel.sql("GROUP_CONCAT(names.id)"),
        (Name[:deprecated].cast("char") + "," +
         Name[:text_name].cast("char") + "," +
         Name[:id].cast("char") + "," +
         Name[:rank].cast("char")).minimum
      ).group(
        Name[:synonym_id].when(present?).then(Name[:synonym_id]).else(Name[:id])
      )
    )
```

Note that `GROUP_CONCAT` is apparently not accessible via Arel extensions, possibly because it's MySQL specific.

### DIFF
--- a/app/classes/checklist.rb
+++ b/app/classes/checklist.rb
@@ -123,7 +123,7 @@ class Checklist
          Name[:id].cast("char") + "," +
          Name[:rank].cast("char")).minimum
       ).group(
-        Name[:synonym_id].when(nil).then(Name[:id]).else(Name[:synonym_id])
+        Name[:synonym_id].when(present?).then(Name[:synonym_id]).else(Name[:id])
       )
     )
     synonyms.each do |row|


nimmolo:
I'm going to go ahead and merge this because 
- it gets rid of the error
- it's very low stakes even if the AR is wrong. It's just calculating site taxa by user, a statistic. It doesn't do CRUD.

Please do feel free to improve the AR if possible, though.

PR #2884, opened 2025-04-07
Remove letter pagination from the Observations index

nimmolo:
Letter pagination for Observations doesn't seem too useful.

The one place where it might be is on SpeciesList show pages, but still.


JoeCohen (approved):

mo-nathan (approved):
Looks good.

PR #2886, opened 2025-04-07
Change default search type on Names index to "Names"

nimmolo:
As suggested by @mo-nathan


JoeCohen (approved):
Not against this change, although IDK the reasons for it.

mo-nathan:
The reason for this change is that we want to get rid of the extra expense of the alphabetic pagination.  The reasoning is that this provides a more convenient way to get a subset of names quickly.  In essence it removes a click and select in a widget that users might not be familiar with.  It's not quite as simple as clicking on a letter, but it's probably more what the person is looking for.

PR #2889, opened 2025-04-08
Remove dependency on Rails.env.test? and get the tests working

mo-nathan:


JoeCohen:
I replaced this with #2890; I messed up some commits to #2889 and couldn't figure out how to recover.

PR #2895, opened 2025-04-11
Revert "Remove letter pagination from the Observations index"

mo-nathan:
Reverts MushroomObserver/mushroom-observer#2884

### DIFF
--- a/app/classes/query/observations.rb
+++ b/app/classes/query/observations.rb
@@ -67,6 +67,15 @@ class Query::Observations < Query::Base
     @model ||= Observation
   end
 
+  def alphabetical_by
+    @alphabetical_by ||= case params[:order_by].to_s
+                         when "user", "reverse_user"
+                           User[:login]
+                         when "name", "reverse_name"
+                           Name[:sort_name]
+                         end
+  end
+
   def self.default_order
     :date
   end
--- a/app/controllers/observations_controller/index.rb
+++ b/app/controllers/observations_controller/index.rb
@@ -212,12 +212,8 @@ class ObservationsController
         include: observation_index_includes
       }.merge(opts)
 
-      # Paginate by letter if sorting by user or name.
-      if %w[user reverse_user name reverse_name].include?(
-        query.params[:order_by]
-      )
-        opts[:letters] = true
-      end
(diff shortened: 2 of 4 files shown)


app/controllers/observations_controller/index.rb
```diff
@@ -204,11 +204,22 @@ def filtered_index_final_hook(query, _display_opts)
...
+      }.merge(opts)
+
+      # Paginate by letter if sorting by user or name.
+      if %w[user reverse_user name reverse_name].include?(
+        query.params[:order_by]
+      )
+        opts[:letters] = true
+      end
+
+      opts
```
nimmolo, line 222: I'd vote to reenable alpha sorting for Species Lists and not Observations. In that case this file (only) would not get reverted. 

That would answer Terri's request.

app/classes/query/observations.rb
```diff
@@ -67,6 +67,15 @@ def model
...
 
+  def alphabetical_by
+    @alphabetical_by ||= case params[:order_by].to_s
+                         when "user", "reverse_user"
+                           User[:login]
+                         when "name", "reverse_name"
+                           Name[:sort_name]
+                         end
+  end
+
```
nimmolo, line 78: Even if we are restoring aplha sorting for Spl only, as I would suggest, we still DO need these lines restored.

nimmolo (changes requested):

nimmolo:
@mo-nathan I saw Ron Pastorino's request. 

Maybe we can permit obs letter sorting only if there are query constraints. That will eliminate the case where we enable the crazy letter sorting on "all obs" but permit it for cases like his, "my observations", and any other case where it's a limited set.

If you like i can modify this PR to do that and you can check it out.

app/controllers/observations_controller/index.rb
nimmolo: Revising this recommendation to accommodate Ron Pastorino's note. 

Will amend this PR accordingly.

mo-nathan:
@nimmolo remind me again of where and why it's loading all the observations?  It doesn't seem like that should ever be necessary given that we have pagination and the feature is explicitly about things that only start with one letter.

PR #2896, opened 2025-04-11
Quick fix to mitigate query explosion

mo-nathan:
For reasons I don't fully understand, the names table is getting joined with the observations table for every query within the definition of scope :observation_query.  This calls `subquery` which in tern tries to find all the synonyms for a name, but when joined with observations it returns over 7,000 rows rather than just 2.  This PR puts in a strategicly placed `.distinct` to avoid the rest of the code from having to thinking about more that the 2 relevant names.  This probably still puts and unnecessary burden on the database, but I don't know of a better way to approach the problem.  @nimmolo may know a better way to handle this sort of case.  Maybe Name needs to be `unscoped`?

### DIFF
--- a/app/classes/lookup/names.rb
+++ b/app/classes/lookup/names.rb
@@ -138,7 +138,7 @@ class Lookup::Names < Lookup
 
     names.reject { |name| name[:synonym_id] } +
       Name.where(synonym_id: limited_id_set(name_ids)).
-      select(*minimal_name_columns)
+      distinct.select(*minimal_name_columns)
   end
 
   def add_subtaxa(names)


nimmolo:
A Name query that calls the scope `observation_query` seems like it must by definition join to Observations, because this scope only exists to get the names of a filtered observation query (i.e. what was formerly called a "coerced query").

There could be other more direct ways to get the same results, in simple cases. 

But when the query is comples, this should work as advertised. Say you have applied multiple filters to an Observation query:

```ruby
Query.new(:Observation, has_public_lat_lng: true, region: "California, USA", clade: "Russulaceae", project: foo) 
```

Now you want the names of those observations. 

Thats's what the scope is for.

mo-nathan:
What is surprising to me is that running a query inside the scope also gets the join.  For example, if I drop a break point in `add_synonyms` and from that break point I run `Name.first.text_name` it adopts the join.  I can understand that the ultimate query returned by the scope needs the join and hence would need a distinct, but I don't see why that's the case for a query that's run as part of the computation needed to create the query in question.  Maybe I'm thinking about it wrong and it's all just part of one big query that comes out of the scope, but that's a bit brain twisting to me.

nimmolo:
That seems weird to me too. Let me know when you want to do the Query orientation meeting so we're all on the same page.

nimmolo:
Merging this because it seems like a hot fix for an unreasonably expensive query, but we should revisit in the meeting tomorrow.

PR #2899, opened 2025-04-12
Add deleted Observation subquery scopes

### LINKED ISSUE #2897: Error showing Observations of Names search
Example:
- https://mushroomobserver.org/names?pattern=Entoloma, http://localhost:3000/names?pattern=Entoloma
- then click Show Observations . 
Result: NoMethodError in ObservationsController#index, undefined method `name_query’ for class Observation

JoeCohen:
- The scopes were accidentally dropped in e0e8deced
- Includes tests to prevent reversion
- Fixes #2897 

### Suggested Manual tests
- Name PatternSearch for `Entoloma` (http://localhost:3000/names?pattern=Entoloma); click `Show Observations`
Result: `Observations of names: [ search pattern: Entoloma ]`
- Location PatternSearch for `Cape Cod` (http://localhost:3000/locations?pattern=Cape+Cod); click `Show Observations`
Result: `Observations of locations: [ search pattern: Cape Cod ]`
- click Latest, Images (http://localhost:3000/images), click `Show Observations`
Result: `Observations of images: [ ]`

### DIFF
--- a/test/models/observation_test.rb
+++ b/test/models/observation_test.rb
@@ -1332,37 +1332,6 @@ class ObservationTest < UnitTestCase
     )
   end
 
-  def test_scope_image_query
-    assert_equal(
-      # Compare ids because comparing AR objects causes FAIL
-      # No visible difference in AR#inspect output.
-      Observation.where.not(thumb_image: nil).order(id: :asc).pluck(:id),
-      Observation.image_query({ order_by: "created_at" }).order(id: :asc).
-                  pluck(:id)
-    )
-  end
-
-  def test_scope_location_query
-    assert_equal(
-      # Compare ids because comparing AR objects causes FAIL
-      # No visible difference in AR#inspect output.
-      Observation.where(location: locations(:burbank)).order(id: :asc).
-                  pluck(:id),
-      Observation.location_query({ pattern: "Burbank" }).order(id: :asc).
-                  pluck(:id)
-    )
-  end
-
-  def test_scope_sequence_query
-    assert_equal(
-      # Compare ids because comparing AR objects causes FAIL
-      # No visible difference in  AR#inspect output.
-      Observation.joins(:sequences).order(id: :asc).uniq.pluck(:id),
-      Observation.sequence_query({ order_by: "created_at" }).order(id: :asc).
-                  pluck(:id)
-    )
-  end
-
   def nybg
     @nybg ||= locations(:nybg_location)
   end
--- a/test/models/observation_test.rb
+++ b/test/models/observation_test.rb
@@ -1332,6 +1332,37 @@ class ObservationTest < UnitTestCase
     )
   end
 
+  def test_scope_image_query
+    assert_equal(
+      # Compare ids because comparing AR objects causes FAIL
+      # No visible difference in AR#inspect output.
+      Observation.where.not(thumb_image: nil).order(id: :asc).pluck(:id),
+      Observation.image_query({ order_by: "created_at" }).order(id: :asc).
+                  pluck(:id)
+    )
+  end
+
+  def test_scope_location_query
+    assert_equal(
+      # Compare ids because comparing AR objects causes FAIL
+      # No visible difference in AR#inspect output.
+      Observation.where(location: locations(:burbank)).order(id: :asc).
+                  pluck(:id),
+      Observation.location_query({ pattern: "Burbank" }).order(id: :asc).
+                  pluck(:id)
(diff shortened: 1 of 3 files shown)


test/models/observation_test.rb
```diff
@@ -1332,6 +1332,37 @@ def test_scope_location_undefined
     )
   end
 
+  def test_scope_image_query
+    assert_equal(
+      # Compare ids because comparing AR objects causes FAIL
+      # No visible difference in AR#inspect output.
+      Observation.where.not(thumb_image: nil).order(id: :asc).pluck(:id),
+      Observation.image_query({ order_by: "created_at" }).order(id: :asc).
```
nimmolo, line 1340: I'd recommend putting these in `Query::ObservationsTest`, because every consequential scope test is now also a Query test. Maybe also test something more distinctive of Images in the image query, like `copyright_holder_has`. 

Anyway, something like this:
```ruby
    images = Image.copyright_holder_has("rolf").pluck(&:id) 
    assert_query(Observation.where(thumb_image: images).order_by_default,
                 :Observation, image_query: { copyright_holder_has: "rolf" })
```

For the AR object comparison to work, i think you need to call `order_by_default` on the first one, not `order(id: :asc)`. 

FWIW I think you can get away without the brackets when calling image_query directly: 
`image_query({ order_by: "created_at" })`
but i don't object to them.

JoeCohen: @nimmo Thanks!  I'll move the tests and use your improved image_query test. 
Will also drop the brackets if that works.

nimmolo: Ok. Note: Brackets ARE needed in the Query `assert_query` test syntax, where it's just describing the Query.

nimmolo: If you have a tight enough scope that you know exactly what fixtures you're expecting, you can use the new alternative `assert_query_scope`, that tests both the query and the scope against an array of defined results. Examples abound.

nimmolo: The key thing to know is that calling a Query now simply calls the scope. That's why `assert_query` in my example above effectively tests the scope, and I'd recommend using that over `assert_equal` because it tests the real use case.

`assert_query_scope` was cooked up for the transition, so i could be sure the scopes matched the old Query SQL. But at this point it's redundant... or you could call it a sanity check. 

In any case, you can be sure that Query calls the scope... because it hardly does anything else anymore. We'll go over that in the meeting.

nimmolo: One other thing to know when testing with `assert_query`. 

Because Query now calls scopes, the following test is now meaningless:
```ruby
assert_query(Observation.image_query(copyright_holder_has: "rolf").order_by_default,
             :Observation, image_query: { copyright_holder_has: "rolf" })
```
Those two lines are identical by definition.

So your expects need to be phrased differently, as in my example in the earlier comment above. 
Here it is rephrased using `assert_query_scope`, with the (redundant) scope:
```ruby
images = Image.copyright_holder_has("rolf").pluck(&:id) 
expects = Observation.where(thumb_image: images).order_by_default
scope = Observation.image_query(copyright_holder_has: "rolf").order_by_default
assert_query_scope(scope, expects,
             :Observation, image_query: { copyright_holder_has: "rolf" })
```
This tests the both the scope and the query against the expects, which are here phrased differently from the scope. 

Even more rigorous, but not always practical, is to provide the results array — an array of expected fixtures — as the `expects`. You'll find many Query tests that do that... often you can arrange your test so it produces a shorter array of results.

nimmolo: The way I found my result fixtures array for whatever test was usually to assert against an empty array. 

This should fail, but the resulting error message will give you the results produced, and if you're pretty sure they are correct, you can use those in the test.

nimmolo: Corrected a typo in my `assert_query_scope(expects, scope,` example.

JoeCohen:
@nimmo Is anything actually using `Observation.sequence_query`?
(I can't think of a way that a user could invoke it.)

nimmolo:
I believe there may be a sequence index link that does use it.

nimmolo:
Can't find it in the code, though. I remember trying to delete it at some point and restoring it, i feel like "observations of these sequences" is something we may need at some point, but i could be wrong.

nimmolo:
Any params that you could filter sequences by: locus, user, etc.

nimmolo:
Merging now because it's raising errors on production.

PR #2901, opened 2025-04-13
Simplify circuitous Name and Location subqueries

nimmolo:
### Name subqueries

`:observation_query` — `:names`, `:clade`, `:lichen`

This changes Name subquery scopes to apply any `:names`,  `:clade`, or `:lichen` param within an `:observation_query` to the outer Name scope directly. In the case where the subquery has that param, it is made to ignore these "sub-params".

#### The issue:
If a user's query starts out as a Name query, and the user clicks "Observations of these names", and then "Names of these observations", our link builders would return them to the original query, because the name params would be cached in the query description as a `subquery`. Our link builder just restores the original Name params. 

But in the case where the query originates as an Observation query, a user's filtering params like `:names`, `:clade` or `:lichen`  are essentially pseudo-subqueries of Name. If the user clicks on "Names of these observations", these params could be more efficiently used as Name filters, when converting the Observation query to a Name query.

Currently an automatically generated link for "Names of Observations (from a user's previous names filter)" calls:
```ruby
Name.observation_query(names: { lookup: "Peltigera", include_synonyms: true })
```
Obviously this is a ridiculous query, but it's possible given the UI, which must be agnostic about query params in providing "related records" links, because any filter could be in the subquery. So the link in this circumstance is currently producing the AR:
```ruby
Name.joins(:observations).distinct.
merge(Observation.names(lookup: "Peltigera", include_synonyms: true))
```
Note the double lookup of names through observations. This is producing a heavy query:
```
  Name Load (863.5ms)  
SELECT `names`.`id`, `names`.`correct_spelling_id`, `names`.`synonym_id`, `names`.`text_name` 
FROM `names` INNER JOIN `observations` ON `observations`.`name_id` = `names`.`id` 
WHERE `names`.`text_name` = 'Peltigera'

  Name Load (248.6ms)  
SELECT DISTINCT `names`.`id`, `names`.`correct_spelling_id`, `names`.`synonym_id`, `names`.`text_name` 
FROM `names` INNER JOIN `observations` ON `observations`.`name_id` = `names`.`id` 
WHERE `names`.`synonym_id` = 1878

  Name Load (239.6ms)  
SELECT DISTINCT `names`.* 
FROM `names` INNER JOIN `observations` ON `observations`.`name_id` = `names`.`id` 
WHERE `observations`.`name_id` = 5077
```
#### Proposed solution:

This PR changes the behavior of `observation_query` so the same call,
```ruby
Name.observation_query(names: { lookup: "Peltigera", include_synonyms: true })
```
now generates:
```ruby
Name.names(lookup: "Peltigera", include_synonyms: true).joins(:observations).distinct
```
producing a much lighter
```
  Name Load (27.6ms)  
SELECT `names`.`id`, `names`.`correct_spelling_id`, `names`.`synonym_id`, `names`.`text_name` 
FROM `names` WHERE `names`.`text_name` = 'Peltigera'

  Name Load (0.3ms)  
SELECT DISTINCT `names`.`id`, `names`.`correct_spelling_id`, `names`.`synonym_id`, `names`.`text_name` 
FROM `names` WHERE `names`.`synonym_id` = 1878

  Name Load (261.9ms) 
 SELECT DISTINCT `names`.* 
FROM `names` INNER JOIN `observations` ON `observations`.`name_id` = `names`.`id` 
WHERE (`names`.`id` IN (5077, 9794)) AND (`names`.`correct_spelling_id` IS NULL)
```
___
`:description_query` — `:names`

Although a `:description_query` is cheap compared to an `:observation_query`, this treats them both the same to simplify the code. It also fixes how `NameDescription` queries parse a `:names` param. (This wasn't working, or tested, for subqueries as currently set up).

Makes sure the `NameDescriptionAPI` sends that param the same way it is sent to `Query` by other models of the API. 
NOTE: This does not change any outward-facing behavior in the API, it only changes its internal communication with `Query`.
___
### Location subqueries

`:observation_query` — `:locations`, `:region` Same idea here.

### DIFF
--- a/app/classes/api2/name_description_api.rb
+++ b/app/classes/api2/name_description_api.rb
@@ -37,9 +37,8 @@ class API2
         created_at: parse_range(:time, :created_at),
         updated_at: parse_range(:time, :updated_at),
         by_users: parse_array(:user, :user, help: :first_user),
-        names: parse_array(:name, :name),
         is_public: true
-      }
+      }.merge(parse_names_parameters)
     end
 
     def post
--- a/app/classes/query/base.rb
+++ b/app/classes/query/base.rb
@@ -65,11 +65,11 @@ class Query::Base
 
   delegate :content_filter_parameters, to: :class
 
-  def self.subquery_parameters
-    parameter_declarations.select { |key, _v| key.to_s.include?("_query") }
-  end
+  # def self.subquery_parameters
+  #   parameter_declarations.select { |key, _v| key.to_s.include?("_query") }
+  # end
 
-  delegate :subquery_parameters, to: :class
+  # delegate :subquery_parameters, to: :class
 
   # Can the current class be called as a subquery of the target Query class?
   def relatable?(target)
--- a/app/classes/query/base.rb
+++ b/app/classes/query/base.rb
@@ -65,6 +65,12 @@ class Query::Base
 
   delegate :content_filter_parameters, to: :class
 
+  def self.subquery_parameters
+    parameter_declarations.select { |key, _v| key.to_s.include?("_query") }
+  end
+
+  delegate :subquery_parameters, to: :class
+
   # Can the current class be called as a subquery of the target Query class?
   def relatable?(target)
     self.class.related?(target, model.name.to_sym)
--- a/app/classes/query/name_descriptions.rb
+++ b/app/classes/query/name_descriptions.rb
@@ -13,7 +13,11 @@ class Query::NameDescriptions < Query::Base
       sources: [{ string: Description::ALL_SOURCE_TYPES }],
       ok_for_export: :boolean,
       content_has: :string,
-      names: [Name],
+      names: { lookup: [Name],
+               include_synonyms: :boolean,
+               include_subtaxa: :boolean,
+               include_immediate_subtaxa: :boolean,
+               exclude_original_names: :boolean },
       projects: [Project],
       name_query: { subquery: :Name }
     )
--- a/app/models/abstract_model/scopes.rb
+++ b/app/models/abstract_model/scopes.rb
@@ -275,6 +275,8 @@ module AbstractModel::Scopes
   module ClassMethods
     # Utility for all subqueries, which are defined on the model
     def subquery(model_name, params)
+      return all if params.blank?
+
       subquery = Query.new(model_name, **params).scope.reorder("")
       merge(subquery).distinct
     end
--- a/app/models/abstract_model/scopes.rb
+++ b/app/models/abstract_model/scopes.rb
@@ -274,6 +274,8 @@ module AbstractModel::Scopes
   # class methods here, `self` included
   module ClassMethods
     # Utility for all subqueries, which are defined on the model
+    # Callers must do their own joins to `model_name` because we can't know
+    # whether the association (has_one, has_many) is singular or plural.
     def subquery(model_name, params)
       return all if params.blank?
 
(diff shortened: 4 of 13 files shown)


nimmolo:
Although any description subquery is cheap compared to an observation subquery, this treats them both the same to simplify the code.

As a result, I had to fix how NameDescription queries parse the names param (it wouldn't have even worked as currently set up), and make sure the NameDescription API sends that param the same way it is sent to Query by other models of the API. 

NOTE: This does not change any outward-facing behavior in the API, it's only internal communication with Query.

nimmolo:
Going to move this logic to the Name scopes, so Query can remain dumb, and so the scopes will still work as expected.

nimmolo:
Need to do the same for params `clade`, `lichen`, `locations` and `region`.

app/classes/api2/name_description_api.rb
```diff
@@ -37,9 +37,8 @@ def query_params
         created_at: parse_range(:time, :created_at),
         updated_at: parse_range(:time, :updated_at),
         by_users: parse_array(:user, :user, help: :first_user),
-        names: parse_array(:name, :name),
         is_public: true
-      }
+      }.merge(parse_names_parameters)
```
nimmolo, line 41: This is a capability addition, rather than a breaking change. It enables NameDescription calls to also parse `include_synonyms` etc., as the API does for observation queries and name queries. 

Necessary, because for Query to be able to convert a Name query to a NameDescription query, `Query::NameDescriptions` must be able to handle the lookup params. That means `NameDescriptionAPI` must be able to send user input as a hash of lookup params to Query. This doesn't change how a user would use the API, it changes how the API packages the user call and sends it to Query.

app/classes/query/base.rb
```diff
@@ -65,6 +65,12 @@ def self.content_filter_parameters
 
   delegate :content_filter_parameters, to: :class
 
+  # def self.subquery_parameters
+  #   parameter_declarations.select { |key, _v| key.to_s.include?("_query") }
+  # end
+
+  # delegate :subquery_parameters, to: :class
+
```
nimmolo, line 73: Keeping these accessors as comments because 
- I will need them on the faceted search PR to distinguish params that should go in the form from subquery params
- I keep adding and deleting these methods on various PRs, and each time i forget the syntax

app/classes/query/name_descriptions.rb
```diff
@@ -13,7 +13,11 @@ def self.parameter_declarations
       sources: [{ string: Description::ALL_SOURCE_TYPES }],
       ok_for_export: :boolean,
       content_has: :string,
-      names: [Name],
+      names: { lookup: [Name],
+               include_synonyms: :boolean,
+               include_subtaxa: :boolean,
+               include_immediate_subtaxa: :boolean,
+               exclude_original_names: :boolean },
```
nimmolo, line 20: Change to treat a `:names` param the same on all Query models.

app/models/abstract_model/scopes.rb
```diff
@@ -274,7 +274,11 @@ module AbstractModel::Scopes
   # class methods here, `self` included
   module ClassMethods
     # Utility for all subqueries, which are defined on the model
+    # Callers must do their own joins to `model_name` because we can't know
+    # whether the association (has_one, has_many) is singular or plural.
     def subquery(model_name, params)
+      return all if params.blank?
+
```
nimmolo, line 281: The new Name/Location subquery parsers may produce an empty params hash, so we now need this guard clause.

PR #2911, opened 2025-04-15
Query — switch to attributes

nimmolo:
- Monkey patches `Class` to add a new attribute declaration method for `Query`, `:query_attr`. 
- Adds a new explicit option for the `:query_param` attribute type, `:accepts`, which our Query validator uses to figure out how to validate the attribute. (It's the same arg previously stored in the `parameter_declarations` hash.)
- Changes Query subclasses to declare attributes using `query_attr`, and changes callers of `parameter_declarations` to access Query `attribute_types` instead. `@params` are still used as the query's "active attributes" (i.e. non-nil).

Using ActiveModel attributes makes it possible for attribute declarations, and our validator, to use other inherited options: `precision`, `limit` and `scale`, that we don't declare on Query attributes yet, but could.

### DIFF
--- a/app/classes/query/base.rb
+++ b/app/classes/query/base.rb
@@ -269,10 +269,10 @@ class Query::Base
   end
 
   # Serialize the query params, adding the model, for saving to a QueryRecord.
-  # The :description column is accessed not just to recompose a query, but to
-  # identify existing query records that match current params. That's why the
-  # keys are sorted here before being stored as strings in to_json - because
-  # when matching a serialized hash, strings must match exactly.
+  # We use this column of QueryRecord to identify an existing query record that
+  # matches current params, and sometimes to recompose a query from the string.
+  # That's why the keys are sorted here before being serialized via `to_json` -
+  # when matching, strings must match exactly.
   #
   # NOTE: QueryRecord[:description] is not a Rails-serialized column; we call
   # `to_json` here to serializate it ourselves. Using SQL to compare serialized
--- a/app/classes/query/base.rb
+++ b/app/classes/query/base.rb
@@ -73,10 +73,13 @@
 #  the attribute should be validated. (We don't differentiate between
 #  `query_param` types at the attribute level because they are all validated
 #  recursively, and nested values may use the same methods as top-level values.)
+#  To keep this from getting too verbose, we assign Query attributes using the
+#  custom method `query_attr` whose second argument is the value of `:accepts`.
+#  See app/extensions/class.rb for the definition of `query_attr`.
 #
-#  We use a special syntax in `:accepts` to declare the validation type of each
-#  attribute. The syntax is important because it tells
-#  Query::Modules::Validation how to parse the attribute value.
+#  `:accepts` expects a special syntax that declares a validation type for each
+#  attribute, and tells Query::Modules::Validation how to parse the attribute
+#  value. It uses the following patterns:
 #
 #  ### Simple values
 #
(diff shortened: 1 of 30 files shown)


nimmolo:
@JoeCohen @mo-nathan — Please examine the attribute declarations for Observation, Name, and Location. 

To keep this DRY, i'm batch-declaring the advanced-search and content-filter attributes the old way, because these are shared between models. But i'm putting the same declarations in the list, commented out, because it confused me many times when working on Query, not remembering that a param was declared by a filter or advanced search. I'm trying to split the difference between DRY and having the attributes explicit and readable.

If i had to choose between deleting the comments and going against DRY, though, I'd go for explicitly declaring the attributes in the model. What do you guys think?

JoeCohen (approved):
It looks good to me. The comments do a good job of explaining how to add more models, params, and subquery's.

I didn't completely understand your question about deleting comments vs. going dry. 
I like explicitly declaring the attributes in the model.

JoeCohen:
Gotcha. Suggestion: 
- Keep the comments
- Prepend them with a comment explaining why you're keeping them.

app/classes/query/base.rb
```diff
@@ -192,44 +203,51 @@ class Query::Base
...
   end
 
   delegate :scope_parameters, to: :class
 
+  # returns keys
   def self.content_filter_parameters
     filters = Query::Filter.all
-    @content_filter_parameters ||= filters.each_with_object({}) do |f, p|
-      p[f.sym] = f.type
+    @content_filter_parameters ||= filters.each_with_object([]) do |f, ary|
```
mo-nathan, line 232: Would this be better as a Set?

app/classes/query/base.rb
```diff
@@ -192,44 +203,51 @@ class Query::Base
 
   attr_writer :record
 
-  def self.parameter_declarations
-    { order_by: :string }
+  query_attr(:order_by, :string)
+
+  def self.attribute_types
+    super.symbolize_keys!
```
mo-nathan, line 209: This seems surprising to me.  Why is this needed?

app/classes/query/base.rb
```diff
@@ -124,38 +120,53 @@
 #    [User]
 #    [Location]
 #
-#  In some cases the array may be ultimately parsed in the scope as a duration
-#  or range of values (e.g. a range of dates, ranks, vote values, etc.). In this
-#  case, two values is the maximum length of the array; more will be ignored.
-#  But for arrays of model instances like [Name], any number of instances or
-#  ids can be passed.
+#  In some cases the scope may be configured to parse the array as a range of
```
mo-nathan, line 123: Would it make sense over to time to replace this usage of array style attributes with hash style attributes, e.g., `{ start: :time, end: :time }`.  Seems like that would help avoid having to dig into the scopes to understand how the array is interpreted.

mo-nathan (approved):
Overall this looks great!  A few minor comments which may best be left for a later PR.

app/classes/query/base.rb
nimmolo: It was surprising to me too. Rails returns the `attribute_types` hash with string keys, not "indifferent access", and accessors with symbols fail. It's not the only thing like that and I got tired of the gotcha.

app/classes/query/base.rb
nimmolo: Done

app/classes/query/base.rb
nimmolo: Hmmm - maybe, but maybe not for forms. I was balancing specificity here with the boilerplate i would have to write for a "range" form field, and in this case it seemed way easier to just have a text field that accepts two date strings. But i haven't rewritten the faceted form for Query yet. I will keep this in mind in case it does turn out to be helpful.

Btw, I believe the scopes I wrote are all consistent in parsing ranges, though someone could write one differently.

PR #2915, opened 2025-04-17
Eliminate User.current from NamesControllertTest

mo-nathan:
In theory this should cause no user visible changes.  Note that I suspect the changes to some of the mailer templates may actually provide a slight improvement for users using "scientific" location formatting since the name formatting is now based on the recipient rather than User.current.

As a side note, my initial approach was to put the `user` argument at the end of various models and POROs when needed.  However, due to the Rubocop, Style/OptionalArguments: Optional arguments should appear at the end of the argument list, I moved it to the front of the parameters in several cases.  This is probably a better convention, but I'm not sure it's worth rejiggering.  Opinions appreciated.

Manual testing of the field slip functionality would be very helpful.

The way I came to this PR is by adding a debugger breakpoint in the readers for User.current and User.current_id.  Then resolving each place where User.current or User.current_id gets called.  If a signature change is required (typically passing user) and it's too much work to update all the callers in this PR, then create a separate function for that.


mo-nathan:
I believe the coverage drops are entirely to new `user_` versions of functions in app/models/observation.rb and app/models/name/format.rb.  These should get covered once the rest of the change over is done.

app/models/name/format.rb
```diff
@@ -8,6 +8,16 @@ def self.included(base)
...
+    str = self[:display_name]
+    if user &&
+       user.hide_authors == "above_species" &&
+       Name.ranks_above_species.include?(rank)
+      str = str.sub(/^(\**__.*__\**).*/, '\\1')
+    end
+    str
+  end
+
   def display_name
```
JoeCohen, line 21: Is this method right? No response needed
```ruby
  def display_name
    str = self[:display_name]
    if User.current &&
       User.current.hide_authors == "above_species" &&
       Name.ranks_above_species.include?(rank)
      str = str.sub(/^(\**__.*__\**).*/, '\\1')
    end
    str
  end
```

JoeCohen:
This could be an old issue. I could not add a member to a project which I own. 
- Went to http://localhost:3000/projects/353/members
- Began typing Nathan W in the `Login Name` box
Screenshot 2025-04-17 at 7 03 28 AM

- Clicked on the `Nathan Wilson (nathan)` line in the dropdown.
Screenshot 2025-04-17 at 7 08 00 AM

- Clicked Add (under that box)
Result: Flash Error User “Nathan Wilson (nathan)” not found.
Screenshot 2025-04-17 at 7 11 53 AM

JoeCohen:
Is this expected/correct? (Again might be an old bug.)
Non-member of non-open project can:
- View Add member and Add alias buttons.
- Add and delete aliases.

Details
- Logged in as me (user 4468) in one browser (Chrome).
  - Went to http://localhost:3000/projects/353

- Logged in as Oregon Mycological Society (OMS) (user 5485) in another browser (Safari)
  - Went to http://localhost:3000/projects/353
  - Clicked the Members tab
Result: Display include Add buttons, although OMS is not a member.  Screenshot 2025-04-17 at 7 25 57 AM

  - Tried to add new member Nathan Wilson
  Result: Flash Error "Only the admins for a project can change member permissions."
  **The message seems off; as if OMS is a Member, but isn't an admin.** 
  Note: users cannot add themselves to this project. Screenshot 2025-04-17 at 7 37 17 AM
  - Under Project Aliases, clicked Add
  - For user `Joseph D. Cohen (Joe Cohen)` typed `blah` in the Name box, then Add
  Result: Added the alias. 
  **Seems wrong that a non-member can add and delete aliases** Screenshot 2025-04-17 at 7 47 14 AM

mo-nathan:
Those don't sound related to this specific PR.  You can verify that by doing the same steps on main.  Does seems like it would be worth creating GitHub issues for them if they are also issues on the current main branch.

app/models/name/format.rb
mo-nathan: It looks reasonable to me.  What are you concerned about?

JoeCohen:
I'll create issues, also do a little more manual testing.
But can I really stress test this? 
Feels like two project members need to be logged in simultaneously.
I don't know how to do that without adding a member to a project I own (or administer). But I can't do that.

nimmolo:
Ah - your idea for `@current_user` is to have that be an ivar of the model, while `@user` continues to be the ivar in controllers and templates. I don't completely get it yet, just noting the pattern.

mo-nathan:
@JoeCohen I don't think you need to worry about stress testing yet.  It's not the only thing between us and multi-threading which is when that could be important.  I don't believe anything I've done so far which be noticeable from a performance perspective.

Issue #2908, opened 2025-04-14
Destroy Comment Error


JoeCohen:
Clicking Destroy Comment throws an Error. (it does destroy the Comment.)

nimmolo:
I think it's likely we're forgetting to start or restart Redis on production.
This error is "Redis::CannotConnectError (Error connecting to Redis on localhost:6379 (Errno::ECONNREFUSED)):". 

My proposed solution is to switch to the gem `SolidCable` and get rid of our Redis dependency.

PR #2920, opened 2025-04-18
Remove Redis dependency on production, add gem `solid_cable`

nimmolo:
See issue #2908. 
This aims to resolve an issue deleting comments that is currently throwing errors on production. The errors seem to be due to Redis not being booted, or not connecting for some other reason. 

The error is reproducible locally: just add and delete a comment from an obs. This PR resolves the error locally for me.

Having to install/deal with/reboot Redis, whether on production or locally, is proving to be a banana peel. The good news is that we don't have to use it anymore. There is a core Rails gem, installed by default in Rails 8, called [`solid_cable`](https://github.com/rails/solid_cable) that runs ActionCable off its own small MySQL database and gets rid of the Redis dependency for Turbo updates. Eliminating Redis was one of the goals of `solid_queue` and `solid_cache`, which we already use, and it seems to me to make sense to add this gem, now that we are having issues on production with comment updates/deletions.

I believe this is our last `redis` dependency, but i'm not sure. I believe removing Redis should be a separate PR in case it breaks anything.

NOTE: After deploy, we will need to 
- alter production's `database.yml`:
```
  cable:
    database: cable_production
    username: $
    password: $
    host: $
    migrations_paths: db/cable_migrate
```
- run `rails db:prepare`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,8 @@ gem("solid_queue")
 # https://github.com/rails/mission_control-jobs
 # Rails-based frontend to Active Job adapters for monitoring jobs
 gem("mission_control-jobs")
+# solid_cable for ActionCable without Redis
+gem("solid_cable")
 
 # sprockets for asset compilation and versioning
 gem("sprockets-rails")
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -469,6 +469,11 @@ GEM
     snaky_hash (2.0.1)
       hashie
       version_gem (~> 1.1, >= 1.1.1)
+    solid_cable (3.0.7)
+      actioncable (>= 7.2)
+      activejob (>= 7.2)
+      activerecord (>= 7.2)
+      railties (>= 7.2)
     solid_cache (1.0.7)
       activejob (>= 7.2)
       activerecord (>= 7.2)
@@ -605,6 +610,7 @@ DEPENDENCIES
   rubyzip (~> 2.3.0)
   simplecov
   simplecov-lcov
+  solid_cable
   solid_cache
   solid_queue
   sorted_set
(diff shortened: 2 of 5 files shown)


JoeCohen (commented):
Looks like a great improvement. Thanks for taking the lead.
A few questions. If it's easier, answer them Monday.
- Should `redis` be removed from Gemfile?
- Will the PR affect mission control? 
- Will it affect jobs that are running on production at the moment we switch over?
- Will we be able to access information about older jobs, e.g., via mission control. 
- What changes to I need to make locally? 
  - Will jobs simply run without the need for redis?
  - When I update the local db I do `rails db:drop:cache DISABLE_DATABASE_ENVIRONMENT_CHECK=1; rails db:create:cache; rails db:migrate:cache` (among other things). Do I need to change that?
  - Anything else I should do locally to get on board?
  - Any changes needed/desirable to README_MACOSX_NOTES.md

JoeCohen:
PS: Any changes needed to CI workflow? `.github/workflows/ci_rails.yml`

nimmolo:
- This doesn't affect jobs or Mission Control. It is separate from Solid Queue, although inspired by the same impulse of simplifying production dependencies. It only affects Turbo broadcasts, and currently I believe that's only Comments. 

- I'm not sure it eliminates the redis dependency, so i feel that should be a separate PR. 

- To get on board locally, I think you only have to run this branch, update your `database.yml` according to the changes in the README on this branch (note that `database.yml` is not in version control, and note this enables solid_cable in dev so we can test it manually), and run `rails db:prepare`. That's it. 

It's working if you can delete a comment manually locally.

nimmolo:
Whoops - i didnt commit the README changes yet. Coming up.

JoeCohen:
Will the real Coveralls please stand up?

Coveralls says: "Coverage remained the same."
Screenshot 2025-04-18 at 3 13 27 PM

Coveralls says: "Coverage decreased (-0.003%) to 94.649%"
Screenshot 2025-04-18 at 3 16 33 PM
and that one line got dropped in `app/views/controllers/projects/aliases/_form.erb`. See https://coveralls.io/builds/73294948.

PR #2928, opened 2025-04-20
Update Symbiota report

JoeCohen:
Updates Symbiota `report` in preparation for MyCoPortal export

### DIFF
--- a/app/classes/report/symbiota.rb
+++ b/app/classes/report/symbiota.rb
@@ -2,7 +2,14 @@
 
 module Report
   # Symbiota-style tsv report.
+  # https://symbiota.org/
+  # https://biokic.github.io/symbiota-docs/
+  # https://github.com/Symbiota/Symbiota
   class Symbiota < TSV
+    # These plus Column labels comprising
+    # Symbiota Standard Field names which are useful for our purposes
+    # https://biokic.github.io/symbiota-docs/editor/edit/fields/#standard-fields
+    # plus MO-specific fields that are useful for uploads to Symbiota portals
     def labels
       %w[
         scientificName
@@ -36,7 +43,7 @@ module Report
       ]
     end
 
-    def format_row(row)
+    def format_row(row) # rubocop:disable Metrics/AbcSize
       [
         row.name_text_name,
         row.name_author,
@@ -45,7 +52,7 @@ module Report
         row.species,
         row.form_or_variety_or_subspecies,
         collector(row),
-        number(row),
+        number(row), # collectors number || "MUOB #{observation.id}", Cf. obs_id
         disposition(row),
         row.obs_when,
         row.year,
@@ -61,11 +68,11 @@ module Report
         row.best_high,
         row.obs_updated_at,
         substrate(row),
-        host(row),
-        field_notes(row),
-        row.obs_id,
-        row.obs_url,
-        image_urls(row)
+        host(row), # MyCoPortal `host` == Sybiota/DWC associatedTaxa
+        field_notes(row), # occurrenceRemarks
+        row.obs_id, # MCP `dpk`; catalogNumber = "MUOB #{observation.id}"
+        row.obs_url, # MO-specific; used in MCP Desciption / verbatimAttributes
+        image_urls(row) # MO-specific
       ]
     end
 
--- a/app/classes/report/symbiota.rb
+++ b/app/classes/report/symbiota.rb
@@ -44,7 +44,8 @@ module Report
         row.genus,
         row.species,
         row.form_or_variety_or_subspecies,
-        *collector_and_number(row),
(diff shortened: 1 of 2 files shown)


app/classes/report/symbiota.rb
```diff
@@ -2,7 +2,14 @@
...
   # Symbiota-style tsv report.
+  # https://symbiota.org/
+  # https://biokic.github.io/symbiota-docs/
+  # https://github.com/Symbiota/Symbiota
   class Symbiota < TSV
+    # These plus Column labels comprising
+    # Symbiota Standard Field names which are useful for our purposes
+    # https://biokic.github.io/symbiota-docs/editor/edit/fields/#standard-fields
+    # plus MO-specific fields that are useful for uploads to Symbiota portals
     def labels
```
JoeCohen, line 13: Is it possible to create a single object which defines, in a single line, (1) a column label and (2) the value of that column for a row, sort of like this fantasy:
```ruby
def labeled_column_value(row)
  {
    scientificName: row.name_text_name,
    scientificNameAuthorship: row.name_author,
    ...
    recordedBy: collector(row),
    etc.
  }
end
(`label` and `formatted_row` would be derived from the above.)

This would make it easer to write.

Comment (JoeCohen) [inline app/classes/report/symbiota.rb:13]: I'd like to standardized the labels to be Symbiota field names (if possible).
Proposed changes
- `host` (obsolete) => `associated_taxa` ([the official Symbiota field name](https://biokic.github.io/symbiota-docs/editor/edit/fields/#associatedTaxa)).
- `fieldNotes` (a description of the contents) => [`occurrenceRemarks`](https://biokic.github.io/symbiota-docs/editor/edit/fields/#occurrenceRemarks) 
- `mushroomObserverId` (not a Symbota field) => [`recordNumber`](https://biokic.github.io/symbiota-docs/editor/edit/fields/#recordNumber)
-  Also add [`fieldnumber`](https://biokic.github.io/symbiota-docs/editor/edit/fields/#fieldnumber) with a value of collector's number / field slip number

**Comment by JoeCohen** [comment]
General question about overall approach: 
This PR is preparation for a report to be used in exports to MyCoPortal. Which is best:
- Create a `Mycoportal` report as subclass of `Symbiota`;
- Replace `Symbiota` with `Mycoportal`;
- Use `Symbiota` plus added post-processing for exports to MyCoPortal.

**Comment by JoeCohen** [comment]
@pellaea Any suggestions:

> General question about overall approach: This PR is preparation for a report to be used in exports to MyCoPortal. Which is best:
> 
> * Create a `Mycoportal` report as subclass of `Symbiota`;
> * Replace `Symbiota` with `Mycoportal`;
> * Use `Symbiota` plus added post-processing for exports to MyCoPortal.

File: app/classes/report/symbiota.rb (hunk 1 above)
Comment (JoeCohen) [reply app/classes/report/symbiota.rb:13]: @nathan: 
> Is it possible to create a single object which defines, in a single line, (1) a column label and (2) the value of that column for a row, sort of like this fantasy:
> 
> ```ruby
> def labeled_column_value(row)
>   {
>     scientificName: row.name_text_name,
>     scientificNameAuthorship: row.name_author,
>     ...
>     recordedBy: collector(row),
>     etc.
>   }
> end
> ```
> 
> (`label` and `format_row` would be derived from the above.) This would be easier to read and write than the existing long definitions of `label` and `format_row`.

File: app/classes/report/symbiota.rb [hunk 2]
```diff
@@ -36,15 +43,16 @@ def labels
...
+    def format_row(row) # rubocop:disable Metrics/AbcSize
       [
         row.name_text_name,
         row.name_author,
         row.name_rank,
         row.genus,
         row.species,
         row.form_or_variety_or_subspecies,
-        *collector_and_number(row),
+        collector(row),
```
JoeCohen, line 54: Changed/added rows so that there are the same number of lines as `labels`;
facilitates side-by-side visual comparison with `labels`.

JoeCohen:
For future reference. The ideal approach -- given enough time, which I don't have now -- could be:
- Fix Report::Darwin to use the official field names
- Make Symbiota a subclass of Darwin, changing only what's different from the superclass.
- Either:
  - Use the revised Symbiota to generate MyCoPortal uploads, OR
  - Create Mycoportal as subclass of Symbiota or Darwin, changing only what's different from the superclass.

> General question about overall approach: This PR is preparation for a report to be used in exports to MyCoPortal. Which is best:
> 
> * Create a `Mycoportal` report as subclass of `Symbiota`;
> * Replace `Symbiota` with `Mycoportal`;
> * Use `Symbiota` plus added post-processing for exports to MyCoPortal.

JoeCohen:
Replaced by #2937

PR #2936, opened 2025-04-22
Eliminate User.current from NamesControllerTest

mo-nathan:
On this branch you can run the tests in `test/controllers/names_controller_test.rb` with no calls to `User.current` or `User.current_id`.  You can test this by putting a breakpoint in those functions and then running the tests.

The surgery here was much more extensive than in the FieldSlipsControllerTest case.  This touched on a lot of functionality and getting all the tests running again was it's own special challenge.


mo-nathan:
Just ran a quick test and this PR takes the test suite down to 15281 calls to User.current or User.current_id.  Before the field slip test change we were at 46149 calls.  After that PR we dropped to 31908.

mo-nathan:
Went through the Coveralls concerns and commented out a bunch of what appears to now be dead code.  Worth reviewing.

test/jobs/inat_import_job_test.rb
```diff
@@ -425,7 +425,7 @@ def test_import_job_prov_name_pnw_style
     new_names.each do |new_name|
       assert_equal(
         @user, new_name.user,
-        "#{new_name.text_name} author should be #{@user.login}"
+        "#{new_name.text_name} owner should be #{@user.login}"
```
JoeCohen, line 428: Good catch

app/helpers/images_helper.rb
```diff
@@ -103,15 +103,6 @@ def show_image_copyright?(image, object)
...
 
-  def show_best_image(user, obs)
-    return unless obs&.thumb_image
-
-    interactive_image(user, obs.thumb_image,
-                      link: observation_path(id: obs.id),
-                      size: :small,
-                      votes: true) + image_copyright(obs.thumb_image, obs)
-  end
-
```
nimmolo: This would be useful to me if we can keep this method commented-out with a note:
```ruby
# For use with matrix box carousels
```

app/helpers/tabs/observations_helper.rb
```diff
@@ -5,12 +5,12 @@ module Tabs
...
-        observation_manage_lists_tab(obs, user),
-        *obs_change_tabs(obs)&.reject(&:empty?)
-      ]
-    end
+    # def show_observation_tabs(obs:, user:)
+    #   [
+    #     observation_manage_lists_tab(obs, user),
+    #     *obs_change_tabs(obs)&.reject(&:empty?)
+    #   ]
+    # end
```
nimmolo, line 13: This can definitely be deleted - these are all now available via icon links in show_obs.

app/controllers/admin/emails/webmaster_questions_controller.rb
```diff
@@ -40,8 +40,8 @@ def create_webmaster_question
         elsif non_user_potential_spam?
           flash_error(:runtime_ask_webmaster_antispam.t)
         else
-          QueuedEmail::Webmaster.create_email(sender_email: @email,
-                                              content: @content)
+          QueuedEmail::Webmaster.create_email(@user, sender_email: @email,
+                                                     content: @content)
```
nimmolo, line 44: If the `QueuedEmail` builder now gets the `@user`, I don't think it needs to get `sender_email` too, it can derive sender_email from `@user`. (see action `new` above, where `@email` is defined.)

app/controllers/names_controller.rb
```diff
@@ -504,11 +512,12 @@ def update_correct_spelling
...
 
   def set_correct_spelling
-    correct_name = Name.find_names_filling_in_authors(@correct_spelling).first
+    correct_name = Name.find_names_filling_in_authors(@user,
+                                                      @correct_spelling).first
     raise(:runtime_form_names_misspelling_bad.t) unless correct_name
     raise(:runtime_form_names_misspelling_same.t) if correct_name.id == @name.id
 
-    @name.mark_misspelled(correct_name)
+    @name.mark_misspelled(@user, correct_name)
```
nimmolo, line 520: To me it seems a little unintuitive to make `user` now the first arg of a name method. Thankfully I don't deal with these much, but if i were calling `@name.mark_misspelled(*args)` I would expect `(correct_name, @user)`

app/models/name_sorter.rb
```diff
@@ -254,40 +254,13 @@ def add_name(spl_line, timestamp = nil)
     return unless name_parse.has_synonym
 
     @has_new_synonyms = true
-    if name_parse.find_synonym_names.empty?
+    if name_parse.find_synonym_names(user).empty?
       @new_name_strs.push(name_parse.synonym_search_name)
     end
     # Keep names in addition to parse for the chosen filter
     @synonym_data.push([name_parse, names])
```
nimmolo, line 261: It looks like the attribute reader `@synonym_data` is no longer needed in this class, if the method `create_new_synonyms` is going away (below). That's the only place it's used.

app/controllers/names_controller.rb
nimmolo: I guess you're doing it everywhere, and trying to foreground the fact that this method depends on the current user.

JoeCohen (approved):
I eyeballed the code. Nothing bad jumped out.
Also did some manual testing, found only an old bug which I wrote up as #2938

nimmolo (changes requested):
A few changes and a couple questions. This is an enormous amount of work, I can only imagine.

app/helpers/tabs/observations_helper.rb
mo-nathan: Sure.

app/controllers/admin/emails/webmaster_questions_controller.rb
mo-nathan: I was trying to do the simplest convention that I could.  The challenge with putting the user argument at the end are functions that slurp up the rest of the arguments.  This case is an example where just looking at the single line of code is is not clear that `@email` is always exactly `@user.email`.  I'll review and see if there are any real inconsistencies.

app/controllers/names_controller.rb
```diff
@@ -342,12 +343,12 @@ def init_edit_name_form
...
       @misspelling      = @name.is_misspelling?
       @correct_spelling = if @misspelling
-                            @name.correct_spelling.real_search_name
+                            @name.correct_spelling.user_real_search_name(@user)
                           else
                             ""
                           end
     end
-    @name_string = @name.real_text_name
+    @name_string = @name.user_real_text_name(@user)
```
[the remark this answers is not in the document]
mo-nathan: Yeah, I have a lot to say about the naming here, but again I tried to just stick to the purpose of this PR.  There's an good question about whether we really want to be passing the user down into these objects just so it can decide if the author should or shouldn't be there based on the current user's preference.  This is definitely an Adolf thing that I'm tempted to just rip out.  It would be way simpler if we just enforced a single format for all names.

app/controllers/names_controller.rb
mo-nathan: It's more about having a simple way to get to the end goal of getting rid of User.current.  See above comment about functions which slurp up remaining arguments.

app/models/name_sorter.rb
nimmolo: Mmm, but you could also say ripping out the method is outside this PR. Isn't this is how confusing dead code starts? 

If a PR is ripping out the only method that uses the code, to me there's no doubt cleanup from that belongs in the same PR. But maybe you reckon `@synonym_data` might be useful again at some point.

app/controllers/admin/emails/webmaster_questions_controller.rb
nimmolo: Based on the way it's established in `new` it seems `@email` is always going to be `@user.email`. A caller could send the a `user` param, but even then...

mo-nathan: Yeah, can't do it in the simplest possible way.  There are several cases where the user.email is not the sender_email.  For example folks who aren't logged in but want to send a comment.  I made the `sender_email` argument default toe `user.email` when `user` is non-nil.  It cleaned up a few cases, but doesn't appear to be a huge win.

app/controllers/herbaria/curator_requests_controller.rb
```diff
@@ -30,6 +30,7 @@ def create
       return unless @herbarium
 
       QueuedEmail::Webmaster.create_email(
+        @user,
         sender_email: @user.email,
         subject: "Herbarium Curator Request",
         content: "User: ##{@user.id}, #{@user.login}, #{@user.show_url}\n" \
```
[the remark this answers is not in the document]
mo-nathan: Fixed.

app/models/name_sorter.rb
mo-nathan: Turned out to be easy enough to get rid of it.

mo-nathan:
Thanks for the reviews!  I think I've responded to all the comments.  I'll plan to merge it tomorrow unless anyone has more concerns.

nimmolo (approved):

PR #2937, opened 2025-04-22
MyCoPortal report unedited

### LINKED ISSUE #2921: Mycoportal report
NEW PLAN: Create Mycoportal report at same level as the Symbiota report (without changing Symbiota).

### Old Tasks

  - [x] Duplicate Symbiota tests as MCP tests
  - [x] Duplicate pending Symbiota report as Mycoportal 
  - [x] Comment format_rows and expects with the label
  - [x] Email Scott with questions about fields
  - [x] UI
  - [x] extension should be `tsv`
  - [x] basic PR
  - [ ] #2942

### Old Tasks

- [x] Review report coverage
- [x] Add tests if that can be done reasonably
- [x] Add comments? E.g.. link to Symbiota documentation
- [x] Consider refactoring, e.g., an array with label & row format on same line.
- [x] RuboCop
- [x] Ready for review
- [x] Questions: 
  - [x] can I define `labels` and `format_row` in the same object?
  - [x] Superclass vs (more) MCP specific
  - [x] Change labels to documented Symbiota/DWC names:
    - [  ] `host` => `associated_taxa`
    - [ ] `fieldNotes` => `occurrenceRemarks`
  - [x] Add any useful missing [Symbiota Standard Fields](https://biokic.github.io/symbiota-docs/editor/edit/fields/#standard-fields)
     - [x] `fieldNumber` The number on a collection slip. Could be used to separate collector's number / collection number / FieldSlip from collection number. If we used `fieldNumber`, maybe `obs.id `would go into `recordNumber` and we could drop `mushroomObserverId`
     - [x] `identificationQualifier` cf., af. could `group` go here? https://dwc.tdwg.org/terms/#dwc:identificationQualifier`
     - [x] `identifiedBy` could get explicit identifier from notes or implicit from user who first suggested the id
     - [x] `identificationReferences` could get something from Namings.
     - [x] `identificationRemarks` could get something from Namings
     - [x] `municipality`
     - [x] `locationRemarks`

JoeCohen:
Creates a `Mycoportal` report and UI.

The report basically duplicates the existing Symbiota report.
It is only the first PR in reviving exports to MyCoPortal, and is not yet suitable for actual exports.
Follow-on PRs will make corrections and revisions based on communications with MCP. 
This PR incurs technical debt (duplication) due to time constrains in getting the exports working.

### DIFF
--- a/app/classes/report/mycoportal.rb
+++ b/app/classes/report/mycoportal.rb
@@ -1,0 +1,170 @@
+# frozen_string_literal: true
+
+module Report
+  # Report for exporting Observations to MyCoPortal
+  # https://mycoportal.org/
+  # https://www.mycoportal.org/portal/api/v2/documentation
+  # MyCoPortal is a Symbiota-based portal for Mycology
+  # https://symbiota.org/
+  # https://biokic.github.io/symbiota-docs/
+  # https://github.com/Symbiota/Symbiota
+  class Mycoportal < TSV
+    # These plus Column labels comprising
+    # Symbiota Standard Field names which are useful for our purposes
+    # https://biokic.github.io/symbiota-docs/editor/edit/fields/#standard-fields
+    # plus MO-specific fields that are useful for uploads to Symbiota portals
+    def labels
+      %w[
+        scientificName
+        scientificNameAuthorship
+        taxonRank
+        genus
+        specificEpithet
+        infraspecificEpithet
+        recordedBy
+        recordNumber
+        disposition
+        eventDate
+        year
+        month
+        day
+        country
+        stateProvince
+        county
+        locality
+        decimalLatitude
+        decimalLongitude
+        minimumElevationInMeters
+        maximumElevationInMeters
+        dateLastModified
+        substrate
+        host
+        fieldNotes
+        mushroomObserverId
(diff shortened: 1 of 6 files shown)


mo-nathan (approved):
Looks fine.  I don't fully understand the differences between mycoportal.rb and symbiota.org.  It does seem like there is a lot of duplication that could be DRYed.

JoeCohen:
MyCoPortal is a Symbiota portal. But it appears to me that MyCoPortal has several fields/values which are different from the Symbiota documentation.
I think Jason intended the Symbiota report to be used for any Symbiota portal, including MCP. (IDK if anyone actually does that.)
Jason suggested that I make `Mycoportal` a subclass of `Symbiota`.  That would avoid lots of the duplication.
But I don't want to do that now because:

- I told Scott I would have a draft Mycoportal report by the end of next week. I need to make numberous changes before I can provide that.
- want to do an actual upload to MCP before the end of May.
- I think there are some issues the Symbiota report, fixing them is a complicating delay.
- For me (probably not for others) subclassing is a complication that would delay things.
- I don't know how changes in the Mycoportal report would interact with a Symbiota superclass. Another complicating delay.

PR #2958, opened 2025-05-02
Enable deep links for observations

mo-nathan:
As part of the ongoing battle against the Spider People, this PR re-enabled direct links to observations for users that are not logged in.  The observation#show page has been greatly slimmed down when the user is not logged in including minimal links to other pages for spiders to go crazy about.  The primary use case for these links are regular users want to share an observation with folks who don't have MO accounts.  I would expect links like this to show up in external articles and such from time to time as well.  So we have to assume that spiders will see these links from time to time so we want something that gives them as little to go on as possible.

The general concept is to switch the observation#show to `/obs/<id>` and to allow those links to show a minimal version of the page to non-logged in users. `/<id>` and `/observations/<id>` links will still work logged in users, but spiders with the cached will continue to pound sand.

To test:
Visit an observation page as a logged in user.
Copy the link.
Logout of your local server.
Visit the copied link.
See that the page doesn't link to any other pages that have complex content.


JoeCohen:
Very nice! Thanks for doing this.
Question: Is it possible to get rid of the links to the User in the Comments?
Screenshot 2025-05-02 at 2 13 09 PM
Comments also sometimes contain other internal links, but it might not be worth the trouble to deal with them.

JoeCohen:
I suppose the Comment thing is really another PR.

JoeCohen (approved):

mo-nathan:
Funny you should ask that.  I already have it in the works.  It’s a bit
ugly so I didn’t get it done this afternoon.

PR #2962, opened 2025-05-03
Fix remaining `user`less calls to interactive_image

JoeCohen:
`interactive_image` now needs user as its first argument, else it throws a "wrong number of arguments" Error.
- Adds `user` argument to one call.
- Deletes unused method which had the other userless call.

### DIFF
--- a/app/helpers/images_helper.rb
+++ b/app/helpers/images_helper.rb
@@ -103,15 +103,6 @@ module ImagesHelper
        image.copyright_holder != object.user&.legal_name)
   end
 
-  def show_best_image(obs)
-    return unless obs&.thumb_image
-
-    interactive_image(obs.thumb_image,
-                      link: observation_path(id: obs.id),
-                      size: :small,
-                      votes: true) + image_copyright(obs.thumb_image, obs)
-  end
-
   # pass an image instance if possible, to ensure access to fallback image.url
   def original_image_link(image_or_image_id, classes)
     id = if image_or_image_id.is_a?(Image)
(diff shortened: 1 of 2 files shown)


app/helpers/images_helper.rb
```diff
@@ -103,15 +103,6 @@ def show_image_copyright?(image, object)
        image.copyright_holder != object.user&.legal_name)
   end
 
-  def show_best_image(obs)
```
JoeCohen, line 106: `show_best_image` looks like dead code.
Is there a reason to keep it (or am I wrong about it being dead.)

nimmolo: @JoeCohen I had asked Nathan to keep it with a comment that it's for getting the top image in a matrix box carousel, which i think we'd like to re-enable after Bootstrap update. 

I think he forgot to add the comment.

JoeCohen: Thanks so much! I'll restore it and add a comment.

nimmolo: @JoeCohen I could be wrong! see my revised comment above

nimmolo: @JoeCohen  Okay, i believe this is dead code after all. 

The way we find obs images involves batch-getting, then a re-sort that puts the thumb_image on top. This method is indeed obsolete then.

JoeCohen: @nimmolo. Thanks for the clarifications. I won't restore it.

PR #2965, opened 2025-05-05
Fix image uploads for users and projects

nimmolo:


app/controllers/projects_controller.rb
```diff
@@ -197,7 +197,7 @@ def set_ivars_for_show
 
   def upload_image_if_present
     # Check if we need to upload an image.
-    upload = params[:project][:upload_image]
+    upload = params[:upload][:image]
```
nimmolo, line 200: Maybe also `return unless params[:upload]` here, as you did above?

test/controllers/account/profile_controller_test.rb
```diff
@@ -54,7 +54,8 @@ def test_add_mugshot
         upload: {
           license_id: licenses(:ccnc25).id,
           copyright_holder: "Someone Else",
-          copyright_year: "2003"
+          copyright_year: "2003",
+          image: file
```
nimmolo, line 58: ah ok there it is

app/controllers/projects_controller.rb
nimmolo: Or use `upload = params.dig(:upload, :image)` which is effectively the same thing

mo-nathan: Switched to dig for both cases.

mo-nathan:
I think Coveralls is confused and I think this an important change to get out.

PR #2968, opened 2025-05-06
Order project observations by thumbnail quality

mo-nathan:
This is a better default for the NEMF project, but I'm not sure it's the right ordering for all projects.  The current default ordering is "Date".  I could also see an argument for "Activity".  In some ways it doesn't matter a lot since changing the order is just a click, but there may be folks who don't what this cheese moved.  It could be an option for a project, but that's a lot more work.  Would love to some thoughts from others.

I do think the `left_outer_joins` change to app/models/abstract_model/ordering_scopes.rb is worth doing regardless of whether we actually change the default ordering.  As the code currently stands the "Thumbnail Quality" sort order causes observations that have no images or no image votes to silently disappear.

### DIFF
--- a/app/controllers/observations_controller/index.rb
+++ b/app/controllers/observations_controller/index.rb
@@ -174,7 +174,8 @@ class ObservationsController
         project = find_or_goto_index(Project, params[:project].to_s)
       )
 
-      query = create_query(:Observation, projects: project)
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):

JoeCohen (commented):
Works fine manually. 
I don't understand the test failures.

PR #2971, opened 2025-05-08
Refactor `flash_notices_html` to use the `flash_notices.erb` partial

nimmolo:
The partial was previously unused. DRYing the code allows a turbo_stream update method to call the same helper without duplicating.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -57,25 +57,23 @@ module ApplicationHelper
   end
 
   # This can be called to display flash notices either in the page or a modal.
-  # Send it an html `id` to be able to find the notices via JS later.
   def flash_notices_html
     return "" unless flash_notices?
 
+    notices = flash_get_notices
     alert_class = case flash_notice_level
                   when 0 then "alert-success"
                   when 1 then "alert-warning"
                   when 2 then "alert-danger"
                   end
-
-    notices = flash_get_notices
     flash_clear
 
-    tag.div(notices, id: "flash_notices",
-                     class: class_names("alert mt-3", alert_class))
+    render(partial: "application/app/flash_notices",
+           locals: { notices:, alert_class: })
   end
 
   def render_turbo_stream_flash_messages
-    turbo_stream.replace("flash", partial: "application/app/flash_notices")
+    turbo_stream.update("page_flash") { flash_notices_html }
   end
 
   # Returns a string that indicates the current user/logged_in/admin status.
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -57,25 +57,25 @@ def container_class
   end
 
   # This can be called to display flash notices either in the page or a modal.
-  # Send it an html `id` to be able to find the notices via JS later.
+  # `flash_notices?` `flash_get_notices` `flash_notice_level` and `flash_clear`
+  # are defined in ApplicationController::FlashNotices.
```
nimmolo, line 61: @mo-nathan @JoeCohen 

Where should this method live? `flash_notices_html` is clearly a template helper method, but the methods it depends on are defined in `ApplicationController::FlashNotices`, which could be confusing. 

Although those methods are not template helpers, they are only used here (with one exception, which may be a mistake).

mo-nathan: Yeah, I agree that should get refactored.  To me helpers don't have to render HTML.  I can see the argument that mucking with session stuff is more of a controller thing than a view thing, but it's a fine line.  I also expect that doing that would require a lot more refactoring and I think we have more important fish to fry.

nimmolo: Yeah, I think the other methods should stay in ApplicationController. They're session getter/setter methods, and they don't feel like they belong in helpers — plus they're called in controllers, so that would break a lot of code.

mo-nathan: I don't think we're on the same page here, but I don't know that it matters much.  In my view, if all but one are only called from helpers, then they belong in with the helpers.  Refactoring `GlossaryTermsController#image_and_term_saves_smooth` to get rid of the `flash_clear` makes sense to me if that enables that change.  Personally I don't see a big issue with calling session getter/setter methods from helpers.  To me session-ness is independent of the Controller/View/Helper distinction especially with regard to flash message which by their nature are temporary and generally have no impact on the long term state of the application.  However, I feel like we're discussing stuff that doesn't matter much in the larger picture, so do whatever feels right to you with a bias towards not going down a rabbit hole.

nimmolo: OK. Easiest thing for this PR is just to merge as-is, no rabbit hole.

nimmolo: I'm noticing that Github does not broadcast comments, so i'm missing yours! 

That's what this PR is about really, getting an extraneous change out of the broadcast PR.

PR #2989, opened 2025-05-12
Move backend for action cable to the regular mo database

mo-nathan:
After checking this code out and running the migration, you'll need to edit your config/database.yml to remove the entry for `cable`.  You'll also eventually want to delete the `cable_development` database schema from your local MySQL instance, but there's no rush for that.

You'll want to test cable usage.  I had to have my Solid Queue process running to get it to work both before and after (`bundle exec rake solid_queue:start`). To test log in as two users in separate windows (use an incognito window or separate browser).  If you bring up the same observation in both windows and add comments, the new comment should appear in both windows.  After you do this, you should be able to see changes in the `solid_cable_messages` table.


JoeCohen (approved):
Ran it locally successfully.
Code makes sense to the extent I can understand it.

PR #2998, opened 2025-05-15
Remove User.current calls from API2ControllerTest and ObservationsControllerShowTest

mo-nathan:
This case was reasonably straight forward, but it suggests that there may be some hard issues ahead related to our logging.  In the case of object creation we can assume the user of the object is the creator.  However, in the case of updates and deletes that's going to be harder to determine.


app/controllers/api2_controller.rb
```diff
@@ -143,15 +143,20 @@ def upload_from_request_body
   end
 
   def render_api_results(args)
+    @user = user_from_key(args[:api_key])
     @api = API2.execute(args)
-    User.current = @user = @api.user
+    User.current = @api.user
```
nimmolo, line 148: I'm gathering that in the API, sounds like you still need to define `User.current` at this point.

mo-nathan: Updated and it looks good.

app/models/user.rb
```diff
@@ -354,6 +354,7 @@ def self.admin_id
   #   user = User.current
   #
   def self.current
+    # trace_tests
```
JoeCohen, line 357: Remove commented-out code?

JoeCohen, line 357: Remove commented-out code?

JoeCohen (approved):

PR #3004, opened 2025-05-20
Display remaining time

### LINKED ISSUE #3003: Remaining time
InatImportJobTracker should display remaining time.

### TODO
- [x] Fix test/models/inat_import_job_tracker_test.rb:15:in `test_timings'
- [x] Enhance test_timings with a test of elapsed.

JoeCohen:
Displays (estimated) time remaining for iNat import.
- Prepares for importing all a user's iNat obss.  
(IMO this feature (or similar) is a necessity for a good User Experience when importing large numbers of observations.)
- Resolves #3003 

### Manual Test
- Create several disposable iNat observations
- Import them, paying attention to the Remaining Time 
Expected result:
Remaining time should
  - Appear
  - Change as time elapses
  - Finish as "00:00:00" when the import is Done
- Delete the disposable iNat observations


app/jobs/inat_import_job.rb
```diff
@@ -231,19 +231,20 @@ def new_obs_params
   #   add an MO provisional name if none exists, and
   #   treat the provisional name as the MO consensus.
   def id_or_provisional_or_species_name
-    prov_name = @inat_obs.provisional_name
-    return @inat_obs.name_id if prov_name.blank?
+    return @inat_obs.name_id if @inat_obs.provisional_name.blank?
```
mo-nathan, line 234: Not a big deal, but why is this change in this PR?

mo-nathan:
The "remaining time" just went up and showed the same values as the elapsed time until it was done.  I just did two observations, but it seemed pretty surprising.

app/models/inat_import_job_tracker.rb
```diff
@@ -22,17 +22,25 @@ def status
...
+  end
+
+  def remaining_time
+    # Can't calculate remaining time until we've imported at least one obs
+    return "" unless imported_count&.positive?
+
+    remaining_importables = importables - imported_count
+    cumulative_avg_import_time = elapsed_seconds / imported_count
+    seconds =
+      (remaining_importables * cumulative_avg_import_time).to_i
```
mo-nathan, line 42: I don't see anything in this calculation that takes into account the time spent on the current import.  I would expect something like:
```
total_expected_time = total_count * avg_import_time
remaining_time = total_expected_time - elapsed_seconds
```
where `avg_import_time` only gets updated after each import completes.

mo-nathan (changes requested):
I saw surprising behavior when I imported just two observations where the remaining time tracked the total time until the process was done.  I think I see the issue and added a comment.

JoeCohen:
Thanks for the review and comments.

app/models/inat_import.rb
```diff
@@ -41,7 +54,52 @@ def add_response_error(error)
...
+
+  def reset_last_obs_start
+    update(last_obs_start: Time.now.utc)
+  end
+
+  def last_obs_elapsed_time
+    return 0 unless last_obs_start
+
+    Time.now.utc - last_obs_start
+  end
```
JoeCohen, line 84: @mo-nathan: Was thinking about handling this via a small `StopWatch` class with `elapsed_time` and `reset` methods.
Reactions?

JoeCohen:
I will later this evening try to add missing coverage to InatImportsController. I don't want to hold up the PR any more.

JoeCohen:
@mo-nathan, @nimmolo 
Any suggestions for covering `/app/controllers/inat_imports/job_trackers_controller.rb#show`, which is just a Turbo endpoint

JoeCohen:
> Update: `assert_turbo_stream` does not seem to work as i described.
> 
> It checks for a `<turbo-stream>` custom HTML element, and I don't think that shows up in the response unless explicitly rendered.

Thanks for checking on that.
I vaguely recall Nathan doing something like a render for Field Slips in order to test it. I'll look at that later.

mo-nathan:
Search for `format: :turbo_stream` in the tests for examples of what I did in order to test turbo stuff in controllers.

JoeCohen:
Thanks @nimmolo & @nathan. 
I complete covered the turbo endpoint via `get(:show, ... format: :turbo_stream)`, then checking `@response.body` for the expected displayed text.

app/models/inat_import.rb
mo-nathan: Sounds worth trying, but not urgent.

mo-nathan (approved):
Looks good!  Thanks for all the work on this.

PR #3006, opened 2025-05-21
Make the login page the root without a redirect for non-logged in user

mo-nathan:
I think this might fix the Let's Encrypt issue.  Checkout the login process process and the root behavior both as a logged in and logged out user.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -265,7 +265,9 @@ MushroomObserver::Application.routes.draw do
   #     resources :products
   #   end
 
-  # Default page "/" is /observations ordered order_by: :rss_log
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -266,7 +266,7 @@ def route_actions_hash
   #   end
 
   # Default page "/" is /observations ordered order_by: :rss_log
```
JoeCohen, line 268: Should this comment be changed?

JoeCohen (approved):
Behavior looks correct, code looks correct.

JoeCohen:
Thanks so much for jumping on this. It would have taken me a long time to do.

PR #3008, opened 2025-05-22
Jobs cleanup

### LINKED ISSUE #3010: Clean out old Jobs
Discard all Jobs which
- finished more than a set time (e.g. one week) ago, or
- failed more than a set time ago.

Documentation:
- https://github.com/rails/solid_queue
- https://github.com/rails/mission_control-jobs

JoeCohen:
Adds a recurring Job which discards stale finished and failed Jobs.

Cleaning up old Jobs (we currently have thousands of them) should make it easier to find and debug more current Job failures. Doing the cleanup on all Job classes obviates the need to have separate cleanup code per class.

Delivers #3010.


config/recurring.yml
```diff
@@ -15,3 +15,6 @@ production:
   repopulate_location_and_observation_center_lat_lng:
     class: UpdateBoxAreaAndCenterColumnsJob
     schedule: every day at midnight
+  discard_stale_jobs:
+    class: DiscardStaleJobsJob
+    schedule: every day at midnight
```
JoeCohen, line 20: Should this also have an `argument`?

JoeCohen:
Coverage decreases are hallucinations and/or have nothing to do with this PR.

config/recurring.yml
mo-nathan: What are you thinking?  This seems reasonable to me.

nimmolo: If discard_date is an arg you will want to pass, then yes, otherwise i guess not?

mo-nathan (approved):
This looks fine.  I'm not clear on how to test this locally, so I haven't looked at it in action, but it seems low risk.

JoeCohen:
@mo-nathan Thanks for the review.
I also couldn't figure out a reasonable local test.
I will merge and deploy it, then keep my eye on it.

PR #3030, opened 2025-06-08
Fix autocomplete

mo-nathan:
The current bad behavior is that if you try to autocomplete a name like "Phylloscypha phyllogena" the JS currently keeps re-asking for a huge list of all the names that start with "p".  The issue as far as I've been able to diagnose has at least three layers.

1. We have a *lot* of names that start with "phyl" and there is a limit of 1000 names.  I bumped this up to 5000 which is higher than the maximum number of names we currently have for each 2-letter combo.
2. The current code is using a parameter "whole" that it's not clear if it is intended to mean "whole words" or use the "whole string".  It appears in the JS to be intended to mean the first, but in Rails as the second.  I updated the Rails so it ignores this value for getting the search token.
3. The JS code is resetting a variable called `this.last_fetch_request` when responses come back based on the first result of the AJAX response.  This part of the response is always a single letter.  I don't see the reason for this logic since the value in `this.last_fetch_request` is already correct when the response comes in.  So I commented out this code.

Now things work seem to work correctly at least in the autocomplete for Field Slip names.  I am pretty uncomfortable with some of these changes given the number of things that rely on this autocomplete code and would like some through review and testing.

### DIFF
--- a/app/classes/autocomplete.rb
+++ b/app/classes/autocomplete.rb
@@ -32,9 +32,7 @@ class Autocomplete
 
   # returns an array of { name:, id: } objects
   def matching_records
-    # unless 'whole', use the first letter of the string to define the matches
-    token = string # whole ? string : string[0]
-    self.matches = rough_matches(token) || [] # defined in type-subclass
+    self.matches = rough_matches(string) || [] # defined in type-subclass
     clean_matches
 
     unless all
--- a/app/classes/autocomplete.rb
+++ b/app/classes/autocomplete.rb
@@ -15,7 +15,7 @@ class Autocomplete
   PUNCTUATION = '[ -\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]'
 
   def limit
-    1000
+    5000
   end
 
   def self.subclass(type)
@@ -33,7 +33,7 @@ class Autocomplete
   # returns an array of { name:, id: } objects
   def matching_records
     # unless 'whole', use the first letter of the string to define the matches
-    token = whole ? string : string[0]
+    token = string # whole ? string : string[0]
     self.matches = rough_matches(token) || [] # defined in type-subclass
     clean_matches
 
--- a/app/javascript/controllers/autocompleter_controller.js
+++ b/app/javascript/controllers/autocompleter_controller.js
@@ -1816,36 +1816,27 @@ export default class extends Controller {
   }
 
   // Process response from server:
-  // 1. first line is string actually used to match;
+  // 1. first line is first character of string actually used to match; [Unused]
   // 2. the last string is "..." if the set of results is incomplete;
   // 3. the rest are matching results.
   //
-  // `this.primer` is a huge array of records usually matching only the first
-  // letter typed, which is assumed not to change too often. `this.matches` is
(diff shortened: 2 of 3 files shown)


JoeCohen (approved):
I tried autocomplete for several objects in several contexts. It works fine.
The Rails code looks good once the questions are resolved.
I don't understand the JS.

app/javascript/controllers/autocompleter_controller.js
```diff
@@ -1832,8 +1832,10 @@ export default class extends Controller {
     this.fetch_request = null;
     // Record string actually used to do matching: might be less strict
     // than one sent in request.
-    if (new_primer.length > 0)
-      this.last_fetch_request = new_primer[0]['name'];
+
+    // Why the following code?  The last_fetch_request is already correct.
+    // if (new_primer.length > 0)
+    //   this.last_fetch_request = new_primer[0]['name'];
```
nimmolo, line 1838: The intention of the commented-out line seems to be handling the case when the user is deleting letters from their string. 
Like if they start typing "pohylo..." and then they delete back to "p" in order to retype "phylo" correctly.

Just need to test manually that this works without the line.

test/controllers/autocompleters_controller_test.rb
```diff
@@ -59,7 +59,7 @@ def test_autocomplete_location
     expect.unshift({ name: "M", id: 0 })
     expect.sort_by! { |loc| [loc[:name], -loc[:id]] }
     expect.uniq! { |loc| loc[:name] }
-    good_autocompleter_request(type: :location, string: "Modesto")
+    good_autocompleter_request(type: :location, string: "M")
```
nimmolo, line 62: By changing the test to the first letter, you may be missing something resulting from the change in `classes/autocomplete.rb` above, which is now taking the whole string rather than the first letter.

app/classes/autocomplete.rb
```diff
@@ -33,7 +33,7 @@ def initialize(params = {})
   # returns an array of { name:, id: } objects
   def matching_records
     # unless 'whole', use the first letter of the string to define the matches
-    token = whole ? string : string[0]
+    token = string # whole ? string : string[0]
```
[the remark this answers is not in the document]
mo-nathan: Yes, the old behavior is the problem.  I don't think we ever want to match against the just the first letter.  I don't see why this was ever a desirable thing.  In my view it is part of the lack of clarity about the meaning of "whole" as discussed in point 2 in the PR description.

app/javascript/controllers/autocompleter_controller.js
mo-nathan: Cleaned up comments.

app/classes/autocomplete.rb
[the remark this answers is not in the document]
mo-nathan: Cleaned up code since I'm not hearing any justification for keeping it.

app/javascript/controllers/autocompleter_controller.js
nimmolo: Manually seems to work great. Thanks for doing this.

PR #3031, opened 2025-06-10
Add search status light for project names

mo-nathan:
Got to the Summary tab of a Project (Projects -> Select a Project).  Note the search panel near the top of the screen.  Type in a name into the search box.  If the substring you've typed in matches a name in the project, then the status light should appear green.  If the string you've put in is not in the project then it should be red.


app/assets/stylesheets/mo/_status_light.scss
```diff
@@ -0,0 +1,85 @@
...
+  height: 20px;
+  border-radius: 50%;
+  border: 2px solid #ccc;
+  background: #f8f9fa;
+  transition: all 0.3s ease;
+  flex-shrink: 0;
+
+  &.red {
+    background: #dc3545;
+    border-color: #dc3545;
```
nimmolo, line 22: Instead of hardcoding all these colors, they should maybe be selected from the theme variables palette.

Especially the text colors like `status-text` should use a background and text color combination defined on the theme to be legible, or else this will probably look weird in other themes.

app/views/controllers/projects/_status_light.html.erb
```diff
@@ -0,0 +1,8 @@
+<!-- ERB Template -->
+<div class="status-light-container">
+  <div class="status-indicator" data-status-light-target="light"></div>
+  <span class="status-text"><%= :show_project_has_name.l %></span>
+  <div class="status" data-status-light-target="status" style="display: none;">
```
mo-nathan, line 5: Pretty sure this is Claude.ai Cruft from the first version that had buttons for turning the status light on/off.

app/javascript/controllers/project-search_controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+  connect() {
+    // console.log("Project search controller connected")
+    // console.log("Names data:", this.namesValue)
+    this.setupInputListener()
+  }
+
+  setupInputListener() {
+    // Find the autocomplete input using querySelector - no target needed
+    const input = this.element.querySelector('input[name="name"]')
```
nimmolo, line 15: Is this Claude's comment? 😣 In general, I would avoid using `querySelector` in Stimulus controllers, unless you _cannot_ know which element you want to watch. I don't think that's the case here.

Declaring targets (you do it at the top of the controller here and with a data-target attribute in the HTML) is the idiomatic Stimulus way: it alerts the dev working on the template that that element is watched by the (named) Stimulus controller, and is probably marginally more efficient. The clincher to me is making developer intent clear.

app/javascript/controllers/project-search_controller.js
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    this.setStatusLight(hasMatch ? 'green' : 'red')
+  }
+
+  setStatusLight(state) {
+    // console.log("Setting status light to:", state)
+
+    // Find status-light controller using Stimulus application
+    const statusLightElement = this.element.querySelector('[data-status-light-target="light"]')?.closest('[data-controller*="status-light"]')
```
nimmolo, line 46: Again, Claude: The light should definitely be a target. This is nonsense.

mo-nathan: In this case, the original proposed code for finding the status light element didn't work.  I believe that was due to some sort of race condition between the two stimulus controllers.  This was what I found that eventually worked.  Again if there's a better way, I'm very open to learning it, but will need your help to figure that out.

PR #3033, opened 2025-06-11
Combined search-status Stimulus controller

nimmolo:
@mo-nathan — things to note:

I've simplified this quite a bit. The key concept here ~is~ was "dispatching events" to communicate between controllers. Claude had "status-light" almost kinda set up for it, but "project-search" wasn't playing ball.

However, I think the "project-search" Stimulus controller could be generalized to the point that it can be combined with "status-light" into a "search-status" controller,  initialized with an array of strings, and able to show if any input matches anything in the array of strings sent to it. But you'll be the better judge of that. 

In that case, we wouldn't have to dispatch an event from here and catch it in the separate "status-light" Stimulus controller.

I assumed you want three translatable messages here for "off", "red" and "green", but you may not.


mo-nathan:
Looks very reasonable to me.  I don't know enough about Stimulus or really JavaScript in general to have much of an opinion.  If we can combine the controllers and still have the status light be easily reused, then that seems great!  Did you see my comment about adding status lights to the SpeciesListsController#show page rather than adding more to the ProjectsController#show page?

PR #3035, opened 2025-06-12
Bundle Update 20250612

JoeCohen:
- Bundle update (inspired by pending Dependabot PR).
- Fixes new RuboCop offenses. See comment below.


JoeCohen:
> CI has a bunch of Lint Rubocop offenses. E.g., https://github.com/MushroomObserver/mushroom-observer/actions/runs/15615003727/job/43985364325?pr=3035 The CI `Naming` offenses seem like RuboCop bugs. The others make sense. Will investigate.

Not bugs. Just a new cop which we offend a lot.

app/classes/api2/species_list_api.rb
```diff
@@ -108,12 +108,12 @@ def build_setter(params)
     private
 
     def validate_set_location!(params)
-      name = params[:place_name].to_s || return
+      name = params[:place_name].to_s
```
mo-nathan, line 111: This seems risky.

script/slowest_pages
```diff
@@ -20,7 +20,7 @@
 ################################################################################
 
 FILE = ARGV[0] || File.expand_path("../log/production.log", __dir__)
-NUM  = ARGV[1].to_i || 10
+NUM  = ARGV[1].to_i
```
mo-nathan, line 23: Why this change?

app/classes/api2/species_list_api.rb
JoeCohen: `to_s` should always be truthy, so `return` should never get evaluated.

script/slowest_pages
JoeCohen: Similar to above. `to_i` should always be truthy, so `10` never happens.

JoeCohen: PS both of the above were flagged by the [Lint/UselessOr cop](https://docs.rubocop.org/rubocop/cops_lint.html#lintuselessor).

app/classes/api2/species_list_api.rb
mo-nathan: Got it.

mo-nathan (approved):
Looks good.  Thanks!

PR #3036, opened 2025-06-14
Fix search wording

mo-nathan:
Updates the Project search wording to be more accurate ("has matching name" vs "has this name") and updates the logic so it lists one of the matching names with a "..." at the end if there is more than one match.

### DIFF
--- a/app/assets/stylesheets/mo/_status_light.scss
+++ b/app/assets/stylesheets/mo/_status_light.scss
@@ -5,16 +5,22 @@
   flex-direction: row;
   align-items: center;
   gap: 8px;
-  padding: 20px;
+  padding: 4px;
 }
 
 .status-indicator {
-  width: 20px;
-  height: 20px;
+  width: 10px;
+  height: 10px;
   border-radius: 50%;
-  border: 2px solid #ccc;
+  border: 2px solid $BODY_BG_COLOR;
   transition: all 0.3s ease;
   flex-shrink: 0;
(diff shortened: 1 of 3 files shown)


JoeCohen (commented):
Really cool.
Questions/comments:

1. Synonyms. Should there be some indication when the Project doesn't have the entered Name, but does have a synonym? When the entered Name isn't considered current?
2. Some issues with how it interacts with browser back & forward.
Example:
- http://localhost:3000/projects/380?q=25L29
- Enter `A`
Result:
Screenshot 2025-06-14 at 11 02 56 AM

- browser back
- browser forward
Result: Type-ahead results still there, but green light disappears
Screenshot 2025-06-14 at 11 06 01 AM

4. Maybe omit the `All Project Names` with a blank light.
Screenshot 2025-06-14 at 10 53 27 AM

I don't understand "All Project Names"
Looks kind of like a radio button.

mo-nathan:
Good catch.  Changed "All Project Names" to "Enter text to search project".  When I tried back / forward it clears the search box and resets to an empty circle.  Maybe it shouldn't show up at all when no text has been entered.  I'm also wondering if it should be on the same line as "Search:" and then the whole box could be tightened up.  @nimmolo what do you think?

nimmolo:
I agree about having it all one one line, although it will still stack on mobile - and also about having the whole thing hidden when there's no input.

PR #3041, opened 2025-06-18
Add from search status

mo-nathan:
Adds a widget for entering the field slip code and a button called "Add" to the search-status form.
The idea is that if there is a field slip code, then add should create a field slip with as much of the name as has been entered.
If there is no field slip code, then it goes directly to the create observation interface.


app/controllers/search_status_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Dispatches to either the ObservationsController#new
+# or FieldSlipsController#new depending on whether
+# field slip code is provided.
+class SearchStatusController < ApplicationController
```
nimmolo, line 6: Thinking about the naming of the controller. 

It doesn't govern _search status_, rather, as you said it governs which kind of resource is created by the link. I'd lean towards something like `NewResourceController` (which is disambiguated by its namespace under `FieldSlips`).

In my opinion, the action name `add` should stick to Rails default CRUD actions (i.e. `new`) for developer-legibility reasons, even though it isn't strictly speaking a CRUD action; it's a router to CRUD actions. But the name `new` renders the intent more legible and allows you to use default Rails route helpers. 

A lot of the CRUD refactor work was renaming actions that had names like this to the CRUD defaults and I do feel it's been beneficial for the overall legibility of the project.

The path could be called `field_slips_new_resource_path`

JoeCohen: Would it be better to namespace it under `Projects`?

JoeCohen (approved):

app/controllers/search_status_controller.rb
mo-nathan: I thought I had responded to this, but now I don't see it anywhere.  In any case the problem with putting it under a namespace is that this functionality pretty much equally applies to the ProjectsController and the SpeciesListsController in terms of where the "Add" button shows up and the FieldSlipsController and the ObservationsController in terms of where it lands.  In my view this makes it reasonable to not put it in a namespace.   I don't care for either `NewResourceController` or `ResourceController` since that seems too generic.  The reason I chose `SearchStatusController` is that it's at least related to other things called "search status".  I could see something like `AddDispatchController` since to me that does a better job of describing what it does.  Switching to `new` vs. `add` for the action is ok with me, but it seems a bit misleading since it's not really about creating anything "new".  It's worth noting that while it will usually end up dispatching to a `new` action, it will go to an `FieldSlipsController#edit` if the field slip already exists.

PR #3059, opened 2025-06-28
iNat Snapshot in notes

### LINKED ISSUE #3026: Prevent Imports from dominating Comment Index
Ideas:
- Somehow suppress adding the Comment to the Index
- Rather than creating a Comment, add all that text to Notes

JoeCohen:
Create the Snapshot of the imported iNat observation in the MO Observation Notes instead of as a separate Comment.

- Resolves #3026 
- Simplifies Snapshot, omitting list of things not imported
- Makes import faster by omitting step of creating a Comment

### Suggested Manual Test
- Import an iNat observation
Expected result:
  - The MO Observation should not include a "Snapshot" Comment
  - The MO Observation notes should include a Snapshot of the iNat Observation
- Clean up the iNat obs:
  - Delete the `Mushroom Observer URL` Observation Field
  - Edit the obs to remove `Imported by Mushroom Observer ...`

### Tasks
- [x] Add Snapshot to the Notes
- [x] Simplify Snapshot, removing the items not imported
- [x] Remove code which adds Snapshot to the Comments
- [x] Manual Test


mo-nathan (changes requested):
I don't believe this is working as expected.  While the notes show up on the observation in a notes field called: "Snapshot_of_Imported_iNat_Data", if I try to edit the notes, those notes are not accessible and if I change or save them, the original notes vanish.

I don't love the name "Snapshot_of_Imported_iNat_Data" for this data and it seems weird to me that it creates a secondary set of keys/values that aren't really keys and values in the notes sense.  In my view, it would make the most sense to have the import create multiple fiels in the notes for each of the iNat values.  You could then get rid of "Snapshot_of_Imported_iNat_Data" and perhaps add "iNat " as a prefix on each of the new fields.  Thus rather than:
```
Notes:
_Collector_: nathan
_Snapshot of Imported iNat Data_: User: nathan
  Observed: 2025-06-20
  Public lat/lng: 41.6467166667,-70.6229027778 +/-6 m
  Place: North Falmouth, Falmouth, MA, USA
  ID: Xanthoconium affine
  Quality Grade: Needs ID
  Suggested IDs:
    [Xanthoconium affine](http://localhost:3000/lookups/lookup_name/Xanthoconium+affine) by nathan 2025-06-28
  Observation Fields: none
```
It would look like:
```
Notes:
_Collector_: nathan
_iNat User_: nathan
_iNat Observed_: 2025-06-20
_iNat Public lat/lng_: 41.6467166667,-70.6229027778 +/-6 m
_iNat Place_: North Falmouth, Falmouth, MA, USA
_iNat ID_: Xanthoconium affine
_iNat Quality Grade_: Needs ID
_iNat Suggested IDs_:
  [Xanthoconium affine](http://localhost:3000/lookups/lookup_name/Xanthoconium+affine) by nathan 2025-06-28
_iNat Observation Fields_: none
```

JoeCohen:
@mo-nathan - Thanks for that review. 
You're right that it isn't working as I expected. I hadn't noticed the issue with invisible/uneditable notes.
I'd also prefer a shorter caption, but I want to keep the concept of a snapshot (else it looks like the Notes reflect the current state of the iNat obs).
IMO additional fields are unnecessary and the underscoring of those captions looks bad (and objectively causes a worse data-ink ratio).
I made some quick revisions, but am not perfectly happy with this result:

> Notes:
>> <ins>Collector</ins>: jdcohenesq
<ins>Snapshot</ins>: User jdcohenesq
Observed 2024-09-04
Public lat/lng 46.0118471242,-121.6004217451 +/-27063 m
Place Skamania County, US-WA, US
ID Caloscypha fulgens
Quality Grade Casual
Suggested IDs
  [Caloscypha fulgens](http://localhost:3000/lookups/lookup_name/Caloscypha+fulgens) by jdcohenesq 2025-05-19
Observation Fields none
>> <ins>Other</ins>: TEST for [MO iNat Import feature](https://mushroomobserver.org/articles/39)

Some options for tweaking this:
- Indent the lines under `Snapshot`
- Alternatively, make it all one line using semicolons instead of line breaks
- Insert a comma or hyphen after each subcaption. E.g., `Observed, 2024-09-04`

Thoughts.

app/classes/inat/obs.rb
```diff
@@ -219,15 +221,13 @@ def snapshop_raw_str
...
-        TAGS: :inat_not_imported.t
+        OBSERVATION_FIELDS: obs_fields(inat_obs_fields)
       }.each do |label, value|
-        result += "#{label.to_sym.t}: #{value}\n"
+        result += "#{label.to_sym.t} #{value}\n"
       end
-      result
+      result.chomp
     end
+    private :snapshop_raw_str
```
nimmolo, line 230: Is this a way to declare a single method private? Neat if so, i didn't know.

nimmolo (commented):
Code looks good, although I have not tested it. There are a number of programming tricks in your inat_obs object i wasn't aware of. Sharp!

app/classes/inat/obs.rb
JoeCohen: Thanks for the review. 
Yes, that's a way to declare a single method (or an array of methods) `private`.
https://docs.ruby-lang.org/en/3.3/Module.html#method-i-private

nimmolo: @JoeCohen 

- The meaning of "Snapshot" is not obvious to me. What if you call it something that makes it clear where it came from, like "iNaturalist imported data:"?
- I can see why the formatting of this has been difficult. I think i would prefer putting a newline at the beginning, so "User" starts on a new line, and using colons after every sub-field, but just not underlining the sub-fields. That way it's clear where each field is, but it doesn't interfere with scanning the major Obs notes fields. 

So like this
___
Notes:
  <ins>Collector</ins>: jdcohenesq
  <ins>iNaturalist imported data</ins>:
  User: jdcohenesq
  Observed: 2024-09-04
  Public lat/lng: 234.25325434235
  Place: Skamania County, US-WA, US
  ID: Mycena
  Quality Grade: Needs ID
  etc

PR #3063, opened 2025-07-01
Dependabot version control

### LINKED ISSUE #3062: Dependabot version updates
On a regular schedule, automatically create PRs which `bundle update`
 - [x] Review [Dependabot documentation](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide)

JoeCohen:
Use Dependabot to create regular PRs which `bundle update`

resolves #3062

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,4 @@ updates:
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"


JoeCohen:
Thanks for the review and comment.
I'll change it to weekly. (Side benefit: we'll learn quicker whether this is actually working.)
I agree that js stuff could be useful. But I don't know enough to implement this, and would like @nimmolo's input. We can always add that later. 
I'll merge it after changing it to weekly.

PR #3065, opened 2025-07-03
bundle update 20250703

JoeCohen:
- `bundle update`
- Deals with "new" RuboCop offenses: new cops, redefined cops, offenses missed because of timing of merges.
Fixes the offense or disables the cop.

> [!NOTE]
> [Style/CollectionQuerying cop added in RuboCop 1.77](https://docs.rubocop.org/rubocop/cops_style.html#stylecollectionquerying)


nimmolo:
Wow. Thanks for doing this. Feels like we're getting there.

JoeCohen:
Thanks for the feedback. Sometimes it feels like a lot running just to stay in place.

PR #3076, opened 2025-07-11
Fix a bunch of little vouchering related stuff

mo-nathan:
Items:
- [x] Allow "Road" in location names since that's what Google likes.
- [x] Remove "(required)" from place where it isn't true.
- [x] Add "Cancel" links to Create Observation, Create Observation List, and Create Project pages.
- [x] Correctly pass the place name from the field slip to the create observation form.
- [x] Fix user autocomplete on project member add page
- [x] Flip default on user trust for open projects
- [x] When Create Project encounters error the page is losing the form data

### DIFF
--- a/config/locales/en.txt
+++ b/config/locales/en.txt
@@ -858,6 +858,7 @@
   add_objects: Add [TYPES]
   all_objects: All [TYPES]
   cancel_and_show: Cancel (Show [TYPE])
+  cancel_to_index: Cancel ([TYPE] Index)
   create_object: Create [TYPE]
   destroy_object: Destroy [TYPE]
   edit_object: Edit [TYPE]
--- a/config/locales/en.txt
+++ b/config/locales/en.txt
@@ -3331,7 +3331,7 @@
   # ADMIN-TYPE PAGES
 
   # project/<id>/members/new
-  add_members_with_gps_trust: Hidden GPS data shared with project admins.
+  add_members_with_editing: Project admins can edit your associated observations.
   add_members_change_status: Change Status
   add_members_denied: "Permission denied: only admins for a project can add new members."
   add_members_title: Add [:users] to [title]
--- a/config/locales/en.txt
+++ b/config/locales/en.txt
@@ -858,7 +858,7 @@
   add_objects: Add [TYPES]
   all_objects: All [TYPES]
   cancel_and_show: Cancel (Show [TYPE])
-  cancel_to_index: Cancel ([TYPE] Index)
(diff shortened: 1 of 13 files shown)


config/locales/en.txt
```diff
@@ -858,6 +858,7 @@
   add_objects: Add [TYPES]
   all_objects: All [TYPES]
   cancel_and_show: Cancel (Show [TYPE])
+  cancel_to_index: Cancel ([TYPE] Index)
```
JoeCohen, line 861:
```suggestion
  cancel_to_index: "[:CANCEL] ([TYPE] [:INDEX])"
```
Works well to automatically translate for many languages, while still allowing manual translation.

JoeCohen:
Old thing: Is it a problem? Quick Create requires an existing Location.
If yes, maybe fix in another PR.

JoeCohen (approved):
Looks great.  I had a few comments, all your call.
Thanks for the cleanup. The little things often get ignored.

mo-nathan:
> Old thing: Is it a problem? Quick Create requires an existing Location. If yes, maybe fix in another PR.

Not clear to me what the behavior should be in this case, but it's certainly an older thing not related to this PR.  Feel free to create an issue around it if you think it's important.  Given that this is implicitly creating an observation with no images, I think it's reasonable to expect a written location.  Note that location aliases should work fine here.

JoeCohen:
> > Old thing: Is it a problem? Quick Create requires an existing Location. If yes, maybe fix in another PR.
> 
> Not clear to me what the behavior should be in this case, but it's certainly an older thing not related to this PR. Feel free to create an issue around it if you think it's important. Given that this is implicitly creating an observation with no images, I think it's reasonable to expect a written location. Note that location aliases should work fine here.

@mo-nathan - To close the loop and make sure we're talking about the same thing, the scenario I had in mind is:
The paper field slip has a written location, but it's not an MO or alias. Example: "Pocasset Town Forest".
If I enter that in the Field Slip form, then Quick Create I get:
Screenshot 2025-07-13 at 10 21 38 AM
IMO that's reasonable behavior, so we need to make sure that people recording field slips know how to:

- create a Location and use it in the Field Slip;
- use a reasonable existing Location instead of the what was written on the field slip.

An alternative (harder to code): send them to Location#create; when that's done return to the Field Slip form with the new Location filled in.

mo-nathan:
Ok, that's an edge case I'm not going to worry about too much.  If you want
to put in an issue about it, that's fine with me.  A simple solution might
be to just update the error message to be a bit more helpful about how to
create a location if they want.

PR #3081, opened 2025-07-11
Cancel tab for InatImport Form

### LINKED ISSUE #3079: `Cancel` link for InatImport
The InatImport form should have a Button (or link) which returns user to Create Observation.

JoeCohen:
Add a `Cancel` tab to the new InatImport view.

- Analog to the `Cancel` tabs added in #3076. See the [Slack conversation](https://mushroomobserver.slack.com/archives/C040TH9FV/p1752280354266689):

> Yeah, cancel buttons/links are unnecessary, but I expect many users don’t think that way.
Just added cancel wording/links to several create pages.  It’s a simple thing to do and if it reduces users friction seems worth it.

- Resolves #3079 

### Manual Test

- http://localhost:3000/inat_imports/new
Expected result: There should be a Cancel tab
- Click the tab
Expected result: Create Observation

### DIFF
--- a/app/controllers/inat_imports_controller.rb
+++ b/app/controllers/inat_imports_controller.rb
@@ -40,7 +40,7 @@
 # 5. The InatImportJob:
 #      Uses the `code` to obtain an oauth access_token
 #      Trades the oauth token for a JWT api_token
-#      Checks if the MO user is trying to import someone else's obss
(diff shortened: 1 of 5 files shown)


mo-nathan (approved):
Looks good and works well.

Issue #3082, opened 2025-07-12
Fix fungarium duplicate code issue


mo-nathan:
Currently it is possible to create two fungaria with the same official herbarium code.  There is one existing example of this: FLAS (https://mushroomobserver.org/herbaria/270 and https://mushroomobserver.org/herbaria/271).

The plan is to merge these using these SQL commands:

```
update herbarium_records set herbarium_id = 271 where herbarium_id = 270;
set @tmp = (select mailing_address from herbaria where id = 270);
update herbaria set mailing_address = @tmp where id = 271;
delete from herbaria where id = 270;
```

and add both a database and rails constraint to ensure the code remains unique.  For the database constraint, Claude.ai provided this migration (which I haven't tested yet):

```
class AddUniqueIndexToHerbariaCode < ActiveRecord::Migration[7.2]
  def change
    add_index(:herbaria, :code, unique: true, where: "code IS NOT NULL")
  end
end
```

We should also update the mailing list of 271 to be:

 Dr. Matthew Smith
 U. of FL, Gainesville
 Department of Plant Pathology
 2527 Fifield Hall
 Gainesville FL 32611-0680

PR #3094, opened 2025-07-14
Validation on Herbarium object to ensure code field is unique

mo-nathan:
Related to #3082.  I attempted to create an appropriate index and kept running into issues.  Even tried switching from using '' as the defeault to nil which I got working from the MySQL side of things, but then it caused Query to fail due to an issue with CONCAT failing with `nil` values vs. blank strings.

This solution works, but is not ideal.  It is only doing the uniqueness constraint in Rails which can fail if multiple folks are trying to add an herbarium with the same code at the same time.  However, this seems extremely unlikely and in any case the consequences aren't that severe.

### DIFF
--- a/app/models/herbarium.rb
+++ b/app/models/herbarium.rb
@@ -63,11 +63,8 @@ class Herbarium < AbstractModel
   # personal_user_id is set to mark whose personal herbarium it is.
   belongs_to :personal_user, class_name: "User"
 
-  # Was unable to create an appropriate index that made Trilogy
-  # happy.
-  # rubocop:disable Rails/UniqueValidationWithoutIndex
-  validates :code, uniqueness: true, allow_blank: true
-  # rubocop:enable Rails/UniqueValidationWithoutIndex
+  # Was unable to create an appropriate index that made Trilogy happy.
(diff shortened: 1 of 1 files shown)


app/models/herbarium.rb
```diff
@@ -60,8 +63,11 @@ class Herbarium < AbstractModel
   # personal_user_id is set to mark whose personal herbarium it is.
   belongs_to :personal_user, class_name: "User"
 
-  # Used by create/edit form.
-  attr_accessor :place_name, :personal, :personal_user_name
+  # Was unable to create an appropriate index that made Trilogy
+  # happy.
+  # rubocop:disable Rails/UniqueValidationWithoutIndex
+  validates :code, uniqueness: true, allow_blank: true
+  # rubocop:enable Rails/UniqueValidationWithoutIndex
```
JoeCohen, line 70:
```suggestion
  validates :code, uniqueness: true, allow_blank: true # rubocop:disable Rails/UniqueValidationWithoutIndex
```

mo-nathan: That's more than 80 characters.

PR #3100, opened 2025-07-15
Add herbarium code index

mo-nathan:
Adds an SQL index to the herbaria table.  This now works correct since the pattern scope for herbaria now does a coalesce to address null values in the code field.


JoeCohen (approved):
Looks good.
though I am (naively) surprised this needs to be done via an index as  opposed to simply making the column unique.

PR #3107, opened 2025-07-15
Batch iNat imports and broadcast `InatImport` status via Turbo, removing `InatImportJobTracker`

### LINKED ISSUE #3099: iNat import status updates without polling
Currently we have a Stimulus controller polling a Turbo endpoint in `InatImports::JobTrackersController` every second to get updates on the status of the iNat import job. All of the info needed are properties of the `inat_import` db object.

The fact that the relevant changes all originate in an update of an AR model means that instead of polling, Turbo can more efficiently update the view over a websocket by `broadcasting` AR updates. The way that works, every time the ActiveRecord object is updated, Rails will send an update to the front-end containing the HTML partial built from the reloaded object, without the overhead of polling.

We need something like this:
```ruby
class InatImport < ApplicationRecord
  after_update_commit lambda { |inat_import|
    broadcast_replace_to(
      [inat_import.user, :inat_import], partial: "inat_imports/current"
    )
  }
```

But to do this, we need to get rid of the intermediary object `inat_import_job_tracker` and use the `inat_import` directly. The partial sent by the broadcast to be built directly off the `inat_import` object. This means that the few class methods in `inat_import_job_tracker` need to be moved to `inat_import` and this object needs to be deleted.

nimmolo:
Resolves #3099 and #4061

## Summary

`InatImportJob` previously ran as a single job that imported every observation in one execution. For large imports this would block a SolidQueue worker for tens of minutes, making the queue unresponsive to cancellations and unrelated jobs.

This PR rewires the job into a chain of short batches:

- `BATCH_SIZE = 10` observations per job (~1 min at ~7s/obs measured in production).
- Each job passes `per_page: BATCH_SIZE` to `Inat::PageParser` so iNat returns exactly one page of that size per API call.
- When a batch fills and `total_results` indicates more pages remain, the job enqueues a continuation (`InatImportJob.perform_later(inat_import, id_above: last_id, continuation: true)`) and exits without calling `done`. The continuation picks up the cursor exactly where the previous job stopped.
- Continuation jobs skip `authenticate` and `ensure_not_importing_others` — the first job already exchanged the OAuth code for a JWT and verified the iNat username; the JWT lives on the `inat_import` record for subsequent batches.

## UI progress tracking

`InatImport` already carries all the attributes the status UI needs — `importables`, `imported_count`, `state`, `started_at`, `ended_at`, `response_errors`. The `after_update_commit` callback on `InatImport` broadcasts a Turbo Stream replace to `[user, :inat_import]` on every DB update. Since `ObservationImporter` calls `inat_import.increment!(:imported_count)` after each observation and the job updates `importables` from `total_results` on the first page, the progress percentage shown in the `Status` component is always the overall fraction of the total import — not just the current batch. No new model columns or separate tracking table needed.

## Test plan

- `test_continuation_job_skips_authentication` — proves continuation jobs never hit OAuth/JWT/users-me endpoints (WebMock raises on unregistered requests).
- `test_continuation_job_resumes_from_id_above` — proves the id_above cursor is forwarded; a request with the wrong cursor would hit an unregistered WebMock stub.
- `test_batch_full_enqueues_continuation` — proves a filled batch enqueues a continuation and leaves state as `Importing` rather than `Done`.


nimmolo:
Merged but carefully reverted. This is not the PR, #4629 is.

PR #3111, opened 2025-07-16
Enable import all

JoeCohen:
Allow an MO user to import all their iNat observations by checking the Import All box.


JoeCohen:
Closed because replaced by #3136.

PR #3115, opened 2025-07-20
Fix issue with editing namings

mo-nathan:
The problem scenario is when a user other than the owner of the observation creates a naming and then goes to edit that naming.  To reproduce:

1) Find an observation owned by another user.
2) Add a new name and vote for it with a value.
3) Edit the naming.

Currently, the vote field will say "No Opinion" rather than the vote value you just cast.
On the branch, the vote field should correctly show you the vote value you cast.

The issue is that it was only looking for votes owned by the owner of the observation.  Since the owner has necessarily not voted for the naming you just added, it gets "No Opinion".

Added a test for this scenario which fails on current main and succeeds on the branch.  I played around a bit with other circumstances and couldn't find any other edge cases that now misbehave, but it's worth poking at it more.  This issue may have been introduced as part of the modal refactor, but it also might be ancient bug that no one ever noticed/reported.

This issue was reported via email to webmaster@ by Igor at Jul 17, 2025, 11:20 PM.

### DIFF
--- a/app/controllers/observations/namings_controller.rb
+++ b/app/controllers/observations/namings_controller.rb
@@ -75,7 +75,7 @@ module Observations
       return redirect_to_obs(@observation) unless check_permission!(@naming)
 
       @consensus = Observation::NamingConsensus.new(@observation)
-      @vote = @consensus.owners_vote(@naming)
+      @vote = @consensus.users_vote(@naming, @user)
 
       if can_update?
         need_new_naming? ? create_new_naming : change_naming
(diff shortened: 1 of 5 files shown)


mo-nathan:
Spotted another case that I think needs fixing.  `owners_vote` is also called in NamingsController#update.

mo-nathan:
Per the test, the other case is if you introduce a new name as part of NamingsController#edit workflow, but it's not yet an approved name.  In this case the vote again used to switch to the owners vote value rather than the owner of the naming.

nimmolo (approved):

JoeCohen (approved):

PR #3138, opened 2025-08-03
Add the Actions menu to Observation Lists when seen in a Project context

mo-nathan:
I think it now makes sense to make these actions available in the Project context.  I removed them originally due to all the clutter.

However, I've also realized that the Project context version of this page now needs the species list pager that I apparently neglected back when I added the Project context, `add_pager_for(@species_list)`.  Just adding that code in `app/views/controllers/species_lists/show.html.erb` doesn't seem to work.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -391,7 +391,8 @@ class ApplicationController < ActionController::Base
   # having the project param stored inside the query record, q.
   def set_project_ivar
     # NOTE: Query param projects is always an array of ids.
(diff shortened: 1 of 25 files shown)


nimmolo:
What i'm doing is 
- changing add_project_banner from assigning content_for(:title) to content_for(:banner_title)
- changing the project show page to also add_page_title for the html title element, but not print it on the page if there is a project banner. 
- moving the order of the header so that the #title (e.g. for a list) may print below the project banner and project tabs. 
- printing a title on show species_list for project and a few other alias pages where it was missing.

PR #3185, opened 2025-08-14
Separate Out Write-In UI from Observation List Create Workflow

mo-nathan:
To test, create a new observation list and enjoy the simpler creation workflow.  Should be more of a standard CRUD style process.
From the Observation List Show page, select the "Write In New Observations" action.  Add newly created observations to the list by entering names and other details to the form.
Checkout the minor updates to the "How to Use" page.


mo-nathan:
There's an issue with handling different locations in the write-in UI.

JoeCohen (commented):
Looks good except for one obscure, weird issue. Possibly an old issue, and IDK whether you care about it.  
Observations created via Write in New Observations are unexpectedly missing from :

- [Activity Logs](http://localhost:3000/activity_logs)
- [Observation Index](http://localhost:3000/observations), when sorted by Activity.

app/controllers/species_lists/shared_private_methods.rb
```diff
@@ -262,182 +72,14 @@ def clean_notes(notes_in)
...
-
     def init_name_vars_from_sorter(spl, sorter)
-      @checklist_names = params[:checklist_data] || {}
       @new_names = sorter.new_name_strs.uniq.sort
       @multiple_names = sorter.multiple_names.uniq.sort_by(&:search_name)
       @deprecated_names = sorter.deprecated_names.uniq.sort_by(&:search_name)
       @list_members = sorter.all_line_strs.join("\r\n")
-      @checklist = nil
       @place_name = spl.place_name
     end
```
nimmolo, line 82:
```ruby 
  def init_name_vars_from_sorter(spl, sorter)
    # ...
    @multiple_names = multiple_names_array
    # ...
  end
 
  # Preassemble a `@multiple_names` 2d array: eager load the "other authors" names 
  # for each multiple name, so the template doesn't have to do an n+1 db query
  def multiple_names_array
    names = sorter.multiple_names.uniq.sort_by(&:search_name)
    names.map do |name|
      [name, name.other_authors.includes([:observations])]
    end
  end
```

app/views/controllers/shared/_form_list_feedback.erb
```diff
@@ -48,7 +48,7 @@ if !@multiple_names.blank?
       @multiple_names.each do |name|
         concat(tag.div(name.display_name.t))
         concat(fields_for(:chosen_multiple_names) do |f_c|
-          name.other_authors.each do |other_name|
+          name.other_authors.includes([:observations]).each do |other_name|
```
nimmolo, line 51: This looks like a db query, which i think we should be aiming to not do at the view layer. 

Can we eager load this association in `SharedPrivateMethods` as suggested above, and do here:
```ruby
@multiple_names.each do |name, other_authors|
  concat(tag.div(name.display_name.t))
  concat(fields_for(:chosen_multiple_names) do |f_c|
    other_authors.each do |other_name|
    # ...
```

mo-nathan: Done.  Not clear to me that it's worth the extra code in the controller, but whatever.

mo-nathan:
> Looks good except for one obscure, weird issue. Possibly an old issue, and IDK whether you care about it. Observations created via Write in New Observations are unexpectedly missing from :
> 
> * [Activity Logs](http://localhost:3000/activity_logs)
> * [Observation Index](http://localhost:3000/observations), when sorted by Activity.

IIRC, that was intentional to avoid floods of observations with no images taking over those UIs.  We could debate whether that's the right thing, but I'm not going to touch it in this PR.

JoeCohen (approved):

nimmolo:
UX wise - This is an improvement. 

This is outside the scope of the PR, but it's never been obvious to me how to add observations to the list. My test users (my friends here) also are flummoxed by that. They make a list and then they say "Now what? I just named something and i don't know what it is or how to use it."

I'm wanting a big ol' button on the species list show page that says "Add or remove observations from this list". I think it should appear here:

SpeciesListAddObs

If you like i can make the button as a PR from this PR, or i can do it after.

PR #3186, opened 2025-08-14
Update .rubocop.yml to allow `included` and `class_methods` blocks

nimmolo:
The `Metrics/BlockLength` cop has been unhelpfully flagging the `included` and `class_methods` blocks of all our concerns. Concerns require these blocks, which must be long by definition: they are how concerns define instance and class methods, respectively, of the classes they're included in.

The cop is configurable to allow specific methods. The default is `refine`.


nimmolo:
PR also refactors a long method in `controllers/concerns/descriptions/permissions.rb`. @JoeCohen can you check my logic there?

JoeCohen (approved):
I also suggest regenerating the rubocop todo list
```ruby
rubocop --auto-gen-config --no-exclude-limit --auto-gen-only-exclude
```

PR #3201, opened 2025-08-18
Position map markers in the right place when there are multiple markers

### LINKED ISSUE #2918: Observation pop-ups on Maps point to wrong location
To replicate: 
- Go to http://localhost:3000/names/6464/map
- Click on several of the red pins marking different Observations
Result:
  - As expected, a pop-up with the Obs #, Location name, and lat/lng appears for each clicked pin.
  - BUT all the pop-ups point to the same Location.
  
Image

nimmolo:
Fixes #2918

Manual testing needed: both the workflow described in #2918, and the create obs workflow selecting a GPS point on the map, need to be tested. This code change may affect interactivity of any map _points_ — not boxes. (Hopefully it only affects what we want to affect!)

### DIFF
--- a/app/classes/mappable/box_methods.rb
+++ b/app/classes/mappable/box_methods.rb
@@ -130,7 +130,7 @@ module Mappable
     # Arbitrary test for whether a box covers too large an area to be useful on
     # a map with other boxes. Large boxes can obscure more precise locations.
     def vague?
-      calculate_area > 24_000 # kmˆ2   or use MO.obs_location_max_area
+      calculate_area > MO.obs_location_max_area
     end
 
     # NOTE: DELTA = 0.20 is way too strict a limit for remote locations.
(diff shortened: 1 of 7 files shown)


JoeCohen (commented):
- This hangs for me when I go to http://localhost:3000/names/6464/map
(Maybe due to js debugger line mentioned by CoPilot.)
- I'm confused because the only fie changed in the PR is a js controller, but [the comment in #2918](https://github.com/MushroomObserver/mushroom-observer/issues/2918#issuecomment-3198057297) says "The solution is to be found in Ruby in maps_helper, not the js."
-Happy to review it again when the PR is Ready for Review.

nimmolo:
@JoeCohen  This branch doesn't hang on http://localhost:3000/names/6464/map for me. 

Be sure to stop and restart your local server when switching branches - I feel like you may not be doing that habitually, you seem to get a lot of "false negatives" when trying out PR branches. 

I find that app constant and asset reloading is unreliable when switching branches.

JoeCohen (approved):
Exercised it and it works great.
Thanks for fixing it.

PR #3203, opened 2025-08-19
Compare users when updating a page section, before deleting modal

### LINKED ISSUE #3200: Simultaneous commenting zapped
Comment draft of user A is zapped when user B simultaneously creates a Comments on the same Observation.

To reproduce (I did this in Chrome; unsure if it's browser related)
- In one window, start to Comment on an Observation;
- In an Incognito Window, login as a different User, and create a Comment on the same Observation.
Result: 
The draft in the first window vanishes, and adding a Comment in that window gets a blank Comment form.

> from User #1539 I. G. Safonov(IGSafonov))
... I am reporting "abnormal computing behavior" regarding making comments in observations. Since this is the second time it happened to me in a couple of months, I believe it's a bug.
Earlier today, I was typing a comment in observation #582315, when suddenly the comment window disappeared and the observation automatically scrolled to the top.  This apparently happened exactly when another user posted a new comment (while I was typing mine) and the page updated itself (is this even possible?).  Hoewver, when I clicked on the "Add comment" icon, the comment window was empty -- my long comment draft was wiped out.
I know that comment drafts are usually saved when, for example, I close the comment window on my own to look something up in the observation, like for verification purposes.
I am using the Google Chrome browser in my chromebook (naturally).
> 
> Thanks for looking into this.
Igor

nimmolo:
Addresses #3200. I won't even describe it because CoPilot is writing better descriptions these days.

Requires the manual test described in #3200. For good measure, manual testers should also test comment updates against another browser window with a CollectionNumber, HerbariumRecord, Sequence, or ExternalLink form modal open. No comment update should close anyone else's modal at this point.

#### Durability against regression:
The PR changes the `modal_form` partial to require a `user`, using the newish Rails `locals` template variables, declared in a magic comment at the top of the template. All partials monitored by `data: { controller: "section-update" }` Stimulus (not only comments) are now also required to supply a `section_update_user_value` as a data attribute — in the case of comment updates, it's the `comment.user`.  If it's not supplied the Stimulus controller will send an alert to the page.

The modal Stimulus controller checks that the users match before closing any modal. In the event of a broadcast update from a comment or any other future broadcast,  the `section-update` Stimulus controller notices the updated div, emits an `updated` event, and sends the `userValue` as a data value in the JS `event.detail`. This is checked by the `modal` Stimulus controller against the modal's `userValue`.

### DIFF
--- a/app/views/controllers/comments/_comment.erb
+++ b/app/views/controllers/comments/_comment.erb
@@ -14,7 +14,7 @@ target_type = tag.span(target.class.name.to_sym.t, class: "small") \
 tag.div(
   class: "list-group-item comment",
   id: dom_id(comment),
-  data: { controller: "section-update" }
+  data: { controller: "section-update", section_update_user_value: user.id }
 ) do
   concat(tag.div(class: "row") do
     [
--- a/app/views/controllers/observations/show/_collection_numbers.erb
+++ b/app/views/controllers/observations/show/_collection_numbers.erb
@@ -1,12 +1,14 @@
+<%# locals: (obs: nil, user: nil) %>
+
 <%
 numbers  = obs.collection_numbers
-can_edit = in_admin_mode? || obs.can_edit?(@user)
+can_edit = in_admin_mode? || obs.can_edit?(user)
 %>
 
 <%=
   tag.div(
     id: "observation_collection_numbers", class: "obs-collection",
-    data: { controller: "section-update", section_update_user_value: @user.id }
+    data: { controller: "section-update", section_update_user_value: user.id }
   ) do
 
     if numbers.any? && can_edit
--- a/app/views/controllers/observations/show/_collection_numbers.erb
+++ b/app/views/controllers/observations/show/_collection_numbers.erb
@@ -6,7 +6,7 @@ can_edit = in_admin_mode? || obs.can_edit?(@user)
 <%=
   tag.div(
     id: "observation_collection_numbers", class: "obs-collection",
-    data: { controller: "section-update" }
+    data: { controller: "section-update", section_update_user_value: @user.id }
   ) do
 
     if numbers.any? && can_edit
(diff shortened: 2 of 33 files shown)


app/views/controllers/observations/show/_collection_numbers.erb
```diff
@@ -1,12 +1,14 @@
+<%# locals: (obs: nil, user: nil) %>
```
nimmolo, line 1: This is an example of Rails template local variable declaration with supplied defaults.

app/views/controllers/shared/_modal_form.erb
```diff
@@ -25,15 +30,11 @@ being updated. Note that all actions are called by data tags on html elements.
 note: MUST REMOVE **.html**.erb file extension from form partials!
 %>
 
-<%
-form_locals = local_assigns.except(:identifier, :title, :form)
-%>
-
```
nimmolo, line 31: Instead of inferring `form_locals` here from top-level locals, make this a declared local variable and require callers to send these under the `form_locals` hash.

app/views/controllers/comments/_comment.erb
```diff
@@ -14,7 +14,7 @@ target_type = tag.span(target.class.name.to_sym.t, class: "small") \
 tag.div(
   class: "list-group-item comment",
   id: dom_id(comment),
-  data: { controller: "section-update" }
+  data: { controller: "section-update", section_update_user_value: user.id }
```
nimmolo, line 17: This is how the `section-update` Stimulus controller knows the owner of the updated section, e.g. the `comment.user`.

JoeCohen (approved):
Works great. Exercised it extensively with simultaneous comments vs comments, fungarium records, sequences, external links.
I also appreciate the added documentation in the code.
Thanks for jumping on this. I couldn't have fixed it.

PR #3208, opened 2025-08-20
New `Observation.within_locations(locs)` scope using `in_box`

nimmolo:
Following discussion on Slack, this PR would add a new scope for Observation queries by location `Observation.within_locations(locs)`, so that a query of a named Location can produce all obs that are mappable within the box of that Location, even if they are not explicitly associated with that Location. 

The PR also:

- Changes results of Observation PatternSearch to use `within_locations`
- Changes results of Observation API to use `within_locations`
- Does not change the results of "Observations at this Location", on a location show page. This still gives you obs explicitly associated with that Location.

The new scope would become the default for searches and the API, while still leaving the `Observation.locations(locs)` scope as a strict association, returning what you would expect if you called `location.observations`. (The latter is AR shorthand, and does not actually produce a pageable query — it calls `Observation.where(location_id: location_id)`, which is the same as `Observation.locations(locs)` in our scopes, and `create_query(:Observation, locations: [locations])` as a reusable, refinable Query.)

For obs with no GPS, they will turn up in the query if they are associated with a location whose center falls within the box, and is smaller than `MO.obs_location_max_area`.

Like the `locations` scope, this new scope also works for multiple Locations, e.g. "Coahuila, Mexico", "Nuevo Leon, Mexico"

#### Notes
This change is useful for searches like `Observation.within_locations("Utah, USA")`, which on this branch turns up any obs found in Utah or the southwest corner of Wyoming. Previously, `Observation.within_locations("Utah, USA")` would barely turn up anything, because it would only match observations associated with the vague location “Utah, USA”.

For searches like `Observation.within_locations("Shenandoah National Park, Virginia, USA")`, the new scope should also be more useful, because `region` wouldn’t necessarily turn up named locations near or within the park that don't share the string "Shenandoah National Park" in the location name. 

However, the new `Observation.within_locations` scope will produce very different results than the `Observation.region` scope for the same strings, when they describe large regions that are not rectangular (e.g. "Coahuila, Mexico" rather than, say, "Wyoming, USA"). 

- `Observation.region("Coahuila, Mexico")` returns obs whose named locations match the substring `Coahuila, Mexico`, which should turn up anything found in Coahuila, and _only_ those found there.
- By contrast, this new version of `Observation.within_locations("Coahuila, Mexico")` (the scope now using `in_box`) will turn up a lot of Texas and Nuevo Leon observations, because they’re within the bounding box of Coahuila. In this case, it’s less accurate than `region`, although maybe useful. 
- A final consideration: This type of query has the potential to be equally or more accurate than `region`, if we move the `Location` model to use polygons.


JoeCohen (approved):
1. This behavior is a big improvement in usability and UX. 
2. Observation Search help needs revision to account for the new behavior.
I pushed a (draft) revision. Please revert or revise as you wish.
3. I thought of a use case for the old Location search. 
I have a bunch of Observations with the same relatively broad Location, e.g., "Oregon, USA" and now want to fine tune them. With this PR, I think there's no longer a UI to find just those Observations. A Pattern Search for my Observations location:"Oregon, USA" includes Observations with a narrower Location.
IMO:  It's worth making this change regardless. This and similar ones are low-probability use cases, and we can deal with complaints when and if they occur.
4. Regarding the API, one can spin this as a breaking change (see 3 above) or simply improved results. My opinion is the same as for 3 above.

@mo-nathan What do you think about 3 and 4 above.

JoeCohen (changes requested):
/app/models/abstract_model/scopes.rb has some coverage gaps, including a new one.
https://coveralls.io/builds/75217904/source?filename=app%2Fmodels%2Fabstract_model%2Fscopes.rb#L520
I think it's worth covering all of them.

JoeCohen:
> @JoeCohen For "Oregon, USA" you can just use the `region` keyword. (There's a full discussion about this on slack). This is a pure location name match, ignoring the bounding box. In the pattern form you can do `region:"Oregon\, USA"`, and you can still get those with this PR!

In the use case I described, all Observations in the region "Oregon, USA" is over-inclusive.
I think that after this PR I cannot get only Observations whose Observation.location.name is "Oregon, USA".

JoeCohen:
P.S.: IMO that's not a blocker for the reasons I gave. 
But I'd like to know what @mo-nathan thinks.

nimmolo:
@JoeCohen Ah, that is correct. Thanks for explaining.

As an admin, you can of course write an AR query for it. I can't picture too many use cases for regular users, though.

An alternative we could try if we end up needing it, is to make the `locations` scope take a hash of arguments, like `names`. That would give the scope (and the search form) some logical branches to work with, as we do for synonyms, subtaxa, etc.

PR #3209, opened 2025-08-21
Add test for bug reported by Igor regarding broken naming interface

mo-nathan:
This is just a quick test of loading the naming form via turbo.  It fails if you revert the change in #3207, but succeeds are current main.


nimmolo:
Incorporated into #3214

PR #3230, opened 2025-08-25
Remove all double-spaces from translation strings

nimmolo:
`Double spaces after a period.  (Like this.)  Or this.` despite being good practice in typing class, are apparently not rendered on the front-end by Textile, they seem to get "squeezed" to a single space.

I discovered this writing #3229, trying to test for the presence of `:edit_affects_multiple_observations.t(type: :herbarium_record)`.

The presence of double spaces in the original multi-sentence translation string makes testing for the presence of the string (by reference to its tag) impossible without string manipulation or regex, and this seems silly. The string should at least match itself as rendered. I found that my simple test for the translation string worked if I just changed the double spaces to single spaces within that string's original English version in `en.txt`.

This PR proposes to get rid of the issue at the source, so we can write simpler tests.

### DIFF
--- a/test/models/translation_string_test.rb
+++ b/test/models/translation_string_test.rb
@@ -37,6 +37,9 @@ class TranslationStringTest < UnitTestCase
     assert_equal("Stuff that we may want to know.", str.reload.text)
   end
 
+  # The reason for this test is that double spaces in translation strings are
+  # "squeezed" to single spaces by Textile, so tests that expect the original
+  # string will always fail against rendered results. Simpler not to allow them.
   def test_no_double_spaces_in_en_txt_original_strings
     substring_test("%.  %")
     substring_test("%!  %")


JoeCohen (commented):
1. I agree with your changes. 
I was in the same typing class. And a couple of years of proofreading paper galleys for law review made me OCD about always using 2 spaces.
But [the standard is now a single space](https://www.google.com/search?q=how+many+spaces+after+a+period&rlz=1C5CHFA_enUS958US958&oq=how+many+spaces+after+a+period&gs_lcrp=EgZjaHJvbWUyBggAEEUYOdIBCTIxMzU4ajBqN6gCALACAA&sourceid=chrome&ie=UTF-8).
We should do that everywhere (including comments, Articles, and in our personal objects).
2. Is it worth writing a test to prevent reversion, or someone unintentionally adding 2 spaces in a new translation string? E.g., search for double space in `en.txt`, if there are any then fail and print the line number or name of the string.

nimmolo:
Yes to a test, good idea. 
I believe the only thing we can test for is `.  ` and `)  `, though — spaces after periods and end-parens. There seem to be 10 legitimate multi-space substrings in there for alignment or whatever reason, although i will scrutinize again before giving them a pass.

Issue #3237, opened 2025-08-26
Need better workflow for creating new projects


mo-nathan:
- [x] Need to be able to add empty observation lists and have them show up in the Observation List tab

Note that you can add aliases by clicking the "Project Aliases" button.

PR #3239, opened 2025-08-27
Show Observation Lists project tab when there are lists but no observations

mo-nathan:


mo-nathan:
Addresses #3237

JoeCohen (commented):
I barely use Projects or Lists. So these points are probably old issues, and probably demonstrate my incomplete grasp of the relationships among Observations, Lists, Projects. So they are not blockers.
- I was surprised that you can have a Project with no Observations, even though the Project has a List which has Observations. 
- The `Add My Observations` button is dangerous. Should it be `Add ALL my Observations` or maybe clicking it asks for confirmation with a warning like "This will ALL your Observations to this Project. Are you sure?"

JoeCohen:
Also surprised that a Project can have Aliases for Locations which aren't listed as Locations for the Project. See screenshot below.
Is there a UI to add a Location to a Project? (I can't figure out how to do that.)
Screenshot 2025-08-27 at 7 06 41 AM

PR #3264, opened 2025-09-08
User rake tasks

mo-nathan:
Started as a request for a simple way to add and verify users.  Turned into a bit more of an exploration of rake driven command line interfaces.  Probably should be an admin page, but this still seems handy and useful as is.  To use:

`rails user:verify` - Asked for a user login or email.  Verifies them.  Deals with cases where there is more than one account with the same email.  Try `nathan@collectivesource.com` to see the more complex UI.
`rails user:add` - Goes through a little command line interface to create users.  Such users are always marked as verified.  Allows for an easy way for admins to create new users who can't be bothered to create their own accounts, but are needed for foray events.
`rails user:list` - Lists all the users.  Debating if this is really useful or if it should get changed around.  It's pretty much as suggested by Claude.ai.  I'm curious to get other folks opinions.

Another interesting question is whether we should have `services` directory parallel to our current `classes` directory.  This appears to be a bit more standard in Rails these days.  Not sure what we put in `services` vs. `classes` or if we'd just rename `classes` to services.

### DIFF
--- a/app/classes/multiple_user_handler.rb
+++ b/app/classes/multiple_user_handler.rb
@@ -1,0 +1,142 @@
+# frozen_string_literal: true
+
+# app/services/multiple_user_handler.rb
+
+class MultipleUserHandler
+  attr_reader :input_handler, :output_handler
+
+  def initialize(input_handler:, output_handler:)
+    @input_handler = input_handler
+    @output_handler = output_handler
+  end
+
+  def handle_users(users, email)
+    display_multiple_users_header(users, email)
+    display_multiple_users_list(users)
+    display_verify_all_option(users)
+
+    handle_user_selection(users)
+  end
+
+  private
+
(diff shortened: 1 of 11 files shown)


JoeCohen (commented):
Thanks for doing this. It's just what we need for NAMA.
If you feel like working on more, maybe add the following. None are blockers.

- `rails user:verify` If user can't be found, redisplay the questions rather than returning to the command line. This would save one step if you make a typo when entering a login or email.
- `rails user:add` Display the passwords. Helps to prevent typos. Or would that be a security risk?
- `rails user:list` I have no use for this.

mo-nathan:
Removed `user:list`.  Made `user:verify` loop until success or a blank value is entered (carriage return).

I'm not entirely comfortable exposing the passwords even in an admin tool like this and I don't see that it's a huge win.

PR #3270, opened 2025-09-11
Dynamically rendered search forms in the search bar

nimmolo:
Makes faceted search forms available from the search bar, via the "+" button. (This button currently links to "Advanced Search".) Clicking the "+" opens a faceted search form for whichever model if available, right in the search bar.

Screen Shot 2025-09-13 at 9 24 35 PM

I'm pretty sure the faceted forms can do everything "Advanced Search" can do, and more. However, this PR keeps "Advanced Search" available from the same URL, "/search/advanced". It just removes the link in the search bar. It wouldn't be hard to keep the link, but I imagine we want to steer people to the new forms.

Screen Shot 2025-09-13 at 9 25 49 PM

This PR could be a precursor to removing "Advanced Search", which I would be in favor of. But if people still want it, we can keep it up.


nimmolo:
Added Observation List search form and tests

app/views/controllers/shared/_search_form.erb
```diff
@@ -2,10 +2,19 @@
 <%
 form_args = { url: { action: :create }, model: search,
               id: "#{controller.search_type}_search_form",
-              class: "faceted-search-form" }
-# form_args[:model] = search if search&:id
+              class: "faceted-search-form pb-4" }
+toggle_classes = %w[btn btn-link navbar-link px-2]
```
JoeCohen, line 6: Should this be frozen or a constant? No answer required.

JoeCohen:
All of my following comments are about Advanced Search, unless otherwise indicated.

app/views/controllers/shared/_search_form.erb
nimmolo: Thanks - actually it doesn't need to be there at all. I moved the toggles to helpers.

mo-nathan:
First time I hit the "+" button after updating to the latest code and running the server:
Screenshot 2025-09-22 at 3 12 38 PM

mo-nathan:
I think what's going on is that the there are a set of widgets that showup when you do some typing and then there is another set of widgets that show up when you click "More V".  However, they seem to get a bit confused from time to time.  Not sure I love the idea in the first place.  I'd be more in favor for each panel either showing all the widgets or none of the widgets.  If there are ones that are relevant for any reason, then they get disabled.

It's also weird that if you just did a search, then the name string shows up, but none of the widgets do.  That's why I clicked on the "More V" link to hopefully see what I had seen last time and instead got a completely different set of widgets.  This seems way too confusing.

mo-nathan:
Rather than "blank", "yes", "no" where "blank" means one "no" (at least for Synonyms and Subtaxa), I think it should just have "no" selected.  In these cases, I don't believe "blank" has any different functional meaning than "no".

mo-nathan:
Searching for "Amanita" with subtaxa and a confidence level of "I'd Call It That" returns nothing which make no sense.  It should at least return anything that with a rating above "Promising" (since it must have at least one "I'd Call It That" vote).  We could debate the range of values for what these different values mean, but I think the simplest rule is everything above the next lower value to the selected value.

nimmolo:
@mo-nathan Thanks for all this feedback.

One reminder: never call `rake assets:precompile`. Once you have called it, and that `public/assets` folder is populated with compiled assets, you are stuck having to call it again manually every time assets are updated. Save yourself the trouble :)

The solution is to delete the entire folder with `rake assets:clobber` and try to remember never to call that, it's never necessary with current Rails.

JoeCohen:
Faceted Search is stuck on pattern if it's done after a Pattern Search.
Locations `More` toggle doesn't work.
Example:

- Observations Patttern Search for `Pseudoclitocybe`
Result (correct): Observations  of names: Pseudoclitocybe, synonyms, subtaxa

- Click Search icon
- Click + icon
- Click `Clear` button
Result (unexpected): 
  - `Names` field is not cleared; still `Pseudoclitocybe`
  - These fields are hidden, with no way to see their values: `Synonyms`, `Subtaxa`, `All proposed`, `Exclude consensus`
  - Clicking Location `More` does nothing.
  IMO all the Location details should be initially hidden. Clicking `More` should toggle the Location details display.
- Click `Search`
Result (unexpected): same as the Pattern Search -- Observations  of names: Pseudoclitocybe, synonyms, subtaxa

JoeCohen:
I don't see an option corresponding to `has_public_lat_lng`.
IDK if that's intentional. 
For me, it's not a big deal if we don't have that option; I've never used it.
I suppose a use case is someone wants to collect a species, so wants to know exactly where it was found.
If you add that option, maybe call it something like `public geolocation` and make it binary (yes, no) or trinary (yes, no, either)

JoeCohen:
`Notes has` and `Comments has` text boxes should be expandable

nimmolo:
> As a newbie, I would not understand this help text:
> 
> Exclude Observations whose consensus is a name listed in "name:value". (When using exclude_consensus you must also use "include_all_name_proposals:yes".)
> 
> What is "name:value"?

In this (and most other cases) I'm reusing the existing help text for each pattern search keyword.

You're right in this case it needs to read differently.

nimmolo:
@mo-nathan Thanks for checking it again. I think those are all good ideas, and (apart from the subtaxa default) I considered most of them. They are too laborious to think about for me right now, though.

I'd like to think instead about merging this as soon as it's _better than what we have now_, and improving it later.

- Having the type switcher available on the faceted form is a request Joe raised too, and I wrote him a detailed response i can't find right now. It would be better, but it's not so simple to implement, first of all because not all the types have faceted forms yet. I spent half a day trying to make it work, and it's a huge headache.
- Remembering which form you have open would be nice to have, and probably involve a cookie, but again more complexity and scope creep from an already complex PR. Already this form is fewer clicks away than advanced search. 
- Scrolled to something... also nice to have, but not MVP. 
- Subtaxa default - The form as is always inherits the last query, or query defaults. To make an exception for this one case seems useful, but unless you feel it's part of an MVP, let's do it in another PR.

Let's get this thing out there, and then i can work on one feature at a time in smaller PRs.

nimmolo:
@mo-nathan This sounds like a bug of the user autocompleter, but not specific to this PR. This PR doesn't alter anything about autocompleters. 

I think it's likely a bug of all autocompleters, but user is the trickiest because it can be hard to find the right Joe.

nimmolo:
This PR now incorporates changes from #3349, so multiple autocompleter and its green checkmark should behave better.

nimmolo:
@mo-nathan - I changed the form prefiller to call `user.unique_text_name` instead of `user.name` in the `created_by` autocompleter when the values are inherited from a previous query, per your catch.

mo-nathan:
Thanks for those fixes.  I'm still getting some weird behavior with the autocompleters.  I can appreciate the desire to get this PR merged and then fix other related issues.  However, I'm guess I'm not sure what the criteria is for something that matters for this PR.  In my view, I would rather take the time before releasing a big shiny new feature to iron things out that are directly related to it and which users will likely notice.  It seems to me that reasonable criteria for this PR include "replace advanced search" and "have the same power as pattern search".  If that means fixing the autocompleters, then we need to fix the autocompleters regardless of whether there are considered "working" in other places in the system.  In my view, these forms raise the bar on the autocompleters since I expect folks will use the faceted search forms to repeat and refine searches over time.

So an example of an issue I just saw is with searching for partial names.  For example, if I do a pattern search for `Amanita user:barky` and then look at the faceted search, it correctly fills it out with `Name: Amanita, Subtaxa: yes, Created by: Nimmo (barky)`.  All good so far.  However, if I switch it to search for a partial genus name, say `Lactari` (I often do hacks like this for more obscure genera since I may not remember the ending), the faceted search form returns nothing.  However, the pattern search `Lactari user:barky` returns 3 observations.  In this case, when I open the faceted search form, the Name field is mysteriously empty.  If I enter `Lactari` in the name and select subtaxa, I get nothing.  The reported search criteria are then `created by: Nimmo, names: , subtaxa`.

Regarding the user autocomplete stuff, the fix you made helps, but it's still very easy to break.  For example, going back to `Name: Amanita, Subtaxa: yes, Created by: Nimmo (barky)`, if I change the created by to `nathan` and click search, you get `created by: Nimmo, names: Amanita, subtaxa`.  It only appears to change if you select an item/get a green checkmark.

Interestingly if you clear the form and then enter `Name: Amanita, Subtaxa: yes, Created by: nathan` (no green checkmark for created by), it "works" but the search criteria shown are `created by: Nathan Wilson, names: Amanita, subtaxa`.  This implies to me that it is looking up my user using my User.login and listing my User.name in the criteria.  I think most users would expect it to match all users whose name or login contain the substring `nathan`.  Interestingly, the pattern search also reports `created by: Nathan Wilson, names: Amanita, subtaxa`.  Pattern searching for `Amanita user:natha` fails with the error: `Invalid or unrecognized value for :user, “natha”, expected user id, login or name; nothing matched.`.  The faceted search fails with the error: `No matching observations found.` with the criteria `created by: , names: Amanita, subtaxa`.  It's sort of the same behavior, but it feels kind of squirrely since it has lost the user input.

mo-nathan:
Looking at the Names faceted search, if I enter "Name: Amanita, Subtaxa: yes, Deprecated: yes", it works fine with the reported criteria as `names: Amanita, subtaxa, deprecated` and when I open the faceted search form again, it is expanded to show the Deprecated field.  However, if I then changed Deprecated to "no", the results seem correct, but the reported criteria are `names: Amanita, subtaxa` and when I go back into the faceted search form the Deprecated field is hidden.  If I open the form, it still has "no" selected.

mo-nathan:
Why does Name:Misspellings have blank, "yes", "no", and "both"?  The help text says: `"no" = ignore misspelled names (default), "yes" = only misspelled names, "either" = ignore whether names are misspelled or not.`, but there is no `either` option.

mo-nathan:
Would be good to be able to search names for "Has observations: no".  I assume this is similar to searching for observations that have no sequence.  I added #3352 since I expect push back on getting this included in this PR.

mo-nathan:
Issues with: `Names: Created`:  I first tried `Created: 2006, Updated: 2025` hoping to get all names that were created in the year 2006 and had been updated in 2025.  The first results is http://localhost:3000/names/24656.  However, the log says this was created in 2010.  So something weird is going on there.  It was updated in 2025.

I then tried entering `2006-2007` in `Names: Created` and got "NoMethodError in NamesController: undefined method `split' for nil" with this stack trace:
```
[app/models/abstract_model/scopes.rb:322:in `datetime_with_levels_of_precision'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/models/abstract_model/scopes.rb:100:in `block (2 levels) in <module:Scopes>'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/models/abstract_model/scopes.rb:81:in `block (2 levels) in <module:Scopes>'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/initialization.rb:107:in `block in initialize_parameter_set'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/initialization.rb:100:in `each'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/initialization.rb:100:in `initialize_parameter_set'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/initialization.rb:93:in `initialize_scopes'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/initialization.rb:75:in `initialize_query'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/classes/query/modules/results.rb:48:in `num_results'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:295:in `set_index_view_ivars'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:274:in `show_index_setup'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:258:in `show_index_of_objects'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:198:in `filtered_index'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:62:in `block in build_index_with_query'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:49:in `each'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller/indexes.rb:49:in `build_index_with_query'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/names_controller.rb:13:in `index'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
[app/controllers/application_controller.rb:148:in `catch_errors_and_log_request_stats'](http://localhost:3000/names?q%5Bcreated_at%5D%5B%5D=&q%5Bmodel%5D=Name&q%5Bupdated_at%5D%5B%5D=2025#)
```

mo-nathan:
Location:Created and Location:Modified seem to have the same issues as Name:Created and Name:Modified.

mo-nathan:
Locations:Edited by:nathan causes:
NoMethodError in LocationsController#index undefined method `id' for an instance of String
```
[app/models/abstract_model/scopes.rb:62:in `block (2 levels) in <module:Scopes>'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/initialization.rb:107:in `block in initialize_parameter_set'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/initialization.rb:100:in `each'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/initialization.rb:100:in `initialize_parameter_set'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/initialization.rb:93:in `initialize_scopes'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/initialization.rb:75:in `initialize_query'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/classes/query/modules/results.rb:48:in `num_results'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:295:in `set_index_view_ivars'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:274:in `show_index_setup'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:258:in `show_index_of_objects'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:198:in `filtered_index'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:62:in `block in build_index_with_query'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:49:in `each'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller/indexes.rb:49:in `build_index_with_query'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/locations_controller.rb:28:in `index'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
[app/controllers/application_controller.rb:148:in `catch_errors_and_log_request_stats'](http://localhost:3000/locations?q%5Bby_editor%5D=nathan&q%5Bmodel%5D=Location#)
```

mo-nathan:
Locations:Created has the same type of autocompleter issues if you try to clear or edit the field.

nimmolo:
> Not surprisingly, Project is similarly weirdly sticky due to the autocompleter. Specifically, if I select a project (green checkmark) and then try to clear it out the green checkmark disappears, but the search still uses the selected project. You can workaround it by clicking the "Clear" button, but I'm sure users will interpret that as a bug.

I'm not getting this. If I 
- search by user "Nathan Wilson (nathan)" and also...
- match a project, get a green checkbox, and then clear the project, the project field is cleared, the project hidden ID field is cleared, and the parameter is not reflected in search results.

Issue #3288, opened 2025-09-14
Edit Field Slip which lacks an Observation


JoeCohen:
To replicate:
- http://localhost:3000/field_slips/qr_reader/new
- enter a code associated with a project, e.g., 2025-NAMA-nnnnnn 
- Click `Add Images or Other Data` (This creates a Field Slip with no associated Observation)
- Click the QR Code icon
- Enter the Code for the Field Slip you just created
Result: It shows that Field Slip
- Actions, Edit Field Slip 
Result: Throws an Error

> NoMethodError in FieldSlips#edit
Showing /Users/joe/mushroom-observer/app/views/controllers/field_slips/_form.html.erb where line #36 raised:
undefined method `unique_text_name' for nil
Extracted source (around line #120):       
       
```ruby
    return observation.collector if observation&.collector

    (user || @current_user).unique_text_name
  end

  def date
```
Trace of template inclusion: #<ActionView::Template app/views/controllers/field_slips/edit.html.erb locals=[]>

Rails.root: /Users/joe/mushroom-observer
```ruby
Application Trace | Framework Trace | Full Trace
app/models/field_slip.rb:120:in `collector'
app/helpers/forms_helper.rb:183:in `block in text_field_with_label'
app/helpers/forms_helper.rb:167:in `text_field_with_label'
app/helpers/autocompleter_helper.rb:35:in `block in autocompleter_field'
app/helpers/autocompleter_helper.rb:30:in `autocompleter_field'
app/views/controllers/field_slips/_form.html.erb:36
app/views/controllers/field_slips/_form.html.erb:20
app/views/controllers/field_slips/edit.html.erb:6
app/controllers/application_controller.rb:148:in `catch_errors_and_log_request_stats'
```

JoeCohen:
Resolved by #3293

Issue #3277, opened 2025-09-12
Field Slip without a project cannot be edited


JoeCohen:
To replicate:
1) Click on QR code icon
2) Enter a field slip code not associated with a project, say ABC-123
3) Click "Quick Create"
4) Go to the Field Slip show page for that Field Slip (may be easiest to go to the Field Slip Index page and find the field slip from there)
5) Click on "Actions"
On `main` you should not see a "Edit Field Slip" action.  On this branch you should.

PR #3293, opened 2025-09-16
Fix edge case bugs from NAMA

mo-nathan:
- #3277
- #3288 
See issues for steps to reproduce.


JoeCohen (approved):
Thank you for fixing both of these.

JoeCohen:
Resolves #3288

PR #3296, opened 2025-09-17
Remove unneeded Create icon from indexes

### LINKED ISSUE #3287: Error when trying to create a Fungarium Record from herbarium_records/nnnn
To replicate:
- Go to http://localhost:3000/herbarium_records/49651
- Click plus sign to create a new Fungarium Records
Result:
NoMethodError in HerbariumRecordsController#new
```ruby
undefined method `name' for nil
Extracted source (around line #153):              
    HerbariumRecord.new(
      herbarium_name: @user.preferred_herbarium_name,
      initial_det: @observation.name.text_name,
      accession_number: default_accession_number
    )
  end

app/controllers/herbarium_records_controller.rb:153:in `default_herbarium_record'
app/controllers/herbarium_records_controller.rb:71:in `new'
```

JoeCohen:
Remove New button (plus sign) from the herbarium_record index.
Records must be created from an Observation. Trying to create a record from the index throws an error.

- Resolves #3287

Similar for collection_numbers index, except displays flash error instead of throwing error.

### Manual Test

- Go to http://localhost:3000/herbarium_records/49651 (or any Fungarium Record)
Result: There should not be a New button (plus sign icon) next to Fungarium Records
- Go to http://localhost:3000/collection_numbers
Result: There should not be a New button (plus sign icon) next to Collection Numbers

### DIFF
--- a/app/helpers/header/rubric_helper.rb
+++ b/app/helpers/header/rubric_helper.rb
@@ -48,6 +48,7 @@ module Header
     ].freeze
 
     # Descriptions also don't get a create button
+    # Herbarium records are created via Observations, not from the index
     def nav_create(user, controller)
       unless user &&
              controller.methods.include?(:new) &&
@@ -66,7 +67,7 @@ module Header
 
     NAV_CREATABLES = %w[
       observations names species_lists projects locations images herbaria
-      herbarium_records collection_numbers glossary_terms field_slips
+      collection_numbers glossary_terms field_slips
       articles publications
     ].freeze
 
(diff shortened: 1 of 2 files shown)


nimmolo (approved):
Good catch. What about CollectionNumbers?

JoeCohen:
@nimmolo: Thanks for reviewing this, and better catch re Collection Numbers.
Add Collection Number should not be available from the index. 
I'll add that to the PR after coffee.

PR #3311, opened 2025-09-22
Consolidate and rename permission helpers

nimmolo:
We currently have overlapping helper methods for checking user permission to edit or delete an object.

- `check_permission(obj)` - the most commonly used. Returns boolean, does not flash if denied. Rubocop doesn't like this method name, because it returns boolean so it should be something like `permission?`. Checks:
  - `in_admin_mode?`
  - `(obj.is_a?(String) || obj.is_a?(Integer)) && obj.to_i == @user.id`
  - `obj.user_id == @user.id`
  - `obj.can_edit?(@user)`, configurable per model, can involve being an editor or member of a group
- `check_permission!(obj)` - also commonly used, calls `check_permission` but flashes "denied" message if denied

- `permission?(obj, message)` - Sort of a private method. Checks `in_admin_mode? || obj.can_edit?(@user)`. Near-duplicate of `check_permission!(obj)` but does fewer checks and requires an `error_message` arg. Returns boolean, flashes message if denied. Not currently called anywhere, except by the below methods:
- `can_edit?(obj)` -  Calls `permission?`. Used in controllers but not views. Flashes specific "cannot update" message if denied
- `can_delete?(obj)` - Calls `permission?`. Used in controllers but not views. Flashes specific "cannot delete" message if denied

This PR consolidates the first three methods around the Rubocop-approved names `permission?` and `permission!`. 

- Checks for all possible permissions in `permission?`, as in the old `check_permission`. Returns boolean.
- Gives `permission!` a default `error_message` arg, the "denied" message from `check_permission!`
- `can_edit?` and `can_delete?` now call `permission!` passing their own error messages

### DIFF
--- a/app/controllers/account/preferences_controller.rb
+++ b/app/controllers/account/preferences_controller.rb
@@ -169,7 +169,7 @@ module Account
     end
 
     def permitted_user_with_valid_email_type?(user)
-      user && check_permission!(user) && EMAIL_TYPES.include?(email_type)
+      user && permission!(user) && EMAIL_TYPES.include?(email_type)
     end
 
     def email_type_setter
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -331,20 +331,6 @@ class ApplicationController < ActionController::Base
     @project = Project.safe_find(project_id)
   end
 
-  def permission?(obj, error_message)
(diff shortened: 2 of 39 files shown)


JoeCohen (approved):
Thanks for cleaning this up and simplifying.
Looked quickly through the code. Didn't manually test, but that that seems unnecessary.

mo-nathan (approved):

PR #3316, opened 2025-09-25
Avoid providing default for Collector for new field flips

mo-nathan:
We found that the current user is usually the wrong thing during foray events with recorders.


JoeCohen (commented):
A win for project recorders -- the people who create the most FieldSlips.
Please merge if you don't like the complication I proposed.
It would be good to have this deployed soon so that we can dig into the backlog of NEMF Field Slips.

PR #3318, opened 2025-09-25
`DateRangeParser` to handle user input date ranges

nimmolo:
This PR moves the date range string parser from `PatternSearch` into a PORO, so it can be used to parse user-input date ranges in the new faceted forms.

Responds to an issue discovered by @mo-nathan testing the Observation search form. The `date` field should accept the same values as the `PatternSearch` keyword `date:val`, e.g. `"2008-2009"`, `"last_year"`, `"08-03"`, etc.


nimmolo:
- [ ] Test expectations have to be adjusted so that queries with a single date do not expect all results after that date.

Issue #3319, opened 2025-09-26
API Notes Hash Parameters


mo-nathan:
I tried setting a note field through the API and have not been able to do it.

`curl -X PATCH "http://localhost:3000/api2/observations?id=589510&api_key=###8&set_notes%5BCap%5D=red"`

seems to be the intent based on app/classes/api2/observation_api.rb around line 313 and various tests in test/classes/api2/observations_test.rb.  Note that %5B = [ and %5D = ].

However, rather than generating params that look like:
```
{:id=>"589510",
 :api_key=>"###",
 :"set_notes[Cap]"=>"red",
 :action=>:observation,
 :method=>"PATCH"}
```
as appears to be intended, the above URL generates params that look like:
```
{:id=>"589510",
 :api_key=>"###",
 :set_notes=>{"Cap"=>"red"},
 :action=>:observation,
 :method=>"PATCH"}
```
which the parser in ObservationAPI fails on resulting in the `Other` notes slot ended up with a value of:
`{"Cap"=>"red"}`.

I'm thinking of updating the parser in ObservationAPI to at least handle params like ` :set_notes=>{"Cap"=>"red"}`.  I'm debating if there is any reason to maintain support for params like `:set_notes[Cap]=>"red"` since I'm not convinced they can actually happen in the wild.

PR #3325, opened 2025-09-27
Update notes field API code to actually work with curl

mo-nathan:
Addresses: #3319 
To test put a valid API key in the following command.  It should update the "Cap" notes field to the value "red" in the given observation:

`curl -X PATCH "http://localhost:3000/api2/observations?id=589510&api_key=###&set_notes%5BCap%5D=red"`

On main this curl command put `{ "Cap" => "red" }` into the "Other" notes field.  On current main I could find no way to change actual notes fields using curl.


JoeCohen (approved):
1. Do you want to delete the commented out code in app/classes/api2/observation_api.rb?

PR #3327, opened 2025-09-28
Send letter param for pagination by letter with hidden field

### LINKED ISSUE #3326: Page nav `form` should sync letter nav input value if present, via Stimulus
From Ron Pastorino:
> For example, if I want to search my observations by name: If I put the name letter in, say "C" ,and then try to put any page number in other than "1", it reverts back to "A" names. I'm forced to go page by page until I get to the desired page.  Not terrible but more tedious than the previous system.

nimmolo:
This PR changes the pagination UI so that paginating within letter results, (e.g. "A" of the names index) maintains the letter context between pages. Currently on `main`, it does not.

It adds a new "letter" hidden field within the "page" UI, that sends the current letter being queried along with the number of the next page.

The hidden field is synced by Stimulus in case the user changes the letter.

Resolves #3326

### DIFF
--- a/app/helpers/header/index_pagination_helper.rb
+++ b/app/helpers/header/index_pagination_helper.rb
@@ -188,7 +188,8 @@ module Header
       ) do |f|
         [
           page_input_group_with_button(f, this_page, max_page),
-          q_param_to_hidden_fields(f) # (Just :q. :id not relevant on next page)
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):
Thanks for doing this, and for the fast turn-around.

PR #3356, opened 2025-10-07
Field slip user fix

mo-nathan:
@JoeCohen noticed that a bunch of FieldSlips have been created without a recorded user.  The issue was introduced as part of the `User.current` clean up.  This PR fixes the issue by adding a validation that requires FieldSlips to have a User and fixes the resulting failing tests.

There should also be a Rails runner script written to update the FieldSlips that currently have no recorded user with the owner of the observation.

Looking more deeply this is potentially an issue with the Name, Naming, and Observation models as well since they have also switched over to the `@current_user` pattern while getting rid of `User.current` and are potentially vulnerable to the same issue.  However, at the moment only Observations have cases where the User is NULL and those appear to be from before the `User.current` clean up started and may be intentional.  Taking a similar approach with this models seems reasonable.  I believe in the case of the Observations it might result in users in essence adopting Observations that don't have a User set.  It would be worth digging more into the Observations case.  I looked a bit at the most recent case, 569283.  It's not clear what happened here, but I do note that it was created using the MO API.  Interestingly the associated images do have an owner who I think would be reasonable to make the owner of the Observation.

### DIFF
--- a/app/classes/api2/observation_api.rb
+++ b/app/classes/api2/observation_api.rb
@@ -176,7 +176,7 @@ class API2
 
         field_slip.update!(observation:)
       else
-        field_slip = FieldSlip.create!(observation:, code: @code)
+        field_slip = FieldSlip.create!(observation:, code: @code, user: @user)
         field_slip.current_user = @user
         field_slip.update_project
         field_slip.save!
--- a/app/models/field_slip.rb
+++ b/app/models/field_slip.rb
@@ -4,12 +4,13 @@
 # code:    string, unique code for field slip, starts with project prefix
 
 class FieldSlip < AbstractModel
-  attr_accessor :current_user
+  attr_reader :current_user
 
   belongs_to :observation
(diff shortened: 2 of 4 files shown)


app/classes/api2/observation_api.rb
```diff
@@ -176,7 +176,7 @@ def add_field_slip_code(observation)
 
         field_slip.update!(observation:)
```
JoeCohen, line 177: Should it also check for `field_slip.user` and do repairs even when a FieldSlip already exists? (lines 175-177)
That would should be belt-and-suspenders after Rails Runner script is created and run.

JoeCohen (commented):
Should there be a test like `test_should_update_user_orphan` 
which asserts that `field_slips(:field_slip_user_orphan) has a user after it is updated to include an observation?

JoeCohen (approved):

PR #3367, opened 2025-10-13
Allow users to select if they want PDF or RTF labels

mo-nathan:
Add label format user preference (<username> -> Edit Preferences).
PDF labels have QR codes and no notes and fit on 3x5 cards
RTF labels have notes and no QR codes and can be arbitrarily long due to the notes.  However, the RTF format is much easier for users to edit.

Adding RTF labels back in is based on a request from Joanne Schwartz.


JoeCohen (changes requested):
It's currently concatenating the name parts in rtf mode. See comment on [app/classes/observation_labels/rtf_labels.rb‎](https://github.com/MushroomObserver/mushroom-observer/pull/3367/files#r2442446473)

Also consider giving the user the option to choose the format on the Download Observations page (in addition to setting default format).  (Not a blocker.)

JoeCohen (approved):
I think you should deploy it now as-is in order to keep Joanne happy and avoid spending more of your time on it right now.

Modifications -- if any -- can be done as future PRs.

Potential future modifications:

- I think Alan does something like this for his iNat observations, and that he has a script to print a label for MO observations. It might be worth asking him what his script does, then doing at least that for MO observations.
- If a user is going to the trouble of printing a lengthy RTF label, I'd think they'd want a QR code if one exists.
- Maybe ditto for sequences, projects, maybe observations lists. it would be interesting what Alan's labels do.
- Maybe change the Action in the Observations menu to read "Download Observations/Print Labels" instead of just "Download Observations".

PR #3376, opened 2025-10-19
Add new search types for fungaria and observation lists

nimmolo:
This just makes the existing search forms for these two models available in the search bar.

@JoeCohen is there a better place to define this array? It could be a constant.


JoeCohen (approved):
I can't think of a better place for the array.

PR #3393, opened 2025-10-21
Remove simplecov coverage for eval

### LINKED ISSUE #3388: Remove erb coverage from Simplecov
Coveralls gives false positives 100% of the time. This is almost certainly due to our asking Simplecov to cover erb files. It's better to have accurate Failure/Passing for code than to cover erb files.

JoeCohen:
See Pull Request Overview for technical description.
See #3388 for rationale.

Resolves #3388

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -28,7 +28,10 @@ else
   SimpleCov.formatter = SimpleCov::Formatter::HTMLFormatter
 end
 
-SimpleCov.start("rails")
+SimpleCov.start("rails") do
+  # An always empty file which is always reported as a coverage decrease
+  add_filter("/channels/application_cable/channel.rb")
+end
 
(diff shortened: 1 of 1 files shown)


JoeCohen:
1. Coveralls reports a relatively big coverage drop for this PR.
I hope that reported drop (a) relates to the fact that we're no longer covering views, which previously had pretty good coverage, and (b) won't be in PRs subsequent to the merger of this one.
2. I also changed the configuration to get rid of one annoyance: `app/channels/application_cable/channel.rb` is always empty, and coveralls always reports a coverage drop in that file. So I configured simplecov to ignore that file.

mo-nathan (approved):
Thanks!

JoeCohen:
Thank you. Now we need to see if it actually works.

PR #3398, opened 2025-10-24
Fix parsing string which ends with nbsp

### LINKED ISSUE #3396: iNat import text_name for nil
Image

JoeCohen:
Resolves #3396

### DIFF
--- a/app/models/name/parse.rb
+++ b/app/models/name/parse.rb
@@ -593,7 +593,7 @@ module Name::Parse
       strip_squeeze.
       # remove trailing spacing characters including non-breaking spaces
       sub(/[\s\u00A0\u202F]+\z/, "").
-      # and the same for lead
+      # and the same for leading
       sub(/\A[\s\u00A0\u202F]+/, "")
   end
 
--- a/app/models/name/parse.rb
+++ b/app/models/name/parse.rb
@@ -589,12 +589,9 @@ module Name::Parse
       gsub(/“|”/, '"'). # let RedCloth format quotes
       gsub(/‘|’/, "'").
       delete("\u2028"). # Unicode RLE that we see occasionally as line separator
-      gsub(/\s+/, " ").
-      strip_squeeze.
-      # remove trailing spacing characters including non-breaking spaces
-      sub(/[\s\u00A0\u202F]+\z/, "").
-      # and the same for lead
-      sub(/\A[\s\u00A0\u202F]+/, "")
+      # normalize whitespace characters including non-breaking spaces
+      gsub(/([\t\r\n\f\v\u00A0\u202F]+)/, " ").
+      strip_squeeze
   end
 
   # Adjust +search_name+ string to collate correctly. Pass in +search_name+.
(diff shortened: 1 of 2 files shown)


app/models/name/parse.rb
```diff
@@ -590,7 +590,11 @@ def clean_incoming_string(str)
       gsub(/‘|’/, "'").
       delete("\u2028"). # Unicode RLE that we see occasionally as line separator
       gsub(/\s+/, " ").
-      strip_squeeze
+      strip_squeeze.
+      # remove trailing spacing characters including non-breaking spaces
+      sub(/[\s\u00A0\u202F]+\z/, "").
```
JoeCohen: CoPilot suggested `u202F` the unicode narrow nonbreaking space.

app/models/name/parse.rb
```diff
@@ -590,7 +590,11 @@ def clean_incoming_string(str)
...
+      # and the same for leading
+      sub(/\A[\s\u00A0\u202F]+/, "")
```
JoeCohen: @nwilson-eol: I think we should do the same thing for the beginning of the string.
Does that makes sense?
If so, should I add a test for that?
Also, what about the middle of the incoming string. (I left that as is because I don't think a nonbreaking space in the middle will break anything.)

test/models/name_test.rb
```diff
@@ -1297,6 +1297,22 @@ def test_name_prov_name_with_spaces
     )
   end
 
+  def test_name_prov_name_with_trailing_nonbreaking_space
+    do_name_parse_test(
+      "Cuphophyllus \"pratensis-IN01\"\u00A0",
```
mo-nathan: I'd include a leading nonbreaking space as well and perhaps even one between the genus and the provisional species name if you take my above recommendation.

mo-nathan (changes requested):
The only thing I'd definitely like to see is for the test to also include a leading non-breaking space.  The idea of dealing with internal non-breaking spaces is not a blocker, but it seems worth attempting.

JoeCohen:
Thanks for the review and the suggestions. I'm in a meeting now, so will implement them after the meeting or tomorrow.

JoeCohen:
The internal one seems easy (at least at first glance).

app/models/name/parse.rb
```diff
@@ -589,7 +589,8 @@ def clean_incoming_string(str)
       gsub(/“|”/, '"'). # let RedCloth format quotes
       gsub(/‘|’/, "'").
       delete("\u2028"). # Unicode RLE that we see occasionally as line separator
-      gsub(/\s+/, " ").
+      # normalize whitespace characters including non-breaking spaces
+      gsub(/([\t\r\n\f\v\u00A0\u202F]+)/, " ").
```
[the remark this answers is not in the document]
JoeCohen: @mo-nathan: What do you think? I'm good either way.
I did it this way 

- to avoid substituting a single space for a single space (the most likely event); and
- `strip_squeeze` will take care of the squeezing.

mo-nathan (approved):
Looks great!  Thanks for the additional changes.

PR #3399, opened 2025-10-25
Make updated_at the Article index default sort

### LINKED ISSUE #3395: Article index default view should be Time Last Modified
This will alert users to the most recent changes.
I'd especially like [2025 MO Improvements](https://mushroomobserver.org/articles/44) to be high in the display by default.

JoeCohen:
Resolves #3395

### DIFF
--- a/app/classes/query/articles.rb
+++ b/app/classes/query/articles.rb
@@ -13,6 +13,6 @@ class Query::Articles < Query
   end
 
   def self.default_order
-    :created_at
+    :updated_at
   end
 end


mo-nathan (approved):
This is fine.  You might consider adding an explicit test, but might not be worth bothering with.

JoeCohen:
Thanks for the review @mo-nathan. I did consider that, but felt it wasn't worth it. 
Also there's no explicit test now for the default for any index.
Will deploy.

Issue #3394, opened 2025-10-21
API notes fields can cause 500s


mo-nathan:
To reproduce locally:
1) Find an API key (http://localhost:3000/account/api_keys) and an observation you own and assign these values to shell variables:
```
% export api_key=<api-key>
% export obs_id=<obs_id>
```
2) Run this curl command:
```
curl -X PATCH "http://localhost:3000/api2/observations?id=$obs_id&api_key=$api_key&set_notes%5BCap%20Color%5D=red"
```
3) Try to edit that observation (http://localhost:3000/observations/$obs_id/edit)

Boom!

The issue has to do with notes fields that contain spaces.  Normally in the web application spaces in notes keys get automatically converted to and from underscores.  However, the API skips that step and stores the value using a key that constains spaces.  This is interpreted by Rails as "frozen" (for reasons I don't fully understand) which throws an error when you try to edit the observation.

The question is whether to fix the API so it stores key values with underscores or if the system should tolerate/support keys with spaces.  A fix to get past the 500 is:
```
diff --git a/app/models/observation.rb b/app/models/observation.rb
index e08b13c44..1c0761bc5 100644
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -649,9 +649,7 @@ class Observation < AbstractModel # rubocop:disable Metrics/ClassLength
   def notes_orphaned_parts(user)
     return [] if notes.blank?
 
-    notes_keys = notes.keys.map(&:to_s).each do |key|
-      key.tr!("_", " ")
-    end
+    notes_keys = notes.keys.map { |key| key.to_s.tr("_", " ") }
     notes_keys - user.notes_template_parts - [other_notes_part]
   end
```

Despite the relatively simple fix described above, it has some issues.  For example, the new advanced search forms expect the key to have underscores and don't work with spaces regardless of how the data is represented in the database.  I also would not be surprised if there are other gotchas lurking down that path.  To really pursue that path we should probably get rid of the entire encoding/decoding with underscores, but I expect that will be a substantial amount of work.

I am leaning towards fixing the API so it translates the keys.  This seems like the fix that will the least harm to existing data and workflows.

JoeCohen:
Translation seems the way to go, at least for now.

PR #3412, opened 2025-10-28
Enforce rules for notes fields

mo-nathan:
The goal is to enforce the rules that notes fields 1) always replace spaces with underscores (a problem we've seen a bit in the wild) and 2) only allow one case pattern for notes fields.  For example, if an observation has a field called `iNat imported data` use that field if the user wants to set `Inat Imported Data`.  The first case pattern used wins.

The implementation is to essentially override the Hash that Observation.notes returns to a hash-like PORO called NormalizedHash that does the enforcement while still getting stored as a YAML serialized hash.


mo-nathan:
Remaining work:
- [ ] Verify that this fixes #3394 
- [ ] Cleanup the rendering of field names on the show observation page to show spaces rather than underscores
- [ ] Cleanup the rendering of field names on notes forms (Edit Observation and Edit Field Slip) to use the correct case pattern ("iNat imported data" rather than "Inat Imported Data")
- [ ] Ensure that the user's Observation Notes Template handles different case patterns correctly

JoeCohen (commented):
I have no comments on the app side of the code.
I wonder if there should be a test(s) focused on the issue(s) you are cleaning up, like this one:

> For example, if an observation has a field called iNat imported data use that field if the user wants to set Inat Imported Data. The first case pattern used wins.

If you don't think such a test is needed/desirable/worth the effort, then please merge.

PR #3415, opened 2025-10-31
Migrate `panel_helper` to Phlex components

nimmolo:
Summary

Replaces ERB helper methods for panel rendering with a standalone `Panel` Phlex component, hopefully improving maintainability and testability.

  Key Changes

  - Adds `phlex-slotable` gem for more configurable content within components
  - Creats `Components::Panel` class to replace `panel_helper.rb`
    - `Panel` has optional slots `heading`, `heading_links`, `thumbnail`, `body`, `footer`
  - Migrates all view files from `panel_block` helper to `Components::Panel.new`
  - Removes `panel_helper.rb` and 4 related helper methods from other files
  - Adds component tests that test the HTML nesting, not just the content
  - Adds some new component test helpers

  Migration Pattern

  Before:
```ruby
# definitions of `shown_html, collapsed_html`, and of other args, omitted

<%= panel_block(heading: :"search_term_group_#{heading}".l,
                collapse:, open:, collapse_message: :MORE.l,
                panel_bodies: [shown_html, collapsed_html]) %>
```
After (note `component.with_#{slot}` syntax):
```ruby
<%= render(Components::Panel.new(
             panel_id: "id",
             collapsible: true,
             collapse_id: "collapse_id",
             expanded: true
          )) do |panel| %>
  <%= panel.with_heading { "Title" } %>
  <%= panel.with_body { shown_html } %>
  <%= panel.with_body(collapse: true) { collapsed_html } %>
<% end %>
```

### DIFF
--- a/app/helpers/panel_helper.rb
+++ b/app/helpers/panel_helper.rb
@@ -2,25 +2,6 @@
 
 # helpers for bootstrap panels
 module PanelHelper
-  # Create a div for notes.
-  #
-  #   <%= notes_panel(html) %>
-  #
-  #   <% notes_panel() do %>
-  #     Render stuff in here.  Note lack of "=" in line above.
-  #   <% end %>
-  #
-  def notes_panel(msg = nil, &block)
-    msg = capture(&block) if block
-    result = tag.div(msg, class: "panel-body")
-    wrapper = tag.div(result, class: "panel panel-default dotted-border")
-    if block
-      concat(wrapper)
-    else
-      wrapper
-    end
-  end
-
   # Help tooltip, note, block.
   #
   # Help tooltip is a span with a title attribute. This has the effect of


app/components/panel.rb
```diff
@@ -0,0 +1,226 @@
...
+    chevron_down = link_icon(:chevron_down, title: :OPEN.l,
+                                            class: "active-icon")
+    chevron_up = link_icon(:chevron_up, title: :CLOSE.l)
+
+    # link_icon returns HTML strings that need to be rendered as HTML,
+    # not escaped as text. Nil check needed as link_icon may return nil
+    # in test environments where icon assets aren't available
+    # rubocop:disable Rails/OutputSafety
+    raw(chevron_down.html_safe) if chevron_down
+    raw(chevron_up.html_safe) if chevron_up
```
mo-nathan: Dug into these cases and apparently none of this is needed any more because of Phlex.  Replacing lines 143 to 153 with just:
```
    link_icon(:chevron_down, title: :OPEN.l,
              class: "active-icon")
    link_icon(:chevron_up, title: :CLOSE.l)
```
appears to do the job.

mo-nathan: This change is also on njw-phlex-discoveries

nimmolo:
@mo-nathan I added the `phlex-slotable` gem to handle passing HTML content to the panel header, thumbnail, body and footer without having to call `raw(content.html_safe)`.

I realize you're busy with other things but if you want to check out the syntax when you get a chance, it's pretty different now.

mo-nathan (approved):
I'm ok with this as far as I've dug into it and I don't want to hold it up.  However, my hope was that Phlex would reduce the code that is actually in ERB files and this PR does not achieve that hope as far as I can see.  If anything it seems to add code to the ERBs and make them more complex.  I recognize that this is an early step, but I just want to raise this concern now to hopefully focus future efforts on the goal of reducing the amount of ERB code in the system.  Perhaps it would help to look at it from the other direction to see what it would look like to implement a page that does not have any ERB code.

PR #3416, opened 2025-10-31
Introduce Phlex components, migrate `alert_block` helper

nimmolo:
This is intended to be a simple first PR introducing Phlex. 

It’s not going to be so clear from this PR what is preferable about Phlex. It’s actually a bit more verbose than ERB here, and `alert_block` just basically wraps a block of HTML anyway. The advantage of Phlex is in being testable and rubocoppable Ruby, and it’s really valuable in complex components where typed args get validated right when you instantiate the class (by rendering it). The Alert component however has no really crucial args that need to be validated. That will come in later components.

#### Summary 

Replaces `alert_block` ERB helper method for alert rendering with a standalone Phlex component. Introduces Phlex syntax.

  1. Adds `phlex-rails` gem, `literal` for typed object properties and `strict_ivars` to raise when an ivar is undefined
  2. Creates `Components::Alert` Phlex component - a Bootstrap alert component that replaces the helper
  3. Adds component tests - 11 tests covering various scenarios (message, levels, blocks, custom attributes)
  4. Migrates all 17 view files - converts all usages of `alert_block` to `render(Components::Alert.new(...))`
  5. Updates `flash_notice_html` and `form_***_feedback` partials to render Alert components - these had divs with class `alert` but didn't use the helper
  6. Removes the old helper - Deletes `alert_block` from `panel_helper.rb`


JoeCohen (commented):
I looked at the phlex documentation and this PR.
Sorry to say that it was in one ear and out the other -- it sounds neat, but I understood almost nothing.
I like automatic coverage and notification of decreases, so I don't like the issues created by strict_ivars in test mode.
PS: 12 test errors

JoeCohen:
Do you mean "There are cases I don't know how to provoke manually
[programatically]."?

PR #3436, opened 2025-11-07
Re-enable lookup with plus signs

### LINKED ISSUE #3435: Name lookup causes warning
To replicate
- Go to https://mushroomobserver.org/info/textile_sandbox
- Enter any good name surrounded by underscores In the `Enter Textile ...` text field, for instance `_Boletus edulis`
- Click the Test button
- Click the link in the `This is what it will look like:` field
Result: Flash warning like `No names match “Boletus+edulis”, maybe you meant one of these?`

Similar for multi-word glossary term, e.g.`glandular dots`, except if gives a Flash Error instead of a warning:
Can’t find glossary term matching ‘glandular+dots’.

mo-nathan:
Resolves #3435

### DIFF
--- a/app/controllers/lookups_controller.rb
+++ b/app/controllers/lookups_controller.rb
@@ -62,7 +62,7 @@ class LookupsController < ApplicationController
   # controller/action after looking up the object.
   # inputs: model Class, true/false
   def lookup_general(model, accepted: false)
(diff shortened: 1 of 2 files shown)


JoeCohen (approved):
Nice! Glad it was that simple.

PR #3437, opened 2025-11-07
Only cache MatrixBox if thumb_image is transferred

nimmolo:


JoeCohen (approved):
Looks OK to me. 
(But I don't understand the underlying problem. I trust you guys on that. (I'm so far behind that I feel like I can't take time to understand it.)

nimmolo:
@mo-nathan checked out the code change on Slack, so i'm going to merge, and hopefully we'll start to be able to tell if this is working.

PR #3453, opened 2025-11-11
Fix second+ image upload handling on CreateObs form

nimmolo:
This PR requires manual testing. To test, upload more than one image to a new observation.

  1. EXIF Extraction Fixes:
  - ✅ Server-side EXIF extraction for saved "good" images on edit forms
  - ✅ JavaScript optimization (skip client-side extraction for "good" images)
  - ✅ GPS visibility fix for owners viewing their own edit forms
  - ✅ Test fixture fallback for EXIF extraction in test environment
  - ✅ Type-safe GPS coordinates (Float instead of String)
  - ✅ Fallback to database values for images without GPS EXIF

  2. Test Improvements:
  - ✅ Combined two geotagged image tests into one to ensure predictable order
  - ✅ Added comprehensive EXIF extraction test for edit forms
  - ✅ Updated test helpers to handle hidden carousel indicators
  - ✅ Fixed flash message assertions (outdated expectations)
  - ✅ Updated Google Maps data constant (UNIVERSITY_PARK)

### DIFF
--- a/app/components/form_camera_info.rb
+++ b/app/components/form_camera_info.rb
@@ -100,11 +100,11 @@ class Components::FormCameraInfo < Components::Base
 
         # Wrap each field so we can hide it when empty
         wrapper_class = class_names("exif_#{field}_wrapper",
-                                     "d-none": !has_value)
+                                    "d-none": !has_value)
 
         span(class: wrapper_class) do
           # Add comma before non-first fields
-          plain(", ") if index > 0
+          plain(", ") if index.positive?
 
           render_gps_part(field, value || "")
         end
--- a/app/components/form_camera_info.rb
+++ b/app/components/form_camera_info.rb
@@ -26,15 +26,16 @@ class Components::FormCameraInfo < Components::Base
   # All props are nilable to handle cases where data may not be available
   prop :img_id, _Nilable(String), &:to_s
   # GPS coordinates can be passed as Float, Integer, or String
-  # Convert to string for display, handling blank values
-  prop :lat, _Nilable(String) do |v|
-    v.present? ? v.to_s : ""
+  # Coerce to Float for semantic correctness
+  # (Phlex will convert to string when rendering)
+  prop :lat, _Nilable(_Union(String, Integer, Float)) do |v|
+    v.present? ? v.to_f : nil
   end
-  prop :lng, _Nilable(String) do |v|
-    v.present? ? v.to_s : ""
+  prop :lng, _Nilable(_Union(String, Integer, Float)) do |v|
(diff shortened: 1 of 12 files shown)


JoeCohen (commented):
I manually tested it by creating an Obs with 2 images, then editing an Obs by adding 2 images. It worked fine.
I don't understand phlex yet, so I don't understand the code. Sorry. I need to catch up on that soon.
The tests are failing in CI.

app/components/form_camera_info.rb
```diff
@@ -23,13 +23,22 @@ class Components::FormCameraInfo < Components::Base
...
-  prop :lng, String, default: ""
-  prop :alt, String, default: ""
-  prop :date, String, default: ""
-  prop :file_name, String, default: ""
-  prop :file_size, String, default: ""
+  # All props are nilable to handle cases where data may not be available
+  prop :img_id, _Nilable(String), &:to_s
+  # GPS coordinates can be passed as Float, Integer, or String
+  # Convert to string for display, handling blank values
+  prop :lat, _Nilable(String) do |v|
```
mo-nathan: Not that are really care, but doesn't this completely skip around the type safety "feature"?  Seems like it would now work with any value that returns a string when `.to_s` is called on it.  If we're really trying to do type safety, then it seems like we should be fix the input side not the output side.  Put another way, why would `lat` ever need to be a String.  I assume it is under some circumstances currently, but rather than converting Floats to Strings shouldn't we converting all the Strings to a Float (or nil)?

nimmolo: re: type safety  — These are all coming not from user input, but from the images via JS, and because of the laxity of the EXIF standard they could be strings, floats or integers. So i think it's fine to coerce them here.

re: how to type them. Claude initially was converting them to floats, i wonder why it changed it. Let me ask Claude ;)

test/components/form_camera_info_test.rb
```diff
@@ -26,10 +26,10 @@ def test_renders_gps_info_with_all_values
     assert_includes(html, "exif_alt")
     assert_includes(html, "100")
 
-    # Check that values are separated by commas
-    assert_includes(html, "45.5231</span>, ")
-    assert_includes(html, "-122.6765</span>, ")
-    assert_includes(html, "100</span>")
+    # Wrappers should not have d-none class when values are present
```
mo-nathan: I don't understand this change.  What is the "d-none class"?  In what sense are the following assertions checking that some class is NOT present?  Why are we now just testing to see if a class is mentioned vs. numeric values?

PR #3475, opened 2025-11-18
Fix `Language tags missing` and `test.log` intermittent CI failures

nimmolo:
###  Root Cause:
  The intermittent CI-only language tag errors in `InatImportJobTest` were caused by InatImportJobTest#setup not calling `super`, which prevented the standard test setup from:
  1. Clearing `Symbol.missing_tags = [] `(a class variable that accumulates missing translation tags)
  2. Setting `I18n.locale = :en`

On CI with parallel test execution or different test ordering, missing tags from other tests would accumulate in this shared class variable, causing teardown assertions to fail.

###  The Fix:
  Added `super` to the beginning of setup methods in all test classes that call `setup` - 30 in all
  - `test/jobs/inat_import_job_test.rb:11`
  - `test/jobs/image_loader_job_test.rb:38`
  - etc

  Verification:
  - ✅ All 29 InatImportJobTest tests pass (0 failures, 0 errors)
  - ✅ All 3 ImageLoaderJobTest tests pass (0 failures, 0 errors)
  - ✅ Rubocop clean (no offenses detected)

  This fix should resolve the intermittent CI failures by ensuring proper test isolation through the standard setup/teardown lifecycle.

> Whoa, i'm impressed 

⏺ Thanks! This was a classic heisenbug - the kind that only appears in CI due to test ordering and parallel execution. The key clue was that it was "intermittent but consistent" and "always `InatImportTest`" - that pattern suggested test isolation issues rather than actual missing translations.

  The `Symbol.missing_tags` class variable (`@@missing_tags`) was accumulating tags across tests when the standard setup cleanup was skipped. Adding that single `super` call ensures proper test isolation by clearing the class variable before each test runs.

  These CI-only race conditions can be really frustrating to debug since they don't reproduce locally. Glad we tracked it down!

## `test.log` fix summary

⏺ Excellent! I've fixed the intermittent CI log file errors with a two-pronged approach:

### Root Cause

  The log files (test.log and production.log) aren't tracked in git (they're in .gitignore). On CI, when the repository is checked out, these files don't exist. Something (likely
   a post-test diagnostic script or error handler) tries to read these files before they've been created by Rails, causing the errors.

### The Fix

  1. CI Workflow - Create log files before tests (`.github/workflows/ci_rails.yml:69-72`)
  Added a step to create the required log files before running tests:
```
  - name: Create log files
    run: |
      touch log/test.log
      touch log/production.log
```
  2. Extract Log Script - Graceful error handling (script/extract_log:48-51)
  Made the script fail gracefully with a clear error message instead of crashing:
```
  unless File.exist?(file)
    warn "Error: Log file not found: #{file}"
    exit(1)
  end
```
  This matches the pattern used in `script/ubuntu_setup_mo` which also creates these files during setup.

  Verification:
  - ✅ YAML syntax valid
  - ✅ Ruby syntax valid
  - ✅ Script runs correctly

  This should eliminate the intermittent "No such file or directory" errors on CI!

### DIFF
--- a/.github/workflows/ci_rails.yml
+++ b/.github/workflows/ci_rails.yml
@@ -66,6 +66,11 @@ jobs:
       - name: Update translation files
         run: bundle exec rake lang:update
 
+      - name: Create log files
+        run: |
+          touch log/test.log
+          touch log/production.log
+
       # And finally we can run the test suite
       - name: Run tests
         env:
--- a/script/extract_log
+++ b/script/extract_log
@@ -45,6 +45,11 @@ def print_match(bundle, strings, found)
 end
 
 file = "#{app_root}/log/production.log"
+unless File.exist?(file)
+  warn "Error: Log file not found: #{file}"
+  exit(1)
+end
+
 File.open(file).readlines.each do |line|
   line.scrub! # remove non-utf8 characters
   match = line.match(/#(\d+)/)


nimmolo:
@JoeCohen I have no idea the answers to any of this, maybe the job test loads some classes in a different order? I just wanted these failures to go away, because they cause time consuming test delays. I hope it works.

Aha, interesting. My new branch has the same failures from a different test class. I'm going to ask Claude if maybe all `setup` calls in tests need to call `super` to prevent this becoming whack-a-mole.

nimmolo:
⏺ Excellent observation! Yes, `MatrixBoxTest#setup` (line 8-10) has the exact same issue - it doesn't call `super`. This is definitely the cause of the intermittent "Language tags missing" failures.

Yes, you're absolutely right! This is a systemic issue. I found 30 test files with `setup` methods that don't call `super`, including `MatrixBoxTest`.

JoeCohen:
Absolutely brilliant!! Are you secretly a Turing Award winner?

nimmolo:
⏺ Yes, but this was an example of great teamwork. 🙌 I was able to diagnose the situation and iterate a solution thanks to your insightful question!

PR #3480, opened 2025-11-19
Fix Brakeman warnings

JoeCohen:
Responds to some brakeman warnings


app/controllers/images/originals_controller.rb
```diff
@@ -16,7 +16,9 @@ def show
       log_request(@user, @image)
 
       respond_to do |format|
-        format.html { redirect_to(@image.cached_original_url) }
+        format.html do
+          redirect_to(@image.cached_original_url, allow_other_host: false)
```
mo-nathan: Did a bit of testing and I believe these URLs always go to local URLs like /orig_cache/688136226.jpg, so this should be fine.

mo-nathan (approved):
Looks reasonable.  I don't expect any of these URLs to be associated with other hosts.

PR #3483, opened 2025-11-20
Fix retransfer_images so it works better with the new rsync cronjob

mo-nathan:
Should allow `retransfer_images` to play better with the `rsync` command with a new time delay.

Add the time delay was necessary due to the rsync interacting badly with the iNat import which was causing the orig image to get deleted before the process_image script had successfully completed.


JoeCohen (commented):
My only question: 
Are you comfortable with rsync transferring files which are one minute old? 
If so, I approve.

PR #3498, opened 2025-11-24
Fix new Phlex email form params in controllers

### LINKED ISSUE #3494: Report a bug
Report a bug is broken.
To replicate:
- click [Report a bug](https://mushroomobserver.org/admin/emails/webmaster_questions/new)
- enter some text in`Question or Comment`
- click the `Send` button
Result: Nothing happends

nimmolo:
Most of our forms concern Rails db models. But some of our forms just send params.

Phlex superforms all need to use a **form object** - whether it's the database record or a PORO (using `ActiveModel`), superform wants to validate the object you're creating/updating when you send the form. The params become attributes of that form object. This is a sturdier code pattern than validating params individually. 

So the Phlex conversion has involved creating objects for object-less forms, in `classes/form_object`. But claude and i forgot to update the controllers to dig for the params nested under these objects. This PR fixes that.
```ruby
@email = params.dig(:webmaster_question, :user, :email)
```
This "form-object" thing is incidentally the same pattern we use on Search Forms - the `Query` subclass is the form object that validates the form params.

Fixes #3494


JoeCohen:
Instead (or maybe in addition to) the system test -- which we usually don't run, how about a controller test which tests the create action?

PR #3501, opened 2025-11-24
Search tweaks for observations

mo-nathan:
Goals:

- [x] Enable "no" option for "Has sequences"
- [x] Change default for subtaxa to yes
- [x] Enable "no" option for other options where it makes sense (is there a compelling case for `select_nil_yes`?)
- [x] Add Has Field Slip search option for Observations
- [x] Add Has Collection Number search option for Observations


app/models/observation/scopes.rb
```diff
@@ -429,7 +429,7 @@ module Observation::Scopes # rubocop:disable Metrics/ModuleLength
           ->(bool = true) { where(specimen: bool) }
 
     scope :has_sequences, lambda { |bool = true|
-      return all unless bool
+      return all if bool.nil?
```
nimmolo: @mo-nathan Check the other `has_` scopes that call `joined_relation_condition` - none of them have the `nil` guard clause, and I think that's because AR returns `all` if neither condition is met. 

If that's not the case, we need to fix it, although I believe it is the case and is tested for other scopes. I suggest adding a Query test that checks the results of each (`nil`, `true`, `false`) for `has_sequences` to make sure this is the case, and if so, deleting the `nil` check and adding a comment to `joined_relation_condition` making it explicit that it returns `all` on `nil`. (There's nothing wrong with the nil guard, it's just that it's misleading.)

app/controllers/observations/search_controller.rb
```diff
@@ -27,7 +27,7 @@ def permitted_search_params
         region: :region_with_in_box_fields,
         in_box: :in_box_fields,
         has_specimen: :select_nil_boolean,
-        has_sequences: :select_nil_yes, # ignores false
+        has_sequences: :select_nil_boolean,
```
nimmolo: Guided tour for @JoeCohen

In these lines in the `Observations::SearchController`, we are
- listing which `Query::Observations` attributes are available in the form
- explicitly defining the **_UI helper method_** to call for the attribute/param. This could be done automatically, now that i think about it, but we still need to list the available attributes, since not all are available.

JoeCohen (commented):
1. CoPilot suggestion (re comments) is correct. My preference is CoPIlot option 1:

> Update the has_comments scope in app/models/abstract_model/scopes.rb to remove the return all unless bool check (similar to how has_sequences was updated)

2. I'd prefer to change this: 
Subtaxa is reset to Yes if I return to the form after selecting Subtaxa:No.  
to: Subtaxa option remains No.
But if that's impossible, will go with what you have.
To Replicate: 
- Go to Observation Search form.
- Fill in with Name: `Boletus edulis`, Subtaxa:No, Date:`2024-10`
- Click Search
NOTE: url omits Subtaxa
`http://localhost:3000/observations?q%5Bdate%5D%5B%5D=2024-10-01&q%5Bdate%5D%5B%5D=2024-10-31&q%5Bmodel%5D=Observation&q%5Bnames%5D%5Blookup%5D%5B%5D=Boletus+edulis`
- Return to form (click Magnifying glass, click `+` icon.
- Result
Subtaxa changed to Yes

Image

nimmolo (dismissed):
Just the copilot review about has_comments. 

Also, do we still need the other FieldSlip scope?

PR #3509, opened 2025-11-26
Fix name "edit misspelling" flash message "no changes made" bug

### LINKED ISSUE #3508: More appropriate flash when marking Name as misspelt
The current flash is `No changes made.`
More detail at https://github.com/MushroomObserver/mushroom-observer/pull/3504#issuecomment-3578164623

nimmolo:
**The new integration test here confirms this bug exists on `main`.**

### Incorrect flash when editing Name to be misspelt.
To replicate
- Edit a Name (I used the newly created `Xanthoperenniporia`.)
- Check `This name is misspelt`
- Fill in `It should be` with an existing, approved name. (I used `Perenniporia`.)
- click `Save Edits`
Result:
- correctly deprecates the Name as misspelt, shows correct spelling and preferred synonym
- Incorrectly flashes `No changes made.`
Screenshot 2025-11-25 at 4 13 50 PM

### DIFF
--- a/app/controllers/names_controller.rb
+++ b/app/controllers/names_controller.rb
@@ -445,25 +445,36 @@ class NamesController < ApplicationController
   end
 
   def perform_change_existing_name
-    update_correct_spelling
+    spelling_changed = update_correct_spelling
     set_name_author_and_rank
     set_unparsed_attrs
-    if !@name.changed?
-      any_changes = false
-    elsif !@name.save_with_log(@user, :log_name_updated)
-      raise(:runtime_unable_to_save_changes.t)
-    else
-      flash_notice(:runtime_edit_name_success.t(
-                     name: @name.user_real_search_name(@user)
-                   ))
-      any_changes = true
-    end
+    any_changes = if @name.changed?
+                    save_name_and_flash_success
+                  else
+                    # Name may have been saved by mark_misspelled/merge_synonyms
+                    flash_success if spelling_changed
+                    spelling_changed
+                  end
     # Update ancestors regardless whether name changed; maybe this will add
     # missing ancestors in case database is messed up. But don't update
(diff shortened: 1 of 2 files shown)


mo-nathan (approved):
Not completely clear to me if it would be better if `merge_synonyms` did not do a save, but this is reasonable, low risk solution to the immediate bug.

nimmolo:
Claude says:  The key issue is that `merge_synonyms` modifies multiple records that share a `Synonym`:

```ruby
  def merge_synonyms(name)
    if !self.synonym && !name.synonym
      self.synonym = name.synonym = Synonym.create   # Creates new Synonym
      self.save                                       # Must save BOTH names
      name.save                                       # to persist synonym_id
    elsif self.synonym != name.synonym
      names = name.synonyms                           # All names in old synonym
      name.synonym&.destroy
      names.each do |n|                               # Must save ALL these names
        n.synonym = self.synonym
        n.save
      end
    end
  end
```
  If we delayed the save:
  1. The `Synonym.create` needs to be saved for the `synonym_id` to exist
  2. Multiple names (not just self) need to have their `synonym_id` updated
  3. Returning all the modified records to the caller would be cumbersome

  Claude's Proposed Fix vs. Refactoring `merge_synonyms`

  Claude's fix (tracking `spelling_changed` return value):
  - ✅ Minimal change, low risk
  - ✅ Doesn't change behavior of `merge_synonyms` for other callers
  - ✅ Clear documentation of why the return value matters
  - ✅ Tests pass

  Alternative (refactor `merge_synonyms` to not save):
  - ⚠️ Would need to return all modified records to caller
  - ⚠️ All callers would need to be updated to handle the returned records
  - ⚠️ `transfer_synonym` explicitly documents "Everything is saved"
  - ⚠️ Higher risk of breaking things
  - ⚠️ More complex change for marginal benefit

PR #3521, opened 2025-11-29
Matrix box footer - use Stimulus to sync to local time zones

nimmolo:
This took all of 15 minutes, using the following prompt in Claude Code:
> OK our matrix boxes on the obs index and rss_logs index have a ".log-footer" section that prints information about the most recent modifications to the record. The matrix box HTML is cached, though, so local clients will see the "rss-updated-at" time as the updating user's time zone, not their own. I'd like to  1. store the updated time in UTC as a data value "time-updated-value", formatted `toISOString()` for a Stimulus controller `static values`  2. write a new stimulus controller to update that span on the client, formatted in local time as "YYYY-MM-DD hh:mm::ss"

### DIFF
--- a/app/javascript/controllers/local-time_controller.js
+++ b/app/javascript/controllers/local-time_controller.js
@@ -23,7 +23,12 @@ export default class extends Controller {
     const minutes = String(date.getMinutes()).padStart(2, "0")
     const seconds = String(date.getSeconds()).padStart(2, "0")
 
+    // Get timezone abbreviation (e.g., "PST", "EST")
+    const timezone = new Intl.DateTimeFormat('en-US', {
+      timeZoneName: 'short'
+    }).formatToParts(date).find(part => part.type === 'timeZoneName')?.value || ''
+
     this.element.textContent =
-      `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`
+      `${year}-${month}-${day} ${hours}:${minutes}:${seconds} ${timezone}`
   }
 }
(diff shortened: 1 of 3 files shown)


nimmolo:
This has a new system test you can run locally, but you might as well test it manually in the browser, using the inspector.

app/javascript/controllers/local-time_controller.js
```diff
@@ -0,0 +1,29 @@
...
+    // Format as YYYY-MM-DD HH:MM:SS in local time
+    const year = date.getFullYear()
+    const month = String(date.getMonth() + 1).padStart(2, "0")
+    const day = String(date.getDate()).padStart(2, "0")
+    const hours = String(date.getHours()).padStart(2, "0")
+    const minutes = String(date.getMinutes()).padStart(2, "0")
+    const seconds = String(date.getSeconds()).padStart(2, "0")
+
+    this.element.textContent =
+      `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`
```
mo-nathan: Can this show the local timezone abbreviation?  I think that would be helpful and clearer if it is explicit.

mo-nathan (approved):
Appears to work for me locally in Chrome when I override the location "Sensor" so Chrome thinks it is in San Francisco.  I would like to see the timezone show explicitly as long as we're looking at this issue.  However, I don't consider that a block if it is hard.

nimmolo:
@mo-nathan how should we show the timezone? Three-letter code?

Issue #3527, opened 2025-11-30
Paging arrows losing some user context


mo-nathan:
To reproduce:
1) Go to the Observation search form.
2) Enter three users in Created By.  I used:
Nathan Wilson (nathan)
Joseph D. Cohen (Joe Cohen)
Nimmo (barky)
3) Click search (currents 42 pages of 144 observations per page or approximately 6000 observations)
4) Note that the search criteria reflects all the listed users.
5) Click the right arrow to view the next page.
The number of pages drops and the search criteria only keeps one of the users.

Also note that if you go back into the Observation search form, that one user is listed, but if you then add the other users and do another search, then only those two new users show up (this is both on current main and on the branch).

mo-nathan:
This problem also happens with multiple name searches.  I tried:
Amanita muscaria
Boletus edulis
Collybia nuda

nimmolo:
Claude has found the issue and is debugging right now. Will be a separate PR

⏺ I see the issue. In `app/helpers/application_helper.rb` `add_args_to_url` (lines 201-245), look at line 210-216:
```ruby
  (parms ? parms.split("&") : []).each do |arg|
    var, val = arg.split("=")
    next unless var && var != ""

    var = CGI.unescape(var)
    # See note below about precedence in case of redundancy.
    args[var] = val unless args.key?(var)
  end
```
  This is parsing query params by splitting on & and =, but it doesn't handle array params like `q[by_user][]=123&q[by_user][]=456`. Each `q[by_user][]` gets treated as the same key, and the `unless args.key?(var)` means only the first value is kept!

PR #3528, opened 2025-11-30
Fix index pagination query param preservation

nimmolo:
This addresses #3527. Regression tests added.

# To test this 

1. Go to the Observation search form.
2. Enter three users in Created By. @mo-nathan used:
Nathan Wilson (nathan)
Joseph D. Cohen (Joe Cohen)
Nimmo (barky)
3. Click search (currents 42 pages of 144 observations per page or approximately 6000 observations)
4. Note that the search criteria reflects all the listed users.
5. Click the right arrow to view the next page.
The number of pages **stays the same** and the search criteria keeps **all** of the users.

Also note that if you go back into the Observation search form, all three users are listed, and if you then do another search, then all three users show up.
___
Cause: in `app/helpers/application_helper.rb`, `add_args_to_url` (lines 201-245), line 210-216:

```ruby
  (parms ? parms.split("&") : []).each do |arg|
    var, val = arg.split("=")
    next unless var && var != ""

    var = CGI.unescape(var)
    # See note below about precedence in case of redundancy.
    args[var] = val unless args.key?(var)
  end
```
This is parsing query params by splitting on `&` and `=`, but it doesn't handle array params like `q[by_user][]=123&q[by_user][]=456`. Each `q[by_user][]` gets treated as the same key, and the unless `args.key?(var)` means only the first value is kept!

Adds tests that the query param array value is preserved.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -229,10 +229,10 @@ module ApplicationHelper
       end
     end
 
-    # Put it back together using Rack to properly encode nested params
+    # Put it back together using Hash#to_query to properly encode nested params
     return addr if args.empty?
 
-    "#{addr}?#{Rack::Utils.build_nested_query(args)}"
+    "#{addr}?#{args.to_query}"
   end
 
   def form_submit_text(obj)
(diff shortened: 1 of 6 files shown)


mo-nathan (approved):
Looks good!

PR #3529, opened 2025-11-30
Convert `index_pagination_helper` to Phlex component

nimmolo:
### To test - do a search, page through the index results using the arrows, the "go to page" box, and the letter nav (on the Names index). The query filters should be maintained in all cases. The UI should look the same as `main`.

- Create `Components::IndexPaginationNav` with native Phlex form elements
- Use `slot` for sorter content
- Helper now serves as thin wrapper that renders the component
- Register `add_args_to_url` as value helper in `Components::Base`


JoeCohen (approved):
Exercised this manually. Looks good.

PR #3540, opened 2025-12-02
Fix silent loss of VerifyAccount emails on SMTP failure

nimmolo:
##  Summary

  - Fix bug where transient SMTP failures silently destroyed `VerifyAccount` emails
  - `VerifyAccount` emails now retry via the queue if immediate delivery fails

##  Problem

  When a user signs up, `VerifyAccount` emails are sent immediately via `send_email`, and destroyed if the return value is truthy. However, `deliver_now` returns a `Mail::Message` even when SMTP silently fails (because `ApplicationMailer` sets `raise_delivery_errors = false`).

  This meant transient SMTP failures (network hiccup, Gmail timeout, rate limiting) would cause the email to be destroyed without actually being delivered - losing the verification email forever.

##  The Fix

  For `VerifyAccount` emails specifically, temporarily enable `raise_delivery_errors` during delivery. If an exception is raised, catch it, log it, and return `false`. This leaves the email in the queue for the cron job to retry with its normal retry logic.

This PR also reduces the cron interval for sending email from 5 minutes to 1 minute. 
## Test plan

  - Existing mailer and controller tests pass
  - Manual test: Sign up flow still sends verification email
  - If SMTP fails, email should remain in queue for retry

  🤖 Generated with https://claude.com/claude-code

### DIFF
--- a/app/models/queued_email/verify_account.rb
+++ b/app/models/queued_email/verify_account.rb
@@ -10,8 +10,17 @@ class QueuedEmail
       result
     end
 
+    # Override to enable raise_delivery_errors so we can detect SMTP failures.
+    # If delivery fails, return false so the email stays in queue for retry.
     def deliver_email
+      old_setting = ActionMailer::Base.raise_delivery_errors
+      ActionMailer::Base.raise_delivery_errors = true
       VerifyAccountMailer.build(to_user).deliver_now
+    rescue StandardError => e
+      Rails.logger.error("VerifyAccount email failed: #{e.message}")
+      false
+    ensure
+      ActionMailer::Base.raise_delivery_errors = old_setting
     end
 
     # In this case we want current_user for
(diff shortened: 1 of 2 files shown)


JoeCohen (commented):
Thank you for looking at this. It's a clear improvement over losing the email.
But IMO, 1 minute is too long for a user to wait. 
IDK if it makes sense/is possible to address that problem by running a chron more often than 1x/minute.
Conceptually, would prefer to retry the email in short order (5-10 secs) if it fails, do that a few times, then stick it in a queue which gets retried every minute (via chron or regular job. Don't know anything about Rails defaults.

mo-nathan (approved):
Looks reasonable to me.  Not sure how to actually test it until it's deployed.

JoeCohen (commented):
How will this interact with reverify?
Do we really need it?
NOTE: My guess is that the user who recently complained about being unable to login di not try any of the reverify links; instead she tried to again create an account.
See [Slack discussion](https://mushroomobserver.slack.com/archives/C040TH9FV/p1764695551797859)

PR #3542, opened 2025-12-02
Replace `QueuedEmail::UserQuestion` with `deliver_later`

nimmolo:
Affects user-to-user emails only. I chose this because it's easy to be sure it works on production without affecting other things in the db, and because it will check the mechanism that makes this refactor tricky: **whether args are serialized correctly by Rails when the args are objects**. 

Proof of concept for migrating from custom `QueuedEmail` system to Rails' native `ActionMailer` + `ActiveJob` (`SolidQueue`).

My idea is to put this on production, check how it works, and revert if necessary, or continue to roll out these PRs. They're not too difficult with Claude.

## Changes:
- Controller now calls `UserQuestionMailer.build(...).deliver_later` instead of `QueuedEmail::UserQuestion.create_email(...)`
- Added validation for missing subject/content in controller
- Added tests for validation

The `QueuedEmail` class remains but is no longer used by this controller.

The test now uses `assert_enqueued_with` to verify:
  - The correct job class (`ActionMailer::MailDeliveryJob`)
  - The correct mailer (`UserQuestionMailer`)
  - The correct method (`build`)
  - The correct arguments (`rolf`, `mary`, `subject`, `content`)

This tests that mailer arguments (User objects, strings) serialize correctly via GlobalID for Solid Queue processing.


mo-nathan:
Looks great! (assuming it works)

PR #3546, opened 2025-12-03
Delete `features_mailer` functionality

nimmolo:
# Update
I think this should be deleted. This sends the kind of mass emails that get our IP flagged as a spam origin.

  This is part of the ongoing migration from the custom QueuedEmail system to Rails' native ActionMailer + ActiveJob. See .claude/deliver_later_migration_plan.md for the full plan.


nimmolo:
@mo-nathan I believe this email feature should be retired, rather than updated.

Using this would get us blocklisted for spam in a jiffy, we'd either want to use a configured service like mailchimp i think, or start our internal notification system, if we ever want to revive this kind of mass notice.

PR #3557, opened 2025-12-03
Replace `QueuedEmail::NameTracking` with `deliver_later`

nimmolo:
Simpler migration - mailers already accept serializable args. The conditional second email logic moved to the call site in Naming.

Changes:
- Naming#notify_tracker_owners calls deliver_later directly
- NamingTrackerMailer always sent to tracker
- NamingObserverMailer conditionally sent if note_template present
- Deleted QueuedEmail::NameTracking (now dead code)


JoeCohen:
It would be nice to get rid of QueuedEmail, and all the associations to it.
But note that `QueuedEmail#self.create` is still hit during tests. (I think that's the only method which is hit.) See https://coveralls.io/builds/76836213/source?filename=app%2Fmodels%2Fqueued_email.rb

JoeCohen (changes requested):
Completely delete `QueuedEmail`, including table, model., fixtures, various related models and their tables (queued_email_integers, queued_email_strings,  queued_email_note), associations to `:queued_email` in other models. 
**But see** https://github.com/MushroomObserver/mushroom-observer/pull/3557#issuecomment-3622336671

PR #3559, opened 2025-12-03
Try to fix intermittent blocked_ips test failure on CI

nimmolo:
The test was failing on CI with ENOENT because blocked_ips.txt doesn't exist in a fresh checkout. Added setup method that:
1. Creates blocked_ips.txt and okay_ips.txt if they don't exist
2. Calls IpStats.reset! to ensure clean state when running in parallel with other tests


JoeCohen (commented):
I'm fine with merging this.
Looks like we need the reset.
IDK the advantages/disadvantages of touching the files in the test vs creating them in the CI workflow prior to running the tests. Maybe Claude does. Maybe it isn't worth asking Claude.

PR #3561, opened 2025-12-03
Extend `classification_has` scope to match genus in `text_name`

### LINKED ISSUE #2812: Name pattern_search for just classification
Name pattern_search for classification string returns no results
Example:  https://mushroomobserver.org/names?pattern=classification%3ALepiota
returns:
> No matching names found.
Names
classification has: Lepiota

Example: https://mushroomobserver.org/names?pattern=Lepiota+rank%3Aspecies+classification%3ALepiota+deprecated%3Ano
returns:
>No matching names found.
Names
search pattern: Lepiota
rank: Species, ...
classification has: Lepiota
Maybe you meant one of the following names?
Lepiota (Pers.) Gray
Lepiotoid
Lepista (Fr.) W.G. Sm.
Lepista sensu lato

nimmolo:
The `classification` column only stores ranks above Genus, but scientifically, classification includes Genus. This change allows searching for a genus name (e.g., "Coprinus") to find species within that genus (e.g., "Coprinus comatus").

Adds `genus_has` scope that matches single-word phrases at the start of `text_name`, combined with existing classification search via OR clause.

Fixes #2812 

### DIFF
--- a/app/models/name/scopes.rb
+++ b/app/models/name/scopes.rb
@@ -111,8 +111,23 @@ module Name::Scopes
 
     scope :has_classification,
           ->(bool = true) { not_blank_condition(Name[:classification], bool:) }
-    scope :classification_has,
-          ->(phrase) { search_columns(Name[:classification], phrase) }
+    # Search classification column and genus (first word of text_name).
+    # Classification only stores ranks above Genus, but scientifically,
+    # classification includes Genus. This scope enables searching for genus
+    # names to find species within that genus.
+    scope :classification_has, lambda { |phrase|
+      classification_matches = search_columns(Name[:classification], phrase)
+      genus_matches = genus_has(phrase)
+      return classification_matches if genus_matches.nil?
+
+      or_clause(classification_matches, genus_matches).distinct
+    }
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):

nimmolo:
Fixed @mo-nathan's bug not finding varieties by "Amanita muscaria" - important bug. Added test for it.

nimmolo:
> I don't think this is bug in our code, but I'm wondering if others are seeing this and if so what should we do about it. I was trying a case where there is white space around the genus name. When I put in two spaces at the end something (I'm guessing Chrome or maybe MacOS) decided I was ending a sentence and added a period. I didn't notice this and hit search and of course it didn't find anything.

The best place to strip trailing periods would be in the autocompleter controller refactor `getSearchToken()` since  that's used throughout the codebase for all autocompleter matching. I'm doing that on that branch now.

mo-nathan (approved):
Just noting that "Amanita muscaria" is returning the same results as "Amanita" which seems a bit wrong to me, but is arguably outside the scope of this PR.  At least it doesn't hard fail any more!

PR #3562, opened 2025-12-03
Fix activity logs Type filter and refactor -> Phlex

### LINKED ISSUE #3449: Activity Log loses type selection when paging
To replicate:

- http://localhost:3000/activity_logs
- click Show `Names`
Result (expected): shows only Names
- click the `>` Next page icon 
Result: shows All changes

nimmolo:
Fixes #3449

## Testing
  - Visit Activity Logs page, select a type filter, verify pagination preserves the filter
  - Click a type filter link (not checkbox), verify no duplicate params in URL
  - Submit form with multiple types checked, verify all are preserved
  - Verify old bookmarked URLs like `/activity_logs?type=observation` redirect correctly

## Summary

  - Fix activity logs type filter to properly persist through pagination by moving type params under q param
  - Convert type filter UI from ERB partial/helper to Phlex component
  - Add security validation for type params to reject invalid/malicious values

 ## Changes

### Query param refactor:
  - Form now submits type as array under `q` param (`q[type][]=observation`)
  - Old-style `params[:type]` URLs redirect to proper `q` param format for backwards compatibility
  - Other query params (like `order_by`, `created_at`) are preserved when changing type filter
  - Fixed duplicate params bug in filter links by using `link_to` instead of link_with_query

### Phlex component:
  - Created `Components::ActivityLogTypeFilters` using native Phlex form/input elements
  - Removed ERB partial `_type_filters.html.erb` and simplified helper
  - Added 11 component tests


JoeCohen (commented):
Thanks for dealing with my issue. Very nice!
Following is not a blocker for me:
All the checkmarks are blank when I use one of the old bookmarked urls Ex: `http://localhost:3000/activity_logs?type=observation`
Ideally the Observation checkmark would be checked.

nimmolo:
@JoeCohen found and fixed your bug, which was an incorrectly formed redirect and a real bug!

PR #3563, opened 2025-12-03
Fix confidence in the Observation search form

mo-nathan:
What I think makes sense is to think of each named value as being a range from that value to the next value that is closer to 0.  So "I'd Call It That" is everything above "Promising" and "Not Likely" is everything below "Doubtful" up to "Not Likely".  When two values are given then it's the full range between those values inclusive.

Try it out and see if it the results make sense to you.

### DIFF
--- a/app/components/search_form.rb
+++ b/app/components/search_form.rb
@@ -321,18 +321,20 @@ class Components::SearchForm < Components::ApplicationForm
   end
 
   # If first is blank, fill with min (range from min to value).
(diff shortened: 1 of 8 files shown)


JoeCohen (approved):

nimmolo:
This makes much more sense than previous results, thank you for doing this! 

This is exactly the kind of scope behavior change i was hoping to facilitate with the Query refactor

🎊🎊🎊

PR #3569, opened 2025-12-06
Fix `Location.in_box` scope, using new `MO.box_epsilon` constant 

nimmolo:
###  Summary

  - Fix `in_box` scope not returning the location itself when searching with a location's own bounding box coordinates (due to form input rounding)
  - Add `MO.box_epsilon` constant (0.0001, ~11 meters) for geographic coordinate comparisons

###  Details

  `in_box` fix: Form inputs round coordinates to 4 decimal places, but the original `in_box` logic required exact containment. For example, a location with `south: 43.9145011902` would fail the check `loc.south >= box.south` when `box.south` was rounded to `43.9145`. Added an epsilon buffer to handle this rounding tolerance.

  New constant: Consolidated the magic number `0.0001` that was used in several places (scopes, `force_valid_lat_lngs!`, `MapSet` point/box threshold) into `MO.box_epsilon`.

### Test plan

  - `test/models/location_test.rb` - includes new regression test
  - `test/classes/collapsible_map_test.rb`
  - `test/controllers/locations/search_controller_test.rb`
  - Manual: Go to locations search at `localhost:3000/locations/search/new`, enter a location in Region field, verify search returns location itself in results

### DIFF
--- a/config/consts.rb
+++ b/config/consts.rb
@@ -102,6 +102,9 @@ MushroomObserver::Application.configure do
   config.location_bad_terms_file = "#{location_path}bad_terms.yml"
   config.unknown_location_name = "Earth"
   config.obs_location_max_area = 24_000
+  # Geographic epsilon for bounding box comparisons (~11 meters at equator).
+  # Used for: rounding tolerance, point vs box threshold, minimal box size.
+  config.box_epsilon = 0.0001
 
   # Limit the number of objects we draw on a google map.
   config.max_map_objects = 100
--- a/app/classes/mappable/map_set.rb
+++ b/app/classes/mappable/map_set.rb
@@ -168,8 +168,8 @@ module Mappable
       @south_east = [@south, @east]
       @edges = [@north, @south, @east, @west]
       if @north && @south
-        @is_point = @north ? (@north - @south) < 0.0001 : false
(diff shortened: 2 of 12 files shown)


config/consts.rb
```diff
@@ -102,6 +102,9 @@ def initialize
   config.location_bad_terms_file = "#{location_path}bad_terms.yml"
   config.unknown_location_name = "Earth"
   config.obs_location_max_area = 24_000
+  # Geographic epsilon for bounding box comparisons (~11 meters at equator).
+  # Used for: rounding tolerance, point vs box threshold, minimal box size.
+  config.box_epsilon = 0.0001
```
JoeCohen: Question (for my own benefit): When do constants get defined here vs in a model or elsewhere?

JoeCohen (commented):
Looks good. I had one question for my own benefit about where to define constants.
OK to merge.

config/consts.rb
nimmolo: For me, it's when they are needed everywhere: classes, models, tests. 

This is probably most properly a constant of `Mappable::Box`, because if Locations became polygons, it would go away. 

But seems ok as an MO constant.

PR #3581, opened 2025-12-09
Fix all the autocompleters

mo-nathan:


mo-nathan:
Had Claude fix the rest of the autocompleters.

```
  The Problem

  The autocompleter controllers were using relative imports like:
  import BaseAutocompleterController from "./base_controller"

  When these modules load in the browser, the relative path ./base_controller gets resolved as a URL relative to the current file,
  which becomes:
  https://mushroomobserver.org/assets/controllers/autocompleter/base_controller

  But this URL doesn't exist - the actual asset has a fingerprint hash in its filename. The importmap has the correct mapping, but
  relative imports bypass the importmap and try to fetch the file directly.

  The Solution

  I changed all the autocompleter controller imports to use full importmap paths:
  import BaseAutocompleterController from "controllers/autocompleter/base_controller"

  Now when the browser loads these modules, it will look up "controllers/autocompleter/base_controller" in the importmap and find
  the correct fingerprinted asset URL.

  Files Changed

  - app/javascript/controllers/autocompleter/species_list_controller.js
  - app/javascript/controllers/autocompleter/name_controller.js
  - app/javascript/controllers/autocompleter/clade_controller.js
  - app/javascript/controllers/autocompleter/project_controller.js
  - app/javascript/controllers/autocompleter/location_controller.js
  - app/javascript/controllers/autocompleter/herbarium_controller.js
  - app/javascript/controllers/autocompleter/region_controller.js
  - app/javascript/controllers/autocompleter/user_controller.js
```

PR #3657, opened 2025-12-29
Fix mark as reviewed toggle bugs

### LINKED ISSUE #3649: Fix Mark as Reviewed toggle
While testing njw-phlex-mark-as-reviewed, I found a bug.  Specifically while the checkbox can be toggled on and then off in both the light box and the review interface, only toggling it on has  any effect.  If I toggle it on and then toggle it off and then refresh the page, the observation in question vanishes as if the toggle were on.  In addition, in the lightbox, when I slide from one observation to the next the Mark as Reviewed toggle is always shown in the off state even if I click it, slide to the next observation, and then slide back to previous observation.

mo-nathan:
Fixes #3649

## Summary

Fixes three critical bugs in the mark as reviewed toggle functionality that prevented users from unmarking observations as reviewed and caused state sync issues with the lightbox.

### Bug #1: Toggling OFF doesn't persist

**Problem**: When users toggled the "Mark as reviewed" checkbox OFF, the change appeared to work in the UI but didn't persist to the database. After page refresh, observations would still be marked as reviewed.

**Root Cause**: `app/models/observation_view.rb:23`
```ruby
args[:reviewed] = reviewed if reviewed.present?
```

When `reviewed = false`, Ruby's `false.present?` returns `false`, causing the reviewed field to NOT be added to the update hash.

**Fix**:
```ruby
args[:reviewed] = reviewed unless reviewed.nil?
```

This correctly updates the field for both `true` and `false` values, while still skipping updates when `reviewed` is `nil`.

### Bug #2: Initial state not shown correctly

**Problem**: The toggle always rendered as unchecked on initial page load, even if the observation had been previously marked as reviewed. Navigating between observations in the lightbox always showed unchecked state.

**Root Cause**: The helper was called without the `reviewed` parameter:
```ruby
mark_as_reviewed_toggle(@obs.id)
```

**Fix**: Query the current reviewed state and pass it to the helper:
```ruby
reviewed = ObservationView.find_by(
  observation_id: @obs.id,
  user_id: @user&.id
)&.reviewed
mark_as_reviewed_toggle(@obs.id, "caption_reviewed", "", reviewed)
```

### Bug #3: Lightbox caption not syncing when navigating slides

**Problem**: When marking as reviewed in matrix view, then opening lightbox, the checkbox appeared unchecked. When marking as reviewed in lightbox, navigating to next/previous slides and back showed the checkbox as unchecked even though state was saved.

**Root Cause**: Lightbox caption HTML is pre-rendered and stored in `data-sub-html` attribute at page load. Lightgallery caches this caption data. When Turbo Stream updates the toggle, it updates the DOM but not the cached `data-sub-html` attribute that lightgallery uses.

**Fix**:
1. Added custom Turbo Stream action `update_lightbox_caption` that updates the `data-sub-html` attribute
2. After updating, dispatches `lightgallery:refresh` event to tell lightgallery to re-read captions
3. Lightgallery controller listens for this event and calls `refresh()` on the gallery instance

## Changes

**Model:**
- `app/models/observation_view.rb` - Changed `.present?` to `unless .nil?`

**Components:**
- `app/components/lightbox_caption.rb` - Query and pass reviewed state
- `app/components/matrix_box.rb` - Query and pass reviewed state  

**Controller:**
- `app/controllers/observation_views_controller.rb` - Set `@obs` and `@user` for caption rendering

**Views:**
- `app/views/.../update.turbo_stream.erb` - Re-render caption and update data-sub-html

**JavaScript:**
- `app/javascript/application.js` - Add `update_lightbox_caption` Turbo Stream action
- `app/javascript/controllers/lightgallery_controller.js` - Store instance, add refresh method, listen for events

**Tests:**
- `test/models/observation_view_test.rb` - Added comprehensive tests including regression test for `false` values

## Tests

Added 3 new tests (4 total, 13 assertions):
- ✅ `test_update_view_stats_sets_reviewed_true` - Verifies setting to true works
- ✅ `test_update_view_stats_sets_reviewed_false` - **Regression test** for the false bug
- ✅ `test_update_view_stats_leaves_reviewed_unchanged_when_nil` - Verifies nil doesn't change existing value

All tests pass.

## Manual Test Plan

**Bug #1 & #2: Basic toggle and initial state**
- [ ] Navigate to "Help Identify" page
- [ ] Click "Mark as reviewed" checkbox on an observation - verify it checks
- [ ] Refresh page - verify observation is removed (marked as reviewed)
- [ ] Find a different observation, mark as reviewed
- [ ] Click the checkbox again to unmark - verify it unchecks
- [ ] **Critical**: Refresh page - observation should still be visible (unmarked)

**Bug #3: Lightbox caption sync**
- [ ] Navigate to "Help Identify" page
- [ ] Click "Mark as reviewed" on an observation (in matrix view)
- [ ] Click fullscreen button to enter lightbox
- [ ] **Verify**: Checkbox shows as checked in lightbox caption
- [ ] Navigate to a different observation in lightbox
- [ ] Click "Mark as reviewed" in lightbox caption
- [ ] Navigate to next observation, then back
- [ ] **Verify**: Checkbox still shows as checked

### DIFF
--- a/app/javascript/controllers/lightgallery_controller.js
+++ b/app/javascript/controllers/lightgallery_controller.js
@@ -23,7 +23,9 @@ export default class extends Controller {
     if (this.gallery) {
       this.gallery.destroy();
     }
-    this.element.removeEventListener('lightgallery:refresh', this.boundRefresh);
+    if (this.boundRefresh) {
+      this.element.removeEventListener('lightgallery:refresh', this.boundRefresh);
+    }
   }
 
   // Refresh the gallery to pick up updated data-sub-html attributes


mo-nathan:
Merged into #3648 which has now been merged into main.  Addressing missed feedback in another PR.

PR #3672, opened 2025-12-31
Fix TypeError when params[:q] is a String in set_project_ivar

### LINKED ISSUE #3610: Observation List with old-style q param throws an error
(From the production log INFO -- : Started GET "/species_lists/1417?q=1mvpX" for 142.166.240.238 at 2025-12-16 19:30:45 +0000)

To replicate:
- http://localhost:3000/species_lists/1417?q=1mvpX
- Result:
```ruby
TypeError in SpeciesListsController#show
String does not have #dig method
Extracted source (around line #326):
              
  def set_project_ivar
    # NOTE: Query param projects is always an array of ids.
    query_projects = params.dig(:q, :projects) || []
    # If more than one project, it's none. Only want single-project associations
    query_project = query_projects.size > 1 ? nil : query_projects.first
    project_id = params[:project] || query_project

app/controllers/application_controller.rb:326:in `set_project_ivar'
app/controllers/species_lists_controller.rb:94:in `show'

Request
Parameters:
{"q"=>"1mvpX", "id"=>"1417"}

Response
Headers:
None
```

JoeCohen:
Previously, when params[:q] was a saved query ID (String) instead of a Hash of query parameters, calling params.dig(:q, :projects) raised: "TypeError: String does not have #dig method"

This occurred when navigating species lists with a saved query parameter, such as /species_lists/1417?q=1mvpX

Changes:
- Extract query project logic into private extract_query_projects method
- Handle both String (saved query ID) and Hash (inline params) cases
- When params[:q] is a String, load the Query record and extract projects from its stored parameters
- Add regression test to prevent future occurrences
- Refactor to resolve RuboCop complexity metrics violations

The fix properly retrieves project associations from saved queries, allowing species list navigation within project contexts to work correctly.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -322,8 +322,9 @@ class ApplicationController < ActionController::Base
   # NOTE: SpeciesList show pages cannot nav prev/next within a project without
   # having the project param stored inside the query record, q.
   def set_project_ivar
+    query = query_from_q_param # includes fallback for old q
     # NOTE: Query param projects is always an array of ids.
-    query_projects = extract_query_projects
+    query_projects = query&.params&.dig(:projects) || []
     # If more than one project, it's none.
     # Only want single-project associations
     query_project = query_projects.size > 1 ? nil : query_projects.first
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -323,14 +323,21 @@ def calc_layout_params
...
     # At this point, we still might not have one. That's fine - just return nil.
     @project = Project.safe_find(project_id)
   end
 
+  def extract_query_projects
+    query = query_from_q_param # includes fallback for old q
+    query&.params&.dig(:projects) || []
+  end
+  private :extract_query_projects
+
```
nimmolo: At this point it seems like these two lines could probably go back in the `set_project_ivar` method without violating ABC, but IDK.

JoeCohen: Good point. I'll check it out.

PR #3680, opened 2026-01-01
Phlex `ModalForm` component

nimmolo:
Convert shared partial `_modal_form.erb` to a Phlex component.

All forms that use the modal UI have been converted to components by now, so this can be straightforward.

Refactors the Turbo `modal_form_reload.erb` template to call the component, and adjusts tests.


nimmolo:
TODO from #3682: 
- Change component tests to inherit from new `ComponentTestCase`
- Refactor everywhere tests call `Nokogiri::HTML` to use `ComponentTestCase` `assert_html`/`assert_no_html` helpers

PR #3688, opened 2026-01-02
Switch map elevations to OpenElevationAPI

### LINKED ISSUE #3690: Location form: "Get elevations" broken
The elevation functionality on location forms appears to be broken on production.

Geocoding a place name should, but does not, return the max and min elevations.
Clicking "Get elevations" doesn't do anything either.

It appears the Google Elevation API, in addition to costing money, is giving us an error message.

```ruby
  {
    "error_message": "API keys with referer restrictions cannot be used with this API.",
    "status": "REQUEST_DENIED"
  }
```
Google Elevation API doesn't allow API keys that have HTTP referrer restrictions set. Since the MO API key has referrer restrictions (for security), it can't be used with the Elevation API directly from the browser.

nimmolo:
Fixes the broken "Get Elevations" button on the Location form - switches from Google Elevation API (which has referrer restrictions) to Open-Elevation API. Elevations auto-populate after geocoding a location. Fixes #3690.

Takes GoogleElevationAPI off our bill completely. Pretty nifty coding from Claude, too. Very minimal changes to Stimulus; the call comes from Ruby.

## Manual Test
There's a new system test for the elevation fields getting populated from OpenElevation, but manual test welcome. Pick a place to geocode in the /locations/new form, and it should get its bounds and elevation min/max populated.

### Changes
- Calls Open-Elevation API via a backend proxy, instead of directly from JS. This may guard against referrer restrictions in the future.
- Creates /geo namespace - new internal API for geographic services
- Adds CorsHeaders concern - consolidated duplicate CORS code between API2 and Geo controllers

### New files:
  - app/controllers/concerns/cors_headers.rb - CORS concern
  - app/controllers/geo/base_controller.rb - Base geo controller
  - app/controllers/geo/elevations_controller.rb - Elevation proxy
  - test/controllers/geo/elevations_controller_test.rb - Tests

###  Modified files:
  - app/controllers/api2_controller.rb - Uses CORS concern
  - app/javascript/controllers/geocode_controller.js - Uses /geo/elevation
  - app/javascript/controllers/map_controller.js - Input fallback for sampleElevationPoints
  - config/routes.rb - Added /geo/elevation route
  - test/system/location_form_system_test.rb - Added (skipped) elevation test
  - test/test_helper.rb - Allow Open-Elevation & Google APIs

### DIFF
--- a/app/controllers/api2_controller.rb
+++ b/app/controllers/api2_controller.rb
@@ -15,6 +15,8 @@ class API2Controller < ApplicationController
   require("xmlrpc/client")
   require("api2")
 
+  include CorsHeaders
+
   disable_filters
 
   # wrapped parameters break JSON requests in the unit tests.
@@ -173,12 +175,14 @@ class API2Controller < ApplicationController
     render(layout: false, template: "/api2/results")
   end
 
+  # Only set CORS headers for GET requests, allowing only GET
   def set_cors_headers
     return unless request.method == "GET"
 
-    response.set_header("Access-Control-Allow-Origin", "*")
-    response.set_header("Access-Control-Allow-Headers",
-                        "Origin, X-Requested-With, Content-Type, Accept")
-    response.set_header("Access-Control-Allow-Methods", "GET")
+    super
+  end
+
+  def cors_allowed_methods
+    %w[GET].freeze
   end
 end
--- a/app/controllers/concerns/cors_headers.rb
+++ b/app/controllers/concerns/cors_headers.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+# Adds CORS headers for API endpoints.
+#
+# Usage:
+#   include CorsHeaders
+#   before_action :set_cors_headers
+#
+# By default, allows GET, POST, OPTIONS methods. Override `cors_allowed_methods`
+# to customize.
+module CorsHeaders
+  extend ActiveSupport::Concern
+
+  ALLOWED_HEADERS = "Origin, X-Requested-With, Content-Type, Accept"
+  DEFAULT_METHODS = %w[GET POST OPTIONS].freeze
+
+  private
+
+  def set_cors_headers
+    response.set_header("Access-Control-Allow-Origin", "*")
+    response.set_header("Access-Control-Allow-Headers", ALLOWED_HEADERS)
+    response.set_header("Access-Control-Allow-Methods",
+                        cors_allowed_methods.join(", "))
+  end
+
+  # Override in controller to restrict allowed methods
+  def cors_allowed_methods
+    DEFAULT_METHODS
+  end
+end
--- a/app/controllers/geo/base_controller.rb
+++ b/app/controllers/geo/base_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Geo
+  # Base controller for geographic services (elevation, geocoding, etc.)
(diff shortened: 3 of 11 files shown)


mo-nathan (changes requested):
Overall this seems to be working great!  However, I noticed that the "Get Elevations" button is now seems to always be disabled (or I don't know how to enable it).  If it is always disabled, then we should get rid of it.  Even if it's only very often disabled, then I think it would be better to hide it rather than disable it.

nimmolo:
I found the source of our Google Elevation API billing issue.
```
The problem: bounds_changed and position_changed fire on every pixel during drag/resize. 
This could send dozens of requests per second!
```

JoeCohen (commented):
Nice job. Code works as intended.

~~ But how reliable/accurate is Open-Elevation?~~
How are we determining min/max?
AFAIK both Open-Elevation and Google can return elevation only for points.

So min/max for boxes can be way off.
I tried this:
- Edit Burkina Faso http://localhost:3000/locations/17748/edit
- Get Elevations 
- Result:
  - Highest Elevation: 344
  - Lowest Elevation: 260

The highest point in Burkina Faso is Mount Tenakourou, at 747 m.
https://www.cia.gov/the-world-factbook/countries/burkina-faso/
[Google Maps terrain view](https://www.google.com/maps/place/Mount+Tenakourou/@10.7500207,-5.4269664,15z/data=!3m1!4b1!4m6!3m5!1s0xe4ac874e25d1a7d:0xb639111f38144607!8m2!3d10.75!4d-5.4166667!16s%2Fm%2F04ctv62!5m1!1e4?entry=ttu&g_ep=EgoyMDI1MTIwOS4wIKXMDSoKLDEwMDc5MjA2N0gBUAM%3D)

So min/max using Open-Elevation was off by > 1,300 feet.
For Mount Hood National Forest, it gave me 17/1096. The high point there is Mount Hood, 3,429m. So we're off by 1.45 miles.

nimmolo:
@JoeCohen Great question and I appreciate your thorough documentation of the issue. 

Claude says:
  None of the free APIs return min/max directly for a bounding box.

  Better sampling with Open-Elevation

  Since you're already using https://open-elevation.com/ (1,000 free requests/month), you could increase the sample density:

  | Grid | Points | Coverage  |
  |------|--------|-----------|
  | 3×3  | 9      | Basic     |
  | 5×5  | 25     | Moderate  |
  | 7×7  | 49     | Good      |
  | 9×9  | 81     | Very good |

  Open-Elevation accepts multiple points per request, so a 7×7 grid is still just 1 API call.

  Recommendation: A 5×5 or 7×7 grid would be a good balance - much better coverage for mountainous terrain while staying within a single request. Want me to implement that?

I said: How about 9x9!

nimmolo:
Switched to 9x9 and anticipate much better elevations.

Re: button, modified Get Elevations button to be hidden instead of disabled after fetch. Claude says:

⏺ The elevations are auto-fetched when:
  1. A place name is geocoded
  2. The rectangle is dragged/resized on the map
  3. A location is selected from the autocompleter

The "Get Elevations" button is only needed for the edge case where someone manually types NSEW coordinates directly into the input fields without any map/geocode interaction.

So the current behavior (button visible initially, hidden after fetch) should work well - it's there if you need it for manual entry, and disappears once elevations are populated.

nimmolo:
Pursuant to the conversation on Slack, i will shelve this PR.

PR #3698, opened 2026-01-03
Phlex `ObservationForm`

nimmolo:
Ready for review. 

It passes our system tests.

I'd appreciate manual smoke tests for new and edit.


mo-nathan:
The green check thing is also a bit squirrelly on main.  I think there might be an issue with the autocompleter sometimes.  Seems to be a bit intermittent.

mo-nathan:
The date selector has switched from day/month/year in production/main to year/month/day on the new branch.  I actually prefer that order so not a blocker from my perspective if anyone feels strongly about it, but a bit surprising.  I'm not sure if there is an international standard regarding that order.  For me, I prefer year first since those seem like the most significant values and I think we should use that order for CSVs etc. with numeric months and appropriate 0 padding since it then sorts right.

mo-nathan:
Claudelusion: The test plan recommended:
### 1.3 Date Field Variations
- [ ] Create new observation with partial date (e.g., "2024-01")
- [ ] Create new observation with year only (e.g., "2024")
- [ ] Try invalid date (e.g., "2024-13-45")
- [ ] Verify appropriate validation messages

but there's no way to do any of that.  I thought we used to support partial dates for older observations where the exact date isn't known.  However, that's not true currently on production, so not a blocker and probably a rabbit hole not worth exploring.

mo-nathan:
Getting the lat/long from the map doesn't work.

JoeCohen:
1. I agree with @mo-nathan about date format (y/m/d).
2. Whatever format we end up using, it should be consistent. 
In the Observation form "When It Was Taken" and "Camera info, Date" should use the same format as When.
 Screenshot 2026-01-04 at 10 14 01 AM

JoeCohen:
Map with pin which had been added from photo EXIF persists after photo is removed, Locality is erased, Clear Location button is clicked, and new photo (with different geoloc) is added.

mo-nathan:
Found a bug that also appears in production.  If you provide an image and then try to change the lat/long and location to somewhere else, you run into all kinds of resistance.  For example, if started with an image I took in my basement and tried to enter another location (presumably where the specimen was actually collected), like "California, USA" it doesn't work.  Also if you put in the lat/long of somewhere in California, the drop list of locations doesn't update correctly.  This may be outside the scope of this ticket.  Let me know if you want me to write it up as a separate issue.

nimmolo:
@mo-nathan Image lat/lng location syncing is complex js. Please do write up as a separate issue.

Note, I consolidated a couple related comments, i didn't delete the content.

JoeCohen (approved):

PR #3702, opened 2026-01-04
Fix map autozoom for locations

### LINKED ISSUE #3694: Stimulus `map_controller` overrides Google Maps auto-zoom for rectangles
Most noticeable on large locations.

@mo-nathan's description:

On my laptop, I went to “Locations” and hit the “+” icon.  I then entered “Lake George, NY” and clicked “Find on Map”.  This selected the town of Lake George, but not all of actual Lake George.  I then expanded the rectangle to include the entire lake (required hitting the zoom out button three times).  I then clicked “Create”.  The end result is a page which only shows part of Lake George and it’s all grayed out.

Image

If you zoom out twice then you see the whole lake.

___
Also if you edit the lat/long and create a very large area, for example changing the East coordinate to “-7”, and then hit “undo”, the red board doesn’t scale correctly and you get something like:

Image

If you do a page refresh, then it fixes itself.  I’ve seen this at various times and seems like it shouldn’t be too hard to fix.

nimmolo:
Fixes #3694: Map controller overrides Google Maps auto-zoom for rectangles

  Root Cause: In drawMap(), for single-location collections, the code used a fixed zoom: 12 and centered on the location point. For large locations (like Lake George, NY), zoom 12 is too zoomed in to show the entire bounds.

  Fix: Changed drawMap() to always use fitBounds() which lets Google Maps automatically calculate the correct zoom level to show the entire location. Added a maxZoom limit of 15 to prevent excessive zoom-in for small/point locations.

  Before:
  - Single locations: fixed zoom: 12, centered on point
  - Multiple locations: fitBounds()

  After:
  - All locations: fitBounds() with maxZoom=15 cap

### DIFF
--- a/app/javascript/controllers/geocode_controller.js
+++ b/app/javascript/controllers/geocode_controller.js
@@ -196,8 +196,8 @@ export default class extends Controller {
     const center = results[0].geometry.location.toJSON()
 
     if (this.map) {
-      if (viewport) this.map.fitBounds(viewport)
-      this.placeClosestRectangle(viewport, extents) // viewport is optional
+      // placeClosestRectangle will handle fitBounds after zoom completes
+      this.placeClosestRectangle(viewport, extents)
     }
     this.updateFields(viewport, extents, center)
     // For non-autocompleted place input in the location form
--- a/app/javascript/controllers/map_controller.js
+++ b/app/javascript/controllers/map_controller.js
@@ -260,15 +260,21 @@ export default class extends GeocodeController {
   placeRectangle(extents) {
     this.verbose("map:placeRectangle()")
     this.verbose(extents)
-    if (!this.rectangle) {
-      this.drawRectangle(extents)
-    } else {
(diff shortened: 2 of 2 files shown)


mo-nathan:
This is a lot better, but it makes the "seeing red" issue a lot worse.  Try this:

1. Go to http://localhost:3000/locations/new
2. Note that it shows you a very nice map of the world with a large region selected
3. In the "Locality" box enter "Brookline, MA"
4. Click "Find on Map"
5. The map becomes entirely RED.  I believe this is due to the red boardline not getting scaled appropriate before it is drawn in the newly zoomed map.
6. Zooming the map in or out will cause a redraw that then shows the map with a correctly scaled red boarder

mo-nathan:
Sometime you just gotta love CC!  I'm now calling this Claudelight!

mo-nathan (approved):
After the bug fix from CC, manual testing looks great!

PR #3708, opened 2026-01-05
Obs form - fix EXIF transfer of date and location data

nimmolo:
### Issues 
1. When uploading geotagged images to the observation form, the EXIF date and location data wasn't being transferred to the observation fields. Users would see the current date (2026) instead of the photo's actual date, and GPS coordinates wouldn't populate.
2. The connection between the "New Locality" field and the map doesn't seem to be working. On production, if I enter "Harare, Zimbabwe" in the locality field, it gives me the option to "+ New locality". This also happens locally. However, if I click the "+ New locality", the map updates in production and not locally.

### Test
Manual. Described above.

### Technical

1. Fix map marker not being assigned in drawMarker()

 The this.marker reference was only assigned in the else branch of drawMarker(), so when the first condition was true (!this.editable && set != null), this.marker remained undefined. This caused downstream code relying on this.marker to fail silently, breaking EXIF date extraction from uploaded images.

  Moved this.marker = marker outside the conditional so it's always assigned after creating the marker.

  2. Outlet access in map_integration_mixin.js and base_controller.js - Added try/catch around mapOutlet accesses since the controller may not be connected when the map is collapsed

  3. updateHiddenId() in base_controller.js - Removed the matches.length > 0 condition so clearHiddenId() is called even when there are zero matches, allowing the "New Location" button to appear

### Super technical

  1. form-exif_controller.js transferExifToObsFields() lines 270-278:
  if (this.hasMapOutlet) {
    this.mapOutlet.calculateMarker(
      { detail: { request_params: { lat, lng } } }
    );
  }
  if (_exif_data.exif_date) {
    const _exifSimpleDate = JSON.parse(_exif_data.exif_date);
    this.observationDate(_exifSimpleDate);  // Date transfer happens HERE
  }

  2. map_controller.js calculateMarker() calls placeMarker(location) (line 512)
  3. map_controller.js placeMarker() lines 171-178:
  placeMarker(location) {
    if (!this.marker) {
      this.drawMarker(location)  // Creates marker but didn't always assign this.marker
    } else {
      this.marker.setPosition(location)
    }
    this.marker.setVisible(true)  // CRASHES if this.marker is undefined
  }

  4. When drawMarker didn't set this.marker, line 178 crashes trying to call .setVisible(true) on undefined
  5. The crash stops execution before reaching this.observationDate(_exifSimpleDate) - so the date transfer never happens
  6. The form keeps its default date (2026) instead of the EXIF date (2006)

### DIFF
--- a/app/javascript/controllers/autocompleter/base_controller.js
+++ b/app/javascript/controllers/autocompleter/base_controller.js
@@ -826,7 +826,7 @@ export default class BaseAutocompleterController extends Controller {
       if (this.lastHiddenTargetValue() != perfect_match['id']) {
         this.assignHiddenId(perfect_match);
       }
-    } else if (!this.ignoringTextInput() && this.matches.length > 0) {
+    } else if (!this.ignoringTextInput()) {
       this.clearHiddenId();
     }
   }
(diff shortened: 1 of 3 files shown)


JoeCohen (commented):
I'm not getting it. For me, this works fine in main. 
(When I create an Observation, and drag an image, the date is correct and the map pin appears when I show the map.)
Maybe it has been fixed elsewhere?

Image

PR #3722, opened 2026-01-08
Phlex `Table` component, `IpsManager` form

nimmolo:
### Turns blocked_ips UI into a Phlex `IpsManager` component.
Huge upgrade of UI: 
- Adds **pagination** for the IPs in the "Okay IPs" and "Blocked IPs" lists, page doesn't take forever to load. Pagination currently set at 50 IPs per page.
- Adds a **live filter** for both lists via the `starts_with` param. Type the beginning of an IP address to filter the list to that block of IPs. Buffered with a 1 second delay.
- Remove or Add IP: Lists update via Turbo,  no more reloading the page.
- To try the `IpsManager` with real data, get the live files from production.
`scp yourname@mushroomobserver.org:/var/web/mo/config/blocked_ips.txt config/blocked_ips.txt`
`scp yourname@mushroomobserver.org:/var/web/mo/config/okay_ips.txt config/okay_ips.txt`

Screen Shot 2026-01-08 at 9 53 34 PM
Screen Shot 2026-01-08 at 9 54 56 PM

### Creates Phlex `Table` component and uses it in admin/blocked_ips/edit 

 Updated 4 tables to use the Table component:

  1. edit.html.erb - okay_ips table (headerless)
  2. edit.html.erb - blocked_ips table (headerless)
  3. _ip_summary.html.erb - Most active users table (5 columns with headers)
  4. _ip_stats.html.erb - Activity stats table (3 columns with headers)
___
  Usage examples from the page:
```erb
  <%# With headers (default) %>
  <%= render(Components::Table.new(@stats, class: "table-striped")) do |t| %>
    <% t.column("ip") { |ip| link_to(ip, path) } %>
    <% t.column("load %") { |ip| @stats[ip][:load] } %>
  <% end %>

  <%# Without headers %>
  <%= render(Components::Table.new(@ips, headers: false, id: "my_table")) do |t| %>
    <% t.column("ip", &:t) %>
    <% t.column("actions") { |ip| patch_button(...) } %>
  <% end %>
```
Note the syntax of the Table columns is somewhat different from `slot` used elsewhere - this pattern is documented at [Phlex.fun](https://www.phlex.fun/components/yielding#vanishing-the-yield) and uses internal methods of the component to render the columns.

The difference is that a `slot` is for passing HTML straight into the component, whereas a `Table.column` is recursively computed from an initial `rows` arg that is the first argument of the table.


JoeCohen (commented):
1. Thanks  for doing this. A huge improvement.
2. I'll look at the code. Maybe something will sink into my head, but no need to wait on me.
3. Following is not a blocker, fix only if easy and you feel like it.
Focus (or something) disappears when backspacing over search strings.
To replicate:
  - Paste a complete ip address into Blocked IPs, Filter by IP Prefix.  (I used `101.47.45.243`)
  - wait a split second
  Result: cursor no longer visible
  - click at the end of that string to put cursor there
  - delete/backspace over last character
  - wait a split second
  Result: last character and cursor disappear
  - Rinse and repeat
  Result: ditto

app/components/ips_manager.rb
```diff
@@ -0,0 +1,187 @@
...
+    @type == :blocked ? "Blocked IPs:" : "Okay IPs:"
+  end
+
+  def render_showing_message
+    if filterable?
+      plain("Showing #{@ips.size} of #{@total_count}")
+      plain(" (page #{@page} of #{@total_pages})") if @total_pages > 1
+    else
+      plain("Showing #{@ips.size}")
+    end
```
JoeCohen: If you prefer, would this work?
```ruby
      plain("Showing #{@ips.size}")
      return unless filterable?

      plain(" of #{@total_count}")
      plain(" (page #{@page} of #{@total_pages})") if @total_pages > 1
```

PR #3725, opened 2026-01-09
Enable `simplecov` to run parallel tests with merged coverage results

nimmolo:
test/test_helper.rb:
  - Removed the restriction that disabled parallelization when coverage was enabled
  - Added `parallelize_setup` to give each worker a unique `SimpleCov.command_name`
  - Added `parallelize_teardown` to call `SimpleCov.result` for proper merging of coverage results into a single report

  config/environments/test.rb:
  - Changed `eager_load` from `false` to `ENV["CI"].present?` for accurate coverage in CI

Result: Tests now run in parallel (4 processes on my machine) while still collecting accurate, merged coverage data. CI should see a nice speedup.

### DIFF
--- a/.github/workflows/ci_rails.yml
+++ b/.github/workflows/ci_rails.yml
@@ -13,6 +13,8 @@ on:
 jobs:
   test:
     runs-on: ubuntu-22.04
+    env:
+      PARALLEL_WORKERS: 4
     steps:
       # check-out repo under $GITHUB_WORKSPACE, so that workflow can access it.
       # https://github.com/actions/checkout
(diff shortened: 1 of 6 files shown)


.github/workflows/ci_rails.yml
```diff
@@ -13,6 +13,8 @@ on:
 jobs:
   test:
     runs-on: ubuntu-22.04
+    env:
+      PARALLEL_WORKERS: 4
```
nimmolo: Our GitHub runner actually only gives us 2 workers, but this will work in case we get a faster runner in the future.

PR #3752, opened 2026-01-20
Add FieldSlip API calls

mo-nathan:
# Summary

This pull request adds comprehensive API support for FieldSlip resources, enabling CRUD operations through the API2 endpoint. The implementation follows established patterns in the codebase for model-based API endpoints.

## Changes

Added API endpoint for querying, creating, and updating field slips with various filter parameters
Implemented JSON and XML view templates for field slip responses
Added scopes to the FieldSlip model for filtering by code, observation, and project
Created comprehensive test coverage for all API operations

## Manual Tests
```
# Some examples to riff on
curl -H "Accept: application/json" "http://localhost:3000/api2/field_slips?code_has=NEMF&detail=low"
curl -H "Accept: application/json" "http://localhost:3000/api2/field_slips?id=123&detail=high"
curl "http://localhost:3000/api2/field_slips?project=389"

# Create a new field slip
curl -X POST "http://localhost:3000/api2/field_slips" \
  -d "api_key=YOUR_API_KEY" \
  -d "code=NEMF-8642" \
  -d "observation=12345"

# Update field slip code
curl -X PATCH "http://localhost:3000/api2/field_slips" \
  -d "api_key=YOUR_API_KEY" \
  -d "id=456" \
  -d "set_code=UPDATED-0001"
```

### DIFF
--- a/app/classes/api2/field_slip_api.rb
+++ b/app/classes/api2/field_slip_api.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+class API2
+  # API for FieldSlip
+  class FieldSlipAPI < ModelAPI
+    def model
+      FieldSlip
+    end
+
+    def high_detail_page_length
+      100
+    end
+
+    def low_detail_page_length
+      1000
+    end
+
+    def put_page_length
+      1000
+    end
+
+    def delete_page_length
+      1000
+    end
+
+    def high_detail_includes
+      [
+        :observation,
+        :project,
+        :user
+      ]
+    end
+
+    def query_params
+      {
+        id_in_set: parse_array(:field_slip, :id, as: :id),
+        created_at: parse_range(:time, :created_at),
+        updated_at: parse_range(:time, :updated_at),
+        by_users: parse_array(:user, :user, help: :creator),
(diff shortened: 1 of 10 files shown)


JoeCohen (commented):
Thanks for adding this needed functionality.
Code looks good.
I didn't manually test because I don't understand what the url should be.

mo-nathan:
@JoeCohen Look at the manual tests and run those from your command line when you have the server running.  You should be able to test the GETs by just going to the URL in the curl commands and ignore the -H bit.  That's for getting curl to return JSON which the browsers do by default.  Not sure how to test the XML version (curl's default) from a browser.  Hopefully you can get things working with curl.  The POST and PATCh examples needs to be run with curl.  They could also be run with wget, but I think the syntax is a bit different.  I'm sure you friendly neighborhood AI could switch it over the wget if ask.

JoeCohen:
I don't understand. I don't see any urls in `test/classes/api2/field_slips_test.rb`.
I think you should merge this, as I'm just holding things up. I also need to be AFK for a while.

JoeCohen:
The question is what most people who use this query would want to see without an additional API call.
You probably have a better idea than I.

PR #3781, opened 2026-01-27
Phlex: use new `mark_safe` for `register_output_helper`

nimmolo:
Per [this pull request on Phlex](https://github.com/yippee-fun/phlex-rails/pull/303), we don't have to use `trusted_html` as much. Available now since @JoeCohen bumped our version of Phlex.

We can basically now tell Phlex that a registered output helper already returns safe HTML.

This updates `Components::Base` to use this, and updates callers.

### DIFF
--- a/.claude/style_guide.md
+++ b/.claude/style_guide.md
@@ -93,6 +93,26 @@ end
 
 ## Phlex Component Style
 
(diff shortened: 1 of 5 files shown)


mo-nathan (approved):
Looks good!

PR #3804, opened 2026-02-03
End reliance on "Voucher Specimen Taken"

### LINKED ISSUE #3801: Stop setting observation.specimen = true based on Voucher Specimen Taken
See parent issue

JoeCohen:


test/jobs/inat_import_job_test.rb
```diff
@@ -434,7 +434,9 @@ def test_import_job_nemf_plischke
     standard_assertions(obs: obs, name: name)
 
     assert(obs.sequences.one?, "Obs should have one Sequence")
-    assert(obs.specimen, "Obs should show that a Specimen is available")
```
mo-nathan: Wouldn't it be better to check that it's not set now since that's the intention of the ticket?  Seems like this test would still pass on main without the change.

PR #3816, opened 2026-02-04
Update consensus algorithm to boost sub-max agreeing votes

### LINKED ISSUE #3815: Update Name Consensus Algorithm
The Problem

The current name consensus calculation penalizes name votes that are below the value "I'd Call It That".  However, the most common usage of the vote "Promising" is more an indication of the user's lack of confidence in the ID vs. saying they think there is something wrong with that naming.  If the user has another vote for some other name with the value "I'd Call It That", then the current algorithm is reasonable.  However, if a user's highest vote is either "Promising" or "Could Be", then then names with those votes should only get added weight rather than being penalized.

Example,

Currently, if user nathan creates a new observation for Amanita amerirubescens and votes "I'd Call It That", then the consensus weight is 86%.
If user darv then votes "Promising", then the consensus weight drops to 78%.

However, instead it should increase.

In the above case, the positive increase should be prorated based on the vote being "Promising", but it should essentially act as a diluted (66%) additional vote at the "I'd Call It That" level.  Specifically, a "Promising" vote would contribute 3.0 * 0.66 * w to sum_val and 0.66 * w to sum_wgt.  Similarly, if darv only voted "Could Be" and that was their highest vote across all names, that should behave the same but at an even lower weight (33%).  "No Opinion" should have no effect.  If the user's maximum value is negative, then the current weighting system should be used.

However, if after voting "Promising" for Amanita amerirubescens, darv then votes some higher value (say "I'd Call It That") for some other name (say Amanita rubescens), then the consensus weight of Amanita amerirubescens should use the current algorithm and the consensus weight for that name should drop to 78% per the example.  The would presumably cause Amanita rubescens to now "win" since it would have a higher consensus weight (the exact value depends on the contribution of the different users).

We will need to figure out how this algorithm change affects the consensus names of all existing observations.  Depending on how often it will change the consensus name, we may want to address this effect by adding a set of additional votes to maintain the current consensus.  However, I need to see some examples of such changes first and know how many observations will be affected.

 Understanding of the Current Algorithm                                                                                                                           

The consensus for a naming is calculated as sum(value * weight) / (sum(weight) + 1.0) in app/models/observation/consensus_calculator.rb:62. When Nathan votes
  "I'd Call It That" (3.0) and Darv votes "Promising" (2.0), therefore Darv's lower vote value mathematically pulls the weighted average down (a "penalty").

Note if a user votes "Promising" on Name A and "Could Be" on Name B, with no "I'd Call It That" anywhere, the "Promising" votes acts as a boost to Name A, but the regular algorithm should apply to Name B so this should be a "penalty".  If they want both to be boosts, then they should give each name the high value they vote on those names.

mo-nathan:
## Summary

- Updates the consensus algorithm so that sub-maximum positive votes (e.g. "Promising") boost the consensus when they represent agreement, rather than paradoxically lowering it
- Adds a Rails runner script (`script/analyze_consensus_change.rb`) for analyzing impact and applying the update to all observations
- Adds comprehensive documentation of the algorithm in `doc/consensus_algorithm.md`
- Adds 6 tests covering boost behavior, edge cases, and negative vote handling

## Background

The original algorithm assumed most users would vote "I'd Call It That" (3.0) when confident. In practice, experienced identifiers routinely vote "Promising" (2.0) to express appropriate humility and allow easier overrides. Under the old algorithm, a second "Promising" vote on a naming that already had an "I'd Call It That" vote would *lower* the consensus score, penalizing agreement.

## How it works

When a user's highest vote across all namings is positive but below 3.0, and the naming already has a higher vote from someone else, the vote is treated as agreement at the naming's maximum level with proportionally reduced weight. The vote value is conceptually elevated to `naming_max` and the weight is reduced to `(vote_value / naming_max) * weight`. This boosts the consensus while maintaining the invariant that the score can never reach the maximum vote value.

The boost does **not** apply when:
- The user has voted "I'd Call It That" on a different naming (their sub-max vote genuinely represents less confidence)
- No one has voted higher on this naming (no stronger opinion to agree with)
- The user's maximum vote is negative

## Analysis script

```bash
# Dry-run analysis (no database changes)
rails runner script/analyze_consensus_change.rb --limit 400

# Apply update to all observations
rails runner script/analyze_consensus_change.rb --update
```

Sample run on 400 candidates: 4 consensus changes out of 132,456 total candidates (~1% rate).

## Test plan

- [x] 6 new tests covering boost/no-boost scenarios all pass
- [x] Existing observation tests unaffected
- [x] RuboCop clean on all modified files
- [x] Analysis script verified in both `--limit` and `--update` modes
- [x] Run full test suite in CI

Fixes #3815


JoeCohen (approved):
The rationale makes sense. It's all well-explained and documented.
I'm concerned about what will happen when the script is run.
Sorry I didn't get a count, but it looked to me like 4-5% of Observations will have consensus changes. 
Although perhaps a small percentage, it's large in terms of absolute numbers.

- Could this try to generate so many emails that Gmail will not send/deliver many?
Maybe the backfilling could be done over time, in small batches

- Probably more important. How will users react when old observations change their ID? (I expect users will complain (or at least ask questions).
Is there a plan to deal with that? I'd rather not spend time responding to emails.
It's instinctively surprising to see some of those changes, particularly from a provisional name supported by DNA to a registered name.

PR #3851, opened 2026-02-16
Make the "admin donations review form" more Superform-idiomatic

nimmolo:
Superform 
- implicitly creates a `form` element, via `super do`, in `view_template`
- has helper methods, customized for MO, for any input type, e.g. `checkbox_field`

So when you see Claude writing a `form_tag` method using Phlex `form`, or using Phlex `input`, you know it's not using Superform and it didn't do its homework. Our Superform patterns for Phlex conversions are detailed in our Claude style guide. Just be sure to ask it to read that, and to familiarize itself with existing form components, before starting a Phlex form conversion.

The first version of `DonationsReviewForm` did a weird thing. It inherited from `ApplicationForm` (Superform) but bypassed it entirely — overriding `form_tag`, rendering manual `<input>` elements, and skipping `super` in `view_template`.

This PR's Changes:
- Use `super do...end` in view_template so Superform handles the form tag, CSRF token, and _method field automatically
- Replace manual hidden+checkbox input pairs with `checkbox_field`
- Remove `form_tag`/`form_attributes` overrides; pass id/style through the constructor instead
- Set `FormObject` model_name to "Reviewed" so Superform namespaces fields as `reviewed[donation_id]`, matching controller expectations
- Remove unused text-center wrapper `div` from table
- Add component test for `DonationsReviewForm`


mo-nathan (approved):
I guess this is fine.  Reading it over I see some of the motivations for the changes, but I don't fully get them.  In the future, please just give me written feedback on my code and let me fix it myself.  It is a better way for me to learn and not make the same mistakes in the future.

PR #3938, opened 2026-02-17
Convert Translation Edit form to Phlex

### LINKED ISSUE #3887: Convert: Translation edit
Convert `app/views/controllers/translations/_form.erb` to a Phlex component
## Tasks

- [x] Review Claude draft and suggest changes
- [x] Ensure no coverage decrease
- [x] Manual tests
- [x] Try to remove some partials. See https://github.com/MushroomObserver/mushroom-observer/pull/3938#pullrequestreview-3835496103

JoeCohen:
Resolves #3887

### Authorization
- [x] Visit `/translations` without logging in → redirected to login
- [x] Log in as a user with no observations → flash error, redirected
- [x] Log in as a non-reviewer and visit `/translations?locale=en` → flash error (reviewer required for official language), redirected
- [x] Log in as a non-reviewer and visit `/translations?locale=el` → success, index renders
- [x] Log in as a reviewer and visit `/translations?locale=en` → success, index renders
- [x] Visit `/translations?locale=bad` → flash error, redirected

### Index
- [x] Index shows translation tags grouped with headers and comments
- [x] Tags missing from the export file appear under an **UNLISTED STRINGS** section at the bottom
In the rails console:
```ruby
lang = Language.find_by(locale: "en")
lang.translation_strings.create!(tag: "my_orphan_tag", text: "Some value")
```

Visit /translations?locale=en and scroll to the bottom. 
To clean up afterward:
```ruby
lang.translation_strings.find_by(tag: "my_orphan_tag").destroy
```

- [x] `?for_page=<token>` filters to only the tags used on that page
- [x] `?for_page=xxx` (expired token) shows flash error and displays all tags

### Edit (turbo_stream)
- [x] Clicking edit on a tag replaces the `#translation_ui` frame with the edit form — no full page reload
- [x] Editing tag with variants shows field for all variants. E.g., `external_link` shows fields for all variants: `external_link`, `external links`, `External Link`, `EXTERNAL_LINKS`
- [x] Edit form for an unofficial language shows the official English string alongside the editable field
- [x] Previous version history appears below the form when versions exist

### Update (turbo_stream)
- [x] Saving a changed value updates the in-page preview string for that tag without a full reload
- [x] Submitting the same value (no change) still succeeds and updates the timestamp. View the timestamp in the rails console:
```ruby
TranslationString.find_by(tag: "one", language: Language.find_by(locale: "en")).updated_at
```

- [x] Saving a tag that has no translation yet creates it (visible on re-edit)
Open a tag in an unofficial language that has no existing translation, type something, save it, then re-open the edit form — the value you typed should now appear in the textarea rather than being empty (or showing the English fallback).


public/test_server1-2/320/1062212448.jpg
mo-nathan: I don't think any of the files in `public` should be part of this PR.  Not sure how these got here, but according to CC:
```
These are test artifacts from running tests with parallel workers. The .gitignore in public/ 
ignores test_server1 and test_server2, but the parallel test infrastructure appends a worker suffix (e.g.,
test_server1-2, test_server2-2), which doesn't match the gitignore patterns. So these leaked into the commit.       
                                                                                                                      
The fix would be to update public/.gitignore to use wildcard patterns like test_server1* and test_server2*. You could leave that as a comment on the PR, or I can fix it on the other branch if you'd like.
```

mo-nathan (dismissed):
Two things:
1. I don't understand why there are files in `public` in this PR.  I suspect that was an accidental things.
2. Seems like this should allow us to delete some of the partials from `app/views/controllers/translations`.  I've noticed that CC doesn't seem to like to do this even though that's one of the major reasons for doing these PRs.

JoeCohen:
Thanks for looking at this. 
- I didn't completely understand the CC comment. I think this branch should get rid its new public files.
Will it work to revise .gitignore and delete the files?

JoeCohen:
I'll look at getting rid of some partials.

mo-nathan:
Yes, I think updating the .gitignore and removing the files is the right thing to do.  I'm not 100% confident about why these showed up.  My only working theory is that they normally get generated and cleaned up when the tests run and that you interrupt a test run at just the right moment to leave this cruft around.  However, there's a small flaw in that theory since the files don't get cleaned up if you just run the tests.  Maybe running the tests generate different files each time?  Maybe they vary each run based on which workers happen to run whatever test it is that is generating these?  I'd feel more confident if I fully understood which tests generated these files and why they don't get cleaned up automatically.  There is also an option of update the tests so they always clean up files like these when the tests stop running.  Certainly the simplest thing to do is remove these from the PR, update the .gitnore and move on.

mo-nathan (approved):

PR #3943, opened 2026-02-20
Enable raise_delivery_errors and log email failures

### LINKED ISSUE #3942: Turn on raise_delivery_errors
Setting `raise_delivery_errors = false` in app/mailers/application_mailer.rb caused us to miss errors in our user verification system for over a month.  Let's see if we can remove this and how bad the side effects actually are.

Adding error handling that logs failures to email-debug.log also seems like a good thing to do at least at first.

mo-nathan:
This is a followup to the issue where verification emails were failing silently.  @pellaea left a comment about why the errors were disabled, but at this point that approach doesn't seem like the right approach to me.  I would first like to see how commonly issues arise.  If there are common patterns, then I think it is better to address those patterns directly rather than turning off alerts for any errors.

## Summary
- Remove `ActionMailer::Base.raise_delivery_errors = false` from `ApplicationMailer`, allowing the production config's `raise_delivery_errors = true` to take effect so email delivery problems are detected promptly
- Add `config/initializers/mail_delivery_error_logging.rb` that patches `ActionMailer::MailDeliveryJob` with a `rescue_from` handler to log delivery failures to `email-debug.log` before re-raising (so Solid Queue also tracks the failure)
- Guard `after_action` delivery callbacks (`webmaster_delivery`, `news_delivery`) with `message.to.blank?` to prevent errors when `mo_mail` returns early for invalid addresses or opted-out users
- Remove dead `noreply_delivery` method (never referenced as an `after_action` by any mailer)

## Test plan
- [x] All 25 existing mailer tests pass, including `test_undeliverable_email` and `test_opt_out`
- [x] 3 new tests in `test/jobs/application_job_test.rb`:
  - Mail delivery failure is logged to `email-debug.log` with mailer name and error details
  - Non-mail job failures are NOT logged to `email-debug.log`
  - Mail delivery failure re-raises the original exception
- [ ] Deploy and monitor `email-debug.log` for any unexpected delivery errors

Fixes #3942

### DIFF
--- a/config/initializers/mail_delivery_error_logging.rb
+++ b/config/initializers/mail_delivery_error_logging.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Log email delivery failures to email-debug.log, then re-raise
+# so Solid Queue can track the failure. This ensures we notice
+# delivery errors instead of silently swallowing them.
+Rails.application.config.after_initialize do
+  ActionMailer::MailDeliveryJob.rescue_from(
+    StandardError
+  ) do |exception|
+    mailer, action = arguments[0..1]
+    args = arguments[3..] || []
+    msg = "DELIVERY FAILED #{mailer}##{action}"
+    msg << " args=#{args.inspect}" if args.any?
+    msg << " error=#{exception.class}: " \
+           "#{exception.message}"
+    Rails.root.join("log/email-debug.log").
+      open("a:utf-8") do |fh|
+        fh.puts("#{Time.zone.now} #{msg}")
+      end
+    raise(exception)
+  end
+end
--- a/config/initializers/mail_delivery_error_logging.rb
+++ b/config/initializers/mail_delivery_error_logging.rb
@@ -8,9 +8,9 @@ Rails.application.config.after_initialize do
     StandardError
   ) do |exception|
     mailer, action = arguments[0..1]
-    args = arguments[3..] || []
+    params = arguments[3] || {}
     msg = "DELIVERY FAILED #{mailer}##{action}"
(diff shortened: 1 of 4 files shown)


config/initializers/mail_delivery_error_logging.rb
```diff
@@ -0,0 +1,22 @@
...
+  ) do |exception|
+    mailer, action = arguments[0..1]
+    args = arguments[3..] || []
+    msg = "DELIVERY FAILED #{mailer}##{action}"
+    msg << " args=#{args.inspect}" if args.any?
+    msg << " error=#{exception.class}: " \
+           "#{exception.message}"
+    Rails.root.join("log/email-debug.log").
+      open("a:utf-8") do |fh|
+        fh.puts("#{Time.zone.now} #{msg}")
```
JoeCohen: Does this print the time in Eastern? If so, is UTC more useful?

JoeCohen (approved):
Thanks for doing this. Looks good to me (but I'm not great at understanding log creation).

config/initializers/mail_delivery_error_logging.rb
mo-nathan: Interesting point and maybe this is something we should clean up.  However, the current email-debug.log uses different time zones:
```
2026-02-21 01:26:15 -0800 MAIL user_question to=200146 email=punkie_2007@yahoo.com
2026-02-21 06:05:55 -0400 MAIL location_change from=55281 to=1290 location=17615 description=
2026-02-21 05:11:00 -0500 MAIL comment from=110873 to=667 object=observation-581232
```
presumably based on the user that is generating the email.

PR #3975, opened 2026-03-04
Add missing Name at supported rank

### LINKED ISSUE #3974: Add missing supported Names
Importer should add iNat names to MO if the name is at a rank supported by MO.

- [x] Add Name for unmatched species
NOTE: iNat's new form of provisional name has the rank "species", rather than a separate rank as in the original (now-modified) announcement. Instead, taxa now have a `provisional` attribute.
- [x] Add Name for unmatched infrageneric
- [x] Add Name for monomial, e.g. genus
- [x] Add Name for complex
- [x] Revise InatImportsJobTest to not create missing Names in advance of the import.
- [x] Separate MO Name creation from InatObs.name. Seems like the latter should not change the db, but instead return `nil` if there's no MO equivalent. Then something in the importer should create the MO equivalent, else use `Fungi`.
- [x] DRY some stuff. Since we're now generally creating missing names, that's how to create provisionals.  Cf. `app/classes/inat/mo_observation_builder.rb#add_provisional_name`, `app/classes/inat/taxon.rb#create_mo_name`
- [x] Create separate issu to ensure created Names are logged. See #4002 `test/jobs/inat_import_job_test.rb#test_import_job_prov_name_pnw_style` creates `Donadinia sp. 'PNW01'` and `Donadinia` but fails to log the latter.

JoeCohen:
Resolves #3974

### Suggested Manual Test

For each iNat rank below:
 
 - Find an iNat fungal taxon which is missing from MO.
   - Get recent iNat fungal taxa at that rank. Example API request https://api.inaturalist.org/v1/taxa?order=desc&order_by=created_at&taxon_id=47170&rank=species
   - In those results look for one taxon name which is not in MO. 
     - To view the names, search the API results for `"name":`
     - Search MO for that name.
 - Create an iNat observation of that taxon (or modify a test iNat observation to be that taxon). 
 - Import it.
   - Expected result: It imports with that name, and the Name is added to MO at the correct rank
 - **CLEAN UP** Refresh the iNat obs, and Delete the `Mushroom_Observer_URL` Observation Field

iNat ranks:
- [x] "phylum"
- [x] "class"
- [x] "order"
- [x] "family"
- [x] "genus"
- [x] "subgenus"
- [x] "section"
- [x] "subsection"
- [x] "complex"
- [x] "species"
- [x] "subspecies"
- [x] "variety"
- [x] "form"


mo-nathan (approved):
Code looks good.  I did not do any manual testing because I don't feel completely comfortable with how to clean everything up on the iNat side.

PR #3990, opened 2026-03-07
Fix location.rb metrics offenses

JoeCohen:
Fixes location.rb metrics offenses and one style offense


mo-nathan (approved):
Looks good.  One minor non-blocking comment about the use of private and public.  Thanks for cleaning this up!

PR #3996, opened 2026-03-08
Fix thumbnail not reassigned when removed via edit form

### LINKED ISSUE #3995: Hi-res image substituted for imported image does not display in index
See this report from @AlanRockefeller 

> image.png [7:40 AM]Looks like there is a bug that is causing the photo to not be displayed on the main page.
> 
> What I did was import an iNat observation, then add the high resolution image, then remove the image that was imported from iNat.
> 
> https://mushroomobserver.org/obs/628987

mo-nathan:
## Summary
- Fixes #3995 — when the thumbnail image is removed via the observation edit form, `thumb_image_id` was left `nil` even when other images remained
- Root cause: the JS clears `thumb_image_id` to `""`, and `update_permitted_observation_attributes` mass-assigns this as `nil` *before* `detach_removed_images` runs, so `remove_image` never detects the thumbnail needs reassignment
- Adds `ensure_thumb_image` after image detachment to reassign the thumbnail to the first remaining image when the current value is missing or invalid

## Test plan
- [x] New regression test: `test_update_observation_remove_thumbnail_reassigns`
- [x] All 14 existing observation update controller tests pass (243 assertions)
- [x] Manual: create observation with 2 images, remove the thumbnail via edit form, verify remaining image becomes thumbnail

### DIFF
--- a/app/controllers/observations_controller/edit_and_update.rb
+++ b/app/controllers/observations_controller/edit_and_update.rb
@@ -130,6 +130,17 @@ module ObservationsController::EditAndUpdate
       img.log_remove_from(@observation)
       flash_notice(:runtime_image_remove_success.t(id: img.id))
     end
+    ensure_thumb_image
+  end
+
+  # Fix for issue #3995: update_permitted_observation_attributes runs before
+  # detach_removed_images, so the form's blank thumb_image_id overwrites the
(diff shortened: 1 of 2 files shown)


JoeCohen (approved):
Thank you! Code looks fine and it works great.

Issue #3846, opened 2026-02-14
Review and enhance log entries


mo-nathan:
Observations and Names both get logs, but they don't seem very reliable or complete.  It would be worth doing a review and audit of the existing logs to see how reliable they are and create a set of follow up tickets to improve these logs.

PR #4007, opened 2026-03-09
Log creation of a Name via the API

### LINKED ISSUE #4002: Log creation of Name by API
The API does not create an rss_log when it creates a Name. (On the other hand creation of Observations is logged via an after_create hook and a parsed @log boolean parameter).

The lack of an initial `rss_log` doesn't seem to break anything, but leads to some undesirable or unexpected results:

- a User who watches the log is left in the dark
- if a log later gets created (via editing the Name) the Name's log and versions are incongruent.  

NOTE: Many Names lack a log:
```ruby
mushroom-observer(dev)> Name.where(rss_log: nil).count
  ...
=> 12340
```

JoeCohen:
Fixes #4002

### DIFF
--- a/app/classes/api2/name_api.rb
+++ b/app/classes/api2/name_api.rb
@@ -77,7 +77,12 @@ class API2
       make_sure_name_doesnt_exist!(parse)
       name = create_name(parse, params)
       save_parents(parse)
+      after_create(name)
       name
+    end
+
+    def after_create(name)
+      name.user_log(user, :log_name_created) if @log
     end
 
     def build_setter(params)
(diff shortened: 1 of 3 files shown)


mo-nathan:
The logging stuff needs be re-written.  This is the goal of the roadmap item #3846 (14: Review and enhance log entries).  Logging is actually significantly harder that I realized when I first implemented this feature.  I think I originally thought most of the logging could be done as callbacks.  I'm not at all convinced of that.  I expect Claude (Code or AI) could help design a much better system than we have at the moment.  My belief at this point is that it would be better to do all logging in the controllers rather than in models or POROs like this where logging is not the primary concern.  Most of the systems I've worked on do logging badly including MO.  I expect there are much better patterns that really do logging right, but we need to understand our real requirements.  In the meantime I don't really care how we bandaid what I suspect is a deeply flawed system.

mo-nathan (approved):
This seems like a fine short term fix until we go after rewriting logging in MO.

PR #4027, opened 2026-03-12
Fix flaky license create test

### LINKED ISSUE #3998: Flaky test: LicensesControllerTest#test_create
## Problem

`LicensesControllerTest#test_create` fails intermittently in parallel test runs.

The test creates a new license and then retrieves it using:

```ruby
license = License.order(updated_at: :asc).last
```

This assumes the newly created license will have the latest `updated_at` timestamp. In parallel test runs, another process may touch a license record at the same time, causing a different license to sort last.

## Failure output

```
FAIL LicensesControllerTest#test_create (2.69s)
    --- expected
    +++ actual
    @@ -1 +1 @@
    -"Creative Commons Non-commercial v4.0"
    +"Creative Commons Wikipedia Compatible v3.0"
    test/controllers/licenses_controller_test.rb:147:in `test_create'
```

## Suggested fix

Instead of relying on timestamp ordering, find the license by its known attributes:

```ruby
license = License.find_by(display_name: display_name)
```

## Location

`test/controllers/licenses_controller_test.rb:146`

JoeCohen:
Fixes #3998

### DIFF
--- a/test/controllers/licenses_controller_test.rb
+++ b/test/controllers/licenses_controller_test.rb
@@ -143,6 +143,10 @@ class LicensesControllerTest < FunctionalTestCase
     end
 
     license = License.find_by(display_name: display_name)
+    assert_not_nil(
+      license,
+      "Expected License to be created with display_name #{display_name.inspect}"
+    )
     assert_equal(display_name, license.display_name)
     assert_equal(url, license.url)
     assert_false(license.deprecated)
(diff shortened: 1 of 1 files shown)


mo-nathan (approved):
Looks good.  I wouldn't be surprised if this anti-pattern exists in other tests.  Seems worth asking Claude to try to hunt them down.

PR #4135, opened 2026-04-16
Limit Add My Observations to 100 per click with count preview

### LINKED ISSUE #4129: Limit Add My Observations
If a Project has no constraints, then "Add My Observations" adds all of the user's Observations to the Project, regardless of date, location/gps, or consensus ID.
“Add My Observations” should probably use both target sets as constraints and target set violations should be considered constraint violations.  If there are no constraints, then “Add My Observations” should probably be hidden.
At minimum it should give you a warning of how many observations it is considering adding (and it should do it as a background job).

mo-nathan:
## Summary
- Clicking "Add My Observations" now opens a Turbo modal showing how many of the member's observations match the project's constraints and aren't already in the project.
- Each click adds at most 100 observations (most-recent first), keeping the request synchronous. Modal button reads "Add All" when count ≤ 100, "Add Next 100" when over.
- Existing location/date filters and the `Project#add_observation` dedup guard are preserved.

Fixes #4129.

Scope limited to the count-preview + cap per Nathan's comments on the issue. Target names/locations and consensus-ID as constraints are deferred to follow-up tickets.

## Test plan
- [ ] Join a project with no constraints and few observations → modal shows "Add All", single click adds them all.
- [ ] Join a constrained project (e.g. one with a location or date range) → modal count reflects only matching obs.
- [ ] With >100 matching obs, click twice → each click adds 100, modal count decreases, flash reports actual count added.
- [ ] Click "Add My Observations" a second time after adding → modal shows "None of your observations match...".
- [ ] Modal renders correctly on mobile / small viewport.
- [x] `bin/rails test test/controllers/projects/members_controller_test.rb` — 36 tests, 303 assertions, 0 failures.
- [x] `bin/rails test test/components/add_obs_modal_test.rb` — 3 tests, 18 assertions, 0 failures.
- [x] `bin/rails test test/controllers/projects/` — 111 tests, 706 assertions, 0 failures.
- [x] RuboCop clean on all changed files.


app/controllers/projects/members_controller.rb
```diff
@@ -169,13 +198,25 @@ def update_trust_status(project, candidate)
...
+      before = project.observations.count
+      project.add_observations(obs)
+      project.observations.count - before
+    end
+
+    # Observations owned by candidate that match project constraints and
+    # are not already in project.observations.
+    #
+    # Can't use candidate.observations due to a bug in in_box.
+    # Specifially, candidate.observations.in_box doesn't return
```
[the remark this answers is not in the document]
mo-nathan: Fixed in b826561d8.

app/controllers/projects/members_controller.rb
```diff
@@ -185,10 +226,7 @@ def add_observations(project, candidate)
         obs = obs.found_between(project.start_date.strftime("%Y-%m,-%d"),
                                 project.end_date.strftime("%Y-%m,-%d"))
```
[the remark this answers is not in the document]
mo-nathan: Good catch. The stray comma made `DateRangeParser.new(str).range` return nil, silently ignoring the project's date constraint. Fixed in b826561d8 to use `"%Y-%m-%d"`. (The bug was pre-existing but clearly wrong.)

test/controllers/projects/members_controller_test.rb
```diff
@@ -163,6 +163,113 @@
...
+      project = projects(:eol_project)
+      params = {
+        project_id: project.id,
+        candidate: target_user.id,
+        commit: :change_member_add_obs.l
+      }
+      put_requires_login(:update, params, target_user.login)
+      expected = :change_member_add_obs_flash.
+                 t(count: target_user.observations.count)
+      assert_flash_text(expected.to_s.gsub(/<[^>]*>/, ""))
```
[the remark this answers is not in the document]
mo-nathan: Fixed in b826561d8. Replaced the ad-hoc tag-stripping `gsub` with a `Regexp` passed to `assert_flash_text` so the test no longer attempts to sanitize HTML.

JoeCohen:
No modal or warning for unconstrained, nonmembership projects.
Maybe this happens because I am a site Admin?
Example:
- go to [Oak-associated Fungi of Washington State](http://localhost:3000/projects/395)
- Add My Observations (Not sure why this button is visible for this project)
- Result: 
  - No modal, No warning
  - Adds all my Observations whose GPS is visible.
  - Add My Observations button disappears.

JoeCohen:
Different subject
Project member page lists Site Admins as "Admin" even when their status is Member.
- [FunDiS Southeast Rare Fungi Challenge](http://localhost:3000/projects/331)
- Add myself as user.
- Result:
- members page lists my status as Admin 
- members/edit lists me as a Member, but not as a Project Admin

app/controllers/projects/members_controller.rb
```diff
@@ -169,26 +198,35 @@ def update_trust_status(project, candidate)
...
+      project.observations.count - before
+    end
+
+    # Observations owned by candidate that match project constraints and
+    # are not already in project.observations.
+    #
+    # Can't use candidate.observations due to a bug in in_box.
+    # Specifically, candidate.observations.in_box doesn't return
+    # the right thing because it incorrectly adds observations not
+    # from the candidate if they have no lat/long data.
```
JoeCohen, line 217: If it's really a bug, seems like his PR is a good opportunity to fix it.
If not a bug, tweak the comment.

app/controllers/projects/members_controller.rb
```diff
@@ -169,26 +198,35 @@ def update_trust_status(project, candidate)
...
       loc = project.location
       if loc
         obs = obs.in_box(north: loc.north, south: loc.south,
                          east: loc.east, west: loc.west)
       end
       if project.start_date && project.end_date
-        obs = obs.found_between(project.start_date.strftime("%Y-%m,-%d"),
-                                project.end_date.strftime("%Y-%m,-%d"))
+        obs = obs.found_between(project.start_date.strftime("%Y-%m-%d"),
+                                project.end_date.strftime("%Y-%m-%d"))
```
JoeCohen, line 227: Seems like it should handle start & end date separately.
Else if there's only one of those dates it will get ignored.

JoeCohen:
For site admins, neither modal nor warning when I added myself to [FunDiS Southeast Rare Fungi Challenge](http://localhost:3000/projects/331). It just added all my observations.

JoeCohen:
This branch simply isn't working for me, even when logged in as someone else. 
When I click Add My Obsevations, it simply adds them without modal or warning.
Not sure what I'm doing wrong.

mo-nathan:
> No modal or warning for unconstrained, nonmembership projects. Maybe this happens because I am a site Admin? Example:
> 
> * go to [Oak-associated Fungi of Washington State](http://localhost:3000/projects/395)
> * Add My Observations (Not sure why this button is visible for this project)
> * Result:
>   
>   * No modal, No warning
>   * Adds all my Observations whose GPS is visible.
>   * Add My Observations button disappears.

@JoeCohen Not what I'm seeing.  I get a modal:
Screenshot 2026-04-19 at 5 39 00 AM

And when I click "Add Next 100" it added 94 observations (71 -> 165).

Can you try reproducing after shutting your local server down and making sure you have the latest changes on the branch?

mo-nathan:
> For site admins, neither modal nor warning when I added myself to [FunDiS Southeast Rare Fungi Challenge](http://localhost:3000/projects/331). It just added all my observations.

Again this is working for me.

mo-nathan:
Here's what CC recommend you try to make sure your system is up to date:
```
  Refresh the local stack:
  git fetch origin njw-4129-limit-add-my-observations                                                                                               
  git checkout njw-4129-limit-add-my-observations                                                                                                   
  git reset --hard origin/njw-4129-limit-add-my-observations   # only if he has no uncommitted work                                                 
  bin/rails tmp:clear                                                                                                                               
  bin/rails assets:clobber                                                                                                                          
  # restart rails server                                               
  That kills Bootsnap cache, Rails tmp cache, and any compiled assets. Most "old code keeps running" bugs in Rails dev come from one of those three.
                                                                       
  Browser side:                                                                                                                                     
  - Hard refresh: Cmd+Shift+R (Mac) / Ctrl+Shift+R (Win).              
  - Or simpler: try the flow in an incognito window, which bypasses all browser caches and service workers.                                         
  - DevTools → Network tab → check "Disable cache" and keep DevTools open while clicking.                  
                                                                                                                                                    
  One-shot verification Joe can paste into the browser Console (same snippet I ran for you — confirms the button is the new version and the endpoint
   returns the modal):                                                                                                                              
  (async () => {                                                                                                                                    
    const link = document.querySelector('a[data-controller="modal-toggle"][data-modal="modal_add_obs"]');                                           
    console.log("button href:", link?.href);                                                             
    if (!link) return console.log("OLD button — not on the PR branch, or assets are stale");                                                        
    const resp = await fetch(link.href, {                                                   
      headers: { "Accept": "text/vnd.turbo-stream.html" },                                                                                          
      credentials: "same-origin"                                       
    });                                                                                                                                             
    const text = await resp.text();                                    
    console.log("status:", resp.status);
    console.log("has modal_add_obs:", text.includes("modal_add_obs"));                                                                              
  })();                                                               
  If button href is null, his HTML is still the old put_button form and the cache/restart is the whole story. If it prints a URL and has            
  modal_add_obs: true, everything is correct and the issue is in whatever Joe does next — watch the Network tab on the next click.       
                                                                                                                                                    
  Sanity-test the code is sound, not just his env:                     
  bin/rails test test/controllers/projects/members_controller_test.rb \                                                                             
                test/components/add_obs_modal_test.rb                  
  Those should all pass locally. If they do, his backend is correct and any remaining issue is front-end / browser-specific.                        
                                                                                                                            
  If after all that it still fails, the next signal we'd want from Joe is the Network tab capture of a single button click (request URL + method +  
  status) and any Console errors — that tells us whether the GET/turbo-stream round-trip is firing at all.                                          
```

app/controllers/projects/members_controller.rb
mo-nathan: Is it still possible to configure a project with only one of those dates?  I don't see a way to do it in the UI.  Looking at the database:
```
mysql> select count(*) from projects where start_date is not null and end_date is null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

mysql> select count(*) from projects where start_date is null and end_date is not null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)

mysql> select count(*) from projects where start_date is null and end_date is null;
+----------+
| count(*) |
+----------+
|      224 |
+----------+
1 row in set (0.00 sec)

mysql> select count(*) from projects where start_date is not null and end_date is not null;
+----------+
| count(*) |
+----------+
|       11 |
+----------+
1 row in set (0.00 sec)

```

mo-nathan: Good catch — fixed in 3a7aced41. Each bound is now applied separately via `found_after` / `found_before`, so a project with only a `start_date` or only an `end_date` correctly constrains the modal count and the subsequent add. Added two regression tests using the `no_start_date_project` and `no_end_date_project` fixtures.

mo-nathan: CC implemented a quick fix, so I added it.  Still not sure it's needed.

app/controllers/projects/members_controller.rb
mo-nathan: Checked empirically — the claim in the comment no longer holds on current code. `candidate.observations.in_box(...)` and `Observation.where(user: candidate).in_box(...)` produce identical SQL and identical results; the `in_box` scope preserves the associations `user_id` filter across its `.or()` branches. Verified on user Evica (38k obs) against a tight box — zero non-candidate observations leak in.

The comment was introduced in May 2025 (`a1335b491`), a few months after the `in_box` scope was reworked in Feb 2025 (`fb9a21232` / `d84632600`), so the bug Nathan observed was real at the time. Whatever reintroduced or surfaced the issue has since been fixed.

Per your suggestion, took the second option — simplified the code to use `candidate.observations` and dropped the comment. ff0a05a1b.

mo-nathan: Apparently no longer a bug.  Did some clean up.

mo-nathan:
> Different subject Project member page lists Site Admins as "Admin" even when their status is Member.
> 
> * [FunDiS Southeast Rare Fungi Challenge](http://localhost:3000/projects/331)
> * Add myself as user.
> * Result:
> * members page lists my status as Admin
> * members/edit lists me as a Member, but not as a Project Admin

Outside the scope of this PR.  Created #4145 to track this issue.  Also discusses what Site Admins should be able to see.

JoeCohen:
@mo-nathan: http://localhost:3000/projects/259 consistently throws an error for me.
I've updated the db, and tried this in different browsers and in incognito mode.
```ruby 
NoMethodError in ProjectsController#show
undefined method '<=' for nil
Extracted source (around line #245):
243
244
245
246
247
248
              

    s_results = results.select do |result|
      Name.ranks[result[:rank]] <= Name.ranks[:Species]
    end

    # This could include groups etc, so we just want to store the genus names.

app/classes/checklist.rb:245:in 'block in Checklist#count_taxa_genera_and_species'
app/classes/checklist.rb:244:in 'Checklist#count_taxa_genera_and_species'
app/classes/checklist.rb:226:in 'Checklist#calc_checklist'
app/classes/checklist.rb:78:in 'Checklist::ForProject#calc_checklist'
app/classes/checklist.rb:130:in 'Checklist#num_taxa'
app/models/project.rb:573:in 'Project#name_count'
app/components/projects/tabs.rb:65:in 'Components::Projects::Tabs#names_tab'
app/components/projects/tabs.rb:40:in 'Components::Projects::Tabs#observation_tabs'
app/components/projects/tabs.rb:20:in 'block (2 levels) in Components::Projects::Tabs#view_template'
app/components/projects/tabs.rb:17:in 'block in Components::Projects::Tabs#view_template'
app/components/projects/tabs.rb:16:in 'Components::Projects::Tabs#view_template'
app/components/project_banner.rb:33:in 'block in Components::ProjectBanner#view_template'
app/components/project_banner.rb:32:in 'Components::ProjectBanner#view_template'
app/helpers/tabs/projects_helper.rb:74:in 'block in Tabs::ProjectsHelper#add_project_banner'
app/helpers/tabs/projects_helper.rb:73:in 'Tabs::ProjectsHelper#add_project_banner'
app/views/controllers/projects/show.rb:28:in 'Views::Controllers::Projects::Show#view_template'
app/controllers/projects_controller.rb:55:in 'ProjectsController#show'
app/controllers/application_controller.rb:149:in 'ApplicationController#catch_errors_and_log_request_stats'
Request
Parameters:

{"id" => "259"}
```

mo-nathan:
Given the error, I wonder if this some weird leftover from the work you've been doing on ranks.  I reloaded the database locally using database-20260424.gz from images, ran db:migrate, and the rest and it loads fine.

If you could add a debugger line inside that loop in app/classes/checklist.rb you should be able to get some insight into what's going on.

JoeCohen (approved):

PR #4196, opened 2026-05-03
Map iNat monomial complexes to MO Group names with parent genus

### LINKED ISSUE #4195: Correctly import iNat supraspecific complexes
iNat supraspecific complexes Ex: [Complex Mappae (Amanita Series Mappae)](https://www.inaturalist.org/taxa/1144777-Mappae) should be imported and displayed as <Genus> lowercase iNat complex name (Amannita mappae complex)

### Remaining tasks

- [x] Ensure binomial species (e.g. _Amanita fulva_) is not confused with corresponding complex (Amanita fulva complex).

JoeCohen:
## Summary

- Resolves #4195  
- Fixes a bug in which an iNat monommial `complex` (e.g. `Mappae`, displayed as `Complex Mappae`) was imported as `Mappae complex`, rather than `Amanita mappae complex` (rank: Group)
(This is an edge case which leaked into MCP and GBIF.)

### Prerequisites

- MO dev server running (`bin/rails server`)
- iNat account `jdcohenesq` linked (or any account with access to
  the test observations below)
- Import via **Preferences → iNat Import**

---

### 1. Monomial complex — MO Name exists ✅

Create an iNat obs whose iNat id is a "monomial complex" example: (taxon: Complex Mappae, iNat id 1144777)
**Import the observation.**

**Expected:**
- Observation is created with name `Amanita mappae complex`
- No new Name record is created (reuses existing)
- Naming and vote are recorded for `Amanita mappae complex`
- External link points to the iNat observation

IMPORTANT: Clean up.
- Delete the MO Observation. 
- Delete the MO_Observation_URL observation field from the iNat Observation

---

### 2. Monomial complex — MO Name does not exist ✅

**Create and import an iNat observation with ID `Paraparmelia  complex` (taxon  992274) **

**Expected:**
- A new MO Name is created: `<Genus> <epithet.downcase> complex`
  at rank Group
- Observation is created with that new Name
- Name appears on the MO Name page with correct rank and no author

IMPORTANT: Clean up.
- Delete the MO Observation. 
- Delete the MO_Observation_URL observation field from the iNat Observation
---

### 3. Binomial complex — regression ✅

**iNat observation:** any observation whose iNat taxon is a binomial
complex, e.g. *Xeromphalina campanella* complex.

**Import the observation.**

**Expected:** behavior unchanged from before this PR — maps to
`Xeromphalina campanella group` (or `complex`) at rank Group.

IMPORTANT: Clean up.
- Delete the MO Observation. 
- Delete the MO_Observation_URL observation field from the iNat Observation

---

### 4. Non-complex taxon — regression ✅

**iNat observation:** any species-level observation (e.g. *Amanita muscaria*).

**Import the observation.**

**Expected:** maps to the species Name as before; no change in behavior.

IMPORTANT: Clean up.
- Delete the MO Observation. 
- Delete the MO_Observation_URL observation field from the iNat Observation

---

### 5. Verify no duplicate Name creation on re-import ✅

**Re-import the same monomial complex observation from test 2.**

**Expected:**
- No duplicate Name or Observation is created
- Import completes without error


JoeCohen:
## CodeQL alert analysis

Two alerts on line 83 (`matching_group_names`), both tracing taint from the same two sources: `JSON.parse(res.body)` in `monomial_complex_name_string` (new, line 116) and the pre-existing `infrageneric_name_string` (line 143), flowing through `full_name_string` into the regex.

### Alert 76 — Regular expression injection

```ruby
::Name.where(::Name[:text_name] =~ /^#{full_name_string}/).
```

`full_name_string` is interpolated into a regex literal without `Regexp.escape`. A malicious string containing regex metacharacters (`.`, `*`, `+`, `(`, `[`, etc.) could alter the pattern's meaning or cause catastrophic backtracking (ReDoS).

**Actual risk**: Low. iNat genus and epithet strings are constrained to Latin letters and spaces — no metacharacters in practice. But CodeQL correctly can't verify that constraint.

### Alert 77 — SQL query built from user-controlled sources

Same line. The regex `/^#{full_name_string}/` is passed to MySQL via Arel's `=~` operator, which emits a `REGEXP` condition. If the interpolated string contained SQL-significant characters it could escape the regex literal in the generated query.

**Actual risk**: Low. ActiveRecord parameterizes the regex value safely, and again the iNat API values are well-formed taxon names. But CodeQL traces the external data flow regardless.

### Pre-existing vs. introduced

The original `matching_group_names` used `@taxon[:name]` directly in this same regex — the same class of taint. My change to `full_name_string` extended the taint flow to also include the two `JSON.parse(res.body)` sources in the `*_name_string` methods, which is what CodeQL is now flagging as the sources (lines 116 and 143).

### Fix if desired

Wrapping with `Regexp.escape` silences both alerts and is correct practice regardless of the practical risk:

```ruby
::Name.where(::Name[:text_name] =~ /^#{Regexp.escape(full_name_string)}/).
```

The same fix should be applied to the pre-existing `@taxon[:name]` interpolation in the original code path to address the underlying pattern consistently.

JoeCohen:
## CodeQL alert 78 — SQL injection (persists after `Regexp.escape`)

The `regexp-injection` alert (76) from the previous scan is gone — `Regexp.escape` resolved it. Only the `sql-injection` alert remains (now renumbered 78).

**Why `Regexp.escape` didn't close this one:**

`Regexp.escape` escapes regex metacharacters (`.`, `*`, `+`, `[`, etc.) but does not escape SQL-special characters — notably single quotes. So the taint path from `JSON.parse(res.body)` → `full_name_string` → REGEXP SQL condition is still open in CodeQL's model. CodeQL sees external data flowing into the SQL query regardless of the regex escaping applied.

There is also a more concrete concern: Arel's `=~` operator (MySQL `REGEXP`) may not bind the pattern as a prepared-statement parameter — it may interpolate it directly into the SQL string. If so, a taxon name containing a single quote (e.g. a hypothetical `"O'Brien"`) could break the query.

**Actual risk:** iNat taxon names consist of Latin letters and spaces and will never contain a single quote in practice. But the structural concern is valid.

**What would close this alert:** Replace the `REGEXP` approach with Arel's `matches`, which generates a parameterized `LIKE` clause:

```ruby
::Name.where(::Name[:text_name].matches("#{Name.sanitize_sql_like(full_name_string)}%")).
```

This is semantically equivalent to `/^#{...}/` (prefix match) while using bound parameters, which CodeQL recognises as safe.

JoeCohen:
## Response to Copilot comments

**Comment 2 (duplication)** — already resolved. The latest commit
extracted `ancestor_genus_name` as a shared private helper used by both
`monomial_complex_name_string` and `infrageneric_name_string`, so there
is no longer any duplicated request/parsing logic.

**Comment 1 (timeouts / error handling)** — valid observation, but this
is a pre-existing pattern in the codebase, not something introduced by
this PR. `infrageneric_name_string` has had the same unguarded
`RestClient` call (now `Inat::APIRequest`) without timeouts or rescue
since it was written. The surrounding import job does catch
`StandardError` in `import_one_result` and `RestClient::ExceptionWithResponse`
in `import_page`, so a network failure here will surface as a job-level
error rather than silently hang forever.

Adding timeouts and finer-grained error handling for the genus lookup is
a reasonable improvement, but belongs in a follow-on issue that addresses
`infrageneric_name_string` and `ancestor_genus_name` together rather
than only the new code path.

JoeCohen:
## Species vs. complex confusion

No. Two independent guards prevent it.

**Direction 1 — iNat species, MO Group name:** When the iNat taxon is a
species (*Amanita fulva*, `rank: "species"`), `complex?` is false so
`matching_group_names` is never called. `matching_names_at_regular_ranks`
runs instead, doing an exact `text_name` match at rank `Species` — it
will never surface a Group-ranked name.

**Direction 2 — iNat complex, MO species:** When the iNat taxon is the
complex (`rank: "complex"`), `matching_group_names` runs with
`where(rank: "Group")`. The MO species *Amanita fulva* (rank `Species`)
is excluded by that filter regardless of what the LIKE pattern matches.

There is one unrelated edge case worth noting: the LIKE pattern
`"Amanita fulva%"` would also match a hypothetical MO Group name
*Amanita fulvella complex* (starts with the same prefix). This is
unlikely in practice and was equally true of the previous REGEXP
approach. If it becomes a concern, the pattern would need a trailing
space before `%` — `"Amanita fulva %"` — to anchor to a word boundary.

JoeCohen:
Alleged .001% coverage decrease appears to be something other than any lines becoming uncovered.

mo-nathan:
When I tried example 1 as myself, it said:

Estimated number of imports: 0
Unlicensed observations: 0 (will not be imported — no iNat license)
Estimated time: 00:00:00

mo-nathan:
Also the example:

iNat observation: https://www.inaturalist.org/observations/282528172
(taxon: Complex Mappae, iNat id 1144777)

takes me to a page with name Entoloma medianox not Complex Mappae.  Guessing there's a typo there of some sort.

mo-nathan:
I found https://www.inaturalist.org/observations/348911797 which I think is an example of a name to try, but I don't know if I should worry about leaving weird MO fingerprints on things.  It would help me during testing to be able to not worry about leaving fingerprints when running iNat imports from localhost.

mo-nathan (approved):
I wasn't able to do all the recommended manual tests, but the code looks good and the purpose is clear, so approved.  Thanks!

JoeCohen:
Thanks for the review, and sorry about the Manual test outline.
What happened was that I repurposed the iNat obs to test my other pending
PR. The Manual test outline should have said to create a new iNat obs and
use it.

PR #4222, opened 2026-05-06
Un-pin gem `mini-racer`

nimmolo:
Had been pinned so that gems would bundle on my older Intel mac. 

Now seems unnecessary, can unpin.

Adds a smoke test that the gem actually works, replacing the test for PRs that kept accidentally un-pinning the gem.


mo-nathan (approved):
Yay!

JoeCohen:
approved

PR #4224, opened 2026-05-06
Phlex `ProjectForm` - use `radio_field` helper for `dates_any` toggle

nimmolo:
## Background
Phlex superforms generally need all form fields to either be attributes of one of our db backed models, like `Project` or `Name` — or, if the form is not a CRUD form, the fields need to be attributes of a custom `FormObject` created for the form, as in the case of admin requests, etc. 

With either type of FormObject, Superform can work its validation magic and generally economize on form generation code by passing the fields to the form field helpers defined in `ApplicationForm`. These helpers are what keep our forms rendering consistent Bootstrap HTML.

### The plot twist
However, some CRUD forms in MO need an extra field or two that is not part of the db-backed model, but it also may not be worth creating a new `FormObject` separate from the db model. In these cases, Claude came up with a solution called `FormProxy` - it allows for a custom field to be added to the form and passed to the field helper, without junking up the model with an attribute that won't get saved, it's only passed to the helper for the form UI.

This works fine, if you know about it. However, Claude's right hand knoweth not what its left hand doeth, and so in @mo-nathan's PR, it found another way to render this field, creating its own render method that duplicates the code in `ApplicationForm#radio_field`. This PR is to remove that duplication.

## Summary
- Addresses @nimmolo's [review feedback](https://github.com/MushroomObserver/mushroom-observer/pull/4076/changes/BASE..842ba12dce1dcf0f1dfe09ac9be0566229e598b4#r3036907380) on #4076: The new Phlex `ProjectForm` was using its own inline radio button render method (`render_radio`) instead of using the shared `radio_field` / `ApplicationForm::RadioField` helper.
- Replaces the inline implementation with the existing `FieldProxy` + `RadioField` pattern (same approach `form_name_feedback.rb` uses for non-model-backed radio groups).
- What is `FieldProxy`? It's a Claude invention for Phlex forms, in cases like this where there's a field that is not an attribute of the form object.  In this case`dates_any` is UI state passed in from the controller, not a `Project` attribute, so `field(:dates_any)` can't read it from the model. A code comment explains this.

The rendered HTML from `radio_field` is identical (same name, value, id, checked attributes); only the rendering path changes.

## Test plan
- [x] `bin/rails test test/components/project_form_test.rb` — 4 tests, 26 assertions pass
- [x] `bin/rails test test/controllers/projects_controller_test.rb` — 44 tests, 323 assertions pass
- [x] `bundle exec rubocop app/components/project_form.rb` — no offenses

### DIFF
--- a/.claude/phlex_style_guide.md
+++ b/.claude/phlex_style_guide.md
@@ -29,6 +29,7 @@ fields. **Always use these helpers** instead of the verbose
 text_field(:name, label: "Name:", size: 40)
 textarea_field(:notes, label: "Notes:", rows: 6)
 checkbox_field(:approved, label: "Approved")
+radio_field(:status, ["active", "Active"], ["inactive", "Inactive"])
 select_field(:rank, rank_options, label: "Rank:")
 static_field(:display_name, label: "Name:", value: @model.name, inline: true)
 read_only_field(:locked_field, label: "Value:", value: @value)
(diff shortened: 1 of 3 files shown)


mo-nathan:
@nimmolo Thanks for catching this!  I tried to get Claude to fix this gap.  Please review the proposed style guide changes and ensure that is the correct intent.  I also created #4225 which describes all the existing places where these rules are currently violated in our code base.  Please review that as well to make sure the proposed approaches make sense and are consistent with your vision for how we should be writing this code.

mo-nathan (approved):

mo-nathan:
Brief summary of what was wrong and what was changed.

Why the original PR #4076 slipped past                                                                                        

ProjectForm hand-rolled a render_radio helper using raw Phlex input(type: :radio, …). The reviewer's correct fix (PR #4224) replaces it with FieldProxy + Components::ApplicationForm::RadioField — the same pattern form_name_feedback.rb already uses.

The previous Claude session had three documentation gaps to walk through:

1. Helper section was model-only. phlex_style_guide.md says "Always use these helpers" and lists text_field, textarea_field, checkbox_field, select_field, etc. — every example is a model attribute. radio_field wasn't even in the list. Internally, all of these call field(field_name) on the model, so once Claude saw dates_any is not a Project column, none of the helpers seemed to apply.                 
2. FieldProxy was framed as outside-form-only. The section header was literally "FieldProxy: Form Fields Outside a Form Context," with examples drawn from FormNameFeedback and image fields. Nothing told Claude that the same proxy is also the right answer for non-model fields inside a Superform.
3. No prohibition on raw input inside forms. With both documented paths apparently inapplicable, Claude fell back on what the rest of the style guide explicitly encourages — native Phlex HTML methods like input(...). There was no rule saying "never hand-roll form controls."

What was changed

.claude/phlex_style_guide.md
  - Added radio_field to the helper list so it's visible.
  - Added a new "NEVER hand-roll form-control HTML inside a form component" hard rule with a decision tree (model field → helper; non-model field → FieldProxy + field class) and the ProjectForm.render_dates_any_radios reference example.
  - Renamed the FieldProxy section to "Fields Without a Superform Field Backing" and rewrote it to call out both uses (outside-form AND inside-form-but-non-model), with the ProjectForm example alongside the FormNameFeedback one.
                                                                                                                                
.claude/rules/phlex_conversions.md
  - Added a per-field decision-tree step during writing.
  - Added a self-review step that explicitly grep-checks the diff for input(, select(, textarea(, option( before opening the PR, with PR #4224 cited as the precedent.

Net effect: a future session that has to render a non-model field inside a Phlex form should now hit the rule on the way in (conversion checklist), see the decision tree (style guide), see a real example (ProjectForm.render_dates_any_radios), and catch any leftover hand-rolled tags on the way out (self-review step).

Issue #4208, opened 2026-05-05
Improving how MO handles imported observations (tracking)


mo-nathan:
Implementation tracking for the proposal in discussion #4106. Resolutions of open questions are captured in the discussion comments and in the comment thread on this issue.

This issue tracks the work in five phases. The schema-consolidation work (#4299, #4565) is sequenced early in Phase 3 — ahead of read-only and sync — so that those, and everything in later phases, build on the unified `ExternalLink` model. Each sub-issue stands alone and can be picked up independently as long as its blockers are done.

## Phase 1 — Foundation

- [x] #4209 — Sources table + `external_id` migration
- [x] #4210 — Visible source indicator
- [x] #4211 — Collector + "Entered by"
- [x] #4221 — Harden iNat import: idempotency check + filter MO back-link annotations

## Phase 2 — Quick wins

- [x] #4212 — Naming vote weights for new imports
- [x] #4213 — Discovery: audit existing imports for post-import edits
- [x] #4543 — Triage orphaned iNat images (one-time backlog from the #4213 audit)

## Phase 3 — Unified model, read-only, sync, and export

Ordered by dependency. The unified external-relationship model (#4299, #4565) lands first; read-only (#4214) and sync (#4215) are built on it. MO→iNat export (#4530) joins this phase rather than a final one — the mirror-correspondence materialization is the same work as #4565 (typed `ExternalLink`s from the same field-5005 set), so there's no reason to defer it. This also means the unified model is in place before Phase 4, so MyCoPortal is built on it from the start.

- [x] #4529 — Structured per-photo provenance *(done; prerequisite for #4215 image sync)*
- [x] #4299 — Consolidate Source into ExternalSite/ExternalLink *(the unified external-relationship model; everything below builds on it)*
- [x] #4565 — Reconcile the iNat "Mushroom Observer URL" field (5005); materialize the complete MO↔iNat correspondence (~161k) as typed ExternalLinks — import / mirror / copy / manual *(builds on #4299)*
- [ ] #4585 — Reflection resolution: compare MO obs vs their iNat counterpart(s) and bring the MO side to a clean reflection *(consumes #4565s links + multi-link report; engine for #4214/#4215)*
[...]

Issue #4215, opened 2026-05-05
Sync imported observations from their source


mo-nathan:
## Goal

Keep imported observations up to date instead of treating them as one-time snapshots.

### Triggers

- Daily background job. Batches external IDs in chunks and queries the source API by ID list, filtered by `updated_since`.
- User-initiated "Sync" button on the observation show page.
- Bulk sync by user (their own imports) and by project (admin-initiated).
- Project-level discovery of new matches handled separately in the discovery issue (#4219); this issue is about updating already-imported observations.

### Sync updates (snapshot side)

- Date, location, GPS, image set (add / remove / license changes — if license becomes unsupported or image deleted on source, delete from MO), notes, DQA grade, sequences, field slip codes.
- New naming if the source ID changed: keep the old naming with a lower vote, add a new naming at the appropriate weight (per #4212).
- GPS privacy: respects source geoprivacy changes. Companion observations with independently entered data are unaffected.

### Sync does NOT touch

- MO namings/votes, comments, collection numbers, fungarium records, project membership, occurrence relationships.

### Operational

- Tracks `last_successful_sync_at` per source (column on `sources`).
- Per-observation `last_synced_at`.
- No fixed schedule assumed; supports initiating full syncs when needed.
- OAuth: project-level iNat API calls use the admin/super-importer’s token; require iNat auth (per #4217) if not already done.

### Deleted source observations

- If the source observation no longer exists, mark the MO obs as orphaned with a note. Keep all MO data.

### Sequences

- String comparison detects new or edited sequences.
- Handle multiple sequence variants per observation (e.g., iNat obs 330136750 has three ITS sequences including a consensus). Related: #4086.

## Blocked by

- #4209 — sources table + external_id (needed for batched ID queries by source)
- #4214 — read-only imported observations (sync respects read_only)

## Notes

Tracking issue: #4208.
Source: discussion #4106 §6.

mo-nathan:
Capturing a design idea raised on PR #4230 (https://github.com/MushroomObserver/mushroom-observer/pull/4230#discussion_r3213172711) for when this issue picks up:

[...]

Issue #4214, opened 2026-05-05
Read-only imported observations and occurrence-based editing


mo-nathan:
## Goal

Lock the snapshot fields on imported observations so they cannot diverge from the source. When a user clicks Edit, MO creates a companion observation linked via an Occurrence; the user edits the companion. This also lets users correct imprecise locations by promoting the companion to primary.

Community fields (namings, votes, comments, sequences, projects, collection numbers, fungarium records) remain editable on the imported observation as today.

## Acceptance

- New `read_only` boolean column on observations (separate from `source` — API/mobile-created native obs are not read-only).
- Set `read_only = true` for imports during creation and during the existing-imports migration.
- Snapshot fields locked on read-only obs:
  - Date, location, GPS, location-hidden status, notes, images.
  - Edits via web form rejected with a helpful inline error.
  - API PATCH on locked fields returns an error response with the same message.
- Edit button on a read-only observation creates a companion observation linked via an Occurrence:
  - Companion copies date, location, GPS, hidden status, notes, image references.
  - User can change coordinates and hidden status independently on the companion.
  - Original stays primary by default; user can promote companion to primary.
  - If the read-only obs is already in an Occurrence with an editable sibling, the Edit button goes to that sibling instead of creating another companion.
- Tests cover both the lock behavior and the companion-creation flow.

## Blocked by

- #4209 — sources table (read_only is independent of `source`, but logic uses `source_id`)
- #4213 — discovery audit (pre-existing edits informs migration approach)

## Blocks

- #4215 — sync (sync respects read_only)

## Notes

Tracking issue: #4208.
Source: discussion #4106 §3.

Joe’s alternative (a Comment from an Importer that holds the snapshot, instead of locked columns + companion observations) was considered and not adopted — it creates more issues than it solves. We use the occurrence-based approach as proposed.

mo-nathan:
**Note re #4299 (unified external-relationship model):**

#4299 sits between Phase 3 and Phase 4 and renames `Observation#external_source` → `#import_link`. This issue is written against the current `Source`-based schema; the rename is mechanical and will be done as part of #4299 (or as a small follow-up if this lands first). Nothing in the design here needs to change for the consolidation — `last_synced_at` etc. move from `Source` to `ExternalSite` cleanly.

PR #4230, opened 2026-05-08
Sources table + external_id migration for imported observations

### LINKED ISSUE #4209: Sources table and external_id migration for imported observations
## Goal

Replace the `source` enum and `inat_id` column on `observations` with a generic data model that supports multiple external data sources (iNat, MyCoPortal, future).

Foundation for the rest of #4208 / discussion #4106.

## Acceptance

- New `sources` table: `id`, `name`, `url`, `description`, `last_successful_sync_at`. Initial seed: iNat. (MyCoPortal added with its own issue.)
- New observation columns: `source_id` (nullable FK), `external_id` (nullable string).
- Uniqueness constraint on `(source_id, external_id)` to prevent duplicate imports.
- Migrate existing rows: map current `source` enum values to `sources.id`; copy `inat_id` to `external_id` with `source_id` set to iNat.
- Drop the old `source` enum and `inat_id` column once migration is verified.
- `last_synced_at` (timestamp) added to observations now (used by the sync issue).

## Notes

Tracking issue: #4208.
Source: discussion #4106 §10 (Data Model & Migration).

mo-nathan:
## Summary

- Replaces the iNat-specific `inat_id` column with the generic `(source_id, external_id)` pair on observations, foundation for the rest of #4208.
- Adds a `sources` table seeded with iNaturalist; unique index on `(source_id, external_id)` closes the dedup race that PR #4223 only narrowed (now landable since the historical duplicates have been removed from production).
- Two-axis model per the #4208 resolution: the `source` enum is preserved as the entry agent (`mo_website`, `mo_android_app`, `mo_iphone_app`, `mo_api`); the new `(source_id, external_id)` pair is the external data source. `mo_inat_import` is dropped from the entry-agent enum.

Fixes #4209.

## Schema (3 migrations)

- `CreateSources` — sources table + iNaturalist seed.
- `AddExternalSourceColumnsToObservations` — `source_id`, `external_id`, `last_synced_at`, unique index on `(source_id, external_id)`.
- `BackfillExternalSourceForInatImports` — every obs with `inat_id IS NOT NULL` (50,061 rows) gets `source_id=iNat` + `external_id=CAST(inat_id AS CHAR)`; rows that had `source=mo_inat_import` (50,060) also get `source` cleared. The one outlier — an obs with `source=mo_website` but a manually-attached `inat_id` — keeps its entry agent and gains the iNat external link.

## Application

- New `Source` model with `iNaturalist` lookup helper.
- `Observation.belongs_to :external_source` — named to avoid colliding with the `source` enum.
- `source_credit` / `source_noteworthy?` rewritten to surface external imports through the source link rather than the dropped enum value.
- iNat importer writes `external_source` / `external_id`; `ActiveRecord::RecordNotUnique` rescue closes the simultaneous-job race that PR #4223 only narrowed.
- `inat_imports_controller` previous-import warning + cleanup use `external_id`.
- `observation_labels` reads the iNat URL via `external_source`.
- New locale key `source_credit_external` with `[name]` / `[url]` interpolation; `source_credit_mo_inat_import` removed.
- Eager-loaded `:external_source` in 6 observation-index call sites (matrix box renders `source_credit` for imports; without the include Bullet raises N+1).

## Tests

- New `Source` model test + `sources.yml` fixture.
- Existing inat / observation tests updated for the new column names; `imported_inat_obs` fixture rewritten to use `external_source` + `external_id`.

## Out of scope (follow-up PR)

- Drop the `inat_id` column from `observations` after a deploy or two of dual-write/dual-read safety net.
- Sweep up any remaining locale/doc references to the old column.

## Test plan

- [x] `bin/rails test` — 5138 tests, 36828 assertions, 0 failures, 0 errors
- [x] `bundle exec rubocop` clean on changed files
- [x] Migration verified on local copy of production DB: 50,061 rows linked via `source_id`, 50,060 with entry-agent `source` cleared, 1 outlier preserved its `mo_website` entry agent + gained the iNat external link, 0 rows remain at `source=5`
- [x] Reviewer check: the `source_credit_external` locale string format (currently `'Imported from "[name]":[url]'`) — links to the source's homepage, not to MO's docs about the import process

### DIFF
--- a/app/classes/inat/mo_observation_builder.rb
+++ b/app/classes/inat/mo_observation_builder.rb
@@ -8,10 +8,11 @@ class Inat
     MO_API_KEY_NOTES = InatImportsController::MO_API_KEY_NOTES
     NAMING_VOTE = Vote::MAXIMUM_VOTE
 
-    def initialize(inat_obs:, user:, import_others: false)
+    def initialize(inat_obs:, user:, import_others: false, inat_source: nil)
       @inat_obs = inat_obs
       @user = user
       @import_others = import_others
+      @inat_source = inat_source || Source.inaturalist
       @skipped_images = 0
       @unlicensed_obs = inat_obs[:license_code].blank? ? 1 : 0
     end
@@ -62,7 +63,7 @@ class Inat
         specimen: inat_obs.specimen?,
         text_name: text_name,
         notes: inat_obs.notes,
-        external_source: Source.inaturalist,
+        external_source: @inat_source,
         external_id: inat_obs[:id].to_s }
     end
 
--- a/app/classes/inat/mo_observation_builder.rb
+++ b/app/classes/inat/mo_observation_builder.rb
@@ -62,8 +62,8 @@ class Inat
         specimen: inat_obs.specimen?,
         text_name: text_name,
         notes: inat_obs.notes,
-        source: inat_obs.source,
-        inat_id: inat_obs[:id] }
+        external_source: Source.inaturalist,
+        external_id: inat_obs[:id].to_s }
     end
 
     # NOTE: 1. iNat users seem to add a prov name only if there's a sequence.
--- a/app/classes/inat/observation_importer.rb
+++ b/app/classes/inat/observation_importer.rb
@@ -28,6 +28,7 @@ class Inat
 
     def import_one_result(result)
       @inat_obs = Inat::Obs.new(result)
+      @observation = nil
       return if unimportable?
       return if date_missing?
       return if already_imported?
@@ -67,12 +68,19 @@ class Inat
     # RecordNotUnique exceptions.
     def already_imported?
       return false unless Observation.exists?(
-        source_id: Source.inaturalist.id,
+        source_id: inat_source.id,
         external_id: @inat_obs[:id].to_s
       )
 
       log("Skipped #{@inat_obs[:id]} already imported")
       true
+    end
+
+    # Cached iNaturalist Source row for this importer instance — avoids
+    # one find_by per imported observation across already_imported? and
+    # the builder's new_obs_params.
+    def inat_source
+      @inat_source ||= Source.inaturalist
     end
 
     def log_with_response_error(msg)
@@ -83,7 +91,8 @@ class Inat
     def create_mo_observation
       builder = Inat::MoObservationBuilder.new(
         inat_obs: @inat_obs, user: @user,
-        import_others: @inat_import.import_others
+        import_others: @inat_import.import_others,
+        inat_source: inat_source
       )
       @observation = builder.mo_observation
       builder
@@ -91,9 +100,11 @@ class Inat
       # The (source_id, external_id) unique index caught a race
       # between simultaneous jobs after `already_imported?` returned
       # false. Treat the same as the pre-check skip.
+      @observation = nil
       log("Skipped #{@inat_obs[:id]} already imported (race)")
       nil
     rescue StandardError => e
+      @observation = nil
       log_with_response_error(
         "Failed to import iNat #{@inat_obs[:id]}: #{e.message}"
       )
--- a/app/classes/inat/observation_importer.rb
+++ b/app/classes/inat/observation_importer.rb
@@ -30,6 +30,7 @@ class Inat
       @inat_obs = Inat::Obs.new(result)
       return if unimportable?
       return if date_missing?
+      return if already_imported?
 
       builder = create_mo_observation
       return unless @observation
@@ -56,6 +57,24 @@ class Inat
       true
     end
 
+    # Last-line-of-defense check against duplicate imports.
+    # Upstream filters (iNat-side `without_field` and the controller's
+    # `clean_inat_ids`) miss observations whose back-link write to iNat
+    # failed silently after a prior import, and the controller filter
+    # is bypassed entirely on import-all runs. The unique index on
+    # (source_id, external_id) is the actual race-safety guarantee;
+    # this pre-check just keeps benign duplicates from emitting noisy
+    # RecordNotUnique exceptions.
+    def already_imported?
+      return false unless Observation.exists?(
+        source_id: Source.inaturalist.id,
+        external_id: @inat_obs[:id].to_s
+      )
+
+      log("Skipped #{@inat_obs[:id]} already imported")
+      true
+    end
+
     def log_with_response_error(msg)
       log(msg)
       @inat_import.add_response_error(msg)
@@ -68,6 +87,12 @@ class Inat
       )
       @observation = builder.mo_observation
       builder
+    rescue ActiveRecord::RecordNotUnique
+      # The (source_id, external_id) unique index caught a race
+      # between simultaneous jobs after `already_imported?` returned
+      # false. Treat the same as the pre-check skip.
+      log("Skipped #{@inat_obs[:id]} already imported (race)")
+      nil
     rescue StandardError => e
       log_with_response_error(
         "Failed to import iNat #{@inat_obs[:id]}: #{e.message}"
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -986,7 +986,7 @@ class Observation < AbstractModel # rubocop:disable Metrics/ClassLength
     end
   end
 
-  # Do we want to prominantly advertise the source of this observation?
+  # Do we want to prominently advertise the source of this observation?
   # Reads source_id (the column) directly to avoid loading the
   # external_source association when only the predicate is needed —
   # source_credit takes care of the actual association load.
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -978,9 +978,9 @@ class Observation < AbstractModel # rubocop:disable Metrics/ClassLength
   end
 
   def external_source_credit
-    if external_source.url.present?
-      :source_credit_external.l(name: external_source.name,
-                                url: external_source.url)
+    obs_url = external_source.observation_url(external_id)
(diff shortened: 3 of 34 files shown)


db/migrate/20260505160002_backfill_external_source_for_inat_imports.rb
```diff
@@ -0,0 +1,75 @@
...
+      UPDATE observations
+         SET source_id = #{inat_source_id},
+             external_id = CAST(inat_id AS CHAR)
+       WHERE inat_id IS NOT NULL
+    SQL
+
+    # Clear the entry-agent column only for the dropped enum value.
+    execute(<<~SQL.squish)
+      UPDATE observations
+         SET source = NULL
```
mo-nathan, line 47: Does this make sense?  I could argue that this should be set to the mo_website value since the import is initiated from the website.  I also note that most recent observation that has a source of NULL in the current database is from 2025-07-07.  Here's the distribution of values for observation.source at the moment:
```
mysql> select count(*), source from observations group by source;
+----------+--------+
| count(*) | source |
+----------+--------+
|   472127 |   NULL |
|    69270 |      1 |
|     5866 |      4 |
|    50064 |      5 |
+----------+--------+
4 rows in set (0.67 sec)
```

The information is recoverable after this migration since the 50K obs will all now have source_ids.  In theory we could move all the data carried in the source column into source ids if added Sources for the website and API.  We could also backfill the 472K that are currently NULL with website.

Overall it feels like just make this one localized change is leaving cruft around that we really should cleanup and that it wouldn't be too hard to clean up.  At minimum this should be a new issue in the backlog.

app/models/observation.rb
```diff
@@ -943,24 +952,38 @@ def turn_off_specimen_if_no_more_records
...
-    mo_api: 4,
-    mo_inat_import: 5
+    mo_api: 4
   }
 
-  # Message to use to credit the agent which created this observation.
-  # Intended to be used with .tpl to render as HTML:
+  # Message to use to credit the source of this observation.
+  # External imports take precedence over the entry agent: an obs
+  # synced from iNat surfaces as "Imported from iNaturalist" even if
```
mo-nathan, line 967: This raises an interesting question regarding the idea of exporting MO obs to iNat.  The new model is that the iNat observation data is "read-only" when it gets imported.  Would we stick with that model if the observation was originally created on "MO"?  Seems like that would be a bit awkward, but I can't think how else we'd do it given that we can't expect iNat to implement read-only obs on their side.  I guess we could use the Occurrence concept and create a read-only "snapshot" obs when the export happens.  The "snapshot" obs would not be the primary observation of the Occurrence, but would reflect the state of the data at the time the export was done.  This could then be used by the sync script to decide if new edits happened on the iNat side.  That's not unreasonable and leads us towards a bit more versioning functionality that seems like it could be useful.

app/classes/inat/observation_importer.rb
```diff
@@ -82,6 +87,12 @@ def create_mo_observation
       )
       @observation = builder.mo_observation
       builder
+    rescue ActiveRecord::RecordNotUnique
+      # The (source_id, external_id) unique index caught a race
+      # between simultaneous jobs after `already_imported?` returned
+      # false. Treat the same as the pre-check skip.
+      log("Skipped #{@inat_obs[:id]} already imported (race)")
```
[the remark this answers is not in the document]
mo-nathan: Fixed in ffd54b6. `@observation` is now nil-set at the top of `import_one_result` and again on both rescue branches in `create_mo_observation`, so a failed import cannot reuse a prior successful obs.

app/classes/inat/observation_importer.rb
```diff
@@ -61,10 +61,15 @@ def date_missing?
...
+    # is bypassed entirely on import-all runs. The unique index on
+    # (source_id, external_id) is the actual race-safety guarantee;
+    # this pre-check just keeps benign duplicates from emitting noisy
+    # RecordNotUnique exceptions.
     def already_imported?
-      return false unless Observation.exists?(inat_id: @inat_obs[:id])
+      return false unless Observation.exists?(
+        source_id: Source.inaturalist.id,
+        external_id: @inat_obs[:id].to_s
+      )
```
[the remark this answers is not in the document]
mo-nathan: Fixed in ffd54b6. Added an `inat_source` memo on the importer instance (`@inat_source ||= Source.inaturalist`) and use it from both `already_imported?` and the builder. One `find_by` per import job instead of one per result.

app/classes/inat/mo_observation_builder.rb
```diff
@@ -62,8 +62,8 @@ def new_obs_params # rubocop:disable Metrics/AbcSize
         specimen: inat_obs.specimen?,
         text_name: text_name,
         notes: inat_obs.notes,
-        source: inat_obs.source,
-        inat_id: inat_obs[:id] }
+        external_source: Source.inaturalist,
+        external_id: inat_obs[:id].to_s }
```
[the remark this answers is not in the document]
mo-nathan: Fixed in ffd54b6. `MoObservationBuilder.new` now takes an `inat_source:` kwarg; the importer passes its memoized Source row through, so the builder no longer does a per-result lookup. The kwarg defaults to `Source.inaturalist` for any standalone callers.

app/models/observation.rb
```diff
@@ -943,24 +952,38 @@ def turn_off_specimen_if_no_more_records
...
+  # the user originally created it via mo_website. Returns nil when
+  # there's nothing noteworthy to credit.
+  # Intended for use with .tpl to render as HTML:
   #   <%= observation.source_credit.tpl %>
   def source_credit
-    :"source_credit_#{source}" if source.present?
+    if external_source
+      :source_credit_external.l(name: external_source.name,
+                                url: external_source.url.to_s)
```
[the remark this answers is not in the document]
mo-nathan: Good catch — fixed in ffd54b6. `source_credit` now branches on `external_source.url.present?` and falls back to a new `source_credit_external_no_url` locale key (`Imported from [name]`, no Textile link) when the URL is blank.

db/migrate/20260505160002_backfill_external_source_for_inat_imports.rb
```diff
@@ -0,0 +1,75 @@
...
+    return if inat_source_id.blank?
+
+    # Restore mo_inat_import only on rows that have no other entry
+    # agent — those are the ones we cleared in `up`.
+    execute(<<~SQL.squish)
+      UPDATE observations
+         SET source = #{INAT_ENUM_VALUE}
+       WHERE source_id = #{inat_source_id}
+         AND source IS NULL
+    SQL
```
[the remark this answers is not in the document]
mo-nathan: Documented as intentionally lossy in ffd54b6. The current production data has zero rows where this matters — the only non-import entry agent that ever appeared on an iNat-linked row was `mo_website` (one row), and it keeps its entry agent through `up`, so the rollback predicate (source IS NULL AND source_id=iNat) never matches it. Adding column-level provenance to make rollback non-lossy felt like overkill for a one-shot migration; flagging an exception in the comment instead.

db/migrate/20260505160002_backfill_external_source_for_inat_imports.rb
mo-nathan: I see the appeal of `mo_website` — the import is web-initiated. But I think the two-axis resolution in #4208 wants the entry-agent column to mean "what action did the user take in MO" (filled out a form, hit the API), not "what HTTP route processed this." Setting `source=mo_website` on an import would conflate "the user created this obs via the new-obs form" with "the user triggered a sync job from a different page" — and the existing `source_credit_mo_website` string ("Created via the Mushroom Observer website") would be misleading on the imported rows.

So I lean toward keeping `source=NULL` for imports here, with `source_id IS NOT NULL` as the "this is an import" signal.

That said, the broader cleanup you raised is a real opportunity:

- 472K rows with `source = NULL` predate the enum and could be backfilled (`mo_website` is the right default for pre-2025-07 obs).
- We could add `Source` rows for the MO website/API and migrate the entire `source` enum into `source_id`, retiring the enum entirely. That collapses both axes into one and arguably simplifies the model — though it gives up the easy distinction between "external import" (read-only in #4214) and "MO-native action" (editable).

I will file a backlog issue capturing both options (backfill the 472K NULLs vs. fully unify the columns into Source) so the design tradeoff gets its own discussion separate from this PR.

app/models/observation.rb
mo-nathan: Worth capturing this for the #4215 sync work. The "snapshot obs as a non-primary member of an Occurrence" idea is appealing because it gives us a concrete versioning anchor without bolting per-field history onto Observation. The sync job could compare the live remote payload to the snapshot row and, if they diverge, either auto-update the snapshot (no MO-side conflict) or flag for review (MO-side edits since last sync would conflict with iNat-side edits).

Out of scope for this PR; I will leave a note on #4215 referencing this comment so the design discussion lands there.

db/migrate/20260505160002_backfill_external_source_for_inat_imports.rb
mo-nathan: Filed as #4231 — captures both options (backfill-only vs. unify-into-Source) plus the status-quo option, for design discussion separate from this PR.

app/models/observation.rb
mo-nathan: Note left on #4215 (https://github.com/MushroomObserver/mushroom-observer/issues/4215#issuecomment-4412600155) capturing the snapshot-obs-in-Occurrence idea for when the sync work picks up.

Issue #3637, opened 2025-12-26
Unify report column indexes - fix intermittent report test failures


mo-nathan:
This issue was found while enabling parallel testing.  There is a thread safety issue with the current implementation as well as an overall complexity to how column data is handled in our reporting system that the following analysis concludes should be fixed.

Two architectural approaches:

  Option 1: Unified Column Set - Include all extended columns in the base query
  Option 2: Current Architecture + Thread-Safety Fixes - Keep varied columns but fix race conditions

  Key Findings

  The Root Problem

  The current architecture concatenates two separate SQL queries:
  - rows_with_location (26 columns via JOIN)
  - rows_without_location (26 columns with blanks)

  During parallel testing, frozen array concatenation (observation_selects + user_selects + ...) and Arel query building have thread-safety issues, causing column indices to occasionally shift by one position.

  Performance Analysis

  For a typical report with 100 observations:
  - Current: ~130KB (26 columns)
  - Unified: ~186KB (31 columns)
  - Overhead: ~56KB (43% increase)
  - Network transfer: 0.045 seconds at 100 Mbps (negligible)

  However: The current system already executes 1 base query + up to 5 separate queries for extended columns. Unifying into a single query could actually improve overall performance.

  Recommendation: Implement Unified Column Set

  This approach:
  - ✅ Eliminates race conditions completely - single query structure, no dynamic concatenation
  - ✅ Simplifies architecture - one source of truth for column positions
  - ✅ May improve performance - consolidates 6 queries into 1
  - ✅ Improves maintainability - easier to understand and debug
  - ⚠️ Adds ~56KB overhead per 100 observations - acceptable tradeoff

  Proposed Implementation Strategy

  1. Phase 1: Add extended columns to base selects as LEFT JOINs/subqueries
  2. Phase 2: Update reports to use unified column positions
  3. Phase 3: Remove extend_data! methods and separate queries
  4. Phase 4: Verify tests pass, measure performance
  5. Phase 5: Remove defensive validation (no longer needed)

  This is a significant architectural change. Would you like me to proceed with implementing the unified column approach, starting with Phase 1?

mo-nathan:
Example test run with relevant failures:
```
nathan@Mac mushroom-observer % rails t
Running 5164 tests in parallel using 14 processes
Started with run options --seed 30388

WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert" 00:00:09,  ETA: 00:00:12

[...]

PR #4237, opened 2026-05-10
Switch report rows from positional to named columns (#3637)

mo-nathan:
In theory at least this fixes the intermittent issue with the report tests.

## Summary

Replaces the `Array<Array>` result + `@vals[N]` indexing in `Report::Row` with a `Hash<String, value>` keyed on SQL column aliases. Eliminates the surface where intermittent column-misalignment errors surfaced under parallel testing — a named lookup can't drift when the column order shifts; if a column is missing from a particular SELECT the lookup returns nil rather than reading whatever happened to land at that index.

Issue: #3637

## Why named accessors first

The user [reported](https://github.com/MushroomObserver/mushroom-observer/issues/3637) the failures are very intermittent (one occurrence after running the full suite 12 times with no issues), the failing tests vary, but they all manifest as **column misalignment** in `Report::Row`. The defensive validation added previously catches the symptom but doesn't fix the cause; the actual root cause hasn't been pinned down.

Switching to named accessors removes the failure mode by construction: the data is keyed by SQL alias, not by index. Even if some other code path is racing on the SELECT (Arel internals, Trilogy result iteration, etc.), the result-hash semantics make the misalignment unrepresentable.

This is the simpler half of the issue's "Unified Column Set" recommendation. The other half (folding extension queries into the base SELECT) is a much larger refactor and deferred — if named accessors don't fully resolve the intermittent failures we can revisit.

### Schema-side

- Every SELECT column carries an explicit `.as("alias")` so `select_all(...).to_a` produces consistent hash keys: `obs_id`, `obs_when`, `obs_lat`, `obs_lng`, `obs_alt`, `obs_specimen`, `obs_is_collection_location`, `obs_vote_cache`, `obs_thumb_image_id`, `obs_notes`, `obs_updated_at`, `user_id`, `user_login`, `user_name`, `name_id`, `name_text_name`, `name_author`, `name_rank`, `loc_id`, `loc_name`, `loc_north`, `loc_south`, `loc_east`, `loc_west`, `loc_high`, `loc_low`.
- `rows_with_location` / `rows_without_location` use `select_all` instead of `select_rows`.
- `blanks_for_location` keeps the legacy semantic of putting `observations.where` in the `loc_name` slot for without-location obs (so `row.loc_name` keeps returning the typed-place string there).

### API-side

- `Report::Row#initialize(hash)` (was Array). All accessors (`obs_id`, `obs_when`, `user_login`, `name_text_name`, `loc_north`, …) look up by name internally; **public signatures unchanged**.
- `Report::Row#val(key)` / `add_val(value, key)` take Symbols identifying extension columns. Extension callers in `mycoportal`, `fundis`, `raw`, `symbiota` updated to pass `:collector_ids`, `:image_ids`, `:sequence_ids`, `:herbarium_labels`, `:herbarium_accession_numbers`, `:field_slips`, `:gps_hidden_flag`, `:gps_hidden_lat`, `:gps_hidden_lng`.
- `ProjectTweaker` mutates the row hash by name.
- `Report::RowExtensions#notes_to_hash` reads `self["obs_notes"]`.

### Removed

- The defensive `validate_column_alignment!` machinery in `Report::Row` and its three test cases. Misalignment is no longer expressible.
- The defensive Time/Hash branch in `notes_to_hash`. Same protection is now structural — the `obs_notes` column is read by name and the result is exactly what the SELECT returned.

### Tests

- All `Report::Row.new(vals = [])` test setups switched to `Report::Row.new("col_name" => value)` Hash style.
- Three column-misalignment-detection tests removed (the validation they tested no longer exists).
- Two `notes_to_hash` defensive tests collapsed to one Hash short-circuit case (Time-handling test removed — `obs_notes` is now read by name and can never be a Time).

## Test plan

- [x] `bin/rails test test/classes/report_test.rb` — 49 tests, 438 assertions, 0 failures, 0 errors
- [x] `bin/rails test` — 5154 tests, 36890 assertions, 0 failures, 0 errors
- [x] `bundle exec rubocop app/classes/report test/classes/report_test.rb` — clean
- [x] Reviewer: confirm no third-party consumers of `Report::Row` rely on positional API beyond what's in this diff. (Searched: only the four extension-consuming reports plus the test file.)

## Out of scope (issue's "Unified Column Set" recommendation)

Folding the six extension queries into the base SELECT (subqueries / LEFT JOINs / LATERAL) is a much larger refactor with its own perf trade-offs. Deferred — if the intermittent failures persist after named accessors land, we can revisit then with concrete evidence about where the residual race lives.

### DIFF
--- a/app/classes/report/base_table.rb
+++ b/app/classes/report/base_table.rb
@@ -26,7 +26,10 @@ module Report
 
     def formatted_rows
       tweaker = ProjectTweaker.new(user:)
-      rows = all_rows.map { |row| Row.new(tweaker.tweak(row)) }
+      rows = all_rows.map do |raw|
+        verify_row_keys!(raw)
+        Row.new(tweaker.tweak(raw))
+      end
       rows = sort_before(rows)
       extend_data!(rows)
       sort_after(rows.map { |row| format_row(row) })
@@ -34,6 +37,23 @@ module Report
 
     def all_rows
       rows_with_location + rows_without_location
+    end
+
+    # Catch select_all returning a row that's missing one or more
+    # of the columns our SELECT requested — the failure mode #3637
+    # was added to detect (intermittent column drop / misalignment).
+    # Under named-hash storage a missing key only becomes a silent
+    # nil downstream, so this guard surfaces it loudly at the row
+    # boundary instead.
+    def verify_row_keys!(raw)
+      missing = Row::BASE_KEYS - raw.keys.map(&:to_s)
+      return if missing.empty?
+
+      raise(
+        "Report::BaseTable row missing expected columns: " \
+        "#{missing.inspect}. Got keys: #{raw.keys.inspect}. " \
+        "Row: #{raw.inspect}"
+      )
     end
 
     private
--- a/app/classes/report/base_table.rb
+++ b/app/classes/report/base_table.rb
@@ -5,6 +5,42 @@ module Report
   class BaseTable < Base
     attr_accessor :query
 
+    OBS_SIMPLE_SELECTS = {
+      obs_id: :id,
+      obs_when: :when,
+      obs_alt: :alt,
+      obs_specimen: :specimen,
+      obs_is_collection_location: :is_collection_location,
+      obs_vote_cache: :vote_cache,
+      obs_thumb_image_id: :thumb_image_id,
+      obs_notes: :notes,
+      obs_updated_at: :updated_at
+    }.freeze
+
+    USER_SELECTS = {
+      user_id: :id,
+      user_login: :login,
+      user_name: :name
+    }.freeze
+
+    NAME_SELECTS = {
+      name_id: :id,
+      name_text_name: :text_name,
+      name_author: :author,
+      name_rank: :rank
+    }.freeze
+
+    LOCATION_SELECTS = {
+      loc_id: :id,
+      loc_name: :name,
+      loc_north: :north,
+      loc_south: :south,
+      loc_east: :east,
+      loc_west: :west,
+      loc_high: :high,
+      loc_low: :low
+    }.freeze
+
     def initialize(args)
       super
       self.query = args[:query]
@@ -83,21 +119,13 @@ module Report
     end
 
     def observation_selects
-      cols = obs_simple_selects
-      cols.insert(2, public_latlng_spec(:lat).as("obs_lat"),
-                  public_latlng_spec(:lng).as("obs_lng"))
-      cols.freeze
+      obs_simple_selects +
+        [public_latlng_spec(:lat).as("obs_lat"),
+         public_latlng_spec(:lng).as("obs_lng")]
     end
 
     def obs_simple_selects
-      [[:id, "obs_id"], [:when, "obs_when"], [:alt, "obs_alt"],
-       [:specimen, "obs_specimen"],
-       [:is_collection_location, "obs_is_collection_location"],
-       [:vote_cache, "obs_vote_cache"],
-       [:thumb_image_id, "obs_thumb_image_id"],
-       [:notes, "obs_notes"],
-       [:updated_at, "obs_updated_at"]].
-        map { |attr, alias_| Observation[attr].as(alias_) }
+      OBS_SIMPLE_SELECTS.map { |aliaz, attr| Observation[attr].as(aliaz.to_s) }
     end
 
     def public_latlng_spec(col)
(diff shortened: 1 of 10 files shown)


app/classes/report/row.rb
```diff
@@ -1,233 +1,185 @@
...
-
-      # name_text_name should be String, not numeric or Time
-      return unless @vals[15].is_a?(Numeric) || @vals[15].is_a?(Time)
-
-      raise_column_misalignment_error(
-        15, "name_text_name", "String", @vals[15].class
-      )
+    # Generic getter — returns nil if the column wasn't selected.
+    def [](key)
+      @vals[key.to_s] || @vals[key.to_sym]
```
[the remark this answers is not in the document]
mo-nathan: Fixed in 0a73f9f. `Row#[]` now does explicit `@vals.key?` checks — string key first, then symbol — so a stored `false` (or any falsey value) is returned as-is instead of getting masked into a "missing column" nil.

app/classes/report/row.rb
```diff
@@ -1,233 +1,185 @@
...
+  # The base set of columns is defined by Report::BaseTable's SELECT
+  # aliases:
+  #   obs_id, obs_when, obs_lat, obs_lng, obs_alt, obs_specimen,
+  #   obs_is_collection_location, obs_vote_cache, obs_thumb_image_id,
+  #   obs_notes, obs_updated_at,
+  #   user_id, user_login, user_name,
+  #   name_id, name_text_name, name_author, name_rank,
+  #   loc_id, loc_name, loc_north, loc_south, loc_east, loc_west,
+  #   loc_high, loc_low,
+  #   obs_where  (only populated for the without-location query)
```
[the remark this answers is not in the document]
mo-nathan: Fixed in 0a73f9f. Dropped the `obs_where` mention from the header comment and replaced it with a description of the actual mapping: `observations.where` is aliased into the `loc_name` slot for without-location obs (preserves the legacy positional behavior where `row.loc_name` returned the typed-place string).

Also added a `BaseTable#verify_row_keys!` defensive guard that raises if any `select_all`-returned row is missing one of `Row::BASE_KEYS`. Under named-hash storage a missing column would silently degrade to nil downstream; the guard surfaces the failure at the row boundary instead. Per-row cost is ~hundreds of nanoseconds (≤100ms on the largest production reports), which the user signed off on given the move toward threaded execution. Two new tests exercise the guard.

nimmolo (changes requested):
Minor quibble, not a blocker.

app/classes/report/base_table.rb
```diff
@@ -36,46 +39,65 @@ def all_rows
...
+
+    def obs_simple_selects
+      [[:id, "obs_id"], [:when, "obs_when"], [:alt, "obs_alt"],
+       [:specimen, "obs_specimen"],
+       [:is_collection_location, "obs_is_collection_location"],
+       [:vote_cache, "obs_vote_cache"],
+       [:thumb_image_id, "obs_thumb_image_id"],
+       [:notes, "obs_notes"],
+       [:updated_at, "obs_updated_at"]].
+        map { |attr, alias_| Observation[attr].as(alias_) }
```
nimmolo: @mo-nathan This is the method i was trying to comment on, `obs_simple_selects`. The array of tuples could maybe be a frozen hash in the class?

nimmolo (changes requested):

app/classes/report/base_table.rb
mo-nathan: I asked CC: "does it make sense for the array in obs_simple_selects in app/classes/report/base_table.rb to be a frozen hash constant?", to which it returned:

```
Tentative yes, but only if you commit to the version of it that's actually useful. The half-version isn't worth the diff.

  What "frozen hash constant" should mean here

  Don't store Arel nodes in the constant — store plain data (symbol → symbol) and build Arel inside the method. Reasons:

  - Arel attribute nodes are constructed via Observation.arel_table[:id], which forces Observation to load when the constant is evaluated. If
  Report::BaseTable autoloads ahead of Observation, that's a sequencing wart. Plain symbols sidestep it.
  - Arel nodes are mostly-immutable in practice but not contractually so; a few visitor paths inside ActiveRecord mutate intermediate state. Reusing
  the same .as("obs_id") instance across every report run isn't known broken, but it's the kind of "fine until it isn't" the other .freezeed methods in
   this file have quietly inherited.
  - Tests can rebuild the schema; cached Arel attrs pinned to the previous arel_table would be stale. Raw symbols don't have that problem.

  So the shape I'd back is:

  OBS_SIMPLE_SELECTS = {
    obs_id:                     :id,
    obs_when:                   :when,
    obs_alt:                    :alt,
    obs_specimen:               :specimen,
    obs_is_collection_location: :is_collection_location,
    obs_vote_cache:             :vote_cache,
    obs_thumb_image_id:         :thumb_image_id,
    obs_notes:                  :notes,
    obs_updated_at:             :updated_at
  }.freeze

  def obs_simple_selects
    OBS_SIMPLE_SELECTS.map { |aliaz, attr| Observation[attr].as(aliaz.to_s) }
  end

  That gives you a single source of truth for the column set that can be diffed against Row::BASE_KEYS directly (OBS_SIMPLE_SELECTS.keys.map(&:to_s)),
  which lines up with the whole point of #3637 — making column drops impossible to ignore.

  The thing that makes it actually pay off

  The current observation_selects does:

  cols.insert(2, public_latlng_spec(:lat).as("obs_lat"),
              public_latlng_spec(:lng).as("obs_lng"))

  That magic 2 is a maintenance hazard: rearrange obs_simple_selects and lat/lng silently move to the wrong column position. If you're going to make
  this declarative, fold lat/lng into the same hash and let the order speak for itself:

  OBS_SELECTS = {
    obs_id:                     :id,
    obs_when:                   :when,
    obs_lat:                    :__public_lat__,
    obs_lng:                    :__public_lng__,
    obs_alt:                    :alt,
    obs_specimen:               :specimen,
    obs_is_collection_location: :is_collection_location,
    obs_vote_cache:             :vote_cache,
    obs_thumb_image_id:         :thumb_image_id,
    obs_notes:                  :notes,
    obs_updated_at:             :updated_at
  }.freeze

  def observation_selects
    OBS_SELECTS.map do |aliaz, attr|
      case attr
      when :__public_lat__ then public_latlng_spec(:lat).as(aliaz.to_s)
      when :__public_lng__ then public_latlng_spec(:lng).as(aliaz.to_s)
      else Observation[attr].as(aliaz.to_s)
      end
    end
  end

  obs_simple_selects then disappears, the insert(2, ...) disappears, and the column order is visible in one place. That's the version where the
  constant earns its keep.

  What I'd push back on

  Doing this only for the obs side. user_selects, name_selects, location_selects, and blanks_for_location are all the same shape (instance methods
  returning a .freezeed Arel array). One declarative constant amid four instance-method returns is the worst of both worlds for a reader — they have to
   mentally context-switch. Either:
  
  - Apply the same {alias_sym => attr_sym}.freeze treatment to all five and have one consistent pattern across the file, or
  - Leave obs_simple_selects alone.

  For a PR that's specifically about defending against column-set drift, the "all five" version is the right amount of effort. Doing the obs one in
  isolation is more churn than insight.

  Want me to sketch the full conversion?
```

I haven't fully wrapped my head around that yet.

mo-nathan: Ok, I get it now.  The bit about "Don't store Arel nodes in the constant" is not something anyone was considering.  I also think the `.insert(2, ...)` is wrong.  I think I see a good approach that generalizes.

nimmolo (approved):

JoeCohen (approved):
Thanks for this fix. (And it has the bonus that the aliases are transparent -- unlike the magic column numbers.)

PR #4239, opened 2026-05-10
Fix typo

jonkiparsky:
Pedantic typo fix.


mo-nathan:
That test failiure is curious.  I'm going to try re-running the tests so we can merge this with a clean bill of health.  However, it looks like you hit a rare intermittent test failure that I don't think I've seen before.

mo-nathan:
Found the issue using Claude Code.  PR is on the way.  It has to do with being an external contributor.  We should support that, so it's worth fixing.

mo-nathan:
#4240 is the relevant PR.

JoeCohen:
Thanks for pitching in.

PR #4236, opened 2026-05-10
Replace hand-rolled form inputs with helpers / FieldProxy (group A of #4225)

nimmolo:
## Summary
Addresses **Group A — quick wins** from #4225. Every form control inside an `ApplicationForm` subclass now goes through a field helper or `FieldProxy` + field class, not a raw `input` / `select` / `textarea` call.

- `external_link_form`: hidden `user_id` and `observation_id` → `hidden_field`
- `name_form`: `approved_rank` → `FieldProxy` + `TextField(type: "hidden")`
- `translation_form`: flat `tag` hidden input → `FieldProxy` + `TextField(type: "hidden")`
- `description_form`: `old_desc_id` and `delete_after` → `FieldProxy` + `TextField(type: "hidden")` (extracted shared `render_flat_hidden_field` helper)
- `occurrence_form`: submit `input` → `submit()` helper
- `occurrence_edit_form`: submit `input` → `submit()` helper

Group B (bare-select sites in `identify_filter_form` and `translation_form`) and Group C (`occurrence_form` / `occurrence_edit_form` array-param checkboxes/radios and the manual `_method=patch` form) remain open per the umbrella issue.

Refs #4225.

## Test plan
- [x] `bundle exec rubocop` clean on all six changed files
- [x] `bin/rails test test/components/external_link_form_test.rb test/components/name_form_test.rb test/components/translation_form_test.rb test/components/occurrence_form_test.rb test/components/occurrence_edit_form_test.rb` — 80 tests, 318 assertions, 0 failures
- [x] `bin/rails test test/controllers/names/descriptions_controller_test.rb -n /merge/` — 2 tests, 8 assertions, 0 failures (covers the `old_desc_id` / `delete_after` path)
- [x] Self-review grep `^\s*(input|select|textarea|option)\(` on all six files — only Group B / Group C call sites remain
- [ ] Smoke-test in browser: create/edit external link, create misspelled name with `approved_rank`, edit a translation, merge two name descriptions, create and edit an occurrence

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4225, opened 2026-05-07
Phlex forms: replace hand-rolled form-control HTML with helpers / FieldProxy


mo-nathan:
## Background

PR #4224 fixed a hand-rolled radio group in `Components::ProjectForm` (introduced in PR #4076) by replacing it with the canonical `FieldProxy` + `Components::ApplicationForm::RadioField` pattern. The Phlex style guide and conversion checklist have been updated to forbid hand-rolling form-control HTML inside any `ApplicationForm` subclass.

This umbrella ticket tracks the remaining violations of that rule discovered while auditing `app/components/*_form.rb`.

**Rule** (from `.claude/phlex_style_guide.md`): inside any `Components::ApplicationForm` subclass, do NOT emit raw `input` / `select` / `textarea` / `option` Phlex tags. Every form control must come from either:

1. An `ApplicationForm` field helper (`text_field`, `radio_field`, `checkbox_field`, `select_field`, `date_field`, `hidden_field`, `submit`, etc.) — for fields that ARE attributes of the model / FormObject.
2. A `FieldProxy` + matching field class (`RadioField`, `TextField`, `CheckboxField`, `SelectField`, …) — for fields that are NOT model attributes.

## Group A — quick wins (one-line replacements)

Each item: replace the hand-rolled `input` with a helper or `FieldProxy`-backed field.

- [x] `app/components/external_link_form.rb:37` — `external_link[user_id]` → `hidden_field(:user_id, value: @user&.id)`.
- [x] `app/components/external_link_form.rb:38` — `external_link[observation_id]` → `hidden_field(:observation_id, value: @observation.id)`.
- [x] `app/components/name_form.rb:40` — `approved_rank` (controller-side flag, not on `Name`) → `FieldProxy` + `TextField` with `type: \"hidden\"`, or move to a FormObject.
- [x] `app/components/translation_form.rb:29` — flat `tag` hidden input → `FieldProxy` (or move to `FormObject::Translation` namespacing if a flat name is no longer needed).
- [x] `app/components/description_form.rb:142` — `old_desc_id` → `FieldProxy` + `TextField` with `type: \"hidden\"`.
- [x] `app/components/description_form.rb:143` — `delete_after` → same pattern.
- [x] `app/components/occurrence_form.rb:21` — `input(type: \"submit\", ...)` → `submit(:create_occurrence_submit.l, center: true, class: \"btn btn-default my-3\")`.
[...]

PR #4245, opened 2026-05-11
Enable bare select for superform `SelectField` (group B of #4225)

nimmolo:
## Summary
Addresses **Group B — needs a small helper extension first** from #4225. Adopts **Option B1**: extend `SelectField` so `wrapper_options[:label] == false` skips the form-group wrapper (matching `TextField`'s existing convention), then convert the two call sites that previously fell back to raw Phlex `select(...)` calls.

- `application_form/select_field`: skip the `form-group` wrapper when `label: false`; extracted `render_select` and `label_text` helpers.
- `identify_filter_form`: `filter[type]` uses `select_field(:type, type_options, label: false, ...)` — `:type` is on `FormObject::IdentifyFilter`. Flipped `type_options` to Superform's `[value, label]` order.
- `form_object/identify_filter`: coerce unknown / `nil` `type` values to `"clade"` in the setter. This moves the "invalid type → clade" fallback from the form component into the FormObject, where it belongs, and is required so the bare `<select>` always renders a selected option.
- `translation_form`: the locale switcher uses `FieldProxy + SelectField` with `label: false`. Locale isn't a `FormObject::Translation` attribute (it's a UI control wired to Stimulus). Passes `id: nil` so Phlex omits the FieldProxy-supplied id and the rendered `<select>` keeps the original markup.

Group A is already merged (#4236). Group C (`occurrence_form` / `occurrence_edit_form` array-param checkboxes/radios and the manual `_method=patch` form) remains open.

Refs #4225.

## Test plan
- [x] `bundle exec rubocop` clean on all four changed files
- [x] `bin/rails test test/components/identify_filter_form_test.rb test/components/translation_form_test.rb` — 44 tests, 143 assertions, 0 failures
- [x] `bin/rails test test/components/account_signup_form_test.rb test/components/application_form_test.rb test/components/external_link_form_test.rb test/components/form_image_fields_test.rb test/components/identify_filter_form_test.rb test/components/name_form_test.rb test/components/names_lookup_field_group_test.rb test/components/project_alias_form_test.rb test/components/sequence_form_test.rb test/components/search_form_test.rb test/components/translation_form_test.rb` (every form that consumes `select_field`) — 181 tests, 830 assertions, 0 failures
- [x] Local before/after parity check via Nokogiri DOM diff on both forms (every variant — default / clade / region / invalid type for the filter form; official / non-official lang, single / multi tag, with / without for_page for translation form). No missing classes, no missing attributes, no nesting differences. Scaffolding deleted before commit.
- [x] Smoke-test in browser: navbar Identify Observations filter (toggle clade/region, submit, paste URL with bogus `?filter[type]=garbage` to confirm fallback still works); Edit Translations modal (switch locale, save a tag).

### DIFF
--- a/app/classes/form_object/identify_filter.rb
+++ b/app/classes/form_object/identify_filter.rb
@@ -3,9 +3,18 @@
 # Form object for the identify observations filter form.
 # Params namespaced as filter[term], filter[type], etc.
 class FormObject::IdentifyFilter < FormObject::Base
+  VALID_TYPES = %w[clade region].freeze
+
   attribute :term, :string
   attribute :term_id, :string
   attribute :type, :string, default: "clade"
+
+  # Coerce unknown / nil type values to "clade" so the form select
+  # always has a valid option to select and downstream callers can
+  # trust the value.
+  def type=(value)
+    super(VALID_TYPES.include?(value) ? value : "clade")
+  end
 
   def self.model_name
     ActiveModel::Name.new(self, nil, "Filter")
--- a/app/components/application_form.rb
+++ b/app/components/application_form.rb
@@ -474,18 +474,31 @@ class Components::ApplicationForm < Superform::Rails::Form
   # @param options [Hash] submit button options
   # @option options [Boolean] :center center the button (default false)
   # @option options [String] :submits_with text shown while submitting
+  # @option options [String] :btn_class Bootstrap button class
+  #   (default `"btn-default"`). Pass the full class name, e.g.
+  #   `"btn-outline-default"` to match the top-nav search bar style.
+  # @option options [Symbol] :as `:input` (default) renders `<input
+  #   type="submit">`; `:button` renders `<button type="submit">value</button>`.
+  #   Default stays `:input` because most controller/system tests look for
+  #   `input[type='submit']` selectors.
   # @option options [String] :class additional CSS classes
   # @option options [Hash] :data additional data attributes
-  def submit(value = submit_value, center: false, submits_with: nil, **options)
+  def submit(value = submit_value, center: false, submits_with: nil, # rubocop:disable Metrics/ParameterLists
+             btn_class: "btn-default", as: :input, **options)
     submits_with ||= :SUBMITTING.l
-    classes = %w[btn btn-default]
+    classes = ["btn", btn_class]
     classes << "center-block my-3" if center
     classes << options[:class] if options[:class].present?
 
     data = { turbo_submits_with: submits_with,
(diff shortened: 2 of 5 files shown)


mo-nathan:
It's a bit of scope creep, but as long as we're in here, can we add a bit of white space around buttons in the identify filter?  Not a blocker.

Screenshot 2026-05-12 at 7 54 24 AM

mo-nathan (approved):
Looks good!  Smoke test passed.  One suggested tweaked that is definitely not a blocker, but seems like an easy pickup.

nimmolo:
Search bar styling changed to match the main search bar -- outlined submit buttons, spacing. 

Screenshot 2026-05-12 at 11 24 17 PM

nimmolo:
Heads up on two things landing in the last commit (1127ab0fa) that go a bit beyond Group B's original "bare select" scope. Calling them out here for review:

**1. `ApplicationForm#submit` grew two new keyword arguments.**

- `btn_class:` — full Bootstrap class controlling the button style. Defaults to `"btn-default"` (no behavior change for existing callers). Pass e.g. `btn_class: "btn-outline-default"` to match the top-nav search bar's button.
- `as:` — renders an `<input type="submit">` (default, `:input`) or `<button type="submit">value</button>` (`:button`). Default stays `:input` because most controller/system tests look for `input[type='submit']` selectors and switching everything to `<button>` would break them.

```ruby
submit(:SEARCH.l, btn_class: "btn-outline-default", class: "px-2")
submit(:SAVE.l)  # unchanged — still gets btn btn-default
```

**2. The identify filter form switched layout from Bootstrap 3 navbar inline-block to flexbox-with-gap.**

- Form class: `navbar-form navbar-left pl-0` → `navbar-flex flex-grow-1 navbar-form px-0 gap-2`. The `gap-2` handles inter-item spacing (replaces the per-button `ml-2`).
- The autocompleter wrap now uses `d-flex flex-grow-1 mb-0`, and the term input gets `class: "flex-grow-1"` so it expands to fill the row. `size: 42` is gone — width is flex-driven now.
- The type select is wrapped in `form-group text-nowrap mb-0` to keep its natural width and not grow.
- Submit buttons use `btn-outline-default px-2` to match the top-nav search bar.

This is the layout pattern the top-nav `_search_bar.html.erb` partial already uses; the identify filter wasn't following it before. Joining the term input and the type select into a single visual control (the Bootstrap 3 `.input-group` style) was attempted in an earlier commit on this branch but broke the Bootstrap navbar styling, so it's reverted and left for a separate ticket.

PR #4246, opened 2026-05-11
Add assert_html_element_equivalent helper to ComponentTestCase

nimmolo:
## Summary
- Adds `assert_html_element_equivalent(expected_html, actual_html, selector:, label: nil, strip_csrf: true)` to `ComponentTestCase` for before/after parity checks during Phlex refactors. Walks the subtree at `selector` recursively and asserts element name, attribute set (order-agnostic), attribute values, text content, and child count all match.
- Sits alongside the existing `compare_html_elements` (which prints a diagnostic and is attribute-order-sensitive). New cross-referencing docstrings explain when to reach for which.
- Strips `<input name="authenticity_token">` values before comparison by default (toggleable via `strip_csrf:`) so rotating CSRF tokens don't cause spurious failures. Path-prefixed failure messages locate the first mismatch; both subtrees are dumped to `/tmp/html_parity_*.html` for inspection.

Motivation: during #4225 Group B I built throwaway scaffolding that did this whole-tree (and was deleted before merge). The selector-scoped, assertion-based, attribute-order-agnostic shape is the version worth keeping around — it complements the existing diagnostic helper without overlapping.

## Test plan
- [x] `bin/rails test test/component_test_case_test.rb` — 16 tests, 89 assertions, 0 failures (covers happy path, every mismatch kind, selector-not-found in either side, CSRF strip default vs. disabled, label propagation, nested-path reporting)
- [x] `bundle exec rubocop test/component_test_case.rb test/component_test_case_test.rb` clean

### DIFF
--- a/test/component_test_case.rb
+++ b/test/component_test_case.rb
@@ -336,6 +336,63 @@ class ComponentTestCase < UnitTestCase
     puts("=" * 60)
   end
 
+  # Assert that the subtree rooted at +selector+ is structurally
+  # equivalent between two HTML strings.
+  #
+  # Compares element name, attribute set (order-agnostic), attribute
+  # values, text content, and child structure recursively. Attribute
+  # order does NOT affect equivalence — only attribute presence and
+  # values do. This is the right helper when verifying that a Phlex
+  # refactor preserves the markup contract a JS/CSS layer depends on:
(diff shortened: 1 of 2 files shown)


mo-nathan (approved):
Looks good!  Now we just have to remember to use it.

PR #4247, opened 2026-05-12
Switch `superform` from @nimmolo's fork to upstream ~> 0.7.0

nimmolo:
## Summary

Drops the `nimmolo/superform` fork pin in favor of the released `~> 0.7.0` gem. Upstream now has @nimmolo's multiple-select work (rubymonolith/superform#64), but the maintainer redesigned Checkbox/Radio along different lines than @nimmolo's collection branch (rubymonolith/superform#65 declined) — so MO's custom field components are here adapted to delegate to upstream's plural `Checkboxes` / `Radios` components via their block syntax.

### Deprecation refactor

Version 0.7 deprecates the `attributes:` keyword on `Components::Base#initialize`, so every `*Field` constructor and every `Field` override + direct call site now spreads `**attributes` instead of passing `attributes:` as a keyword. Eliminates the `[DEPRECATION] Passing 'attributes:' keyword to ...` warning that fired on every render.

### Component changes

- **`SelectField`** — passes `options:` instead of deprecated `collection:` to upstream; reads `@options` from the parent ivar.
- **`RadioField`** — delegates to `Components::Radios` via block form. Stringifies the value so Phlex doesn't dasherize symbol attrs (e.g. `:mycoportal_image_list`). Uses a value-derived `index` so the rendered id is value-based (matching MO's pre-upstream convention). Computes `checked` ourselves with stringified comparison — upstream `Radio`'s `field.value == @value` doesn't coerce types, and MO routinely pairs boolean/symbol field values with string option values.
- **`CheckboxField`** — keeps inheritance from upstream `Checkbox`; branches on (a) boolean mode → `super`, (b) block-given → existing `cb.option(value)` matrix pattern used by `PermissionsForm`, (c) positional options → delegate to `Components::Checkboxes` via block, wrapping each option in MO's `.checkbox` div.
- **`HiddenField`** (new) — replaces the three `TextField.new(proxy, type: "hidden")` call sites in `description_form`, `name_form`, `translation_form`. Purely for brevity and consistency.

## Test plan

- [x] `bundle exec rubocop app/components/` clean (164 files)
- [x] `bin/rails test test/components/` — 875 tests, 4536 assertions, 0 failures
- [x] `bin/rails db:test:prepare && bin/rails test` (full suite) — 5160 tests, 36915 assertions, 0 failures
- [x] Smoke test in browser: any form that uses radio buttons (e.g. project create with `dates_any`, observations download format select), any matrix-checkbox form (description permissions), any plain checkbox form (login `remember_me`, name edit `locked`), translation modal (locale dropdown), identify navbar filter (clade/region toggle).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,7 +70,7 @@ gem("literal")
 # Enable Slot API for Phlex
 gem("phlex-slotable")
 # Superform for type-safe forms with Phlex
-gem("superform", github: "nimmolo/superform", branch: "nimmo-add-radio-fields")
+gem("superform", "~> 0.7.0")
 # Strict ivars: raises a NameError if an ivar is nil (undefined). Must be
 # required in config/boot.rb to work: https://github.com/yippee-fun/strict_ivars
 # gem("strict_ivars", require: false)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,15 +5,6 @@ GIT
     mo_acts_as_versioned (0.6.6)
       activerecord (>= 3.0.9)
 
-GIT
-  remote: https://github.com/nimmolo/superform.git
-  revision: 25449f8adf9febad3531ba3e3e3038629c031fbb
-  branch: nimmo-add-radio-fields
-  specs:
-    superform (0.6.1)
-      phlex-rails (~> 2.0)
-      zeitwerk (~> 2.6)
-
 GEM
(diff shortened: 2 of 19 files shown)


mo-nathan (approved):
Looks good!  Smoke test passed.  Code makes sense.  Thanks!

Issue #4258, opened 2026-05-14
ERB ↔ Phlex form-comparison audit: divergences requiring decisions


nimmolo:
# ERB ↔ Phlex form-helper audit

Phlex form conversion needs ERB/Phlex parity tests; this audit catalogs every divergence between the two, so we know which need fixing, and which are intentional improvements that should be accepted as the new canonical shape.

## Already addressed via stacked PRs (status as of 2026-05-13)

- **PR #4253** — Autocompleter parity nit: drop `ml-3` from ERB `autocompleter_has_id_indicator`; switch Phlex AutocompleterField from `autocompleter: true` (boolean → bare `data-autocompleter`) to `autocompleter: "true"` (string → `data-autocompleter="true"`).
- **PR #4254** — SelectField nil-key fix: coerce `nil` keys to `""` in `Components::ApplicationForm::SelectField#options` so `[nil, "(No Project)"]` pairs render as `<option value="">…` (matching Rails' `select` helper) rather than `<option>…` which would submit the label text as the form value. Includes a regression test.
- **PR #4255** — Native text-year date helper: rewrite ERB
  `date_select_with_label` to emit day-select + month-select +
  year-text-input directly (matching Phlex `DateField`); delete
  `app/javascript/controllers/year-input_controller.js` and its
  stale references in `form-exif_controller.js`. Drops unused
  `start_year`/`end_year`/`order` kwargs from field-slip caller.

## Item 0 — Empty label-row wrapper in ERB `text_label_row` (PROPOSED PR)

All ERB `*_with_label` helpers that share `text_label_row`
(`text_field_with_label`, `text_area_with_label`,
`select_with_label`, `autocompleter_field`, …) **unconditionally**
emit a flex wrapper around the label, even when there's no help
icon / between text / label_end content:

```html
<div class="d-flex justify-content-between">
  <div><label class="mr-3" for="…">…</label></div>
  <div></div>           <!-- empty when no help/between/label_end -->
</div>
```

Phlex's `FieldLabelRow#simple_label?` (in
`app/components/application_form/field_label_row.rb`) short-circuits
to bare `<label class="mr-3">…</label>` when `help_slot`,
`between_slot`, and `label_end_slot` are all absent.

**Recommended fix:** modify `text_label_row` in
`app/helpers/forms_helper.rb` to do the same short-circuit, so the
ERB matches Phlex's emission. Visual rationale: when only one of
the two flex children is non-empty (the label-containing one),
[...]

PR #4264, opened 2026-05-14
Parameterize value in ERB radio_with_label for= attribute

nimmolo:
## Summary

[Issue #4258](https://github.com/MushroomObserver/mushroom-observer/issues/4258) item 9 — recommended fix.

`radio_with_label` interpolated the value verbatim into the label's `for=`:

```ruby
args[:form].label("#{args[:field]}_#{args[:value]}") …
```

But Rails' `radio_button` internally sanitizes the value when generating the input's id (lowercase, whitespace → underscore, other non-word chars stripped). So for a value like `"Hello World"`:

```html
<label for="widget_flavor_Hello World">
  <input type="radio" id="widget_flavor_hello_world" …>
</label>
```

The for-attr never matched the input id, breaking label click-focus and screen-reader association.

Apply `parameterize(separator: "_")` to the value before interpolation. It's the public-API equivalent of Rails' internal `sanitized_value` and matches Phlex `RadioField#index_for`, so ERB and Phlex emit identical for-attrs.

## Test plan

- [x] New regression tests in `forms_helper_test.rb`:
  - `test_radio_with_label_for_attr_matches_sanitized_radio_button_id` — exercises the original bug case (`"Hello World"`); asserts both the radio input `id=` and the label `for=` end up `widget_flavor_hello_world`.
  - `test_radio_with_label_for_attr_with_symbol_value` — common MO caller pattern (`:txt`, `:rtf`, `:csv` from `species_lists/downloads`); asserts the for/id pair for symbols.
- [x] 4 tests / 18 assertions pass
- [x] `bundle exec rubocop` clean on both changed files
- [ ] Visual eyeball on a page with radio buttons that uses non-trivial values (e.g. `species_lists/download` with `:txt`/`:rtf`/`:csv`, `support/donate` with `:amount` integers + "other"). Behavior change is invisible for typical values; only visible when a value would have failed sanitization before.

### DIFF
--- a/app/helpers/forms_helper.rb
+++ b/app/helpers/forms_helper.rb
@@ -117,15 +117,26 @@ module FormsHelper # rubocop:disable Metrics/ModuleLength
   end
 
   # Bootstrap radio: form, field, value, label, class, checked
+  #
+  # The label's `for=` matches the id Rails' `radio_button` generates for
+  # the input — which sanitizes the value (lowercase, spaces → underscores,
+  # special chars stripped). Without sanitization here, a value like
+  # "Hello World" or a non-trivial symbol would yield a label
+  # `for="field_Hello World"` while the input id is `"field_hello_world"`,
+  # breaking label click-focus and screen-reader association.
+  # `parameterize(separator: "_")` is the public-API equivalent of Rails'
+  # internal `sanitized_value` and matches Phlex `RadioField`'s `index_for`.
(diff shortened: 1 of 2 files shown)


nimmolo:
Folded into #4261 as commit `fddef7f28` — keeping both sides of the for-attr parity (Phlex items 4 + ERB item 9) in a single review.

PR #4279, opened 2026-05-17
OccurrenceResolveForm: Superform + drop OccurrenceResolveModal wrapper

nimmolo:
## Summary

Net deletion of one component + one test.

- **`OccurrenceResolveForm` → Superform.** Now inherits `Components::ApplicationForm` instead of `Components::Base`. The `<form>` tag, CSRF token, `_method` hidden field, and the create-vs-edit action URL come from `form_action` + the model — drops the manual `authenticity_token_field` and the parallel `render_create_form` / `render_edit_form` methods (Superform's one-form-per-component model differentiates by the supplied `form_action`).
- **`OccurrenceResolveModal` deleted.** The thin wrapper combined the form with its modal chrome. The two view callers (`field_slips/new.html.erb`, `edit.html.erb`) now render `Components::Modal` directly with `OccurrenceResolveForm` in the body. Audited: this is the only auto-opening server-rendered modal in MO, so no reuse opportunity justified a named wrapper. Uses `Components::Modal`'s `auto_open:` prop (introduced in #4278, now on main).

## Test plan

- [x] `bin/rails test test/components/occurrence_resolve_form_test.rb` — 6 tests, 44 assertions
- [x] `bin/rails test test/controllers/field_slips_controller_test.rb` — 70 tests, 343 assertions
- [x] RuboCop clean on `occurrence_resolve_form.rb`
- [ ] Manual smoke test on a field-slip with project gaps to verify the modal still opens and the form submits
- [ ] CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4292, opened 2026-05-18
TrustSettingsForm: Superform + render via ModalTurboForm

nimmolo:
Refactor the form to use Superform and the existing modal component.

## Summary

- Splits `Components::TrustSettingsModal` into `Components::TrustSettingsForm` (Superform) rendered inside `Components::ModalTurboForm`. Same pattern as OccurrenceResolveForm in #4279.
- Form uses `ApplicationForm`'s `hidden_field` for `target`, `FieldProxy` for the `commit` radio group (radio name lives outside the model), and `submit(... name: "save")` to avoid colliding with the radios' `commit` name (the controller switches on `params[:commit]` to determine which trust level to apply).
- Removes hand-rolled `<input>` / `<form>` tags per `.claude/rules/phlex_conversions.md`.
- Replaces the modal component test with a focused form test (CSRF/_method/action, radio names+values, per-level pre-selection, help text, save-button `name="save"` collision avoidance, cancel button `data-dismiss="modal"`).

## Test plan

- [x] `bin/rails test test/components/trust_settings_form_test.rb` — 6/6 pass
- [x] `bin/rails test test/controllers/projects/members_controller_test.rb` — 40/40 pass
- [x] `bundle exec rubocop` clean on touched files
- [x] CI green
- [x] Manual: click "Trust Settings" on a project member, change radio, save


mo-nathan:
The manual functionality looks good.

However, I asked CC to "Review the changes on nimmo-trust-settings-form.  Determine if these changes would be made by Claude Code given the current Phlex standards.  If not, what revisions of these standards would have implemented these changes?"

It responded:
```
Would Claude Code make these changes today? No, mostly not. The diff goes beyond what the standards prescribe.

  What Claude Code would produce given the current standards:
  - TrustSettingsForm as a Superform subclass, using radio_field and submit helpers — covered.
  - Rendered via ModalTurboForm with form_class: and form_locals: — covered (the Pattern B section at lines 92-122, plus the OccurrenceResolveForm
  precedent).
  - FieldProxy if the radio field weren't model-backed — covered.

  What it wouldn't arrive at on its own:
  1. The form spans .modal-body + .modal-footer. Nothing in the guide mentions modal section structure. A naive conversion (what #4292 originally shipped)
  puts everything inside .modal-body and synthesizes a text-right mt-3 button row — that drifts from the pre-Phlex chrome but doesn't violate any 
  documented rule. This is the whole reason PR #4293 had to add the :form_content slot and this branch had to opt into it.
  2. self.owns_modal_sections? as a class-method opt-in. This is a convention introduced by this branch, with no precedent in the codebase or guide.
  3. Threading modal_ids: { body:, flash: } through ModalTurboForm → form. Also new convention.
  4. wrapper_options: { wrap_class: "mb-2" } to preserve radio row spacing. This was caught by an HTML-parity diff against the pre-refactor modal, not by
  following the standards.

  What the standards would need to add to make Claude Code arrive here on its own:

  - A "Form Inside a Modal" section in phlex_style_guide.md. Specify that:
    - The <form> tag must span both .modal-body and .modal-footer (so submit-in-footer is inside the form, matching pre-Phlex chrome and Bootstrap
  expectations).
    - Forms designed for modal rendering emit their own .modal-body / .modal-footer divs in view_template and declare self.owns_modal_sections? = true.
    - Such forms accept a modal_ids: { body:, flash: } kwarg, which ModalTurboForm will pass automatically.
    - Don't synthesize footer chrome (text-right mt-3 etc.) as a stand-in for a real .modal-footer — restructure to use :form_content. Add this as a "Modal
   Form Anti-Patterns" callout next to the existing option() and nested-<form> callouts.
  - Make the HTML-parity diff mandatory for modal/form ERB→Phlex conversions, not just a debugging technique. .claude/rules/phlex_conversions.md currently
  lists steps 1-5 (read guides, render, test). Add step 6: "If the component replaces an ERB modal or form, diff rendered HTML against the pre-refactor
  markup. Surface any class/margin drift (mb-2, mt-3, text-right, etc.) and either keep parity via wrapper_options: { wrap_class: ... } or call out the
  change explicitly in the PR description." This matches the testing.md "HTML Diff Technique" but elevates it from debug-only to required-pre-merge.
  - A short note about wrap_class: on collection fields (RadioField, CheckboxField). Pre-refactor modals/forms often put .mb-2 on per-row Bootstrap .radio
  / .checkbox wrappers; Superform's default omits it. The existing FieldProxy example at line 604 uses wrap_class: but doesn't flag why — adding "use
  wrap_class: to preserve pre-refactor row spacing" would close that gap.

  Net: Claude Code today would ship a working but visually-drifted modal form, like #4292 did. The branch's three additions (form spans both sections,
  owns_modal_sections? opt-in, wrap_class preservation) are exactly what the standards are missing. Want me to draft the style-guide patch?
```

Do these changes to the Phlex standard seem reasonable?  If so, let's add them to this PR to avoid more churn in the future.

mo-nathan (approved):
Approved with the caveat of potentially revising our Phlex standard to avoid this sort of churn in the future.

JoeCohen (approved):

nimmolo:
Added doc patches (commit 894bb8397) addressing the conventions Nathan's CC review flagged as missing from the style guide:

- **`.claude/phlex_style_guide.md`** — new "Form Inside a Modal" subsection (Pattern A: form spans `.modal-body` + `.modal-footer` via `:form_content`; Pattern B: form lives entirely in `:body`); explicit "why two ids" for `modal_ids: { body:, flash: }`; "Modal Form Anti-Patterns" callout against synthesizing `text-right mt-3` footer chrome; the existing `wrap_class:` example annotated with WHY (preserve pre-refactor `.mb-2` row spacing).
- **`.claude/rules/phlex_conversions.md`** — adds step 6: mandatory HTML-parity diff for ERB→Phlex modal/form conversions. Lists the common drops (`.mb-2`, footer-row classes, custom data-attrs, submit placement) and the preserve-or-call-out rule.

The "must provide both modal-body and modal-footer" framing is qualified: only for forms with a distinct footer-button row. Forms that are all body content (one inline submit, confirmation-style forms) just use the `:body` slot — explicit Pattern B with example.

nimmolo:
@mo-nathan good call, I have added the Phlex documentations in several places.

Luckily i think we're done with modal form conversions, with this batch of PRs.

nimmolo:
Coveralls is wrong, there's 100% coverage on all files (not all lines, all files) changed by this PR.

PR #4298, opened 2026-05-18
bundle update for May 18, 2026

mo-nathan:


JoeCohen:
I reproduced the coverage decrease locally.
- In main, ran the test suite and coveralls show the relevant line is covered. app/components/checklist/contents.rb#observed_summary_text line 72 `:checklist_observed_higher_only.t(higher: higher, ...`
- Pulled the branch, ran the test suite, and coveralls show the relevant line is not covered.
Screenshot 2026-05-18 at 2 21 23 PM

I also get multiple deprecation warnings like the following in the branch:

> The `helpers` method is deprecated and will be removed in the next
> minor version of phlex-rails.
> 
> If you absolutely need to access the underlying Rails view context,
> you can do that via the `view_context` method, though this is
> not recommended.
> 
> It is much safer to use one of the built-in helper adapters or to
> register your own adapter via `register_output_helper` or
> `register_value_helper`.
> 
> See https://www.phlex.fun/rails/helpers

IMO neither of these is a blocker given that the PR was motivated by a high severity Dependabot Alert

PR #4300, opened 2026-05-18
Delete Components::AddObsModal — controller renders Modal via Phlex view

nimmolo:
## Summary

`AddObsModal` was a thin wrapper around `Components::Modal` (hand-rolled modal markup pre-#4293, then a 9-line `view_template` that just composed Modal after #4293's slot API).

Replaces it with a Phlex view at `app/views/controllers/projects/members/add_obs_modal.rb` that composes `Components::Modal` directly with `:body` and `:footer` slots. Controller renders the view via `render(Views::Controllers::Projects::Members::AddObsModal.new(...), layout: false)`.

## Why a view (not Components/)

- `app/components/` houses reusable Phlex components (`Modal`, `Panel`, field helpers, form components). This modal isn't reusable — it's the rendering of one specific controller action.
- `app/views/controllers/` houses Phlex templates for specific controller actions — exactly what this is. Same location as `occurrences/new.rb`, `field_slips/new.html.erb`, the existing Members `Edit` / `Index` / `New` views.

## Why not render `Components::Modal` directly from the controller

The natural-looking shape is:

```ruby
render(Components::Modal.new(id: "modal_add_obs", title: ...)) do |m|
  m.with_body { p { plain(body_text) } }
  m.with_footer { ... }
end
```

This **doesn't work from a controller** because Phlex Slotable just stores the slot block without `instance_exec`ing it on the Modal instance — when Modal later invokes the block, `self` is whatever it was at capture time. From a Phlex view (like `occurrences/new.rb`), capture-time `self` IS a Phlex component, so `p {}` / `button {}` resolve on the Phlex DSL. From a controller, those calls fall through to `Kernel` methods (`Kernel#p` prints to stdout, returns nil) and contribute nothing to the rendered HTML.

Verified empirically with a one-off test: `m.with_body { p { plain("hello") } }` called from a non-Phlex test class produces no `<p>` in the rendered modal.

Workarounds (rejected):

- Use Rails view helpers in the slot blocks: `m.with_body { view_context.tag.p(body) }` and `m.with_footer { safe_join([cancel, " ", submit]) }`. Returns html_safe strings, Phlex captures them. Works but reads worse and mixes idioms inside the controller.
- Inline the entire modal markup via `view_context.tag.div(...)` chains. Loses Modal composition.

The Phlex view file is the clean answer and matches MO's existing pattern.

## Test moves

- Deletes `test/components/add_obs_modal_test.rb` (the component is gone).
- Folds the `body_text` + `submit_label` branch coverage into `test/controllers/projects/members_controller_test.rb` (where the modal HTML is now exercised end-to-end through the controller). New tests cover:
  - "Add All" submit label (under-limit)
  - "Add Next N" submit label (over-limit, via `add_obs_batch_limit` stub)
  - Cancel button always present
  - Submit button absent when count is zero
  - Modal title in `.modal-header`
  - Submit form's action URL + `_method=put` + `commit` / `target` query params (dispatch contract)
  - Partial-batch body text

## Test plan

- [x] `bin/rails test test/controllers/projects/members_controller_test.rb` — 45/45 pass
- [x] CI green
- [x] Per-file coverage 100% on `members_controller.rb` and `add_obs_modal.rb`
- [x] Manual: navigate to a project as a member, click "Add My Observations", verify the modal renders + submit/cancel buttons in `.modal-footer`


nimmolo:
Coverage is 100% on files changed in this PR.

PR #4301, opened 2026-05-18
Stabilize Checklist::Contents coverage (seed-dependent line 72)

mo-nathan:
## Summary

Coverage of `app/components/checklist/contents.rb:72` — the `elsif higher.positive?` branch of `observed_summary_text` — was seed-dependent: missed under `--seed 61015`, hit under `--seed 1580`. Diagnosis confirmed with stderr instrumentation across both seeds.

## Root cause

`test_checklist_marks_deprecated` in `test/controllers/checklists_controller_test.rb` did:

```ruby
observation = Observation.joins(:name).find_by(name: { deprecated: true })
user = observation.user
get(:show, params: { user_id: user.id })
```

With no `ORDER BY`, MySQL returns rows in implementation-defined order. Under different seeds / parallel-worker assignments, `find_by` returned different deprecated observations:

| Seed | obs returned | user | branch hit | line 72? |
|---|---|---|---|---|
| 61015 | 457573027 (`Baaaa...`) | dick | species + higher (L67-68) | no |
| 1580 | 98434105 (Petigera) | vouchered_only_user | higher only (L72) | yes |

The test's only assertion (`.checklist a` text matches `) *\z`) passed either way, so the variance was silent — it only showed up in coverage reports. No dedicated component test existed for `Components::Checklist::Contents`, so all coverage of its branches came through this single render.

## Fix

Two small changes:

- **New** `test/components/checklist/contents_test.rb` — five tests exercising every branch of `observed_summary_text` with a stub data object (both-positive, species-only, higher-only, singular-`taxon`, and both-zero). Coverage of lines 67-68, 70, and 72 is now seed-independent.
- `test_checklist_marks_deprecated` orders the `find_by` by id so it deterministically returns the same row across seeds and workers. Ends the silent variance.

## Test plan

- [x] `bin/rails test test/components/checklist/contents_test.rb` — 5 tests, 21 assertions, 0 failures.
- [x] `bin/rails test test/controllers/checklists_controller_test.rb test/models/checklist_test.rb test/components/checklist/` — 37 tests, 222 assertions, 0 failures.
- [x] `bundle exec rubocop` clean on both changed files.
- [x] Full-suite runs under both `--seed 61015` and `--seed 1580` (verified by user) now consistently cover line 72.

### DIFF
--- a/test/components/checklist/contents_test.rb
+++ b/test/components/checklist/contents_test.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+
+require("test_helper")
+
+# Component tests for Components::Checklist::Contents.
+#
+# Exercises the three branches of `observed_summary_text` directly so
+# coverage is independent of whatever data the controller tests happen
+# to render. Before this file existed, coverage of the higher-only
+# branch (`elsif higher.positive?`) depended on whichever observation
+# `Observation.joins(:name).find_by(name: { deprecated: true })` in
+# `test_checklist_marks_deprecated` happened to return — an unordered
+# `find_by` whose result varied with test seed / parallel-worker
+# state. See #4XXX for the diagnosis.
+class Components::Checklist::ContentsTest < ComponentTestCase
+  # Both species and higher-level observed → :checklist_observed_summary
+  # branch (lines 67-68 in contents.rb).
+  def test_summary_both_species_and_higher_observed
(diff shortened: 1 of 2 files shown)


JoeCohen (approved):
Looks good.
(I assume you will deal with the Copilot comments about the comments.)

PR #4303, opened 2026-05-19
Modal: knobs for header/controller/body_class; refactor Confirm + Spinner

nimmolo:
## Summary

Extends `Components::Modal` with three props so the two remaining hand-rolled singleton modals (`ModalConfirm`, `ModalProgressSpinner`) can compose Modal instead of inlining their own `.modal > .modal-dialog > .modal-content` markup. Per the Modal docstring follow-up note, this finishes the "predate this and inline their own modal markup" cleanup.

**Why central composition:** every MO modal goes through one source of truth for chrome styling. A future BS3 → BS4 / BS5 migration touches `Components::Modal` once, not N hand-rolled modals.

## New Modal props

- **`controller:`** (String, default `"modal"`) — Stimulus controller wired to the modal root. `ModalConfirm` uses `"confirm-modal"` (the Turbo confirm-replacement dialog has its own controller that mutates the title text and confirm-button action target at runtime).
- **`header:`** (Boolean, default `true`) — when `false`, skip rendering the `.modal-header` div entirely. `ModalConfirm` renders its title inside `.modal-body` so the Stimulus controller can target it without also tracking a header element; `ModalProgressSpinner` has no title at all. `aria-labelledby` is still emitted on the root — point it at the in-body element via `title_id:`.
- **`body_class:`** (String, default `nil`) — extra CSS classes appended to `.modal-body`. `ModalConfirm` wants `py-4`, `ModalProgressSpinner` wants `text-center`.

## Refactored components

- **`ModalConfirm`** — composes Modal with `header: false`, `controller: "confirm-modal"`, `body_class: "py-4"`, `title_id: "mo_confirm_title"`. Renders the title `<h4>` + buttons via `with_body` / `with_footer` slots.
- **`ModalProgressSpinner`** — composes Modal with `header: false`, `dialog_class: "modal-dialog modal-sm"`, `body_class: "text-center"`, `body_id`/`title_id` overrides, plus `extra_data` for the non-dismissible settings (`keyboard: false`, `backdrop: static`) and the `section-update` auto-hide listener.

## HTML parity diff (verified via harness vs. `origin/main`)

- Both modals: `class="modal"` → `class="modal fade"` (Modal default; adds BS3 fade-in animation, consistent with every other MO modal).
- `ModalConfirm` body: `.modal-body.py-4` gains `id="mo_confirm_body"` (Modal's default body_id derivation). No external CSS/JS targets this id.

No other markup differences. No external refs broken.

## Test plan

- [x] 3 new Modal tests for `controller:` / `header:` / `body_class:`.
- [x] `ModalConfirm` test strengthened: header absence, title inside `.modal-body.py-4`, aria-labelledby targeting the in-body title.
- [x] `ModalProgressSpinner` test strengthened: header + footer absence, `.modal-sm` dialog, `.text-center` body, aria-labelledby targeting the caption span.
- [x] `bin/rails test test/components/modal_test.rb test/components/modal_confirm_test.rb test/components/modal_progress_spinner_test.rb` — 12 tests, 92 assertions
- [x] CI green
- [x] Manual: trigger a Turbo confirm somewhere (e.g. a `data-turbo-confirm` form action) and verify the styled modal appears + fade-in animates; manually trigger the spinner via a save action and verify

### DIFF
--- a/app/components/modal.rb
+++ b/app/components/modal.rb
@@ -81,6 +81,23 @@ class Components::Modal < Components::Base
   # `modal_<identifier>_body` convention).
   prop :title_id, _Nilable(String), default: nil
   prop :body_id, _Nilable(String), default: nil
+  # Stimulus controller wired to the modal root. Defaults to `modal`
+  # (the standard show/hide controller). Override for singleton layout
+  # modals with their own controllers (e.g. `confirm-modal` for the
+  # Turbo confirm dialog). Multiple controllers can be passed as a
+  # space-separated string.
+  prop :controller, String, default: "modal"
+  # Render the `.modal-header` div (close button + title). Set to
+  # false for headerless modals where the "title" element lives
+  # inside the body for Stimulus targeting (`ModalConfirm`) or where
+  # the modal has no user-visible title at all (`ModalProgressSpinner`).
+  # `aria-labelledby` is still emitted on the modal root; point it at
+  # the in-body element via `title_id:`.
+  prop :header, _Boolean, default: true
+  # Extra CSS class(es) appended to the `.modal-body` div, e.g.
+  # `"py-4"` for ModalConfirm or `"text-center"` for
+  # ModalProgressSpinner. Joined with the base `"modal-body"`.
+  prop :body_class, _Nilable(String), default: nil
 
   slot :title_content
   slot :body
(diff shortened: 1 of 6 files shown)


nimmolo:
Coverage is at 100% for files changed in this PR

PR #4315, opened 2026-05-21
Project + species_list membership, `ImagesEditForm` Phlex

nimmolo:
## Summary

Replaces the per-row `<group>[id_<id>]=1/0` hash shape with Rails-idiomatic has_many-through array shape across all four forms in MO that let users toggle project / species-list membership via checkboxes:

| Form | Old wire shape | New wire shape |
|------|----------------|-----------------|
| `SpeciesListForm` (Phlex) | `project[id_<id>]=1/0` | `species_list[project_ids][]=<id>` |
| `ObservationForm` projects panel (Phlex) | `observation[project][id_<id>]=1/0` | `observation[project_ids][]=<id>` |
| `ObservationForm` lists panel (Phlex) | `observation[list][id_<id>]=1/0` | `observation[species_list_ids][]=<id>` |
| Image edit form (was ERB) | `project[id_<id>]=1/0` | `image[project_ids][]=<id>` |

Same observable behavior; cleaner contract.

## What's in scope

- **Three Phlex form components** updated (`SpeciesListForm`, `ObservationFormProjects`, `ObservationFormLists`).
- **One ERB → Phlex conversion** along the way (`Components::ImageEditForm`, replacing `app/views/controllers/observations/images/edit.html.erb` + its `_fields_for_images` and `_project` partials). Doing the conversion here means the controller contract is already set for the eventual full Phlex conversion of the image flow.
- **Four controller-side update paths** (`species_lists#update_projects`, `observations_controller/shared_form_methods#update_projects` and `#update_species_lists`, `observations/images_controller#update_related_projects`) read the new shape. Custom iterators preserved — they're the server-side enforcement layer that disabled-checkboxes-don't-submit alone can't provide.
- **`Components::ApplicationForm::CheckboxField#option(value, **overrides)`** extended to accept per-option attribute overrides. Backwards-compatible — no existing callers passed overrides.

## Why a custom iterator (vs. Rails' `*_ids=` setter directly)

Rails' has_many-through `project_ids=` writes to the join table **immediately** on a persisted parent — INSERT/DELETE on assign. Two consequences we cared about:

1. **Permission edge case**: each form intentionally shows projects-the-record-belongs-to-but-the-user-isn't-a-member-of as **disabled** checkboxes. Per HTML spec, disabled inputs don't submit. If we used Rails' setter directly, a tampered submission could detach those projects (or attach unauthorized ones). The custom iterator only toggles `@user.projects_member` projects, so non-member attachments are preserved by omission.
2. **Failure-reload trap**: when validation fails, we re-render the form with the user's submitted choices. Setting `model.project_ids = submitted` for that preview would **silently commit** the join-table changes on a persisted parent — even though the save itself failed. To avoid that we pass `@submitted_project_ids` (and `@submitted_list_ids`) through the view and the form uses that array for `checked:` directly, without touching the model.

## `ignore_proj_conflicts` moved

The obs form's "ignore project warnings" checkbox previously lived under `project[ignore_proj_conflicts]` (alongside the per-project hash). With the projects shape now a flat array under `observation[project_ids][]`, the checkbox now lives at `observation[ignore_proj_conflicts]`. Controller reads `params.dig(:observation, :ignore_proj_conflicts)`.

## Sentinel hidden inputs

Each panel emits one `<input type="hidden" name="…[]" value="">` so the params key is always present even when every checkbox is unchecked — Rack drops empty arrays otherwise. Controllers `compact_blank` to strip the empty value.

## Test plan

Component + controller + integration tests in this branch:

- `bin/rails test test/components/` — 1042 tests, 0 failures
- `bin/rails test test/controllers/observations/` + `test/controllers/observations_controller_project_list_test.rb` + `test/controllers/species_lists_controller_test.rb` — all green
- `bin/rails test test/integration/capybara/images_integration_test.rb` — 2 tests green
- `bundle exec rubocop` clean on every file touched

**Manual smoke test (localhost dev server):**

Each form, both users. Pick a few projects you're a member of, toggle, save, verify the wire shape + result in the DB.

#### As nathan
- New species list: http://localhost:3000/species_lists/new
- Edit species list (2300, "Cape Cod Specimens"): http://localhost:3000/species_lists/2300/edit
- New observation: http://localhost:3000/observations/new
- Edit observation (1, "Xylaria polymorpha group"): http://localhost:3000/observations/1/edit
- Edit image (1997302): http://localhost:3000/images/1997302/edit

#### As Joe Cohen 
- New species list: http://localhost:3000/species_lists/new
- Edit species list (2331, "2025 NEMF possible duplicates"): http://localhost:3000/species_lists/2331/edit
- New observation: http://localhost:3000/observations/new
- Edit observation (603265, "Amanita sp. 'multisquamosa-01'"): http://localhost:3000/observations/603265/edit
- Edit image (1996914): http://localhost:3000/images/1996914/edit

Spot-check via DevTools Network panel: form posts should include `<model>[project_ids][]=<id>` (and `observation[species_list_ids][]=<id>` for the obs form's lists panel), not the old `project[id_<id>]=1/0` hash.


nimmolo:
All touched files have 100% coverage.

PR #4317, opened 2026-05-21
Fix and DRY `page_title` / `document_title` methods

### LINKED ISSUE #4316: Browser tab <title> shows escaped textile/HTML markup for observation + image pages
## Problem

The browser tab `<title>` for observation and image pages shows literal textile source (`_Russula virescens_`) or literal HTML tags (`<i>Russula</i>`) instead of the formatted name. The `<title>` element is plain-text — it doesn't render HTML — so any markup in the string surfaces as escaped text.

Affects (at least):
- Observation edit pages
- Observation show pages
- Image show pages

## Where it lives

Two callsites pass rich-formatted names into a plain-text context:

- `app/views/controllers/observations/edit.html.erb:2` — `add_edit_title(@observation.user_format_name(@user), @observation)`. `user_format_name` returns textile source (`_Russula virescens_`).
- `app/views/controllers/images/show.html.erb:2` — `add_show_title(@image.format_name.t.small_author, @image)`. `.t` produces HTML (`<i>Russula</i>`).

Both flow through the document-title builders in `app/helpers/header/title_helper.rb`:

- `show_document_title` (line 66): `[type_tag, "id:", string].safe_join(" ")`
- `edit_document_title` (line 94): `[:EDIT.l, show_document_title(...)].safe_join(" ")`

`safe_join(" ")` is fine for the on-page heading (where HTML renders), but wrong for the document title (where it doesn't).

## Proposed fix

Strip HTML / textile from the string in `show_document_title` / `edit_document_title` before joining. One chokepoint, fixes every callsite that flows through these helpers — including any future ones — without per-callsite vigilance.

Implementation sketch:

```ruby
def show_document_title(string, object)
  [
    :"#{object.type_tag.to_s.upcase}".l,
    "#{object.id}:",
    plain_text_for_title(string)
  ].safe_join(" ")
end

def plain_text_for_title(string)
  # Handles both textile source (_x_, *x*) and already-rendered HTML
  # (<i>x</i>, <b>x</b>) by stripping tags then unwrapping textile.
  ActionController::Base.helpers.strip_tags(string.to_s).
    gsub(/[_*]+/, "")
end
```

(Exact stripping rules tbd — `strip_tags` + a textile-regex pass should handle the common cases; might need to use MO's `.t.html_to_ascii` or similar if one already exists.)

## Test plan

- Add specs for `show_document_title` / `edit_document_title` covering: plain text (unchanged), HTML-wrapped (`<i>x</i>` → `x`), textile source (`_x_` → `x`), mixed.
- Manual smoke: visit an obs edit, an obs show, and an image show — confirm browser tab text is clean.

nimmolo:
Fixes #4316.

## Problem

HTML document `<title>` for obs/image pages (shown in the tab of browsers and phones) was showing escaped textile source (`_Russula virescens_`) or literal HTML tags (`<i>Russula virescens</i>`). The `<title>` element renders as plain text — any markup leaks through literally.

But generally speaking, MO's title system relied on views to supply the right title strings for both the page title and the document title. This seemed like it could be simplified, and that this would also help with Phlex refactoring of view templates. 

## Fix

Since each model builds its page title differently from the name of the record, give each model `page_title` / `document_title` methods, and make the title helper call them on the page @object.

### New helper API

Callsites simplify from
```erb
add_show_title(@observation.format_name.t.small_author, @observation)
```
to
```erb
add_show_title(@observation, user: @user)
```
(Note - only Observation has to supply the `user`, because of our `user_format_name` complication. Most callers just send the object, `add_show_title(@project)`.) 

### Per-model methods

Each model gains:

- `page_title(user = nil)` — rendered HTML/textile string for the visible page heading. Optional `user` arg lets some models apply user-specific name prefs (hide_authors above_species).
- `document_title` — plain text for the browser tab `<title>`. No textile, no HTML. The helper prepends `"<TYPE> <id>:"` so this string doesn't include either.

`AbstractModel` provides defaults returning the localized type-tag label (e.g. `"Publication"`, `"External Link"`), so models without anything more meaningful just work.

Updated:
Observation, Image, SpeciesList, Project, Location, Article, Herbarium, HerbariumRecord, Sequence, CollectionNumber, GlossaryTerm, VisualGroup, VisualModel, Description, Name, User.

### Observation: the one special case

The obs show heading wraps the consensus name in a link to the name page (with a `(Site ID)` flag when the owner's preferred naming differs). That logic lives in `observations_helper#observation_show_title` + `owner_naming_line` — view-layer work that doesn't belong on the model. The title helper has access via the helper chain, so it special-cases `Observation` in `page_title_for`.

### Why models for everything else

- Single chokepoint: every flow through `add_show_title` / `add_edit_title` benefits, including future ones, without per-callsite vigilance.
- Models already own the textile/plain pair: `format_name` and `text_name` (or similar) — `page_title` and `document_title` just route to them.
- Per-class plain-text source (e.g. `obs.text_name`, the denormalized binomial column on Observation) avoids fragile string-stripping at output time.

## Drive-by cleanups

- **Standardize `string_with_id(name)`** across the various `unique_format_name` / `unique_text_name` methods that were inlining `name + \" (#{id || \"?\"})\"`. Replaced in Article, Project, Location, Naming, Name::Format. AbstractModel's `string_with_id` already handled the `id || \"?\"` fallback.
- **Remove stray `add_pager_for(@image)`** from `locations/edit.html.erb` — `@image` is never set on `LocationsController`, so the call was a copy-paste no-op from another template.
- **Fix `herbarium_records/edit.html.erb`** which was joining `format_name.t + herbarium_label` — same string twice (they're aliases).
- **Fix herbarium edit title**: `add_edit_title(:HERBARIUM.l, @herbarium)` → `add_edit_title(@herbarium)` uses the actual herbarium name + institution code via `Herbarium#format_name`, instead of the generic \"Herbarium\" label.

## Tests

Rewritten `test/helpers/header/title_helper_test.rb`: the old `title_tag_contents` strip-textile assertions are replaced with `document_title_for` / `page_title_for` assertions verifying the new architecture's contract (model returns plain text for the doc title; the helper composes `"<TYPE> <id>: <plain>"`).

## Test plan

- [x] `bin/rails test test/helpers/` — passes
- [x] `bin/rails test test/controllers/` — 1924 tests, 0 failures
- [x] `bundle exec rubocop` clean on every touched file (119 inspected)
- [x] Manual smoke: visit obs show / obs edit / image show / SL show / herbarium show — confirm browser tab text is clean AND visible heading renders normally (italics, links where applicable)

Fixes #4316.

### DIFF
--- a/app/helpers/header/title_helper.rb
+++ b/app/helpers/header/title_helper.rb
@@ -132,9 +132,17 @@ module Header
     end
 
     # contents of the <title> in html <head>
+    #
+    # The doc title is plain text — the browser tab doesn't render
+    # HTML — so any markup in the source string surfaces literally
+    # ("_Russula_" or "<i>Russula</i>"). We textilize first so
+    # textile-source markers (`_x_`, `*x*`, etc.) get converted to
+    # HTML, then strip both the resulting tags AND any already-
+    # rendered HTML (e.g. an `.t.small_author` upstream). Finally
+    # decode HTML entities (`&amp;` → `&`).
     def title_tag_contents(title, action: controller.action_name)
       if title.present?
-        title.strip_html.unescape_html # removes tags and special chars
+        title.to_s.t.strip_html.unescape_html
       else
         action.tr("_", " ").titleize
       end
--- a/app/helpers/header/title_helper.rb
+++ b/app/helpers/header/title_helper.rb
@@ -14,9 +14,20 @@
 #
 module Header
   module TitleHelper
-    def add_show_title(string, object)
+    # Per-model `page_title(user)` returns the rich/textile string
+    # shown in the visible page heading; `document_title` returns the
+    # plain string for the browser tab `<title>`. The doc title
+    # renders as plain text, so we keep textile/HTML out of it. Models
+    # without those methods fall back to a localized type-tag label.
+    #
+    # `user:` lets the page heading apply user-specific naming prefs
+    # (e.g. hide_authors, deprecated-with-preferred-synonym link
+    # wrapping). Document title ignores user — the tab text shouldn't
+    # vary by viewer.
+    def add_show_title(object, user: nil)
       add_page_title(
-        show_page_title(string, object), show_document_title(string, object)
+        show_page_title(page_title_for(object, user), object),
+        show_document_title(document_title_for(object), object)
       )
     end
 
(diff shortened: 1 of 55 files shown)


nimmolo:
TODO: 
- Is there already an MO method or model method that produces a title without textile, that we can pass to the document title method? Likely
- The page templates are calling `add_show_title(title_string, object)` but that calls `add_page_title(title, document_title)` which accepts two arguments, the second should be the non-textile string, which is often cached on the record. 
- `add_show_title` (and `add_edit_title`)'s signature may need to accept two strings plus the object, not just one. because the accessor for the plain text name may vary per model (check this!). Then it should pass the second plain text string to "show document title". The add edit title prolly also needs to do this. 
- Would be slicker to standardize the attribute names for both. Or maybe the titling helpers file gets a lookup of where columns/methods to find both title versions, for each model, so we only have to pass the `object` in to any `add_*_title` method. 
- Maybe also make `strip_textile` a String extension.

mo-nathan:
Took a while for me to reproduce this.  I had to:
1. Got to an observation that I own.
2. Click the "Edit" icon.
3. Open the JS console (Cmd+Opt+J)
4. Type in `document.title` and hit return.

On current main this produces something like:
`Mushroom Observer: Edit Observation 641530: **__Amanita__** **__ristichii__** Tulloss`

Being on an Edit page is crucial.  Also the weird formatting isn't visible directly in Chrome.

nimmolo:
Here it is in Safari:

Screenshot 2026-05-21 at 6 30 31 PM

PR #4319, opened 2026-05-22
Fix map cluster Show All returning nothing for shared-location observations

### LINKED ISSUE #4318: New map UI failing on obseravtions that share a location
Currently, if you go to this page:
http://localhost:3000/observations/map?q%5Bmodel%5D=Observation&q%5Border_by%5D=thumbnail_quality&q%5Bprojects%5D%5B%5D=327

The is a yellow dot with a "2" in it in central Pennsylvania.  If you click on it you see that there are two observations of Pseudofistulina radicata.

Image

However, if you click on "Show All", it goes to:

http://localhost:3000/observations?q%5Bmodel%5D=Observation&q%5Border_by%5D=thumbnail_quality&q%5Bprojects%5D%5B%5D=327&q%5Bin_box%5D%5Bnorth%5D=41.1973&q%5Bin_box%5D%5Bsouth%5D=41.1973&q%5Bin_box%5D%5Beast%5D=-78.5096&q%5Bin_box%5D%5Bwest%5D=-78.5096

which shows nothing.

Image

The issue appears to be related to the north param being identical to the south param and the east param being identical to the west param.  If you add a bit of buffer:

http://localhost:3000/observations?q%5Bmodel%5D=Observation&q%5Border_by%5D=thumbnail_quality&q%5Bprojects%5D%5B%5D=327&q%5Bin_box%5D%5Bnorth%5D=41.1974&q%5Bin_box%5D%5Bsouth%5D=41.1972&q%5Bin_box%5D%5Beast%5D=-78.5095&q%5Bin_box%5D%5Bwest%5D=-78.5097

the two observtions show up.

Image

mo-nathan:
## Problem

On the new JS map UI, clicking a cluster of observations that share one location and then clicking **Show All** (or **Map All**) lands on an empty page (#4318). Two *Pseudofistulina radicata* observations at "Parker Dam State Park, PA" produce a "2" badge, but Show All returns nothing. The generated URL has `north == south` and `east == west` — a zero-area box.

## Root cause

The cluster popup builds its Show All / Map All links from the union of the member marker boxes (`computeClusterOutlineBox`). When every member sits at one point — here because location 22185 is itself a point-geometry location (its bounding box equals its center) — that union collapses to a zero-area box, which `clusterQueryUrl` passed into `q[in_box]` with no buffer. Two issues compounded:

1. `Mappable::MapSet` rounded emitted coordinates to **4 decimals**, so the stored `location_lat` `41.1972999573` became `41.1973` — nudging the box edge *past* the observation's own coordinate.
2. With a zero-area box, the server's `in_box` then required the full-precision coordinate to fall exactly on that rounded edge (`location_lat >= south` with `south == 41.1973 > 41.1972999573` → false), so it matched nothing.

The server-side popup helper (`map_popup_helper.rb`'s `mapset_box_params`) already buffers its query box to avoid exactly this; the new JS path never ported that.

## Fix

- **Emit 6-decimal coordinates** (`COORD_PRECISION`) from `Mappable::MapSet`. This is headroom, not a precision claim — it shrinks the worst-case emission rounding error so a small buffer reliably clears it.
- **Buffer the `in_box` edges by ±0.00001°** in `clusterQueryUrl`, clamped to valid lat/lng ranges. For a point cluster this expands the query box to roughly an **8 ft square (~53 ft²)** — about as tight as GPS data supports — instead of the **~12 acres** a ±0.001° buffer would cover. Covers Show All, Map All, and the per-species row links (all route through `clusterQueryUrl`).

## Verification

- `map_set` + `collapsible_map` + `map` component + `maps_controller` suites pass; added two `MapSetTest` cases asserting 6-decimal emission on the point and box paths.
- RuboCop clean on both Ruby files.
- Against the real data (project 327 / location 22185): the new buffered box returns both observations (`[402121, 402125]`); the old zero-area box returned `[]`.

## Testing note

There is no JS unit-test runner in the repo, and the only map system test stubs around Google Maps, so the JS buffering math is not covered by an automated test. The server-side behavior it relies on is verified above. Recommend a manual browser check against the issue's URL before merge.

Fixes #4318

### DIFF
--- a/app/classes/mappable/map_set.rb
+++ b/app/classes/mappable/map_set.rb
@@ -39,6 +39,12 @@ module Mappable
     MIXED_COLOR     = "#C69B71" # observations in different consensus bands
     LOCATION_ONLY_COLOR = "#3B79CC" # bootstrap primary; no obs to classify
 
+    # Decimal places kept when emitting marker coordinates to the JS map.
+    # At 6 places the worst-case rounding error is ~5.5 cm, well below GPS
+    # accuracy, so the cluster popup's Show All / Map All box no longer
+    # rounds an obs's own coordinate outside the box it builds (#4318).
+    COORD_PRECISION = 6
+
     attr_reader :north, :south, :east, :west, :is_point, :is_box,
                 :north_east, :south_east, :south_west, :north_west, :lat, :lng,
                 :north_south_distance, :east_west_distance, :center, :edges
@@ -161,8 +167,8 @@ module Mappable
     end
 
     def update_extents_with_point(loc)
-      lat = loc.lat.to_f.round(4)
-      lng = loc.lng.to_f.round(4)
+      lat = loc.lat.to_f.round(COORD_PRECISION)
+      lng = loc.lng.to_f.round(COORD_PRECISION)
       if @north
         @north = lat if lat > @north
         @south = lat if lat < @south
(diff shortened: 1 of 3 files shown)


JoeCohen (approved):
Looks good. Thanks for the fast fix.
I did some manual testing and didn't see any problems.
Is there something specific you'd like me to try?

PR #4326, opened 2026-05-23
Bundle update with Rails ~> 7.0

### LINKED ISSUE #4325: Relax Rails component version pin
We are currently pinned at  `~> 7.2.2.1`.
I want to update to at least 7.3 to address reported security issues.
Therefore relax it `~> 7.0`.

(I think i pinned it to `~> 7.2.2.1` because that was our version when Dependabot tried to upgrade to 8.x.)

JoeCohen:
Resolves #4325


JoeCohen:
The update to Rack causes test Errors.
CC initially said:
> Now I have the full picture.

`store_query_in_session`(query) at [queries.rb:201](vscode-webview://0i7ijagn5nnse4qoc3c6a3mso4ctegpn0nd7j9f1c8rp3qp7hql0/app/controllers/application_controller/queries.rb#L201) assumes query is a Query object with an .id method. But somewhere upstream, query is already an integer ID.
>
> The guard query.save unless query.id is trying to say "persist if not yet saved" — but when query is an Integer, calling .id on it raises NoMethodError (Ruby no longer aliases Object#id to object_id).
>
> Suggested fix
>
> Add a type guard at the top of store_query_in_session:
```ruby
def store_query_in_session(query)
  query = Query.find(query) if query.is_a?(Integer)
  query.save unless query.id
  session[:query_record] = query.id
  session[:search_type] = query.search_type
end
```
> Or, if an integer ID is a fully valid input and there's no need to save/re-fetch, handle both types:
```ruby
def store_query_in_session(query)
  if query.is_a?(Integer)
    session[:query_record] = query
  else
    query.save unless query.id
    session[:query_record] = query.id
    session[:search_type] = query.search_type
  end
end
```
> The right fix depends on which branch the caller intends — do you want to say "the caller should never pass a raw ID here (and we should fix the caller)", or "both forms are acceptable inputs"? If the former, the real fix is tracing back through the test to find where the integer is being passed and fixing that call site instead.
---
Then when asked why the Errors started only after the update:

> Now I have enough to give a complete explanation.
>
> Why the errors started after bundle update --all
>
> The Gemfile on main pinned every Rails component to ~> 7.2.2.1. That constraint allows >= 7.2.2.1, < 7.2.3 — a ceiling just below the patch that caused trouble. On this branch, the pin was loosened to ~> 7.0, so bundle update --all was free to resolve to 7.2.3.1 for all Rails components.
>
> That one-patch jump had a second effect: actionpack 7.2.3.1 broadened its Rack constraint from < 3.2 to < 3.3, which allowed Rack to jump from 3.1.21 → 3.2.6 as well. You can see both facts in the lock diff:
```ruby
-      rack (>= 2.2.4, < 3.2)
+      rack (>= 2.2.4, < 3.3)

-    rack (3.1.21)
+    rack (3.2.6)
```
> What actually broke
>
> The crash is at [queries.rb:201](vscode-webview://0i7ijagn5nnse4qoc3c6a3mso4ctegpn0nd7j9f1c8rp3qp7hql0/app/controllers/application_controller/queries.rb#L201):
```ruby
def store_query_in_session(query)
  query.save unless query.id  # ← query is an Integer here
  ...
end
```
> query.id fails because query is already an Integer (a raw QueryRecord ID), not a Query object. This is the mechanism; the triggering change is almost certainly in Rack 3.2's query-string parsing.
>
> The "Show Observations at this Location" link is built by [locations_helper.rb:109](vscode-webview://0i7ijagn5nnse4qoc3c6a3mso4ctegpn0nd7j9f1c8rp3qp7hql0/app/helpers/tabs/locations_helper.rb#L109):
```ruby
query = Query.lookup(:Observation, locations: location.id)
add_q_param(observations_path, query)
```

>query.q_param returns { model: :Observation, locations: <integer_id> }, which add_q_param serializes into ?q[model]=Observation&q[locations]=123. When Rails parses that on the next request, params[:q] is an ActionController::Parameters hash, and [query_from_q_param_hash](vscode-webview://0i7ijagn5nnse4qoc3c6a3mso4ctegpn0nd7j9f1c8rp3qp7hql0/app/controllers/application_controller/queries.rb#L244) reconstructs the query via:
```ruby
Query.lookup(q_param[:model].to_sym,
             **q_param.except(:model).to_unsafe_hash)
```

> Rack 3.2 introduced changes to how nested query-string parameters are parsed (related to the security hardening in cgi 0.5.1, which was split out as an explicit dependency in 7.2.3.1). The likely result is that the integer value "123" is now parsed or coerced differently — either it comes through as an actual Integer where code previously expected a String, or the nested key locations leaks to a top-level param — causing a different code path to hand a raw Integer to store_query_in_session.
>
> Summary in one line: loosening ~> 7.2.2.1 → ~> 7.0 unlocked Rails 7.2.3.1 + Rack 3.2, and Rack 3.2 changed query-string parsing in a way that breaks the q-param round-trip, ultimately passing a raw integer ID where a Query object is expected.

---

Joe: You earlier suggested two alternate fixes for the bug. In light of what you found out abour rack, which fix do  you recommend?

> Neither fix targets the right place.
>
> Both patches treat store_query_in_session as the problem, but it's just where the symptom surfaces. That method has always expected a Query object — and before Rack 3.2, it always received one. The Rack 3.2 change broke something upstream in the q-param round-trip; patching the receiver just papers over it.
>
> Recommended fix: repair the source of the Integer, most likely in query_from_q_param_hash at [queries.rb:244](vscode-webview://0i7ijagn5nnse4qoc3c6a3mso4ctegpn0nd7j9f1c8rp3qp7hql0/app/controllers/application_controller/queries.rb#L244). That method reconstructs a Query from the q hash param — if Rack 3.2 is now coercing the numeric string "123" into an actual Integer 123, the reconstructed params hash will have integer values instead of strings, which could cascade into a raw ID leaking out of that method instead of a Query object.
> 
> The place to start is adding a log or debugger breakpoint in query_from_q_param_hash to inspect what q_param.except(:model).to_unsafe_hash actually contains under Rack 3.2, then either:
>
> Stringify or symbolize the values before passing to Query.lookup, or
Adjust query_from_q_param_hash to handle the type change explicitly
That said, if you need a temporary guard while you investigate, Fix 1 (fetch the Query object from the Integer) is the safer of the two — it preserves search_type storage and doesn't silently change behavior. Fix 2 drops search_type for the Integer path, which could introduce a subtler bug.

nimmolo (commented):
Chatty Claude said:

> The right fix depends on which branch the caller intends — do you want to say "the caller should never pass a raw ID here (and we should fix the caller)", or "both forms are acceptable inputs"? If the former, the real fix is tracing back through the test to find where the integer is being passed and fixing that call site instead.

If i recall correctly, the intent was "both forms are acceptable". This is where we wanted to accept either a query record ID (the old alphabetized QueryRecord param) or a Query object.

End of story. The simple fix seems correct:

```ruby
def store_query_in_session(query)
  if query.is_a?(Integer)
    session[:query_record] = query
  else
    query.save unless query.id
    session[:query_record] = query.id
    session[:search_type] = query.search_type
  end
end
```

JoeCohen:
@nimmolo Here is CC's response. What do you think?

> Disagree. The "both forms" intent applies to query_from_q_param — that method deliberately handles both an alphabetized string ID (old permalinks) and a hash (current format). The Integer reaching store_query_in_session was never an intended input; it was a bug artifact.
> 
> Two problems with the proposed guard:
> 
> Missing search_type — the Integer branch skips session[:search_type] = query.search_type, leaving stale search type in the session from a prior request.
> 
> Masks future regressions — if something upstream breaks again and passes an Integer, the guard silently swallows it. The test failures we just saw are the correct signal; we want them to fail loudly.
> 
> The .symbolize_keys fix repairs the actual cause. The guard would be defensive coding against a bug that no longer exists.

Issue #4331, opened 2026-05-24
Phlex move initiative: relocate single-use forms from app/components/ to app/views/


nimmolo:
## Why these PRs

This issue tracks an ongoing initiative spawning many PRs over the next few weeks. Posting it here so reviewers have context for the wave of small mechanical refactors.

We're separating Phlex classes into two clear locations based on how they're actually used:

- **Reusable components** (rendered from more than one controller) stay in `app/components/` under the flat `Components::<Name>` namespace.
- **Single-use views** (only rendered for one controller's pages, including its own turbo_stream responses) move to `app/views/controllers/<controller>/<name>.rb` under `Views::Controllers::<Controller>::<Name>` — a deep namespace that mirrors the controller tree.

Both inherit from `Phlex::HTML` via `Components::Base`. The split is for organization and intent, not capability. Codified in `.claude/rules/phlex_conversions.md` (PR #4324).

## Why it matters

- **`app/views/` stays organized like the ERB tree it replaces.** A Phlex single-use view sits next to the action templates it serves — the same place a partial used to live. Reviewers find it where they expect.
- **`app/components/` stays small and meaningful.** It contains real building blocks, not page-specific fragments. Putting a page-specific fragment there implies \"reuse me elsewhere\" — an invitation that rarely turns out well.
- **Eventually action templates themselves migrate to Phlex.** When `index.html.erb` becomes `index.rb`, it joins the same `app/views/controllers/<controller>/` directory as the partial-equivalents already there. The structure scales.

## What the move looks like

For each form being moved:

1. Create `app/views/controllers/<path>/<name>.rb` with the new namespace.
2. Delete `app/components/<old_name>.rb`.
3. Update all call sites (grep `Components::<ClassName>`).
4. Move + rename the test from `test/components/` to `test/views/controllers/<path>/<name>_test.rb`.
5. Run the controller's full test suite + integration/system tests.
6. Lint clean; per-file coverage at 100%.

PR #4327 (articles), #4328 (comments), #4329 (sequences), and #4323 (images/licenses) are early examples of the pattern; #4321 (account/api_keys) was the exemplar that established it.

## Scope

A grep-based audit of `app/components/` (verify per-component before acting — the rough numbers are imprecise):

- **~152 total** Phlex components (excluding base/infrastructure).
[...]

PR #4345, opened 2026-05-24
occurrences: move + consolidate OccurrenceForm + OccurrenceEditForm; fix 2 N+1s

nimmolo:
Per #4331. Moves + consolidates the occurrence forms; system test exposed two pre-existing N+1s (fixed in scope).

## Form consolidation

`Components::OccurrenceForm` + `Components::OccurrenceEditForm` → single `Views::Controllers::Occurrences::Form` branching on `model.persisted?`. Mode differences (kwargs, stimulus controllers, fallback radio, inline primary detail, HTTP method) preserved exactly.

## Project-confirmation modal — three changes

- **Skip button** added between Cancel and Add All. Submits `occurrence_projects[resolution]=skip`; controllers already no-op on anything other than `add_all`, so this just exposes the existing branch. Whether users *should* be able to skip is a product call — pending #4349.
- **`FormObject::OccurrenceProjects`** replaces the decorative `Occurrence.new` the form was passing to Superform. Real fields now bind under `occurrence_projects[*]`. `OccurrencesController#create` reads from either `:occurrence` (new-form POST) or `:occurrence_projects` (modal repost) via one helper.
- **Custom `resolve_projects` → CRUD.** `resources :occurrences do resource :projects, only: :update, controller: "occurrences/projects" end` → `PATCH /occurrences/:occurrence_id/projects` → `Occurrences::ProjectsController#update`. In the spirit of MO's (mostly) pure-CRUD transition: no custom-route plumbing, cleaner reads.
- Component renamed `OccurrenceResolveForm` → `OccurrenceProjectsForm` to match the FormObject. Stays in `app/components/` (also rendered from field_slips).

## N+1 fixes (exposed by new edit-form system test)

- `naming_consensus#load_occurrence_namings`: added `:observation` to Naming preload.
- `occurrences_controller/show#ordered_observations`: added `:rss_log`.

## Try it locally

- New form (Nathan's obs with no occurrence): `/occurrences/new?observation_id=641129`
- Edit form (multi-obs occurrences Nathan owns): `/occurrences/3184/edit` (2 members), `/occurrences/2630/edit` (3 members)
- Project gap modal appears when selected obs span projects not shared by all — check a candidate that's in a project the source isn't.

## Tests / coverage

24 view + 74 controller (incl. new `Occurrences::ProjectsControllerTest`) + 8 component + 6 FormObject unit + 5 system tests pass. Lint clean. All touched Ruby files at 100% (closed a pre-existing gap in `naming_consensus#clean_votes` per fix-on-touch policy).


JoeCohen (dismissed):

JoeCohen (changes requested):
After focusing on #4349, I don't see a use case for Skip.
An Occurrence should be in all Projects associated with any of the Occurrence's Observations.
I can't think of a use case for anything else. 
An I spoke to @mo-nathan about this; he and I see it the same way.

PR #4359, opened 2026-05-25
ImageEditForm: fix license select showing ID instead of name

### LINKED ISSUE #4358: Edit Image License Display
The Edit Image, license menu shows numbers (enum's of the licenses?) instead of License names.
Image

JoeCohen:
Fixes #4358

### DIFF
--- a/app/components/image_edit_form.rb
+++ b/app/components/image_edit_form.rb
@@ -49,7 +49,7 @@ class Components::ImageEditForm < Components::ApplicationForm
     date_field(:when, inline: true,
                       label: "#{:form_images_when_taken.l}:",
                       help: :form_images_when_help.t)
-    select_field(:license_id, @licenses,
+    select_field(:license_id, superform_license_options,
                  label: "#{:LICENSE.t}:",
                  help: :form_images_license_help.t)
     # Two-paragraph help: notes-specific + textile-syntax help.
@@ -107,6 +107,10 @@ class Components::ImageEditForm < Components::ApplicationForm
   # only the image owner OR a member of the project can toggle.
   def cannot_modify_project?(project)
     model.user_id != @user.id && !project.member?(@user)
+  end
+
+  def superform_license_options
+    @licenses.map { |display, id| [id, display] }
   end
 
   def project_checked?(project_id)
--- a/test/components/image_edit_form_test.rb
+++ b/test/components/image_edit_form_test.rb
@@ -34,6 +34,18 @@ class ImageEditFormTest < ComponentTestCase
     assert_html(html, "select[name='image[when(2i)]']")
   end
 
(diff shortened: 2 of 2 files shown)


JoeCohen:
@nimmolo: Should this fix be generalized? 
When CC first suggested the fix in this PR, I asked if things should be DRYed up. CC said:

> The question is whether it makes sense to DRY up the two components through a shared License method.
> 
> Pros:
> 
> - Removes three independent swap implementations (FormImageFields, DescriptionForm, and the one we're adding to ImageEditForm) from the view layer, where the concern doesn't belong
> - The model already owns this data; the order it's produced in is a model-level decision
> - Adding a future caller can't silently get the wrong order by forgetting to swap
> 
> Cons:
> 
> - available_ids_and_names is only meaningful given Superform's specific convention, which is an application-layer detail leaking into the model
> - The name doesn't convey the ordering rationale — a future reader sees two near-identical methods and has to dig to understand why
> - The better fix might be in SelectField itself: if it accepted [text, value] like Rails helpers do, there'd be nothing to swap anywhere
> 
> My take: The real smell is that Superform's non-standard [value, text] convention forces every caller to remember to swap. The ideal fix is making SelectField accept the Rails-standard [text, value] order. Short of that, a method on License is a reasonable DRY-up, but it trades one leaky concern for another. Adding it only to License also leaves DescriptionForm and the bulk-updater still doing manual swaps.
> 
> For this PR I'd recommend the minimal fix: just swap the pairs in ImageEditForm (matching what FormImageFields already does), leaving the broader consolidation for a follow-up that fixes all callers at once — or ideally fixes SelectField.
>

nimmolo:
@JoeCohen thanks for fixing this. 

I agree about fixing SelectField and i opened an issue about it on superform's GH repo weeks ago. Haven't heard back from them, although they did eventually merge some of my other PRs.

I think we should fix our version, though. It's a gotcha every time. I can do that PR.

app/components/image_edit_form.rb
```diff
@@ -109,6 +109,10 @@ def cannot_modify_project?(project)
     model.user_id != @user.id && !project.member?(@user)
   end
 
+  def superform_license_options
+    @licenses.map { |display, id| [id, display] }
+  end
+
```
nimmolo, line 115: It seems like the controller should just build `@licenses` in the reverse order, unless there's another user of that ivar.

On the other hand, this can just be reverted if we fix SelectField.

nimmolo (changes requested):

nimmolo:
Thanks for digging into this.
I agreed with CC and did a PR #4364 to fix `SelectField` that should replace this PR. We don't have to flip `@options` arrays anymore.  

Also opening a PR on Superform to the same effect.

PR #4447, opened 2026-06-05
Description#put_together_name: fix source_type ||= local-shadow bug

nimmolo:
## Summary

`Description#put_together_name` had a Ruby parser quirk that quietly flattened every description's title to the public-source format, regardless of the record's actual `source_type`.

The line was `source_type ||= :public`. Ruby parses `||=` as `source_type = source_type || :public`. On the left-hand side, the parser sees an assignment target and declares a fresh local variable named `source_type` — which shadows the AR `source_type` reader for the rest of the method. The RHS then reads that newly-declared local (always `nil`), `||` picks `:public`, and the local is bound to `:public` from there on.

The net effect: the method computed `tag = :"description_part_title_public"` (or `_with_text` if `source_name` was present) every single time. A project-source description rendered as "Draft for Project Name by Mary Newbie" only because `source_name` was set; a project-source description with no `source_name`, or any user-source description with no `source_name`, rendered as "Public Description". User-source descriptions where the user got nil'd out rendered as "Public Description of …" instead of "?'s Description of …".

The fix assigns to a differently-named local (`type`) so the RHS stays an actual method call on the record:

```ruby
type = source_type.presence || :public
tag = :"description_#{full_or_part}_title_#{type}"
```

`.presence` keeps the original "fall back to public when the attribute is nil" intent that motivated the original (buggy) commit.

## Locale tags

With the fix in place, partial-title rendering now uses the per-source-type tag instead of always landing on `public`. Three tags were missing — pre-fix they were unreachable, so no one ever noticed:

- `description_part_title_project: "[USER]'s Draft"`
- `description_part_title_source: "[:DESCRIPTION]"`
- `description_part_title_foreign: "[:DESCRIPTION]"`

These are the no-text variants for descriptions where `source_type` is set but `source_name` is blank. Fixtures cover this: `bolete_project_private_location_desc` is `source_type: :project` with no `source_name`, and was rendering as "Public Description" before the fix.

## Tests

- `test_user_sourced_description_with_unknown_user` was previously reinforcing the bug — it asserted the buggy "Public Description of …" output for a user-source description with `user_id: nil`. Updated to assert the correct `?'s …` user-source format.
- `test_partial_text_name_public_source_with_text` updates `peltigera_alt_desc.source_name` and asserts the source_name renders verbatim.
- New tests cover each `source_type`'s partial / full title shape (public / user / project / source — with and without `source_name`).
- New test `test_nil_source_type_falls_back_to_public_format` calls `update_columns(source_type: nil)` and asserts the public-format fallback still kicks in.

## Test plan

- [x] `bin/rails test test/models/description_test.rb`
- [x] `bin/rails test test/models/ test/controllers/names/descriptions/ test/controllers/locations/descriptions/ test/integration/capybara/name_descriptions_integration_test.rb` (859 tests, 0 failures)
- [x] `bundle exec rubocop app/models/description.rb test/models/description_test.rb`

### DIFF
--- a/app/models/description.rb
+++ b/app/models/description.rb
@@ -535,8 +535,11 @@ class Description < AbstractModel
   # include the title of the parent object), in plain text.  [I'm not sure
   # I like this here.  It might violate MVC a bit too flagrantly... -JPH]
   def put_together_name(full_or_part)
-    source_type ||= :public
-    tag = :"description_#{full_or_part}_title_#{source_type}"
+    # NOT `source_type ||= :public` — Ruby would treat the LHS as a
+    # new local that shadows the AR reader for the rest of the
+    # method, flattening every description to the public title format.
+    type = source_type.presence || :public
+    tag = :"description_#{full_or_part}_title_#{type}"
     user_name = begin
                   user.legal_name
                 rescue StandardError
--- a/config/locales/en.txt
+++ b/config/locales/en.txt
@@ -1018,6 +1018,9 @@
   # descriptions in the show_name page.
   description_part_title_public: Public [:DESCRIPTION]
   description_part_title_user: "[USER]'s [:DESCRIPTION]"
+  description_part_title_project: "[USER]'s Draft"
+  description_part_title_source: "[:DESCRIPTION]"
+  description_part_title_foreign: "[:DESCRIPTION]"
 
   # Name/location description types. These are used when listing the available
   # descriptions in the show_name page, but this time the optional text field was
--- a/test/models/description_test.rb
+++ b/test/models/description_test.rb
@@ -204,9 +204,114 @@ class DescriptionTest < UnitTestCase
     desc = name_descriptions(:peltigera_user_desc)
     desc.update(user_id: nil)
 
-    assert(desc.text_name.start_with?("Public"),
-           "text_name of user sourced Description with unknown user should " \
-           "start_with Public ")
+    # User-source description with unknown user: the user-source
+    # title format still applies (the description's
+    # `source_type` is "user" regardless of whether the user
+    # record exists), but the `[USER]` placeholder falls back to
+    # the "?" sentinel from `put_together_name`'s rescue branch.
+    # Prior to the source_type-shadow bug fix this asserted
+    # "Public" because the buggy local-variable shadow made every
+    # description render with the public-source title format.
+    assert(desc.text_name.start_with?("?'s"),
+           "text_name of user-sourced description with unknown user " \
+           "should start with the user-source format and an `?` " \
+           "placeholder — got #{desc.text_name.inspect}")
+  end
+
+  # ---- source_type-specific title formats ------------------------
+  #
+  # Each `Description` source_type maps to its own i18n key — see
+  # `description_{full,part}_title_<type>` in `en.txt`. The
+  # `put_together_name` shadow bug masked these differences for
+  # years (every description rendered with the `_public` format
+  # regardless of its actual `source_type`); pin each format
+  # explicitly so regressions don't silently re-flatten everything
+  # back to "Public" again.
+
+  # `description_part_title_public_with_text: "[TEXT]"`. A public
+  # description with a `source_name` just renders the source_name
+  # verbatim.
+  def test_partial_text_name_public_source_with_text
+    desc = name_descriptions(:peltigera_alt_desc) # source_type :public
+    desc.update(source_name: "Wikipedia")
+    assert_equal("public", desc.source_type)
+
+    assert_equal("Wikipedia", desc.partial_text_name)
+  end
+
+  # `description_part_title_user_with_text: "[TEXT] by [USER]"`.
+  def test_partial_text_name_user_source_with_text
+    desc = name_descriptions(:peltigera_user_desc) # user mary
+    desc.update(source_name: "Mary's Take")
+
+    assert_equal("user", desc.source_type)
+    assert_equal("Mary's Take by Mary Newbie", desc.partial_text_name)
+  end
+
+  # `description_part_title_user: "[USER]'s [:DESCRIPTION]"` (no
+  # text variant).
+  def test_partial_text_name_user_source_without_text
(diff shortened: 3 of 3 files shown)


JoeCohen:
Still getting deprecation warnings and Skipped tests
```ruby
joe@iMac mushroom-observer % rails t
Running 6050 tests in parallel using 12 processes
Started with run options --seed 2409

DEPRECATED: Use assert_nil if expecting nil. This will fail in Minitest 6.                                                                                    ] 14% Time: 00:00:16,  ETA: 00:01:38
DEPRECATED: Use assert_nil if expecting nil. This will fail in Minitest 6.=======================                                                             ] 58% Time: 00:00:37,  ETA: 00:00:26
WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"===========                                          ] 71% Time: 00:00:44,  ETA: 00:00:18

 SKIP Views::Controllers::Observations::Show::AssociatedObservationsPanelTest#test_siblings_render_with_occurrence_link (49.97s)
        Need an occurrence with multiple observations
        test/views/controllers/observations/show/associated_observations_panel_test.rb:29:in 'Views::Controllers::Observations::Show::AssociatedObservationsPanelTest#test_siblings_render_with_occurrence_link'

 SKIP Views::Controllers::Observations::Show::SpeciesListsPanelTest#test_no_species_lists_renders_panel_chrome_only (58.10s)
        Need obs without species_lists
        test/views/controllers/observations/show/species_lists_panel_test.rb:15:in 'Views::Controllers::Observations::Show::SpeciesListsPanelTest#test_no_species_lists_renders_panel_chrome_only'
```

JoeCohen:
## Manual Tests for PR 4447 — `Description#put_together_name` source_type fix

Before this fix, every description title rendered as "Public Description [of …]"
regardless of source_type. These tests verify each source_type now renders its
own format.

---

### 1. Public description (regression — should be unchanged)

1. Navigate to any name, e.g. `/names/1` (or search for "Agaricus campestris").
2. Click the default public description tab / link.
3. **Expect:** The description title reads something like **"Public Description of Agaricus campestris"** (full) or **"Public Description"** (tab/partial).

---

### 2. User-source description

1. Log in as a user who has created a User-source description, or create one:
   - Go to any name → "Create Description" → set Source Type = **User** → leave
     Source Name blank → save.
2. View the name page with that description active (or navigate to the description directly).
3. **Expect (no source name):** Title reads **"[Username]'s Description of [Name]"** (full) or **"[Username]'s Description"** (partial).
4. Edit the description, set Source Name = "My custom title" → save.
5. **Expect (with source name):** Partial title reads **"My custom title by [Username]"**; full title includes "of [Name]".

Before the fix, step 3 showed "Public Description of …" — that was the bug.

---

### 3. Project-source description (no source name)

1. Go to a project that has descriptions attached (e.g., a project with location
   or name descriptions), or create one:
   - Navigate to any name → "Create Description" → Source Type = **Project** →
     leave Source Name blank → associate with a project → save.
2. View the description.
3. **Expect (no source name):** Partial title reads **"[Username]'s Draft"**.

Before the fix this showed "Public Description".

---

### 4. Project-source description (with source name)

Same as above but set Source Name = "Boletes of North America" (or similar).

- **Expect:** Partial title reads **"Draft for Boletes of North America by [Username]"**;
  full title reads **"Draft of [Name] for Boletes of North America by [Username]"**.

---

### 5. Source-type description (third-party source)

1. Create a description → Source Type = **Source** → Source Name = "MycoBank" → save.
2. View the description.
3. **Expect (with source name):** Partial title reads **"MycoBank"** (or "Description From MycoBank"); full title reads **"Description of [Name] from MycoBank"**.
4. Repeat with Source Name blank.
5. **Expect (no source name):** Partial title reads **"Description"** (the `:DESCRIPTION` locale key).

---

### Quick smoke test path (read-only, no account needed)

- Browse to a name with multiple descriptions of different source types (the
  description tabs in the page header show partial titles). Verify the tabs are
  labeled with the correct owner/source rather than all reading "Public Description".

JoeCohen:
Cannot create a name description associated with a project using CC's suggested Manual Test #3 (the test might be nonsense). 
No "Source Type" option available when I do this:

3. Project-source description (no source name)
Go to a project that has descriptions attached (e.g., a project with location
or name descriptions), or **create one:
Navigate to any name → "Create Description" → Source Type = Project →**
leave Source Name blank → associate with a project → save.

nimmolo:
@JoeCohen I don't know how that's supposed to work, but i don't think this PR changes the source select. (My orevious Phlex Descriptions PR did though). 

We may need @mo-nathan input.

nimmolo:
CC reply for @JoeCohen :

Good catch — that scenario can't be exercised the way I wrote it. My instructions for Test #3 were wrong. Non-admin users don't see "Project" in the Source Type dropdown:

- `Description::BASIC_SOURCE_TYPES = ["public", "source", "user"]` (model). That's the dropdown a regular user gets on the new-description form.
- "project" and "foreign" are admin-only (added when `in_admin_mode?` is true).
- The intended path to a project-source description is via the **"Create New Draft For:"** panel on the name show page — each linked project goes to `new_name_description_path(@name.id, project: <id>, source: :project)`, which the controller (`initialize_description_source`) handles by setting `source_type = "project"`, `source_name = <project title>`, and the project association directly. The dropdown is skipped entirely.

So the corrected Test #3:

> 3. Project-source description:
>    Navigate to a name page where you're a member of a project that doesn't already have a description for that name. Scroll to the **Create New Draft For:** panel → click the project name → fill in description fields → save. Confirm the resulting description title format (uses `:description_full_title_project` / `_part_title_project` translations, e.g. "Project description from \<Project Name\>" rather than the public-source title).

The actual PR fix (`source_type ||= :public` → `type = source_type.presence || :public` in `Description#put_together_name`) still applies to a project description created via that path — the title-format selection still went through the buggy `||=` and would have flattened to the public format. Worth re-testing along the corrected #3 path.

PR #4476, opened 2026-06-08
species_lists/index + listing: use Components::ListGroup

nimmolo:
## Summary

Mirrors the names/index refactor from #4474 for the species_lists index. Replaces hand-rolled `div.list-group` + `div.list-group-item` markup with `Components::ListGroup`, splitting wrapper responsibility from row contents.

## Why

The Index view rolled its own `div(class: "list-group")` and Listing emitted its own outer `div.list-group-item.d-flex.justify-content-between.align-items-start` wrapper — two halves of the same Bootstrap pattern hand-coded across two files. Same shape as the names/index refactor, same fix.

## Changes

- **`Index#render_list`** now renders a single `Components::ListGroup`, iterates `@objects`, and emits each row via `list.item(class: "d-flex justify-content-between align-items-start") { render(Listing.new(...)) }`. The layout classes that used to live on Listing's wrapper now ride on the item the component emits.
- **`Listing#view_template`** drops the outer `<div>` and just emits contents — `render_info` + optional `render_manage_section`. Standalone-render output no longer includes a `list-group-item` wrapper; that's the ListGroup's responsibility.
- **`listing_test.rb`** flips its first assertion from `assert_html(".list-group-item.d-flex")` to `assert_no_html(".list-group-item")`, pinning the new "contents only" contract.

## Scope

Listing has only one other caller in the codebase — the listing_test itself (`grep -rn 'SpeciesLists::Listing\|Listing\.new'`). No production callers outside the index, so no `bare:` prop / fallback wrapper needed. The change is fully contained.

## Test plan

- [x] `bin/rails test test/views/controllers/species_lists/listing_test.rb test/controllers/species_lists_controller_test.rb` — 42 tests, 353 assertions, green
- [x] `bundle exec rubocop` on touched files — clean
- [x] CI: full suite + per-file coverage check via `script/coveralls_pr_check.py 4476`


nimmolo:
Taking copilot's suggestions, but solving them by converting `observations/species_lists/edit.html.erb` to Phlex too.

Issue #4216, opened 2026-05-05
MyCoPortal integration as an import source


mo-nathan:
## Goal

Import fungarium specimen records from MyCoPortal (https://www.mycoportal.org), particularly to populate Rare Fungi Challenge projects.

## Acceptance

- New `sources` row: MyCoPortal.
- Import via MyCoPortal API v2 (public, no auth for reads). Stable record IDs, collector info, GPS coordinates, images with licensing, taxonomic synonym handling.
- Background job creates MO observations with `source_id = MyCoPortal`.
- Same `read_only` rules as iNat (per #4214) and same sync rules (per #4215).
- Naming vote weight: Promising (per #4212).
- Cross-link is one-way (MO → MyCoPortal); no need to push anything back.
- Deduplication via `institutionCode`:
  - Filter out records with `institutionCode: "iNat"` (we get those from iNat directly).
  - Filter out records with `institutionCode: "MUOB"` (those originated on MO).
  - Verified: Multifurca ochricompacta has 49 records; 21 are iNat-sourced, the rest are fungaria (TENN-F, BRIT, …).
- Location mapping uses the existing `within_locations` GPS-based lookup.
  - Refining over-coarse matches via iNat / OSM data is **out of scope** here and tracked in a separate issue.
- Project admins see a "Search MyCoPortal" affordance on the project Update tab. The full search/import surface is built in #4219.
- Only super importers can initiate actual imports.

## Blocked by

- #4209 — sources table
- #4214 — read-only
- #4215 — sync

## Notes

Tracking issue: #4208.
Source: discussion #4106 §8.

mo-nathan:
**Note re #4299 (unified external-relationship model):**

#4299 lands before this issue (Phase 3.5 in #4208) so MyCoPortal is built on the unified `ExternalSite` + `ExternalLink` model from the start. The same code path that creates an `ExternalLink(relationship: "import")` for iNat handles MyCoPortal — no separate `Source` row, no retrofit later. `ExternalSite#observation_url(external_id)` gets a case branch for MyCoPortal returning `.../individual/index.php?occid=<external_id>`.

If we need to distinguish MyCoPortal records that originate elsewhere (e.g. records aggregated from fungaria) from straight mirrors, we can add an `aggregated_from` value to the relationship enum then; the initial set in #4299 is `{ import, export, mirror, cross_reference, unknown }`.

PR #4509, opened 2026-06-11
Weight imported naming votes by source confidence (#4212)

### LINKED ISSUE #4212: Naming vote weights for new imports
## Goal

Set the auto-generated naming(s) on newly imported observations to reflect source confidence and MO's current taxonomy, instead of always voting maximum confidence on a single substituted name. Existing imports are left as-is (a separate decision; not part of this issue).

## Background (current behavior)

The iNat importer creates **one** naming per observation (`Inat::MoObservationBuilder`), voting `Vote::MAXIMUM_VOTE` (3.0) on it. When iNat carries a provisional name, MO *substitutes* the provisional as the sole consensus and drops the iNat Community ID entirely (it survives only as a bare `Name`, with no naming/vote).

### Confidence weight (the "lead" naming)

The name we propose as the consensus — the **lead** — is voted by the iNat obs's signals (`Promising` = 2.0, `Could Be` = 1.0, max = 3.0):

| iNat observation | lead vote |
| --- | --- |
| Has sequence/DNA data (`inat_obs.sequences.present?`) | **3.0** |
| No sequence, has a provisional name | **2.0** Promising |
| No sequence, no provisional, Research Grade | **2.0** Promising |
| No sequence, no provisional, non-Research Grade (`needs_id`/`casual`) | **1.0** Could Be |

### Multi-naming model

The importer now proposes **every relevant name** and lets `calc_consensus` choose the consensus from the votes:

- **Provisional present and different from the Community ID:** create a naming for **both**. The provisional leads (highest vote) **unless** it is deprecated in MO.
- **Any iNat name (Community ID or provisional) deprecated in MO:** also create a naming for its **best preferred synonym** (`Name#best_preferred_synonym`).
- **The lead** is the provisional when present and not deprecated, else the accepted name (a deprecated name's preferred synonym, falling back to itself when it has no approved synonym). The lead carries the weight above; **every other proposal is Could Be (1.0)**. The lead naming is created first so it wins `calc_consensus` ties.
- Provisional **equal** to the Community ID collapses to a single naming.

Scenario covered: the iNat Community ID is an accepted name (e.g. `Leccinum coffeatum`), there is a provisional name (e.g. `Leccinum sp. 'ID01'`), and the provisional is deprecated in favor of the accepted name. The accepted name leads; the deprecated provisional follows at Could Be (and its preferred synonym dedups against the Community ID).

## Out of scope

- **MyCoPortal weighting** ("MyCoPortal → Promising") moves to the MyCoPortal integration issue **#4216** — there is no MyCoPortal import path to attach it to until that lands.
- **Sync re-weighting** (keep the old naming, add a new one on source-ID change) stays with the sync issue **#4215**.
- **Re-weighting already-imported observations.** Tracked separately if/when we decide to.

## Notes

Revised after team review of how provisional names and deprecations appear on iNat (previously this issue proposed a single substituted naming). Two deltas from the original write-up: MyCoPortal moved to #4216, and sequence/DNA presence is a new top-confidence trigger.

Tracking issue: #4208. Source: discussion #4106 §7.

mo-nathan:
## Summary

Sets the auto-generated naming(s) on newly imported observations from source confidence and MO's current taxonomy, instead of always voting maximum on a single substituted name. Closes #4212.

### Confidence weight (the lead naming)

`Inat::MoObservationBuilder` no longer hardcodes `Vote::MAXIMUM_VOTE`. The **lead** (consensus-candidate) naming is voted from the iNat obs's signals:

| iNat observation | lead vote |
| --- | --- |
| Has sequence/DNA data | **3.0** |
| No sequence, has a provisional name | **2.0** Promising |
| No sequence, no provisional, Research Grade | **2.0** Promising |
| No sequence, no provisional, non-Research Grade (`needs_id`/`casual`) | **1.0** Could Be |

### Multi-naming model

Previously, when iNat had a provisional name MO *substituted* it as the sole consensus and dropped the iNat Community ID. Now the importer proposes every relevant name and lets `calc_consensus` choose:

- Provisional present and different from the Community ID → a naming for **both**; the provisional leads **unless** it is deprecated in MO.
- Any iNat name (Community ID or provisional) deprecated in MO → also propose its **best preferred synonym** (`Name#best_preferred_synonym`).
- The lead (provisional if not deprecated, else the accepted synonym; falling back to a deprecated name itself when it has no approved synonym) carries the weight above; every other proposal is **Could Be**. The lead is created first so it wins `calc_consensus` ties.
- Provisional equal to the Community ID collapses to a single naming.

This handles the reviewed case where the Community ID is accepted (`Leccinum coffeatum`), there's a provisional (`Leccinum sp. 'ID01'`), and the provisional is deprecated in favor of the accepted name: the accepted name leads, the deprecated provisional follows at Could Be and dedups against the Community ID.

`proposed_namings(community, provisional, lead_vote)` is a pure function, so every branch is unit-tested directly — including the deprecated-provisional and no-approved-synonym cases that no recorded iNat fixture can produce.

## Scope notes (deltas from the issue as originally written, recorded on #4212)

- **MyCoPortal weighting** moved to #4216 (no MyCoPortal import path yet).
- **Sequence/DNA presence** added as a new top-confidence (3.0) trigger.
- **Multi-naming + deprecation handling** added after team review (the original issue proposed a single substituted naming).

## Test plan

- [x] `test/classes/inat_mo_observation_builder_test.rb` — `naming_vote` (all four weight branches) and `proposed_namings` (single accepted; deprecated Community ID → synonym; provisional leads; deprecated provisional prefers the accepted name; deprecated-without-synonym fallback; provisional == Community ID; reuse of an existing MO provisional name).
- [x] `test/jobs/inat_import_job_test.rb` — per-fixture lead weights; provisional imports now assert two namings (Community ID + provisional).
- [x] `bin/rails test test/jobs/inat_import_job_test.rb test/classes/inat_mo_observation_builder_test.rb test/classes/inat_observation_importer_test.rb test/controllers/inat_imports_controller_test.rb` — 97 runs, 0 failures.
- [x] `app/classes/inat/mo_observation_builder.rb` at 100% line coverage (integration covers all but the deprecated/existing-name branches; the unit file covers those).
- [x] RuboCop clean.


app/classes/inat/mo_observation_builder.rb
```diff
@@ -272,7 +271,25 @@ def adjust_consensus_name_naming
...
+    # everything else (needs_id / casual, no sequence) is Could Be.
+    def naming_vote
+      return Vote::MAXIMUM_VOTE if inat_obs.sequences.present?
+
+      if inat_obs.provisional_name.present? || research_grade?
+        Vote::NEXT_BEST_VOTE # Promising
+      else
+        Vote::MIN_POS_VOTE   # Could Be
+      end
+    end
```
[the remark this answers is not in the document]
mo-nathan: Good call — fixed in 68f68daae. Rather than cache on the builder, I memoized at the source in `Inat::Obs#sequences`, since that's where the field scan happens and both call sites (`naming_vote` and `add_inat_sequences`) already route through it — so one `@sequences ||= …` benefits every caller without the builder needing its own accessor. It's safe because the extraction is pure and `@obs[:ofvs]` isn't mutated during an import, and an empty result is a truthy `[]` so `||=` caches it correctly instead of re-running.

app/classes/inat/mo_observation_builder.rb
```diff
@@ -75,30 +66,66 @@ def collector_attrs
...
 
-    def need_new_prov_name?(parsed_prov_name)
-      Name.where(text_name: parsed_prov_name.text_name).none?
+    # The name proposed as the obs's consensus: the provisional name when
+    # present, else the Community ID, corrected to its preferred synonym when
+    # deprecated in MO. calc_consensus confirms it from the votes, where it
+    # carries the highest weight. (#4212)
+    def lead_name
+      @lead_name ||= preferred(prov_name || community_name)
+    end
```
JoeCohen, line 99: Is there an edge case here that may be too hard to resolve automatically:

- The iNat obs has a provisional name (in a observation field), 
- but the iNat id is legitimately something different.

Example: The provisional name sequence matches a new sequence for an old type; on MO two have not been synonymized.

JoeCohen:
Should the importer prefer (a) a provisional name which lacks a sequence over (b) a non-research grade iNat leading identification?
Example: https://www.inaturalist.org/observations/362781998

JoeCohen:
A nice example how your algorithm picks the right name. https://www.inaturalist.org/observations/344618980
- Research Grade _Mycena pura_
- Provisional Species Name _Mycena sp. 'pura PNW15'_ supported by a sequence
- Comment from Steve mentioning the provisional name, but he does not suggest it as an identification.

app/classes/inat/mo_observation_builder.rb
JoeCohen: This one was a near miss: https://www.inaturalist.org/observations/342013813
- Provisional Species Name `Trichoglossum sp. 'hirsutum-PNW01'` supported by a sequence
- iNat ID is _Trichoglossum hirsutum var. irregulare_ because the provisional matched a brand-new type sequence. 
- It was imported with the registered name only because 4 days ago I manually added the registered name to MO and deprecated the provisional name. If I hadn't done that, the importer would have used the provisonal name. 

This is a hard case and maybe there's nothing the importer can do about it.  (And It's not the end of the world if the importer prefers the provisional.)

**IMO this is not a blocker.**

Out of scope: The ultimate solution could be:
- some job/script which finds these synonyms in MycoMatch (IDK if that's possible maybe @dwmccheyne would know) and updates the MO Name objects;
- periodically checking iNat taxa db for updates
- crawling (I don't think this is available via an API)[ iNat Taxa Changes](https://www.inaturalist.org/taxon_changes?utf8=%E2%9C%93&filters%5Bsplit%5D=0&filters%5Bmerge%5D=0&filters%5Bswap%5D=0&filters%5Bstage%5D=0&filters%5Bdrop%5D=0&filters%5Bchange_group%5D=&filters%5Btaxon_scheme_id%5D=&filters%5Bstatus%5D=&filters%5Bvoted%5D=&filters%5Biconic_taxon_id%5D=47170&filters%5Btaxon_name%5D=&filters%5Btaxon_id%5D=&filters%5Bancestor_taxon_name%5D=&filters%5Bancestor_taxon_id%5D=&filters%5Bsource_id%5D=&filters%5Buser_id%5D=) for updates

JoeCohen: I think this is a real case: https://www.inaturalist.org/observations/196703704
But the importer says it's not licensed. That looks like an out-of-scope bug to me.
I added Issue #4519 and will start working on it.

JoeCohen (approved):
Approve. Code looks good, behavior is great.
The only issues I flagged are IMO non-blockers and/or out of scope.

app/classes/inat/mo_observation_builder.rb
JoeCohen: Above bug is fixed in [Import observations with by-nc-sa license](https://github.com/MushroomObserver/mushroom-observer/pull/4520#top) #4520 (which is merged into main)
If you're seeing incorrect licensed observation estimates, you could merge #4520 into this branch.

mo-nathan:
Yeah, until the data is in MO I don't see a way to resolve the Provisional Name -> Real Name issue.  Having a separate mechanism that monitors these conclusions from other locations would be great.

PR #4539, opened 2026-06-15
rubocop: wire in rubocop-capybara + rubocop-minitest

nimmolo:
## Summary

Wires in the two RuboCop plugins that `bundle exec rubocop` was suggesting based on the project's already-installed gems, then works through the offenses they surface — sweeping the safe ones, hand-fixing the small-count manual ones, restructuring helpers around `Minitest/TestMethodName`, and permanently disabling a few cops whose autocorrect either erases author intent or is loud enough to not pay for itself.

- `Gemfile`: add `rubocop-capybara` + `rubocop-minitest` next to the other `rubocop-*` gems in the `test, :development` group.
- `.rubocop.yml`: list both as plugins.
- `.rubocop_todo.yml`: only the entries for cops still deferred (or permanently disabled with rationale) remain; everything cleared during the PR is stripped.

### Tier 1 — safe-autocorrect mechanical rewrites (12 cops)

Ran `bundle exec rubocop --autocorrect --only "<CopName>"` for each, then deleted the corresponding `.rubocop_todo.yml` entry:

- `Capybara/SpecificFinders`, `Capybara/SpecificActions`
- `Minitest/AssertEmpty`, `Minitest/AssertEqual`, `Minitest/AssertInDelta`, `Minitest/AssertInstanceOf`, `Minitest/AssertKindOf`, `Minitest/AssertMatch`, `Minitest/AssertNil`, `Minitest/AssertPathExists`, `Minitest/AssertRespondTo`, `Minitest/LiteralAsActualArgument`, `Minitest/RefuteFalse` (via the `Rails/RefuteMethods` chain trick — `assert_equal(false, x)` → `refute(x)` → `assert_not(x)`), `Minitest/ReturnInTestMethod`, `Minitest/UnreachableAssertion`
- `Capybara/FindAllFirst` (4 sites across 3 system tests: `all(sel).first` and `find(sel, match: :first)` → `first(sel)` — the cop is labeled "unsafe" because of nil-vs-wait semantic differences, but in all 4 sites the calling code immediately accesses the element, so the wait-and-raise semantic is strictly better than a silent nil).

### Tier 2 — initially swept, **then fully reverted** (3 cops, commit `8883b9ad5`)

`Minitest/AssertIncludes` (152), `Minitest/AssertOperator` (188), `Minitest/AssertPredicate` (399) were originally swept here and then reverted on 2026-06-16. Reasoning: `assert(arr.include?(x))` reads more naturally than `assert_includes(arr, x)`, `assert(x > y)` reads more naturally than `assert_operator(x, :>, y)`, `assert(x.present?)` reads more naturally than `assert_predicate(x, :present?)`. All three now live in `.rubocop_todo.yml` as `Enabled: false` with explanatory comments — see the file for the exact wording.

The audit-before-pushing lesson from that sweep is preserved in [project_rubocop_plugin_cops_followups.md](https://github.com/MushroomObserver/mushroom-observer/blob/main/.claude/projects/...) — short version: `Minitest/AssertIncludes` and `Minitest/AssertOperator` autocorrect introduced 3 real semantic bugs (`Project#member?` is a user-membership predicate the cop misread as Enumerable; `assert(qb[i] = qa[i].subquery_of(model))` was misread as a comparison). Grep on the method side of the autocorrect, not the receiver name.

### Follow-up A — `Minitest/TestMethodName` (23 sites, 11 files)

The cop only flags **public** methods on TestCase subclasses; its autocorrect renames helpers to `test_*` and breaks every call site. Solution: append `private :method_name` after each helper's `end` — the cop respects non-public visibility, and call sites stay unchanged. Helpers patched: `assert_email_generated`, `assert_show_observation`, `assert_show_location`, `assert_api_failed`, `assert_create_species_list`, `assert_edit_species_list`, `init_for_list_checkbox_tests`, `init_for_project_checkbox_tests`, `prove_sorting_links_include_contribution`, the 4 `assert_observation_lat_lng_*` helpers, etc.

### Follow-up B — small-count manual cops (~10 sites)

- **`Minitest/AssertWithExpectedArgument`** caught two real bugs: `inat_obs_test.rb:229` was `assert(2, mock_observation("inocybe")[:tags].length)` — first arg `2` is always truthy, so the assertion never actually checked the count (real length is 3; asserted as such). `string_extensions_test.rb:94` was `assert("<p>a</p>", "a".tp_nodiv)` — same shape, switched to `assert_equal`. `general_extensions.rb:358` was an `assert(nil, msg || …)` flunk idiom, now `flunk(msg || …)`. Remaining sites in `naming_test`, `vote_test`, `sequence_test`, `account_controller_test` are legitimate `assert(predicate, message)` patterns where the message was a method call (the cop's heuristic treats it as a second value arg); wrapped each message with a string-interp prefix (`"Errors: #{…}"`, `"Validation: #{…}"`, `"Save failed: #{…}"`) so the cop sees a string literal.
- **`Minitest/NonPublicTestMethod`** in `search_form_test.rb` (5 tests): the `private` keyword was misplaced *above* 5 length-validator tests, hiding them from MiniTest entirely. Moved `private` below them. Running those tests surfaced a pre-existing bug: `assert_html(html, sel, "msg")` passed a third positional arg `assert_html` doesn't accept — dropped the bogus message.
- **`Minitest/UnspecifiedException`** (2 sites): `assert_raises(<string>)` actually treats the string as the *failure message*, not the expected exception class — the test passes on any exception. Switched to `assert_raises(ActiveRecord::RecordInvalid)` / `assert_raises(RuntimeError)` to match the actual raised type.
- **`Minitest/UselessAssertion`** (2 sites): deleted `donation_test.rb` and `publication_test.rb` — each was a single Rails-scaffolded `def test_truth; assert(true); end` placeholder.

## Inline `# rubocop:disable/enable` directives — intentionally untouched

An earlier revision ran `--auto-gen-config` against the full cop list and let `--autocorrect-all` remove the 244 inline `# rubocop:disable/enable` comments that the regenerated `.rubocop_todo.yml` made redundant. Wrong shape — the inline disables call attention to specific decisions at each call site (`Metrics/CyclomaticComplexity`, `Metrics/AbcSize`, etc.), and a blanket `.rubocop_todo.yml` deferral buries them. This PR regenerates only the plugin-specific entries and leaves the inline directives in place.

## Permanently disabled cops (with reasoning, in `.rubocop_todo.yml`)

- `Minitest/AssertIncludes`, `Minitest/AssertOperator`, `Minitest/AssertPredicate` — see Tier-2 revert above.
- `Minitest/AssertTruthy` — autocorrect rewrites `assert_equal(true, x)` → `assert(x)`, relaxing literal-`true` to any truthy value; the original form documents the author's intent.
- `Minitest/EmptyLineBeforeAssertionMethods` — 5902 offenses, pure cosmetic churn across nearly every test file.
- `Minitest/MultipleAssertions` — 1564 offenses; MO's component-test conventions actively encourage many assertions in a single test rendering once, so threshold-policing doesn't fit.
- `Minitest/RefuteFalse` — same reasoning as `AssertTruthy`: `assert_equal(false, x)` documents that the method returns literal `false`, not any falsy value.

## Drive-by fixes caught in passing

- `parallel_test_config_service_test.rb` (`test_setup_config_files_write_error`): swapped a `FileUtils.chmod(0o444, config_path)` provocation (flaky on CI runners that execute as root) for a `File.stub(:write, raise StandardError)` matching the sibling delete-error test. Was responsible for the -0.006% coveralls drop reported earlier.
- `api2/observations_test.rb` (`test_editing_observation`): killed 6 N+1 re-fetches of the same project / species lists — `Project.find(proj.id).observations.include?(…)` × 4 + `splN.reload.observations.include?(…)` × 4 → one `proj.observations.reload` per state phase. 16 queries → 6.
- `api2_extensions.rb` (`assert_last_observation_correct`): three `assert_equal(@lat, …)` / `@long` / `@alt` calls fired Minitest's `assert_equal nil, …` deprecation when test methods left those ivars unset. Switched to the existing `assert_equal_even_if_nil` helper already used for `@vote` / `@orig`.

## After

`bundle exec rubocop` exits with `no offenses detected` and emits no SuggestExtensions tip.

## Test plan

- [x] `bundle exec rubocop` — no offenses
- [x] `bin/rails test` on every touched file — green
- [x] System tests for the 3 files touched by `Capybara/FindAllFirst` — green
- [x] CI green (after FindAllFirst push)
- [x] coveralls green


JoeCohen (commented):
1. I didn't understand why all the new stuff in `.rubocop_todo.yml` is in that file rather than `.rubocop.yml`. Is that how we have to disable stuff in that plugin? Won't it all be overwritten when the file is automatically regenerated (which I do from time to time).
2. I do understand why you disabled the three cops you mentioned in Slack.
3. I don't understand why the others are disabled.

You have stronger opinions than me about the 3 bad cops. So I won't stand in your way about them.

nimmolo:
Good catch. The intern has moved the disables to rubocop.yml.

All the reasoning for each disable was written by our careful intern in the comment block above the disable! I think the reasoning sounds pretty sound, but they didn't survive the current move. I'll add them back.

This one's pretty self explanatory in the first line:
```
# Offense count: 5902
# This cop supports safe autocorrection (--autocorrect).
# Permanently disabled. The autocorrect inserts a blank line before
# every assertion that follows non-whitespace — pure cosmetic churn
# across nearly every test file (5900+ lines), with no readability
# win for MO's style.
Minitest/EmptyLineBeforeAssertionMethods:
```

PR #4546, opened 2026-06-16
location form: surface Google Maps geocode failures (#4535)

### LINKED ISSUE #4535: Edit Location Find on Map nonfunctional
> 6:43 PM (43 minutes ago)
> to webmaster
> 
> (from User #1539 I. G. Safonov(IGSafonov))
> Hello, Joseph:-
> 
> It appears that the "Find on Map" button in the location edit page is not working at this time.  Clicking on it does absolutely nothing. I found out while editing the Winona Lake, IN, location.
> 
> Thanks,
> 
> Igor

nimmolo:
## Summary

Fixes the user-facing symptom in #4535: clicking "Find on Map" on the location edit page appears to do absolutely nothing. The underlying problem is **not** in this codebase — it's in the Google Cloud Console — but the existing JS swallowed the failure to `console.log`, so users saw no signal.

@JoeCohen heads-up — see "What's actually broken" below; this PR is the code-side hardening, but the real fix needs to happen on the Google Cloud project.

## What's actually broken

Captured the browser console while reproducing locally (debug script logged to `window._logs`):

```
[showBox] called. opened: true hasPlaceInputTarget: true placeInputTarget.value: santa barbara hasLocationIdTarget: false map_type: location ignorePlaceInput: false
Geocoding Service: This API key is not authorized to use this service or API.
Geocode was not successful: MapsRequestError: GEOCODER_GEOCODE: REQUEST_DENIED: The webpage is not allowed to use the geocoder.
```

`map#showBox` fires correctly. The Stimulus targets and data attributes are all present. `geocode_controller#geolocatePlaceName` calls Google's Geocoding API. Google rejects with `REQUEST_DENIED` because the Maps API key has the **Maps JavaScript API** enabled (the map tile renders fine) but does **not** have the **Geocoding API** enabled. Those are separate Google Cloud products with separate enablement and separate restrictions.

Pre-existing tests (`test/system/location_form_system_test.rb` — `test_format_new_location_name` + `test_edit_location_geocode`) fail locally for the same reason. They presumably pass in CI because the CI environment uses a different key, or because the issue is recent.

## What this PR changes

Adds a single `showGoogleMapsError(label, error)` helper on `GeocodeController` and routes **every** silent Google Maps failure surface through it. Each path now injects a Bootstrap `.alert-danger` into `#gmaps_flash` — a div the location form renders above the place input — with a labelled error message and a dismiss button.

Failure surfaces covered:

| Surface | Before | After |
|---|---|---|
| Maps JS loader (bad key, network) | `console.error("error loading gmaps: …")` | "Google Maps failed to load: …" |
| `window.gm_authFailure` (Google's global on any Maps auth failure) | undefined → silent | "Google Maps authentication failed: the API key is missing, restricted, or doesn't have the required services enabled" |
| Geocoding API (place name → lat/lng) | `console.log("Geocode was not successful: …")` | "Geocoding failed: …" |
| Geocoding API (lat/lng → place name) | same silent `console.log` | "Geocoding failed: …" |
| Elevation API (`ElevationService.getElevationForLocations`) | `else`-branch was empty | "Elevation lookup failed: \<status\>" |

`gm_authFailure` is the catchall — Google calls it on any Maps JavaScript API auth failure (bad key, wrong referer, billing disabled). Hooked once per page load via a `window._mo_gmauth_installed` guard so multiple maps on one page don't reinstall it.

After this PR, an admin or contributor seeing "Find on Map does nothing" will at least get `Geocoding failed: REQUEST_DENIED. The map service may be misconfigured — please report this to a site administrator.` instead of the silent failure. Other Maps service breakages (key rotation, quota, referer drift) surface the same way.

## Diagnostic: per-API smoke test

`test/system/google_maps_api_smoke_test.rb` — a stand-alone system test that visits the location form and exercises **each** Google Maps service in turn. On any failure the assertion message names the specific API surface, so a developer running it locally knows exactly which Cloud Console product is misconfigured rather than chasing a silent regression.

```bash
bin/rails test test/system/google_maps_api_smoke_test.rb
```

Concrete failure on a key missing the Geocoding API (verified locally on this branch):

```
Geocoding API: `#gmaps_flash` populated with "Geocoding failed:
GEOCODER_GEOCODE: REQUEST_DENIED: The webpage is not allowed to use
the geocoder.". Enable the corresponding product on the Maps API key
in the Google Cloud Console.
```

Steps it checks (each is a separate Cloud Console product):

1. **Maps JavaScript API** — `#map_div` canvas renders; `gm_authFailure` didn't fire (`#gmaps_flash` still empty after page load).
2. **Geocoding API** — fill the place input, click "Find on Map", expect `#location_display_name.geocoded` and no flash.
3. **Elevation API** — click "Get Elevations", expect no flash.

## Files

- `app/javascript/controllers/geocode_controller.js` — `showGoogleMapsError(label, error)` helper; geocoder catches + elevation status handler route through it.
- `app/javascript/controllers/map_controller.js` — loader catch routes through the helper; install `window.gm_authFailure` once per page load.
- `app/views/controllers/locations/form.rb` — render the `<div id="gmaps_flash">` slot above the place input on the editable variant.
- `test/views/controllers/locations/form_test.rb` — assert the slot and the `data-map-target='placeInput'` data attribute.
- `test/system/google_maps_api_smoke_test.rb` — the per-API smoke test described above.

## What this PR does NOT do

It does not fix #4535's actual cause. The fix lives in the Google Cloud Console:

1. Open the Google Cloud project that owns the production Maps API key.
2. Enable the **Geocoding API** on the key (it's a separate enableable product from Maps JavaScript API).
3. Verify the referer restrictions still cover prod hostnames and `localhost:3000` (per the existing system-test setup).

Once that lands, both the production "Find on Map" button and the system tests on this branch will start working again.

## Test plan

- [x] `bin/rails test test/views/controllers/locations/form_test.rb` — 6 tests / 42 assertions, green
- [x] `bin/rails test test/system/google_maps_api_smoke_test.rb` — reproduces the Geocoding-API-denied case locally with the expected error message
- [x] `bundle exec rubocop` on touched Ruby — clean
- [ ] CI green
- [x] Google Cloud Console fix applied (out of band — owner: site admin); re-run the smoke test to confirm

### DIFF
--- a/app/javascript/controllers/geocode_controller.js
+++ b/app/javascript/controllers/geocode_controller.js
@@ -76,8 +76,7 @@ export default class extends Controller {
         this.respondToGeocode(results)
       })
       .catch((e) => {
-        console.log("Geocode was not successful: " + e)
-        // alert("Geocode was not successful for the following reason: " + e)
+        this.showGeocodeError(e)
       });
   }
 
@@ -108,9 +107,32 @@ export default class extends Controller {
         this.respondToGeocode(results)
       })
       .catch((e) => {
-        console.log("Geocode was not successful: " + e)
-        // alert("Geocode was not successful for the following reason: " + e)
+        this.showGeocodeError(e)
       });
+  }
+
+  // Surface a Google Maps geocode failure to the user. Without this the
+  // catch handler was a silent `console.log` — issue #4535: clicking
+  // "Find on Map" appeared to do nothing because the Geocoding API on
+  // MO's Google Cloud key was disabled and every request returned
+  // `REQUEST_DENIED`. The injected `.alert-danger` lives in
+  // `#geocode_flash`, a div the location form renders above the
+  // place input.
+  showGeocodeError(error) {
+    console.log("Geocode was not successful: " + error)
+    const flash = document.getElementById("geocode_flash")
+    if (!flash) return
+
+    const message = `Geocoding failed: ${error?.message || error}.
+      The map service may be misconfigured — please report this to
+      a site administrator.`
+    flash.innerHTML = `
+      <div class="alert alert-danger" role="alert">
+        <button type="button" class="close" data-dismiss="alert"
+                aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        ${message}
+      </div>
+    `
   }
 
   // Remove certain types of results from the geocoder response:
--- a/app/javascript/controllers/geocode_controller.js
+++ b/app/javascript/controllers/geocode_controller.js
@@ -76,7 +76,7 @@ export default class extends Controller {
         this.respondToGeocode(results)
       })
       .catch((e) => {
-        this.showGeocodeError(e)
+        this.showGoogleMapsError("Geocoding failed", e)
       });
   }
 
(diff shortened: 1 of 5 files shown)


JoeCohen:
Fixed on the Google Cloud Console side. It now works in production and test/system/location_form_system_test.rb  passes locally.

Issue #4543, opened 2026-06-15
Triage orphaned iNat-imported images (source photo deleted on iNat)


mo-nathan:
## Problem

The #4213 audit found **38 observations / 77 imported images** where the iNat **source photo has since been deleted on iNat** — the MO image is an orphan with no live iNat counterpart. These need a one-time remediation, and unlike the go-forward sync policy (#4215) the decision **cannot be automated**: it is per-image and three-way — **keep** (adopt as MO-native), **delete** (the photo was deliberately removed on iNat), or **ask** the original observer. A single observation can mix all three.

## Initial review

First-pass review of all 38 (decision + reason columns) is in the attached `inat_orphan_reviewed.csv`, embedded below. Tally:

| decision | obs | meaning |
|---|---:|---|
| Adopt | 16 | keep as MO-native (field slips, collection bags, host/habitat shots, extra angles of the same specimen) |
| Delete | 7 | orphan is a different species — correctly removed on iNat |
| Adopt? | 11 | lean keep, needs the observer's confirmation |
| Delete? | 3 | lean delete, needs the observer's confirmation |
| Complicated | 1 | mix within one observation |

So **23 are decided** (16 keep / 7 delete) and **15 need observer input**.

## Patterns worth noting

- A large share of the keeps are exactly the contextual photos MO values and iNat curators routinely strip: **field slips, collection bags, host plants, habitat/environment shots, extra angles**.
- The firm deletes are uniformly **"different species"** — a photo that was removed on iNat because it did not belong to the observation.
- **Who removed these?** iNat observation media is **owner-controlled** — curators' powers are taxonomic (taxon changes, merges, flags), not editing other users' observation photos — so these removals are most likely the **observation owners' own decisions** (deletes = owners pulling a photo that did not belong; keeps = owners trimming context shots that MO wants to keep). iNat does not expose per-photo deletion attribution, so the agent cannot be confirmed for any specific case.
- **Content-matcher caveat (586383):** the observer combined two MO images into a single *composite* for the iNat photo, so neither MO image matches the live iNat photo and both were flagged as orphans. The aHash matcher cannot detect composites; the right action here is to import the composite. Relevant to the #4529 backfill.

## Mechanism

[...]

Issue #4529, opened 2026-06-14
Structured per-photo provenance for imported images


mo-nathan:
## Goal

Record each imported MO image's **source photo id structurally** — not in `original_name` — so sync can map MO images to iNat photos: add new photos, propagate deletions (copyright/legal), and avoid clobbering edited (rotated / higher-res) images.

## Background

The importer already captures provenance — `Inat::ObsPhoto#original_name` writes `"iNat photo_id: <id>, uuid: <uuid>"` "for purposes of syncing updates" — but `original_name` is a free-text, user-editable display field and is **already overwritten** on some observations (e.g. obs 642622, 640048). iNat photos carry an explicit `:position`, and the importer adds them in that order, so MO `created_at` order tracks iNat position.

The #4213 audit (#4528) found **764 imports with an image-count mismatch** between MO and iNat, and confirmed the provenance fragility — making this the prerequisite for reliable image sync.

## Acceptance

- A structured store for `(MO image, external site, external photo id)` — a column on the image↔obs join, or an image-level `ExternalLink` (consistent with #4299).
- The importer writes it going forward.
- **Backfill** existing imports: parse `original_name` where the `iNat photo_id:` format survives; fall back to **position-order inference** (iNat `:position` vs MO `created_at` order) where it doesn't. `log()` the rows that can't be mapped (count mismatch / ambiguous order).
- Unblocks #4215's image sync (add new, delete-on-source, preserve edited).

## Relates

Tracking: #4208. Blocks #4215. Consistent with #4299.

mo-nathan:
## The provenance gap is far larger than "some `original_name` fields were overwritten" — it's a second, un-instrumented import path

Quantified across the full corpus while reviewing the #4213 audit. The gap isn't an edge case; it's the default state of most imported images.

**The numbers.** Of **186,830** imported images (`notes LIKE 'Imported from iNat%'`), only **24,277 (13%)** carry the `original_name` `"iNat photo_id: <id>, uuid: <uuid>"`. **162,553 (87%)** have a blank `original_name` despite identical `"Imported from iNat …"` notes.

**Root cause — two parallel import paths, not overwriting.** The Background above attributes the gap to `original_name` being user-edited. That does happen, but it's a rounding error next to the real cause: there are (at least) two import code paths running concurrently and only one records the photo id.

[...]

PR #4556, opened 2026-06-16
Backfill iNat image provenance: parse + position + content-match (#4529)

mo-nathan:
## Summary

The **backfill half of #4529** — recovering iNat photo provenance for the 186,838 already-imported images. Three runner scripts, escalating in cost as the cases get harder. Builds on #4555 (the `images.source_id` / `external_id` columns + go-forward capture).

### 1. `script/backfill_inat_image_provenance.rb` — parse path (no network)

The ~24,285 images whose `original_name` still holds `"iNat photo_id: <id>, uuid: <uuid>"` (imports by `keep_filenames != "toss"` users): parse the id, write `external_id` + `source_id`.

### 2. `script/infer_inat_image_provenance.rb` — position-order inference

The ~159,656 whose `original_name` was nulled by `keep_filenames = "toss"`: fetch each observation's current iNat photos ordered by `position` (via the audit's batched, rate-limited, public `Inat::ImportAudit::Fetcher`), order the imported MO images by creation, and map one-to-one by position. **Only maps when counts line up**; `not_found` / `fetch_error` / `count_mismatch` / `position_conflict` are logged, not guessed.

### 3. `script/content_match_inat_image_provenance.rb` — aHash content-match fallback

Position inference *skips* any observation whose image count drifted (deletes, higher-res re-adds, swaps). This recovers them by matching on **content** instead of order: average-hash (aHash) each MO image and each current iNat photo via ImageMagick (8×8 grayscale, aspect-preserving sizes both sides), pair by smallest Hamming distance (greedy, unique, within `--threshold`). Matched images get their photo id; **unmatched MO images are real orphans** (overlaps #4543), **unmatched iNat photos were added after import** — both logged.

aHash is resolution-invariant, so a higher-res re-add still matches at distance 0.

## Validation (against `original_name` ground truth)

| Method | Result |
|---|---|
| Position inference | reproduced the true photo ids **12/12** and **100/100** sampled |
| Content match | reproduced the true photo ids **76/76** matched images |

Worked examples on real count-mismatch obs:
- **565650** (MO 3 / iNat 4, a higher-res swap): content-match recovered all 3 at **hamming 0** and flagged the 1 extra iNat photo.
- **561533** (MO 20 / iNat 0, a #4543 orphan): correctly produced 20 `orphan_mo_image`, zero false matches.

Timing (100-obs visual-check sample, one fetch batch): **~0.12s/obs** → full position pass ≈ ~60–86 min across ~217 batches at ~3–4 req/min (well under iNat's rate guidance).

## Running it (production, when ready)

```bash
bin/rails runner script/backfill_inat_image_provenance.rb            # parse (fast)
bin/rails runner script/infer_inat_image_provenance.rb --limit 200   # stage, then no --limit
bin/rails runner script/content_match_inat_image_provenance.rb --limit 50  # fallback, then no --limit
```

All three are idempotent (only touch images still lacking `external_id`), `update_columns` (no callbacks/timestamps/notifications), `--dry-run` + CSV reports.

## Not run here

The full passes have **not** been executed against any database — dry-run + ground-truth validation + small live `--limit` samples only. Run staged against production when ready.

Refs #4529. Stacked on #4555.


mo-nathan:
Closing without merging — these four scripts (parse / position-inference / aHash content-match / loader) were one-off backfill tooling, not intended for `main`.

The backfill has been computed locally and loaded into production: **187,895 of 188,307 imported images (99.78%)** now carry structured `(source_id, external_id)` provenance. The 412 remaining are unrecoverable — 101 deleted-on-iNat (→ #4543) and 311 orphaned images with no observation. Go-forward capture is live via #4555.

PR #4570, opened 2026-06-19
Button component - new API and sweeping refactor

nimmolo:
## Summary

Sweeps all remaining raw `class: "btn btn-*"` callers out of `app/components/` and `app/views/`, completing the work started by the `button_to` → CrudButton tier. Every btn-class string is now expressed via `variant:` / `size:` (or `toggle_variant:` / `toggle_size:` on `Components::Dropdown`) rather than hand-rolled Bootstrap class strings.

## What changed

**New APIs:**
- `Components::ButtonContent` module — shared `button_content` extracted from both `Components::Button` and `Components::Link`. Includes a `label:` kwarg: when `true`, icon-adjacent text renders as `d-none d-sm-inline ml-1` (responsive) instead of `sr-only`. Used by the top-nav "Add" button.
- `Components::Dropdown` — now includes `ButtonStyling` and accepts `toggle_variant:` / `toggle_size:` props. Plain dropdowns (no `toggle_variant`) are unchanged.
- `InputGroupAddon` — callers now pass `button_variant:` / `button_size:` instead of a raw `button_class:` string. `button_class` removed from `WRAPPER_OPTIONS`.
- `preferences/form.rb` — `addon_button_defaults` extracted so `render_keep_filenames_select` and `external_addon` share the common `button_variant: :outline, button_size: :sm, button_icon: :new_window` base and merge their particulars on top.

**Converted callers (tier by tier):**
- Layout components: `top_nav.rb`, `top_nav/login.rb`, `sidebar/admin.rb`, `sidebar/user.rb`
- Remaining view-layer components: `namings/row.rb`, `visual_groups/edit.rb`, and others
- `InputGroupAddon` + `FileField` (`span.file-field.btn.btn-default` → `Button(tag: :span)`)
- `Components::Dropdown` toggle in `header/sorter.rb` (last raw caller)

**Link hierarchy refactor** (prerequisite, same branch): flattened `Link::Object::*` namespace, unified dispatch to `type:` kwarg, fixed CRUDBase `variant:` regressions.

## Test plan

- [x] `bin/rails test test/components/` — all component tests green
- [x] `bin/rails test test/views/layouts/` — sorter + other layout tests green
- [x] `bin/rails test test/views/controllers/account/preferences/form_test.rb` — addon parity test green
- [x] `bundle exec rubocop` — no offenses on all changed files
- [x] Parity tests written before each tier conversion to lock in pre-change HTML structure


app/classes/tab/comment/new.rb
```diff
@@ -18,7 +17,7 @@ def path
   end
 
   def html_options
-    { class: @btn_class, icon: :add }
+    {}
```
mo-nathan: Is this override of the inherited version from app/classes/tab/base.rb needed any more?  There are bunch of similar cases.

Using CC:

  There are 10 of these redundant {} overrides in the PR's tab classes:

  comment/new.rb
  location/new.rb
  collection_number/new.rb
  project/alias_new.rb
  description/create.rb
  external_link/new.rb
  sequence/new.rb
  name/new_description.rb
  name/new.rb
  herbarium_record/new.rb

mo-nathan:
> Copilot wasn't able to review this pull request because it exceeds the maximum number of files (300). Try reducing the number of changed files and requesting a review from Copilot again.

Seems like it would be worthwhile having CC break this up into two PRs so Copilot can weigh in.

app/classes/tab/comment/new.rb
nimmolo: Do you mean the `html_options` kwarg in general? Or just here on the comment new tab. If you mean in general, i think i understand the question.
 
I played with getting rid of it and having callers own the options, but IMO the kwarg is still needed. There are a couple dozen of them that pass options in a way that seems proper to the Tab and not the caller.

mo-nathan: No, I mean cases where you're replicating what the parent class does which is return {}.  No point in including it if it is just doing what the parent class is already returning.

nimmolo: True! Thanks

Issue #4299, opened 2026-05-18
Consolidate Source into ExternalSite/ExternalLink


mo-nathan:
## Goal

Collapse the `Source` table + `Observation#source_id` / `external_id` columns (added in #4209) into the existing `ExternalSite` / `ExternalLink` M:N relationship. After this change, an observation can hold multiple typed relationships with external platforms (import, cross-reference, future mirror, etc.) rather than at most one `Source`.

This is the schema cleanup we identified during PR #4235, deferred to its own issue so #4235 could ship the visible-source banner against the current model.

Tracking issue: #4208.

## Why now

- **MyCoPortal (#4216) lands next.** Building it on the unified model from the start avoids a retrofit — the same code path that creates an `ExternalLink(relationship: "import")` for iNat handles MyCoPortal.
- **Resolves the "redundant iNat link" duplication** Joe noted on #4235: an obs imported from iNat and also cross-referenced to iNat today has two rows (Source pointer + ExternalLink). The backfill collapses these.
- **Keeps the import path open for new relationship types** (mirror, future exports) without further schema churn.

## Schema changes

`external_sites`:
- Add `description` (text)
- Add `last_successful_sync_at` (datetime)

`external_links`:
- Add `external_id` (string) — the platform's stable id for that observation
- Add `relationship` enum: `{ import, export, mirror, cross_reference, unknown }` — initial set; new values added via migration as needed
- Add `last_synced_at` (datetime)

`observations`:
- Drop `source_id`, `external_id`

Drop the `sources` table.

## URL pattern

`ExternalSite#observation_url(external_id)` keeps the same shape it has on `Source` today — a per-site case statement returning the per-obs URL. `external_id` alone is sufficient for iNat and MyCoPortal as currently understood; future sources that need more get their own case branch.

## Backfill

For each observation with `source_id` present:
- `find_or_create_by` an `ExternalLink` for the matching `ExternalSite` (looked up by name).
- Set `external_id`, `relationship: "import"`, derive `url` from `base_url + external_id`.
- If a `cross_reference` `ExternalLink` already exists for the same (observation, site), upgrade it in place to `import` rather than creating a duplicate. This is the "redundant iNat link" fix.

Existing `external_links` rows default to `relationship: "cross_reference"` (their historical meaning).

## Code changes

- Update import jobs to write `ExternalLink(relationship: "import")` instead of `observations.source_id` / `external_id`.
[...]

PR #4572, opened 2026-06-20
Drop Source table + source columns (#4299 phase 2)

mo-nathan:
## Summary

Phase 2 of #4299 — drops everything the phase-1 polymorphic `ExternalLink` model superseded, now that import provenance lives on import `ExternalLink`s.

**Stacked on the Phase 1 PR (#4568).** This PR's base is the phase-1 branch so the diff shows only phase-2 changes; once #4568 merges to `main`, GitHub retargets this to `main`.

## Deploy sequencing (important)

This must deploy **after** phase 1 **and after the backfill has run in production** — the backfill reads `observations.source_id` / `images.source_id`, the exact columns this drops. The order is:

1. Merge + deploy #4568 (phase 1).
2. Run `script/backfill_external_links_from_sources.rb` (`APPLY=1`) on prod; verify.
3. Merge + deploy this PR.

They can land close together, but not simultaneously — step 2 has to happen in between. (Per the runner-scripts-not-migrations convention, the backfill is not folded into the drop migration.)

## Schema (drop migration `20260620175348`)

- `observations`: drop `source_id`, `external_id`, `last_synced_at` + their unique index and the FK to `sources`.
- `images`: drop `source_id`, `external_id` + their index.
- drop the `sources` table.

Order within the migration: drop FK → drop index → drop columns → drop table (MySQL requires the FK gone before its backing index, and the index gone before its columns). `down` recreates the structure.

## Code

- Delete the `Source` model; remove `belongs_to :external_source` from `Observation` and `Image`.
- **`import_audit`** (`runner` + `row_builder`): select iNat-imported observations via their import `ExternalLink` (not `observations.source_id`), and read the iNat id + URL from the link — the runner passes `external_id` through to the row builder, which builds the URL via `ExternalSite#observation_url`.
- Drop the now-unused `source_credit_external_no_url` locale key (`link_url` always resolves).
- Comment cleanups referencing the old columns.

## One-time backfill removed

Deletes `script/backfill_external_links_from_sources.rb` and its test. It's a one-time runner that will have already run in prod between the two deploys, and it references the columns this migration drops. (It remains available in the #4568 diff / history for the prod run.)

## Fixtures / tests

- Delete `sources.yml` and `source_test.rb`.
- Drop the source columns from the `imported_inat_obs` fixture — its provenance is the existing import `ExternalLink` fixture.
- Update the audit tests and `rss_logs_controller_test` to read the iNat id/URL from the import link.

## Test plan

- [x] `bin/rails test` — full suite green: **6555 tests, 42939 assertions, 0 failures, 0 errors**.
- [x] Migration `up` + `down` exercised; schema confirmed (no `sources` table, no source columns/indexes/FK).
- [x] RuboCop clean on all changed files.

Relates to #4208. Phase 2 of #4299.


mo-nathan:
Coveralls failure due to total lines dropping rather than newly uncovered lines.  The total uncovered lines remains at 813.

nimmolo (approved):

PR #4600, opened 2026-06-25
Fix nil observation_id in add_missing_views_corresponding_to_votes

### LINKED ISSUE #4599: refresh_caches script output omits observation ids
refresh_caches script sends webmaster emails like this:

```
Cron Daemon
Jun 24, 2026, 10:15 PM (6 hours ago)
to webmaster

User 38275 has reviewed observation  (insert).
User 4468 has reviewed observation  (insert).
User 4468 has reviewed observation  (insert).
User 4468 has reviewed observation  (insert).
User 4468 has reviewed observation  (insert).
User 1 has reviewed observation  (insert).
...
```

JoeCohen:
Fixes `refresh_caches` output which omitted Observation id. Example bad output:

>User 38275 has reviewed observation  (insert).

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -240,7 +240,8 @@ class Vote < AbstractModel
     # (user 0 is used for anonymous votes, ignore those)
     Vote.where.not(user_id: 0).joins(join).
       where(observation_views: { id: nil }).
-      select(:observation_id, :user_id, :updated_at).map do |vote|
+      select("votes.observation_id, votes.user_id, votes.updated_at").
+      map do |vote|
       unless dry_run
         ObservationView.create!(
           observation_id: vote.observation_id,
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -242,17 +242,17 @@ class Vote < AbstractModel
       where(observation_views: { id: nil }).
       select("votes.observation_id, votes.user_id, votes.updated_at").
       map do |vote|
-      unless dry_run
-        ObservationView.create!(
-          observation_id: vote.observation_id,
-          user_id: vote.user_id,
-          last_view: vote.updated_at,
-          reviewed: 1
-        )
+        unless dry_run
+          ObservationView.create!(
+            observation_id: vote.observation_id,
+            user_id: vote.user_id,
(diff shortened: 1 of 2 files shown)


app/models/vote.rb
```diff
@@ -240,18 +240,19 @@ def self.add_missing_views_corresponding_to_votes(dry_run: false)
...
       where(observation_views: { id: nil }).
-      select(:observation_id, :user_id, :updated_at).map do |vote|
-      unless dry_run
-        ObservationView.create!(
-          observation_id: vote.observation_id,
-          user_id: vote.user_id,
-          last_view: vote.updated_at,
-          reviewed: 1
-        )
+      select("votes.observation_id, votes.user_id, votes.updated_at").
```
nimmolo: Not that i think this is a blocker, but I don't believe this line (243 only) needs to change. Ask CC to write it in pure AR/Arel, and it should give you what we had before. 

It seems like it might be done to resolve some ambiguity, but it's a single-table query, no joins - so these columns should not be ambiguous.

JoeCohen: @nimmolo CC says:
Bug: script/refresh_caches was logging lines like "User 4468 has reviewed observation  (insert)." with a blank observation id.

Cause: In add_missing_views_corresponding_to_votes, the query does a LEFT OUTER JOIN between votes and observation_views. Both tables have an observation_id column. The original .select(:observation_id, :user_id, :updated_at) generates unqualified column names in SQL, and MySQL resolves the ambiguity by returning observation_views.observation_id — which is NULL for every row in the result (those are exactly the rows where no matching observation_views record exists).

Fix: Qualify the column names using Arel so MySQL reads from the right table:

select(arel_table[:observation_id], arel_table[:user_id], arel_table[:updated_at])
Reviewer pushback: A reviewer claimed the select line didn't need to change because "it's a single-table query, no joins." That premise is wrong — the JOIN is right there in the method. However, their stylistic suggestion to use pure AR/Arel (rather than a raw SQL string) was valid.

Verification: Running the original query in the Rails console against the live database confirmed that results.first.observation_id returns nil, proving the select was the root cause.

Issue #4575, opened 2026-06-21
Ambiguous imported rank


JoeCohen:
When importing https://www.inaturalist.org/observations/132754043

> Failed to import iNat 132754043: Failed to create name "Leucocoprineae": The name, "Leucocoprineae", doesn't parse as a valid scientific name.

The problem is that the ICN is unambiguous in only one direction: Genus => rank above Genus, but not vice versa.
-ineae is the canonic suffix for Suborder. https://www.iaptglobal.org/_functions/code/madrid?m=art17
The canonic ending for a Tribe is -eae, and if the source genus ends in "inxx", the derived Tribe ends in  -ineae.

Proposed solution: If rank is ambiguous or doesn't parse, use the iNat rank.

JoeCohen:
app/classes/inat/mo_observation_builder.rb#post_name calls our api
```ruby
      api = API2.execute(params)
```
with params 
```ruby
{method: :post, action: :name, api_key: "f7kikza8b51bnn5napbn4v41aghcdvog", name: "Leucocoprineae", rank: "Tribe"}
```
API returns the error.

JoeCohen:
CC explains

**The chain of events:**

1. `make_sure_name_parses!` calls `Name.parse_name("Leucocoprineae", rank: "Tribe")`.

2. `parse_name` cycles through all below-genus parsers (subgenus, section, subspecies, species,
   etc.) — all fail for a single uppercase word with no spaces — and falls through to
   `parse_genus_or_up("Leucocoprineae", false, "Tribe")`.

3. Inside `parse_genus_or_up`, it calls `guess_rank("Leucocoprineae")`.

4. `guess_rank` walks `TEXT_NAME_MATCHERS` in order. The critical entries are:

```ruby
   RankMatcher.new("Suborder",   /^\S+ineae$/),   # checked first
   RankMatcher.new("Tribe",      /^\S+eae$/),      # checked second
```

"Leucocoprineae" ends in the five characters i-n-e-a-e = "ineae". The Suborder
regex /^\S+ineae$/ matches it, so guess_rank returns "Suborder".

5. Back in parse_genus_or_up:

- rank (from the API call) = "Tribe"
- Name.ranks_above_genus.include?("Tribe") → true
- The guard condition: !force_rank && guessed != "Genus" && guessed != rank → "Suborder" != "Tribe" → true
- → raises RankMessedUp internally

[...]

PR #4631, opened 2026-07-01
Fix parse failure due to nomenclature issue.

jonkiparsky:
@JoeCohen I took the liberty of picking this one up. Fix seems sensible, hope I'm not stepping on ongoing work.

### DIFF
--- a/test/classes/inat_mo_observation_builder_test.rb
+++ b/test/classes/inat_mo_observation_builder_test.rb
@@ -228,6 +228,39 @@ class InatMoObservationBuilderTest < UnitTestCase
                  "spelling note when proposed name corrects the obs taxon")
   end
 
+  # --- preferred_rank: ICN suffix overrides iNat rank for above-genus names --
+
+  def test_preferred_rank_suffix_wins_over_inat_rank
+    # Leucocoprineae ends in -ineae (Suborder per ICN). iNat assigns "tribe";
+    # MO should create it as Suborder, which is what the suffix mandates.
+    assert_equal("Suborder",
+                 builder_for.send(:preferred_rank, "Leucocoprineae", "tribe"))
+  end
+
+  def test_preferred_rank_falls_back_for_unsuffixed_name
+    # A plain single-word name has no disambiguating suffix; keep iNat's rank.
+    assert_equal("Genus",
+                 builder_for.send(:preferred_rank, "Boletus", "genus"))
+  end
+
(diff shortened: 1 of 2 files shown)


mo-nathan:
Addresses #4575

test/classes/inat_mo_observation_builder_test.rb
```diff
@@ -228,6 +228,39 @@ def test_observation_taxon_naming_reason_corrected_spelling
                  "spelling note when proposed name corrects the obs taxon")
   end
 
+  # --- preferred_rank: ICN suffix overrides iNat rank for above-genus names --
+
+  def test_preferred_rank_suffix_wins_over_inat_rank
+    # Leucocoprineae ends in -ineae (Suborder per ICN). iNat assigns "tribe";
+    # MO should create it as Suborder, which is what the suffix mandates.
+    assert_equal("Suborder",
+                 builder_for.send(:preferred_rank, "Leucocoprineae", "tribe"))
```
mo-nathan, line 237: From my reading of the original issue, this is wrong.  "Leucocoprineae" is a Tribe, so we should be following iNat's conclusion rather than assuming it's a Suborder just because it ends in -ineae.  The issue is that the prefix "Leucocoprin-" (from Leucocoprinus) happens to end in "in".  Per @JoeCohen's comment at the end of #4575 we need to come to the conclusion one way or another that this is Tribe not a Suborder.  Also note that Tribe is a much lower taxonomic rank (below Family) than Suborder (between Order and Family).  My recommendation would be to go with what iNat says.  If you want to go a step further, then you could check that it has a reasonable suffix for the declared rank - in this case "-eae" for Tribe.  However, that's probably creates more problems than it solves.

mo-nathan (changes requested):
One way or another we need to conclude that "Leucocoprineae" is a Tribe.

JoeCohen:
@jonkiparsky Thank you for picking up this issue. Please "take the liberty" any time you like. 
If you do pick up other issues, please le me know so we're not stepping on each others' toes.

JoeCohen:
I agree with @mo-nathan.
- It's a tribe. ( I've already created that tribe in the live db https://mushroomobserver.org/names/116612, so that name now imports correctly)
- We should go with iNat's name for any future ambiguous name endings.  (This also might allow us to prune some code.)

JoeCohen:
Closing in favor of #4915

PR #4656, opened 2026-07-03
Fix 2 remaining pre-existing system test failures

nimmolo:
## Summary

Fixes the last 2 pre-existing, unrelated system test failures found while investigating the confirm-modal Turbo regression in #4655, so the full system test suite can go green and be turned on in CI.

- **`AdminModeSystemTest#test_admin_mode_toggle_from_inat_import_page`**: `InatImportsController#show` does `InatImportJobTracker.find(params[:tracker_id])` unconditionally, but neither the test nor the `rolf_inat_import` fixture supplied a `tracker_id`. Added a `rolf_tracker` fixture and pass its id explicitly in the test's `visit(...)` URL, matching the pattern already used in `test/controllers/inat_imports_controller_test.rb`.
- **`AutocompleterSystemTest#test_observation_search_region_autocompleter`**: stale selector. `Components::Panel` deliberately omits `data-target` for id-based collapse targets — see the comment at `app/components/panel.rb:152` — so Bootstrap reads the trigger's `href="#id"` and calls `preventDefault()`, avoiding a Turbo-frame navigation bug. The test was never updated to the `aria-controls` selector that replaced it; `search_bar_collapse_system_test.rb` already uses the correct pattern.

## Test plan

- [x] `bundle exec rubocop` on both touched Ruby files — no offenses
- [x] `test/system/admin_mode_system_test.rb` — 2/2 pass
- [x] `test/system/autocompleter_system_test.rb` — 14/14 pass

### DIFF
--- a/test/fixtures/inat_import_job_trackers.yml
+++ b/test/fixtures/inat_import_job_trackers.yml
@@ -8,3 +8,6 @@ lone_wolf_tracker:
 
 ollie_tracker:
   inat_import: <%= ActiveRecord::FixtureSet.identify(:ollie_inat_import) %>
+
+rolf_tracker:
+  inat_import: <%= ActiveRecord::FixtureSet.identify(:rolf_inat_import) %>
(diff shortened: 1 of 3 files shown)


nimmolo:
Folded into #4655 — these two fixes needed the confirm-modal Turbo fix to actually pass in CI (this branch alone still hits the #mo_confirm failures), so merging them into one PR made more sense than keeping them separate.

PR #4710, opened 2026-07-06
Fix shadowed update_field_slip method in ObservationsController::Create

nimmolo:
## Summary

`ObservationsController` includes both `Create` and `EditAndUpdate`, and both defined a private `update_field_slip` method. Ruby's module resolution order means `EditAndUpdate#update_field_slip` always wins — `Create#update_field_slip` was 100% dead code, both in tests and in production. The `create` action's field-slip attachment was silently running through `EditAndUpdate`'s `assign_field_slip`/`clear_field_slip` logic instead of its own, undetected because both paths happen to produce an equivalent end result for the existing test scenario.

Confirmed via `ObservationsController.instance_method(:update_field_slip).owner` returning `ObservationsController::EditAndUpdate` even when called from the `create` action.

- Renamed `Create`'s version to `attach_new_field_slip` so it actually runs.
- Added a comment on the method explaining why the distinct name matters.
- Added a dedicated regression test pinning `ObservationsController.instance_method(:attach_new_field_slip).owner` to `ObservationsController::Create`, so a future name collision fails loudly instead of silently.
- The existing `test_create_observation_with_field_slip` test still passes with the real `Create` logic now executing.

Surfaced while chasing a coverage gap on a separate PR (#4705); pulled out here as its own standalone fix since it's unrelated to that PR's actual scope (the `User.current` removal sweep).

## Test plan

- [x] `bin/rails test test/controllers/observations_controller_create_test.rb` — green (63 tests)
- [x] `bundle exec rubocop app/controllers/observations_controller/create.rb test/controllers/observations_controller_create_test.rb` — no offenses

### DIFF
--- a/app/controllers/observations_controller/create.rb
+++ b/app/controllers/observations_controller/create.rb
@@ -61,7 +61,7 @@ module ObservationsController::Create
     save_collection_number
     save_herbarium_record
     strip_images! if @observation.gps_hidden
-    update_field_slip
+    attach_new_field_slip
     flash_notice(:runtime_observation_success.t(id: @observation.id))
     redirect_to_next_page
   end
@@ -254,7 +254,11 @@ module ObservationsController::Create
            location: new_observation_path)
   end
 
-  def update_field_slip
+  # NOTE: named distinctly from EditAndUpdate#update_field_slip - both
+  # modules are included into ObservationsController, and a same-named
+  # method here would be silently shadowed by EditAndUpdate's version
+  # (included after Create), never actually running.
+  def attach_new_field_slip
     field_code = params[:field_code]
     return if field_code.blank?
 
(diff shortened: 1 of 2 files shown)


mo-nathan:
Thanks for catching and isolating this — the shadowing diagnosis (and the `instance_method(:update_field_slip).owner` confirmation) is spot on.

Heads-up that #4715 addresses the same shadowing, so the two overlap and will conflict. It takes a slightly different tack: instead of renaming `Create`'s copy to un-shadow it, it extracts the shared logic into an `ObservationsController::FieldSlips` concern with a single `update_field_slip` that returns a status (`:assigned`/`:invalid`/…), and each action surfaces an invalid code its own way — `update` re-renders the edit form, `create` warns and keeps the observation (since it's already saved by then) with a create-specific message rather than the `edit_observation_field_slip_invalid` one.

I did carry over the best idea from here: your regression test pinning the method's owner, as `test_update_field_slip_is_not_shadowed` (credited in the commit). That guard is a nice touch that the extraction approach benefits from too.

Since #4715 subsumes this fix plus the DRY cleanup, probably cleanest to land that and close this.

mo-nathan (changes requested):
Close since this is replaced by #4715

Issue #4726, opened 2026-07-08
Audit: move suitable crontab scripts to Solid Queue recurring jobs


nimmolo:
## Summary

Audit of the production crontab to identify which scheduled scripts should become Solid Queue recurring jobs (`config/recurring.yml` + `app/jobs/*.rb`), matching the pattern already established by `InatImportRecoveryJob` / `DiscardStaleJobsJob` / `UpdateBoxAreaAndCenterColumnsJob`.

**Existing Solid Queue capacity** (relevant constraint): production runs a single supervisor (systemd `solid_queue:start`; the Puma plugin is deliberately disabled in production per #4639/PR #4657 to avoid double-claiming) — one dispatcher, one worker process, 5 threads. `config/queue.yml` is fully commented out — no queue-priority split, no per-queue config at all. Three recurring jobs already exist: `recover_stuck_inat_imports` (every 3 min), `discard_stale_jobs` and `repopulate_location_and_observation_center_lat_lng` (both daily at midnight). Adding several new recurring jobs onto this same undifferentiated 5-thread pool is a real capacity question this issue raises rather than ignores.

## Proposed for this round: Bucket A (convert now)

These already boot Rails and/or are simple ActiveRecord cleanup, idempotent, no external infra dependency:

| Cron script | Schedule | New job | Notes |
|---|---|---|---|
| `repair_observation_vote_cache.rb` | daily 5:28am | `RepairObservationVoteCacheJob` | Already Rails-booted (`bin/rails runner`), already ActiveRecord, and the only script that documents its own idempotency in-code ("safe to re-run; calc_consensus has an idempotent early-return"). Nearly a direct lift — best first candidate. |
| `refresh_name_lister_cache` | daily 3:33am | `RefreshNameListerCacheJob` | Already Rails-booted, single big query + regenerates `public/name_list_data.js` from scratch each run (idempotent). |
| `refresh_caches` | daily 5:13am | `RefreshCachesJob` | Already Rails-booted, ActiveRecord cache/integrity fixups, each step is a "make sure X is correct" repair (idempotent per-step). Runtime: minutes (full-table scans) — occupies a worker thread for a while, factor into the queue-priority work below. |
| `check_for_broken_references` | weekly Sun 7:13am | `CheckForBrokenReferencesJob` | Already Rails-booted, heavy ActiveRecord, idempotent (converges to nothing-left-to-fix). Runtime: minutes across many tables — same thread-occupancy note as above. |
[...]

PR #4730, opened 2026-07-08
Port `refresh_caches` script into 4 Solid Queue jobs, staggered by measured cost

nimmolo:
## Summary

From #4726 (crontab → Solid Queue recurring-job audit, Bucket A) — this job doesn't depend on any of the other Bucket A job PRs, so it's not chained through them.

Ports `script/refresh_caches` into Solid Queue jobs, bringing the Ruby logic directly into the job layer. Originally landed as a single `RefreshCachesJob` running all 10 of the script's tasks in one daily slot; timed each task locally (`dry_run: false`) against a recent production-size DB copy to decide how to stagger it, and that timing changed the shape of this PR:

```
UserStats.refresh_all_user_stats                         8.929s
Observation.refresh_content_filter_caches                 6.640s
Vote.update_observation_views_reviewed_column              4.765s
Observation.refresh_needs_naming_column                    1.559s
Occurrence.refresh_has_specimen_cache                       1.539s
Observation.make_sure_no_observations_are_misspelled        1.147s
User.cull_unverified_users                                  0.635s
Name.fix_self_referential_misspellings                      0.083s
Name.make_sure_names_are_bolded_correctly                   0.046s
Synonym.make_sure_all_referenced_synonyms_exist              0.033s
```

Three tasks dominate the ~25s local total (production's own single-job runtime measured considerably higher, ~1m41s). Split those three out into their own jobs so each can be staggered independently and none of the seven cheap tasks queues behind whichever of the three is having a slow night:

- `RefreshAllUserStatsJob` — `UserStats.refresh_all_user_stats`
- `RefreshContentFilterCachesJob` — `Observation.refresh_content_filter_caches`
- `UpdateObservationViewsReviewedColumnJob` — `Vote.update_observation_views_reviewed_column` (also the one most likely to get slower as data grows — its "missing views" half creates one `ObservationView` row at a time rather than in bulk)

The remaining 7-task job kept the name `RefreshCachesJob` at first, but that overclaims once it's not the full original set — half of what's left (misspelling fixes, culling unverified users) was never a cache refresh to begin with. Renamed to `MiscDataRepairsJob`.

`config/recurring.yml`: the single `5:13am` slot became four entries, staggered ~4-5 minutes apart and off `recover_stuck_inat_imports`' every-3-minutes marks, same reasoning as the existing stagger comment on the jobs above it (independent jobs, no cost to spacing them out).

`Name.propagate_generic_classifications` stays disabled everywhere, per the original script's own comment — it's not vetted enough to run unattended yet.

Each job's own test stubs its task(s) to verify the orchestration contract (task runs, `dry_run` is forwarded, messages are logged) rather than re-testing the underlying repair logic those model methods already cover — plus one real end-to-end run per job to catch any kwarg/signature drift the stubs would hide.

Updates the "Used by RefreshCachesJob" doc comments in `vote.rb`, `observation.rb`, and `propagate_generic_classifications.rb` to point at whichever job now actually calls each method.

Deletes `script/refresh_caches` — no other live references remain. (`test/models/cache_test.rb` has an already-disabled, already-stale commented-out test referencing older method names entirely; left alone as out of scope for this PR.)

`Vote.update_existing_views_corresponding_to_votes`/`add_missing_views_corresponding_to_votes` (called by `UpdateObservationViewsReviewedColumnJob` via `Vote.update_observation_views_reviewed_column`) build their SQL joins from a raw string, which is a `.claude/rules/no_raw_sql.md` violation — that rewrite has been moved to #4731 instead (stacked on #4702, at Joe's request), since #4702 already touches those same methods for an unrelated bug fix.

## Test plan

- [x] `bundle exec rubocop` clean on all touched `.rb` files.
- [x] `bin/rails test test/jobs/` — all green, including all four new/renamed job test files.
- [x] `bin/rails test test/models/vote_test.rb test/models/observation_test.rb test/models/name/propagate_generic_classifications_test.rb` — all green (no regressions).
- [x] Verified each of the four `config/recurring.yml` schedules parses via `SolidQueue::RecurringTask.from_configuration(...).valid?`.


nimmolo:
This takes awhile on production, and could easily be split up into different jobs so that it staggers.

```
real    1m41.062s
user    0m23.731s
sys     0m2.455s
```

mo-nathan (approved):
Looks good.  I'm not convinced that daily cleanup scripts like these are where we actually want to be, but that's a bigger issue.  It would be nice if the Intern could setup some sort of monitor for these to understand how much work they are really doing.  I'd rather see these get triggered by some event that implies that cleanup is needed or ideally eliminated entirely.

PR #4733, opened 2026-07-08
Port `check_rss_logs` script into 2 Solid Queue jobs, split by measured cost

nimmolo:
## Summary

From #4726 (crontab → Solid Queue recurring-job audit, Bucket A) — this job doesn't depend on any of the other Bucket A job PRs).

Ports `script/check_rss_logs` into Solid Queue jobs, bringing the Ruby logic directly into the job layer. The original was a raw `mysql` CLI script (bash, no Rails boot at all) — rewritten in ActiveRecord/Arel per `.claude/rules/no_raw_sql.md`. Scheduled via `config/recurring.yml`, on the `:maintenance` queue. Removes the corresponding line from `config/etc/crontab`.

Reuses `RssLog::ALL_TYPE_TAGS` (already defined on the model) instead of re-deriving the seven type columns by hand. The `notes NOT REGEXP '^[0-9]{14}'` / `notes REGEXP '^[0-9]{14}'` conditions translate directly to Arel's `matches_regexp`/`does_not_match_regexp` predicates — verified they produce the exact same `REGEXP` SQL MySQL used before.

For each target type:
- an "orphan" note (doesn't start with a 14-digit timestamp) whose type_id points at a row that's actually gone gets that type_id nulled out.
- a "real" (timestamped) note whose type_id points at a deleted target gets the whole row deleted.

Both checks share a `dangling_reference_query` liveness guard (`where.not(column => ref_model.all)`) — a type_id is only touched when the referenced row genuinely no longer exists. This matters because of #4763/#4764: roughly 845 observations were falsely orphaned by a 2016 mass-deletion malfunction and are still live, so an orphan-looking note doesn't always mean the target is gone. Without the guard this job would null those still-live observations' type_id, severing their only reverse link to their own RSS history.

Plus a ghost check: timestamped notes with all seven type_id columns nil.

**Kept the dangling-type_id deletion check deliberately**, even though it duplicates part of what `CheckForBrokenReferencesJob` (#4732) already covers for `RssLog` on a weekly cadence — this job runs daily, so a dangling reference gets caught same-day instead of waiting up to a week.

Deletes `script/check_rss_logs` — no other references remain.

## Timing: script vs. job

Production was taking ~3m40s per run on the old script (224s, then 218s on a repeat run the next day — consistent, not a one-time backlog cost). Measured both the old script and the ported job locally, same DB copy, both timed as full processes (`time script/check_rss_logs` vs `time bin/rails runner "CheckRssLogsJob.perform_now(...)"`, so both include Rails boot):

```
script/check_rss_logs (bash):  47.97s
CheckRssLogsJob (Ruby):          9.82s
                                -------
ratio:                          4.89x
```

The job is roughly 4.9x faster locally, most likely because it replaces ~30 separate `mysql` CLI process spawns (7 types × 4 queries + 2 ghost queries, each its own connection) with a single persistent ActiveRecord connection — confirmed via `EXPLAIN` that the query plans themselves aren't meaningfully different (the `NOT IN` subquery this port preserves 1:1 already gets materialized with an indexed lookup by this MySQL version, so an anti-join rewrite wouldn't help).

## Split by measured per-type cost: observation vs. everything else

Added always-on (not gated behind `verbose:`) elapsed-time logging around each type's queries to find out which types actually dominate the runtime, rather than guessing. Locally, `:observation`'s two checks alone were ~46% of the job's total time — `observations` is by far the largest of the seven reference tables (605K rows locally vs. 70K for the next largest, `names`), so both its orphan-null check and its bogus-type-delete check cost proportionally more.

Split into two jobs so the six cheap types don't queue behind observation's disproportionate share:

- `CheckObservationRssLogsJob` — just `:observation`.
- `CheckOtherRssLogsJob` — the other six types, plus the ghost-row check (not type-specific, so it only needs to run once across the two jobs).

`CheckRssLogsJob` itself is now an abstract base holding the shared orphan/bogus-type/ghost logic (arel predicates, the per-query timing instrumentation, log formatting) — never scheduled directly; `#check_types` and `#check_ghosts?` are the two hooks subclasses override. `config/recurring.yml`'s single slot became two entries, staggered 5 minutes apart and off `recover_stuck_inat_imports`' every-3-minutes marks, same reasoning as the existing stagger comment on the jobs above it.

## Test plan

- [x] `bundle exec rubocop` clean on all touched `.rb` files.
- [x] `bin/rails test test/jobs/check_rss_logs_job_test.rb test/jobs/check_observation_rss_logs_job_test.rb test/jobs/check_other_rss_logs_job_test.rb` — the abstract base's two hook defaults, `CheckObservationRssLogsJob`'s orphan/bogus/dry-run/no-ghost-check behavior, `CheckOtherRssLogsJob`'s orphan/bogus/ghost/no-observation-check behavior, and (one regression test per concrete job) that a still-alive target's type_id survives a run.
- [x] `bin/rails test test/jobs/` — 99 tests, all green (no regressions).
- [x] `bin/rails test test/jobs/queue_config_test.rb` — confirms both new jobs' `:maintenance` queue is covered by a worker.
- [x] Verified both `config/recurring.yml` schedules parse via `SolidQueue::RecurringTask.from_configuration(...).valid?`.


nimmolo:
This script is expensive.

```
real    3m44.454s
user    0m0.408s
sys     0m0.258s
```

nimmolo:
Needs to be adjusted for #4764

Issue #3589, opened 2025-12-13
Evaluate the current code base for thread safety when in production


mo-nathan:
We would like to enable multi-threaded execution in our production environment, but we need to analyze the current code base to see how much work this will be and to create a plan for achieving this goal.

nimmolo:
Follow-up on the Dec 2025 thread safety analysis linked above — re-verified the remaining class-variable issues against current `main` before planning next steps.

## Updated findings

- `app/models/queued_email.rb` is gone (removed in "Remove `QueuedEmail` everywhere"), so its `@@all_flavors` item from the report is moot.
- `app/models/language.rb`'s `@@last_update` is dead code — its only reader, `self.update_recent_translations`, is already fully commented out with no live caller anywhere in the app. This is a deletion, not a thread-safety refactor.
- `app/classes/run_level.rb` is currently inert — `RunLevel.silent`/`.normal` have zero callers anywhere in `app/`/`script/`/`test/` (they used to be called from `QueuedEmail`, which is gone). `RunLevel.is_normal?` has exactly one caller (`AbstractModel:649`) and always returns `true` today. Decision: drop it entirely rather than fix its thread-safety — it's dead weight, not a mechanism worth preserving.
- Several stale WIP branches already exist from around the original report (`thread-safe-run-level`, `thread-safe-textile`, `thread-safe-test-helper`, `thread-safe-test-language-extensions`, `thread-safe-controller-tests`), all unmerged. The two directly relevant ones took the wrong approach: they convert `@@foo` (class variable) to `@foo` (class instance variable via `self.foo`). A class-instance variable is still one shared value per class, read/written by every thread in the process — this fixes Ruby's "class variables leak into subclasses" quirk, not the actual race condition. They shouldn't be resurrected as-is. The test-side branches are additionally too stale to use even as reference (they predate `component_test_case`/`mailer_test_case`).

## Per-file plan (5 real items, not 8)

1. **`app/models/user_group.rb`** (HIGH — real race on cache init/reset) — convert `@@all_users`/`@@one_users`/`@@reviewers` to `Rails.cache.fetch(...)`; convert `clear_cache_for_unit_tests`'s reset to `Rails.cache.delete(...)`.
[...]

PR #4743, opened 2026-07-09
Convert UserGroup's meta-group caching to Rails.cache/Concurrent::Map

nimmolo:
## Summary

Part of the #3589 thread-safety cleanup. `@@all_users`/`@@one_users`/`@@reviewers` cache DB-backed meta-groups in class variables — concurrent threads could race during cache initialization, triggering multiple DB queries that overwrite each other's cached values.

~~`all_users`/`reviewers` (single-value caches) convert to `Rails.cache.fetch`. `one_user` (keyed per user id) needs a real thread-safe map rather than a single cache slot — converts to `Concurrent::Map`...~~ Superseded below.

Also removes `app/models/user_group.rb` from `.rubocop_todo.yml`'s `Style/ClassVars` exclude list, now that there's nothing left to exclude. Adds `test/models/user_group_test.rb` (didn't exist before) covering `all_users`/`one_user`/`reviewers`.

## Fix: index the column, drop caching entirely (review feedback)

Review caught that both `Rails.cache` and the class-var it replaced were treating a symptom. `all_users`/`reviewers`/`one_user` are only worth caching because `find_by_name` is slow, and it's slow only because `user_groups.name` has no index — every miss is a full table scan (~158K rows: one `"user #{id}"` group per user, plus meta-groups).

Added a unique index on `user_groups.name` (names are 1:1 with groups by the model's own doc comment) and deleted all caching — `Rails.cache.fetch`, `expires_in`, `Concurrent::Map`, `clear_cache_for_unit_tests`. With the index, `find_by_name` is a microsecond indexed lookup either way, so there's nothing left to cache — and no caching means no shared class state at all, which is the actual thread-safety objective.

One pre-existing duplicate name blocked the unique index: project #231 ("MMHC Mycoflora Project")'s member and admin groups were both named identically, missing the `.admin` suffix every other project's admin group uses. The migration renames it to match the established convention before adding the index.

Replaced the thread-safety test (which was proving `Concurrent::Map`'s atomicity — no longer applicable with no cache) with a test proving concurrent `one_user` calls each get back the correct group.

## Follow-up: request-scoped memo (review feedback)

Nathan pushed back on dropping caching entirely: `all_users`/`reviewers`/`one_user` are called many times within a single request — once per description in a name/location's reader/writer/admin-group listing (`Description::DetailsAndAltsPanel`), not just once. Confirmed empirically: 0 calls on the observations index (matrix boxes don't render description permission info at all), but 2 calls on the observation show page for a name with 2 descriptions.

Even with the unique index making each `find_by_name` call cheap, a lightweight per-request memo avoids the repeat work within that one request. Added one using `Thread.current[...]` — matching `Textile`'s name-lookup cache pattern (#4741) rather than a class ivar or `Rails.cache`, so it's isolated per request/thread and doesn't reintroduce the cross-request state this PR just eliminated.

Reset via a new `ApplicationController` before_action (`reset_user_group_cache`) so a request that never touches these doesn't inherit stale groups from whatever request ran on the same pooled thread before it — the same landmine pattern fixed for `Textile` in #4741. Also reset in `test_helper.rb`'s `setup` for unit tests that call these methods directly without a full request.

## Test plan

- [x] `bundle exec rubocop app/models/user_group.rb test/models/user_group_test.rb test/test_helper.rb db/migrate/20260711120000_add_unique_index_to_user_groups_name.rb` — no offenses
- [x] `bin/rails db:migrate` — index added; verified the duplicate-name rename ran correctly (project #231's admin group renamed to `"MMHC Mycoflora Project.admin"`)
- [x] `bin/rails test test/models/user_group_test.rb` — 4 tests, 0 failures
- [x] `bin/rails test test/mailers/location_change_mailer_test.rb test/mailers/name_change_mailer_test.rb test/models/description_test.rb test/models/user_test.rb test/controllers/projects_controller_test.rb test/controllers/account_controller_test.rb test/controllers/names/descriptions_controller_test.rb test/controllers/names/descriptions/publish_controller_test.rb test/controllers/names/descriptions/merges_controller_test.rb test/controllers/locations/descriptions/merges_controller_test.rb test/views/controllers/descriptions/permissions/form_test.rb` — 185 tests, 0 failures (every caller of `UserGroup` class methods)
- [x] `bin/rails test test/models/project_test.rb` — 60 tests, 0 failures
- [x] `bundle exec rubocop app/models/user_group.rb app/controllers/application_controller.rb test/models/user_group_test.rb test/controllers/application_controller_test.rb test/test_helper.rb` — no offenses
- [x] `bin/rails test test/models/user_group_test.rb test/controllers/application_controller_test.rb` — 22 tests, 0 failures — new memoization tests (stub-based call counts) plus a request-boundary regression test proving the memo resets between requests, mirroring #4741's Textile test
- [x] `bin/rails test test/models/ test/controllers/application_controller_test.rb test/classes/textile_test.rb` — 986 tests, 0 failures (post-merge with main, which now includes #4741/#4744)


mo-nathan (changes requested):
Reading this over it struck me as an odd thing to be using Rails.cache for vs. either memoizing or making more efficient in at the database.  I asked the Intern about it and it suggested adding an index and ditching the caching.

In full Internese:

index user_groups.name instead of caching around a full table scan

The thread-safety goal is right, but I think both the old class-vars and the new cache are treating a symptom. The reason all_users/reviewers/one_user are worth caching at all is that find_by_name is slow — and it's slow only because user_groups.name has no index. The table has ~158K rows (one "user #{id}" group per user, ~95K users, plus the meta-groups), so every miss is a full 158K-row scan.

Fix the cause and the caching layer becomes unnecessary:

1. Add a unique index on user_groups.name. Names are 1:1 with groups, so this is almost certainly an oversight. It turns find_by_name into a microsecond indexed lookup for every caller — cached or not.
2. Then drop the caching and derive on demand. With the index, all_users/reviewers/one_user can just do their find_by_name each call. That's inherently thread-safe (no shared class state at all — which is the actual objective here), with none of the follow-on complexity: no Rails.cache, no expires_in, no Concurrent::Map, no clear_cache_for_unit_tests, no marshaling UserGroup records into the cache.

A few specific reasons the current caching isn't pulling its weight even on its own terms:
- The cache store is SolidCache (DB-backed), so a "cache hit" is itself an indexed DB round-trip plus an unmarshal — barely different from an indexed find_by_name — and on a miss it still runs the 158K scan. So it mostly swaps one indexed DB lookup for another while adding machinery.
- expires_in: 1.hour introduces up to an hour of staleness that the previous code didn't meaningfully have.
- The Concurrent::Map for one_user can grow to ~95K entries per process over its lifetime (one per user ever accessed) — unbounded-ish memory that on-demand derivation avoids.

If profiling later shows all_users/reviewers are called many times within a single request (they're used in description permission checks), the right add is a lightweight request-scoped memo — not a cross-request cache. But with the index in place I'd expect the plain indexed lookup to be cheap enough to skip even that.

Net: add the index, delete the cache (old and new). That resolves the thread-safety concern more simply than either version and removes the slowness for every caller, not just the cached paths.

nimmolo:
Added the memoization because it is called more than once on show obs and show name.

PR #4745, opened 2026-07-09
Fix thread safety of Symbol's missing-tags and Language's tracking

nimmolo:
## Summary

Part of the #3589 thread-safety cleanup — the last of the six items from that audit.

~~`Symbol.missing_tags` (`@@missing_tags`)... converts to a `Concurrent::Array`...~~ Superseded below — review found this half over-engineered.

`LanguageTracking`'s `@@tags_used` is a different shape: it's primed by `ApplicationController#track_translations` per-request and read back later in that same request via `note_usage_of_tag`/`tags_used`/`save_tags` — request-scoped state, not global. A `Concurrent::Hash` shared across threads would still leak tags between concurrent unrelated requests, so this converts to `Thread.current[...]` instead, matching `Textile`'s name-lookup cache (`app/classes/textile.rb`) for the same reason. `@@last_clean` (throttles the temp-file cleanup sweep) IS genuinely global/cross-request, so it becomes a plain class-level instance variable on `Language` instead. **Nathan/Intern review confirmed this half is correct as-is — no changes requested.**

Adds `test/models/concerns/language_tracking_test.rb` (no unit test existed before, only a Capybara integration test) covering the basic `track_usage`/`note_usage_of_tag`/`tags_used`/`ignore_usage` contract plus a concurrent stress test proving `tags_used` is thread-isolated.

Also removes `app/extensions/symbol.rb` and `app/models/concerns/language_tracking.rb` from `.rubocop_todo.yml`'s `Style/ClassVars` exclude list (`symbol.rb`'s other, unrelated exclude entries for different cops are untouched).

## Fix: revert missing_tags to a plain Array (review feedback)

Review caught that the `missing_tags` half was over-engineered for what's actually a test-only feature. Every setter/reader of `Symbol.missing_tags` lives under `test/` (`test_helper.rb`, `session_extensions.rb`) — in production the setter is never called, so the `instance_variable_defined?` guard in `Symbol#l` is always false and the `<<` never fires. It's completely inert outside a single-threaded test worker, so `Concurrent::Array` added synchronization overhead to the single hottest method in the app (`Symbol#l`) for a diagnostic that's never touched concurrently in practice. The real driver was the `Style/ClassVars` exclusion removal, not an actual thread-safety need.

Reverted to a plain class-instance-var `Array` — `Style/ClassVars` still passes (it's not a `@@` class variable), with zero `Concurrent` overhead in the app's hottest path.

## Test plan

- [x] `bundle exec rubocop app/extensions/symbol.rb app/models/concerns/language_tracking.rb test/models/concerns/language_tracking_test.rb` — no offenses
- [x] `bin/rails test test/classes/symbol_extensions_test.rb test/integration/capybara/language_tracking_integration_test.rb test/controllers/translations_controller_test.rb test/models/concerns/language_tracking_test.rb` — 65 tests, 0 failures
- [x] `bin/rails test test/classes/` — 1913 tests, 0 failures (broad sanity check — every test's teardown asserts on `Symbol.missing_tags`)


JoeCohen (commented):
Approve (provided merge conflicts are fixed)

mo-nathan (changes requested):
This one appears to be half right.  The Intern thinks there are much better way to do the missing_tags bit and implies a small speed increase for "the single hottest method in the app".  The tags_used, on the other hand, is actually legit and this is a reasonable approach.

More details,
the tags_used change is right; the missing_tags change is over-engineered for a test-only feature

This PR bundles two unrelated conversions with very different justifications. One is warranted, one isn't.

Keep: LanguageTracking#tags_used → Thread.current. This is genuine per-request state that's live in production. track_translations is a before_action (application_controller.rb:55) that turns tracking on for logged-in users viewing a non-official locale or reviewers (@user && @language && (!@language.official || reviewer?)); as the page renders, each Symbol#l records its tag, save_tags writes them to a temp file, and the Translations editor reads them back (translations_controller.rb:220) to show a translator which strings the previous page used. The old @@tags_used genuinely races here — a concurrent non-tracking request running ignore_usage (setting it nil) would wipe a tracker's in-progress list, and two simultaneous translators would cross-contaminate. Thread.current is the correct isolation (matches the Textile per-request pattern), and last_clean → a class-level instance var is fine since it's a genuinely global cleanup throttle. No change requested on this half.

Drop the Concurrent::Array for Symbol.missing_tags. This one is test-only. Every setter/reader lives under test/ (test_helper.rb:201/207 sets it to [] before each test and asserts it's empty after; session_extensions.rb likewise). In production the setter is never called, so @missing_tags is never defined, the guard (instance_variable_defined?(:@missing_tags)) is false, and the << never fires — it's completely inert. So this puts a Concurrent::Array plus guard machinery into Symbol#l (the single hottest method in the app) purely for a test diagnostic, and the only place it's ever touched concurrently is system tests (app-running-in-a-thread) — a test concern, not a production race. The real driver is the Style/ClassVars exclusion being removed, not an actual thread-safety need.

Cheaper ways to satisfy that without the production complexity:
- Minimal: a plain class-level instance var holding a plain Array (@missing_tags ||= []) clears Style/ClassVars with zero Concurrent overhead — appropriate since prod never touches it and unit tests are single-threaded per worker.
- Better: move the missing-tag detection out of production Symbol#l entirely. I18n already has a missing-translation hook (I18n.exception_handler / raise-on-missing); let the test framework detect untranslated tags through that instead of having #l accumulate onto the Symbol class. That deletes the production branch rather than making it thread-safe.

So: keep the tags_used/last_clean changes as-is; swap the missing_tags Concurrent::Array for a plain class-instance-var Array (or, preferably, relocate that detection to I18n's missing-translation handling so Symbol#l stays clean).

Issue #4735, opened 2026-07-09
Job-ify `Image#process_image` (currently blocks the upload request synchronously)


nimmolo:
## Problem

`Image#process_image` (`app/models/image.rb:764`) runs **synchronously**, blocking the HTTP request/response cycle, on every image upload — including the observation-create path (`app/controllers/observations_controller/shared_form_methods.rb`), `application_controller.rb`, `images/uploads_controller.rb`, and `image_api.rb`. It shells out to `script/process_image`, which does:

1. Convert original to JPEG if needed
2. Strip GPS via exiftool (if hiding coordinates)
3. Set width/height in the database
4. Generate 5 resized copies via ImageMagick (`convert`)
5. Transfer all files to the image server(s) via scp/rsync
6. Update `transferred` + touch related observations
7. Email webmaster on error

That's a genuinely heavy, slow chain (multiple ImageMagick conversions plus a network file transfer) sitting inline in the upload request.

`Image#transform` (rotate/mirror, calls `script/rotate_image`) is already backgrounded via a bare shell `&` trick (`system("script/rotate_image #{id} #{operator}&")`) — but with no Solid Queue supervision, no retry, and no observability.

## Tradeoff

Main tradeoff to design around: if processing goes async, the image's resized/GPS-stripped versions aren't ready the instant the upload request returns, so the UI needs to handle a "processing" state gracefully (placeholder / poll / turbo-stream update) rather than assuming everything is ready synchronously — matches the pattern already established by `ImageDhashJob` for perceptual-hash computation.

Debate about the desirability of this is invited.

## Prior art: PR #2135

[`nimmo-process-image-ruby` / PR #2135](https://github.com/MushroomObserver/mushroom-observer/pull/2135) (draft, opened 2024-05, never merged) replaces the bash script chain with a pure-Ruby `Image::Processor` class using the `image_processing`/`mini_magick` (ImageMagick wrapper), `mini_exiftool_vendored` (EXIF), and `rsync` gems.

**It does not solve the synchronous-blocking problem.** `Image#process_image` still calls `processor.process` synchronously in that PR — the goal was maintainability/testability (per the PR's own description), not async execution.

[...]

PR #4747, opened 2026-07-09
Port image processing to Ruby: Image::Processor (#4735 PR 1/4)

nimmolo:
## Summary

First PR in the job-ify-image-uploads plan for #4735 (full 4-PR breakdown posted as a comment on that issue). Ports the resize/reorient/transfer logic from `script/process_image` into a real Ruby class, `Image::Processor`, based on the draft in #2135.

The real motivation isn't just "no more shelling out" -- `script/process_image` flips `width`/`height`/`transferred` via a raw `mysql` `UPDATE`, so `after_update_commit` never fires today even if we added a callback. Writing through ActiveRecord (`image.update`) is what makes that possible, and it's a precondition for the Turbo Stream broadcast work planned later in this series, not a separate concern.

`Image#process_image` now calls `Image::Processor#process` directly instead of building and shelling out to `MO.process_image_command` (removed, now dead). Still skipped in test env exactly as before -- `Image::Processor` gets its own fast, direct unit tests instead of going through that guard.

`rotate` and `self.retransfer_images` are ported too (mirroring script/rotate_image and script/retransfer_images) but aren't wired to any call site or job yet -- that's the next PR, along with a from-scratch `verify_images` port.

### #2135 had never actually run

I used #2135's draft as the starting point, but close reading + writing real tests turned up quite a bit more than the two bugs I'd originally scoped in from a first read. None of this is a knock on that draft -- it was exploratory and never merged -- but worth being upfront that "port with the known bugs fixed" undersold the gap once I actually ran it against real files:

- A dead guard line in the rotate path (`nil unless operations.include?(orientation)` doesn't return/guard anything).
- `self.retransfer_images` read `@transferred_any` on `self` (the class) instead of on the `processor` instance that actually sets it -- the transferred flag was never marked.
- `mark_image_record_transferred_and_touch_obs` was `private` but called on a different instance's object from `self.retransfer_images` -- would raise `NoMethodError` the moment that path ran.
- `IMAGE_SERVER_DATA` was assigned in the class body by calling `image_server_data` (an *instance* method, defined *after* the assignment) -- the class couldn't even load.
- `IMAGE_SERVERS` included `:local`, so the transfer loop tried to copy local files onto themselves.
- The `ssh://` write-target path dropped the required `:` between `user@host` and the remote path, which rsync needs to know it's a remote target at all.
- `*_file` vs `*_filepath` -- `convert_source_to_destination` and `transform_full_size_file` called undefined `*_file` methods; only `*_filepath` was ever defined.
- `image_config.yml`'s `:sizes:` lists size names (`:thumbnail`) but the transfer code compares against subdirectory strings (`"thumb"`) -- direct comparison never matched, so `remote2`'s narrower size list silently transferred nothing.
- GPS stripping used `MiniExiftool#[]=`/`#save` with pseudo-tags like `"GPS:all"` -- these aren't real tags MiniExiftool recognizes, so `save` returns `false` and (since the return value was never checked) the file is left untouched. Fixed by shelling out to `exiftool` directly for this one operation, matching what `script/process_image` already did.

Extracted the server-config building (`IMAGE_SERVER_DATA`, the size↔subdir translation, the ssh path-joining) into `Image::Processor::ServerData` to keep `Image::Processor` itself under the class-length cop and separate "what servers exist" from "process this one image."

New gems: `image_processing`, `mini_magick`, `mini_exiftool_vendored`, `rsync`.

### Update: fixed a parallel-test-worker race in the path constants

`LOCAL_IMAGES_PATH`/`IMAGE_SERVER_DATA`/`IMAGE_SERVERS` were originally frozen constants computed once at class-load time from `MO.local_image_files`/`MO.image_sources`. Those two are deliberately lazy methods (see their comments in `config/consts.rb`) specifically so each parallel test worker gets its own suffixed directory (`test_images-3` vs `test_images-7`, etc.) -- caching their result in a frozen constant meant whichever forked worker happened to autoload `Image::Processor` before its own `TEST_ENV_NUMBER` was set would permanently lock in the wrong (unsuffixed) path for that worker's process lifetime, colliding with every other worker writing to the same shared directory.

I found this while working on the follow-up PR, which adds a second test file also referencing these constants -- with only this PR's own test file in the mix it doesn't reproduce in practice (stress-tested: 4 runs of 946 tests, 0 failures), but the design is unsafe under a different test distribution (CI's own parallelization included), and it's this class's own flaw regardless of which PR happened to expose it. Converted to plain class methods (`local_images_path`, `image_server_data`, `image_servers`) that recompute on every call.

## Test plan

- [x] `bin/rails test test/models/image/processor_test.rb` -- new, direct coverage of `process` (jpg + non-jpg/tiff conversion), GPS stripping (verified against a real geotagged fixture), `rotate` (including fetching the original back from a remote when missing locally), `self.retransfer_images`, and the `ServerData` helpers
- [x] `bin/rails test test/models/image_test.rb` -- existing `process_image` coverage, updated the one test that stubbed the old shell command
- [x] `bin/rails test test/classes/image_script_test.rb` -- shell-script integration tests unaffected (scripts still exist, untouched)
- [x] `bin/rails test test/controllers/observations_controller_create_test.rb test/controllers/observations_controller_update_test.rb` -- upload/update paths that call `process_image`
- [x] Stress-tested the parallel-worker-race fix: `bin/rails test test/models/image/processor_test.rb test/models/` (946 tests, full 12-way parallelization) x4, 0 failures
- [x] `bundle exec rubocop` clean on all new/changed files

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -132,6 +132,16 @@ gem("xmlrpc")
 gem("fastimage")
 # for detecting file type of uploaded images
 gem("mimemagic")
+# An interface between Ruby and ImageMagick/Vips, used to resize/reorient
+# uploaded images
+gem("image_processing")
+gem("mini_magick", "~> 5.0")
+# Reads and writes EXIF/GPS data; "vendors" a recent exiftool so we don't
+# depend on a system install
+gem("mini_exiftool_vendored")
+# Wraps the rsync binary, used to transfer processed images to the image
+# server(s)
+gem("rsync")
 
 # Gems used for iNat import
 gem("oauth2")
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -240,6 +240,7 @@ GEM
       domain_name (~> 0.5)
     i18n (1.14.8)
       concurrent-ruby (~> 1.0)
+    image_processing (2.0.2)
     importmap-rails (2.2.3)
       actionpack (>= 6.0.0)
       activesupport (>= 6.0.0)
@@ -285,6 +286,13 @@ GEM
     mimemagic (0.4.3)
       nokogiri (~> 1)
       rake
+    mini_exiftool (2.14.0)
+      ostruct (>= 0.6.0)
+      pstore (>= 0.1.3)
+    mini_exiftool_vendored (9.2.7.v1)
+      mini_exiftool (>= 1.6.0)
+    mini_magick (5.3.1)
+      logger
     mini_mime (1.1.5)
     mini_racer (0.21.0)
       libv8-node (~> 24.12.0.1)
@@ -346,6 +354,7 @@ GEM
       snaky_hash (~> 2.0, >= 2.0.4)
       version_gem (~> 1.1, >= 1.1.9)
     os (1.1.4)
+    ostruct (0.6.3)
     parallel (2.1.0)
     parser (3.3.11.1)
       ast (~> 2.4.1)
@@ -375,6 +384,7 @@ GEM
       rexml (>= 3.4.2, < 4)
     prettyprint (0.2.0)
     prism (1.9.0)
+    pstore (0.2.1)
     psych (5.3.1)
       date
       stringio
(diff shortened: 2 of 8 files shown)


nimmolo:
Combined into #4751 along with #4748 (tightly coupled, no longer need separate review units). Closing in favor of that.

PR #4748, opened 2026-07-09
Add RetransferImagesJob, VerifyImagesJob, RotateImageJob (#4735 PR 2/4)

nimmolo:
## Summary

Second PR in the job-ify-image-uploads plan for #4735 (stacked on #4747 -- `Image::Processor`, still open). Converts the three remaining shell scripts to real, scheduled Solid Queue jobs, using `Image::Processor` from the first PR.

- `RetransferImagesJob` wraps `Image::Processor.retransfer_images` (already ported in #4747) -- a thin recurring wrapper, no new logic.
- `VerifyImagesJob` wraps a fresh Ruby port of `script/verify_images`: `Image::Processor::Verifier` lists every local and remote image file (by subdir/filename => byte size), uploads anything missing or mismatched on a server that's meant to have it, then deletes local copies once they're confirmed to match on every server that carries that subdirectory -- skipping any size in `MO.keep_these_image_sizes_local`, which never gets deleted locally. Tested against the exact same fixture data `test/classes/image_script_test.rb`'s shell-script test uses, but asserting on resulting file state and the returned `{uploaded:, deleted:}` rather than replicating the shell script's exact log-line output.
- `RotateImageJob` replaces `Image#transform`'s bare `system("script/rotate_image ... &")` fire-and-forget. That also let me drop the hardcoded 5-minute delay before re-hashing (a guess at how long the untracked background script would take) -- since the job is awaited, it enqueues `ImageDhashJob` immediately after `Image::Processor#rotate` actually finishes, a real completion signal instead of a guess.

Extracted the low-level per-file copy logic (`copy_file_to_server` et al) out of `Image::Processor` into `Image::Processor::FileTransfer`, a stateless module, so `Verifier`'s bulk sync and `Processor`'s per-image transfers share one implementation instead of risking two copies of the same rsync/scp logic drifting apart.

*(A parallel-test-worker race I originally found and fixed while building this PR turned out to be a defect in `Image::Processor` itself, not specific to anything added here -- moved to #4747, see that PR's description. This PR is based on top of that fix.)*

### Schedule

Registered both new recurring jobs in `config/recurring.yml`:

```
retransfer_images:  every day at 12:10am
verify_images:      every day at 1:01am
```

Staggered off the existing midnight maintenance jobs and off the every-3-minutes inat-import recovery job's `:00`/`:03`/`:06`... marks (same reasoning as #4746 earlier this week). Verified both parse to the intended cron expressions via `Fugit.parse(...).to_cron_s`.

## Test plan

- [x] `bin/rails test test/models/image/processor/verifier_test.rb test/jobs/retransfer_images_job_test.rb test/jobs/verify_images_job_test.rb test/jobs/rotate_image_job_test.rb` -- new coverage
- [x] `bin/rails test test/models/image_test.rb test/models/image/` -- updated `transform` tests, rest unaffected
- [x] `bin/rails test test/classes/image_script_test.rb test/controllers/observations_controller_create_test.rb test/controllers/observations_controller_update_test.rb test/controllers/images/` -- unaffected
- [x] `ruby -ryaml -e "YAML.load_file('config/recurring.yml')"` -- valid YAML
- [x] `bundle exec rubocop` clean on all new/changed files

### DIFF
--- a/app/jobs/retransfer_images_job.rb
+++ b/app/jobs/retransfer_images_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# Recurring safety net: re-transfer any image still marked untransferred
+# (e.g. a process that died mid-upload before reaching the image server).
+class RetransferImagesJob < ApplicationJob
+  queue_as(:maintenance)
+
+  def perform
+    Image::Processor.retransfer_images
+  end
+end
--- a/app/jobs/rotate_image_job.rb
+++ b/app/jobs/rotate_image_job.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Rotates/mirrors an image outside the request cycle. Replaces the old
+# bare `system("script/rotate_image ... &")` fire-and-forget: since this
+# runs as a real, awaited job, ImageDhashJob can be enqueued right after
+# rotate finishes instead of guessing at a fixed delay.
+class RotateImageJob < ApplicationJob
+  queue_as(:default)
+
+  def perform(image_id, ext, orientation)
+    image = Image.find_by(id: image_id)
+    return unless image
+
+    Image::Processor.new(image: image, ext: ext).rotate(orientation)
+    ImageDhashJob.perform_later(image_id)
+  end
+end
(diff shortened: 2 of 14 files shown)


nimmolo:
Combined into #4751 along with #4747 (tightly coupled, no longer need separate review units). Closing in favor of that.

PR #4755, opened 2026-07-09
Move import-source notice to top of Details panel

nimmolo:
## Summary

The `ImportedSourceBanner` on the observation show page was a full-width panel rendered above the carousel/details row. On small screens this gives it higher priority than the observation itself, which seems inappropriate for an attribution.

This PR moves it into the top of the observation details panel as a second `.panel-body` (with `border-bottom` to separate it from the main body), holding the same credit link + help icon as before. Still prominent and unmissable, but leaves the images up top. `Views::Controllers::Observations::Show::Details::ImportSource` is the new component; the old `ImportedSourceBanner` file and test are deleted.

Screenshot 2026-07-09 at 4 42 15 PM

## Rename: `ObservationDetailsPanel` -> `Details`

While in there, renamed `ObservationDetailsPanel` to `Details` and nested its four sub-panels under it, since none of them are actually panels in their own right (they're just row-groups inside the one real `Details` panel) and they're only ever rendered from within it:

- `CollectionNumbersPanel` -> `Details::CollectionNumbers`
- `HerbariumRecordsPanel` -> `Details::HerbariumRecords`
- `SequencesPanel` -> `Details::Sequences`
- `ExternalLinksPanel` -> `Details::ExternalLinks`

Callers within `Details` itself (and `show.rb`, which nests `Details` the same way) reference these unqualified now -- they're direct nested constants of the referencing class, not sibling references, which is the safe case for Zeitwerk autoloading (see `.claude/rules/phlex_reference.md`'s constant-resolution guidance). External callers -- the two naming-form pages and the four sub-record CRUD controllers' turbo-stream responses -- keep full qualification, since they're not nested under `Details`.

## Test plan

- [x] `bin/rails test` -- full suite green
- [x] New `Details::ImportSourceTest` (revived/renamed from the deleted `ImportedSourceBannerTest`) plus two new `DetailsTest` cases covering the conditional body-slot wiring (present + border-bottom for an imported obs, absent for a non-imported one)
- [x] Browser-checked the rendered obs-show page


mo-nathan (changes requested):
Please don't merge this until we have a chance to talk about it

JoeCohen (commented):
I'm OK with this placement and likely with whatever placement @mo-nathan wants.

nimmolo:
Merged into nimmo-inline-link-block-centralization for further work; closing this PR.

PR #4772, opened 2026-07-12
Parallelize `lang.rake` multi-language tasks; skip entirely if no change

nimmolo:
## Summary

This PR changes `lang:update` in two ways.

**1. Skips entirely when nothing changed.** `script/lang_update_if_needed.sh` (now what `script/deploy.sh` calls instead of `rake lang:update` directly) hashes `config/locales/en.txt` and compares it against the hash from the last successful run, cached in `tmp/lang_update_synced_hash` (gitignored, not committed). If the hash matches, it skips `rake lang:update` entirely -- pure overhead on the common case of a deploy that doesn't touch translations. When `en.txt` has actually changed, it runs `lang:update` exactly as before and then updates the cached hash.

**2. Parallelizes when it does run.** `bin/rails lang:update` (and `lang:check:all`, `lang:strip:all`, `lang:export:unofficial`, etc.) loop over every unofficial/all `Language` sequentially. Each language's export/import/update/strip/check work is fully independent, so this parallelizes cleanly -- and since these tasks run standalone (CI, a git hook, or a developer directly) with no Solid Queue workers running, they can't be background jobs; they have to parallelize in-process.

Adds `ConcurrentEachWithConnection` (`app/classes/concurrent_each_with_connection.rb`) -- a small, normal Zeitwerk-autoloaded class, not rake-file-local top-level `def`s -- that runs a block over a list of items on a fixed-size thread pool, each thread checking out its own `ActiveRecord` connection via `connection_pool.with_connection`. `lib/tasks/lang.rake`'s `:unofficial`/`:all` tasks now call into it instead of a plain sequential `each`.

Measured ~30% wall-clock reduction on `lang:update` when it does run: 11.8s sequential -> ~8.2-8.5s parallel, consistent across repeated runs.

## Error handling: every failure surfaces, not just the first

One item's failure doesn't stop the others from completing, but the overall call still fails loud once everything settles. If exactly one item fails, that exact error (class + message) is re-raised unchanged. If two or more fail -- plausible for a task touching 16 locales unattended during a deploy -- a single combined `RuntimeError` lists every one of them (`"N errors: ClassA: msg; ClassB: msg; ..."`) rather than silently dropping N-1 failures behind whichever happened to be collected first.

## Why a class instead of rake-file-local helpers

Rake `.rake` files aren't Zeitwerk-autoloaded, so top-level `def`s living in `lib/tasks/lang.rake` aren't reachable from an ordinary MiniTest file without manually loading the rake file -- not a viable pattern for a committed test. Extracting the pool logic into `app/classes/concurrent_each_with_connection.rb` makes it a normal, unit-testable class; `lib/tasks/lang.rake` builds a fresh instance per call (`lang_task_pool` method) rather than a top-level constant, since rake files are evaluated before the `:environment` task runs and Zeitwerk can't resolve the constant yet at file-load time. Not memoized -- the wrapper object only holds `pool_size` (an Integer); the real `Concurrent::FixedThreadPool` is created fresh inside every `#call`, so caching the wrapper saved nothing.

## Thread-safety of the parallelized work

Each `Language`'s export/import/update/strip/check work touches only its own instance variables and its own per-locale files. The only class-level state involved -- `Language.verbose`, `Language.safe_mode`, `Language.locales_dir` -- is set once by the `:setup` prerequisite chain (`environment`, `login`, `verbose`, `safe_mode` tasks) before any parallel work starts, and never written to during it, so concurrent reads are safe. (Traced every method reachable from the parallelized tasks by hand, not just the state I already knew about -- there's a separate, genuinely non-thread-safe memoized cache on `Language.for_locale`, but it's used by page-render helpers like `TranslatorsCredit`, not by anything in this call path.)

## Output ordering under `Language.verbose`

`lang.verbose(msg)` does a plain `puts` when `Language.verbose` is on (the default unless `silent=yes`). Running per-language work on a thread pool means interleaving across locales is *possible* in principle. Re-verified this properly: the original "byte-identical" claim in the test plan was checked with `silent=yes`, which suppresses `Language.verbose` entirely -- so that comparison couldn't have caught interleaving even if it happened, since there was no verbose output to interleave. Re-ran `bin/rails lang:update` (verbose, no `silent=yes`) three times in a row and captured the output each time: all three runs produced the exact same per-locale ordering, matching submission order (`pl, en, el, es, ru, pt, de, fr, it, tr, jp, zh, uk, ar, be, fa`), with no interleaved/out-of-order lines observed. This is an empirical property of this specific workload at `pool_size: 4` (short, roughly-uniform per-item work tends to drain a `Concurrent::FixedThreadPool`'s queue close to FIFO order), not a guarantee `ConcurrentEachWithConnection` makes -- interleaving remains possible in principle and would be harmless if it happened (each line is already locale-tagged, e.g. "Stripping fr", and nothing parses this task's stdout), just not what was actually observed.

## Test plan

- [x] `bin/rails test test/classes/concurrent_each_with_connection_test.rb` -- covers: runs the block for every item; defaults to `pool_size: 4`; each item gets its own active DB connection; single-item failure re-raises that exact error after every item finishes; **multiple differing failures raise one combined error naming all of them**; correctly processes more items than the pool size; requires a block; propagates an error raised while iterating `items` itself (not just inside posted work).
- [x] `bundle exec rubocop app/classes/concurrent_each_with_connection.rb test/classes/concurrent_each_with_connection_test.rb lib/tasks/lang.rake` -- clean.
- [x] `bin/rails test test/style/` -- clean.
- [x] `bin/rails lang:update` (verbose, **not** `silent=yes`) run 3x in a row against the dev DB -- succeeds every time, `git status` shows no unexpected tracked-file diff, per-locale output order identical and non-interleaved across all 3 runs (see "Output ordering" above for what this does and doesn't prove).
- [x] `bin/rails lang:check:all silent=yes` also exercises the parallel path successfully.
- [x] Timing comparison: sequential 11.820s vs. parallel 8.296s/8.139s/8.484s/8.208s across 4 runs.


nimmolo:
Related question that came up alongside this parallelization work: could we speed up deploy further by checking `config/locales/*.yml` into git and skipping `lang:update` at deploy time entirely? Investigated this — write-up below, since it's directly relevant to what this PR is optimizing.

## How translations actually flow today

- `en.txt` is the only hand-authored, git-tracked file. Every other locale file — `en.yml` and all non-English `.txt`/`.yml` — is a generated build artifact, gitignored since 2014 (commit `49c8720`, right when the DB-backed `acts_as_versioned` pipeline was finalized).
- **The database (`translation_strings` table) is the real source of truth**, not the files. `lang:update` is really five steps: `check:official` (validate `en.txt` syntax) → `import:official` (diff `en.txt` into the DB) → `strip:all` (delete DB rows for tags no longer in English) → `update:all` (regenerate every locale's `.yml` from the DB) → `export:unofficial` (regenerate every non-English `.txt` from the DB).
- Rails only ever loads `.yml` at boot (default `I18n.load_path` behavior — `.txt` is never read by the app itself).
- The live "edit this page's translations" feature (`TranslationsController#update`) writes directly to `TranslationString` rows, patches the current worker's in-memory `I18n.backend`, and then synchronously regenerates that locale's `.yml` and `.txt` on disk — in the same request, no queue, no git involved.
- `lang:update` runs in `script/deploy.sh` right after `db:migrate`, and in CI right before the test suite.

## Would checked-in ymls go out of sync with the DB?

Yes, immediately and by design — that's the whole point of the live-edit feature. The moment any logged-in contributor edits a translation in production, the DB (and that locale's on-disk `.yml`) diverges from whatever `.yml` snapshot is sitting in git. There's no path back from a production edit into git. So a checked-in `.yml` isn't "the current translations plus occasional drift" — it's a snapshot that starts stale the instant it's committed, for every non-English locale (which are edited almost exclusively through the live UI, never through hand-edited `.txt` + `lang:update import`).

## The actual risk of checking ymls in

The dangerous part isn't the divergence itself — it's what happens at the next deploy if you also drop the `lang:update` step to make that checkin "count":

- If deploy stops regenerating `.yml` from the DB, deploying would overwrite production's live-edited, DB-current `.yml` files with whatever stale snapshot a developer last committed — silently reverting every translator's edit since that commit. That directly contradicts the goal of letting users add/edit translation strings on production.
- If deploy still regenerates `.yml` from the DB (keeping today's correctness), then checking `.yml` into git buys you nothing at deploy time — you're back to running the exact same regeneration step, just with an extra, perpetually-stale copy of the output sitting in version control.

## Pros / cons of checking ymls into git anyway

**Pros:** diffable history for the one true hand-edited source path (though `en.txt` already gives you that); a "known-good fallback" `en.yml` if the DB is ever corrupted; slightly easier to eyeball what production last shipped.

**Cons:** every non-English `.yml`/`.txt` becomes noisy, frequently-conflicting generated diffs with no independent value (exactly why they were pulled from git in 2014); a second source of truth that can only be kept honest by discipline (a developer forgetting to regenerate+commit `en.yml` after editing `en.txt` is a new footgun, not a removed one); and it doesn't actually let you skip the deploy-time regeneration without the data-loss risk above.

## On the testing footgun specifically

Checking ymls in would make this worse, not better. The current footgun (`en.yml` stale relative to `en.txt` after a checkout/rebase, papered over by the `check_lang_update_before_test.sh` / `post_checkout_lang_db.sh` / `post_rebase_lang_db.sh` Claude Code hooks) exists precisely because the generated file isn't in git and has to be rebuilt on demand. If it were checked in, staleness would come from a worse source: two committed files (`en.txt` and `en.yml`) that a merge/rebase can trivially leave inconsistent with each other in a way that's syntactically valid but semantically wrong — and nothing would force a rebuild to catch it, since "the file is right there in git" removes the trigger the current hooks key off of (`en.txt`'s blob hash changing).

## A way to actually speed up deploy without any of this

Don't touch the git-tracking question at all — apply the same trick the local dev hooks already use: skip `lang:update` at deploy unless `en.txt` changed since the last deploy (a cached git-blob-hash comparison, same as `check_lang_update_before_test.sh` does for tests). Most deploys don't touch translations, so this eliminates the ~8-12s cost entirely on the common path with zero correctness risk — nothing regenerates, so nothing can revert a live edit, because nothing needed to change. On the rarer deploy that does touch `en.txt`, the full pipeline still runs exactly as today, benefiting from this PR's parallelization. That combination — skip-if-unchanged plus this PR's parallelization for when it does run — gets the deploy-speed win without reopening the DB-vs-git sync problem at all.

app/classes/concurrent_each_with_connection.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+# Runs a block once per item, up to `pool_size` at a time, each on its
+# own ActiveRecord connection. For work that has to parallelize
```
JoeCohen: Something missing, or a dangling modifier.

JoeCohen (approved):
The intern's comments are often garrulous, redundant, and -- for me -- unreadable.
That used to be a smell. Today, maybe that's the new world. 
See my comment and @mo-nathan's response at https://github.com/MushroomObserver/mushroom-observer/pull/4847#discussion_r3610679177
Also see [Add a code-comments rule: explain *why* (only when unclear), one source of truth 
](https://github.com/MushroomObserver/mushroom-observer/pull/4850)

app/classes/concurrent_each_with_connection.rb
nimmolo: Fixed in 11671069bb - that sentence was a dangling fragment ("For work that has to parallelize in-process..." had no subject/verb of its own). Rewrote as one complete sentence and tightened the surrounding comment while I was in there. Also removed the duplicated "why in-process, not a background job" rationale that had been repeated in `lib/tasks/lang.rake`'s comment (now just points back to this class), and scoped `lang_task_pool`'s "the only shared state is X" claim more precisely after re-checking it against a skeptical-review finding about `Language.for_locale`'s separate memoized cache.

Issue #3797, opened 2026-02-02
01: Bootstrap 3 -> 4 Migration


mo-nathan:
This migration has been underway for quite a while and will continue to occupy a reasonable amount of development attention. Bootstrap 3 is out of date and is no longer receiving security updates. There are no known security issues as a result of this lack of support, but we need to get off this legacy package and into a more current version. We expect once we get to Bootstrap 4, the migration to Bootstrap 5 will be easier. One of the biggest pieces of work is switching from ERB files for the frontend (original Ruby on Rails standard frontend package) to Phlex (a more maintainable, future-facing solution).

### Done

- **Icon library** — chosen and implemented. It's a home-grown SVG sprite (`vendor/assets/images/icons/mo-icons.svg`, built by the private `MushroomObserver/icon-library` repo), rendered via `<use>` in `Components::Icon`. No Glyphicon-name mapping remains anywhere in the app.
- **`col-xs-*` grid classes** — centralized behind `Components::Column`. Only 2 raw references remain, both intentional: `Components::Column` itself (the single source of truth — this is what the Grid-constants recommendation below was for), and one documented "keep the bug for visual parity" quirk in `app/views/layouts/header.rb`. The BS4 cutover is now a one-file change here, not a sweep.
- **BS3 responsive visibility helpers** (`hidden-xs`, `visible-*`) — centralized behind `Components::Column`'s `show_at:`/`hide_at:` kwargs and `.mobile_hide_classes`/`.mobile_only_classes` class methods, built on the `.d-*` bridge classes in `_utilities.scss`. 15 call sites use it. Two raw-class references remain outside this: `matrix-table_controller.js` (tracked below, part of the cutover PR) and a prose comment in `_icons.scss` (not a class reference). The BS4 cutover is now a one-file change here too, not an 18-file sweep.
- **Raw Bootstrap class sweeps** (`btn`, `panel`, `collapse`/`in`, `dropdown`, `modal`, `carousel`, `navbar`, `input-group`, `media`, `well`) — done. Full file-by-file trail is in the comments below.
[...]

PR #4775, opened 2026-07-12
Dummy icon-library comparison page for #3797

nimmolo:
## Summary

Dummy comparison page to help pick a replacement for Bootstrap 3 Glyphicons — the last "design decision required" item blocking further work on #3797. Not meant to merge to main; check it out locally, look at the page, pick a library, then this branch/PR can be closed and deleted.

## How to view it

1. Check out this branch and start the dev server (`bin/rails server`).
2. Log in as any user.
3. Visit **`http://localhost:3000/test_pages/icon_library_comparison`**.

You'll see all 63 semantic icon names MO currently uses (`Components::Icon::GLYPHS`), each rendered at large size in three columns: the current Bootstrap 3 Glyphicon, a candidate from Bootstrap Icons, and a candidate from Font Awesome Free (both loaded via CDN for this throwaway page, not vendored).

**The candidate name pairings are best guesses for visual comparison, not verified semantic matches** — that's the entire point of the page: look at the options side by side and pick, don't trust the guessed pairing.

## What happens once a library is picked

The actual migration is small and contained: `Components::Icon::GLYPHS`' values change from Glyphicon suffixes to the new library's names, and `Components::Icon#span_class` swaps the `glyphicon glyphicon-*` class template for whatever the chosen library expects — one file. Every call site in the app already goes through `Icon(type: ...)` Kit sugar, so nothing else needs to change.

## Test plan

- [x] Page renders successfully (200, all 63 rows) — verified via a throwaway functional test hitting the route directly.
- [x] `bundle exec rubocop` on all 3 touched files — clean.
- [x] `bin/rails test test/style/` — clean.
- No permanent test coverage added — this page is explicitly throwaway and gets deleted once the decision is made.


JoeCohen:
Mildly prefer Bootstrap because the qr code icon is closer.
I also wonder if it avoid some confusion to use Bootstrap icons with Bootstrap.

PR #4810, opened 2026-07-15
Force iNat confirm-form links to the UI host

### LINKED ISSUE #4809: confirm form links should target iNat UI
The target for links on the confirm form should be `www.inaturalist.org`, never `api.inaturalist.org`. 
Currently if the user provides an api url when importing, the target is the api.
To replicate:
- http://localhost:3000/inat_imports
- Check Import Others, Consent, Observations from an iNat search URL
- Fill in text field with `https://api.inaturalist.org/v1/observations?project_id=fundis-rocky-mountain-rare-fungi-challenge`
- Submit

JoeCohen:
Fixes #4809
See [Copilot Overview.](https://github.com/MushroomObserver/mushroom-observer/pull/4810#pullrequestreview-4705824622)

### DIFF
--- a/app/classes/inat/confirm_url_builder.rb
+++ b/app/classes/inat/confirm_url_builder.rb
@@ -37,6 +37,7 @@ class Inat::ConfirmURLBuilder
     requested_obs_url&.then { |u| "#{u}&licensed=false" }
   end
 
+  # Will the requested-obs URL always return the same results?
   def stable_result_set?
     return true if @model.inat_ids.present?
     return false unless (query = requested_obs_query)
--- a/app/classes/inat/confirm_url_builder.rb
+++ b/app/classes/inat/confirm_url_builder.rb
@@ -100,9 +100,10 @@ class Inat::ConfirmURLBuilder
       merge("iconic_taxa" => IMPORTABLE_ICONIC_TAXA_ARG).to_query
   end
 
+  # Always link to the iNat UI, never the API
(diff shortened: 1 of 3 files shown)


nimmolo (commented):
It seems like "import inat query within dates" would be one of the most likely natural user inclinations, but it is not allowed per the url validator. I can understand it not being considered stable, but i'm not sure what that means.

JoeCohen:
@nimmolo Thanks for flagging this. It's been on my mind. 
`stable` is a CC construct. I think it means something like "idempotentish". I think it was a way of addressing something which I saw as an issue, but probably should be ignored.
Issue: Links on the confirm form get a different number of hits than the displayed count. That will happen if iNat observations meeting the constraints get added/deleted after the count is created, but before the link is clicked.
Very unlikely for regular users
Superimporters should just live with it.

JoeCohen:
@nimmolo IMO, the behavior is correct. (I forgot that I'd already worked this through and tested it.)
But if I'm hallucinating, it would be great to have an example of bad behavior.
Expected behavior: 
- The confirm form displays an "As-of" time stamp.
- If a user-requested URL is not "stable", then:
  - the confirm form includes the warning "Actual counts may change after [the as-of] time."
  - The counts, urls, and actual items imported do not change.
- An iNat query with date params is not "stable" because its results could change;
observations could be added/deleted after the confirm form is displayed.
So clicking on a link will return more/fewer observations than the count shown on the confirm form.

PR #4824, opened 2026-07-17
Cache-bust image rendition URLs with an `updated_at` token in `Image::URL#source_url`

### LINKED ISSUE #4808: Image renditions are served from stable URLs, so rotate/mirror shows stale images until a hard refresh (cache-busting)
## Problem

Rotating or mirroring an image (and, more generally, any reprocessing that rewrites an image's files in place) does not visibly update on the site until a **hard refresh**. After a rotate/mirror:

- the **show-image** page shows the old orientation,
- the **show-observation** page shows the old orientation,
- a normal reload of either page still shows the old image — only a hard refresh (cache bypass) shows the corrected one.

Manually reproduced on both pages.

## Root cause

Image renditions live at **stable URLs** keyed only on id + size — `.../orig/<id>.jpg`, `.../320/<id>.jpg`, etc. `Image::URL#source_url` (`app/models/image/url.rb`) produces that path with **no version/cache-busting token** (`Image#image_url`, `app/models/image.rb:356`, passes only `size`/`id`/`transferred`/`extension`). A rotate rewrites the *contents* of those files but not their URLs, so the browser — and any CDN/proxy in front of the image server — keep serving the previously-cached bytes under the unchanged URL.

## Proposed fix

Append a cache-busting version token to image `src` URLs that changes whenever the image content changes. `image.updated_at` is the natural token: the rotate pipeline bumps it (`Image::Processor#rotate` → `update_image_record_width_height_and_transferred`, then `transferred = true` at transfer completion), so once reprocessing finishes the URL flips and browsers/CDNs fetch fresh.

- Thread `updated_at` (or a dedicated version) into `Image::URL` via `Image#image_url(size)`.
- Append `?<updated_at.to_i>` in `Image::URL#source_url`.

Result: any page load **after** reprocessing completes — show-image, show-observation, matrix boxes, carousels, everywhere — gets a new URL and the correct image, with no hard refresh. The timing is naturally correct: a load *during* processing still shows the old image (it genuinely isn't done yet); the token flips when the job finishes.

## Caveat to verify

The cache-buster only works if the image server (nginx on `images.mushroomobserver.org`) and any CDN **vary their cache by query string** (the common default). If query strings are stripped, bust via a path segment instead. Worth confirming that vhost config as part of this work.

## Relationship to #4752

This is a **prerequisite** for #4752 (broadcast image-processing completion). #4752's Turbo broadcast re-renders the image component when `transferred`/`gps_stripped` change — but the re-rendered `` points at the same stable URL, so with a warm browser/CDN cache the broadcast (or a Turbo morph, or a plain reload) swaps in visually-identical markup and the new image never appears. Cache-busting is what makes the broadcast — and normal reloads — actually show the updated image. See the review comment on #4752.

## Scope note

Fixing this addresses the "show-observation page correct on a normal load" half of the rotate/mirror caching problem. The "show-image page auto-reloads the moment processing finishes" half is the Turbo-broadcast work in #4752, which builds on this.

mo-nathan:
## Summary

Fixes #4808. First of two PRs replacing the closed #4752 (whose base branch died with #4749's closing) — this is the cache-busting prerequisite; the surviving Turbo-broadcast work follows in a PR stacked on this one.

Image renditions live at stable URLs keyed only on id + size (`.../320/<id>.jpg`), so a rotate/mirror — async in production since #4751's `RotateImageJob` — rewrites the file contents under an unchanged URL, and browsers/CDNs keep serving the previously-cached bytes until a hard refresh. Manually reproduced on both the show-image and show-observation pages (see #4808).

### Change

`Image::URL#source_url` now appends `?<version>` when a version token is present, and `Image#image_url` threads `updated_at.to_i` through as that token. The rotate pipeline bumps `updated_at` when processing finishes, so the URL flips exactly when fresh bytes exist — a page load *during* processing still shows the old image (it genuinely isn't done yet), and any load after completion fetches fresh. This fixes normal reloads everywhere an `Image` record is rendered (show pages, matrix boxes, carousels).

### What deliberately does NOT get a token

- **Class-level `Image.url`/`Image.all_urls`** (id-only call sites: textile `!image>` embeds, map popups, comment-user thumbnails, data exports) — no record in hand to read `updated_at` from, so no token unless the caller passes `version:` explicitly. The Fundis/report exports in particular depend on stable URLs.
- **Placeholder URLs** (`/place_holder_*.jpg`) — static assets.
- **`Image::URL#full_filepath`** — also builds local *filesystem* paths (`Image#full_filepath`, `source_exists?`), where a query string would be wrong. The token is applied only in `source_url`.

### Caveat to verify at deploy time (from #4808)

The buster only works if nginx on `images.mushroomobserver.org` (and any CDN in front of it) varies its cache by query string — the common default. Worth confirming that vhost config before relying on this in production; if query strings are stripped there, the token needs to move into a path segment instead.

## Test plan

- [x] `bin/rails test test/models/image_test.rb` — new coverage: every size's instance URL carries the `?<updated_at.to_i>` token; token changes when `updated_at` changes; placeholders never get a token; class-level `Image.url` is token-free without `version:` and tokened with it
- [x] `bin/rails test test/models/image_test.rb test/components/ test/classes/report_test.rb` — 1005 tests, 0 failures (report URL stability confirmed)
- [x] `bin/rails test test/controllers/observations/downloads_controller_test.rb` — the one controller test asserting exact rendition URLs
- [x] `bundle exec rubocop` clean on all changed files

### DIFF
--- a/app/components/matrix/table.rb
+++ b/app/components/matrix/table.rb
@@ -37,7 +37,10 @@ class Components::Matrix::Table < Components::Base
   # read this through `cache_key_for`. Phlex's automatic class +
   # method + line digest doesn't survive into the controller's
   # check, so we encode the version explicitly.
-  CACHE_VERSION = "v1"
+  # v2: image URLs in the fragment now carry a cache-busting
+  # ?<updated_at> token (#4808) -- fragments cached under v1 embed
+  # tokenless URLs and must be regenerated.
+  CACHE_VERSION = "v2"
 
   # The cache key MatrixBox fragments are stored under, used by
   # both the Phlex `low_level_cache` write inside this component and
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -358,7 +358,8 @@ class Image < AbstractModel # rubocop:disable Metrics/ClassLength
       size: size,
       id: id,
       transferred: transferred,
-      extension: extension(size)
+      extension: extension(size),
+      version: updated_at&.to_i
     )
   end
 
@@ -367,7 +368,8 @@ class Image < AbstractModel # rubocop:disable Metrics/ClassLength
       size: size,
       id: id,
       transferred: args.fetch(:transferred, true),
-      extension: args.fetch(:extension, "jpg")
+      extension: args.fetch(:extension, "jpg"),
+      version: args[:version]
     )
   end
 
--- a/app/models/image/processor.rb
+++ b/app/models/image/processor.rb
@@ -255,9 +255,16 @@ class Image
 
     # This also sets transferred to false to save db writes.
     # Needed in rotate, and later overwritten in process.
+    # updated_at is forced explicitly: a mirror (or 180/vertical flip)
+    # leaves width/height unchanged, and transferred can already be
+    # false -- with no dirty attribute Rails skips the UPDATE entirely,
+    # so updated_at (and the #4808 cache-busting URL token derived from
+    # it) would never change and browsers would keep the pre-transform
+    # bytes. Rotations dodge this only by luck (width/height swap).
     def update_image_record_width_height_and_transferred
       width, height = FastImage.size(full_size_filepath)
-      @image.update(width: width, height: height, transferred: false)
+      @image.update(width: width, height: height, transferred: false,
+                    updated_at: Time.zone.now)
     end
 
     def make_file_sizes
--- a/app/models/image/url.rb
+++ b/app/models/image/url.rb
@@ -25,7 +25,7 @@ class Image
       small: "/place_holder_320.jpg"
     }.freeze
 
-    attr_accessor :size, :id, :transferred, :extension
+    attr_accessor :size, :id, :transferred, :extension, :version
 
     def initialize(args)
       size = args[:size]
@@ -35,6 +35,7 @@ class Image
       self.id          = args[:id]
       self.transferred = args[:transferred]
       self.extension   = args[:extension]
+      self.version     = args[:version]
     end
 
     def url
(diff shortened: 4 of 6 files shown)


mo-nathan:
Tracked down the mirror report — fixed in f23e4c6376. It wasn't an overlooked code path in this PR (the URL token applies to every rendition URL regardless of operation); it was a no-dirty-attributes accident in `Image::Processor` that mirror happens to trigger and rotate happens to dodge.

`Image::Processor#update_image_record_width_height_and_transferred` did `@image.update(width:, height:, transferred: false)`. A ±90 rotate swaps width/height, so the record goes dirty, the UPDATE runs, and `updated_at` bumps — token flips. A mirror leaves width/height unchanged, and when `transferred` is already `false` — always the case in dev, where there are no writable image servers so `Verifier#mark_transferred` never runs, and also true for any not-yet-transferred image in production — **nothing is dirty, Rails skips the UPDATE entirely, and `updated_at` never moves.** Same URL, so the browser correctly serves its cached pre-mirror bytes.

The evidence from this morning's test session (dev DB): the two mirrored images (2036092, 2036094, both `-h`) had file mtimes of 05:16/05:18 today but `updated_at` stuck at yesterday 07:28; the two rotated ones (2036091, 2036093) bumped `updated_at` to 05:16/05:17. All four `RotateImageJob`s ran and finished cleanly — only the record write was skipped.

The fix forces the timestamp in that update (`updated_at: Time.zone.now`), so any reprocessing that rewrites the files always flips the token — mirror, vertical flip, and 180° rotation included. Regression test added at the exact scenario (dims already correct, `transferred` already false); verified it fails without the fix.

One note for #4825: in production a mirror still flips `transferred` (true → false → true on transfer completion), so the completion broadcast fires. In dev there's no transfer step, hence no "processing finished" flip and no broadcast — the page catches up on the next load via the token instead. That's inherent to dev having no image servers, not something #4825 needs to handle.

PR #4829, opened 2026-07-17
Quiet ActionCable and Solid Cable log chatter from #4825's per-image `turbo_stream_from` subscriptions

mo-nathan:
## Summary

Since #4825 subscribes every rendered image to a `[image, :processed]` stream, a single page load floods the dev log (and would flood `production.log` on deploy — it logs at `:info`) with dozens of lines from two sources.

### 1. ActionCable info chatter

`Registered connection (...)` (ActionCable's `InternalChannel`) and `Turbo::StreamsChannel is streaming from... / transmitting the subscription confirmation` — one pair per subscription, all at `:info` through `config.action_cable.logger`. Fixed by giving ActionCable its own WARN-level logger in development (stdout) and production (`log/production.log`). Warnings and errors still log.

### 2. One `SELECT MAX(id) FROM solid_cable_messages` per subscription

Solid Cable already silences its recurring 0.1s polling queries — `Listener#listen` wraps them in `#with_polling_volume`, which uses `ActiveRecord::Base.logger.silence` — but the one-off `MAX(id)` lookup in `Listener#add_channel` (runs on every new stream subscription) is not wrapped. New initializer `config/initializers/solid_cable_silence_subscribe.rb` prepends a module wrapping `add_channel` in that same silencer. Gem-version-specific (solid_cable 3.0.12), commented for removal when fixed upstream — this is arguably a gem gap worth an upstream issue/PR.

## Test plan

- [x] `bin/rails test test/channels/solid_cable_subscribe_silencer_test.rb` — behavioral probe (add_channel runs inside the silencer) + prepend-is-active assertion against the real Listener
- [x] Verified in the development environment: module is first ancestor of the Listener, cable logger level is WARN
- [x] `bundle exec rubocop` clean on all changed files

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -116,6 +116,14 @@ MushroomObserver::Application.configure do
 
   # Highlight code that enqueued background job in logs.
   config.active_job.verbose_enqueue_logs = true
+
+  # Quiet ActionCable's per-connection/per-subscription info chatter
+  # ("Registered connection", "Turbo::StreamsChannel is streaming
+  # from..."). Since #4825 subscribes every rendered image to a
+  # [image, :processed] stream, a single page load emits dozens of
+  # those lines. Warnings and errors still log.
+  config.action_cable.logger =
+    ActiveSupport::Logger.new($stdout, level: Logger::WARN)
   # Solid Queue as the queue adapter locally, as on production.
   config.active_job.queue_adapter = :solid_queue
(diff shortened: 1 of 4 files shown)


mo-nathan:
Closing unmerged — superseded by the subscription scope-down (see that PR's description): with index-page subscriptions removed, the noise this PR silenced largely disappears, and silencing queries/cable chatter was the wrong trade regardless — the visible dev-log queries are exactly what surfaced the underlying design problem.

PR #4844, opened 2026-07-19
`i18n`: consolidate ALL-CAPS/lowercase twin translation tags; add `Symbol#ti`

### LINKED ISSUE #4843: More efficient localizations
## Current Database situation
There are several things that could be improved regarding the `translation_strings` table:
- it is not indexed. Currently, this slows down the generation of `#{locale}.yml` and `.txt` files during `lang:update`; lookups are more expensive than they have to be.
- it has _case-**in**sensitive_ collation, but _case-sensitive_ lookup keys, e.g. UPPERCASE and lowercase keys for the most commonly repeated titles and labels, like "Collection Number"/"collection number". The collation means that a query for 
```ruby
TranslationString.where(key: "COLLECTION_NUMBER").distinct
``` 
returns the lower case value.
- has inconsistencies between titleizing patterns, even within the same language. This arises from translators doing their own titleizing, sometimes at odds with each other.

## Current `i18n` mechanism
MO uses the default rails i18n backend, which reads translations from .yml files. 

When users supply translations, they are stored in the `translation_strings` db, added to the `yml` file where data is read at runtime, and added to the annotated `txt` file used in the translation UI. The `.yml` files are what is used for lookups by Rails, and the `txt` files are used only in the translators' interface. 

The English `.txt` file, however, is the master copy of what keys are valid; any deletions from this file will result in deletions of all corresponding translations in the database, via the Rake lang scripts, and any additions make new strings available for translation in production.

`lang:update` takes awhile to regenerate all those files: it's half of our downtime for each deploy.

## Solid Cache as alternative

It turns out Rails can also be configured to read translations out of its cache, rather than from `.yml` files. The way this works is the keys and translations are read out of the db the first time, and cached by key and locale. Changes to translations or keys bust the cache. The first loads can be a bit more expensive, because they're db queries, but after populating, they're quicker than file reads. The cache also survives reboots and deploys, unless explicitly dumped.

- It loads slightly faster for users than `.yml`
- It eliminates the file generation steps of `lang:update`, saving time at deploys
- No more `.yml` files, and only one `en.txt` file

### Changes this would require

The cache is populated by db queries of the `translation_strings` table. Because these currently don't work (see above), either:

- The table collation must be changed to become case-sensitive, or
- The translation keys must be made case **in**sensitive, which is to say, `snake_case`.  
- That could mean separate lowercase/uppercase translations survive at keys like `:collection_number`/`:collection_number_ti` (the former `:COLLECTION_NUMBER`), or it could mean just keeping one string, and automatically titleizing the lowercase string to produce the other. 
- Having one translation per "twin", we'd need programmatic titleization of the strings. It turns out the titleizing is different per locale, but still pretty predictable. 

## Standardizing titleization per locale

Rails' `.titleize` only works in English, and not well enough even for that. I examined diffs in each language to determine algorithms for titleization in each lang.
 
After applying these algorithms within a new Symbol method `.ti`, i re-audited the diffs between what we would produce automatically, and user-supplied translations. Extensive audit results below.

## Production db changes

I just made the 44 translation string updates/corrections suggested below on the production db, in all languages, which should eliminate any deviation between what is currently/correctly shown in a locale on production, and what the i18n simplification, including the migrations, will produce on production.

nimmolo:
## Important

> [!WARNING]
> When you get on this branch, you must run `db:migrate` **_before_** `lang:update`, as `deploy.sh` does in production. If you run it in the opposite order, you'll lose the theme name and standalone-tag translations in your db. That's the only risk - minor but worth noting.

> [!NOTE]
> This PR does not change the mechanism of i18n from files to cache. It only removes the `:UPPERCASE`/`:lowercase` "twin strings" duplication in `en.txt` and introduces a method to derive the titleized localized version from the lowercase string. Title strings are very thoroughly tested in the test suite, but in any case, the worst bug this PR can produce is an inappropriately lowercased string, which can be fixed in a follow-up by calling `.ti` on the string.

#### Which of the 399 dropped tags needed a migration

`translation_strings.tag` has a case-insensitive collation, so a naive `where(tag: ...)`/`find_by(tag: ...)` query can silently match either casing of a tag and return the wrong row. Every result below was re-verified with an exact, case-sensitive Ruby string comparison on the fetched rows, not trusted to the DB's `=`/`IN` match.

- **69 of the 399 uppercase tags had no lowercase counterpart in `en.txt`** (`OK`, `DQA`, `CURATOR`, `THEME`, etc.). These had translated content in some languages that a plain `en.txt` key rename would not move in the database — only `TranslationString.rename_tags` does that. Without a migration, these 69 would lose all their translations. The migration renames these 69 plus the 5 theme-name tags (`Agaricus`/`Amanita`/`BlackOnWhite`/`Cantharellaceae`/`Hygrocybe`), 74 total.
- **The remaining 330 dropped tags already had a lowercase counterpart in `en.txt`.** Audited every (tag, language) pair across all languages: zero pairs have a translation under the ALL-CAPS tag but none under the lowercase counterpart. Deleting these 330 from `en.txt` with no migration loses nothing.
- The same migration also adds a `[language_id, tag]` index `translation_strings` has never had, not even on `language_id` (the `has_many :translation_strings` foreign key). `Language#strip` and `Language#translation_strings_hash` scan by `language_id` on every `lang:update`, and the migration's own 74 renames each scan by `tag`; adding the index before the renames run means both benefit.

## Summary

Fixes #4843.
- Simplifies MO's translation strings API by eliminating capitalized duplicates of common simple one/two-word strings, like `:OBSERVATION`/`:observation`.
- Adds a new String method, `.ti`, which titleizes the string in a way appropriate to the locale, based on an audit of existing ALL-CAPS translations against algorithmic derivations from their lowercase counterparts: `.l.titleize` in English/Spanish/Portuguese/Polish (Latin-script, where `.titleize` reproduces the translators' per-word capitalization correctly); a Unicode-aware per-word capitalize in Greek/Russian/Ukrainian/Belarusian (Cyrillic/Greek, where `.titleize`'s ASCII-only regex is a silent no-op) and Turkish (with `:turkic` i/I mapping applied per word); plain `.l.capitalize` (sentence-case) in French and Italian, whose apostrophe-elided articles ("d'activité") both title-casing approaches mishandle; and a no-op in German and locales with no letter-casing concept (Arabic, Farsi, Chinese, Japanese).
```ruby
:species_lists.l == "observation lists"
:species_lists.ti == "Observation Lists"
```

### Why the change
Storing two independently-translated strings per concept adds translator busywork, and conflates two different concerns: **translation** (what does this word mean in French) and **presentation** (how is it cased in this context). Casing is a rendering decision, not a linguistic one, in the overwhelming majority of cases here — nearly all 47 sites are single words, where "capitalize each word" is unambiguous and language-agnostic. Collapsing to one canonical string per concept plus a `.ti` method:

  - Eliminates translation string key lookup mismatches like `:Vote.t` — there's no second tag left to typo against.
  - Halves translator burden for every one of these ~332 concepts, going forward, across all 15 languages.
  - Fits the existing Symbol API shape — `.l`/`.t`/`.tl`/`.tp`/`.tpl` are already "same lookup, different post-processing." `.ti` is one more.

### Why not keep the "Title Case" version and downcase it, instead

In German, nouns are always capitalized even in the middle of sentences, so downcasing programmatically for contexts where this would be appropriate in other languages would make the German incorrect. (In German translations, the "lowercase" version of a translated string is expected to be Title Case already wherever it's a noun, which is appropriate.)

For the rest, which words get capitalized turned out to differ by locale, so we audited each one's existing ALL-CAPS translations against algorithmic derivations from the lowercase tag rather than assume. English/Spanish/Portuguese/Polish, Greek/Russian/Ukrainian/Belarusian, and Turkish translators predominantly wrote per-word title case; French and Italian render as sentence-case instead (e.g. "Listes d'espèces") because both routinely use apostrophe-elided articles that a per-word title-case pass would mis-capitalize ("D'espèces"). In locales with no letter-casing concept, there is no change.

### Why the twin-tag pattern was a correctness risk, not just translator busywork

`en.txt` carried a parallel ALL-CAPS/lowercase "twin" tag for every title-cased label (`VOTES`/`votes`, `OBSERVATIONS`/`observations`, etc.), so callers had two spellings to choose from and typos went undetected. A mismatched-case symbol like `:Votes.t` silently fell through `Symbol#localize`'s exact-match → downcase-fallback path instead of raising, costing a fresh, uncached `translation_strings` query on every single call — 574 on one home-page load.

This PR eliminates the twin-tag pattern entirely. Every translation tag is now lowercase-only; title-cased display goes through a new, locale-aware `Symbol#ti` instead of a separate ALL-CAPS twin tag. A new RuboCop cop (`MO/LowercaseTranslationTag`) and an `en.txt` content-shape test make the whole bug class impossible to reintroduce.

## What changed

- **`Symbol#ti`** (`app/extensions/symbol.rb`) — `Symbol.titleize_localized(localize(*))`, locale-aware per the Summary above: `.titleize` for `TI_TITLEIZE_LOCALES` (en/es/pt/pl), a new Unicode-aware `Symbol.capitalize_each_word` for `TI_WORD_CAPITALIZE_LOCALES` (el/ru/uk/be) and `TI_TURKIC_LOCALES` (tr, with `:turkic` mapping per word), a no-op for `TI_NO_OP_LOCALES` (de/ar/fa/zh/jp), and plain sentence-case `.capitalize` otherwise (fr/it). Deliberately not textilized (checked every ALL-CAPS tag's content for Textile markup, found none — some callers interpolate the result into HTML attributes, where inserted markup would be wrong). The `[:tag.ti]` embedded-ref syntax routes through the same helper. Covered by `SymbolExtensionsTest#test_titleize_localized_all_locales`, which exercises every branch across all 16 of MO's production locales using real translated content.
- **`Symbol#localize_expand_arguments`** and **`Symbol#localize_recursive_expansion`** each independently depended on ALL-CAPS tags too (`[TYPE]`/`[TYPES]` argument placeholders, `[:tag]` embedded refs) — fixed both, and extended the embedded-ref syntax with an optional `.ti` suffix (`[:tag.ti]`).
- **Deleted all 399 ALL-CAPS tags from `en.txt`** (330 ALL-CAPS/lowercase twin pairs + 69 standalone ALL-CAPS tags with no lowercase twin — see the audit section above) and rewrote every call site — ~1100+ sites total, including roughly 150 built **dynamically at runtime** (`table_name.upcase.to_sym.l`, ALL-CAPS hash-shorthand keys, opaque symbols routed through a shared helper like `FieldLabelRow#resolved_label_text`) that a static regex sweep structurally could not find. Two full audit passes closed this gap; see #4843 for the methodology.
- **Migrated the 69 standalone tags plus the 5 theme-name tags** (`Agaricus`/`Amanita`/`BlackOnWhite`/`Cantharellaceae`/`Hygrocybe`) via `TranslationString.rename_tags`, in `db/migrate/20260718120000_lowercase_standalone_translation_tags.rb` — see the audit section above. `BlackOnWhite` in particular has a different translation per language (Spanish "Blanco y Negro", French "Noir sur blanc", etc.), not just a repeated Latin genus name.
- **Added `ID`/`OK` acronyms** to `config/initializers/inflections.rb` so `.titleize` reproduces them correctly, and renamed the two Zeitwerk-autoloaded classes this collided with (`Components::IdBadge` → `IDBadge`, `API2::ObjectNotFoundById` → `ObjectNotFoundByID`, 15 files total) — verified with a full `Rails.application.eager_load!`.
- **New RuboCop cop**, `lib/rubocop/cop/mo/lowercase_translation_tag.rb` (MO's first custom cop) — flags any Symbol literal used with `.l`/`.t`/`.tl`/`.tp`/`.tpl`/`.ti` whose name contains an uppercase letter.
- **New test**, `LocalizationFilesTest#test_no_uppercase_tag_keys` — asserts no tag key in the DB-backed English translation set contains an uppercase letter.

## Test plan

- [x] Full suite green locally (`bin/rails test`), confirmed by the user after two rounds of fixes for regressions this sweep itself introduced (missing-tag misses from dynamically-built symbols the static sweep couldn't see, caught by the existing `Symbol.missing_tags` teardown check)
- [x] `bundle exec rubocop` clean across the full repo, including the new cop running against its own sweep output
- [x] Collation-safe audit (case-sensitive comparison on fetched rows, not trusted to the DB match) confirmed zero data loss across all 330 twin-pair tags in every non-English language
- [x] The 69 standalone tags plus 5 theme-name tags are covered by a real `TranslationString.rename_tags` migration, run and verified locally against a production checkpoint
- [x] `Rails.application.eager_load!` clean after the `ID` acronym class renames


app/classes/api2/error/object_not_found_by_id.rb
```diff
@@ -2,7 +2,7 @@
 
 class API2
   # Referenced object id doesn't exist.
-  class ObjectNotFoundById < FatalError
+  class ObjectNotFoundByID < FatalError
```
JoeCohen, line 5: Not a blocker. I'd prefer `ObjectNotFoundById` and similar (fewer special inflections) here and elsewhere. For me it's easier to always follow the Rails rules than remembering all the exceptions. This has tripped me in the past (I think for API vs Api).
(And fewer/no special inflections seems more in the spirit of this PR.)

app/classes/tab/sidebar/indexes/herbaria.rb
```diff
@@ -3,7 +3,7 @@
 # Sidebar indexes nav: herbaria index.
 class Tab::Sidebar::Indexes::Herbaria < Tab::Base
   def title
-    :HERBARIA.t
+    :herbaria.ti
```
JoeCohen, line 6: Out of scope: Might be a good time to change Herbaria/Herbarium to Fungaria/Fungarium so that code and UI are the same.

app/components/button/submit.rb
```diff
@@ -8,14 +8,14 @@
 # `disable_with:` defaults to `name:` when set.
 #
 # @example Plain text
-#   Button(type: :submit, name: :SAVE.l)
+#   Button(type: :submit, name: :save.ti)
```
JoeCohen, line 11: Out of scope: Maybe an opportunity to DRY all the buttons which submit to `submit.ti`, getting rid of `save`, etc.

app/components/form/search.rb
```diff
@@ -449,7 +449,7 @@ def array_to_newlines(val)
 
   def render_form_buttons
     div(class: "text-center") do
-      submit(:SEARCH.l, class: "d-inline-block mx-3")
+      submit(:search.ti, class: "d-inline-block mx-3")
```
JoeCohen, line 452: ?? maybe `submit.ti` instead of `search.ti`? See comment above about DRYing submit buttons.

app/components/link/location.rb
```diff
@@ -38,7 +38,7 @@ def render_where_link
     a(href: url_for(observations_path(where: @where)),
       class: "index_observations_at_where_link") do
       render_label(@where)
-      plain(" [#{:SEARCH.t}]") if @click
+      plain(" [#{:search.ti}]") if @click
```
JoeCohen, line 41: Could this be `submit`?

app/controllers/herbaria_controller.rb
```diff
@@ -227,11 +227,11 @@ def modal_identifier
   def modal_title
     case action_name
     when "new", "create"
-      :new_object.t(type: :HERBARIUM)
+      :new_object.t(type: :herbarium)
     when "edit", "update"
       render_to_string(Views::Layouts::Header::ObjectTitle.new(
                          object: @herbarium, mode: :edit,
-                         title: :HERBARIUM_RECORD.l
+                         title: :herbarium_record.ti
```
JoeCohen, line 234: See out of scope comment above re herbarium => fungarium

app/views/controllers/account/api_keys/form.rb
```diff
@@ -66,7 +66,7 @@ def render_inline_edit_layout
                            class: "form-control border-none")
 
         render(Components::InputGroup::Addon.new) do
-          submit(:SAVE.l, submits_with: submits_text)
+          submit(:save.ti, submits_with: submits_text)
```
JoeCohen, line 69: See comment about save => submit

app/views/controllers/account/api_keys/form.rb
```diff
@@ -138,7 +138,7 @@ def submit_text
     end
 
     def submits_text
-      :SAVING.l
+      :saving.ti
```
JoeCohen, line 141: maybe `submitting`

app/views/controllers/account/preferences/form/email_section.rb
```diff
@@ -33,7 +33,7 @@ def render_email_section
       render_email_group(label_key, fields)
     end
     Help(content: :prefs_email_note.tp, class: "mt-4")
-    submit(:SAVE_EDITS.l, center: true)
+    submit(:save_edits.ti, center: true)
```
JoeCohen, line 36: submit?

app/views/controllers/account/preferences/form.rb
```diff
@@ -51,7 +51,7 @@ def render_login_section
     password_field(:password, label: :prefs_password_new)
     password_field(:password_confirmation,
                    label: :prefs_password_confirm)
-    submit(:SAVE_EDITS.l, center: true)
+    submit(:save_edits.ti, center: true)
```
JoeCohen, line 54: submit?

app/views/controllers/account/profile/form.rb
```diff
@@ -29,20 +29,20 @@ def form_action
 
     def view_template
       super do
-        submit(:UPDATE.l, center: true)
+        submit(:update.ti, center: true)
```
JoeCohen, line 32: submit?

app/views/controllers/theme/color_themes.rb
```diff
@@ -9,7 +9,7 @@ def view_template
       trusted_html(:color_themes_text.tp)
       p do
         MO.themes.each do |name|
-          a(href: url_for(action: name)) { plain(name.to_sym.l) }
+          a(href: url_for(action: name)) { plain(name.underscore.to_sym.l) }
```
JoeCohen, line 12: Are these right? It's what I see after updating preferences to Spanish:

Image
<br/>
Image

JoeCohen:
I added this comment for convenience. The PR should be straightforward, but because it's so large, I asked the intern to suggest some manual tests.

> Based on the nimmo-i18n-titlecase-typo-sweep branch (the ALL-CAPS/lowercase twin-tag elimination + Symbol#ti sweep, commit [724ca4de00](https://github.com/MushroomObserver/mushroom-observer/commit/724ca4de00)), here are manual UI tests. This sweep touched ~545 files, so these are prioritized at the highest-risk spots: the original regression site, dynamically-built tags (.upcase/hash-shorthand — invisible to a static regex sweep), the theme-tag migration, and a representative cross-section of ordinary title-cased labels.

## Manual Tests: i18n titlecase / twin-tag sweep

### 1. Original regression site — image vote widget
1. Go to any observation `Show` page with at least one image, e.g. `/observations/<id>`.
2. ~Open the image's vote dropdown/tooltip (the vote-count control).~
3. ~**Expected:** label reads `"<N> Votes"` (capital V), not `"<N> votes"`, `"<N> VOTES"`, or a raw `votes` string.~
The vote widget is a thin progress bar under an image, not literally a dropdown — that was imprecise wording on my part. Here's the actual path:

Simplest route — the image's own Show page:

Go to any image's page directly: /images/<id> (e.g. click any thumbnail from an observation, then click through to the full image page — or just visit /images/1 in dev).
Below the full-size image you'll see a thin horizontal progress bar (the vote meter) with vote-value links ((x), 1, 2, 3, 4) underneath it.
Hover your mouse over the progress bar itself (not the buttons) — a native browser tooltip pops up after ~1 second showing "<N> Votes". That tooltip text is the .ti call you're checking.
Optionally click one of the vote value links to cast a vote and confirm the meter/tooltip updates.

5. Open Rails dev log (or `bin/rails server` console) while loading the page.
   **Expected:** no spike of repeated identical translation-lookup queries (this was the original 574-query regression from PR #4811 — confirms the mismatched-case fallback path is gone).

### 2. "OK"/"Cancel" acronym in confirm modals
1. Trigger any destructive confirm modal — e.g. go to an object you own (a `Location`, `Herbarium Record`, etc.) and click its delete/destroy icon.
2. **Expected:** modal buttons read `OK` (both letters capitalized) and `Cancel` — not `Ok`.

### 3. Theme name dropdown
1. Go to `/account/preferences/edit` (Account → Preferences) while logged in.
2. Open the **Theme** select dropdown.
3. **Expected:** options read `Agaricus`, `Amanita`, `Black on White`, `Cantharellaceae`, `Hygrocybe`, `Random` — properly capitalized, no literal tag names (`agaricus`, `BLACK_ON_WHITE`, etc.) and no blank options.

### 4. CRUD flash-message casing
1. Edit any object you have permission to edit (e.g. a `Location` or `Species List`) and save a trivial change.
2. **Expected:** flash notice reads e.g. `"Location was successfully updated."` — the object-type word is capitalized correctly, not `"location"` or `"LOCATION"`.
3. Repeat for a `create` action (e.g. add a new Herbarium Record) and a `destroy` action.
   **Expected:** same correct capitalization in each flash message.

### 5. Spot-check tabs / page titles across controllers
Visit each of the following and check the page title bar, context-nav dropdown (top-right tabs menu), and any in-page tab strip:
- `/observations` (index) — page title + sidebar tabs
- `/names` (index)
- `/projects/<id>` — project banner tabs (Summary / Names / Locations / Species Lists / Admin)
- `/herbaria`
- `/species_lists`
- `/glossary_terms`

**Expected for each:** every label is proper Title Case (e.g. `Species Lists`, not `species Lists` or `SPECIES LISTS`), no page shows a raw translation key (`translation missing: en...`) or a literal `:tag.ti` artifact anywhere in the rendered page.

### 6. Observation form labels
1. Go to `/observations/new` (or edit an existing observation you own).
2. Check field labels, section headings (Notes, Images, Projects, Species Lists), and the submit button.
3. **Expected:** all labels/headings are correctly capitalized; nothing renders as a raw symbol or lowercase where a heading is expected.

### 7. (Optional, if you have iNat import test data) iNat import audit snapshot
1. Run an iNat import against a test iNat account and open the resulting `/inat_imports/<id>` status/audit page.
2. **Expected:** the audit row's raw-data snapshot text (e.g. `User: <name>`) is **sentence case** (`Symbol#l.upcase_first`), not title-cased — this path was deliberately kept different from the `.ti` rewrite.

JoeCohen:
Maybe out of scope: The intern thinks there should be a vote meter/progress bar on the show_image page. I'm not seeing one. Is the intern seeing some dead code?

I do see this on show_observation, which is a prgres bar but differes from what the intern describes.
Screenshot 2026-07-19 at 11 35 22 AM


>Confirmed — the vote widget has class: "vote-section require-user", and there's CSS: when the <body> (or a wrapping ancestor) has class .no-user, everything with .require-user is hidden via display: none. So the fix is simple:

> You must be logged in to see the vote meter/progress bar at all — it's invisible to anonymous visitors by design (voting requires an account). Log in first, then revisit the image page:

> Log in at /account/login.
Go to an image with existing votes, e.g. /images/<id> for an image you know has votes (or vote on one yourself first — voting is available even with 0 prior votes).
The progress bar + vote links should now appear below the image.
If you were already logged in and still don't see it, let me know and I'll dig into whether @image.vote_cache/votes: is being passed correctly for that page instead.

JoeCohen:
Intern says:

>2. "OK"/"Cancel" acronym in confirm modals
Trigger any destructive confirm modal — e.g. go to an object you own (a Location, Herbarium Record, etc.) and click its delete/destroy icon.
Expected: modal buttons read OK (both letters capitalized) and Cancel — not Ok.

- The buttons I see (for Comment and Observation) are `Cancel` and `Destroy <object>`. IMO, it would be simpler to have `OK` everywhere. (Not a blocker.)
- Also a redundant "Are you sure?", which is a preexisting issue. (Not a blocker.)
Screenshot 2026-07-19 at 12 05 31 PM

app/classes/api2/error/object_not_found_by_id.rb
nimmolo: Necessary to add the acronym for the correct capitalization of translation strings.

app/classes/tab/sidebar/indexes/herbaria.rb
nimmolo: Agree, i think that might want to be a separate sweep.

app/components/button/submit.rb
nimmolo: I considered this last year, but i felt like there are enough cases where the submit button says "Add", "Remove", "Save" etc and it makes more sense UX wise.

nimmolo:
@JoeCohen These ideas are interesting but would be probably better in separate issues. 

This PR is moving a lot of code around trying to change the mechanics of translation strings without changing any results, so we can be sure it's not messing up anything. If i also change output it makes it too slippery and confusing.

app/views/controllers/theme/color_themes.rb
nimmolo: These are not right. Will investigate.

mo-nathan:
I agree that we should avoid changing functionality as part of this PR.

app/extensions/symbol.rb
```diff
@@ -307,6 +310,16 @@ def tpl(*)
...
   end
 
+  # Localize, then title-case the result. No textilizing: title-cased
+  # tags are short UI labels, never Textile-formatted body text, and
+  # some callers interpolate the result straight into an HTML attribute
+  # (e.g. a `title:`) where inserted markup would be wrong. Use this
+  # instead of authoring a separate ALL-CAPS twin tag for a title-cased
+  # presentation of an existing lowercase tag.
+  def ti(*)
+    localize(*).titleize
```
mo-nathan: This seems like the wrong direction to do this in.  It seems like there's a lot less computation and risk in going from "This is a Title" -> "this is a title" than going from "this is a title" -> "This is a Title".  If you are choosing one why chose the one with less infomration?

JoeCohen:
The intern says (IDK if it's correct):

> 4. CRUD flash-message casing
Edit any object you have permission to edit (e.g. a Location or Species List) and save a trivial change.
Expected: flash notice reads e.g. "Location was successfully updated." — the object-type word is capitalized correctly, not "location" or "LOCATION".

When I edited an Observation the object-type word in the3 flash was all lower case:
Screenshot 2026-07-19 at 4 26 39 PM

app/extensions/symbol.rb
nimmolo: Good question. 

TL;DR: Dealbreaker is in German, where something like "Observation Lists" has two words that must be capitalized, and downcasing them programmatically in some other context would be always wrong - nouns are always capitalized in Deutsch. 

The audit turned up no cases where there were twin strings like "This is a Title"/"this is a title".  The all-caps keys seem to have been restricted to single or double word, very simple strings like "SPECIES_LISTS", where there is no middle word that might be lowercase. In any language, we are capitalizing both anyway when used as a title. 

Longer title strings that don't have a lowercase counterpart seem to be always one-offs with lowercase keys. There were no "Alice in Wonderland"/"alice in wonderland" strings that the intern found when i asked for the audit.

I don't know of a language where Title Case is incorrect (and should be something like Sentence case) for what is effectively a two word proper name string. Stylistically it could of course be different, but `.ti` produces a correct string for all of these in any language i'm aware of.

JoeCohen:
The intern expects that object name in a create flash should be capitalized. It is actually lower case.

>CRUD flash-message casing
Edit any object you have permission to edit (e.g. a Location or Species List) and save a trivial change.
Expected: flash notice reads e.g. "Location was successfully updated." — the object-type word is capitalized correctly, not "location" or "LOCATION".
**Repeat for a create action (e.g. add a new Herbarium Record) and a destroy action.
Expected: same correct capitalization in each flash message**

Screenshot 2026-07-19 at 4 35 12 PM
<br/>
Screenshot 2026-07-19 at 4 35 50 PM

db/schema.rb
```diff
@@ -182,7 +182,7 @@
     t.string "accession_number", limit: 80, null: false
   end
 
-  create_table "image_gap_checkpoints", charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
+  create_table "image_gap_checkpoints", charset: "utf8mb3", force: :cascade do |t|
```
mo-nathan: Wrong collation

db/schema.rb
```diff
@@ -936,10 +936,11 @@
   create_table "translation_strings", id: :integer, charset: "utf8mb3", force: :cascade do |t|
     t.integer "version"
     t.integer "language_id", null: false
-    t.string "tag", limit: 100
+    t.string "tag", limit: 100, collation: "utf8mb3_bin"
```
mo-nathan: I need to be convinced that this is the right collation.  If it is, this is going to crate a painful future.

nimmolo:
@JoeCohen The flash message is the same on main, and "no change from main" is what we're going for. Unless it's a diff from main, please don't flag it.

db/schema.rb
nimmolo: This collation is wrong! It should have been reverted.

db/schema.rb
nimmolo: Indeed

JoeCohen:
Re not flagging changes from main. Understood.
But not sure I'll have time to check both main and branch for everything. So I'll just flag stuff which seems crazy.

nimmolo:
Fwiw, the whole test suite tests translation strings thousands of times. 

The worst case scenario for this PR is that something gets downcased or inappropriately capitalized, and we have to fix it, e.g. by adding/removing `.ti`. No big deal, and not worth the time testing everything manually, IMO.

app/extensions/symbol.rb
nimmolo: Actually Title Case is incorrect in most languages. 

Maybe i'm the drunk intern today. PR modified.

PR #4846, opened 2026-07-19
Coerce scalar request params so hash-shaped probes don't 500

mo-nathan:
## Summary

Two errors showed up in `#alerts` from automated SQL-injection scanners (sqlmap-style time-based blind payloads, `… or sleep(30) …`):

- `ActiveRecord::StatementInvalid: can't quote ActionController::Parameters` in `set_locale` (from `?user_locale[…]=…`)
- `Literal::TypeError` on `IndexPaginationNav`'s `letter_param` prop (from `?letter[…]=…`)

**Neither probe injected anything** — the payload arrives as a hash *key*, so the param is an `ActionController::Parameters` object, and MO's existing defenses (ActiveRecord's quoting layer and the Literal `String` typecheck) rejected it before it reached SQL. No `sleep(30)` ran; nothing was exposed. But each probe raised a **500** instead of serving the page, which is what spammed `#alerts` — and it's a cheap way for a scanner to generate error noise.

This hardens the entry points so a garbage param *shape* is treated as "not given" rather than crashing a String-only sink.

## Change

Adds `ApplicationController#string_param(key)` — returns `params[key]` only when it's a `String`, else `nil` — registered as a Phlex value helper so views can use it too. Routes the GET-reachable raw scalar reads through it:

| Site | Sink that used to 500 |
| ---- | --------------------- |
| `params_locale` (`user_locale`) | `Language.find_by(locale: …)` — AR can't quote a `Parameters` |
| `index_nav.rb` (`letter`) | `IndexPaginationNav`'s `letter_param, _Nilable(::String)` |
| `images_controller` (`size`) | `params[:size].to_sym` — `NoMethodError` on a `Parameters` |
| `observations/index` view + `locations_controller` (`where`) | passed to a tab / form-value string context |
| `inat_imports_controller` (OAuth `state`) | `InatImport.find_by(id: …)` |

Left alone (already safe): params coerced via `.to_s` (`pattern`, most `where` reads), one `where` read already wrapped in `rescue`, and `find_by(id: params[:id])` sites whose `:id` is a route segment (always a scalar string). `Api2` query params are a separate entry point with their own error rendering and are out of scope here.

Deliberately **not** a global "reject hash-shaped scalars" filter — MO uses nested params legitimately everywhere (`q[…]`, `observation[…]`), so the coercion has to be per-read-site.

## Test plan

All MO devs have a prod-data dev DB. On a running instance, these are the probe shapes — each should now serve the page normally (before this change, each returned a 500):

- `/account/login/new?user_locale[x]=1` — renders the login form (locale falls back to default)
- `/articles?letter[x]=1` — renders the articles index (no letter filter)
- `/images/<id>?size[x]=1` — renders the image (default size)

A valid scalar still works as before — e.g. `/articles?user_locale=fr` switches to French, `/images/<id>?size=large` uses the large size.

_Automated coverage: `test/integration/rails-dom-testing/scalar_param_injection_test.rb` hits the exact probe shapes and asserts a normal response._


JoeCohen (approved):
Same comments (non-blocking) as for [Scrub invalid UTF-8 from incoming requests via Rack::UTF8Sanitizer](https://github.com/MushroomObserver/mushroom-observer/pull/4847):

- ~verbose intern comments~ (Just noticed you dealt with this in PR #4850)
- Still think it could be worth flagging these issues now and dealing with them in Rails 8 (on top of the current fixes). I'm guessing that would avoid rendering some pages, and would damp some robot activity.

PR #4847, opened 2026-07-19
Scrub invalid UTF-8 from incoming requests via `Rack::UTF8Sanitizer`

mo-nathan:
## Summary

Automated scanners POST malformed bytes (an invalid UTF-8 byte like `0xFF` in a form value). When those bytes reach a String operation deep in the app they raise `ArgumentError: invalid byte sequence in UTF-8` — a 500, and an `#alerts` notification. Seen recently on `POST /account/login` (`login#create`) and `POST /support/confirm` (`support#confirm`).

The scanner never injects anything — this is a robustness/noise problem, the encoding analog of #4846 (which coerced hash-shaped scalar params). Rails already catches *some* invalid-byte params itself and returns a 400 (`ActionController::BadRequest: Invalid encoding for parameter`), but others slip past that check into a raw String op and 500.

## Change

Adds the `rack-utf8_sanitizer` gem and inserts `Rack::UTF8Sanitizer` at the front of the middleware stack (`config.middleware.insert_before(0, …)`). It scrubs invalid UTF-8 out of every request's params and headers — the default `:replace` strategy swaps bad bytes for the Unicode replacement character — before any Rails or app code touches them. The garbage request is then processed normally (a junk login just fails auth) instead of erroring.

One boundary fix covering every endpoint, rather than patching `login#create` / `support#confirm` individually — the next scanner would otherwise just find a different action. The gem's default `SANITIZABLE_CONTENT_TYPES` covers `application/x-www-form-urlencoded` (the form POSTs in the alerts) plus JSON and query strings/headers.

## Test plan

All MO devs have a prod-data dev DB. Against a running instance, reproduce the probes with `curl` (the `%FF` is an invalid UTF-8 byte in a form value):

```
curl -i --data 'user[login]=%FFabc&user[password]=x' https://<host>/account/login
curl -i --data 'donation[amount]=5&donation[who]=%FFbad&donation[email]=a%FEb' https://<host>/support/confirm
```

- **Before** this change: a 500 (or a bare `ActionController::BadRequest`) and an alert.
- **After**: the request is served normally — the login page re-renders / the donation flow proceeds with the bad bytes replaced by `�`, no 500.

A valid request is unaffected — normal logins and donations behave exactly as before, since well-formed UTF-8 passes through untouched.

_Automated coverage: `test/integration/rails-dom-testing/utf8_sanitizer_integration_test.rb` POSTs invalid-byte form bodies to both endpoints and asserts a non-5xx response. (Verified it fails without the middleware — the requests raise on the bad encoding — and passes with it.)_

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -52,11 +52,9 @@ module MushroomObserver
     # Configure sensitive parameters which will be filtered from the log file.
     config.filter_parameters += [:password]
 
-    # Scrub invalid UTF-8 out of every incoming request before anything
-    # (including Rails' own param-filtering for the logs) touches it.
-    # Automated scanners POST malformed bytes; without this they reach a
-    # String op deep in the app and raise `invalid byte sequence in
-    # UTF-8` (a 500). Inserted at the front of the stack so it runs first.
+    # Scrub invalid UTF-8 out of every request at the front of the stack,
+    # before anything (including Rails' log param-filtering) touches it --
+    # else a scanner's malformed bytes reach a String op and 500.
     config.middleware.insert_before(0, Rack::UTF8Sanitizer)
 
     # Tells rails not to generate controller-specific css and js stubs.
--- a/config/application.rb
+++ b/config/application.rb
@@ -52,6 +52,13 @@ module MushroomObserver
     # Configure sensitive parameters which will be filtered from the log file.
     config.filter_parameters += [:password]
 
+    # Scrub invalid UTF-8 out of every incoming request before anything
+    # (including Rails' own param-filtering for the logs) touches it.
+    # Automated scanners POST malformed bytes; without this they reach a
+    # String op deep in the app and raise `invalid byte sequence in
+    # UTF-8` (a 500). Inserted at the front of the stack so it runs first.
+    config.middleware.insert_before(0, Rack::UTF8Sanitizer)
+
     # Tells rails not to generate controller-specific css and js stubs.
     config.generators.assets = false
 
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,11 @@ end
 # See https://github.com/trilogy-libraries/trilogy/tree/main/contrib/ruby
 gem("trilogy")
 
+# Scrubs invalid UTF-8 bytes out of incoming request params/headers at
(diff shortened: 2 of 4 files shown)


config/application.rb
```diff
@@ -52,6 +52,13 @@ class Application < Rails::Application
     # Configure sensitive parameters which will be filtered from the log file.
     config.filter_parameters += [:password]
 
+    # Scrub invalid UTF-8 out of every incoming request before anything
```
JoeCohen: The intern loves redundant comments, especially long redundant comments. 
Cf. test/integration/rails-dom-testing/utf8_sanitizer_integration_test.rb, Gemfile.

A year ago, I would have removed some of them, shortened others as much as possible.
**Question**: Nowadays, is that exactly wrong?
The redundancy could help an intern avoid reversions.
(On the other hand, if one of those files changes in the future, the intern has to clean up all of them. If it doesn't we have he opposite problem.)

JoeCohen (approved):
Thank you! 
See comment re the intern's comments.
Should we flag these kinds of attacks in an Issue for future reference?
Rails 8 has a rules-based way of (temporarily) blocking IPs. Could that reduce some server load/noise?

config/application.rb
mo-nathan: Good question, and I mostly land where you'd have a year ago. The reversion-avoidance value is real, but it lives almost entirely in *one* well-placed "why" at the load-bearing site (here, where the middleware is wired and `insert_before(0)` is non-obvious). Repeating the same rationale in the Gemfile and the test doesn't add proportional safety -- a reader has the whole file/PR in view -- but it does triple the staleness surface, which is exactly the cleanup-on-change problem you flagged: change the behavior and a missed copy becomes a *lying* comment, which is worse than none.

So I kept the one authoritative comment here and trimmed the Gemfile and test docstring to brief pointers back to it (0b2909cd89). One source of truth to update; pointers everywhere else. And separately noted for myself that the smell isn't "commenting" -- it's over-explaining and duplicating; a single *why* earns its keep, three copies of it don't.

mo-nathan:
I don't think there's much value in flagging these as long there's no noticeable impact.  Feels too much like whack-a-mole.  That's why these more systematic approaches are better.

PR #4853, opened 2026-07-19
Resync read-only reflections from their iNaturalist source + `Sync now` button (#4215)

mo-nathan:
Part of #4215 / #4208. **Slice 2** of the read-only-reflection work, **stacked on #4852** (`njw-4214-readonly-reflections`) — base branch is that PR, not `main`. Together they are the "read-only reflections + resync" pair: read-only only makes sense once "edit it at the source and resync" is a real path, so the two **ship together.** GitHub will show only this slice's diff once #4852 merges.

## What this does

`Inat::ObservationResyncer` re-fetches a reflection's current iNaturalist data via the existing public, deletion-aware `Inat::ObsFetcher#fetch_batch` and refreshes the **scalar core in place** — date, location, GPS, notes — using the same field mapping the importer uses on create. It runs **below** the read-only edit guard from #4852: that guard blocks the web/API *edit actions*, while this service writes the record directly, so it can refresh a reflection the user is (deliberately) not allowed to hand-edit.

Triggered by a **"Sync now" button** on the imported-source banner → `Observations::InatResyncsController` → `InatObservationResyncJob` (background, because the iNat fetch is rate-limited). `Observation#resyncable_by?` gates who sees and can trigger it (owner or editor; admin server-side).

## Safety (per #4215)

- **Transient fetch failure** → nothing is touched (`fetch_batch` distinguishes a failed batch from a genuinely-absent id).
- **Source deleted on iNat** → `:source_deleted`; every MO record is kept and the loss is logged. No data is destroyed.
- **Only read-only reflections are refreshed** — the still-editable import backlog (`reflected_at` nil) is left alone, so a resync can never clobber MO-side edits.
- **Change detection uses `saved_changes`**, not an in-memory `changed?`: setting `location` fires a callback that rewrites `where` to the location's name, so an assigned-value comparison never converges. `where` is only set when no Location resolves. `last_synced_at` (already on `external_links`) records the last check.

## Scope

This slice refreshes the **scalar core only**. Deliberately deferred to follow-up PRs:

- image add / remove / relicense / delete diffing;
- sequence string-comparison diffing;
- naming re-weight on taxon change — which first needs a **provenance marker on votes/namings** (they have none today; `external_site_id` lives only on `external_links`) plus the consensus-as-one-external-voter change;
- the **daily batch** resync job (this PR is the user-initiated path).

## Test plan

On a dev DB with production data, take a read-only reflection — one of your own imports, or any imported observation with `reflected_at` set (`Observation.find(<id>).update_column(:reflected_at, Time.current)`):

1. **Sync now.** On the observation's show page, the imported-source banner shows a **"Sync now"** button (only for you if you own it / can edit it). Click it → you're redirected back with a "Syncing…" flash and an `InatObservationResyncJob` is enqueued.
2. **Scalar refresh.** If the iNat source's date / location / notes differ from the MO copy, after the job runs the MO observation's date / location / notes now match the source, and the import link's `last_synced_at` is set. Re-syncing unchanged data is a no-op (no spurious activity-log entry).
3. **No clobber of the backlog.** On an imported observation whose `reflected_at` is still `NULL`, there is no Sync button, and calling the resync service leaves it untouched (`:not_a_reflection`).
4. **Deleted source is safe.** For a reflection whose iNat observation no longer exists, a sync keeps all MO data and logs that the source is gone (no deletion).

_Automated coverage: `test/classes/inat/observation_resyncer_test.rb` (synced / unchanged / deleted-source / transient-failure / non-reflection), the controller, the job, `Observation#resyncable_by?`, and the banner Sync button._


mo-nathan:
The INAT tool tip really gets in the way of the expanded area.  It shows stuff that I could not read and it was difficult get the tool tip to go away after the click.  In my opinion the tool tip should go away after I click on the button.  I also don't understand why it is "INAT" vs "iNat".

Screenshot 2026-07-23 at 1 43 43 PM

mo-nathan:
The date needs a time in the panel especially given the iNat caching delay.  It's important to know that I did a sync 1 minute ago as oppposed to half an hour ago.  Date alone doesn't cut it.

mo-nathan:
Here's a draft of an article to post when this goes live:

[imported-observations-sync.txt](https://github.com/user-attachments/files/30358635/imported-observations-sync.txt)

mo-nathan:
## `Sync now` isn't reachable from the occurrence — I think this needs resolving before this goes live

@nimmolo — I expect you'll have opinions here, since you're the one who moved the button into `InfoFrame`.

While drafting the community article for this work I hit a gap in where `Sync now` is reachable from.

In `Views::Controllers::Observations::ExternalLinks::InfoFrame`, the read-only note and the sync button are both emitted from `render_own_row` and gated on `@obs.reflection?`. `render_sibling_row` — the row that shows an occurrence sibling's link to the same site — renders only the external link plus the `(MO #N)` attribution. So a sibling row gives no indication that the record it points at is a read-only mirror, and offers no way to refresh it.

The consequence is that `Sync now` only exists on a reflection's *own* page. That would be fine if the reflection were where people worked, but it isn't the workflow we're steering them toward. The recommended way to correct something in a reflection's mirrored core, when you can't edit the iNat record, is to make your own observation of the same collection and group the two as an occurrence — at which point your observation is the primary and *that* is the page you're living on. From there you can see the reflection's iNat link and click through to it, but you have to navigate to the reflection first to press a button that refreshes data being displayed to you on the page you were already on.

Since the article tells people to use exactly that path, I don't want to publish it describing a detour. I'd like this resolved before this ships.

### Roughly what I think it needs

- Move the gating off `@obs` and onto the row: a sibling row whose observation is a resyncable reflection should get the same read-only note and `Sync now` button that an own row gets.
- Evaluate the permission against the sibling, not the page's observation — `resyncable_by?` has to be asked of the observation actually being synced, not of `@obs`.
- Point the button at that observation's `resync_observation_path`, and make sure the Turbo confirm and the resulting flash make it clear *which* record is being synced when it isn't the one whose page you're on.

### The design question I'd like your read on

I've been describing this as "per-link sync buttons," which is the right presentation — the button belongs next to the link it refreshes. But the underlying action is per-*observation*: `resync_observation_path(id)` refreshes a reflection's whole scalar core from its import link. Today that distinction is invisible, because a reflection has one import link. It stops being invisible if an observation can ever mirror more than one source, and it reads oddly the moment a site pane shows several rows.

So: do we keep per-link buttons as a pure presentation layer over a per-observation action and accept that two rows could trigger the same sync, or is it worth making resync take the link rather than the observation now, while there's only one caller? I lean toward the former for this PR — it's the smaller change and nothing today produces the ambiguous case — but you've been closer to this code than I have lately.

Also worth deciding: whether a sibling row's note should differ in wording from an own row's. "Read-only reflection of its imported source" reads naturally on the reflection's own page and slightly oddly on someone else's.

app/classes/inat/observation_resyncer.rb
```diff
@@ -0,0 +1,194 @@
...
+      attrs[:where] = inat_obs.where if location.nil?
+      attrs
+    end
+
+    # The iNat obs is gone: keep every MO record intact, record the loss on
+    # the activity log, and still stamp last_synced_at so the check ran.
+    def handle_deleted
+      mark_synced
+      @observation.log(:log_observation_source_deleted,
+                       user: @observation.user)
```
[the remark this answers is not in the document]
mo-nathan: Fixed in 1515690e8b - both log calls now use the triggering user the resyncer already captures (`@user`), via a `logging_user` helper. Since the daily-batch slice will pass no user, that falls back to `User.admin` rather than the owner, matching the `current_user || User.admin` system-actor pattern in `Location` and `NameDescription`. Covered by two new tests in `observation_resyncer_test.rb` (triggering user, and the unattributed/admin case).

app/models/observation.rb
```diff
@@ -1128,6 +1128,25 @@ def import_link
     end
   end
 
+  # True when this observation is a read-only reflection of an imported
+  # source (#4214): its scalar core (date/location/notes/name) mirrors
+  # the source and is kept current by resync (#4215), so MO-side edits to
+  # those fields are blocked. `reflected_at` is stamped at import time for
```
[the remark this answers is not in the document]
mo-nathan: Fixed in 1515690e8b - the doc now describes what resync actually refreshes (date/location/GPS/notes) and says explicitly that `name` is excluded because iNat identifications are mirrored at import time only, with afterwards-tracking waiting on the identification-sync slice of #4215.

config/locales/en.txt
```diff
@@ -1127,6 +1129,15 @@
   source_credit_external: '"Imported from [name]":[url]'
   source_credit_external_text: Imported from [name]
   source_credit_help_link: About imported observations
+  observation_reflection_read_only_note: Read-only reflection of its imported source. To change it, edit it at the source and resync.
+  observation_resync_button: Sync now
```
nimmolo: This seems like it should be a top-level tag, :sync_now, since it may be reused for other sites, right?

config/locales/en.txt
```diff
@@ -1127,6 +1129,15 @@
...
+  observation_resync_confirm: If you recently edited this observation on iNaturalist, it may take up to 10 seconds for those changes to appear there. Ready to sync now, or would you rather wait?
+  observation_resync_started: Syncing this observation from its source. This may take a moment.
+  observation_resync_denied: You do not have permission to sync this observation.
+  observation_resync_synced: Synced with iNaturalist.
```
nimmolo: Should maybe be made flexible, "Synced with [site]"

config/locales/en.txt
```diff
@@ -1127,6 +1129,15 @@
...
+  observation_resync_unchanged: Already up to date with iNaturalist.
+  observation_resync_source_deleted: This observation's iNaturalist source could not be found. It may have been deleted.
+  observation_resync_failed: Couldn't sync with iNaturalist right now. Please try again later.
```
nimmolo: These too, maybe [site] is a pass-in arg

nimmolo (commented):
I like the crud routes

mo-nathan:
While there isn't a fully automated way that an observation can have more than one reflection, both the UI and the field slip functionality allows a user to build such an occurrence and this is a very realistic workflow (likely to happen next week).  All that's needed are two iNat observations that are for the same specimen (e.g., same field slip code).  If both iNat obs get reflected into MO, then you would end up with an occurrence which has two reflections.  However, from a user perspective it still would make sense for there to be a single "Sync now" button which would sync all of the reflections.  I've reallized that users really don't care which reflections are getting synced, but would want all the reflections to be synced a essentially the same time.

config/locales/en.txt
mo-nathan: Fixed in 9a7a7a3fbe - now a top-level sync_now tag (lowercase, per the generic-verbs section), rendered with .ti for the button label.

config/locales/en.txt
mo-nathan: Fixed in 9a7a7a3fbe - now 'Synced with [site].'; the resyncer passes ExternalSite::INATURALIST_NAME.

config/locales/en.txt
mo-nathan: Fixed in 9a7a7a3fbe - unchanged/source_deleted/failed (and the confirm dialog, which also named iNaturalist) all take [site] as an arg now. The confirm uses the info frame's site_name prop; the flash messages get the site from the resyncer.

mo-nathan:
Following up on the design question above, resolved in discussion with @mo-nathan — sync becomes **occurrence-wide**. Rework in progress on this branch:

- **One `Sync now` button** (at the bottom of each expanded site pane) on every observation whose occurrence contains at least one reflection. It enqueues a **single job** that refreshes every reflection in the occurrence — `Inat::ObsFetcher#fetch_batch` already takes a list of ids, so the whole occurrence is one rate-limited API call and one aggregate report. Multi-reflection occurrences are a realistic near-term workflow (two iNat obs for the same field-slip code), and users want all reflections current at once, not per-link control.
- **Anyone logged in can trigger it.** The owner/editor gate goes away: resync applies no user input, it converges on source-canonical data — the same thing the planned scheduled batch will do with no user at all.
- **Attribution goes to the admin account**, matching that future scheduled job (supersedes the per-triggerer logging from the earlier review round).
- **Mixed outcomes** (rare) get one aggregate flash: refreshed/up-to-date counts, source-deleted called out explicitly, worst outcome drives the alert level.
- **Spam guard:** if every reflection in the occurrence was synced within the last 10 seconds, the click reports "just synced" and skips the fetch.
- This supersedes the per-site "Synced with [site]" wording from the last push — completion messages become occurrence-level and source-generic. The reusable `sync_now` tag stays.

mo-nathan:
Occurrence-wide sync landed in 1f19abc4f3, as described above: one `Sync now` button per site pane shown to any logged-in viewer, one batched job per click (all reflections in a single `fetch_batch` call), admin-owned log attribution, aggregate mixed-outcome flash broadcast to every member's page, 10-second re-sync guard, and read-only notes on sibling reflection rows. The earlier per-site "Synced with [site]" flash tags are replaced by occurrence-level wording.

mo-nathan:
Correction to the guard from 1f19abc4f3, per discussion: it checked `last_synced_at`, which the job stamps only on COMPLETION — so it didn't debounce clicks at all while a sync was still queued/fetching. Fixed in 406f4dad44: enqueueing writes a 10-second solid_cache key (keyed on the occurrence's reflection set), so re-clicks are swallowed from initiation, whichever member page they come from.

PR #4861, opened 2026-07-21
`i18n`: consolidate `:ALL_CAPS`/`:all_caps` translation tags, add `Symbol#ti`

### LINKED ISSUE #4843: More efficient localizations

nimmolo:
## Important

> [!WARNING]
> When you get on this branch, you must run `db:migrate` **_before_** `lang:update`, as `deploy.sh` does in production. Running it in the opposite order loses the theme name and standalone-tag translations in your db.

> [!NOTE]
> This PR does not change the mechanism of i18n from files to cache. It only removes the `:UPPERCASE`/`:lowercase` "twin string" duplication in `en.txt` and adds a `Symbol#ti` method that derives a title-cased version from the lowercase string, using locale-specific rules.

## Summary

Fixes #4843.

### Eliminates ALL-CAPS duplicate translation tags

These "twins" only exist for common strings (`:OBSERVATION`/`:observation`, etc.). To confirm they could be safely deleted, we checked every language for translated strings where:

- **The ALL-CAPS and lowercase versions differed in more than casing.** We determined the correct value for each -- sometimes that meant preferring the ALL-CAPS tag's wording, sometimes correcting an unrelated terminology issue found along the way (e.g. several locales' "herbarium"/"fungarium" tags used the generic botanical word instead of the fungarium-specific term MO actually means; confirmed against real mycological institutions' own naming conventions per locale), and in at least one case neither side was right and both needed replacing. If the lowercase tag was already correct, we left it alone -- it becomes the sole surviving translation once the ALL-CAPS tags are dropped. Where it wasn't, we updated the production DB directly. 42 of these, each individually reviewed before correcting.
- **The ALL-CAPS tag existed but the lowercase one didn't.** 2 of these; the value was copied over so it isn't lost when the ALL-CAPS tag is dropped.

Full audit data -- deviation charts per language -- is in #4843.

### New titleizing algorithm, `Symbol#ti`

After analyzing casing patterns across all existing ALL-CAPS translations, we landed on an algorithm that title-cases each localized string appropriately for its locale. Locales are grouped as either Title Case, like English (or rather, Title Case _with_ Connecting Words Excepted), Sentence case no matter how many words, or skipped, because the locale doesn't use an alphabet with capital letters.

The locale-specific casing rules were derived by auditing every existing ALL-CAPS translation against what `.ti` would produce, against a fresh production checkpoint -- not assumed. Four locale families emerged:

- per-word title-case (en/es/pt/el/uk/be, plus tr with Turkic i/I mapping);
- sentence-case (pl/ru/fr/it -- confirmed as the dominant convention for pl/ru via a cross-reference against uk);
- no titleizing (de, where nouns are already capitalized mid-sentence, and locales with no letter-casing concept at all: ar/fa/zh/jp).

See #4843 for the full audit.

  ```ruby
  :species_lists.l  == "observation lists"
  :species_lists.ti == "Observation Lists" # both words capitalized in English
  :species_lists.ti == "Listes d'espèces"  # sentence case in French
  :species_lists.ti == "Listas de Observaciones" # connecting word "de" stays lowercase in Spanish

  :species_lists.l  == "Artenlisten" # German nouns are already capitalized
  :species_lists.ti == "Artenlisten" # .ti is a no-op here -- .l is already correct
  ```

### Why the change

TL;DR: less translator busywork, and helpful groundwork for #4811 (moving translations to a cache) -- twin ALL-CAPS/lowercase tags meant two different casings of the same concept could collide under `translation_strings.tag`'s case-insensitive collation; a cache keyed on tag doesn't have to work around that anymore once there's only one casing per concept.

Two independently-translated strings per concept is translator busywork that conflates **translation** (what does this word mean) with **presentation** (how is it cased here) -- a rendering decision, not a linguistic one, for the ~332 affected concepts (nearly all single words). One canonical string plus `.ti`:
- Eliminates key-lookup typos like `:Vote.t` (no second tag to typo against).
- Halves translator burden per concept, across all 15 languages.
- Fits the existing `.l`/`.t`/`.tl`/`.tp`/`.tpl` API shape -- `.ti` is one more "same lookup, different post-processing."

### Why not just downcase the Title Case version instead

German capitalizes nouns mid-sentence, so its "lowercase" tag is often already Title Case -- downcasing it programmatically would make the German wrong. For the rest, per-word capitalization habits genuinely differ by locale (confirmed by audit, not assumed), so a single global rule wasn't safe either.

## What changed

- **Production DB prep** -- the 44 translation fixes described above (42 corrected, 2 recovered from a missing lowercase counterpart), so the correct value survives once the ALL-CAPS tags are dropped.
- **`Symbol#ti`** (`app/extensions/symbol.rb`) -- locale-aware dispatcher (`Symbol.titleize_localized`) covering the four locale families above. Per-word title-casing (`capitalize_each_word`) preserves any word that already has an uppercase letter past its first position (handles acronyms -- "API", "ICN", "ДНК", "IDs" -- with no inflection config needed), treats apostrophes/hyphens/parentheses as part of the attached word (needed for Portuguese gender-suffix notation like "editor(a)"), and applies small per-locale connector-word exceptions (English "at"/"by"/"or"/"with"/"and", Spanish/Portuguese/Greek/Ukrainian articles and prepositions). Sentence-case locales use `capitalize_first_letter_only` rather than `String#capitalize`, since the latter also downcases the rest of the string, destroying embedded acronyms the lowercase tag already stores correctly. The `[:tag.ti]` embedded-ref syntax routes through the same helper. Covered by `SymbolExtensionsTest`, exercising every branch across real translated content in all 16 production locales.
- **Deleted all 399 ALL-CAPS tags from `en.txt`** (330 twin pairs + 69 standalone tags with no lowercase twin) and rewrote every call site -- ~1100+ sites, including ~150 built dynamically at runtime that a static regex sweep couldn't find; two full audit passes closed that gap (see #4843).
- **Migration** (`db/migrate/20260718120000_lowercase_standalone_translation_tags.rb`) renames the 69 standalone tags plus 5 theme-name tags via `TranslationString.rename_tags`, and adds a `[language_id, tag]` index the table never had.
- **`ID`/`OK`/`ICN` acronyms** added to `config/initializers/inflections.rb`; two Zeitwerk-autoloaded classes renamed to match (`Components::IDBadge`, `API2::ObjectNotFoundByID`).
- **New RuboCop cop**, `lib/rubocop/cop/mo/lowercase_translation_tag.rb` -- flags any Symbol literal used with `.l`/`.t`/`.tl`/`.tp`/`.tpl`/`.ti` containing an uppercase letter.
- **New test**, `LocalizationFilesTest#test_no_uppercase_tag_keys` -- asserts no tag key in the DB-backed English translation set contains an uppercase letter.

## Test plan

- [x] Full suite green locally, confirmed by the user after fixing regressions this sweep itself introduced (missing-tag misses from dynamically-built symbols, caught by the existing `Symbol.missing_tags` teardown check)
- [x] `bundle exec rubocop` clean across the full repo, including the new cop running against its own sweep output
- [x] Collation-safe audit (case-sensitive comparison on fetched rows, never trusted to the DB's case-insensitive `=`/`IN` match) confirmed zero data loss across all 330 twin-pair tags in every non-English language
- [x] The 69 standalone tags plus 5 theme-name tags are covered by a real migration, run and verified against a production checkpoint
- [x] `Rails.application.eager_load!` clean after the acronym-driven class renames

---

Separately, the same production-checkpoint audit surfaced real translation-content bugs beyond casing (typos, wrong words, missing translations) across 11 locales -- documented and resolved as production-database edits via #4843's comment thread, independent of this PR's code changes.


JoeCohen (approved):
I'm good with this.
Note that for some symbols translators now see a call to `.ti`. Example (Spanish)

- Production 
INDEX_OBJECT: [TYPE] [:INDEX]
- This PR
index_object: [TYPE] [:index.ti]

This is a little more confusing. But IMO, not worth messing with.

mo-nathan (approved):
My concerns have more to do with solid cache at this point than with this PR.  This checks out as far as I can tell.  Still makes me nervous, but we can fix issues on the other side.

PR #4876, opened 2026-07-23
Delete confirmed-complete parity tests

nimmolo:
## Summary

Parity tests exist to verify a Phlex conversion didn't silently change behavior, comparing an `old_html`/`new_html` pair via `assert_html_element_equivalent`. Per `.claude/rules/phlex_reference.md`'s existing convention ("Delete the `_Old` copy and the parity test together... once the refactor is confirmed green"), they're meant to be temporary — once confirmed, they're testing history, not current behavior.

Confirmed zero `*_Old.rb` files remain anywhere under `app/`, meaning every conversion these were protecting is complete. Deleted every remaining parity test found across the repo:

- `test/components/image/reuse_form_test.rb` — `test_toggle_link_parity` + `OldReuseFormToggleLink`
- `test/components/help/sweep_parity_test.rb` — entire file (all 4 tests were parity tests)
- `test/views/controllers/comments/index_test.rb` — `test_list_group_parity_with_legacy_comment_row_render` + `LegacyCommentList`
- `test/views/controllers/occurrences/projects/form_test.rb` — `test_cancel_link_parity` + `OldCancelLink`
- `test/views/controllers/observations/show/namings/row_test.rb` — `test_matching_observations_link_parity` + `OldMatchingObsLink`
- `test/views/controllers/observations/show/namings/footer_buttons_test.rb` — `test_suggest_button_parity` + `OldSuggestButton`

Every real behavioral test in each of these files is untouched.

## What's deliberately NOT touched

- `test/component_test_case.rb`'s `assert_html_element_equivalent` helper (and its own dedicated test, `test/component_test_case_test.rb`) — this is reusable test infrastructure for the *next* Phlex-to-Phlex refactor, not itself a parity test. Nothing in the production test suite calls it anymore after this PR, but it's worth keeping on hand rather than deleting and re-writing it next time this pattern is needed.
- `test/components/application_form_helper_parity_test.rb` — the filename says "parity," but it contains no old/new HTML comparison anywhere; every test asserts directly on a current-behavior fixture form's output. It's a real regression suite that happens to have "parity" in its name (its own header comment: "Pins behaviors where the Phlex helpers were historically diverging... exists to prevent regression"). Left alone.

Also fixed a stale citation in `.claude/rules/phlex_reference.md`: the "multiple anonymous stub Phlex views" section pointed at `comments/index_test.rb`'s `LegacyCommentList`/`NewCommentList` pair as an example of the named-nested-class pattern — `LegacyCommentList` is now gone, so it no longer demonstrated what it claimed to. The code example directly above it is already self-contained, so just dropped the now-inaccurate pointer sentence.

## Test plan

- [x] `bundle exec rubocop` on all changed files — clean.
- [x] Full test run of all 6 touched test files — 58 tests, all green.
- [x] `bin/rails test test/classes/localization_files_test.rb` — green (this is the duplicate-def scanner the phlex_reference.md section is about).


nimmolo:
Checked coveralls (build [80742523](https://coveralls.io/builds/80742523)) for a per-file coverage regression from this deletion, not just the headline diff percentage.

Traced every production class the deleted parity tests could plausibly have exercised — `Components::Button`, `Components::Button::Get`, `Components::Help`, `Components::Help::CollapseBlock`, `Components::Form::LocationMap`, `Components::Icon`, `Components::Collapsible`, `Components::ListGroup`, `Components::ListGroup::Item`, `comments/index.rb`, `comment_row.rb` — including the 13 parity tests removed from `test/components/collapsible_test.rb` (541 lines) covering call sites like `Form::TableAccordion`, `Observations::Form::Details`, `Views::Layouts::TopNav`, `Header::IndexBar::FilterCaption`, `Projects::Aliases::Form`, and `InatImports::Form`.

All of them report 100% coverage, 0 missed lines, in this PR's build. Most of those `Collapsible`-related parity tests rendered `Components::Collapsible` directly with hand-copied params rather than actually invoking the real production call site, so they never contributed coverage to those files to begin with — nothing to restore here.

PR #4877, opened 2026-07-23
Script to backfill MCP ExternalLinks

### LINKED ISSUE #4819: backfill script MCP ExternalLinks
### Tasks

- [x] Decide ExternalLink.relationship (Mirror or Export)
- [x] "100000 processed" => "100000 images processed"
- [x] track time
- [x] Also backfill Observations
- [x] target_type: "Image",
- [x] external_id: nil,
- [x] speed it up?
- [x] relationship: "export",
- [x] last_synced_at: nil,
- [x] import_target: nil,
- [x] external_created_on: nil
- [x] url: nil,
- [x] local dry run
- [x] local backfill
- [x] Coverage
- [x] What to do about obss, imgs missing on MO?
- [x] double check time needed to run script locally: 
6-7 minutes 
- [x] add 1-second delay between writes?
Rejected because 6-7 minutes acceptable if run at low usage time.

JoeCohen:
Rails runner script to create the baseline backfill of ExternalLink's to MyCoPortal for both images and observations

#### Setup

1. Download a backup of MUOB in UTF-8 format
to tmp/mycoportal_dwca/ using the instructions at
<https://docs.symbiota.org/Collection_Manager_Guide/Downloading/downloading_copy>

2. Optional paranoid check of local db.  Confirm that there are no existing External Links to MyCoPortal
From the Rails console:

```ruby
 ExternalLink.where(target_type: "Observation", external_site: ExternalSite.find_by_name("MyCoPortal"), relationship: :export).count
 => 0

 ExternalLink.where(target_type: "Image", external_site: ExternalSite.find_by_name("MyCoPortal"), relationship: :export).count
 => 0

```

#### Local Dry Run

From the command line: `rails runner script/backfill_mycoportal_export_links.rb`
(Takes 2m 22s on a MacBookPro with Apple M5 Pro chip, 24 GB memory)

Expected Result:

```txt
  Images:
    created: 1231908
    already present (skipped): 0
    invalid (see warnings above): 0
    not found in MO: 297
    unparseable: 0
  Observations:
    created: 388856
    already present (skipped): 0
    invalid (see warnings above): 0
    not found in MO: 1876
    unparseable: 0
  missing report: -> /Users/joe/mushroom-observer/tmp/mycoportal_dwca/mycoportal_backfill_missing.csv
  skipped report: -> /Users/joe/mushroom-observer/tmp/mycoportal_dwca/mycoportal_backfill_skipped.csv
  elapsed: 2m 22s
```

Note: 

- Observations created + Observations not found == [MCP Collection Statistics](https://mycoportal.org/portal/collections/misc/collprofiles.php?collid=36) 390,732 specimen records 
- Images created + Images not found == total rows in multimedia.csv (1,232,205)
(This is one greater than shown in [MCP Collection Statistics](https://mycoportal.org/portal/collections/misc/collprofiles.php?collid=36) 1,232,204 total media)
Reason: The DwCA backup supplied by MCP has one more image row than the media total.)

### Local Applied Run

`rails runner script/backfill_mycoportal_export_links.rb --apply`
(Takes 41m 3s on a MacBookPro with Apple M5 Pro chip, 24 GB memory)

Expected result:

```
   Images:
    created: 1231907
    already present (skipped): 1
    invalid (see warnings above): 0
    not found in MO: 297
    unparseable: 0
  Observations:
    created: 388856
    already present (skipped): 0
    invalid (see warnings above): 0
    not found in MO: 1876
    unparseable: 0
  missing report: -> /Users/joe/mushroom-observer/tmp/mycoportal_dwca/mycoportal_backfill_missing.csv
  skipped report: -> /Users/joe/mushroom-observer/tmp/mycoportal_dwca/mycoportal_backfill_skipped.csv
  elapsed: 41m 40s
```

NOTE: - Images created + Images not found == [MCP Collection Statistics](https://mycoportal.org/portal/collections/misc/collprofiles.php?collid=36) 1,232,204 total media)
(The script noticed the duplicate row in the MCP-supplied DwCA backup, and avoided creating a duplicate ExternalLink.)


script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,488 @@
...
+# One-time backfill of ExternalLink (relationship: :export) rows for every
+# Observation and Image already present in MyCoPortal (MCP)
+# (Based on a Darwin Core Archive (DwC-A) backup of the MUOB collection)
+
+# The backup must be downloaded manually first (see USAGE below)
+# and placed in DEFAULT_DWCA_DIR (tmp/mycoportal_dwca/).
+
+# This script finds the newest such zip and extracts
+# occurrences.csv / multimedia.csv
+# The extracted CSVs are deleted bye default once the run finishes;
```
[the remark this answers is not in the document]
JoeCohen: Already fixed as of 2024840 (typo corrected: "by default") - no change needed here.

script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,488 @@
...
+# Pass --keep-csvs to retain them
+# Pass --occurrences/--multimedia to skip the zip entirely and use
+# already-extracted files instead.
+#
+# Observations/Images MCP has that don't exist in MO are written to a report,
+# bydefault <dwca-dir>/mycoportal_backfill_missing.csv
+# pass --missing-out to override.
```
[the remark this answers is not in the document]
JoeCohen: Already fixed as of 2024840 - both this line and line 22 corrected to "Pass ... by default" - no change needed here.

test/script/backfill_mycoportal_export_links_test.rb
```diff
@@ -0,0 +1,619 @@
...
+    multimedia_csv = write_csv(
+      %w[coreid identifier MetadataDate],
+      [{ "coreid" => "1", "identifier" => image_url(image.id),
+         "MetadataDate" => "2019-07-22 17:05:18" }]
+    )
+    occurrences_csv = write_csv(%w[id catalogNumber],
+                                [occurrence_row(1, "MUOB 1")])
+    @missing_out = Tempfile.new(["missing", ".csv"]).path
+    @skipped_out = Tempfile.new(["skipped", ".csv"]).path
+    subject = BackfillMycoportalExportLinks.new(
```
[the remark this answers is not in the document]
JoeCohen: Already fixed as of 2024840 - all report-path Tempfiles (@missing_out/@skipped_out) now go through a `temp_csv_path` helper that retains the Tempfile object in `@tempfiles` for the test's duration, mirroring the same fix already applied to `write_csv` for the CSV input files. Covers all the flagged locations (93, 207, 332, 412, 493, 553, and the remaining one) - no change needed here.

test/script/backfill_mycoportal_export_links_test.rb
```diff
@@ -0,0 +1,619 @@
...
+
+  def test_picks_the_newest_matching_zip
+    image = images(:in_situ_image)
+    dwca_dir = Dir.mktmpdir("test_dwca")
+    older = build_dwca_zip(dwca_dir, "MUOB_backup_2020-01-01_000000_DwC-A.zip",
+                           [occurrence_row(999, "MUOB 999")],
+                           [multimedia_row(999, image_url(999_999_999))])
+    # File.utime needs a plain Time, not an ActiveSupport::TimeWithZone --
+    # this is filesystem-mtime manipulation, not app-facing date logic.
+    old_time = ::Time.now - 100 # rubocop:disable Rails/TimeZone
```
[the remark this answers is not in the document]
JoeCohen: Already fixed as of 2024840 - swapped to `Time.current.to_time - 100`, no `rubocop:disable` needed anymore - no change needed here.

script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,488 @@
...
+  def record_missing(target_type:, occid:, mo_id:, image_id: nil)
+    increment_stat(target_type, :mo_missing)
+    @missing << [target_type, occid, mo_id, image_id]
+  end
+
+  # Already linked -- report which ExternalLink covers this record
+  # alongside the plain already_present count, for traceability.
+  def record_skipped(target_type:, occid:, mo_id:, link_id:, image_id: nil)
+    increment_stat(target_type, :already_present)
+    @skipped << [target_type, occid, mo_id, image_id, link_id]
```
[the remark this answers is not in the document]
JoeCohen: Already fixed as of 2024840 - missing/skipped rows now stream straight to CSV via a new `ReportWriter` (`CSV.open` + incremental `<<` per row as each is discovered) instead of buffering into growing `@missing`/`@skipped` Arrays for the whole run - no change needed here.

script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,436 @@
...
+# Re-running this script doubles as a reconciliation check:
+# in a dry run, a nonzero "would create" count means MCP has
+# observations/images MO doesn't think it exported.
+
+# USAGE:
+#
+#   Before running the script, create a backup of the MUOB data and
+#   download it to tmp/mycoportal_dwca/
+#   using the instructions at
+#   https://docs.symbiota.org/Collection_Manager_Guide/Downloading/downloading_copy
```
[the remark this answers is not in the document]
JoeCohen: @copiot:This is incorrect. Rubocop excepts URLs from the line-length cop

JoeCohen:
@mo-nathan

Is it correct to leave these blank:
- `ExternalLink.last_synced_at` I don't interpret this as a sync. 
- `ExternalLink.url`. Seems like we don't do this for iNat mirrors, and it doesn't matter because the view derives this on the fly.
  - for images. In any event, I cannot provide an image URL. MO validation requires the URL domain to be the ExternalSite. But AFAIK, MCP image URLs are external to MCP (they're mostly links to MO).
  - for observations. I could provide this.

Other: 
- Script took about 2hrs 15 minutes on my local (desktop with Intel CPU). [Took under 1 hr on Apple M5 Pro laptop]
Will that create issues when run on production?
Intern's main suggestion for improved speed is to use `insert_all` instead of `create`. See Stage 5 at https://github.com/MushroomObserver/mushroom-observer/issues/4819#issuecomment-5060472176 
The intern points out that `insert_all` bypasses callbacks and validations. That makes me nervous.

- What to do about observations and images which exist on MCP but not on MO?
IMO, just ignore them. 
  - Users delete images on MO and we should respect that. Also MO users get deleted; we shouldn't restore them.
  - We lack privileges to delete MCP observations or images.

mo-nathan:
> Is it correct to leave these blank:
> 
> * `ExternalLink.last_synced_at` I don't interpret this as a sync.

Yes.

> * `ExternalLink.url`. Seems like we don't do this for iNat mirrors, and it doesn't matter because the view derives this on the fly.

Yes.  The only reason this is lingering on are the MCP links.  I recommend moving the MCP links that you (or the Intern) can't turn into a unique MCP link to a comment or notes.  Getting to the point where all the MCP external links are derived from the external id is the goal.  Then we can actually get rid of the `url` field.

>   * for images. In any event, I cannot provide an image URL. MO validation requires the URL domain to be the ExternalSite. But AFAIK, MCP image URLs are external to MCP (they're mostly links to MO).

I don't think Image external links are all that important.  The observation URLs are way more important and the image data can be effectively derived from them.

> Other:
> 
> * Script took about 2hrs 15 minutes on my local (desktop with Intel CPU). [Took under 1 hr on Apple M5 Pro laptop]
>   Will that create issues when run on production?

This is where running a script locally and then exporting/importing the data could make a difference.

>   Intern's main suggestion for improved speed is to use `insert_all` instead of `create`. See Stage 5 at [backfill script MCP ExternalLinks #4819 (comment)](https://github.com/MushroomObserver/mushroom-observer/issues/4819#issuecomment-5060472176)
>   The intern points out that `insert_all` bypasses callbacks and validations. That makes me nervous.

I'm not too worried about callbacks and validations on ExternalLinks.  You could ask the intern about what callbacks and validations exist and if that can be precomputed brefore doing the insert_all.

> * What to do about observations and images which exist on MCP but not on MO?
>   IMO, just ignore them.

Why is the Intern even raising this as an issue?  I'm cool with ingoring them for the backfill step.

>   * Users delete images on MO and we should respect that. Also MO users get deleted; we shouldn't restore them.

Again why is this relevant?  This script is just creating / clenaing up ExternalLinks for MCP.  No other data is getting added, right?

>   * We lack privileges to delete MCP observations or images.

Yeah, of course.  Why is this relevant?

JoeCohen:
@mo-nathan Thanks for all the responses! 
All the questions were mine, not the intern's. For several I was trying to think ahead/deal with edge cases.
I'll ask the intern tomorrow about moving data from my local db to the server.
But won't actually do that until you give me the go-ahead.

script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,443 @@
...
+#   Before running the script, download a backup of MUOB in UTF-8 format
+#   to tmp/mycoportal_dwca/ using the instructions at
+#   https://docs.symbiota.org/Collection_Manager_Guide/Downloading/downloading_copy
+#   (The download will have a name like MUOB_backup_2026-07-22_190455_DwC-A.zip)
+#
+#   # Dry run by default
+#   bin/rails runner script/backfill_mycoportal_export_links.rb
+#
+#   # Write to the database (Idempotent)
+#   APPLY=1 bin/rails runner script/backfill_mycoportal_export_links.rb
```
mo-nathan: It's interesting that your Claude seesion went the `APPLY=1` approach.  During older Claude sessions it would do this for me as well, but during more recent ones it switched to `--commit` or `--apply` as an argument after the script.  Personally, I prefer the flag at the end of the command vs. the environment variable approach.  I reviewed this in a Claude session and here's the analysis and recommendation:

```
Good catch — there's no stored convention, so each session just picked one of the two common idioms. They're functionally near-equivalent — both are a boolean the script reads, both default safe if the script treats "absent" as dry-run — but they aren't equivalent in their failure modes and ergonomics:

Where the env var wins:
- It passes through wrapper layers without forwarding — relevant for rake tasks (where argv parsing is genuinely awful, rake foo -- --apply), and it's the Rails/rake house style (VERSION=... rails db:migrate). For bin/rails runner script.rb, though, argv passes through cleanly, so this advantage mostly doesn't apply to how we actually run backfills here.

Where the flag wins:
1. No ambient persistence. APPLY=1 cmd is per-invocation, but the mechanism invites export APPLY=1 — which then silently arms every later script in that shell that honors the same variable. A flag cannot linger by construction. For a destructive/not-destructive toggle, "explicit at every invocation, impossible to leave switched on" is the property you want.
2. Typos can fail loudly. APLY=1 is undetectable — the script just quietly dry-runs (or worse, if someone inverts the default). A script that parses argv strictly can abort on --aply / any unknown arg. Env-var typos can never be validated; flag typos can.
3. Self-documentation. The flag shows up in the script's usage line and in ps output; an env var is invisible unless you read the source.
4. Workflow. The natural loop is: run dry, read output, up-arrow, append  --apply, run. Appending to the end of a history line is exactly how people actually promote a dry run.

Recommendation for runner scripts: standardize on the trailing --apply flag, with three attached rules that matter more than the choice itself:
- Dry-run is always the default; --apply is the only way to write.
- The script rejects unknown argv with a non-zero exit (typo → abort, never silent dry-run).
- Dry-run output states explicitly that nothing was written and echoes the exact command to apply.

Recommendation for Rake tasks: inline APPLY=1 prefix, same semantics — with the one hygiene rule that it's always written inline on the command, never exported, so the lingering-arm hazard can't materialize.
```

I'm working on a PR to enforce this in all future Claude sessions.

script/backfill_mycoportal_export_links.rb
```diff
@@ -0,0 +1,443 @@
...
+    ExternalLink.where(target_type: target_type, target_id: ids,
+                       external_site: @site, relationship: :export).
+      pluck(:target_id, :id).to_h
+  end
+
+  def create_export_link(target_type:, target_id:, external_id: nil,
+                         external_created_on: nil)
+    if @apply
+      begin
+        ExternalLink.create!(user: @admin, target_type: target_type,
```
mo-nathan, line 343: This appears to be and N+1 issue since it is called inside a loop.  Would it be better/faster to do this in batches?

script/transfer_mycoportal_export_links.rb
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+#  USAGE::
+#
+#    # Phase 1 (dev box): after running
+#    # backfill_mycoportal_export_links.rb with APPLY=1 locally, export
```
mo-nathan: Using an `--apply` flag would be better than `APPLY=1`.

script/transfer_mycoportal_export_links.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  def apply_one(row, existing)
+    key = [row[:target_type], row[:target_id]]
+    return @stats[:already_present] += 1 if existing.include?(key)
+
+    create_link(row)
+  end
+
+  def create_link(row)
+    ExternalLink.create!(
```
mo-nathan: Again this appears to be N+1-ish which will make this quite slow.  As long any validation checks are checked on the data before it all gets inserted, I expect this would be a lot faster with an insert_all.  As long as the data in the local data is reasonably close to current (< half an hour), the chances of real issue seem extremely low especially if the time saved is creater than the age of the database + the faster update.

script/backfill_mycoportal_export_links.rb
JoeCohen: The lookups are already batched — `process_occurrence_batch`/`process_image_batch` fetch `known_ids` and `existing_links` for the whole `BATCH` (2000 rows) in one query each before the per-row loop runs, so it's not a read-side N+1. Only the write (`ExternalLink.create!`) is per-row.

That's intentional here: this script runs against the local dev DB (not production), and the per-row `create!` lets an individual invalid record be caught, logged with its `target_type`/`target_id` + the validation error, and skipped without aborting the rest of the batch — that's what feeds the `invalid` stat in the summary. A bulk `insert_all` would need validation done out-of-band first to keep that per-row error isolation, which felt like more machinery than a one-time/reconciliation backfill over a small "missing" subset needs.

Happy to revisit if a live run's missing-count turns out large enough that write throughput actually matters here.

script/transfer_mycoportal_export_links.rb
JoeCohen: Same batching as the backfill script — `existing_keys` looks up already-present rows per `target_type` per `BATCH` (2000), so only `create_link`'s insert is per-row, not the lookup.

Agreed this is the one place it's worth fixing, since keeping production load low is this script's whole reason for existing (the export/apply split was added specifically so production never runs the DwC-A parse — see the script header). Your point about pre-validating and using `insert_all` for the write side is on target for that goal. I'll leave `create_link` as-is for now per this task's scope (comment-only, no code changes), but a follow-up switching it to a validated `insert_all` batch — keeping the per-row `already_present`/`invalid` stats by validating each row's `target_type`/`target_id`/`url` length before the bulk insert — seems like the right next step given the production-load motivation.

JoeCohen: Fixed. `apply_batch` now builds an `ExternalLink` per candidate row, validates all of them in memory (`#valid?`), and writes the valid ones with a single `insert_all` per batch instead of one `create!` per row.

To keep the pre-validation itself batched (not just the write), `preload_targets` loads the real target records for the whole batch (one query per `target_type`) and caches each row's target directly on the association before calling `#valid?` — otherwise the `validates :target, presence: true` check would trigger its own query per row, reintroducing the same problem on the read side. A row whose target doesn't exist (or has an unrecognized `target_type`) fails validation and is logged/counted as `invalid`, same as before.

script/transfer_mycoportal_export_links.rb
JoeCohen: Fixed — `backfill_mycoportal_export_links.rb` now takes `--apply` instead of the `APPLY=1` env var, matching `transfer_mycoportal_export_links.rb`'s style.

While in there I also applied the three rules you laid out, on both scripts:
- Dry run stays the default; `--apply` is the only way to write. (Already true for both, now expressed the same way.)
- `Options.parse` / `parse_argv` now raise on any leftover positional argument after `OptionParser#parse!` — previously a typo'd bare argument would just get silently ignored rather than aborting.
- The backfill script's dry-run summary now says explicitly "nothing was written" and prints the exact command to re-run with `--apply` (including `--dwca-dir` when it's non-default).

`transfer_mycoportal_export_links.rb`'s `--apply FILE` doesn't have a separate dry-run/preview mode to echo a command for — passing `--apply` there already is the explicit, opt-in write step, so rules 1 and 3 were satisfied by its existing `--export`/`--apply` split. Didn't add a preview mode for it since that wasn't part of what was flagged here; happy to add one if you'd like it.

JoeCohen:
I returned this to a draft PR because I found a bug. The script created 1 ExternalLink per Image, even if the Image had been exported to multiple MyCoPortal observations (i.e., from multiple MO Observations.).
IDK if that makes any real difference when we update MyCoPortal, but I would prefer that the ExternalLink's accurately reflect what's in MCP.

script/transfer_mycoportal_export_links.rb
```diff
@@ -0,0 +1,210 @@
...
+#    already_present. Idempotent and resumable -- re-running the same
+#    CSV is harmless. Lookups are batched (BATCH rows/query) and new
+#    rows are validated in memory, then written with a single insert_all
+#    per batch rather than one ExternalLink.create! per row (#4877
+#    review) -- keeps production DB round trips to a handful per BATCH
+#    instead of one per row.
+
+require "optparse"
+require "csv"
+
```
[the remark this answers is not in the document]
JoeCohen: Not needed here, for two independent reasons:

1. `Set` has been a Ruby core class since 3.2 (autoloaded, no `require` needed) — this repo runs Ruby 3.3+/3.4, so `Set.new` works in any context, not just under Rails.
2. Even disregarding that, this script only ever runs via `bin/rails runner`, which boots the full Rails env first — ActiveSupport itself uses `Set` extensively and requires it during boot, well before this script's top-level code executes.

It's also the existing convention in this codebase: `script/hash_inat_photos.rb`, `script/materialize_external_links.rb`, and `script/standalone_image_dhash.rb` all use `Set` without an explicit `require "set"` either. Adding it here would be harmless but inconsistent with the sibling scripts and not fixing anything actually broken — leaving as-is.

Issue #4565, opened 2026-06-18
Reconcile the iNat "Mushroom Observer URL" field (5005) — complete MO↔iNat correspondence set (~161k)


mo-nathan:
## Goal

Establish observation-level provenance for the **complete** set of MO↔iNat correspondences, using iNat's "Mushroom Observer URL" observation field (id **5005**) as the authoritative iNat-side signal — **161,166** iNat observations carry it. It has been written consistently by all three historical MO↔iNat mechanisms, so it is the single most complete signal of which iNat observations correspond to an MO observation.

## Background — three mechanisms, one field

1. **iNat → MO import** (~50k) — the current importer. MO creates the obs from iNat (`source_id` = iNaturalist, `external_id` = iNat obs id) and writes the MO URL into iNat field 5005 (`Inat::ObservationImporter#update_mushroom_observer_url_field`, `MO_URL_OBSERVATION_FIELD_ID = 5005`). Image provenance done in #4529.
2. **MO → iNat mirror** (~25k) — the one-time third-party transfer script. MO-native obs pushed to iNat; MO side carries `"Mirrored on iNaturalist as <url>"`. Tracked in #4530.
3. **iNat-team historical** (no longer maintained) — bulk back-references the iNat team created; findable only via an iNat field-5005 report.

The 161k is a **superset**, not a fourth disjoint bucket — it already contains the import and mirror populations.

## Findings — refined 2026-06-19 (field-5005 report `observations-750205.csv`, 161,166 rows)

Parsed the MO obs id from each URL, matched against MO, classified by MO-side signal. The parser was tightened to strip query strings (`?q=…`) and handle the legacy `/observer/show_observation/<id>` path — **unparseable dropped 3,188 → 119**.

| Pathway | iNat obs |
|---|---:|
| **historical** (iNat-team + manual) | 81,434 |
| **import** (#4529) | 50,987 |
| **mirror** (#4530) | 25,564 |
| `mo_missing` | 2,506 |
| `import_diff` | 556 |
| unparseable URL | 119 |

(MO-image totals per pathway from the first pass — historical ~226k, import 188,649, mirror 81,998 — are approximately unchanged; the net-new historical population is still ~226k images not covered by #4529/#4530.)

### The "historical" bucket is not one pathway

Sub-classifying it dissolves the assumption that everything non-import / non-mirror is the iNat-team bulk:

| Sub-category | Count | Signal |
|---|---:|---|
[...]

PR #4944, opened 2026-07-29
Derive API2 `external_link` url from `external_id`

AlanRockefeller:
# API2: derive `external_link` url from `external_id`

## Problem

`GET /api2/external_links` returns `"url": ""` for every link identified by
`external_id` rather than by the `url` column — which is every link
`script/materialize_external_links.rb` created. The target is in the database
and the web UI displays it correctly; only the API can't see it.

Observation 2799, external_link 272867:

```
GET /api2/external_links?id=272867&detail=high&format=json
→ {"id":272867,"url":"","observation_id":2799,
   "owner":{"id":0,"login_name":"admin"},
   "external_site":{"id":2,"name":"iNaturalist"}}
```

while the page renders the target fine:

```
GET /external_links/272867   (Turbo-Frame: external_link_frame_272867)
→ <a href="https://www.inaturalist.org/observations/4675274">
     2016-12-01: Copied by iNaturalist</a>
```

Confirmed in production:

```
+--------+-------------------------------------------+-------------+--------------+---------+
| id     | url                                       | external_id | relationship | user_id |
+--------+-------------------------------------------+-------------+--------------+---------+
|   4934 | https://www.inaturalist.org/.../61261786  | NULL        |            0 |     123 |
| 272867 | NULL                                      |     4675274 |            4 |       0 |
+--------+-------------------------------------------+-------------+--------------+---------+
```

## Cause

`_external_link.json.jbuilder` serialized the raw attribute:

```ruby
json.url(object.url.to_s)
```

but `ExternalLink` deliberately nils `url` whenever `external_id` is present —
the XOR invariant from #4565 — and provides `link_url` for reading it back:

```ruby
def normalize_external_id_and_url
  self.external_id = nil if external_id.blank?
  self.url = nil if external_id.present?
end

def link_url
  url.presence || external_site&.observation_url(external_id)
end
```

`ExternalSite#observation_url` is documented as "the single source of truth for
an ExternalLink's URL — links store only `external_id`, not the full URL"
(#4299). The serializer was the one reader not going through it.
`_external_link.xml.builder` had the same issue.

## Change

- `json.url(object.link_url.to_s)` / `xml_string(xml, :url, object.link_url)`
- also serialize `external_id`, so clients don't have to reverse-parse the URL
  to recover the site's own identifier
- two controller tests: one for the `external_id` form, one pinning the
  `url`-column form so existing behavior stays fixed

The tests build the record with `ExternalLink.create!` rather than adding a
fixture, deliberately: fixtures skip `before_validation`, so a fixture with
both `external_id` and `url` set (as `imported_inat_obs_inat_link` has) does
not reproduce production rows. Going through the model exercises the invariant
that causes the bug.

## Scope / follow-up

Not fixed here, but same root cause and probably worth its own issue:
`Query::ExternalLinks`'s `url_has` searches `ExternalLink[:url]` directly, so
URL search never matches an `external_id`-backed link:

```
?url=https://www.inaturalist.org/observations/61261786   → 1 result
?url=https://www.inaturalist.org/observations/4675274    → 0 results
```

Fixing that means matching a derived value in SQL, which is a bigger design
call than this serializer change — happy to open a separate issue.

## Impact

There are ~11,273 of these rows on my account alone. To an API consumer each
one announces "this observation is linked to iNaturalist" while providing no
way to learn which observation, and it can't safely be ignored either, since
it's indistinguishable from a real link whose target failed to serialize.

## Testing

Style-checked against `.rubocop.yml` (all lines ≤ 80, methods well under the
30-line limit), but I have no Ruby/Rails/MySQL environment set up, so **I have
not been able to run the suite** — the two new tests need a real run before
merge.

### DIFF
--- a/app/views/controllers/api2/_external_link.json.jbuilder
+++ b/app/views/controllers/api2/_external_link.json.jbuilder
@@ -2,7 +2,8 @@
 
 json.id(object.id)
 json.type("external_link")
-json.url(object.url.to_s)
+json.url(object.link_url.to_s)
+json.external_id(object.external_id)
 json.created_at(object.created_at.try(&:utc))
 json.updated_at(object.updated_at.try(&:utc))
 json.observation_id(object.observation_id)
--- a/app/views/controllers/api2/_external_link.xml.builder
+++ b/app/views/controllers/api2/_external_link.xml.builder
@@ -6,7 +6,8 @@ xml.tag!(
   url: object.show_url,
   type: "external_link"
 ) do
-  xml_string(xml, :url, object.url)
+  xml_string(xml, :url, object.link_url)
+  xml_string(xml, :external_id, object.external_id)
   xml_datetime(xml, :created_at, object.created_at)
   xml_datetime(xml, :updated_at, object.updated_at)
   xml_minimal_object(xml, :observation, :observation, object.observation_id)
--- a/test/controllers/api2_controller_test.rb
+++ b/test/controllers/api2_controller_test.rb
@@ -336,6 +336,45 @@ class API2ControllerTest < FunctionalTestCase
     assert_equal(expected, files)
   end
 
+  # A link identified by `external_id` stores no `url` -- the model's
+  # `normalize_external_id_and_url` deliberately drops it, and
+  # `ExternalSite#observation_url` is "the single source of truth for an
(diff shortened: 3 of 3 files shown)


mo-nathan (approved):
The change looks good to me.  No idea why Coveralls is grumpy, but it's very clear that the actual coverage numbers have not change compared to main.

PR #4968, opened 2026-08-01
Move the thumbnail map into `Show::Details` under the location info

### LINKED ISSUE #4967: Move map into the details
The map should be in the details subpanel on the show observation page just under the location information.  It should continue to obey the user permission.

mo-nathan:
## Summary

Fixes #4967. Inspired by an in-person conversation with Igor at NEMF.

Moves the thumbnail map out of its own "Map" panel (second row, right column) and into the "Observation Details" panel, rendered as a line directly under the location info — after the Collection Location / GPS lines, before "Who:". Per the issue discussion, the panel chrome is gone entirely: no "Map" heading and no hide-map icon; it's obviously a map.

The map continues to obey the user permission: it only renders for logged-in viewers with the `thumbnail_maps` preference on (the gate moved from `Show` into `Details`), and hidden GPS is still obfuscated via `public_lat`/`public_lng` inside the component. Logged-out viewers see no map, same as before.

Because `Show::Details` is shared, the map now also appears on the naming new/edit pages (confirmed as fine in the issue discussion). The `Inat::ObservationResyncer` broadcast re-renders `Details` with `user: nil`, so the map drops out of that re-render the same way the GPS line already does — consistent with the existing logged-out-fallback tradeoff there.

## Changes

- `Views::Controllers::Observations::Show::ThumbnailMapPanel` → `Views::Controllers::Observations::Show::Details::ThumbnailMap`: same map body (Stimulus `thumbnail-map` wiring, zoom buttons, pin, bounding box, coordinate projection), wrapped in a chrome-less `li#observation_thumbnail_map` instead of a `Panel`.
- `Show::Details` renders it between the GPS and "Who:" lines when `@user&.thumbnail_maps`.
- Removed the now-unreachable hide-map pieces: `Tab::Observation::HideThumbnailMap`, the `javascript#hide_thumbnail_map` action + route, the `session[:hide_thumbnail_maps]` setter, and the `show_observation_hide_map` / `show_observation_thumbnail_map_hidden` locale strings. The preference itself is unchanged and still editable on the account preferences page.

## Test plan

- [ ] Log in and visit any observation with coordinates (e.g. `/observations/657608`): the map appears in the details panel under the location lines; zoom buttons and click-through to the full map page still work; the old standalone "Map" panel is gone.
- [ ] Visit `/observations/657608/namings/new`: the details panel there includes the map too.
- [ ] Log out and view the same observation: no map.
- [ ] Turn off "View thumbnail map on Observation page?" in account preferences: no map.
- [ ] View an observation with hidden GPS as a non-owner: the pin is still at the obfuscated location.


JoeCohen (approved):
I like it.

app/views/controllers/observations/show/details.rb
```diff
@@ -50,9 +52,16 @@ def render_when_where_who
...
     ObservationFragment(type: :where_gps, obs: @obs, user: @user)
+    render_thumbnail_map if @user&.thumbnail_maps
     ObservationFragment(type: :who, obs: @obs, user: @user)
   end
 
+  def render_thumbnail_map
+    render(Views::Controllers::Observations::Show::Details::ThumbnailMap.new(
+             obs: @obs
+           ))
+  end
```
nimmolo, line 63: Seems like the `ThumbnailMap` Phlex view should be moved into `ObservationFragment` like the others above. 

This means adding another `type` to the dispatcher there. 

I think the "render if" logic should be in the component too. ("return unless")

nimmolo (changes requested):

app/views/controllers/observations/show/details.rb
nimmolo: On the "render if" part: fixed in fea3bb7b — `ThumbnailMap` now self-gates with `return unless @user&.thumbnail_maps`, matching how When/Where/WhereGps/Who each self-gate. Details calls it unconditionally now.

On the `ObservationFragment` part: there's no second caller, so moving it is not the way to go. It only renders here.

test/views/controllers/observations/show/details_test.rb
```diff
@@ -89,6 +89,38 @@ def test_renders_external_links_body_last_when_no_links_yet
...
+    assert_operator(map_index, :<, who_index)
+  end
+
+  def test_does_not_render_thumbnail_map_when_pref_off
+    @user.thumbnail_maps = false
+
+    html = render(panel_with(@obs))
+
+    assert_no_html(html, "#observation_thumbnail_map")
+  end
```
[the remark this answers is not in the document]
nimmolo: Fixed in 3f903659 — the test now forces coordinates on the fixture obs (it has no lat/lng by default, so the GPS fragment never rendered before) and matches exact class tokens instead of substring (`"obs-where"` was matching both the Where line and the GPS line's `obs-where-gps` class). It now asserts the full ordering: Where < GPS < map < Who.

nimmolo (approved):

Issue #4894, opened 2026-07-23
Rework lightbox caption from `data-sub-html` string attribute to a real, Turbo-updatable DOM element


nimmolo:
## Problem

The lightbox caption isn't a real DOM element — it's an HTML string stashed in the `data-sub-html` attribute of each thumbnail's theater button (`<a class="theater-btn" data-sub-html="...">`), read by the third-party `lightGallery` library (`app/javascript/controllers/lightgallery_controller.js`, v2.7.2, pinned via importmap) when the lightbox opens. That string is built server-side by `Components::ImageFragment::LightboxCaption`, captured to HTML, and threaded into the attribute.

This makes the caption fundamentally incompatible with Turbo Streams' model, which expects to target persistent, addressable DOM elements — not attribute strings on elements that may or may not currently be showing their contents.

### Concrete evidence from #4892

Two symptoms already hit in production code because of this:

1. **Reviewed-toggle updates.** `ObservationViewsController#lightbox_caption_stream` can't use a plain `turbo_stream.update`/`replace` — the caption isn't a live element to target. Instead there's a custom Turbo Stream action, `update_lightbox_caption`, registered by `lightgallery_controller.js`, that reaches into `theaterBtn.dataset.subHtml` directly and calls `gallery.refresh()` to force lightGallery to re-read it. This is a second, parallel update mechanism that exists only because the normal one doesn't apply here.

2. **Vote UI, worse.** `Components::ImageFragment::VoteInterface` needed a `context:` prop specifically to give its lightbox copy (`context: :lightbox`) a `lightbox_`-prefixed id, and `Images::VotesController#update`'s turbo-stream response updates *both* the in-page and lightbox-prefixed ids. That only works while the lightbox happens to be **open** — lightGallery has injected the caption's HTML into real DOM at that point, so the prefixed id exists to target. If the lightbox is **closed** when a vote happens elsewhere (or on any other slide), the stored `data-sub-html` string is never touched, so reopening the lightbox (or navigating to that slide) shows stale vote state. Copilot flagged this exact gap on the PR review. Fixing it the same way the reviewed-toggle flow does — refreshing `data-sub-html` after every vote too — is awkward here specifically: an image can belong to more than one observation, so `Images::VotesController` doesn't have a single lightbox instance's identity to refresh. This isn't a bug to patch, it's the architecture pushing back.

Both symptoms are one root cause: the caption isn't a real, standing DOM element, so nothing can point at it directly.

[...]

PR #4972, opened 2026-08-01
Redesign image vote UI: theme-independent meter, `ButtonGroup` links

### LINKED ISSUE #4971: Switch colors of voting slider when in the Amanita theme
Currently the slider background is white and the value is shown in light gray.  It would be better if the background was light gray and the value was in white like it used to be.

mo-nathan:
## Summary

Fixes #4971. Another tweak from the in-person conversation with Igor at NEMF.

In the Amanita theme the image vote meter ("voting slider") rendered as a white track with a light-gray value fill. Root cause: the #4894 contrast fix (d76ffa7e95) hardcoded `.progress.vote-meter` to `background-color: white` with a `$link-color` fill -- and Amanita's link color is `$light_tan`, which reads as light gray against white, so the value was nearly invisible.

Rather than reintroducing per-theme `$VOTE_METER_*` override variables (the original approach here, Amanita-only), the meter now uses colors that don't depend on any theme at all: a half-opacity `rgba($white, 0.5)` track and a solid `$white` fill, both composited against the vote section's dark overlay (`rgba(0,0,0,0.8)`) that already sits behind every theme's meter. No per-theme override needed, and no new theme variables -- every theme's meter now looks the same. The already-cast vote (`.image-vote`) is similarly inverted to black-on-white, so it reads as a distinct, inert pill rather than blending into the link-colored value buttons around it.

Also restructures the vote links themselves: they were plain text separated by literal `|` characters, now a real `Components::ButtonGroup` (`.vote-btn-group`) with CSS divider borders between members instead. The already-cast vote renders as a non-interactive `Button(tag: :span, class: "active")`, matching the pattern already used by `Views::Controllers::Interests::Index`'s filter pills. The clear-vote button is now an icon (glyphicon-remove) instead of literal `"(x)"` text, with `"Clear"` (`:clear.ti`) as its screen-reader-only accessible name.

While wiring up tooltips on the restructured vote buttons, found that `Button::CRUDBase#button_html_options` combines its own tooltip data attributes (`tooltip_target`/`placement`/`title`) with a caller-supplied `data:` hash via a plain `Hash#merge` -- which replaces the whole `data:` hash rather than combining it, silently dropping the Stimulus tooltip wiring on any `CRUDBase` button that also passes its own `data:` (as the vote buttons do, for `image_id`/`value`). Switched to `deep_merge`. The vote-clear/vote-value buttons also now pass `placement: "bottom"` for their tooltips.

## Test plan

- [x] Hover an observation thumbnail (or open the lightbox) on the default/BlackOnWhite theme: the vote meter shows a white value bar on a translucent-white/gray track; vote links show as a bordered strip, not pipe-separated text.
- [x] Switch to the Amanita theme: the meter and vote links look the same as BlackOnWhite -- no theme-specific difference anymore.
- [x] Cast a vote, then reload: your own vote renders as a black-on-white pill, non-interactive.
- [x] Hover the clear-vote (X icon) and value buttons: tooltips appear, positioned below the button.

### DIFF
--- a/.claude/rules/code_comments.md
+++ b/.claude/rules/code_comments.md
@@ -1,7 +1,7 @@
 # Code comments: explain *why* (only when unclear), one source of truth
 
-Two rules for every comment in this codebase — app code, tests, config,
-`Gemfile`, scripts, everywhere.
+Three rules for every comment in this codebase — app code, tests,
+config, `Gemfile`, scripts, everywhere.
 
 ## 1. Comment the *why*, and only when it isn't already clear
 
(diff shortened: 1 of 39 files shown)


nimmolo:
I would consider doing it in a theme-independent way, for one because this image overlay UI does seem theme-independent, but also because single-theme overrides like this are more trouble to maintain. The fewer the better. 


E.g. `$white` for the bar and one of the `$gray` variables for the bar background. Also need a way to highlight "the user's own vote", currently a lightened version of the theme's `$primary`. Maybe the inverse: the voted text is a `$gray-800`, because it's inactive. The other vote words are active links.

Issue #5017, opened 2026-08-07
Entering corrected coordinates: viewport jumps to the locality dropdown, and a pasted `lat, lng` pair is not split


mo-nathan:
Correcting bad coordinates is a routine task right now — roughly 230 NEMF 2026 observations carry the foray ID-table point (39.8416, -77.5443) that came off a photo of the field slip rather than the collection site. #5016 makes the Geolocation checkbox actually release the locality autocompleter, which removes one obstacle, but two more remain. They are independent fixes and could be split; they are filed together because they obstruct the same task.

## 1. The page scrolls back up to the locality dropdown while typing coordinates

Typing into the latitude or longitude field jumps the viewport back up to the locality field, repeatedly, mid-entry.

The path: each keystroke runs `map#bufferInputs` → `sendPointChanged`, which fires the autocompleter's `swap` with the new point. `swap` bails early when nothing changed —

```js
if (type === this.TYPE && !paramsChanged) { return; }
```

— but `paramsChanged` is true on *every keystroke*, because each character changes the lat/lng being sent. So the full swap re-runs per character, ending in `constrainedSelectionUI`, which for the constrained type does:

```js
this.focused = true; // so it will draw the pulldown immediately
this.refreshPrimer(); // directly refresh the primer w/request_params
```

(`app/javascript/controllers/autocompleter/map_integration_mixin.js:152`)

Drawing the pulldown on the locality field — which sits *above* the coordinate inputs — is what pulls the viewport up. The 1000 ms debounce in `sendPointChanged` limits how often, but does not prevent it while typing at a normal pace.

Worth confirming in the browser before fixing, since the reasoning above is read off the code rather than observed in a profiler. Plausible fixes: don't redraw the pulldown when the autocompleter input is not the focused element, or hold the swap until the coordinate inputs lose focus rather than debouncing per keystroke.

## 2. Pasting a coordinate pair into the latitude field should split it

iNat renders coordinates as a single string, e.g. `40.0028333333, -77.3633033333`. Getting that into MO means splitting it by hand into two fields, for every observation.

The parsing capability is already present and already imported — `geo-coordinates-parser`'s `convert` is used in `validateLatLngInputs`, and it handles a combined pair:

```js
let coords = convert(origLat + " " + origLng)
```

What blocks the paste case is the guard at the top of that same method:

```js
if (!this.hasLatInputTarget || !this.hasLngInputTarget ||
[...]

Issue #5014, opened 2026-08-07
Coordinate helpers treat `0` as absent, so the equator and prime meridian are rejected


mo-nathan:
`GeocodeController`'s coordinate helpers treat `0` as absent, so a point on the equator or the prime meridian is rejected as if it had no coordinates:

```js
// validateLatLngInputs
if (!lat || !lng) return false

// sendPointChanged
if (lat && lng) { …swap the autocompleter… }
```

The range check immediately below the first one is correct (`lat > 90 || lat < -90 || lng > 180 || lng < -180`) — it is only the falsy test that is wrong. `0` is a valid coordinate; `null`, `""` and `NaN` are not.

Copilot raised this on #5013 and suggested `Number.isFinite(lat)` / `Number.isFinite(lng)`, which looks right.

## Priority: low, because nothing legitimate hits it today

Checked against production data:

| | count |
|---|---|
| lat = 0 **and** lng = 0 (Null Island) | 90 |
| lat = 0 with a real lng | 22 |
| lng = 0 with a real lat | 22 |

All 44 of the single-zero records are corrupt rather than legitimate — one coordinate was lost, and the location name proves it:

| observation | stored | location says | |
|---|---|---|---|
| 110709 | lat 0, lng −123.22 | Cypress Provincial Park, West Vancouver BC | lat should be ~49.4 |
| 276369 | lat 0, lng 19.6875 | Howard Co., Maryland | lng should be ~−76.9 |
| 157522 | lat 16.99, lng 0 | El Remate, Guatemala | lng should be ~−89.7 |
| 263059 | lat 52.0, lng 0 | Barneveld, Netherlands | lng should be ~5.6 |
| 234280 | lat 0.6756, lng 0 | Yasuní National Park, Ecuador | lng should be ~−76 |

So there is no current victim. The last row is worth noting though: a genuine observation sits at latitude 0.68 in Ecuador, so near-equator latitudes are real MO data — an exact 0 is only a matter of time, and Greenwich/Ghana/eastern Spain make lng = 0 equally plausible.

## Scope

`validateLatLngInputs` has six call sites, so this is wider than the one path that surfaced it. In particular:

```js
if (center = this.validateLatLngInputs(false)) {
```

is an assignment-in-condition that currently does **not** take its branch for a zero coordinate. Fixing the falsy test changes that — probably correctly (a point on the equator should draw a marker), but it touches map drawing and rectangle handling, which is why it was kept out of #5013.

[...]

PR #5076, opened 2026-08-14
Manual geocoding entry - Fix `#5017` viewport jump / paste-split and `#5014` zero-coordinate falsy bug

nimmolo:
## Summary

Fixes four bugs in the observation form's coordinate-entry UX, all obstructing the same correction workflow described in issue #5017 (which also references #5014).

- **Viewport jumps to the locality dropdown while typing coordinates** (#5017 item 1). Root cause: `constrainedSelectionUI()` force-sets `this.focused = true` on the locality autocompleter to redraw its "localities containing this point" dropdown as coordinates change, and the redraw path (`drawPulldown()` and two related spots) unconditionally called `.focus()` on that field afterward, stealing keyboard focus and scrolling the page to it. Fixed by adding a `silentRedraw` flag: `constrainedSelectionUI()` sets it before a programmatic redraw, the three `.focus()` call sites skip stealing focus while it's set, and it's cleared by the next real interaction with that field (so a redraw triggered by the user actually typing there still reclaims focus as before). This also covers the `location_containing` → `location_google` Google-geocode fallback (`refreshGooglePrimer`, which force-sets focus "even if input lost focus" for its own reasons) since the flag persists through that whole cascade.
- **Pasting a `lat, lng` pair doesn't split across fields** (#5017 item 2). iNat exports coordinates as a single string (e.g. `40.0028333333, -77.3633033333`). `GeocodeController#validateLatLngInputs` already parses combined pairs via `geo-coordinates-parser`, but its guard clause bailed whenever either field was empty, before that parsing ever ran. Added `GeocodeController#splitPastedCoordinates`, wired to a `paste` action on both lat/lng fields, which parses the pasted text directly and splits it across both fields when it's a full pair; an ordinary single-value paste isn't intercepted.
- **A latitude or longitude of exactly 0 is treated as absent** (#5014). Several falsy checks (`!lat || !lng`, `lat && lng`, `center?.lat`) rejected the equator/prime meridian as if no coordinate were given. Replaced with `Number.isFinite` checks in `GeocodeController` and in the autocompleter's own `swap()` (`hasNewParams`), which had the identical bug for a different call path.
- **Re-entering coordinates after a prior match silently does nothing** (found via manual testing during review). Once a search auto-fills the locality field with a match, correcting to a second point left the field stuck on the old name. `refreshPrimer()`'s "is my new search string just a longer version of the last one" caching heuristic is a text-search optimization that doesn't apply to `location_containing`/`location_google` (their results are keyed off `request_params`, i.e. lat/lng, not the search string — the server ignores the string entirely for these). It ran anyway, and the leftover locality text from the prior match looked like a valid "refinement" of itself once truncated by an intermediate search, so the fetch for the new point was silently skipped. Now skipped for `ACT_LIKE_SELECT` types, which already have a correct `request_params`-based dedup check.

## Test plan

- [x] `bin/rails test test/system/observation_form_system_test.rb` -- full file, including 5 new system tests covering all four fixes, plus regression-checked each fix by reverting/re-confirming failure before restoring
- [x] `bundle exec rubocop` clean on all touched Ruby files
- [x] Verified all fixes directly in a running dev server (Chrome), including the pre-fix broken behavior for comparison

### DIFF
--- a/app/javascript/controllers/autocompleter/base_controller.js
+++ b/app/javascript/controllers/autocompleter/base_controller.js
@@ -46,6 +46,14 @@ const INTERNAL_OPTS = {
   ROW_HEIGHT: null,
   SCROLLBAR_WIDTH: null,
   focused: false,
+  // Set by constrainedSelectionUI() right before a redraw it triggers
+  // programmatically (e.g. the map controller refreshing the locality
+  // dropdown while the user is typing coordinates elsewhere), so that
+  // redraw doesn't also steal keyboard focus. Cleared by the next real
+  // interaction with this field (ourFocus/ourChange), so a genuine
+  // redraw triggered by the user actually typing here still reclaims
+  // focus as before.
+  silentRedraw: false,
   menu_up: false,
   old_value: null,
   stored_id: 0,
@@ -213,8 +221,11 @@ export default class BaseAutocompleterController extends Controller {
     }
     if (type == undefined) { return; }
 
-    // If same type but different lat/lng params, still update the location
-    const hasNewParams = detail?.request_params?.lat && detail?.request_params?.lng;
+    // If same type but different lat/lng params, still update the location.
+    // 0 is a valid coordinate (equator/prime meridian) -- Number.isFinite,
+    // not truthiness.
+    const hasNewParams = Number.isFinite(detail?.request_params?.lat) &&
+      Number.isFinite(detail?.request_params?.lng);
     const paramsChanged = hasNewParams &&
       (detail.request_params.lat !== this.request_params?.lat ||
        detail.request_params.lng !== this.request_params?.lng);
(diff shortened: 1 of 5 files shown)


JoeCohen:
Loops (instead of freezing) if I enter a non-existent Locality. (Maybe this should branch off #5055?)
To replicate:

1. http://localhost:3000/observations/new
2. Clear the Locality field, enter an nonexistent Location (I used `Null Island`)
3. Create 
Result (expected): Flash Error "Unknown country ‘Null Island’
Click ‘Create’ to use this location name or edit the text below to correct any typo."
4. Create
Result (unexpected): Page redisplays with same Flash.

nimmolo:
Confirmed unrelated to this PR — `validators.rb` has no diff here, and it reproduces identically on `main`. Root cause: `ObservationsController::Validators#validate_place_name` never passed `approved:` to `Location.dubious_reasons_for`, so the resubmit-to-confirm flow could never pass. Fixed in #5079.

JoeCohen:
Another preexisting bug, which I just wrote up: Kiribati for Null Island #5080

JoeCohen (approved):
Approve, but note another preexisting bug #5080.
(Might be the ultimate edge case.)

PR #5104, opened 2026-08-17
Make observation show page's "About this Taxon" panel Turbo lazy-loaded

### LINKED ISSUE #5093: Make Show Observation "About this Taxon" Turbo lazy-loaded by a button or link
@JoeCohen @mo-nathan 

Currently the obs title links to the name page, and so does the "About this Taxon" panel.
The latter provides other links, like query results and external reference links. It takes awhile to load.

The idea is to make the panel collapsible, and collapsed by default, and make the panel's collapse toggle _fetch the panel content separately via Turbo_, saving some queries on the obs show page. I haven't audited recently but i think it's more than half the page's db time just to make this panel.

Part of #4049.

nimmolo:
Fixes #5093.

## Summary

The "About this Taxon" panel on the observation show page on this branch is skipped on page load and collapsed by default. If a user opens the panel, its content is fetched lazily via Turbo, instead of being eager-rendered on every page load. This addresses both the info-overload/visual-clutter goal from the issue and the DB cost the audit found (~100ms / 8 queries per load, previously always paid whether or not the panel was ever opened).

## Approach

- `Views::Controllers::Observations::Show::NameInfoPanel` now renders a collapsed panel shell with an empty `turbo_frame_tag` placeholder. The heading's collapse toggle carries `fallback_href:` (so `Link::CollapseToggle` sets `data-target` explicitly, per `Components::Panel#collapse_toggle_data`'s comment) plus `data-turbo-frame:`, mirroring the existing `Observations::ExternalLinksController`/accordion-badge pattern -- not Panel's own `collapsible:` auto-toggle, which deliberately omits `data-target` for id-based targets and would block Turbo from ever firing.
- New `Observations::NameInfoPanelsController#show` renders the panel's actual content (moved unchanged into `Views::Controllers::Observations::NameInfoPanels::Show`) into the matching Turbo Frame when the request carries a `Turbo-Frame` header; a direct non-frame visit redirects to the observation page instead of rendering a bare fragment.
- `Observation.show_includes_tree` / `not_logged_in_show_includes` drop the `descriptions`/`interests`/`description` keys from the `name:` subtree -- that preload moved to the new controller's own lightweight find, since the panel is now its only consumer. `name: { synonym: :names }` stays, since `ConsensusNameLink` (the page title, shown to every viewer, not just when the panel is open) still needs it -- confirmed by running the full observation-show test suite, which caught this via `ActiveRecord::StrictLoadingViolationError` before the fix.

## Test plan

- [x] `bin/rails test test/controllers/observations/name_info_panels_controller_test.rb` -- Turbo-Frame request renders content; non-frame request redirects; login required; not-found handling
- [x] `bin/rails test test/views/controllers/observations/name_info_panels/show_test.rb` -- fragment wraps content in the matching frame, renders both On MO / On the web link groups
- [x] `bin/rails test test/views/controllers/observations/show/name_info_panel_test.rb` -- placeholder renders collapsed with an empty frame, toggle link targets the fetch route
- [x] `bin/rails test test/controllers/observations_controller_show_test.rb` -- full obs-show suite green (also updated one assertion whose content moved into the lazy frame)
- [x] Verified in a running dev server: panel renders collapsed on page load; clicking the toggle expands it and the real content (On MO / On the web links) loads via the Turbo Frame fetch

### DIFF
--- a/app/controllers/observations/name_info_panels_controller.rb
+++ b/app/controllers/observations/name_info_panels_controller.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Observations
+  # Turbo Frame content for the observation show page's "About this
+  # Taxon" panel (#5093). The panel starts collapsed with an empty
+  # placeholder frame; expanding it sends a real GET carrying a
+  # `Turbo-Frame` request header, fetched here instead of eager-loading
+  # the name subtree (synonyms/descriptions/interests) on every
+  # observation page view. A direct (non-frame) visit falls back to
+  # redirecting to the observation instead of rendering a bare fragment.
+  class NameInfoPanelsController < ApplicationController
+    before_action :login_required
+    before_action :find_observation!
+
+    # `Descriptions::List#visible?` reads each description's `.user`,
+    # so `descriptions: :user` avoids N+1 per description in the panel.
+    # This subtree used to live in `Observation.show_includes_tree` --
+    # moved here since this frame is now its only consumer.
+    NAME_INCLUDES = {
+      name: [{ synonym: :names }, { descriptions: :user },
+             :interests, :description]
+    }.freeze
+
+    def show
+      if request.headers["Turbo-Frame"]
+        render(Views::Controllers::Observations::NameInfoPanels::Show.new(
+                 obs: @observation, user: @user
+               ), layout: false)
+      else
+        redirect_to(permanent_observation_path(@observation))
+      end
+    end
+
+    private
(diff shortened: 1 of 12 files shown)


JoeCohen:
Would it be possible to rework the panel and its caption so that the link to the Name page is always visible?
That link is probably the one I use the most on the show_observation page.

mo-nathan:
I don't think this is enough of win that it's worth doing.

PR #5106, opened 2026-08-17
Create obs slowness: Add missing `observations.user_id` index; fix project-checkbox N+1

### LINKED ISSUE #5103: Create-observation page: missing `observations.user_id` index + project-checkbox N+1
Two distinct bugs on the create-observation page (`/observations/new`), both found in the same audit. Tracked together since fixing them is one PR.

## Audit setup

Reported slow by a user with many projects/species-lists. Compared two dev-DB users on `/observations/new`, 2-3 warm requests each:

| Section | `nathan` (28 projects, 45 species-lists) | `barky` (1 project, 0 lists) | Scales with project/list count? |
|---|---|---|---|
| `init_project_vars_for_new` | ~34-46ms, 3 queries | ~30ms, 3 queries | No |
| `init_list_vars` | ~50-56ms, 3 queries | ~52ms, 3 queries | No |
| `defaults_from_last_observation_created` | ~585-591ms, 6 queries | ~577-602ms, 6 queries | **No — identical for both** |
| `render_new_view` (Phlex form render) | ~546-746ms, 58-63 queries | ~56-76ms, 6 queries | **Yes — ~10x, tracks project+list count almost exactly** |

That split is what separates the two bugs below: one is a fixed cost paid by every user regardless of profile, the other is real N+1 that only shows up for users with many projects.

## Bug 1: `observations` has no index on `user_id`

`observations` has no index on `user_id` at all. Every "this user's most recent/own observations" query -- `Observation.recent_by_user`, `@user.observations`, etc. -- does a full table scan.

```
$ bin/rails runner 'puts ActiveRecord::Base.connection.indexes("observations").map { |i| "#{i.name}: #{i.columns}" }'
needs_naming_index: ["needs_naming"]
index_observations_on_name_id: ["name_id"]
index_observations_on_location_id: ["location_id"]
index_observations_on_occurrence_id: ["occurrence_id"]
index_observations_on_collector_user_id: ["collector_user_id"]
index_observations_on_inat_import_id: ["inat_import_id"]
index_observations_on_reflected_at: ["reflected_at"]
```

`collector_user_id` is indexed; `user_id` is not.

`defaults_from_last_observation_created` costing a near-constant ~580-600ms regardless of which user loads the page -- identical for both the 28-project user and the 1-project user -- is what pointed at a fixed, universal cost rather than a many-projects problem. Isolated it to a single line: `Observation.recent_by_user(@user).last` (`app/models/observation/scopes.rb`):

```ruby
scope :recent_by_user, lambda { |user|
  includes(:location, :projects, :species_lists).
    where(user_id: user.id).reorder(:created_at)
}
```

```
$ bin/rails runner '
  u = User.find_by(login: "barky")
  rel = Observation.where(user_id: u.id).order(created_at: :asc).limit(1)
  puts rel.explain.inspect
'
+----+-------------+--------------+------------+------+---------------+------+---------+------+--------+----------+-----------------------------+
| id | select_type | table        | partitions | type | possible_keys | key  | key_len | ref  | rows   | filtered | Extra                       |
+----+-------------+--------------+------------+------+---------------+------+---------+------+--------+----------+-----------------------------+
|  1 | SIMPLE      | observations | NULL       | ALL  | NULL          | NULL | NULL    | NULL | 544048 |     10.0 | Using where; Using filesort |
+----+-------------+--------------+------------+------+---------------+------+---------+------+--------+----------+-----------------------------+
```

`type: ALL`, `possible_keys: NULL`, `rows: 544048` -- a full scan of every row in `observations` (544k in this snapshot) plus a filesort for the `ORDER BY`, on every single load of the create-observation page, for every user, regardless of how many observations that user has.

**Ask**: add an index on `observations.user_id` (a composite `(user_id, created_at)` would additionally make `recent_by_user`'s `ORDER BY created_at` an index-order scan instead of a filesort). `user_id` is read constantly elsewhere in the app (`@user.observations`, various `by_user`/`recent_by_user` scopes, permission checks), so this isn't narrowly scoped to the one call site that surfaced it.

## Bug 2: N+1 in project-checkbox form rendering

`render_new_view`'s cost tracks project+list count almost exactly (~10x between the two test users, matching their ~10x difference in combined project+list count), and query count scales the same way (58-63 vs 6).

Traced to `Views::Controllers::Observations::Form::Projects#render_project_checkbox` (`app/views/controllers/observations/form/projects.rb`):

```ruby
def render_project_checkbox(project)
  @form.checkbox_field(
    :project_ids,
    label: false,
    disabled: !project.user_can_change_membership?(@observation, @user)
  ) do |cb|
    ...
```

`user_can_change_membership?` → `Project#member?` (`app/models/project.rb`):

```ruby
def member?(user)
  user && user_group.users.member?(user)
end
```

`user_group.users` is a fresh query per project unless already loaded. `ObservationsController::SharedFormMethods#init_project_vars` only preloads `include: :user_group`, not the deeper `user_group.users`, so every project checkbox fires its own membership query -- one extra query per project the user belongs to. Species-list checkboxes don't have an equivalent per-item permission recheck, which is why this scales with project count specifically, not list count.

**Ask**: deepen the preload in `init_project_vars` to `include: { user_group: :users }` (or equivalent) so `member?`'s `user_group.users.member?(user)` hits the already-loaded association instead of issuing a new query per checkbox.

## Related

Found while auditing why `/observations/new` is slow (separate from #5093/#5101/#5102, which were about the observation *show* page).

nimmolo:
Fixes #5103.

## Summary

Two independent performance bugs on the create-observation page (`/observations/new`), fixed together since #5103 tracks both for one PR.

### Bug 1: missing `observations.user_id` index

`observations` had no index on `user_id` at all -- every "this user's own/most recent observations" query (`Observation.recent_by_user`, `@user.observations`, various `by_user` scopes) did a full table scan. Added a composite `(user_id, created_at)` index so `recent_by_user`'s `ORDER BY created_at` is satisfied by index order instead of a filesort.

Before:
```
type: ALL, possible_keys: NULL, key: NULL, rows: 544048, Extra: Using where; Using filesort
```
After:
```
type: ref, key: index_observations_on_user_id_and_created_at, rows: 54 (barky's own count), Extra: (none)
```

### Bug 2: N+1 in project-checkbox form rendering

`ObservationsController::SharedFormMethods#init_project_vars` only preloaded `include: :user_group`, not the deeper `user_group.users` that `Project#member?` reads (`user_group.users.member?(user)`) for every project checkbox's disabled state. Deepened the preload to `include: { user_group: :users }` -- one extra query avoided per project the user belongs to, instead of one query per checkbox.

## Combined effect

`/observations/new` for a user with 28 projects / 45 species-lists (the case that motivated #5093's original audit):

| | Before | After |
|---|---|---|
| Total request | ~1325ms | ~352ms |
| Queries | 94 | 68 |

## Test plan

- [x] `bin/rails test test/controllers/observations_controller_project_list_test.rb` (project/list checkbox rendering, 14 tests)
- [x] `bin/rails test test/controllers/observations_controller_create_test.rb` (110 tests, all touch `init_project_vars` via the create form)
- [x] `bin/rails test test/controllers/observations_controller_update_test.rb` (48 tests, edit form shares the same preload)
- [x] Verified the index fix via `EXPLAIN` (full scan -> indexed lookup) and the N+1 fix via direct query counting (28 projects: ~28+ queries -> 4)
- [x] Verified end-to-end in a running dev server, logged in as a 28-project/45-list user: `/observations/new` dropped from ~1325ms/94 queries to ~352ms/68 queries

### DIFF
--- a/app/controllers/observations_controller/shared_form_methods.rb
+++ b/app/controllers/observations_controller/shared_form_methods.rb
@@ -154,9 +154,13 @@ module ObservationsController::SharedFormMethods
     @accession_number = herb_params[:accession_number]
   end
 
+  # `user_group: :users` (not just `:user_group`) -- the projects form's
+  # per-checkbox `Project#user_can_change_membership?` reads `member?`,
+  # which is `user_group.users.member?(user)`. Without the deeper
+  # preload that's a fresh query per checkbox (#5103).
   def init_project_vars
     @projects = @user.projects_member(order: :title,
-                                      include: :user_group)
+                                      include: { user_group: :users })
   end
 
   # Failure-reload path: capture the user's just-submitted project_ids
--- a/db/migrate/20260817133242_add_user_id_index_to_observations.rb
+++ b/db/migrate/20260817133242_add_user_id_index_to_observations.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddUserIDIndexToObservations < ActiveRecord::Migration[7.2]
+  # `observations` had no index on `user_id` at all -- every "this
+  # user's own/most recent observations" query (Observation.
+  # recent_by_user, @user.observations, various by_user scopes) was a
+  # full table scan. Composite with `created_at` so recent_by_user's
+  # `ORDER BY created_at` is an index-order scan, not a filesort.
+  def change
+    add_index(:observations, [:user_id, :created_at])
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2026_08_09_175755) do
+ActiveRecord::Schema[7.2].define(version: 2026_08_17_133242) do
   create_table "api_keys", id: :integer, charset: "utf8mb3", force: :cascade do |t|
     t.datetime "created_at", precision: nil
     t.datetime "last_used", precision: nil
@@ -643,6 +643,7 @@ ActiveRecord::Schema[7.2].define(version: 2026_08_09_175755) do
     t.index ["needs_naming"], name: "needs_naming_index"
     t.index ["occurrence_id"], name: "index_observations_on_occurrence_id"
     t.index ["reflected_at"], name: "index_observations_on_reflected_at"
+    t.index ["user_id", "created_at"], name: "index_observations_on_user_id_and_created_at"
   end
 
   create_table "occurrences", id: :integer, charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|


mo-nathan (approved):
Looks good.  Nice speed up!

Issue #5094, opened 2026-08-16
Obs Show - Button UI to add obs directly to a Project or FieldSlip


nimmolo:
Two PRs.

## Projects
There's currently an "Add to Observation List" link in the title bar of the `SpeciesListsPanel` on `Observations::Show`.

As the `ProjectsPanel` comment block says:
```ruby
# "Projects" panel on the observation show page. Read-only list of the
# projects this observation belongs to. Unlike `SpeciesListsPanel`,
# there's no add/remove-membership affordance here -- observations
# aren't added to or removed from a project from this page -- so the
# panel simply hides when the observation belongs to no projects.
```

This should be redone to work the same as the `SpeciesListsPanel`.

## FieldSlips
`Observation::Show::Details` has a `render_field_slip` method, gated at the call site.

This should be refactored to render an "Attach to Field Slip" modal link (intro "Field Slip:", plus `.inline-icon-link` with the "+" icon and an "Attach" tooltip) if there's no field slip, which pops up a `.modal-sm` form for adding a field slip code, with the usual validation. (Prefix, iirc). This may require a new controller endpoint for attaching an obs to a FieldSlip.

If there's a field slip already attached to the obs, print the link to the slip as currently. 

@JoeCohen @mo-nathan please lmk if i've got this right.

nimmolo:
Done via #5123 (Project UI) and #5124 (Field Slip UI).

PR #5123, opened 2026-08-18
Add UI to add/remove an observation to/from a `Project`

nimmolo:
## Summary

Part of #5094 — the Projects half only (the issue asks for two PRs; the FieldSlip half is separate).

> [!NOTE] 
> Run `dev_setup_macos --icons-only` to see the new icon on the Add button. 
> This needs to be updated on production during `deploy`, with `deploy --icons`, or afterwards with `deploy --icons-only`.

Redoes the observation show page's `ProjectsPanel` to work the same way as `SpeciesListsPanel`:

- When the observation belongs to no projects, the panel's whole heading becomes an "Add to a Project" link — shown only if the user is a member of at least one project, otherwise the panel doesn't render at all.
- When it belongs to some, the heading is the bare "Projects" title with an icon-only "manage projects" link, and each project in the list gets an inline REMOVE button when the user is a member of that project.

New `Observations::ProjectsController#edit`/`#update` mirrors `Observations::SpeciesListsController` exactly (same route shape, same `find_or_goto_index` + `params[:commit]` dispatch pattern), using the existing `Project#add_observation`/`#remove_observation` model methods and `Project#user_can_change_membership?` (== `member?`) for the permission gate.

## Constraint handling

Unlike species lists, `Project` membership has constraint logic — date range, bounding box, target name/location (`Project#violates_constraints?`). Adding an observation to a project that violates one of these is allowed, not blocked, matching the full observation form's own behavior (`Views::Controllers::Observations::Form::Projects`) — the constraint is a warning, not a gate. `Views::Controllers::Projects::ListItem` (extended with optional `observation:`/`remove:`/`add:`/`violation_kinds:` props, reused from the projects index page) shows an inline warning next to a project that would be violated, reusing `Project#violation_kinds_for` and the same `form_observations_projects_kind_*` translation keys the create form already uses.

## Other notes

- The panel's REMOVE-button membership check reuses `@user.projects_member`, already loaded for the manage-link decision above it, instead of eager-loading `projects.user_group` on `Observation.show_includes_tree` (which would load every rendered project's full member list) or querying per project in the render loop.
- Removing an observation reports the full count of siblings removed (via `:removed_from_project_with_siblings`, matching `ObservationsController::SharedFormMethods#flash_project_removal`'s established pattern), since `Project#remove_observation` removes every sibling sharing the same `Occurrence`, not just the clicked one.
- An invalid `params[:commit]` value routes through `:runtime_invalid.t` rather than being interpolated directly into the flash message.
- `Observations::ProjectsController#set_project_ivars` preloads `location`/`target_names`/`target_locations`, so `Edit`'s per-project `violation_kinds_for` call isn't an N+1 across the "other projects" list.
- No sorter/`Query::Projects` on the new `Observations::Projects::Edit` page (unlike `SpeciesLists::Edit`, which has one) — scoped out as a nice-to-have, not part of the issue's core ask. Users typically belong to far fewer projects than observation lists.
- Reused `Tab::Observation::ListActions` (a single cancel-to-obs link) as-is for the new edit page's context nav — it's already fully generic, despite being named/commented for the species-lists page originally.
- No dedicated "manage projects" icon exists; reused `:manage_lists`, same as `Tab::Observation::ManageLists`.
- Navigating directly to the edit page for an observation with no project memberships renders an empty page rather than redirecting — same behavior as the equivalent `SpeciesLists::Edit` path.
- Adding/removing an observation that's already in/out of a project is a silent no-op that still flashes success, matching `Project#add_observation`'s existing return semantics and `SpeciesList`'s equivalent controller.

## Test plan

- [x] `bin/rails test test/controllers/observations/projects_controller_test.rb`
- [x] `bin/rails test test/views/controllers/observations/show/projects_panel_test.rb`
- [x] `bin/rails test test/views/controllers/projects/list_item_test.rb`
- [x] `bin/rails test test/classes/tab/observation/tabs_test.rb`
- [x] `bin/rails test test/models/project_test.rb`
- [x] `bin/rails test test/controllers/observations_controller_show_test.rb` (full file, confirms the panel's membership check doesn't regress existing project-panel rendering under `strict_loading`)
- [x] `bin/rails test test/style/no_queries_in_phlex_views_test.rb`

### DIFF
--- a/app/classes/tab/observation/manage_projects.rb
+++ b/app/classes/tab/observation/manage_projects.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# "Manage projects for this observation" link. Caller must guard on
+# `user.projects_member.any?` — the icon is reused from
+# `Tab::Observation::ManageLists`, since both are "manage membership"
+# affordances and no dedicated projects glyph exists.
+class Tab::Observation::ManageProjects < Tab::Base
+  def initialize(observation:, q_param: nil)
+    super()
+    @observation = observation
+    @q_param = q_param
+  end
+
+  def title
+    :show_observation_manage_projects.l
+  end
+
+  def path
+    with_q_param(edit_observation_projects_path(@observation.id), @q_param)
+  end
+
+  def html_options
+    { icon: :manage_lists }
+  end
+
+  def model
+    @observation
+  end
+end
--- a/app/controllers/observations/projects_controller.rb
+++ b/app/controllers/observations/projects_controller.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module Observations
+  # Add or remove one Observation from a Project, from the observation's
+  # own show page -- mirrors Observations::SpeciesListsController.
+  class ProjectsController < ApplicationController
+    before_action :login_required
+
+    # Form (table of post_button links) to let user add/remove one
+    # observation at a time from a project they're a member of.
+    def edit
+      return unless (@observation = find_observation!)
+
+      set_project_ivars
+      render_phlex_edit
+    end
+
+    def update
+      return unless (@project = find_project!) &&
+                    (@observation = find_observation!)
+      return unless permission_to_change_membership?
+
+      set_project_ivars
+      dispatch_commit
+    end
+
+    private
+
+    def permission_to_change_membership?
+      return true if @project.user_can_change_membership?(@observation, @user)
+
+      flash_error(:permission_denied.l)
+      redirect_to(project_path(@project.id))
+      false
+    end
+
+    def dispatch_commit
+      case params[:commit]
+      when "add"
+        add_observation_to_project(@project, @observation)
+      when "remove"
+        remove_observation_from_project(@project, @observation)
+      else
+        flash_error("Invalid mode: #{params[:commit].inspect}")
+        render_phlex_edit(
+          location: edit_observation_projects_path(id: @observation.id),
+          status: :unprocessable_content
+        )
+      end
+    end
+
+    def render_phlex_edit(**render_opts)
+      render(
+        Views::Controllers::Observations::Projects::Edit.new(
+          observation: @observation,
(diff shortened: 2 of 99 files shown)


app/classes/tab/observation/manage_projects.rb
```diff
@@ -0,0 +1,29 @@
...
+    @observation = observation
+    @q_param = q_param
+  end
+
+  def title
+    :show_observation_manage_projects.l
+  end
+
+  def path
+    with_q_param(edit_observation_projects_path(@observation.id), @q_param)
```
nimmolo, line 19: `with_q_param`???

app/controllers/observations/projects_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+      else
+        flash_error(:runtime_invalid.t(type: '"mode"', value: params[:commit]))
+        render_phlex_edit(
+          location: edit_observation_projects_path(id: @observation.id),
+          status: :unprocessable_content
+        )
+      end
+    end
+
+    def render_phlex_edit(**render_opts)
```
nimmolo: rename `render_phlex_edit` and ban this bullshit with a cop. Intern just finished a huge sweep renaming all of these consistently `render_edit_view` app-wide.

app/controllers/observations/projects_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+        **render_opts
+      )
+    end
+
+    def set_project_ivars
+      member_ids = @observation.project_ids.to_set
+      all_projects = @user.projects_member(order: :title)
+      @obs_projects, @other_projects =
+        all_projects.partition { |project| member_ids.include?(project.id) }
+    end
```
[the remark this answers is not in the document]
nimmolo: Fixed in 83fb030953 - added :user to set_project_ivars' include list (ListItem's meta row reads project.user per row).

PR #5124, opened 2026-08-18
Add "Attach to Field Slip" UI to the observation show page

nimmolo:
## Summary

Part of #5094 — the FieldSlip half (the Projects half is a separate PR, #5123).

`Details#render_field_slip` (observation show page) previously rendered nothing at all when the observation had no field slip. It now shows an icon-only "Attach" link (`.inline-icon-link`, `+` icon) next to the "Field Slip:" label, gated the same way `Observations::FieldSlipsController` itself gates the action: `in_admin_mode? || obs.can_edit?(user)`.

The link goes to a full-page round trip — same shape as `Observations::SpeciesListsController` / the Projects half's `Observations::ProjectsController` — rather than a modal, per discussion on this PR (the issue's ".modal-sm" detail wasn't load-bearing; the existing "Add to an Observation List" link is also a full page, not a modal).

## Design

New `Observations::FieldSlipsController#edit`/`#update`:

- `edit` renders `Observations::FieldSlips::Form` — a single code field, backed by `FormObject::FieldSlipAttach` purely for Superform's `<form>` tag machinery. The submitted param is the bare top-level `field_code`, not a namespaced value — that's deliberate, since it's what the shared `validate_field_slip`/`update_field_slip` pair (`ObservationsController::FieldSlips`) already reads.
- `update` calls that same `validate_field_slip`/`update_field_slip` pair, mirroring exactly how `ObservationsController#update`'s own field-slip handling calls them (see `update_field_slip_or_flag_error` in `observations_controller/edit_and_update.rb`) — including the double-check pattern (`validate_field_slip` first, `update_field_slip`'s `:invalid`/`:too_many` branches again after) for the same rare race the original comment documents (slip changes between validation and application).
- Also includes `FieldSlipProjectJoinable` — `assign_field_slip`'s auto-join-project side effect depends on `join_field_slip_project`, defined there.

No new business logic was written for the actual attach — this PR only adds a new entry point into logic that already exists and is already exercised by the full observation create/update forms.

## Behavior

- The attach form submits via Superform's standard PATCH spoofing (a `persisted?` override, matching `FormObject::AdminSession`), not a hand-rolled `method:` attribute — consistent with every other Superform-backed edit form in the app. A new cop, `MO/NoHandRolledFormTag`, keeps future Superform forms on this path; it's cleared against the 4 other pre-existing legitimate `form_tag` overrides elsewhere in the codebase with targeted, justified `rubocop:disable`s (each needs a runtime-computed attribute merge or a runtime GET/PATCH branch the default `form_tag` can't express).
- Visiting or submitting the attach page when the observation already has a field slip redirects back to the observation with a warning, rather than detaching the existing slip.
- The attach link's selector class merges with (rather than replaces) its Tab-derived class, via `Components::InlineLinkBlock.item_class` — the same merge helper `InlineCRUDLinks` uses. The tab has its own `attach_observation_to_field_slip` `alt_title` so the derived class stays collision-free with other "attach observation to X" tabs.
- A rejected field-slip code stays in the field on re-render, per `validate_field_slip`'s documented contract that the form comes back with the user's input intact.
- A blank/no-op submission flashes and re-renders like any other rejected input, rather than silently redirecting back to the observation.
- A successful attach flashes its own confirmation — distinct from the full observation form's `:assigned_new`/`:assigned_existing` case, which stays silent since its generic "Observation successfully updated" flash already covers it (same reasoning already applied to `Location` during observation create; see `Locationable#save_location`'s commented-out `flash_notice(:runtime_location_success...)`).
- `dispatch_update_field_slip`'s `case` dispatch reuses the shared module's `add_field_slip_error` for `:invalid`/`:too_many` flash-setting, but is its own implementation rather than merged into `update_field_slip_or_flag_error` — this page's entire submission *is* the field-slip attach, so its assigned case needs a confirmation flash the full form's version correctly omits.
- `assign_field_slip` returns `:assigned_new`/`:assigned_existing` (via `FieldSlip#previously_new_record?`) instead of one bare `:assigned`, so `dispatch_update_field_slip` can flash the "already attached" notice without its own redundant `FieldSlip.exists?` check. The other two callers of `update_field_slip` (`ObservationsController`'s create/update paths) don't switch on the assigned case at all, so this is a no-op for them.

**Investigated and accepted as-is, not changed:**

- **A project-constraint force-join is unreachable from this page.** The full observation form's `ignore_proj_conflicts` checkbox lets an admin force-attach a slip to a project the observation violates; this standalone page has no such control, so a violating project always ends up spare (slip attached, project left unset) with a warning, even for an admin. An admin who needs to force the join can use the full observation edit form instead — adding that control here would expand this page well past "type a code."
- **The new attach link appears on the seldom-used, standalone naming form pages** (`namings/new`, `namings/edit`) with no unsaved-work guard, since `Details` renders there too. Real, but not a new class of risk — `SpecimenPanel` on the same page already has plain, unguarded navigate-away links (collection numbers, sequences), and the field-slip line itself already had this exposure whenever a slip was already attached. Also, nobody uses these pages, they use the modal naming form.
- **The `in_admin_mode? || obs.can_edit?(user)` permission gate is duplicated** across the controller, the view, and the Tab's doc comment. Checked whether the view could call the registered `permission?(obs)` helper instead to collapse this to one definition — every real controller that renders `Details` forwards its own `@user`, so no production request would ever notice. But `permission?` reads the ambient controller `@user`, not the view's own `user:` prop, and `details_test.rb` constructs `Details` directly with an explicit `user:` (e.g. `mary`) without logging that user into the test controller — routing through `permission?` would silently check the test harness's own empty session instead of the user the test is trying to verify. Composing from explicit args at each call site avoids that; left alone.

## Test plan

- [x] `bin/rails test test/controllers/observations/field_slips_controller_test.rb`
- [x] `bin/rails test test/views/controllers/observations/show/details_test.rb`
- [x] `bin/rails test test/classes/tab/observation/tabs_test.rb`
- [x] `bin/rails test test/controllers/observations_controller_show_test.rb` (full file, confirms `Details` changes don't regress existing show-page rendering)
- [x] Every regression test above confirmed to actually fail if its corresponding behavior is reverted (checked locally, one change at a time, then restored)

### DIFF
--- a/.claude/hooks/block_banned_words.sh
+++ b/.claude/hooks/block_banned_words.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+# Claude Code PreToolUse hook.
+# Fires before Edit/Write/MultiEdit (file content) and before Bash
+# calls that create/edit a git commit or a GitHub issue/PR/comment
+# (git commit -m, gh issue/pr create|edit|comment|review, gh api
+# ...body=, and any --body-file content). Blocks the action if the
+# new text contains one of the user's banned intensifier/hedge words.
+#
+# Why: personal style rules, corrected multiple times in conversation
+# and previously tracked only as memory (which the assistant has to
+# recall and apply itself, and can miss). See
+# .claude/rules/gh_pr_issue_formatting.md ("No \"real\"/\"genuine\"/\"actually\" as intensifiers")
+# and project memory feedback_banned_words_real_genuine.md /
+# feedback_banned_words_canonical_consume.md /
+# feedback_never_is_a_narrative_smell.md.
+#
+# Word list: real, genuine(ly), actual(ly), exactly, never, ever,
+# canonical, consume, "at all" -- used as intensifiers/hedges, they
+# carry no information and read as arguing for a claim instead of
+# just stating it. Whole-word matches only (so e.g. "actualization",
+# "eventual" don't false-positive); case-insensitive.
+#
+# Blunt by design, same philosophy as block_pii_in_gh.sh: false
+# positives get fixed by rephrasing, not by bypassing the hook.
+set -euo pipefail
+
+INPUT="$(cat)"
+TOOL="$(printf '%s' "$INPUT" | jq -r '.tool_name // ""')"
+
+# POSIX ERE doesn't treat `\b` as a word boundary (BSD/macOS grep in
+# particular) -- use the same `(^|[^[:alnum:]_])...([^[:alnum:]_]|$)`
+# portable boundary as check_any_phlex_props_on_save.sh.
+WORD_RE='(^|[^[:alnum:]_])(real|genuine(ly)?|actual(ly)?|exactly|never|ever|canonical|consume|at all)([^[:alnum:]_]|$)'
+
+TEXT=""
+case "$TOOL" in
+  Edit|Write|MultiEdit)
+    FILE="$(printf '%s' "$INPUT" | jq -r '.tool_input.file_path // ""')"
+    # Only source files -- not every Write (e.g. scratch/log files
+    # outside the repo) needs this scrutiny.
+    case "$FILE" in
+      .claude/*|*/.claude/*) exit 0 ;;
+      *.rb|*.rake|*.erb|*.scss|*.md) ;;
+      *) exit 0 ;;
+    esac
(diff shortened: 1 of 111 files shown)


nimmolo:
Rename `render_phlex_edit`

mo-nathan (approved):
Manual test looks good.

PR #5126, opened 2026-08-18
Show importable cap

### LINKED ISSUE #5122: Show info about requested imports exceeding cap
Issues:
- Expected Imports on the confirm form is incorrect if it exceeds MAX_IMPORTABLES. Ex:
>Requested Observations: 3789
...
Expected imports: 3710

- If an import gets to Done when it reaches MAX_IMPORTABLES, but there are still expected imports, the 
Tracker simply displays `imported: 2500 of 3710 observations`

Instead
- Confirm form should list Expected Imports as 2500 with a terse explanation and a message telling user they can import more in a separate import after this one is finished.
- Tracker should say something like `imported: 2500 of 3710 observations, you can import the remainder using additional imports`

JoeCohen:
Fixes a problem where requesting more than `InatImport::MAX_IMPORTABLE` (2,500) observations
silently truncated the import without explanation.

### Manual test plan

- [x] Since triggering a real >2,500-observation iNat import is impractical for
manual QA, Edit app/models/inat_import.rb:94 from MAX_IMPORTABLE = 2_500 to e.g. MAX_IMPORTABLE = 5, save.
Test through the browser as normal — no restart of bin/rails server or the Solid Queue worker needed.

This lowers the cap to 5 for the running server process (revert by restarting
the dev server). Pick an `inat_username` (or an iNat URL / id list) known to
have more than 5 importable fungi/slime-mold observations for the "over cap"
tests, and one with 5 or fewer for the "under cap" tests.

### Confirm page

- [x] Start a new iNat import for a username with **more** importable observations than the (lowered) cap. On the Confirm page, "Expected imports" shows the cap value (5), not the true higher count.
- [x] On that same Confirm page, a new line shows the number of observations in excess of the cap, with a note that they can be imported later in a new import.
- [x] On that same Confirm page, "Estimated time" reflects only the capped count (5 × the shown per-observation rate), not the full requested count.
- [x] Start a new iNat import for a username with **fewer or equal** importable observations than the cap. The Confirm page shows the true count in "Expected imports" and does **not** show the excess/over-cap line or note.

### Tracker page

- [x] Proceed with the over-cap import from above. While the import is still running (Importing state), the Tracker shows a count of observations not yet imported in excess of the cap.
- [x] While still Importing, the Tracker does **not** show any "import these now" link or message — only the running excess count.
- [x] Let the import finish (Done). The import stops at the capped count (imported count equals the cap), and the Tracker now also shows a link/message saying the remaining observations can be imported now in a new import.
- [x] Click that "import now" link. It lands on the New iNat Import form prefilled with the same username (and any ids/URL/"import all" used in the original import), ready to submit.
- [x] Submit that prefilled new import. It completes without re-importing any observations already imported in the first run (already-imported ones are skipped). (It is expected that Total Ignored, Already imported, and Exceeding per-import limit counts will be incorrect because those counts rely on the iNat `MushroomObserver URL` field, which is generally not written for test purposes.)
- [x] Run (or re-check) an import that stays **under** the cap end to end. The Tracker never shows the excess count or the "import now" line at any point, from start through Done.

### Super-importer scope (if you have super-importer access)

- [x] As a super-importer, run an over-cap import for another user's observations with "Import others" checked. Once Done, the Tracker's "import now" reimport link still has "Import others" set, not lost/reset.

### Cleanup

- [x] Reset set app/models/inat_import.rb#MAX_IMPORTABLE = 2_500. Double-check before committing anything.

### DIFF
--- a/app/classes/inat/observation_importer.rb
+++ b/app/classes/inat/observation_importer.rb
@@ -22,6 +22,7 @@ class Inat
     def import_page(page)
       page["results"].each do |result|
         return false if inat_import.reload.canceled?
+        return false if inat_import.reached_import_cap?
 
         import_one_result(JSON.generate(result))
       end
--- a/app/models/inat_import.rb
+++ b/app/models/inat_import.rb
@@ -156,6 +156,12 @@ class InatImport < ApplicationRecord
     imported_count.to_i >= MAX_IMPORTABLE
   end
 
+  # Observations beyond MAX_IMPORTABLE that a single run of this size
+  # won't import. 0 when count is at or under the cap.
+  def self.excess_over_cap(count)
+    [count.to_i - MAX_IMPORTABLE, 0].max
+  end
+
   def ignored_total_count
     ignored_not_importable_count +
       ignored_date_missing_count +
--- a/app/models/inat_import.rb
+++ b/app/models/inat_import.rb
@@ -156,6 +156,12 @@ class InatImport < ApplicationRecord
     imported_count.to_i >= MAX_IMPORTABLE
   end
 
+  # This run will never import more than MAX_IMPORTABLE, regardless of how
+  # many observations are actually available.
+  def capped_total_importables
+    [total_importables.to_i, MAX_IMPORTABLE].min
+  end
+
   # Observations beyond MAX_IMPORTABLE that a single run of this size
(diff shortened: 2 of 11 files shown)


mo-nathan (approved):
Look good.  Thanks!

Issue #5092, opened 2026-08-16
Add image rotate/flop icons to Image transform UI


nimmolo:
Also, reconsider the "Vote & Next" UI, changing it to a regular vote overlay, or otherwise DRYing it.

nimmolo:
Done via #5127.

PR #5127, opened 2026-08-18
Add rotate/mirror icons to `ImagePanel`'s transform controls

nimmolo:
## Summary

Part of #5092 — the rotate/flop-icons half. The issue's second ask ("reconsider the Vote & Next UI") is a separate, unstarted piece.

`Views::Controllers::Images::Show::ImagePanel`'s rotate/rotate/mirror/Original/EXIF controls row is now a Bootstrap `.btn-group`, with icons on the transform operations:

- **Rotate Left / Rotate Right / Mirror** — icon + visible label, outline-styled buttons (new `:rotate_left`/`:rotate_right`/`:flip` icon keys — the artwork already existed in the built sprite, so no `icon-library` changes were needed).
- **Show Original Image / Show EXIF Header** — stay plain text (no icon invented for them), but are now `.btn`-styled buttons so they sit correctly as members of the group instead of looking like stray links next to it.

## Design

- `Components::ImageFragment::OriginalLink` gets a new `button:` prop and `Components::ImageFragment::EXIFLink`'s previously-hardcoded `variant: :strip` becomes a configurable `variant:` prop, both forwarded into the existing `Link`/`Button` calls. Their other caller, `lightbox_caption.rb`, doesn't pass either, so its plain-link rendering is unchanged.
- `ButtonGroup(class: "btn-group-sm") { ... }` replaces the old `" | "`-separated plain-text row.

## Depends on #5136

This transform-controls row mixes plain `.btn`s (Original/EXIF links) with `form.button_to`-wrapped CRUD buttons (Rotate Left/Right, Mirror) inside one `.btn-group`. Bootstrap's own `.btn-group` CSS doesn't see through the `<form>` wrapper, so the form-wrapped members get none of the group's border-radius squaring, margin collapse, or height matching on their own.

That fix was originally written directly in this PR but has moved to #5136, since it's a general CSS supplement (`app/assets/stylesheets/mo/_links_buttons_alerts.scss`) unrelated to this PR's icon additions, not something scoped to the image page. #5136 needs to merge first for this transform-controls row to render with correct border-radius/height joining — until then, the group's form-wrapped members render with the pre-existing BS3 float-based layout (mismatched height, square corners against their neighbors).

## Test plan

- [x] `bin/rails test test/classes/icon_glyph_sync_test.rb`
- [x] `bin/rails test test/components/image_fragment/original_link_test.rb`
- [x] `bin/rails test test/components/image_fragment/exif_link_test.rb`
- [x] `bin/rails test test/controllers/images_controller_test.rb`
- [x] `bin/rails test test/system/image_show_system_test.rb`
- [x] Manual visual check (after merging #5136 locally): image show page's transform controls (joined, consistent height, text/icon-label baselines aligned)


nimmolo:
Add BS4 bridge CSS for InputAddOn similarly, to fix IndexPagination and others.

PR #5130, opened 2026-08-19
Truncate oversized import digests

### LINKED ISSUE #5129: oversize import digest
The ImportJob doesn’t notify interested user about imported Observations as soon as each Observation is imported. Instead, it accumulate notices of interest in the imported Observations, and sends a single digest per user when the Import is Done.
A single digest can get too large, generating the following error when the job is done:
```ruby
2026-08-17 13:58:52 -0700: InatImportJob #<InatImport:0x00000001313ad860> {id_above: 256671873, continuation: true} Import digest failed: SolidQueue::Job::EnqueueError: ActiveRecord::ValueTooLong: Trilogy::ProtocolError: 1406: Data too long for column 'arguments' at row 1 (trilogy_query_recv)
```
This silently blocks
1. emailing that digest to that user, and
2. any subsequent emails that are lined up.

JoeCohen:
- Caps each digest at 500 observations to prevent overflowing`solid_queue_jobs.arguments.
- Isolates each user's `deliver_later` call in its own rescue so one failure can't cascade.
- Delivers #5129 

### Manual Test ✅

I ran the exact large import which surfaced the bug.
It completed with 4 truncated digests ("This email shows the first 500 of [n] matching observations ... ." )
and without an Import Digest failure.

### DIFF
--- a/app/classes/inat/import_digest.rb
+++ b/app/classes/inat/import_digest.rb
@@ -36,7 +36,7 @@ class Inat
       by_observation = namings.group_by(&:observation_id)
       total = by_observation.size
       capped = capped_namings(by_observation)
-      log_truncation(user, total) if capped.size < total
+      log_truncation(user, total) if total > MAX_DIGEST_OBSERVATIONS
 
       InatImportDigestMailer.build(
         receiver: user, namings: capped, total_observations: total
--- a/app/classes/inat/import_digest.rb
+++ b/app/classes/inat/import_digest.rb
@@ -11,18 +11,60 @@ class Inat
       new(inat_import).deliver
     end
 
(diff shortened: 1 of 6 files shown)


nimmolo:
@JoeCohen Does the email currently print a digest comment for every obs imported? 

That itself sounds like it could be dropped, not just capped at 500. Who is reading those? Maybe only digest the errors?

JoeCohen:
@nimmolo: It's the batched counterpart for what happens whenever someone creates an Observation.
Normally, a callback will immediately email every who has an Interest in the taxon. 
Imports, however, do not send those emails immediately. Instead they cumulate all the info that would normally be sent in    after creation. When the Import is Done, it sends each  (interested) user a digest of all the info that would normally be in separate emails.

Issue #4932, opened 2026-07-27
Have `/qr/<code>` go to the Create Observation page rather than the Create Field Slip page


mo-nathan:
Now that you can enter the field slip code on the Create Observation page, there is no good reason that the `/qr/<code>` URLs as encoded on an MO field slip should go to the Create Field Slip page rather than directly to the Create Observation page with the field slip code correctly filled in.

Changing to the Create Observation page would allow users using field slips to create new observations to immediately add images instead of having to first fill in the field slip data and then click "Add Images or Other Data" to get to the Create Observation page. Adding the images up front would help resolve the location information immediately from the images. However, project aliases should still be applied to a value in the Locality field. Similarly the Collector field should check project aliases if a project is known based on the field slip code for an explicit user override. The Scientific Name field makes the "ID" field from the field slip form unneeded. However, the "Id by" field should move down to the notes area and be resolved by project aliases if appropriate.

It will be important to get the notes right. Currently the Create Field Slip page automatically displays the fields: Odor / Taste, Trees / Shrubs, Substrate, Habit, Other. This behavior would have to be added to the Create Observation page. However, the user's own `notes_template` should appear before the field slip fields. If any of the field slip fields are already in a user's `notes_template`, then they should only be shown in the position indicated by the user's `notes_template`. This is also where the "Other Codes" field and the "Check if Other Codes is an iNat ID" check box should be added.

If the page is initially rendered with a field slip code, then the "Specimen Available" checkbox should be checked. In addition, if the field slip code text box is empty and the user adds anything to it, then the "Specimen Available" checkbox should become checked. Note that users should still be able to explicitly uncheck the "Specimen Available" checkbox when there is content in the code text box, and if they edit that text it should remain off. However, if they completely blank out the field and then type into it again, it should again become checked.

---

# Design decisions

Worked out in discussion 2026-07-28. Where these differ from the description above, these win.

## Guiding principle

The field slip workflow must "just work" for someone who has never used MO. Conceptually: *scan the field slip, see "the" observation.* Most users are not expected to understand occurrences, so the flow must never require that concept.

## Scope: the Create Field Slip page stays, transitionally

The page is not removed. It stops being where a scanned slip lands, but remains reachable (the Field Slips tab, the code-only entry form, `field_slips#edit`) as a backup workflow in case the observation form turns out not to cover the foray case as well as expected. It can be removed later once the new flow is demonstrably sufficient.

[...]

PR #5177, opened 2026-08-23
Edit on a read-only reflection opens a companion observation (#4214)

mo-nathan:
## Summary

Part of #4214 / #4208: the "Edit creates a companion" half of read-only reflections, which #4852 left as a flash-and-bounce.

Edit on a read-only reflection now opens a **companion observation** in the same occurrence:

- If the occurrence already has a member that is not a reflection and the user may edit, Edit goes to that member's edit form.
- Otherwise `Observation::Companion` creates one: copies the snapshot (date, `where`/location, GPS, altitude, hidden flag, collection-location flag, specimen flag, notes, collector text — `collector_user_id` is re-derived by the model's own callback), gives it the user's naming of the reflection's consensus name, shares the reflection's images and thumbnail (logged as reuse), links the two in an occurrence with the reflection as primary (or joins the reflection's existing occurrence), and adds it to the reflection's projects — members of an occurrence share project membership (#4932). Created by the editing user with source `mo_website`, so it is an ordinary editable observation; resync keeps ignoring it.
- The edit form opens with a flash explaining what happened (two variants: companion created / existing companion). A full occurrence surfaces its validation error and bounces to Show.

Update on a reflection is unchanged (still blocked with the existing warning), so a stray PUT cannot spawn companions. The Show page's read-only note now says Edit opens a linked companion.

## Test plan

Setup: an observation imported from iNat on your account (a reflection — `reflected_at` set), ideally with photos and in a project.

1. Click Edit on it. You land on the edit form of a *new* observation with the same date, location, GPS, notes and photos, and a flash saying a companion was created and linked. Judge the flash wording here.
2. The original's Show page now shows a matching-observations panel with the companion; the original is primary; both are in the same projects.
3. Click Edit on the original again: you land on the same companion (no new observation), with the "existing companion" flash.
4. Save a change on the companion (e.g. GPS): the reflection is unchanged.
5. Sync now on either member still works and leaves the companion alone.

article: yes
Edit an imported observation through a linked companion

https://claude.ai/code/session_01JPMv2YTuAkun3UNCgigTPu

### DIFF
--- a/app/controllers/observations_controller/edit_and_update.rb
+++ b/app/controllers/observations_controller/edit_and_update.rb
@@ -25,7 +25,7 @@ module ObservationsController::EditAndUpdate
   #   @good_images                      list of images already attached
   #
   def edit
-    return unless editable_or_redirect?
+    return unless editable_or_redirect?(companion: true)
 
     init_license_var
     init_new_image_var(@observation.when)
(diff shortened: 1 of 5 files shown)


mo-nathan:
Closing in favour of a fresh PR from the same branch: Copilot's first review errored, and an errored PR keeps erroring (see .claude/rules/copilot_review_comments.md).

Issue #4585, opened 2026-06-23
Reflection resolution: reconcile MO observations against their iNat counterparts


mo-nathan:
## Goal

After #4565 materializes the raw MO↔iNat correspondences as `ExternalLink`s, reconcile each MO observation against its corresponding iNat observation(s) so the MO side becomes a clean **reflection** — a reflection MO obs equals what a fresh import of that iNat obs would produce at a point in time. This is the engine #4214 (read-only reflection + occurrence editing) and #4215 (sync) build on.

## Single iNat obs ↔ single MO obs: comparison cases

Comparing an iNat obs to its MO obs by content (images are primary — see "Comparison granularity"):

1. **Identical**, or **MO ⊆ iNat** (iNat is a superset, e.g. more images) → the existing MO obs is the (possibly early) reflection; sync proceeds normally.
2. **iNat ⊊ MO** (MO has extra/adopted images or data not on iNat) → split the MO-only data into a **new** MO obs; reduce the existing MO obs to match the current iNat obs (it becomes the reflection).
3. **Disjoint** → import the iNat obs into a **new** reflection MO obs; link the two MO obs via an Occurrence.
4. **Overlapping** → move the MO-unique data to a **new** MO obs; the existing MO obs becomes an (older) reflection of the iNat obs.

End state in every case: each reflection MO obs equals a fresh import of its iNat obs.

## Multiple iNat obs → one MO obs

Real data has MO obs referenced by several iNat obs — a mix of legitimate iNat-side duplicates of one collection and user error. We do **not** force one link per obs (that constraint is removed in #4565) and we do **not** auto-clean these; careful human review is required and new cases will keep appearing.

Choosing the reflection for such an MO obs:
- If an **import link already exists**, it determines the reflection (set at import time). Example: MO 634003 (iNat 10592075 + 10339038) — the lower id 10339038 is correct, and the existing import link already captures that.
- Otherwise (ambiguous), prefer the **highest iNat id**. Example: MO 233468 (iNat 4749134 + 2712928) — both are subsets of the MO obs by images, and the more recent 4749134 (more images) is the better reflection.

**Do not spawn a new reflection MO obs per extra iNat obs** — that would require writing the MO-URL field back to iNat, which we want to avoid. Non-chosen iNat obs remain as additional `ExternalLink`s on the MO obs, flagged for human review.

`only_one_import_per_target` is retained: an MO obs has at most one `import` reflection.

## Comparison granularity

[...]

PR #5196, opened 2026-08-24
Exclude MO's own back-link field from the iNat snapshot

mo-nathan:
## Summary

The iNat snapshot stored in an imported observation's notes lists every iNat observation field — including field 5005, **"Mushroom Observer URL"**, which MO itself writes onto the iNat observation after import as a back-link.

At import the field isn't present yet, so the stored snapshot records `Observation fields: none`. On resync, iNat returns the observation *with* the field, so the freshly-built snapshot reads `Observation fields: … Mushroom Observer URL: https://mushroomobserver.org/…` and differs from the stored one — so the daily resync (#4215) flags `notes` as changed for **every back-linked reflection**. It's MO seeing its own back-link reflected back through iNat.

Measured against a production snapshot: a full first pass reported 5,742 of 5,746 reflections "changed", driven by this. A 15-reflection convergence sample went from **15/15 `notes` changes to 4/15** with this fix (the remaining 4, plus `lat`/`lng`/`specimen` drift, are per-obs changes that reflect iNat drift).

**Fix:** exclude field 5005 in `Inat::Obs#obs_fields`, so the snapshot mirrors iNat's own data rather than MO's annotation of it. This makes the resync convergent (no spurious first-pass rewrite of ~every reflection, no burst of `log_observation_resynced` entries) and drops a redundant self-link from imported notes.

Only the snapshot is affected — `inat_obs_fields` stays intact for the #4585 extract/comparator, which needs the raw field list.

## Test plan

- `bin/rails test test/classes/inat_obs_test.rb` — green, including `test_obs_fields_excludes_mo_url_back_link` (field 5005 is dropped, other fields kept) and `test_obs_fields_none_when_only_mo_url_back_link`.
- Verified against a fresh production snapshot: the convergence sample's `notes` change rate fell from 15/15 to 4/15.

article: no

https://claude.ai/code/session_01JPMv2YTuAkun3UNCgigTPu

### DIFF
--- a/app/classes/inat/obs.rb
+++ b/app/classes/inat/obs.rb
@@ -378,7 +378,14 @@ class Inat
       "#{self[:location]} +/-#{self[:public_positional_accuracy]} m"
     end
 
+    # Excludes MO's own "Mushroom Observer URL" back-link (field 5005): MO
+    # writes it onto the iNat obs after import, so keeping it here would
+    # make every back-linked reflection's snapshot differ from its stored
+    # (pre-back-link) form on the first resync. The snapshot mirrors iNat's
+    # own data, not MO's annotations of it.
     def obs_fields(fields)
+      fields = Array(fields).
+               reject { |f| f[:field_id] == MO_URL_OBSERVATION_FIELD_ID }
       return :none.t if fields.empty?
 
       "\n#{one_line_per_field(fields)}"
--- a/test/classes/inat_obs_test.rb
+++ b/test/classes/inat_obs_test.rb
@@ -164,6 +164,38 @@ class InatObsTest < UnitTestCase
     assert_nil(mock_obs.when)
   end
 
+  # MO's own "Mushroom Observer URL" back-link (field 5005) is excluded
+  # from the snapshot -- MO writes it onto the iNat obs after import, so
+  # keeping it would make the reflection's snapshot differ from its stored
+  # form on the first resync.
+  def test_obs_fields_excludes_mo_url_back_link
+    obs = Inat::Obs.new(JSON.generate(
+                          ofvs: [
+                            { field_id: MO_URL_OBSERVATION_FIELD_ID,
+                              name: "Mushroom Observer URL",
+                              value: "https://mushroomobserver.org/1" },
+                            { field_id: 42, name: "Voucher Number",
+                              value: "AN 0432" }
+                          ]
+                        ))
+    result = obs.obs_fields(obs.inat_obs_fields)
+
+    assert_not(result.include?("Mushroom Observer URL"),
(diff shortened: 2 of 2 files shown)


JoeCohen:
@Nathan:
Would it be better to add that field during the ImportJob?
The Job currently:
- Creates an MO Observation with a snapshot
- Writes iNat's `Mushroom Observer URL` field

What if the job then modifies the MO Observation's snapshot?
This would also let the sync pick up any changes to that field.

mo-nathan:
I don't understand what you're suggesting when you ask "What if the job then modifies the MO Observation's snapshot?"

I can kind of see the point about the MO URL field changing presumably manually on iNat, but that also creates kind of a weird circularity.  The sync script looks up the observations that have changed on iNat based on the presence of a value for MO URL.  It then finds the MO obs by looking for iNat obs number in the ExternalLinks.  The question then becomes what do do we do if the MO URL field disagrees with the MO obs number from the ExternalLinks table?

mo-nathan:
Following up on the back-link integrity question (thanks @mo-nathan for the framing):

**The ExternalLink stays authoritative for the sync.** Discovery keys on the *presence* of the `Mushroom Observer URL` field; the iNat→MO mapping uses the import `ExternalLink`. So a wrong field value can't misroute a sync, and force-writing the field back to a computed value would be circular — we don't do that.

**Value comparison — added to #5198** (`Inat::ReflectionBatchResyncer#check_back_link`, commit 5e89081496). During the batch resync, each fetched obs's field-5005 value is compared to the MO obs its `ExternalLink` points at. A mismatch (edited on iNat, or pointing at a different obs) is routed to #alerts by the job for a human to review — no auto-repair. Editing the field bumps the observation's `updated_at`, so the incremental `updated_since` fetch still surfaces it. A 60-obs sample is 60/60 clean today, so this is a tripwire, not a known problem.

**Completeness reconciliation — new issue #5200, and it does not need to land before NAMA.** The value check can't catch the field being *cleared or deleted* (the `field:` filter excludes those). A periodic (monthly-ish) job verifies every import `ExternalLink`'s observation still appears in a `field:Mushroom Observer URL` search, alerting on any that fell out — which also subsumes the deferred deletion-detection sweep.

On the snapshot suggestion, @JoeCohen: field-value integrity is a separate concern from what we store in the notes snapshot. Folding the field's value back into the snapshot would re-introduce the circular self-reference this PR removes (and the every-reflection-churns diff), so the exclusion here stands as-is; the tracking lives in the two checks above instead.

JoeCohen:
In my mind "At import the field isn't present" is the root of problem. 
Why not fix that, rather than work around it later.

PR #5207, opened 2026-08-25
Add new cop `MO/NoUncapturedTestLogging`; fix three tests it flagged

nimmolo:
## Summary

Adds `MO/NoUncapturedTestLogging`: flags a test method that deliberately triggers a failure (a literal `raise`/`fail` inside a `.stub(...)` block or `define_singleton_method`) without also stubbing/capturing the resulting `Rails.logger`/`warn` call the production rescue path makes. Left un-stubbed, that log line hits the test logger's `$stdout` destination, dumping into the suite's console output looking like a failure in an unrelated test -- the pattern this PR was prompted by (an iNat retry warning and an `InatImportDigestMailer` error log line showing up in unrelated test runs).

- Excludes `assert_raises`/`assert_nothing_raised` -- the raise is meant to propagate to the test's own assertion, not be swallowed by a rescue that logs.
- Excludes stubs of stdlib/vendor I/O primitives (`File`, `Tempfile`, `Open3`, `Dir`, `FileUtils`, `IO`, `RestClient`) -- confirmed by reading each one's production rescue path (`Image#save_to_temp_file`, `Observation.define_a_location`, `IpStats#blocked_ips_current?`, `FieldSlip::QRDecoder.remote_codes`, `ParallelTestConfigService`) that MO's own rescues around these don't call `Rails.logger`, only domain-class failures do.
- Excludes tests wrapped in MiniTest's `capture_io` -- it redirects `$stdout`/`$stderr` for the whole block, so a `warn`/`puts` inside it is already suppressed.

Fixed the three tests the cop found with a `Rails.logger` call in their production rescue path (`Inat::ImportDigest`'s mailer-isolation test, `InatMoObservationBuilder`'s name-override-resolution test, and `Inat::ObservationImporter#create_crosslink`'s validation-failure test) to stub `Rails.logger` and assert on the captured message instead of letting the message print.

The remaining flagged tests were individually triaged by reading each one's production rescue path and confirmed as false positives -- no logging call in the rescue, or the rescue routes through a job-log file / `output_handler` / injectable callback that isn't console output. Each is inline-disabled with the specific reason found at that site.

## Test plan

- [x] Rubocop -- clean repo-wide
- [x] Every touched test file -- all green (18 files, individually run)

article: no

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,7 @@ require:
   - ./lib/rubocop/cop/mo/no_raw_link_or_button_to
   - ./lib/rubocop/cop/mo/no_render_phlex_method_name
   - ./lib/rubocop/cop/mo/no_resolved_errors_add_type
+  - ./lib/rubocop/cop/mo/no_uncaptured_test_logging
   - ./lib/rubocop/cop/mo/prefer_kit_syntax
   - ./lib/rubocop/cop/mo/redundant_render_view_invalid
   - ./lib/rubocop/cop/mo/self_status_outside_invalid_method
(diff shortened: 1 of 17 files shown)


nimmolo:
Superseded by #5211 -- replaced the RuboCop-cop approach with a runtime hook that measures test console output directly instead of inferring it from test-side syntax. The cop flagged 19 tests for 3 leaks; the runtime hook has no false positives by construction. Closing this one.

Issue #5186, opened 2026-08-23
Duplicate namings: same user, same name, same observation (231 groups in production)


mo-nathan:
A single observation can carry more than one naming by the same user for the same name. There is no reason a duplicate naming — same `(observation_id, user_id, name_id)` — is correct: the naming form updates a user's existing naming in place, so one per user per name per observation is the intended shape.

## Scope in production (snapshot ~2026-08-21)

- **231 duplicate `(observation, user, name)` groups** → 235 extra namings, across 229 observations and 69 users, out of ~840,025 namings.
- Long-standing, not a single recent regression: the extra namings span every year from 2008 to 2026 (roughly 10–25 per year). Only ~19 are from 2026.
- The ~19 recent ones sit on occurrence members and come from the field-slip review flow (see fix below). The other ~216 predate that flow and come from some other path — a double-submit of the naming form or an API/import race is the likely candidate, not yet pinned down.

## Already fixed: the field-slip review path

`FieldSlip::Extractor::NameProposer#propose_naming` always constructed a new `Naming`, so saving the same slip review twice (a re-review, a rescan) stacked identical namings on the observation. Fixed to reuse the reviewer's existing naming for that name and re-cast the vote (matches the naming form's own behavior). This closes the ~19 recent groups' source; the historical source is still open.

## Cleaning up the 231 existing groups is not a plain delete

Two complications make "keep the earliest, delete the rest" lossy:

- **92 of the 231 groups have votes from other users** (200 such votes) on one of the duplicate namings. Deleting a naming deletes its votes, so those need to be migrated onto the kept naming — and de-duped where a voter voted on more than one of the duplicates.
- **44 groups have differing reasons** across the duplicates (same user and name, but different used-reasons — e.g. "Recognized by sight" on one, reference text on another). The keeper should be chosen to preserve the richest reasons, not blindly the earliest.
- The remaining 139 groups are clean (only the proposer's own vote, identical reasons) and merge trivially.

Proposed repair: a dry-run-default runner that, per group, keeps one naming (most used-reasons, tie broken by earliest), migrates and de-dups its votes onto the keeper, drops the extras, and recomputes each observation's consensus. ID-level reporting; `--apply` to write.

## Open questions for the team

1. **Keeper rule when reasons differ** — is "most used-reasons, tie by earliest id" the right choice, or should differing reasons be merged onto the keeper rather than dropped with the loser?
[...]

PR #5212, opened 2026-08-25
Enforce one Naming per `(Observation, User, Name)` (#5186)

mo-nathan:
## Summary

Enforces the invariant that a user may propose a given `Name` on a given `Observation` at most once, via a **unique index on `namings(observation_id, user_id, name_id)`** plus a **matching model validation** on `Naming`.

This is the constraint half of the #5186 duplicate-naming work. The *prevention* paths landed in #5204 (name-merge now folds a colliding naming via `Naming#absorb`; the iNat import find-or-reuses instead of stacking a second naming). This PR closes the invariant at the database level so no path can create a duplicate, and adds a graceful validation for the two write paths that don't pre-check:

- the plain `create` double-submit race (`name_been_proposed?` is a soft, race-able check), and
- `create_new_naming` — a user re-points a naming that's locked (already upvoted by others) to a name they've already proposed.

Both call `validate_object(@naming)`, so the validation surfaces as a flash error rather than a `RecordNotUnique` 500. The unique index remains the hard backstop for the true check-then-insert race.

## Deploy ordering — important

The migration adds a `unique: true` index, which **fails if any duplicate `(observation_id, user_id, name_id)` rows still exist**. So the sequence is:

1. #5204 deploys (stops new duplicates from the merge/import paths).
2. Run `projects/5186/dedup_namings.rb --apply` on production to collapse the existing duplicates (233 groups / 233 extra namings in the current snapshot).
3. Merge + deploy this PR.

If a stray duplicate slips in between 2 and 3, the migration fails loudly (no partial state) rather than silently skipping — re-run the dedup and retry.

## Notes

- **Occurrence members keep their own copies.** Each occurrence member is a distinct `Observation`, so `(observation_id, user_id, name_id)` still differs across them.
- **Orphaned namings are removed first.** The dedup script deletes any naming whose observation is gone (NULL or dangling FK) before the duplicates are folded, so none remain when the index is built.
- The validation is guarded to only run when all three key columns are set (`check_requirements` already reports missing ones) and excludes the record itself on update.

## Test plan

- [x] `bin/rails test test/models/naming_test.rb test/models/name/merge_test.rb test/models/observation/merged_naming_test.rb`
- [x] `bin/rails test test/controllers/observations/ test/controllers/observations_controller_create_test.rb`
- [x] `bin/rails test test/classes/api2/observations_test.rb test/classes/inat_mo_observation_builder_test.rb`
- Reviewer: confirm the deploy ordering above — this PR should merge/deploy only after the dedup script has run on production.

article: no

### DIFF
--- a/app/models/naming.rb
+++ b/app/models/naming.rb
@@ -408,6 +408,20 @@ class Naming < AbstractModel
                                                         !@current_user
   end
 
+  # A user may propose a given name on an observation only once; the DB
+  # unique index (#5186) is the hard backstop, this is the graceful one
+  # so a double-submit or a re-pointed naming flashes an error instead of
+  # raising RecordNotUnique. Skipped when the key columns aren't all set
+  # (check_requirements already reports those).
+  validate :name_not_already_proposed_by_user
+  def name_not_already_proposed_by_user # :nodoc:
+    return unless observation_id && user_id && name_id
+
+    scope = Naming.where(observation_id:, user_id:, name_id:)
+    scope = scope.where.not(id:) if persisted?
+    errors.add(:name, :validate_naming_duplicate) if scope.exists?
+  end
+
   private
 
   # Move the other naming's votes onto this one, keeping the higher value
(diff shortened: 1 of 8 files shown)


mo-nathan:
Closing and reopening from the same branch: Copilot's review errored on this PR (it stays stuck once it errors). New PR carries the same description.

Issue #5139, opened 2026-08-19
4636 #3 - Migrate remaining view-layer residuals in `index_active_params` shortcuts


nimmolo:
Part of #4636. Depends on #5137 (foundation) and on the audit done in #5138.

A small number of `index_active_params` shortcuts have residual custom logic beyond a simple alias, found during #5138's audit. Two kinds:

**View-layer chrome.** A shortcut sets a controller ivar for something `Query` has no business owning — banner/button/permission-check chrome, not the page title (already fully generic via `add_index_title`/`FilterCaption`). For each, keep the ivar but derive it by reading the already-resolved record off the query instead of a second independent lookup.

- [x] `ObservationsController::Index#project` — #5217. `@project` derived once, in `filtered_index_final_hook`, from `query.params[:projects]` -- covers both the `project` URL shortcut and a bookmarked `q[projects][]=` query.
- [x] `CollectionNumbersController#observation` — #5217. Same pattern; also fixed the old code calling `Observation.find(params[:observation])` directly, which raised `RecordNotFound` (500) on a bad id instead of flashing/redirecting.

**Missing Query-level validation/scope.** A shortcut hand-rolls value normalization, or a value-dependent dispatch, in the controller that belongs on the Query/model layer instead. A scope is a plain lambda -- it can branch on its input and `return` early, so a discriminator-style shortcut doesn't need controller-side dispatch, just a smarter scope.

- [x] `RssLogsController#type` — #5217. `RssLog` gained a `types` scope (array/scalar input, "all"/"none" handled once, centrally) and `Query::RssLogs` declares `types: [:string]` with `param_alias: :type`.
- [x] `Observations::IdentifyController#filter` — #5217. `Observation` gained an `identify_filter` scope (structural-hash `query_attr`, same shape as `Comment#target`) picking `clade`/`region` based on a sibling `filter[type]` field; renamed the form/URL namespace to `identify_filter[...]` to match.
- [x] `UsersController#pattern` — #5217. `User.pattern` scope now checks `User.exact_match` (numeric id or verified email) first and returns early, same technique as the two scopes above. Confirmed the exact match has to keep winning outright rather than get unioned with the fuzzy search -- `User.pattern("1")` fuzzy-matches ~17,000 users in production, so a single OR'd query would break id-based lookup for most ids. `SearchController` has its own separate, structurally different copy of the same exact-match check (redirects to a show page before any Query exists, for every searchable model) -- left untouched.

All five done in #5217.

PR #5221, opened 2026-08-25
Give `Observation` an `identify_filter` scope; drop clade/region dispatch (#5139)

nimmolo:
## Summary

Part of #5139. `Observations::IdentifyController#filter` hand-dispatched a single clade-or-region autocompleter submission (`filter[type]`/`filter[term]`) to two separate query-building methods. Moves that dispatch onto a new `Observation.identify_filter` scope, and declares it as a structural-hash `query_attr` on `Query::Observations` -- the same shape `Comment#target` (`{ type:, id: }`) already uses for its own type-discriminated dispatch.

### What changed

- `Observation::Scopes#identify_filter` -- picks `clade`/`region` based on `filter[:type]`, falling back to `none` for anything else.
- `Query::Observations` -- `query_attr(:identify_filter, { type: :string, term: :string })`.
- `Observations::IdentifyController` -- `filter`/`clade`/`region` (3 methods) collapse into one `identify_filter` method that permits the nested hash and calls `create_query` directly, merging in `needs_naming: @user`/`order_by: :rss_log` (session-scoped context the URL doesn't supply, so this doesn't go through `create_query_from_url_params`).
- URL/form namespace renamed `filter[...]` -> `identify_filter[...]` to match the attr name directly -- this is an internal identify-page form, not a bookmark-compat surface, so there's no reason to route it through a mismatched alias. `FormObject::IdentifyFilter`'s `model_name` override (which forced the old `"Filter"` namespace) is gone; Superform's default derivation already produces `identify_filter`.
- Added `query_identify_filter`/`query_term` to `en.txt` for the index-bar filter caption (a new top-level attr name needs its own label).

## Test plan

- [x] Rubocop -- clean
- [x] `test/controllers/observations/identify_controller_test.rb` -- green, params renamed to `identify_filter[...]`
- [x] `test/views/controllers/observations/identify/form_test.rb` -- green, field name/id assertions updated
- [x] `test/models/observation_test.rb` -- green, new `test_scope_identify_filter` covering the clade/region/unknown-type dispatch
- [x] `test/classes/query/observations_test.rb` -- green

article: no

### DIFF
--- a/app/classes/form_object/identify_filter.rb
+++ b/app/classes/form_object/identify_filter.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 # Form object for the identify observations filter form.
-# Params namespaced as filter[term], filter[type], etc.
+# Params namespaced as identify_filter[term], identify_filter[type],
+# etc. -- matches Query::Observations' identify_filter query_attr
+# directly, no controller-side param translation needed.
 class FormObject::IdentifyFilter < FormObject::Base
   VALID_TYPES = %w[clade region].freeze
 
@@ -14,9 +16,5 @@ class FormObject::IdentifyFilter < FormObject::Base
   # trust the value.
   def type=(value)
     super(VALID_TYPES.include?(value) ? value : "clade")
-  end
-
-  def self.model_name
-    ActiveModel::Name.new(self, nil, "Filter")
   end
 end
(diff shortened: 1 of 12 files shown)


nimmolo:
Consolidated into #5217 -- combining both #5139 scope conversions into one PR rather than one per shortcut, per repo sweep convention. Closing this one.

PR #5258, opened 2026-08-28
Add script to standardize an iNat import's project/field-slip links

mo-nathan:
## Summary

Adds `script/standardize_inat_import_project_slips.rb`, a dry-run-by-default, idempotent runner that reviews the observations created by one iNat import and settles their project and field-slip associations against a target project (e.g. the 2026 NAMA Yoop project).

```
bin/rails runner script/standardize_inat_import_project_slips.rb <import_id> <project_id>          # dry run
bin/rails runner script/standardize_inat_import_project_slips.rb <import_id> <project_id> --apply  # live
```

Motivated by a test import of 647 iNat observations for the 2026 NAMA Yoop foray, where the field-slip scanner (run during import) produced a mix of states: some observations got the right `2026-NAMA` slip, some got a reused old `2025-NAMA` slip, some carry the code only in their notes, and some are duplicates of native foray observations already scanned into the project.

## Per-observation logic (first match wins)

1. **iNat-linked native partner.** A native (non-import) observation whose notes cite this observation's iNat number and which carries a field slip is the same physical specimen scanned during the foray. The import observation joins that native's occurrence and takes its (correct-year) slip; the native stays primary, since an import reflection may not be an occurrence's primary. An import observation that already had an occurrence is merged in with `Occurrence.merge!`, so a reused `2025` slip yields to the native's `2026` slip.
2. **A reused/spare slip** (wrong project, no native counterpart) is reassigned to the target project, keeping its printed code.
3. **A target-prefix code sitting in the notes** (collector wrote it in the iNat description rather than photographing a scannable slip) gets its slip attached via `FieldSlip::Attacher`.
4. **Anything else** is simply added to the target project.

Every imported observation ends up a member of the target project. If the `join_in_use` attach in case 3 lands a reflection as an occurrence's primary while a native member is present, `keep_native_primary` hands primary back to the oldest native member; an all-reflection occurrence keeps the reflection primary. `FieldSlip#project=` is bypassed with `project_id=` in case 2, since its add-slip permission gate silently keeps `nil` for a backfill.

## Validation

Run against the local test import (443 into project 407), 647 observations:

- **Dry run:** 147 already-standardized, 351 added-bare, 71 joined-native, 68 attached-from-notes, 9 reassigned-slip, 1 merged-into-native. 0 constraint violators.
- **Apply:** all 647 land in the project; occurrence joins/merge, slip attach, and slip reassign all persist; the one merge case (a `2025-NAMA` reflection into its native's `2026-NAMA` occurrence) leaves the native primary.
- **Re-run:** clean no-op (351 + 224 + 72), confirming idempotency.

## Test plan

- [x] Dry-run, `--apply`, and idempotent re-run against a local import.
- Reviewer: on a copy of production, dry-run against an import id + the target project id and eyeball the buckets before applying.

article: no

### DIFF
--- a/script/standardize_inat_import_project_slips.rb
+++ b/script/standardize_inat_import_project_slips.rb
@@ -1,0 +1,252 @@
+# frozen_string_literal: true
+
+# Standardizes the project and field-slip associations of the
+# observations created by one iNat import, against a target project
+# (e.g. the 2026 NAMA Yoop project). Idempotent: safe to re-run.
+#
+#   Dry run (default -- reports what WOULD change, writes nothing):
+#     bin/rails runner script/standardize_inat_import_project_slips.rb \
+#       <import_id> <project_id>
+#   Live run:
+#     bin/rails runner script/standardize_inat_import_project_slips.rb \
+#       <import_id> <project_id> --apply
+#
+# For each imported observation, in priority order:
+#
+#   1. iNat-linked native partner. A native (non-import) observation whose
+#      notes cite this observation's iNat number, and which carries a
+#      field slip, is the same physical specimen scanned during the foray
+#      -- its slip (the correct-year one) wins. The import observation
+#      joins that native's occurrence; the native stays primary (an
+#      import reflection may not be an occurrence's primary). An import
+#      observation that already sits in an occurrence is merged in, so a
+#      reused 2025 slip yields to the native's 2026 slip.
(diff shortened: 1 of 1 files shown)


mo-nathan:
Closing without merging, per the plan on #5259: the standardization belongs in the importer/website, not a shell script on the production server. The script stays visible here as the logic reference for the importer implementation.

PR #5263, opened 2026-08-28
Make `Project#violations` build an AR relation, paginate with SQL

### LINKED ISSUE #5256: Make `Project#violations` build an AR relation instead of a Ruby Array
## Summary

`Project#violations` (used by `Projects::ViolationsController#index`, paginated manually as of #5255/#5250) computes violations by iterating every visible observation in Ruby (`visible_observations.includes(...).filter_map { |obs| ... }`), then sorting the full result in Ruby. That's `O(all visible observations)` work on every page load, even though only one page's worth gets displayed.

`Project#count_violations` already proves the existence check is fully SQL-expressible: it collects violating-observation ids via four AR-scope-backed helpers (`collect_date_violation_ids`, `collect_bbox_violation_ids`, `collect_target_name_violation_ids`, `collect_target_location_violation_ids`), each `.pluck`/`.ids`-ing a relation and merging into a `Set` — no per-observation Ruby loop. `#violations` doesn't reuse any of this.

## Proposed design

Condense the four `collect_*_violation_ids` calls into one shared method that builds the violating-observation-id `Set` (the merge step `count_violations` already does), and have both `count_violations` and `#violations` call it:

1. Shared id-collecting method (e.g. `violating_observation_ids`) — the same four-scope merge `count_violations` already does, extracted so both callers use it.
2. `#violations` becomes relation-based: `Observation.where(id: violating_observation_ids)`, ordered by `sort_name` (needs a join to `names`).
3. `Projects::ViolationsController#index` paginates that relation with LIMIT/OFFSET instead of Ruby array slicing — still via the manual `PaginationData` approach from #5255, not `build_index_with_query`, so the session-clobbering bug that fix addressed doesn't come back. The relation-vs-array question and the "which pipeline touches session" question are independent; this only changes the former.
4. `violation_kinds_for` (which kinds apply, and why — needed for the "Details" column) only runs on the current page's observations, not the whole project.

## Why this is worth doing

- Bounds per-request cost to page size instead of the full project's observation count — the performance problem `count_violations`'s comment already flags for `#violations`.
- Removes the duplicate id-collecting logic between `count_violations` and `#violations`.

## Risks / things to verify

- Getting the `sort_name` ordering right via a `names` join (currently done by sorting `Violation` POROs in Ruby).
- Confirming the SQL-based existence check (the four `collect_*` scopes) agrees with the Ruby-based `violation_kinds_for` check for every observation — if they've drifted apart, unifying them either fixes a latent inconsistency or surfaces one that needs a closer look before shipping.

## Context

Split out from #5250/#5255 (`Projects::ViolationsController#index` pagination fix) — that PR fixed the session-clobbering and page-clamp-redirect bugs and added pagination over a plain Ruby `Array`, which works but doesn't get the performance benefit a relation-based rewrite would.

nimmolo:
## Front-end behavior changes

- Some projects will show a **new** violation: an observation with neither GPS nor an assigned Location, in a project with a location constraint, now correctly appears (previously invisible).
- Some projects will show **one fewer** violation: an observation whose Location is the project's constrained Location, but whose GPS falls outside that Location's bounding box, is now correctly excluded (previously a false positive).
- Both affect the violations page's row list and the count badge on the project's page, since both now share the same id set.
- Pagination page size/boundaries are unchanged for unaffected projects — only the slicing mechanism changed (SQL LIMIT/OFFSET instead of Ruby array indexing).

## Summary

Closes #5256.

`Project#violations` iterated every visible observation in Ruby and sorted the full result, even though the violations page only displays one page at a time. `Project#count_violations` already proved the existence check is SQL-expressible via four `collect_*_violation_ids` scopes merged into a Set — extracted that merge into `violating_observation_ids`, shared by `count_violations` and the new `violating_observations` (an ordered AR relation built from those ids).

`Projects::ViolationsController#index` now paginates `violating_observations` with LIMIT/OFFSET and only computes violation kinds for the current page's observations, via the new `violations_for`. Removed the `skip_bullet` override — the eager loading needed for it turned out to be sufficient, confirmed by the controller/view tests passing without it.

## Bug found and fixed along the way

Unifying `count_violations` and `#violations` onto one SQL id set surfaced two pre-existing drifts between the SQL bbox scopes and `Location#found_here?` (the Ruby check `violation_kinds_for` uses), confirmed by direct reproduction against a rolled-back transaction:

- An obs with neither GPS nor an assigned Location was invisible to the SQL scopes (one requires lat present, the other requires a location to join against), but `found_here?` flags it as a violation. Added `obs_missing_geoloc_and_location` to cover it.
- An obs whose Location is the project's Location was wrongly flagged when its GPS fell outside that Location's bbox — `found_here?`'s first branch (`obs.location == self`) exempts this regardless of GPS precision, but the SQL scopes had no equivalent exemption. Added the missing `location_id` exclusion to both existing scopes.

These drifts predate this PR (they were already live in `count_violations`'s badge count), but this PR extends their reach to the violations list, so fixing them here keeps the new unified path correct instead of inheriting a known gap.

## Changes

- `Project#violating_observation_ids` — shared Set-building method (the merge `count_violations` already did), now also used by `violating_observations`.
- `Project#violating_observations` — ordered, unpaginated AR relation of every violating observation.
- `Project#violations_for(observations)` — builds `Violation` structs for a given collection (a page, or the full set for `#violations`).
- `Project#violations` — now `violations_for(violating_observations)`, same public contract (`Array<Violation>`), used by existing tests.
- Two bbox scopes fixed to match `Location#found_here?`; one new scope added for the missing case.
- `Projects::ViolationsController#index` paginates via LIMIT/OFFSET instead of Ruby array slicing; `skip_bullet` removed.
- New tests: `violating_observations` matches `violations`, LIMIT/OFFSET pagination slices correctly, `violations_for` skips non-violations, and the two bbox parity fixes.

## Test plan

- [x] `bin/rails test test/models/project_test.rb`
- [x] `bundle exec rubocop` on every touched file

article: no

### DIFF
--- a/app/controllers/projects/violations_controller.rb
+++ b/app/controllers/projects/violations_controller.rb
@@ -10,21 +10,23 @@
 module Projects
   class ViolationsController < ApplicationController
     before_action :login_required
-    # Cannot figure out the eager loading here.
-    around_action :skip_bullet, if: -> { defined?(Bullet) }, only: [:index]
 
-    # @violations is a plain Ruby Array (Project#violations), not a
-    # Query/AR relation, so it's paginated directly with
-    # PaginationData rather than through build_index_with_query's
-    # Query-driven machinery -- see the class doc on PaginationData
-    # for this exact "without Query" usage.
+    # `@project.violating_observations` is an AR relation (not a
+    # Query), so it's paginated directly with PaginationData rather
+    # than through build_index_with_query's Query-driven machinery --
+    # see the class doc on PaginationData for this "without Query"
+    # usage. Unlike the array-slicing shown there, the current page
+    # is fetched with LIMIT/OFFSET, so only one page of observations
+    # gets its violation kinds computed.
     def index
       return unless find_project!
 
-      @violations = @project.violations
+      observations = @project.violating_observations
       @pagination_data = number_pagination_data
-      @pagination_data.num_total = @violations.length
-      @violations = @violations[@pagination_data.from_to] || []
+      @pagination_data.num_total = observations.count
+      page = observations.offset(@pagination_data.from).
+             limit(@pagination_data.num_per_page)
+      @violations = @project.violations_for(page)
       render_index_view
     end
 
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -317,23 +317,12 @@ class Project < AbstractModel # rubocop:disable Metrics/ClassLength
     member?(user)
   end
 
-  # SQL-based count over the four violation kinds (#4136). Each branch
-  # plucks ids of OFFENDING observations and merges them into a Set
-  # for dedup; total cost is O(violations) rather than the
-  # O(visible_observations) cost of the full Ruby iteration in
-  # `#violations`. Called from the project show page's
-  # `render_violations_button` (inlined from the former
-  # `Tabs::ProjectsHelper#violations_button`), so any per-project
-  # work multiplies by the number of projects rendered.
+  # Called from the project show page's `render_violations_button`
+  # (inlined from the former `Tabs::ProjectsHelper#violations_button`),
+  # so any per-project work multiplies by the number of projects
+  # rendered.
   def count_violations
-    return 0 unless constraints?
-
-    ids = Set.new
-    collect_date_violation_ids(ids)
-    collect_bbox_violation_ids(ids)
-    collect_target_name_violation_ids(ids)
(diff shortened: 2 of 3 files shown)


nimmolo:
Superseded by #5264 — collapses the same fix into a single OR-combined relation instead of the id-collection-then-refetch approach here, with better benchmarked results.

Issue #5242, opened 2026-08-27
Show Project Locations Error on nil Scientific Name


JoeCohen:
An ArgumentError occurred while GET </projects/407/locations> was processed by locations#indexExceptioncomparison of NilClass with String failedHostnamemushroomobserverBacktraceapp/controllers/concerns/projects/location_grouping.rb:12:in 'Enumerable#sort_by'
app/controllers/concerns/projects/location_grouping.rb:12:in 'Projects::LocationGrouping#build_grouped_locations'
app/controllers/projects/locations_controller.rb:13:in 'Projects::LocationsController#index'
app/controllers/application_controller.rb:170:in 'ApplicationController#catch_errors_and_log_request_stats'

PR #5308, opened 2026-09-03
Require `Location#name`/`scientific_name` at model and DB level

### LINKED ISSUE #5244: Validate Locations for presence of scientific_name
About 7 Locations were created without a scientific_name. This is possible because scientific_name presence is not currently validtated.
Those Locations caused Errors upon show Project Location. See #5242, #5243.

The intern diagnosed that the Locations had been created from the webserver console. I have confirmed that this is possible. https://mushroomobserver.slack.com/archives/D2AAX55NK/p1787874124434329

The validation would prevent that.

nimmolo:
## Summary

Fixes #5244: `Location#name`/`Location#scientific_name` are both DB-nullable with no presence validation on `scientific_name`, which let ~7 Locations get created via console access with a nil `scientific_name` (per the issue: confirmed possible from a webserver console session). Those rows caused crashes elsewhere (#5242, fixed in #5243 for the single-nil case).

## Fix

Two layers, since the issue's diagnosis is that these rows came from direct console access, which can call `.save(validate: false)` or write columns directly, bypassing Model validations entirely -- posted this reasoning on #5244 before starting the migration.

- **Model**: adds `validate_scientific_name` (parallel to the existing `validate_name`), called from `check_requirements`. Also hardens `validate_name` itself: `name.empty?` would raise `NoMethodError` on a nil `name` (should have been `name.to_s.empty?`, matching the pattern the new `validate_scientific_name` uses).
- **DB migration**: `NOT NULL` on both `locations.name` and `locations.scientific_name` -- the layer a console session can't route around. Confirmed locally that `update_columns` (which skips validations) is blocked by the DB constraint once applied. No `default:` backfill passed deliberately -- if a null row exists when this runs, the migration should fail loudly rather than silently paper over bad data with a placeholder.

A fresh production DB checkpoint (pulled mid-review) shows zero Locations with a nil or blank `name` or `scientific_name`, so the migration should apply cleanly with no backfill needed -- worth a final `NULL` count check against production directly right before running it, since a checkpoint is a point-in-time snapshot.

## Locale tag

`Location#validate_scientific_name`'s error message needs a `scientific_name` locale tag (used as `field: :scientific_name` interpolation, matching the existing `name` tag). Added to `en.txt` and regenerated `en.yml` via `rails lang:update` (gitignored, not committed) -- surfaced by `test_construct_location_empty_form`'s missing-tags check.

## Test call sites

The new validation broke a number of existing test call sites that create a `Location` directly with just `name:`, bypassing the `display_name=` setter (which normally derives both `name` and `scientific_name` together from one input -- see `Location#display_name=`). Added `scientific_name:` to each, using `Location.reverse_name(name)` to derive the correct value:

- `test/models/location_test.rb` (2 sites)
- `test/classes/autocomplete_for_region_test.rb`, `test/classes/inat_obs_test.rb`, `test/models/abstract_model_test.rb`, `test/models/field_slip_extract_test.rb`
- `test/system/observation_form_system_test.rb` (the shared `UNIVERSITY_PARK` constant, plus two `pasadena` sites and one multi-line `alpha`/`beta` pair an earlier automated pass missed -- single-line regex, these span two lines with string concatenation)
- `test/controllers/locations_controller_test.rb` (3 sites), `test/jobs/inat_import_job_test.rb` (8 sites, one shared name string)

Confirmed the app's creation paths (`ObservationsController`/`Locationable`) already call `display_name=` after `Location.new`, so no production code needed a change. Also confirmed one `Location.new(name: "Earth")` in `test/components/map_test.rb` is not saved (an in-memory probe for a component test), so it doesn't hit the validation and needs no fix.

## Verified

- `bin/rails test test/models/location_test.rb` -- 38/38 pass
- `bin/rails test test/classes/autocomplete_for_region_test.rb` -- 5/5 pass
- `bin/rails test test/classes/inat_obs_test.rb` -- 25/25 pass
- `bin/rails test test/models/abstract_model_test.rb` -- 33/33 pass
- `bin/rails test test/models/field_slip_extract_test.rb` -- 43/43 pass
- `bin/rails test test/system/project_violations_target_location_modal_system_test.rb` -- 1/1 pass
- `bin/rails test test/components/map_test.rb` -- 26/26 pass
- `bin/rails test test/jobs/inat_import_job_test.rb` -- 64/64 pass
- `bin/rails test test/controllers/locations_controller_test.rb` -- 69/69 pass
- `bin/rails test test/system/observation_form_system_test.rb` -- 25/27 pass; the 2 failures (`Ferrum::PendingConnectionsError` on one test, and `test_zero_latitude_triggers_locality_lookup` failing on a DOM assertion) are unrelated to this change -- confirmed via `log/test.log`, which shows zero server errors and nothing `scientific_name`/validation-related near either failure. `test_zero_latitude_triggers_locality_lookup` is documented as flaky pre-existing (Nathan's PR #5239 note); being fixed separately, on a different branch.
- `bundle exec rubocop` on every touched Ruby file -- no offenses.

article: no

### DIFF
--- a/app/controllers/concerns/projects/location_grouping.rb
+++ b/app/controllers/concerns/projects/location_grouping.rb
@@ -20,11 +20,8 @@ module Projects
       [groups, ungrouped]
     end
 
-    # `scientific_name` isn't guaranteed on legacy Location rows -- fall
-    # back to `name` so a nil doesn't blow up `sort_by`'s comparison
-    # against other locations' non-nil scientific names.
     def sort_key(loc)
-      loc.scientific_name || loc.name
+      loc.scientific_name
     end
 
     def sorted_targets(project)
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -567,10 +567,6 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     UNDERSTOOD_CONTINENTS[a_continent]
   end
 
-  def self.countries_by_count
-    CountryCounter.new.countries_by_count
-  end
-
   def self.location_name_cache
     Rails.cache.fetch(:location_names, expires_in: 15.minutes) do
       (Location.pluck(:name) + Observation.pluck(:where) +
@@ -897,8 +893,8 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
   # merge (Descriptions::Merges#perform_merge) would leave it.
   def merge_primary_descriptions(old_loc)
     return unless description && old_loc.description
-    return unless description.source_type == :public
-    return unless old_loc.description.source_type == :public
+    return unless description.source_type == "public"
+    return unless old_loc.description.source_type == "public"
     return unless description.mergeable_notes?(old_loc.description)
 
     description.merge_notes_from(old_loc.description)
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -984,6 +984,7 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     validate_elevation
     validate_user
     validate_name
+    validate_scientific_name
   end
 
   def check_hidden
@@ -1031,8 +1032,22 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
   def validate_name
     if name.to_s.size > 1024
       errors.add(:name, :validate_location_name_too_long)
-    elsif name.empty?
+    elsif name.to_s.empty?
       errors.add(:name, :validate_missing, field: :name)
+    end
+  end
+
+  # #5244: ~7 Locations were created with a nil scientific_name via
+  # direct console access, which bypasses this validation -- see the
+  # NOT NULL DB constraint (added in the same migration as this
+  # validation) for the enforcement that reaches that path too.
+  def validate_scientific_name
+    if scientific_name.to_s.size > 1024
+      errors.add(:scientific_name, :validate_too_long,
+                 field: :scientific_name, max: 1024)
+    elsif scientific_name.to_s.empty?
+      errors.add(:scientific_name, :validate_missing,
+                 field: :scientific_name)
     end
   end
 end
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -726,15 +726,6 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     result
   end
 
-  def self.dubious_country?(name)
-    !understood_country?(country(name))
-  end
-
-  def self.fix_country(name)
-    c = country(name)
-    name[0..(name.rindex(c) - 1)] + COUNTRY_FIXES[c]
-  end
-
   def self.find_by_name_with_wildcards(str)
     find_using_wildcards("name", str)
   end
@@ -898,12 +889,20 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     move_remaining_descriptions(old_loc)
   end
 
+  # A conflict (both sides have a non-blank value for the same note
+  # field) is left alone -- old_loc's description survives the
+  # location merge as a secondary description (see
+  # `move_remaining_descriptions`) for manual resolution via the
+  # descriptions/merges UI, same as a controller-driven description
+  # merge (Descriptions::Merges#perform_merge) would leave it.
   def merge_primary_descriptions(old_loc)
     return unless description && old_loc.description
     return unless description.source_type == :public
     return unless old_loc.description.source_type == :public
+    return unless description.mergeable_notes?(old_loc.description)
 
-    description.merge(old_loc.description)
+    description.merge_notes_from(old_loc.description)
+    old_loc.description.destroy
   end
 
   def move_remaining_descriptions(old_loc)
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -137,6 +137,7 @@ class Location < AbstractModel # rubocop:disable Metrics/ClassLength
     "hidden"
   )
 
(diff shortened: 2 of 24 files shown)


app/models/location.rb
```diff
@@ -1031,8 +1032,22 @@ def validate_user
...
-    elsif name.empty?
+    elsif name.to_s.empty?
       errors.add(:name, :validate_missing, field: :name)
     end
   end
+
+  # #5244: ~7 Locations were created with a nil scientific_name via
+  # direct console access, which bypasses this validation -- see the
+  # NOT NULL DB constraint (added in the same migration as this
+  # validation) for the enforcement that reaches that path too.
```
JoeCohen: The intern loves these comments about the way things used to be.  
I don't (and often remove them from  my PRs).
What do you think?

JoeCohen (changes requested):
Why not have the model add a `scientific_name` if it's not provided, but the `name` is provided?  And vice versa? 

When do the validations run? I'm able to do this from the console:
```ruby
mushroom-observer(dev)> loc = Location.create(user_id: 4468, north: 1, south: -1, west: -1, east: 1, name: "Unscientific, USA")
  TRANSACTION (0.1ms)  BEGIN
  User Load (0.6ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 4468 LIMIT 1
  TRANSACTION (0.1ms)  ROLLBACK
=> 
#<Location:0x00000001294a4118
...
mushroom-observer(dev)> loc
=> 
#<Location:0x00000001294a4118
 id: nil,
...
 name: "Unscientific, USA",
 scientific_name: nil,
...
 center_lng: nil>
```
It would be nice to get notice of he problem (or better, to automatically add he scientific_name) instead of waiting for a failed `save`.

The are some newly added uncovered lines. 
This would also be a good opportunity to cover all the missed lines in the model.

app/models/location.rb
nimmolo: I also do not like the narrative comments. Particularly the ones that are intra-PR, there are whole tiny dramas in some of them about how we got to this code, through various commits of the PR.

nimmolo:
@JoeCohen Good idea - adding that now, the validator will fill the missing format name.

**On validations from the console:**
- `Location.create(...)`, `.new.save`, `.update(...)` — do run validations by default, from the console exactly as from app code. Joe's own console example demonstrates this: `Location.create(...)` with no `scientific_name` came back with `id: nil` — it failed validation and was never persisted. It just didn't raise or print an obvious error, so it's easy to miss.
- `update_column`/`update_columns`, `update_all`, `insert_all`/`upsert_all`, raw SQL, or an explicit `save(validate: false)` — these skip validations, whether run from the console or from app code. This is the actual mechanism for how the ~7 bad rows (nil `scientific_name`) got into the DB — some path other than a validated `create`/`save`, which is why the `DB NOT NULL` constraint is the layer that closes it for good; the model validation alone wouldn't have done.

app/controllers/concerns/projects/location_grouping.rb
```diff
@@ -20,11 +20,8 @@ def build_grouped_locations(project)
       [groups, ungrouped]
     end
 
-    # `scientific_name` isn't guaranteed on legacy Location rows -- fall
-    # back to `name` so a nil doesn't blow up `sort_by`'s comparison
-    # against other locations' non-nil scientific names.
     def sort_key(loc)
-      loc.scientific_name || loc.name
+      loc.scientific_name
     end
```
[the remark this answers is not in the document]
nimmolo: This was a deliberate simplification, not an oversight. This PR adds both a model validation and a DB `NOT NULL` constraint on `locations.scientific_name` (see the PR description) -- together they make a nil `scientific_name` impossible for any row that reaches this code path, including the deploy window: this PR ships the model change and the migration in the same commit, and the migration fails loudly (no default backfill) if any row is already null when it runs, rather than silently leaving one behind. The regression test that used to force a nil `scientific_name` via `update_column` (bypassing validations) to exercise this fallback was removed in this same PR, since the DB constraint now rejects that write outright.

nimmolo:
Addressed both points:

**Auto-derivation** -- added a `before_validation :derive_missing_name_field` callback on `Location`: if `name` is present and `scientific_name` is blank, it derives `scientific_name` via `Location.reverse_name(name)` (and vice versa). Your console example now derives `scientific_name` automatically instead of failing validation silently. If both are supplied, both are kept as given -- the callback only fills in whichever one is missing.

**Coverage** -- went through every previously-uncovered line in `location.rb` and either added a test or removed the code:

- Added tests for `validate_user`, `title_display_name`, `textile_name`, `clean_name(leave_stars: true)`, `understood_continents`, `understood_states`, `countries_by_count`, and both `validate_name`/`validate_scientific_name` error branches.
- Deleted `Location.fix_country`/`dubious_country?` -- `fix_country` referenced an undefined `COUNTRY_FIXES` constant (would raise `NameError` if called), and neither method has a caller anywhere in the app.
- Found a bug while chasing the remaining uncovered lines: `merge_primary_descriptions` called `description.merge(old_loc.description)`, but no `merge` method exists on `Description`/`LocationDescription` anywhere in the codebase -- would raise `NoMethodError` the first time two locations being merged both had a public `LocationDescription`. Fixed by extracting the reusable core of `Descriptions::Merges#perform_merge` (the conflict check + field/author/editor copy) onto `Description` as `mergeable_notes?`/`merge_notes_from`, so the controller concern and `Location#merge_primary_descriptions` share the same code instead of one hand-copying the other's logic. Added tests for both the successful-merge and conflict cases.

PR #5309, opened 2026-09-03
Fix flaky `test_zero_latitude_triggers_locality_lookup`

nimmolo:
## Summary

Fixes the flaky `test_zero_latitude_triggers_locality_lookup` system test, previously noted as flaky in PR #5239.

## Root cause

The test dispatches synthetic `lat=0`/`lng=0.5` input events, expecting the autocompleter to swap into `location_containing` mode. The only fixture location whose bounding box geographically contains that point is `unknown_location` ("Earth"), a whole-globe box — and `Autocomplete::ForLocationContaining` deliberately rejects overly broad boxes as "vague" (`Mappable::BoxMethods#vague?`, `calculate_area > MO.obs_location_max_area` = 24,000 km²). With zero non-vague matches, the client falls back to `location_google`, and the test's `location_containing` assertion fails.

This is not the timing/race bug the earlier `wait_for_autocompleter_request_params` helper (PR #5239) was written for — that helper resolves `request_params` state correctly, confirmed via live browser debugging. The gap is that the server-side lookup finds nothing to match against, so the helper has nothing to wait for.

## Fix

Adds a purpose-built, appropriately sized `Location` fixture (`north: 0.7, south: -0.3, east: 1.0, west: 0.0`, 12,368 km², under the vague-area threshold) containing the test's target point, created at the start of the test and destroyed in an `ensure` block.

## Verified

- `bin/rails test test/system/observation_form_system_test.rb -n test_zero_latitude_triggers_locality_lookup` — passes, run twice to confirm stability (1 tests, 6 assertions, 0 failures, 0 errors, 0 skips both times).
- `bundle exec rubocop test/system/observation_form_system_test.rb` — no offenses.

article: no

### DIFF
--- a/test/system/observation_form_system_test.rb
+++ b/test/system/observation_form_system_test.rb
@@ -1280,13 +1280,28 @@ class ObservationFormSystemTest < ApplicationSystemTestCase
     university_park.destroy
   end
 
-  # Bug fix: a latitude or longitude of exactly 0 (equator or
-  # prime meridian) used to be treated as "no coordinate" by a falsy
-  # check and silently rejected. Under the old bug, a latitude of 0
-  # would keep the autocompleter in plain "location" mode; the fixed
-  # code swaps it into "location_containing" like any other valid
-  # point, and carries the exact (0-valued) params along.
+  # Bug fix: a latitude or longitude of 0 (equator or prime meridian)
+  # used to be treated as "no coordinate" by a falsy check and
+  # silently rejected. Under the old bug, a latitude of 0 would keep
+  # the autocompleter in plain "location" mode; the fixed code swaps
+  # it into "location_containing" like any other valid point, and
+  # carries the 0-valued params along.
+  #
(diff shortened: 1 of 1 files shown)


JoeCohen:
In e684dd28b7ae2e7c499019638cc6d3c5e28ff5a2 in https://github.com/MushroomObserver/mushroom-observer/pull/4921, my intern did the same fix, but with slightly different coordinates as NULL_ISLAND. (I changed it to NULL_ISLAND_BOX.)

nimmolo:
@JoeCohen - good catch. I checked with the intern and it says although the fixes are very similar, this is the only one that will survive #5308 and has a better cleanup step, so i'm going to push this one onto your branch to eliminate that merge conflict.

JoeCohen:
> @JoeCohen - good catch. I checked with the intern and it says although the fixes are very similar, this is the only one that will survive #5308 and has a better cleanup step, so i'm going to push this one onto your branch to eliminate that merge conflict.

Thanks for the push.

PR #5353, opened 2026-09-09
Normalize `inat_username` to lowercase; match logged-in iNat user case-insensitively

mo-nathan:
## Summary

A user who typed their iNat username as "Lothlin" got the wrong-user error — "lothlin is logged into iNat. To import iNat observations of Lothlin make sure that Lothlin is logged into iNat on your device" — even though it is the same account. iNat logins are lowercase-only (capitals are rejected at signup), so the entered value and the login iNat's `users/me` returns can differ only by case, and `InatImportJob#ensure_not_importing_others` compared them with `==`.

## Changes

- `User#inat_username=` and `InatImport#inat_username=` strip and downcase, so the stored value always matches iNat's lowercase form however the user typed it.
- `InatImportJob#ensure_not_importing_others` compares with `casecmp?` — records saved before normalization can still hold mixed case. (The comparison was also extracted to `fetch_inat_logged_in_user` for a metrics limit.)
- `Validators#not_importing_all_anothers?` (the super-importer warning) compares case-insensitively for the same reason.
- The estimator queries send the downcased login to iNat's search as `user_login`, the form iNat stores.

## Test plan

- [x] `bin/rails test test/jobs/inat_import_job_test.rb test/models/inat_import_test.rb test/models/user_test.rb test/controllers/inat_imports_controller_test.rb` (250 tests)
- [x] New job test: an `InatImport` row holding "Capitalized" imports successfully when iNat reports the lowercase login, and the username persisted onto the user is lowercase
- [x] Manual: enter your iNat username with a capital letter on the import form; the import should proceed instead of showing the wrong-user warning

article: yes
Fix iNat import rejecting usernames typed with capitals

### DIFF
--- a/app/controllers/inat_imports_controller/estimators.rb
+++ b/app/controllers/inat_imports_controller/estimators.rb
@@ -90,7 +90,10 @@ module InatImportsController::Estimators
     args = listing_url? ? url_query_args : {}
     args[:only_id] = true
     args[:id] = params[:inat_ids] if listing_ids?
-    args[:user_login] = params[:inat_username]&.strip unless import_others?
+    unless import_others?
+      args[:user_login] =
+        params[:inat_username]&.strip&.downcase
+    end
     args
   end
 
@@ -100,7 +103,10 @@ module InatImportsController::Estimators
     args[:only_id] = true
     args[:taxon_id] ||= IMPORTABLE_TAXON_IDS_ARG
     args[:id] = params[:inat_ids] if listing_ids?
-    args[:user_login] = params[:inat_username]&.strip unless import_others?
+    unless import_others?
+      args[:user_login] =
+        params[:inat_username]&.strip&.downcase
+    end
     args
   end
 
@@ -142,7 +148,7 @@ module InatImportsController::Estimators
     if import_others?
       LICENSED_FILTER
     else
-      { user_login: params[:inat_username]&.strip }
+      { user_login: params[:inat_username]&.strip&.downcase }
     end
   end
 
--- a/app/controllers/inat_imports_controller/estimators.rb
+++ b/app/controllers/inat_imports_controller/estimators.rb
@@ -90,10 +90,7 @@ module InatImportsController::Estimators
     args = listing_url? ? url_query_args : {}
     args[:only_id] = true
     args[:id] = params[:inat_ids] if listing_ids?
-    unless import_others?
-      args[:user_login] =
-        params[:inat_username]&.strip&.downcase
-    end
+    args[:user_login] = normalized_inat_username unless import_others?
     args
   end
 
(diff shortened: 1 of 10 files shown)


app/controllers/inat_imports_controller/estimators.rb
```diff
@@ -100,7 +103,10 @@ def after_taxon_query_args
     args[:only_id] = true
     args[:taxon_id] ||= IMPORTABLE_TAXON_IDS_ARG
     args[:id] = params[:inat_ids] if listing_ids?
-    args[:user_login] = params[:inat_username]&.strip unless import_others?
+    unless import_others?
+      args[:user_login] =
+        params[:inat_username]&.strip&.downcase
+    end
```
JoeCohen: Duplicates lines 93-96. Can/should it be DRYed?

app/jobs/inat_import_job.rb
```diff
@@ -81,17 +81,11 @@ def authenticate
...
-      raise("iNat API user request failed: #{e.message}")
-    end
-
-    inat_logged_in_user = JSON.parse(response.body)["results"].first["login"]
+    inat_logged_in_user = fetch_inat_logged_in_user
     log("inat_logged_in_user: #{inat_logged_in_user}")
-    return if inat_logged_in_user == inat_username
+    # casecmp: records saved before inat_username was normalized to
+    # lowercase can still hold the login with different case.
+    return if inat_logged_in_user.casecmp?(inat_username.to_s)
```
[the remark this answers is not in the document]
JoeCohen: Copilot might be over-defensive. IDK if iNat will ever return a nil/blank login.
If the line is revised, its alignment should be fixed. (Copilot seems to drop leading spaces a lot, which means ha we can't just click "Commit suggestions".)

app/models/user.rb
```diff
@@ -287,6 +287,12 @@ class User < AbstractModel # rubocop:disable Metrics/ClassLength
   # go through +change_password+.)
   before_create :crypt_password
 
+  # iNat logins are canonically lowercase; normalize so comparisons
+  # against values from iNat match however the user typed it.
+  def inat_username=(val)
+    super(val.is_a?(String) ? val.strip.downcase : val)
+  end
```
JoeCohen: Is this DRYable? (Duplicates app/models/inat_import.rb lines 76-81.)
Also I am starting to avoid the ternary operator in favor of if/else/end, as the latter results in better tests and coverage of both conditions.

app/jobs/inat_import_job.rb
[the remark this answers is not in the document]
mo-nathan: Yeah, I generally have CC review Copilot's comments and do the implmentation for reasons like this.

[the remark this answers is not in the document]
mo-nathan: Fixed in cfbc1a7094 - to_s on both sides keeps the comparison nil-safe (and the line is aligned properly, done by hand rather than committing the suggestion). Whether iNat can return a nil login is unknown; the guard costs one method call.

app/controllers/inat_imports_controller/estimators.rb
mo-nathan: Fixed in cfbc1a7094 - the three reads share a normalized_inat_username helper, which also let the two unless-blocks collapse back to one-liners.

app/models/user.rb
mo-nathan: Fixed in cfbc1a7094 - the setter moves to a NormalizesInatUsername concern included by both models, written if/else instead of a ternary per your preference.

Issue #5345, opened 2026-09-08
Scope a `Kamal`-based production deploy (replace `script/deploy.sh`)


nimmolo:
Umbrella issue for moving production off `script/deploy.sh`'s bare-metal stop/deploy/restart onto Docker + Kamal, plus the groundwork that move requires. Tracks a series of PRs, not one cutover -- the Dockerfile and a Kamal spike are two of many parts below.

## Sequence

- [x] Multi-stage production `Dockerfile` (#5355)
- [x] Fix stale MySQL-on-app-box docs in `README_PRODUCTION_INSTALL` (#5356)
- [ ] Add `strong_migrations` gem; grandfather existing migrations; adopt the expand/contract migration convention and a maintenance-window PR marker (independent of Kamal -- do this first)
- [ ] Inventory and classify production's `public/` directory (tracked / generated / operational / ad-hoc); land a persistent-volume plan for the generated and ad-hoc-drop classes
- [ ] Add Kamal + a minimal `config/deploy.yml`; deploy to a throwaway droplet as the first disposable environment (wires in the `public/` volumes above, `db:migrate` as a pre-deploy hook, image registry + secrets)
- [ ] Build the front routing tier (nginx or HAProxy) with cookie-routing groundwork and the maintenance-mode mechanism, replacing `deploy.sh`'s maintenance page
- [ ] Cut DNS over to the new box; run in parallel first; decide whether to keep or decommission the old box
- [ ] Split Solid Queue: a co-located file-local-queue worker role stays on the web droplet, everything else moves to a separate droplet
- [ ] Route anonymous (not-logged-in) traffic to a separate droplet via the front tier
- [ ] (Later, decoupled) MySQL version bump, rehearsed against a disposable environment, cut over via replicate-and-promote or a maintenance window
- [ ] (Later, decoupled) Ubuntu/Ruby base image bump -- expected to fall out of containerization rather than needing a separate cutover

## Decided architecture

**MySQL**: already on a separate droplet; the app container reaches it over the network. No placement change, no data migration.

[...]

PR #5356, opened 2026-09-09
Fix `README_PRODUCTION_INSTALL`'s stale MySQL-on-app-box assumption

nimmolo:
Fixes the stale MySQL-on-the-app-box assumption in `README_PRODUCTION_INSTALL`, flagged in #5345's comments (Nathan): production MySQL runs on a separate droplet, not the app box.

## Summary

Changes:
- Dropped `mysql-server` from the app-box package install; kept `mysql-client`/`libmysqlclient-dev` (CLI tools + gem compilation only), with a note on why.
- Labeled the remote-database section as the current production shape.
- Dropped the `mysqld` process check from the post-install sanity checklist (it runs on the separate DB droplet, not here).

## Test plan

- [x] Documentation-only change, no code touched.

article: no

### DIFF
--- a/README_PRODUCTION_INSTALL
+++ b/README_PRODUCTION_INSTALL
@@ -433,9 +433,9 @@ mrtg> crontab -e
 root> iptables-save
 
 # Make sure all the critical processes started automatically.
-# Look for sshd, nginx, puma, postfix. mysqld runs on the separate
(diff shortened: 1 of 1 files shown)


mo-nathan (approved):
I'd accept Copilot's comment, but otherwise this looks good.

PR #5368, opened 2026-09-12
Update RuboCop & dependencies

JoeCohen:
## Summary

Updates RuboCop to 1.90.0
Pins json to < 3.0 to fix incompatibility with active_support 7.2.3.2

article: no

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -209,6 +209,12 @@ group :test, :development do
   gem("rubocop-thread_safety", require: false)
 end
 
+# Pin json below 3.0:
+# activesupport 7.2's JSON encoder passes
+# legacy `quirks_mode:` option to JSON.generate,
+# json 3.0 rejects that with ArgumentError (instead of ignoring).
+gem("json", "< 3.0")
+
 group :test do
   # Use capybara to simulate user-browser interaction
   gem("capybara")
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -258,7 +258,7 @@ GEM
     jbuilder (2.15.1)
       actionview (>= 7.0.0)
       activesupport (>= 7.0.0)
-    json (3.0.2)
+    json (2.21.2)
     jwt (3.2.0)
       base64
     language_server-protocol (3.17.0.6)
@@ -656,6 +656,7 @@ DEPENDENCIES
   image_processing
   importmap-rails
   jbuilder
+  json (< 3.0)
   literal
   mimemagic
   mini_exiftool_vendored
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -258,7 +258,7 @@ GEM
     jbuilder (2.15.1)
       actionview (>= 7.0.0)
       activesupport (>= 7.0.0)
-    json (2.21.2)
+    json (3.0.2)
     jwt (3.2.0)
       base64
     language_server-protocol (3.17.0.6)
@@ -360,7 +360,7 @@ GEM
       version_gem (~> 1.1, >= 1.1.14)
     os (1.1.4)
     ostruct (0.6.3)
-    parallel (2.1.0)
+    parallel (2.2.0)
     parser (3.3.12.0)
       ast (~> 2.4.1)
       racc
@@ -463,7 +463,7 @@ GEM
     rqrcode_core (2.1.0)
     rsync (1.0.9)
     rtf (0.3.3)
-    rubocop (1.90.0)
+    rubocop (1.91.0)
       json (>= 2.3)
       language_server-protocol (~> 3.17.0.2)
       lint_roller (~> 1.1.0)


JoeCohen:
Intern's explanation and plan re test failures:

## Root cause

The most recent commit on this branch, `20debc8bff` ("Update RuboCop & dependencies"), bumped the `json` gem from **2.21.2 → 3.0.2** as a side effect of updating RuboCop (`rubocop`/`execjs` only require `json >= 2.3`/`>= 2`, so bundler was free to pick the newest 3.x). `json` 3.0.0 made a breaking change: **unknown keyword options now raise `ArgumentError` instead of being silently ignored.**

Rails' `activesupport` 7.2.3.2 (the version this app is pinned to) still calls:

```ruby
::JSON.generate(jsonified, quirks_mode: true, max_nesting: false)
```

in `ActiveSupport::JSON::Encoding::JSONGemEncoder` — `quirks_mode:` is a legacy no-op that json 2.x accepted and ignored. json 3.x rejects it, so **every call to `Hash#to_json` (and anything built on it) now raises** `ArgumentError: unknown keyword: quirks_mode`.

Reproduced directly, outside the test suite:

```ruby
gem "json", "2.21.2"; JSON.generate({a: 1}, quirks_mode: true, max_nesting: false)  # => works
gem "json", "3.0.2";  JSON.generate({a: 1}, quirks_mode: true, max_nesting: false)  # => ArgumentError: unknown keyword: quirks_mode
```

`Query#serialize` (`app/classes/query.rb:646`) just happens to be the first `to_json` call this particular test path hits (via `Query.get_record` → `serialize`), but this isn't a bug in `Query` or in the test — it's a dependency incompatibility. It's likely to surface anywhere else `to_json` runs too (other `QueryRecord` lookups, Turbo Stream/JSON responses, etc.), not just this one test.

## Plan

1. **Pin `json` below 3.0 in the `Gemfile`**, near the `rubocop*` gem group (`Gemfile:202-209`), with a comment explaining why it's pinned:

   ```ruby
   # Pinned below 3.0: activesupport 7.2's JSON encoder still passes the
   # legacy `quirks_mode:` option to JSON.generate, which json 3.0 now
   # rejects with ArgumentError instead of ignoring.
   gem("json", "< 3.0")
   ```

2. Run `bundle lock --update json` (or `bundle install`) to bring `json` back down to `2.21.2` in `Gemfile.lock`, while leaving `rubocop` at `1.91.0` and `parallel` at `2.2.0` (the versions actually intended by the dependency-update commit).
3. Verify the lockfile diff only changes `json` back to `2.21.2` and nothing else regresses.
4. Re-run the failing test: `bin/rails test test/views/controllers/species_lists/downloads/form_test.rb` — expect it to pass.
5. Since this bug isn't test-specific, do a broader sanity check: run `bin/rails test` (or at least a cross-section covering `Query`/`QueryRecord` and any JSON/Turbo Stream response specs) to confirm nothing else was silently broken by the json 3.0 bump before this fix landed.
6. Note for the commit/PR: this should probably be a follow-up fix to `20debc8bff` (e.g. "Pin json gem below 3.0 (incompatible with activesupport 7.2's JSON encoder)"), since the RuboCop bump itself is fine — only its incidental `json` transitive bump is the problem.

nimmolo:
Sounds like we should try deleting the `quirks_mode` call and see/fix what breaks, rather than pinning older `json`.

JoeCohen:
> Sounds like we should try deleting the `quirks_mode` call and see/fix what breaks, rather than pinning older `json`.

@nimmo: That call is in active_support.
We triigger it from the call to `to_json` in app/classes/query.rb`#serialiize:

```ruby
  # Serialize the query params, adding the model, for saving to a QueryRecord.
  # We use this column of QueryRecord to identify an existing query record that
  # matches current params, and sometimes to recompose a query from the string.
  # That's why the keys are sorted here before being serialized via `to_json` -
  # when matching, strings must match exactly.
  #
  # NOTE: QueryRecord[:description] is not a Rails-serialized column; we call
  # `to_json` here to serializate it ourselves. Using SQL to compare serialized
  # strings is more efficient than making it a Rails-serialized column and using
  # Ruby to compare hashes (in whatever order), because in a serialized column
  # you can't use SQL on the column value, you have to compare parsed instances.
  def serialize
    attributes.compact.sort.to_h.merge(model: model.name).to_json
  end
```

`to_json` is not a Rails method;  it's defined by the json gem.

But ActiveSupport overrides `Hash#to_json` (in active_support/core_ext/object/json.rb) to route through ActiveSupport::JSON.encode, which calls ActiveSupport::JSON::Encoding::JSONGemEncoder#encode, which calls JSON.generate(jsonified, quirks_mode: true, max_nesting: false). That quirks_mode: keyword is what json 3.0.2 rejects with ArgumentError.

JoeCohen:
@nimmolo: Did you intend to suggest monkey-patching ActiveSupport?

JoeCohen:
@nimmolo, @mo-nathan: The best way to deal with this may be:
- Do not upgrade `rubocop`; (we might need to pin it, and possibly, others temporarily)
- Move to Rails 8.1 ASAP
- Then we can unpin `rubocop` (and any others).

The gem conflict isn't fixed until 8.1. The intern says:
> [T]he conflict exists in Rails 8.0 (confirmed against v8.0.5.1, the latest 8.0.x release: still calls JSON.generate(jsonified, quirks_mode: true, max_nesting: false)). It's fixed in Rails 8.1 (confirmed against v8.1.3.1: calls JSON.generate(jsonified) with no keyword arguments) and on main.
>
> So: fixed in 8.1+, not fixed in any 8.0.x release. Upgrading this app from Rails 7.2 to Rails 8.0 alone would not resolve the conflict — only 8.1+ would.

JoeCohen:
PS: We call `to_json` in about 94 places in 31 files.

JoeCohen:
### Gems depending on `json` gem

Direct dependents (gems in Gemfile.lock that declare json themselves):

execjs
faraday
rubocop
Indirect dependents (pull in json transitively through one of those three):

| via execjs | via faraday	| via rubocop |
| -----| ----- |  ----- |
| autoprefixer-rails  |  faraday-follow_redirects | rubocop-capybara |
| bootstrap-sass (Gemfile) | google-cloud-env | rubocop-minitest |
| terser (Gemfile) | googleauth | rubocop-performance |
| | signet  |  rubocop-rails |
| | oauth2 (Gemfile)  |  rubocop-thread_safety |
| | google-apis-core 
| | google-apis-iamcredentials_v1	
| | google-apis-storage_v1	
| | google-cloud-core	
| | google-cloud-storage (Gemfile)	

That's 21 gems total (3 direct + 18 indirect).

json itself is not listed in the Gemfile — it's pulled in purely transitively. Of the chains above, the ones actually reachable from our own Gemfile are: bootstrap-sass, terser (→ execjs), oauth2, google-cloud-storage (→ faraday), and rubocop/rubocop-capybara/rubocop-minitest/rubocop-performance/rubocop-rails/rubocop-thread_safety (→ rubocop's own json dependency).

PR #5374, opened 2026-09-15
Use `subgen.` for MB search

### LINKED ISSUE #5360: Use subgen for MycoBank subgenus search
The Name, Nomenclature, MycoBank search uses "subg." when searching for a subgenus. The result is that the search returns no hits. It should instead use "subgen."

JoeCohen:
## Summary

Fixes MycoBank search for subgeneric Name.

## Manual Test plan

This test needs a Name at rank Subgenus for which MycoBank has an entry and which does not have an ICN ID in MO.
Example: [Tricholoma subg. Humidicutis](http://localhost:3000/names/107225)

- Go to [Tricholoma subg. Humidicutis](http://localhost:3000/names/107225)
- Click `MycoBank Search`
Expected: 
Goes to Basic searching on MycoBank, and displays a result. ✅
(In main, "No record matches the specified search terms.")

review: standard
article: yes
Fix MycoBank search for subgenus


app/classes/tab/name/mycobank_search.rb
```diff
@@ -12,6 +12,15 @@ def title
 
   def path
     "#{MYCOBANK_HOST}#{MYCOBANK_BASIC_SEARCH_PATH}" \
-      "/field/Taxon%20name/#{@name.sensu_stricto.gsub(" ", "%20")}"
+      "/field/Taxon%20name/#{taxon_name.gsub(" ", "%20")}"
```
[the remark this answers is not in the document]
JoeCohen: Fixed in [d9b7902](https://github.com/MushroomObserver/mushroom-observer/pull/5374/commits/d9b7902c4fed8378cfcafa30875ccc3b573a189b)

app/classes/tab/name/external_base.rb
```diff
@@ -23,4 +23,13 @@ def html_options
...
     @name
   end
+
+  private
+
+  # CGI.escape encodes a space as "+", which only decodes back to a
+  # space in form/query-string parsing -- not in a URL path segment.
+  # "%20" decodes correctly in both, so convert to that instead.
+  def url_encode(str)
+    CGI.escape(str).gsub("+", "%20")
```
nimmolo: `ERB::Util.url_encode` does this in one call, no `gsub` needed — it percent-encodes everything outside the RFC 3986 unreserved set (`A-Za-z0-9_-.~`), so a space becomes `%20` directly:

```ruby
ERB::Util.url_encode("Tricholoma subg. Humidicutis")
# => "Tricholoma%20subg.%20Humidicutis"
```

`CGI.escape` was written for `application/x-www-form-urlencoded` bodies (hence the `+`), which is the mismatch the `gsub("+", "%20")` is working around here. `ERB::Util.url_encode` is built for path/query segments and doesn't have that problem, so `external_base.rb`'s helper could simplify to:

```ruby
def url_encode(str)
  ERB::Util.url_encode(str)
end
```

`ERB` is stdlib and already required by Rails, so no new dependency.

JoeCohen: Fixed in c0bf56c8b6 - switched to `ERB::Util.url_encode`, no `gsub` needed.

Issue #5357, opened 2026-09-10
iNat importer should use /obs/# for of the Observation Show page


mo-nathan:
The iNat importer currently uses just `/#` which we don't allow non-logged in users to access.  This is causing issues with users trying to access MO links when they aren't logged in.  This is a especially problematic on some cell phones where new browser windows are configured to act like they aren't logged in even when another window on the phone is.  I believe this is an option "feature" at least on iPhones that in my view causes more headaches than actually safety.

Note that once the iNat importer is pushing out the correct link, we'll need to run a long slow backfill process to fix all the incorrect links in the Mushroom Observer URL values on iNat.

We need to do a more complete audit and wherever possible us the `/obs/#` form for the show observation pages even to the point of re-writing explicit URLs of the other forms (`/#` and `/observations/#`) for logged in users.

mo-nathan:
Site-wide audit. Mechanics first, verified against a dev server: all three forms route to `ObservationsController#show` (exempt from `login_required`); the logged-out 403 comes from `check_for_spider_block` (`application_controller/indexes.rb:85`, called at `observations_controller/show.rb:23`), which allows only the `permanent_observation_path` (`/obs/:id`) form. Anonymous: `/obs/578761` → 200, `/578761` → 403, `/observations/578761` → 403.

**Biggest single lever — an anonymous 301.** Change `check_for_spider_block` to redirect anonymous GET `/:id` and `/observations/:id` to `/obs/:id` (query string preserved) instead of 403ing. That fixes every wrong link already in the wild — all existing iNat field values, years of notification emails, forum posts, bookmarks — with no backfill. Spider exposure is unchanged since `/obs/` is already open to them. This also downgrades the iNat backfill from repair to hygiene.

**Emitters producing blocked forms today:**

1. iNat writeback — `inat/observation_importer.rb:238` writes `#{MO.http_domain}/#{id}` into field 5005. (`MO_URL_FIELD_VALUE_ID_RE` already accepts `/obs/`, so re-import dedup is unaffected by the change.)
2. Six mailers hand-build the bare form: `observation_links.rb`, `naming_tracker_mailer.rb`, `naming_observer_mailer.rb`, `observer_question_mailer.rb`, `occurrence_change_mailer.rb`, `inat_import_digest_mailer.rb` — the email-opened-logged-out case on phones.
[...]

PR #5398, opened 2026-09-21
Run the iNat URL backfill as a scheduled job

mo-nathan:
## Summary

Moves the iNat "Mushroom Observer URL" backfill (#5357) out of a hand-run script and into a Solid Queue recurring job, since the pass runs for about five weeks and cron is being retired for Ruby work.

Logged-out visitors get a 403 on the `/:id` and `/observations/:id` forms, so every value iNat holds should read `https://mushroomobserver.org/obs/<id>`. `Inat::URLBackfill` walks iNat's search for observations carrying field 5005, newest first, rewrites the values that need it, and resyncs the MO reflection behind each one from the payload the search already returned.

The resync is inline rather than left to the nightly batch (`InatReflectionBatchResyncJob`). An old iNat observation stops changing, so the batch's `updated_since` fetch stops returning it — and the sequence, taxon and image engines shipped long after most of these were imported. Walking the corpus for the URL rewrite is the occasion to reconcile them. Measured on production data, about 28% of a sample of 333 had drift worth writing, almost all vote reweights from the taxon engine.

### Shape

- **Time-boxed.** `:maintenance` has two threads and writes are paced at one per second, so a run stops after 30 minutes and the cursor resumes it the next night. `max_writes` is available as a second bound but the clock is what binds.
- **One run at a time.** `limits_concurrency(key:, to: 1, duration: 4.hours, on_conflict: :discard)` — a retry from /jobs or a console `perform_now` would otherwise walk the same cursor as the scheduled run and rewind it. `duration` has to outlast the longest run (it defaults to 3 minutes, which a time-boxed run outlives, leaving the limit unenforced), and `:discard` drops a run that would overlap rather than queueing it behind.
- **The cursor is a file**, `tmp/inat_url_backfill_cursor`, not a column. The pass is temporary; a column would ride along in every database dump taken while it runs and need a second migration to remove. Losing the file costs one read-only re-walk — the values there are already in `/obs/` form — not lost corrections.
- **Credential**: a long-lived iNat OAuth access token in `inat: { backfill_access_token: }` (production credentials) or `INAT_BACKFILL_ACCESS_TOKEN`. The API takes the 24h JWT minted from it; a pass this long outlives any one JWT, so a 401 mints a fresh one and retries once. With no token configured the job logs and returns, so the recurring entry is harmless before the token is installed and after the pass is retired.

### What reaches #alerts

Only errors, plus one deliberate nag:

- **Rejected writes** — one message per run carrying the count, not one per failure (distinct alert texts are distinct Slack posts). The values stay unrewritten and the next pass over them retries.
- **Engine alerts** from the inline resync, the same treatment `InatReflectionBatchResyncJob` gives them.
- **Auth failure** raises, so `ApplicationJob`'s `rescue_from` reports it and Solid Queue records the failure. This is the one that matters: with the token revoked, every nightly run would otherwise fail silently for weeks.
- **Pass complete** — a daily message asking for the job to be removed, repeating until someone does it. It renders red like every other notification; there is no informational styling.

Everything else goes to `job.log`, and only as events: a line per value rewritten or reflection changed, not per observation walked. Across 76,049 imported observations the untouched ones are the overwhelming majority, and logging them buries what needs checking.

### Also: "Sync now" alerts went nowhere

`Inat::ObservationResyncer` built a `ReflectionResync` and did not read its `alerts`, so a user-triggered resync that hit an unparseable taxon, an ambiguous locus pairing or a photo that failed to import produced a message that reached nothing — not `#alerts`, not `job.log`, not the flash. `InatObservationResyncJob` now drains them, matching the scheduled batch. Found while mapping the alert plumbing for this PR; it is a one-line fix that belongs with the rest of the resync alerting rather than in a separate PR.

## Test plan

Before the deploy, on the server: add the OAuth access token to the production credentials as `inat: { backfill_access_token: ... }`, and **restart the Solid Queue workers** — they cache decrypted credentials, so a Puma restart does not reach them.

For the reviewer:

1. With no token configured, `InatURLBackfillJob.perform_now` should return quietly and log "no iNat OAuth access token configured" to `job.log` — nothing to #alerts, no failed execution.
2. With a token, `InatURLBackfillJob.perform_now(time_limit: 1.minute, max_writes: 1)` from the console makes at most one write. Check `job.log` for one event line naming the iNat id, the MO id and what the resync changed, then confirm the value at the iNat observation reads `https://mushroomobserver.org/obs/<id>`.
3. `tmp/inat_url_backfill_cursor` should hold the id of the last observation handled. Re-running picks up below it; deleting it restarts from the newest.
4. Press "Sync now" on a reflection whose iNat photos include one MO cannot import, and confirm the alert lands in #alerts (it reached nothing before this PR).

The pass is done when the job alerts that it is complete; removing the `inat_url_backfill` entry from `config/recurring.yml`, this job, `Inat::URLBackfill` and the credential is the last step.

article: no


mo-nathan:
Parked, not ready for review.

Scope changed after this was opened: the pass now covers the ~15k read-only reflections only, not every iNat observation carrying field 5005. At that size it is roughly 14.2k writes — two hand-run passes at `--max-writes 8500`, not a five-week unattended drip — so the recurring job is not what the work needs right now, and the walk is better driven from MO's reflection list than from an iNat search that fetches ~160k observations to skip most of them.

Kept rather than closed because the iNat-driven walk returns for the full resync and reflection-building pass over the ~100k linked observations (imports, exports, manual links), which does want a long-running job.

The `Inat::ObservationResyncer` alert fix that rode along here is independent and has moved to #5399; this branch still carries a duplicate of it.

